PR #1, opened 2016-12-05
Initial setup and connecting to contentful

weerd:
This PR does some initial setup for the project as well as integrating the Contentful Delivery Client as a Service Provider in Laravel, so we can grab an instance of the client from the _container_ instead of creating new instances every time we make a request to the Contentful Content Delivery API.

This PR includes a quick route with a closure to make a request to grab the data for a `campaign` content type on contentful.

---

@DFurnes @deadlybutter 

cc: @angaither


DFurnes:
Lookin' good! 🚢

PR #2, opened 2016-12-06
Using Contentful Laravel package

weerd:
This PR switches the Contentful Laravel Service Provider from the custom one in the last PR to one provided by the [Contentful-Laravel](https://github.com/contentful/contentful-laravel) package provided by Contentful.

It also adds a nice little helper function to make calling the client easier, since the service provider by the package doesn't make it super helpful with how it retrieves the package instance.

Lastly, after running the `php artisan vendor:publish` command to add the package config file, the Laravel notification and pagination views were published as well and added to repo. Might be useful down the line... 🔮 

---
@DFurnes @deadlybutter 

cc: @angaither


DFurnes:
🚢 🚢 🚢

PR #3, opened 2016-12-06
Adding campaigns resource

weerd:
This PR adds some Campaign resource logic to retrieve content from Contentful and display both a listing of all campaigns, as well as a single campaign view.

It also adds some basic tests, but the tests currently reference the Contentful API instead of using Mocks, so that will be an update in an upcoming PR.

Another thing to note... Contentful only returns fields for a content type that actually has some value in the field. If the field is not filled out, then the field is not included in the JSON response. This may pose a little logistical problem with templating and expecting some values to be available when they are not. This is just something to be aware of, and we can definitely build our consuming client and displaying content in a way that accounts for this. Just an FYI! :tipping_hand_man: 

---
@DFurnes @deadlybutter 

cc: @angaither


app/Http/Controllers/CampaignController.php
```diff
@@ -0,0 +1,34 @@
...
+        return view('campaigns.index', ['campaigns' => $campaigns]);
+    }
+
+    /**
+     * Display the specified resource.
+     *
+     * @param  int  $slug
+     * @return \Illuminate\Http\Response
+     */
+    public function show($slug)
```
weerd: Probs but we're just prototyping for now, playing w/ Contentful. I think w/ backwards compatibility will happen once we have a 🚥  green light.

DFurnes:
Looks good to me! ⛵️

PR #4, opened 2016-12-08
Possible approach for requesting cover images

weerd:
This PR provides a possibly solution that might provide a path to how we handle retrieving different cropped sizes of specifically the Cover Images for a Campaign, but it could very likely be generalized to apply to other images retrieved from Contentful.

This is definitely up for discussion, but I wanted to experiment with an approach.

image

---
@DFurnes @deadlybutter 

cc: @angaither

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -7,23 +7,17 @@ use Contentful\Delivery\ImageOptions;
  */
 
 /**
- * Get cover image URL for a specified asset by the crop type.
+ * Get image URL for a specified asset by the crop type.
  *
  * @param  \Contentful\Deliver\Asset $asset
  * @param  string $crop
  * @return string
  */
-function get_cover_image_url($asset, $crop = 'large')
+function get_image_url($asset, $crop = 'landscape')
 {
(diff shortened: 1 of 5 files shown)


weerd:
One thing I tried to do with this too, is upload a 2x sized cover image to contentful (2880x1240; normally they're 1440x620), specifically because I wanted to play around with having the highest quality image on Contentful and using it's Images API to do the resizing when requesting.

app/helpers.php
```diff
@@ -1,5 +1,39 @@
...
+ *
+ * @param  \Contentful\Deliver\Asset $asset
+ * @param  string $crop
+ * @return string
+ */
+function get_cover_image_url($asset, $crop = 'large')
+{
+    $validCrops = ['large', 'square'];
+
+    if (!in_array($crop, $validCrops, true)) {
```
DFurnes: Could this read from the `$options` array below rather than a separate `$validCrops` whitelist?

```php
if (! array_key_exists($crop, $options)) {
  // ...
```

PR #5, opened 2016-12-08
Localization

DFurnes:
This pull request adds simple localization support to the Campaign class based on the Laravel locale.

🇺🇸 🇲🇽 🇧🇷


app/Models/Campaign.php
```diff
@@ -37,17 +37,19 @@ public static function findBySlug($slug)
...
-
-        if (! $campaign->count()) {
+        $campaigns = $client->getEntries($query);
+        if (! $campaigns->count()) {
             throw new ModelNotFoundException;
         }
 
-        return $campaign->offsetGet(0);
+        $campaign = $campaigns[0];
+        $campaign->setLocale(app()->getLocale());
```
weerd: maybe combine with the return statement?

```php
return $campaign->setLocale(app()->getLocale());
```

or is that less readable?

app/Models/Campaign.php
```diff
@@ -37,17 +37,19 @@ public static function findBySlug($slug)
...
-        $campaign = $client->getEntries($query);
-
-        if (! $campaign->count()) {
+        $campaigns = $client->getEntries($query);
+        if (! $campaigns->count()) {
             throw new ModelNotFoundException;
         }
 
-        return $campaign->offsetGet(0);
+        $campaign = $campaigns[0];
```
weerd: `$campaign = $campaign->offsetGet(0);` thoughts on using the method on the Contentful class?

app/Models/Campaign.php
DFurnes: Could definitely do that! No strong preferences either way.

PR #8, opened 2016-12-09
Prototype: Snapshots

DFurnes:
This pull request is a lil' attempt at reproducible test results for the campaign repository – when running as part of a test suite, it short-circuits and uses a preloaded JSON file from the `tests/snapshots` directory rather than requesting the data from Contentful.

This would ensure that changing an entry via the web interface wouldn't break the test results.

🌵🚦


DFurnes:
Yeah, no idea! Definitely something to think about and keep an eye on. 👁

PR #9, opened 2016-12-12
Adding Campaign entity class to handle Contentful response and improve it.

weerd:
This PR adds a new `Campaign` entity class to help us work with the class instances from Contentful's API PHP package a lot better. 

One issue we ran into with the provided PHP package and inherently with the API response from Contentful is that if a field in a content type is _not_ filled out, then that field is _not_ included in the response. This means that we can't always know what is/isn't present in a Campaign response, which would result in some ugly templates with a lot of `isset()` calls. A helper function to deal with this was also considered, but wouldn't address some of the other issues with getting content from the different instances provided. Anyhoo, we decided it best to provide a `Campaign` entity class where we can dynamically set the provided property and the empty one automatically set to null. It also handles extracting data from the `DynamicEntry` data type, making template easier by just dynamically requesting properties off the `Campaign` entity, like `$campaign->title`.

**NOTE**: Still working on adding tests, but wanted to put this up to get some eyes on it!

What template shows for now:
image

Also, @DFurnes I didn't add an extraction method for the `Asset` instance which contains media like the cover image. We can, but I wasn't sure, since in the template we want specific sizes, and that's done via that helper method I set up to grab `landscape` or `square` urls for now. Alternative is to provide an associative array with the generated urls:

```php
[
 'landscape' => 'cdn.contentful.com/asdadas/asdasdasd?landscape',
 'square' => 'cdn.contentful.com/asdadas/asdasdasd?square',
]
```
Only caveat is that we won't always know what kind of crop a requested image needs... feels like that's more determined by the template and where that image is placed.

---
@DFurnes @deadlybutter 

cc: @angaither


app/Entities/Campaign.php
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * Create instance of the Campaign class.
+     *
+     * @param \Contentful\Delivery\DynamicEntry $entity
+     */
+    public function __construct(DynamicEntry $entity)
+    {
+        foreach(array_keys(get_object_vars($this)) as $property) {
+            $this->set($property, $entity->{'get'.ucwords($property)}());
```
DFurnes: Will this still call the method (for example, `$entity->getSolutionStatement`) if the field isn't set on the DynamicEntry? If so, do we need to wrap this line in a try/catch?

deadlybutter:
looks sweet!

app/Entities/Campaign.php
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * Dynamically set the campaign's attributes.
+     *
+     * @param  string $property
+     * @param  mixed $data
+     * @return void
+     */
+    public function set($property, $data)
+    {
+        return $this->$property = $this->getByDataType($data);
```
DFurnes: I'm still a little confused why we're exposing a public `set` API for changing the contents of a campaign object (as opposed to having the `__get` handler reach into the DynamicEntry for each requested property). I remember having a conversation IRL about this but forget the outcome.

app/Entities/Campaign.php
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * Get array of string data from supplied array of mixed data types.
+     *
+     * @param  array $array
+     * @return array
+     */
+    private function getArrayData($array)
+    {
+        return array_map([$this, 'getByDataType'], $array);
```
DFurnes: This is clever! 👏

app/Entities/Campaign.php
```diff
@@ -0,0 +1,207 @@
...
+
+namespace App\Entities;
+
+use Contentful\Delivery\Asset;
+use Contentful\Delivery\DynamicEntry;
+
+/**
+ * The Campaign entity.
+ *
+ * @property  string $title
```
DFurnes: We probably don't need `@property` docblocks since these are concrete fields.

weerd: Great! Those were a pain in the 🍑 to add for each, haha. Should I leave the `@property` listing at the top of the class?

```php
...
* @property  string $title
 * @property  string $slug
 * @property  string $callToAction
 * @property  string $active
...
```

app/Entities/Campaign.php
weerd: It will but I tested it and that returns `null` so it's technically ok. Thoughts?

The reason why this was crappy in the template was that you can't pass a method call to `isset($campaign->getSomePropertyHere())`, so the templates would have required wrapping everything in some nasty logic for each item.

However, when called in this fashion, if it isn't set, it will return `null`, which is what the property is set to by default in this entity class anyhoo :) Hopefully I explained that in a way that makes sense 😬

app/Entities/Campaign.php
weerd: We could technically set it to `private`; sorry that's my mistake from when I was playing around with stuff. We tried the experiment IRL with `__get()` but we also needed to address the nested items that would then have to be filtered. We ended up agreeing that it wasn't as useful 😉

app/Entities/Campaign.php
weerd: 😄 💡

app/Entities/Campaign.php
weerd: I'll double check while writing tests just to make sure 💃

app/Entities/Campaign.php
DFurnes: Ah right, bummer.

PR #11, opened 2016-12-15
Update for translation bug

weerd:
This PR by @DFurnes restructures how we thing of the `Campaign` entity to now include a parent `Entity` class that handles working w/ the returned Contentful `DynamicEntry` for grabbing all the supplied data on a content type.

It also includes some Forge-ification to make things look nicer.

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -3,119 +3,21 @@
 namespace App\Entities;
 
 use Contentful\Delivery\Asset;
-use Contentful\Delivery\DynamicEntry;
 
 /**
  * The Campaign entity.
+ *
+ * @property string $title
+ * @property string $callToAction
+ * @property Asset $coverImage
+ * @property array $problemFact
+ * @property array $solutionFact
+ * @property array $solutionStatement
+ * @property array $facts
+ * @property string $faqs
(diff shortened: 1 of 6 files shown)


app/Entities/Campaign.php
```diff
@@ -3,116 +3,22 @@
...
+ * @property array $problemFact
+ * @property array $solutionFact
+ * @property array $solutionStatement
+ * @property array $facts
+ * @property string $faqs
  */
-class Campaign
+class Campaign extends Entity
 {
     /**
```
weerd: It's all gone! haha. I like that now it only has to have the methods specific to this particular entity type!

app/Entities/Entity.php
```diff
@@ -0,0 +1,120 @@
...
+            return $value;
+        }
+
+        if ($value instanceof DynamicEntry) {
+            return new self($value);
+        }
+
+        if (is_array($value)) {
+            return collect($value)->map(function($value) {
+                return new Entity($value);
```
weerd: Can you not do `new self()` like above cause it's in the closure?

weerd:
Looking awesome! ❤️  the improvements!

PR #13, opened 2016-12-16
Use localized images & add query string for locale.

DFurnes:
This PR adds support for pulling localized images to the `get_image_url` helper method, and adds a middleware to set the application's locale by using the `?locale=` query string. 🌐


app/Http/Kernel.php
```diff
@@ -30,6 +30,7 @@ class Kernel extends HttpKernel
             \Illuminate\View\Middleware\ShareErrorsFromSession::class,
             \App\Http\Middleware\VerifyCsrfToken::class,
             \Illuminate\Routing\Middleware\SubstituteBindings::class,
+            \App\Http\Middleware\SetLocale::class,
```
weerd: Huzzah!

PR #14, opened 2016-12-16
Update to Problem Fact markup.

weerd:
This PR makes a quick update to the HTML markup for the Problem Fact section.


weerd:
👍  IRL!

PR #21, opened 2017-01-19
Northstar authentication! :lock:

deadlybutter:
This pull request adds Northstar authentication to phoenix-next and makes the necessary config & db changes to make it work. I also got rid of a few files no longer needed.


app/Models/User.php
```diff
@@ -2,28 +2,15 @@
...
-
-    /**
-     * The attributes that should be hidden for arrays.
-     *
-     * @var array
-     */
-    protected $hidden = [
-        'password', 'remember_token',
-    ];
+    protected $table = 'users';
```
DFurnes: Is this necessary? Laravel should pluralize the classname by default.

config/auth.php
```diff
@@ -2,6 +2,8 @@
 
 return [
 
+    'model' => App\Models\User::class,
```
DFurnes: Can you add a comment specifying that this is a hack until DoSomething/gateway#65 is fixed?

deadlybutter: :indeed:

app/Models/User.php
deadlybutter: Not necessary, had some generic bland SQL errors and was working through all possible things going wrong. I shall remove!

PR #25, opened 2017-02-03
Adds custom templates for blocks and some front-end stuff!

DFurnes:
#### Changes
Here's what's in this round:

💥 Removes Laravel's default packaged theme and Vue components. (aeee808)

🔨 Adds our custom Webpack config and hooks everything up. ~~__Remaining question:__ How do we configure Forge/Envoyer to run this command before deploys?~~ Added deploy hooks. (2351e3b)

🎨 Adds some custom styles for container block widths and a lil' empty "block" widget.

📄 Adds some templates for the blocks specified in Contentful – if the `type` field matches a Blade template, it'll be rendered onto the page via the `@includeIf`. Otherwise it'll be ignored. (bdb6cd4)

#### Screenshot

screen shot 2017-02-03 at 5 31 22 pm


weerd:
@DFurnes regarding "How do we configure Forge/Envoyer to run this command before deploys?"

There's a **Deployment Hooks** tab which lets you click on any of the below _main_ steps and add pre/post steps to each of them. We could add the webpack stuff somewhere in there 😄 

image

package.json
```diff
@@ -1,18 +1,28 @@
...
-    "vue-resource": "^1.0.3"
-  }
+    "private": true,
+    "scripts": {
+        "start": "npm run clean && npm run modernizr && webpack",
+        "build": "npm run clean && npm run modernizr && NODE_ENV=production webpack",
+        "clean": "rm -rf public/dist",
+        "modernizr": "modernizr -c node_modules/@dosomething/forge/modernizr.json -d public/dist/modernizr.js"
+    },
+    "babel": {
```
weerd: thoughts on adding the babel config stuff in a `.babelrc` file? Helps cleanup the package.json file and get it out of the way since it's unlikely to ever get amended much.

DFurnes:
Awesome! I just added steps to install NPM dependency and build assets after the "Install Composer Dependencies" step, so hopefully that works. 🔨

package.json
DFurnes: Hah! I've actually gotten in the habit of putting as much config as possible in the `package.json` in order to clean up the directory listing. 😆

weerd: Haha, we can leave it. I just find keeping the `package.json` cleaner since I'm in there more often, and stuff that can be "set and forget" as separate files when possible. But not a big deal either way 👍

PR #26, opened 2017-02-07
Adding login/logout links in main nav.

weerd:
This PR adds a quick nav link(s) to allow users to login or logout via the already setup auth via Northstar.


deadlybutter:
fancy schmancy 👍

PR #27, opened 2017-02-07
Adds Heroku

deadlybutter:
This pull request allows phoenix-next to run on Heroku, additionally it adds an `app.json` file so review apps can be generated on the fly for pull requests.

- Adds Procfile -- Used by Heroku to start the process
- Adds app.json -- This has a few env vars set, but most are copied from the staging dyno at build time. Additionally it defines a mysql service and a post deploy script
- Adds setup.sh -- Runs migrations, seed & builds fender assets
- Changed database config -- This was needed to support mysql strings that are provided from the cloud service

cc @blisteringherb


deadlybutter:
(Im also going to squash these commits, ha....)

app.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "phoenix-next",
+  "description": ":sparkles:",
+  "scripts": {
+    "postdeploy": "sh bin/setup.sh"
```
DFurnes: Can we move this into `bootstrap/` so that we don't add a new top-level directory?

bin/setup.sh
```diff
@@ -0,0 +1,3 @@
+php artisan migrate --seed --force
+npm install --only=dev
```
DFurnes: Do we want this `--only=dev` flag?

config/database.php
```diff
@@ -1,5 +1,19 @@
 <?php
 
+$mysql_host = env('DB_HOST', 'localhost');
+$mysql_port = env('DB_PORT', '3306');
+$mysql_database = env('DB_DATABASE', 'forge');
+$mysql_username = env('DB_USERNAME', 'forge');
+$mysql_password = env('DB_PASSWORD', '');
+
+if (env('CLEARDB_DATABASE_URL')) {
```
DFurnes: Can you add a comment above here explaining that this is used for Heroku environments?

app.json
deadlybutter: fine with me!

bin/setup.sh
deadlybutter: the heroku npm buildback does the normal npm install, but it leaves out the dev dependencies (so yes, intentional)

config/database.php
deadlybutter: 👍

bin/setup.sh
DFurnes: Ah, that makes sense! Can you add a `# comment` above explaining that?

deadlybutter: sure thing!

deadlybutter:
Seems like everything still works after those changes, let me know if the comments help clear things up

DFurnes:
Woohoo! 🎉 This is so cool:

screen shot 2017-02-07 at 5 22 08 pm

PR #30, opened 2017-02-09
Add a flexbox grid component & light visual touches.

DFurnes:
#### Changes
This pull request adds a "Flex" component for building the layout on the new feed page, and a "Block" component which is just responsible for the box appearance. This allows us to implement the evenly spaced grid from the wires so far (whereas the previous implementation with floats left gaps and awkward spacing in rows).

#### Screenshots
screen shot 2017-02-09 at 4 02 09 pm


DFurnes:
And just for proof, here it is working perfectly in IE 11! 😉 Some minor bugs in IE 10, but nothing bad.

bs_win7_ie_11 0

DFurnes:
And [on mobile as well](https://cloud.githubusercontent.com/assets/583202/22804715/502c3296-eee8-11e6-9782-51043c2d8a03.png). 📱

resources/assets/components/header.scss
```diff
@@ -0,0 +1,15 @@
+// HACKZ: Make the header a bit smaller for nice demo.
```
weerd: Speaking of hackz, should we also start thinking about removing some of the legacy Forge HTML structure. Not sure how much will apply in the new experience compared to the old experience. Might be worth thinking about at some point in the near future!

DFurnes: Agreed!

weerd:
image

🕺

DFurnes:
[As a dev, I want to have reportback data available in the browser and display it in each block.](https://trello.com/c/e8BCMXNI/266-as-a-dev-i-want-to-have-reportback-data-available-in-the-browser-and-display-it-in-each-block)

PR #34, opened 2017-02-13
Collect user emails on welcome page

deadlybutter:
When users click into the new campaigns, we won't have the experience to show them. We want to capture their email, however, to let them know when we do launch. We'll just export the table as a CSV & import into MailChimp later (We don't expect a massive amount of people in this list...)

### DIFF
--- a/app/Http/Controllers/WaitingListController.php
+++ b/app/Http/Controllers/WaitingListController.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use Request;
+use Session;
+use App\Models\WaitingListUser;
+
+class WaitingListController extends Controller
+{
+
+    /**
+     * Store a users email.
+     *
+     * @param  Request $request
+     * @return Response
+     */
+    public function store(Request $request) {
+        $user = new WaitingListUser;
+        $user->email = Request::input('email');
+        $user->save();
+
+        Session::flash('flash_message', 'Thanks! We\'ll let you know when we lauch.');
+
+        return redirect('/');
+    }
+
+}
--- a/app/Http/Controllers/WaitingListController.php
+++ b/app/Http/Controllers/WaitingListController.php
@@ -2,9 +2,8 @@
 
 namespace App\Http\Controllers;
 
-use Request;
-use Session;
 use App\Models\WaitingListUser;
+use Illuminate\Http\Request;
 
 class WaitingListController extends Controller
 {
@@ -16,13 +15,11 @@ class WaitingListController extends Controller
      * @return Response
      */
     public function store(Request $request) {
-        $user = new WaitingListUser;
-        $user->email = Request::input('email');
-        $user->save();
+        WaitingListUser::create([
+           'email' => $request->input('email')
+        ]);
 
-        Session::flash('flash_message', 'Thanks! We\'ll let you know when we lauch.');
-
-        return redirect('/');
+        return redirect('/')->with('flash_message', 'Thanks! We\'ll let you know when we lauch.');
     }
 
 }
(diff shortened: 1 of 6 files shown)


deadlybutter:
[As a developer, I want to collect emails from the construction page (exportable CSV).](https://trello.com/c/RTGFmrVd/299-as-a-developer-i-want-to-collect-emails-from-the-construction-page-exportable-csv)

DFurnes:
@deadlybutter Didn't we decide to do this with a [MailChimp embedded signup form](http://kb.mailchimp.com/lists/signup-forms/add-a-signup-form-to-your-website), like we've used for other landing pages? I think we'd rather avoid the work of storing, exporting, & importing this as a CSV.

deadlybutter:
@DFurnes No we talked with Kika earlier, apparently the embedded signup form is no longer supported & we'd have to use the mailchimp api

app/Http/Controllers/WaitingListController.php
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Store a users email.
+     *
+     * @param  Request $request
+     * @return Response
+     */
+    public function store(Request $request) {
+        $user = new WaitingListUser;
+        $user->email = Request::input('email');
```
DFurnes: This should use the passed request object, not the facade: `$request->input('email');`

app/Http/Controllers/WaitingListController.php
```diff
@@ -0,0 +1,28 @@
...
+{
+
+    /**
+     * Store a users email.
+     *
+     * @param  Request $request
+     * @return Response
+     */
+    public function store(Request $request) {
+        $user = new WaitingListUser;
```
DFurnes: Can you change this to add `email` as a fillable on the model and use the "create" static method?
```php
WaitingListUser::create([
   'email' => $request->input('email');
]);
```

app/Http/Controllers/WaitingListController.php
```diff
@@ -0,0 +1,28 @@
...
+        $user->email = Request::input('email');
+        $user->save();
+
+        Session::flash('flash_message', 'Thanks! We\'ll let you know when we lauch.');
```
DFurnes: This can be moved in with the redirect below:

```php
return redirect('/')->with('flash_message', 'Thanks! We\'ll let you know when we lauch.');
```

deadlybutter:
Updated, everything appears to still work

deadlybutter:
had to destroy the app and rebuild it because the app.json changed, but all works now

app/Http/Controllers/WaitingListController.php
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * Store a users email.
+     *
+     * @param  Request $request
+     * @return Response
+     */
+    public function store(Request $request) {
+        WaitingListUser::create([
+           'email' => $request->input('email')
```
DFurnes: Oh, sorry I didn't think of this before – what happens if someone submits an empty form here? Should we add a validation rule to mark this field as required? Ignore the submission?

deadlybutter: 🤔  The submission goes through but it seems to make an empty entry. Not the end of the world? We can add validation if we're worried

DFurnes: I think adding some lightweight controller validation makes sense here. Could also use `firstOrCreate` to avoid making duplicate rows if a user submits their email more than once.

deadlybutter: sounds good to me

deadlybutter:
Added the `required|email` rule, feel free to give it a shot!

DFurnes:
OMG I FORGOT I COULD _DO_ THAT! 😄

deadlybutter:
🚂 🚋 🚋

THE MERGE TRAIN IS BACK CHOO CHOOOO

PR #35, opened 2017-02-13
Fixes Heroku Review apps

deadlybutter:
Phew this was a brain tease, let me try to explain all of this,

`"NPM_CONFIG_PRODUCTION": "false",` - This forces dev modules to be installed when the build pack runs `npm install`. Basically making the setup.sh script have 1 less step

`Fideloper/Proxy` - When Heroku spins up a review app, it uses HTTPS. However internally behind the load balancer, it's treated as HTTP. We needed to tell our app to trust all proxies & AWS headers. (In the future this might have to be conditional based on environment, but I assume it won't be)

`"postinstall": "npm run build"` - Originally this was in `setup.sh` but that step only runs when the dyno is created, not per commit. I moved it to after `npm install` since that's simple & guaranteed to run. If someone wants to veto this I can move it behind a script that checks for ENV variable (Basically, post npm install, if this is Heroku, run npm build). But that seemed overkill?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 Homestead.json
 Homestead.yaml
 .env
+.DS_Store
--- a/app.json
+++ b/app.json
@@ -6,6 +6,7 @@
   },
   "env": {
     "APP_DEBUG": "true",
+    "NPM_CONFIG_PRODUCTION": "false",
     "APP_KEY": {
       "required": true
     },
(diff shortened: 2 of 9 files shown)


deadlybutter:
the proof this works should be in the deployment above (when it finishes) ^

deadlybutter:
[As a dev, I want to see the latest frontend changes on Heroku review apps](https://trello.com/c/nyfTNpK8/289-as-a-dev-i-want-to-see-the-latest-frontend-changes-on-heroku-review-apps)

PR #36, opened 2017-02-15
Phoenix Next API

weerd:
This PR enables a couple proxy endpoint in Phoenix-Next. The `api/v1/reporbacks` & `api/v1/signups` endpoints act as proxies for any client side JS ajax calls; the requests hit Phoenix-Next first and then it sends the request to Phoenix-Ashes to get/post data. This is a temporary pathway until [Rogue](https://github.com/DoSomething/rogue) is fully up and running to process requests.

The following are example JS calls incorporating the [Gateway JS](https://github.com/DoSomething/gateway-js) package for helping setup the API calls in JavaScript:

```javascript
// app.js

import { RestApiClient } from '@dosomething/gateway';

const Phoenix = new RestApiClient;

// GET all promoted Reportbacks
Phoenix.get('api/v1/reportbacks', {status: 'promoted'})
  .then((response) => {
    console.log(response);
  });

// GET specific Reportback
Phoenix.get('api/v1/reportbacks/10')
  .then((response) => {
    console.log(response);
  });

// GET all Signups for specific campaign
Phoenix.get('api/v1/signups', {campaigns: '362'})
  .then((response) => {
    console.log(response);
  });

// GET specific Signup
Phoenix.get('api/v1/signups/4269')
  .then((response) => {
    console.log(response);
  });

// POST Signup
Phoenix.post('api/v1/signups', {
    user_id: '1600000',
    campaign_id: '1436',
    source: 'phoenix-next'
  })
  .then((response) => {
    console.log(response);
  });
```

_NOTE:_ Aiming to make an update to `Gateway JS` so that the `Phoenix` class (which extends the `RestApiClient` class) has some specific methods that handle things easier, rather than creating a client from the `RestApiClietnt`. Both work, but it'd be nice since we already have a dedicated client in the package 😉


app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -0,0 +1,75 @@
...
+     * ReportbackController constructor.
+     *
+     * @todo once Rogue is ready, this will all change to request
+     * Reportbacks from Rogue instead of PhoenixLegacy.
+     */
+    public function __construct(PhoenixLegacy $phoenixLegacy)
+    {
+        $this->phoenixLegacy = $phoenixLegacy;
+
+        $this->middleware('api');
```
DFurnes: I've liked the pattern we've used in Northstar of applying middleware groups (like `web` or `api`) in the routes file, and individual middleware (like `auth` or `role`) in the controller.

app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Display a listing of the resource.
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function index()
+    {
+        $reportbacks = remember($this->makeCacheKeyFromRequest(), 30, function() {
```
DFurnes: This caching (and cache key generation) should live within the API client.

app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -0,0 +1,75 @@
...
+            return $this->phoenixLegacy->getAllReportbacks(request()->query());
```
DFurnes: Let's use a Request `$request` parameter here rather than the helper method.

app/Http/Controllers/Api/SignupController.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request)
+    {
+        return $this->phoenixLegacy->storeSignup(
```
DFurnes: We should have controller validation here so we don't attempt to make requests without any of these values.

app/Http/Controllers/Controller.php
```diff
@@ -10,4 +10,13 @@
...
 {
     use AuthorizesRequests, DispatchesJobs, ValidatesRequests;
+
+    /**
+     * Get the cache key for the current request.
+     *
+     * @return string
+     */
+    protected function makeCacheKeyFromRequest() {
+        return make_cache_key(request()->path(), request()->query());
```
DFurnes: I think it makes more sense to create the cache key based on the API request, not the current page's request.

app/Http/Controllers/Api/ReportbackController.php
weerd: 👍

app/Http/Controllers/Api/ReportbackController.php
weerd: Yeah, I wasn't actually 100% sure that it was cool to do that in the `index()` method, which is why I went for the helper method, but doesn't seem like it's a big deal and double-checked and it works fine. Will update :)

weerd:
@DFurnes updated!

app/Exceptions/Handler.php
```diff
@@ -62,4 +67,39 @@ protected function unauthenticated($request, AuthenticationException $exception)
...
+        }
+
+        if ($exception instanceof ValidationException) {
+            $code = 422;
+        }
+
+        $response = [
+            'error' => [
+                'code' => $code,
+                'message' => $exception->getMessage(),
```
DFurnes: We should probably [hide error details in this response](https://github.com/DoSomething/northstar/blob/d3f583dca778bce96cf83b87be136a29466b8983/app/Exceptions/Handler.php#L139-L146) on production, so we don't accidentally expose sensitive information in error strings. (Also thinking that a standard "handler" makes sense as a package, but that's a task for another time!)

weerd: ah, good point. Yeah, I pulled some of this code from stuff I had done in Gladiator and was thinking it might be best to make a package in the future 👍

weerd:
@DFurnes for reference, the default error message set for `ValidationException` in Laravel is:

> The given data failed to pass validation.

Doesn't reveal the specifics.

PR #37, opened 2017-02-16
Legacy Id

deadlybutter:
This pull request adds a `legacy_id` (drupal id) to the user model & will automatically fill itself in the background.

### DIFF
--- a/app/Providers/EventServiceProvider.php
+++ b/app/Providers/EventServiceProvider.php
@@ -28,7 +28,12 @@ class EventServiceProvider extends ServiceProvider
     {
         parent::boot();
 
-        User::creating(function($user) {
+        // Went with 'saving' over 'creating' in the event
+        // a request fails & we want to auto-retry next on the authentication / update.
+        User::saving(function($user) {
+            if ($user->legacy_id) return;
+
+            // If the User does not have a legacy Drupal id cached, fetch from Northstar.
             $northstarUser = gateway('northstar')->asClient()->getUser('id', $user->northstar_id);
             $user->legacy_id = $northstarUser->drupal_id;
             $user->save();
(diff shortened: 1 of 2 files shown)


deadlybutter:
[As a dev, I want to grab the user's Drupal ID for phoenix-next.](https://trello.com/c/9jPU9NVj/253-as-a-dev-i-want-to-grab-the-user-s-drupal-id-for-phoenix-next)

app/Providers/EventServiceProvider.php
```diff
@@ -27,6 +28,15 @@ public function boot()
...
+        // Went with 'saving' over 'creating' in the event
+        // a request fails & we want to auto-retry next on the authentication / update.
+        User::saving(function($user) {
+            if ($user->legacy_id) return;
+
+            // If the User does not have a legacy Drupal id cached, fetch from Northstar.
+            $northstarUser = gateway('northstar')->asClient()->getUser('id', $user->northstar_id);
+            $user->legacy_id = $northstarUser->drupal_id;
+            $user->save();
+        });
```
weerd: @deadlybutter mind elaborating on why this is happening in the `EventServiceProvider` and not on the `User` model?

deadlybutter: @weerd where on the User model could we accomplish something similar? I figured all events had to be in the provider

DFurnes: I've seen model events both in the service provider and as part of a constructor on the model… no real preference on my end. For what it's worth, [Laravel docs](https://laravel.com/docs/5.3/eloquent#events) seem to encourage putting it here.

weerd: Cool cool. I didn't mean to imply it was wrong or anything. I was just curious :)

deadlybutter:
I can play around with it now! Worst comes to worst we at least have a reliable fallback

PR #38, opened 2017-02-16
First pass at rendering the feed client-side.

DFurnes:
This pull request is a bit of a whopper, and _aaaalmost_ gets us to feature parity with the server rendered version. For now, this is a plain ol' props & state React application. Welp, here goes:

📁 Allows us to serializes entities from the server into a client-side variable. (6c9371e, 4399b2e)

⌛️ Adds a lil' helper method to wait for DOM Ready. (f547afb)

🔳 Updates the server-rendered view to display "loading" placeholders on the grid. (a09fb44)

📦 Adds React, Lodash, and [Marked](https://github.com/chjj/marked). (0e54ebb)

📝 Configures a `markdown()` helper so we can render Markdown on the client. (783a68d)

👤 Adds a totally incomplete Figure element. (8693521)

🔡 Adds a base `Block` component & specific `CallToActionBlock`, `CampaignUpdateBlock`, and `PlaceholderBlock` components that compose it. (b3c9e30)

📐 Adds `Flex` and `FlexCell` components for easily building flexbox grids. (0c727f8)

📟 Adds the `Feed` component for displaying the whole feed. This is the "root level" app view where we'll be keeping application state, at least for the moment. (2673734)


deadlybutter:
I love the formatting on this description 👏

DFurnes:
This is still missing the reportback block (ironically?), which will come in a future PR. 😄

app/Entities/Entity.php
```diff
@@ -57,7 +58,7 @@ public function __get($property)
         }
 
         if ($value instanceof DynamicEntry) {
-            return new self($value);
+            return new Entity($value);
```
weerd: why not `self`?

resources/assets/helpers.js
```diff
@@ -0,0 +1,31 @@
+import marked from 'marked';
+
+/**
+ * Wait until the DOM is ready.
+ *
+ * @param {Function} fn
+ */
+export function ready(fn) {
+  if (document.readyState != 'loading'){
```
weerd: `!==`

DFurnes:
@deadlybutter I just added Phoenix legacy URL, username, and password to the environment config on Heroku, and also set `APP_DEBUG` to `false` so we don't accidentally expose those credentials in a stack trace.

deadlybutter:
@DFurnes gotcha 👍 , you'll also need to add them to the app.json here for review apps to recognize it https://github.com/DoSomething/phoenix-next/blob/dev/app.json#L32

DFurnes:
Oooh, gotcha. I'll do that now.

resources/assets/components/CampaignUpdateBlock.js
```diff
@@ -0,0 +1,26 @@
...
+  const { title, content } = props.fields;
+
+  const author = 'Lizzy Divine';
+  const jobTitle = 'Campaign Lead';
+  const avatar = 'http://images.contentful.com/e5mq1t8pfsum/6bZ26cuCC4YmAgYA6YasQs/2ce7b11d25384edc35fcfe7c24002bce/Pic_1_PP_DITL_EDivine.jpeg';
+
+  return (
+    <Block>
+      <h2>{title}</h2>
+      <div dangerouslySetInnerHTML={markdown(content)} />
```
weerd: hmmmm `dangerouslySetInnerHTML` 🤔

deadlybutter:
@DFurnes I think we might need to also recreate the review app since we did a app.json change, if it doesnt work let me know & I can do that

DFurnes:
@deadlybutter Nope, it [looks good to me](https://phoenix-next-pr-38.herokuapp.com/campaigns/teens-for-jeans)! 👌

deadlybutter:
I just realized we have a meeting, but I'll make sure to take a look first thing tomorow morning!

app/Entities/Entity.php
DFurnes: Using `self` would create a new instance of any subclass, so all the "blocks" would have become `Campaign` instances rather than generic `Entities`. It probably makes sense to add a little class mapping here in the future! (Or have them all as dumb entities? Idk.)

resources/assets/helpers.js
DFurnes: Oy! Good catch. 😉

resources/assets/components/CampaignUpdateBlock.js
DFurnes: :warning:‼️

resources/views/campaigns/show.blade.php
```diff
@@ -8,13 +8,17 @@
...
-        <div class="flex wrapper">
-            @foreach ($campaign->activity_feed as $block)
-                <div class="flex__cell {{ $block->displayOptions->map(function($c) { return '-'.$c; })->implode(' ') }}">
-                    @includeIf('blocks.'.$block->type, ['block' => $block])
-                </div>
-            @endforeach
+    <div id="app">
+        <div class="block-container">
+            <div class="flex wrapper">
+                @foreach ($campaign->activity_feed as $block)
```
deadlybutter: Isn't this automatically overwritten by the React render?

resources/views/campaigns/show.blade.php
```diff
@@ -8,13 +8,17 @@
...
 
-    <div class="block-container">
-        <div class="flex wrapper">
-            @foreach ($campaign->activity_feed as $block)
-                <div class="flex__cell {{ $block->displayOptions->map(function($c) { return '-'.$c; })->implode(' ') }}">
-                    @includeIf('blocks.'.$block->type, ['block' => $block])
-                </div>
-            @endforeach
+    <div id="app">
+        <div class="block-container">
```
deadlybutter: Won't react immediately override this?

resources/views/campaigns/show.blade.php
DFurnes: Just some "placeholder" content intended to give a little less of a "flash" while waiting for first render on slower internet connections or computers/phones.

deadlybutter: ah gotcha

app/Entities/Entity.php
weerd: ahh. ok!

DFurnes:
[(3) As a dev, I want to have reportback data available in the browser and display it in each block.](https://trello.com/c/e8BCMXNI/266-3-as-a-dev-i-want-to-have-reportback-data-available-in-the-browser-and-display-it-in-each-block)

PR #39, opened 2017-02-17
React-ify the reportback block.

DFurnes:
This pull request adds the reportback block, but rendered on the client! This is important because we'll  need to be able to load more pages of reportbacks as the user scrolls through the feed.

🗂 Updates the Phoenix Legacy API call to load user profile info & _not_ flatten the reportback items. This is a change in behavior from how the legacy gallery works, but makes sense because we don't want a user's name & impact line repeated more than once. (5dc4f45)

💪 Tidies up the `Flex` and `FlexCell` components & sets a default empty array for modifiers prop. This allows us to use a `FlexCell` without a specific width modifier (without errors). (5b22e15)

🔳 Uses the right class name on server-rendered placeholder so it looks right. (ee0a499)

🏞 Adds the "Reportback Block" component! There's also some code to "normalize" the Contentful feed and Phoenix Legacy API call into one state object, which can be cleaned up into a Redux store in another PR. (c12816a)

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -57,9 +57,8 @@ class CampaignController extends Controller
         $campaign = $this->campaignRepository->findBySlug($slug);
         $phoenixNid = data_get($campaign, 'legacy_campaign_id', '1144');
 
-        $response = $this->phoenixLegacy->getAllReportbacks(['campaigns' => $phoenixNid, 'status' => 'promoted']);
-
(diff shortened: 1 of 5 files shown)


DFurnes:
[(3) As a dev, I want to have reportback data available in the browser and display it in each block.](https://trello.com/c/e8BCMXNI/266-3-as-a-dev-i-want-to-have-reportback-data-available-in-the-browser-and-display-it-in-each-block)

PR #40, opened 2017-02-17
Exposing services object

weerd:
This PR exposes specific environment variables defined in `.env` so that they are made available to the client. It specifically exposes DoSomething services urls so that they can be configured based on development or production needs.

This will be very useful for packages like [Gateway JS](https://github.com/DoSomething/gateway-js) and its services specific API client classes, so they know which service to hit to request data when used locally during development versus production.

Also, for example usage:

```javascript
console.log(services.PHOENIX_URL);
console.log(services.PHOENIX_LEGACY_URL);
console.log(services.GLADIATOR_URL);
console.log(services.NORTHSTAR_URL);
```


package.json
```diff
@@ -1,34 +1,35 @@
...
-        "@dosomething/webpack-config": "^1.1.1",
-        "babel-preset-es2015": "^6.22.0",
-        "babel-preset-react": "^6.22.0",
-        "modernizr": "^3.3.1",
-        "webpack": "^1.11.0",
-        "webpack-dev-server": "^1.11.0"
-    }
+  "private": true,
+  "scripts": {
+    "start": "npm run clean && npm run modernizr && NODE_ENV=development webpack",
```
weerd: Updated this to add `NODE_ENV=development`.

package.json
```diff
@@ -1,34 +1,35 @@
...
+    "lodash": "^4.17.4",
+    "marked": "^0.3.6",
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2"
+  },
+  "devDependencies": {
+    "@dosomething/webpack-config": "^1.1.1",
+    "babel-preset-es2015": "^6.22.0",
+    "babel-preset-react": "^6.22.0",
+    "dotenv": "^4.0.0",
```
weerd: And added `dotenv` package. Not sure why GitHub decided I changed the entire file :man_shrugging:

webpack.config.js
```diff
@@ -1,4 +1,5 @@
 var webpack = require('webpack');
+var dotenv = require('dotenv').config();
```
deadlybutter: Are we sticking with using `const` instead of var when applicable? If so all of these should _probably_ be const?

package.json
DFurnes: I think npm CLI always changes the file to 2-spaces, whereas our EditorConfig enforces 4. 😬

We can probably change that in the EditorConfig.

.env.example
```diff
@@ -16,17 +16,21 @@ CACHE_DRIVER=redis
...
 QUEUE_DRIVER=sync
 
-CONTENTFUL_CONTENT_DELIVERY_API_KEY=contentfulContentDeliverApiKey
-CONTENTFUL_CONTENT_PREVIEW_API_KEY=contentfulContentPreviewApiKey
-CONTENTFUL_SPACE_ID=contentfulSpaceId
+CONTENTFUL_CONTENT_DELIVERY_API_KEY=
+CONTENTFUL_CONTENT_PREVIEW_API_KEY=
+CONTENTFUL_SPACE_ID=
+
+GLADIATOR_URL=
```
DFurnes: 🎯

webpack.config.js
weerd: We are, but not in the webpack config, so we don't have to babelify the config I believe... :man_shrugging:

DFurnes: Yeah, I think Node 4 still has some wonky behavior with `let` and `const`.

PR #43, opened 2017-02-23
Adds Reactions to the Reportback Block

deadlybutter:
❤️  This PR adds the ability to react to Reportbacks by adding a new `Reaction` component. For the moment I named everything based around Reportbacks, but I could see one day abstracting this out to be a more generic reaction component?

**NOTE**: This PR does _not_ make any new Phoenix API calls. You can see in the `//TODO`'s where those need to go (next PR)

🎨  I added some super simplistic styling so you can see the functionality of this, when we have final visuals we can come back to this

📁  I also moved the Reportback Block into a new folder since it has a scss file asscociated with it now. 
**Question**: Should all components live in a folder? Was a bit confusing at first seeing some were & some weren't

----

@DFurnes I'm stealing your PR style and I have 0 regrets


resources/assets/components/ReportbackBlock/index.js
```diff
@@ -0,0 +1,69 @@
...
+      reacted: !this.state.reacted,
+      total: this.state.total + (this.state.reacted ? -1 : 1),
+    });
+    //TODO: Make API call to change the Reaction state (POST|DELETE)
+  }
+
+  render() {
+    return (
+      <div className="reaction" onClick={this.onReact}>
+        <div className={`reaction__button ${this.state.reacted ? '-reacted' : ''}`}></div>
```
DFurnes: Totally optional, but the `classNames` helper is useful for conditional classes like this!

```js
<div className={classNames('reaction__button', {'-reacted' : this.state.reacted})}></div>
```

resources/assets/components/ReportbackBlock/index.js
```diff
@@ -0,0 +1,69 @@
...
+          <p>{this.state.total}</p>
+        </div>
+      </div>
+    );
+  }
+}
+
+const ReportbackItem = (props) => {
+  const item = props.reportback.reportback_items.data[0];
+  const image = item.media.uri;;
```
DFurnes: `;;` 👀

resources/assets/components/ReportbackBlock/index.js
deadlybutter: thats nifty!

DFurnes:
> __Question:__ Should all components live in a folder? Was a bit confusing at first seeing some were & some weren't

Yeah, I agree on it looking a little weird. I'm on board if we want to try that!

deadlybutter:
@DFurnes I'll work on the file re-organization then after Kudo's stuff -- otherwise this PR should be updated & good to go

resources/assets/components/ReportbackBlock/index.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  render() {
+    return (
+      <div className="reaction" onClick={this.onReact}>
+        <div className={classnames('reaction__button', {'-reacted' : this.state.reacted})}></div>
+        <div className="reaction__meta">
+          <p>{this.state.total}</p>
+        </div>
+      </div>
```
weerd: Might be useful to follow the current markup for reactions on Phoenix-Ashes...

resources/assets/components/ReportbackBlock/reportback.scss
```diff
@@ -0,0 +1,17 @@
+@import '~@dosomething/forge/scss/toolkit';
+
+.reaction {
+  .reaction__button {
+    width: 29px;
+    height: 25px;
+    background-image: url('../../images/heart-empty.svg');
+
+    &.-reacted {
+      background-image: url('../../images/heart-filled.svg');
```
weerd: Just noting for future reference, it could be cool to use CSS to change the style of the heart from only an outline to having a filled color; would save on base64 encoding 2 svgs and just use a single one with more versatility to style it :)

resources/assets/components/ReportbackBlock/index.js
deadlybutter: @weerd See PR description, will work on markup/style when we know what the final thing is

resources/assets/components/ReportbackBlock/reportback.scss
deadlybutter: @weerd oh yeah i forgot we could style svg's

resources/assets/components/ReportbackBlock/index.js
weerd: Sounds good.

PR #44, opened 2017-02-24
Send the Reactions to Phoenix-Ashes

deadlybutter:
- This PR takes advantage of the new request parameter in the report backs endpoint by specifying the currently logged in user 👤 

- Updated the legacy_campaign_id with a config variable so we can debug easier across environments / setups 🛠 

- Parse the Kudos current user in the reaction component ❤️ 

- Adds a Reaction Controller + relevant API functions in the Legacy provider. Called these functions in the Reactions component. 🔩 

- Creates a new "verifyToken" middleware. This lets the server validate the client asking it to do something is in fact that user. With this I added some "scriptifying" to the master template. 🔐  

_STILL TO COME IN LATER PR('S)_
* Handling unauthenticated users
* Fixing the cache (Currently breaks the kudos experience when you try refreshing the page)
* Making an addition to GatewayJS for 1. Reactions 2. This new auth method I introduced.


deadlybutter:
I'm just gonna add the POST/DELETE requests and I'll request review ⌛️

app/Http/Controllers/Api/ReactionController.php
```diff
@@ -0,0 +1,47 @@
...
+
+        if (!Auth::check()) {
+            return response()->json(['error' => 'Not authenticated'], 401);
+        }
+
+        $reportbackItemId = $request->input('reportback_item_id');
+        $termId = $request->input('term_id');
+        $userId = Auth::user()->northstar_id;
+
+        return response()->json($this->phoenixLegacy->storeReaction($reportbackItemId, $termId, $userId));
```
deadlybutter: Need to make this have a delete variant based on the value param

deadlybutter:
@DFurnes @weerd Not sure why reportbacks aren't showing on the review app but it's working locally for me, feel free to take a look!

deadlybutter:
Oh I bet it's because I changed the campaign id.

app.json
```diff
@@ -43,7 +43,8 @@
     },
     "PHOENIX_LEGACY_PASSWORD": {
         "required": true
-    }
+    },
+    "PHOENIX_LEGACY_CAMPAIGN_NID": "1283"
```
weerd: Booya! "Patient Playbooks" is my goto test campaign!

.env.example
```diff
@@ -33,6 +33,7 @@ PHOENIX_URL=
 PHOENIX_LEGACY_URL=
 PHOENIX_LEGACY_USERNAME=
 PHOENIX_LEGACY_PASSWORD=
+PHOENIX_LEGACY_CAMPAIGN_NID=
```
DFurnes: I don't want to nix it since I know we're still in rapid prototyping mode, but this won't work super well once we have more than one campaign. It feels like a better solution would be to have separate "Staging NID" and "Production NID" fields for each campaign on Contentful?

app/Http/Controllers/CampaignController.php
```diff
@@ -55,9 +56,10 @@ public function index()
     public function show($slug)
     {
         $campaign = $this->campaignRepository->findBySlug($slug);
-        $phoenixNid = data_get($campaign, 'legacy_campaign_id', '1144');
+        $phoenixNid = data_get($campaign, 'legacy_campaign_id', config('services.phoenix-legacy.campaign'));
```
DFurnes: Wait, will this fallback to the config value ever be triggered if Contentful isn't down? I think I might be a little confused how this is intended to be used.

app/Http/Controllers/Api/ReactionController.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function update(Request $request)
+    {
+        if (!Auth::check()) {
```
DFurnes: This should be handled by applying the `auth` middleware to this route.

app/Http/Middleware/VerifyToken.php
```diff
@@ -0,0 +1,33 @@
...
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @param  string|null  $guard
+     * @return mixed
+     */
+    public function handle($request, Closure $next, $guard = null)
+    {
+        $northstarId = $request->input('northstar_id');
```
DFurnes: I'm not really sure this is how we want to provide access tokens to the front-end. Can we use Laravel's built-in sessions for this in the meantime?

app/Http/Controllers/Api/ReactionController.php
deadlybutter: Hm, thats what I originally tried and the Auth:user() was not being set

.env.example
weerd: Oh yeah, maybe that would be a bit better.

deadlybutter: I agree, this was just a temporary measure in the meantime. I don't know if we want to do a staging + thor ID though, I think that could get confusing quick

app/Http/Controllers/CampaignController.php
deadlybutter: From what I can tell the `legacy_campaign_id` is not set in the Contentful campaign right now, so the hardcoded value (currently 1144) is being used. Id check contentful but i cant login anymore...

app/Http/Middleware/VerifyToken.php
deadlybutter: As I said above that was my original attempt, I must have done it wrong? I can give it another shot

.env.example
deadlybutter: I think the easiest / non-engineering solution would be everyone agreeing to use either Thor or Staging

resources/views/layouts/master.blade.php
```diff
@@ -47,6 +47,8 @@
 <script type="text/javascript" src="{{ asset('dist/app.js') }}"></script>
 
 {{ isset($state) ? scriptify($state) : scriptify() }}
+{{ Auth::check() ? scriptifyAuthDetails(Auth::user()) : '' }}
```
weerd: How does this look on the `window.` side of things?

app/Http/Controllers/CampaignController.php
DFurnes: Oh, weirdness! It should be set but maybe I typo'd the field name.

resources/assets/components/ReportbackBlock/index.js
```diff
@@ -3,27 +3,48 @@ import Block from '../Block';
...
+
+    this.phoenix.post('api/v1/reactions', {
+      'reportback_item_id': this.props.itemId,
+      'term_id': this.props.reactions.term.id,
+      'value': value,
+      'northstar_id': window.AUTH.northstar_id || '',
+      'access_token': window.AUTH.access_token || '',
+      'reaction_id': this.state.reactionId,
+    })
+    .then(res => {
```
weerd: can we go with full names rather than abbreviations? my 👴ness took a sec to realize what this was

app/Http/Middleware/VerifyToken.php
DFurnes: Give it another go and give me a shout if you run into any errors. 💥

app/Http/Controllers/CampaignController.php
weerd: I see it in the contentful field...

weerd: Yep! It's been defaulting to `1144` because of typo in the string.... camel case not snake case:

```php
$phoenixNid = data_get($campaign, 'legacyCampaignId', config('services.phoenix-legacy.campaign'));
```

DFurnes: Oh doy! Good catch. 🎣

deadlybutter:
Closing this PR in favor of https://github.com/DoSomething/phoenix-next/pull/50

PR #45, opened 2017-02-27
Adding new Media Uploader component

weerd:
This PR adds a new `MediaUploader` component, that should be able to technically handle a variety of media, but for the time being is meant to only really deal with JPEG and PNG, and will throw a basic error if other file types are uploaded. When a file is successfully uploaded it is placed inside the media uploader container for previewing. See below:

mediauploader

The `MediaUploader` component accepts a `label` prop for overriding the text label shown, otherwise it resorts to using a default:

```javascript
<MediaUploader label="Send us your photo" />
```

The MediaUploader also does a bit of processing to remove EXIF data _only_ from JPEG files (since EXIF data only exists on JPEG, TIFF and RAW files, so no need to worry about PNG).

### Next Steps

This is the first component needed for the Reportback Uploader. The next steps will be to create a Reportback Uploader component and import this Media Uploader component inside of it.

### DIFF
--- a/resources/assets/components/MediaUploader/index.js
+++ b/resources/assets/components/MediaUploader/index.js
@@ -1,0 +1,74 @@
+import React from 'react';
+import { processFile } from '../../helpers';
+import './media-uploader.scss';
+
+class MediaUploader extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.handleChange = this.handleChange.bind(this);
+
+    this.state = {
+      file: null,
+      filePreviewUrl: null
+    };
+  }
+
+  handleChange(event) {
+    event.preventDefault();
+
+    this.readFile(event.target.files[0]);
+  }
+
+  readFile(file) {
+    let fileReader = new FileReader;
+    let blob;
+
+    fileReader.readAsArrayBuffer(file);
+
+    fileReader.onloadend = () => {
+      try {
+        blob = processFile(fileReader.result);
+
+        this.setState({
+          file: blob,
+          filePreviewUrl: URL.createObjectURL(blob)
+        });
+      }
+      catch(error) {
+        // @todo: need a nice way to handle this, display message?
+        console.log(error);
+      }
+    }
+  }
+
+  render() {
+    let { filePreviewUrl } = this.state;
+    let content = null;
+
+    if (filePreviewUrl) {
+      content = (<img src={filePreviewUrl} />);
+    } else {
+      content = (<span>{this.props.label}</span>);
+    }
+
+    return (
+      <div className="media-uploader">
+        <label htmlFor="media-uploader">
+          {content}
+          <input type="file" id="media-uploader" name="media-uploader" onChange={this.handleChange} />
+        </label>
+      </div>
+    );
+  }
+}
+
+MediaUploader.propTypes = {
+  label: React.PropTypes.string
+};
+
+MediaUploader.defaultProps = {
+  label: 'Upload Media'
+};
+
+export default MediaUploader;
--- a/resources/assets/helpers.js
+++ b/resources/assets/helpers.js
@@ -1,4 +1,5 @@
 import marked from 'marked';
+import get from 'lodash/get';
 
 /**
  * Wait until the DOM is ready.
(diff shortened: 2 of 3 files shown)


weerd:
_NOTE_: this won't show up on the branch PR deploy because I had only dumped the component on the main campaign page for me to build it out!

resources/assets/helpers.js
```diff
@@ -29,3 +29,125 @@ export function markdown(source) {
...
+ */
+function getFileType(file) {
+  let type = '';
+  let dv = new DataView(file, 0, 5);
+  let byte1 = dv.getUint8(0, true);
+  let byte2 = dv.getUint8(1, true);
+  let hex = byte1.toString(16) + byte2.toString(16);
+
+  switch(hex) {
+    case '8950':
```
DFurnes: I don't want to be the "switch killer" but I feel like this might be more readable as an key-value mapping:

```js
import get from 'lodash/get';

return get({
  '8950' => 'image/png',
  '4749' => 'image/gif',
  '424d' => 'image/bmp'
}, hex, null);


```

resources/assets/components/MediaUploader/index.js
```diff
@@ -0,0 +1,74 @@
...
+      }
+    }
+  }
+
+  render() {
+    let { filePreviewUrl } = this.state;
+    let content = null;
+
+    if (filePreviewUrl) {
+      content = (<img src={filePreviewUrl} />);
```
DFurnes: Do we want to add some CSS so this expands to fill the container (and appears as a square crop)?

resources/assets/components/MediaUploader/index.js
```diff
@@ -0,0 +1,74 @@
...
+    } else {
```
deadlybutter: More of a stylistic question -- do we prefer 

```js
if (x) { ... 
```

or 
```js
x ? ... : ...
```

DFurnes: I think it depends how much work is happening in the conditional. This could probably work nicely as an inline ternary in the JSX markup.

weerd: I was debating that myself, just thought it might be a bit hairy to read the ternary with both options being markup, particularly if the markup grows in the future... :man_shrugging:

resources/assets/components/MediaUploader/index.js
weerd: Yeah, I was planning on handling that a bit more in the upcoming PR when I pop it into the RB uploader. Figured it might be a bit easier to deal with when it's contained in something and then aiming to expand it. It currently does expand to the full width of the enclosing container, but depending on the aspect ratio for it to fill...

resources/assets/helpers.js
weerd: oh sweet @DFurnes. Not as familiar with `get()` from Lodash, but that works too Mr. Switchkiller. Bit cleaner as well.

resources/assets/components/MediaUploader/index.js
weerd: Also figured the if would be ok, since it sits outside the `return (...)` block of markup. I figure we can use a ternary when it's for JSX inside the `return` and whatever else if it's outside? Thoughts?

resources/assets/components/MediaUploader/index.js
DFurnes: Gotcha, makes total sense! 👍

resources/assets/components/MediaUploader/index.js
weerd: Oh, dang. @DFurnes your comment didn't show before. I can change it or fix it up in upcoming PR once I see if it needs to be more elaborate markup...

DFurnes: It's all good! 😄

weerd:
[(13) As a user, I want to be able to reportback on Phoenix-next (hold on Cropping though) so that users can RB on Phoenix-next](https://trello.com/c/uHyDi2RL/264-13-as-a-user-i-want-to-be-able-to-reportback-on-phoenix-next-hold-on-cropping-though-so-that-users-can-rb-on-phoenix-next)

PR #47, opened 2017-02-28
Set "tweet" appearance for short campaign updates.

DFurnes:
#### Changes
This pull request updates the __Campaign Update__ block to display any content shorter than 144 characters as a "tweet", with no title and larger font size. It also pulls author information from the "additional content" blob on the custom block, and sets some defaults for optional fields.

screen shot 2017-02-28 at 11 06 42 am


resources/assets/components/CampaignUpdateBlock/index.js
```diff
@@ -0,0 +1,33 @@
...
+import classnames from 'classnames';
+import { markdown } from '../../helpers';
+import Block from '../Block';
+import Figure from '../Figure';
+import DEFAULT_AVATAR from './default-avatar.png';
+import './campaign-update.scss';
+
+const Byline = ({author, jobTitle = 'DoSomething.org Staff', avatar = DEFAULT_AVATAR}) => (
+  <Figure image={avatar}>
+    <strong>{author}</strong><br/>
```
weerd: `<br/>` 😱  :trollface:

resources/assets/components/CampaignUpdateBlock/index.js
```diff
@@ -0,0 +1,33 @@
...
+    <p className="footnote">{jobTitle}</p>
+  </Figure>
+);
+
+const CampaignUpdateBlock = (props) => {
+  const { title, content, additionalContent } = props.fields;
```
weerd: do all the rest of the props get assigned to the `additionalContent` var? That's kinda cool :)

resources/assets/components/CampaignUpdateBlock/index.js
DFurnes: Awww yeeee! 😈

resources/assets/components/CampaignUpdateBlock/index.js
DFurnes: The `additionalContent` field is the JSON text area on the "Custom Block" content type, so it's already a separate collection of nested fields!

weerd: Gotcha, cool!

PR #48, opened 2017-02-28
Add a real implementation for the Figure component.

DFurnes:
#### Changes
This pull request is just tidying up after #47. It adds a real __Figure__ component that isn't just hardcoded markup, and a helper class that can be used for prefixing modifier class names:

```js
modifiers('alpha'); // → ['-alpha']
modifiers('alpha', 'beta'); // → ['-alpha', '-beta']
```

🎨


resources/assets/helpers.js
```diff
@@ -32,6 +32,16 @@ export function markdown(source) {
 }
 
 /**
+ * Prefix a class name or array of class names.
+ * @param {String|Array} classes
+ */
+export function modifiers(...classes) {
+  if (! Array.isArray(classes)) classes = [classes];
+
+  return classes.map(className => `-${className}`);
```
weerd: Nice!

PR #50, opened 2017-02-28
Send the Reactions to Phoenix-Ashes - Try #2

deadlybutter:
This PR lets users make reactions on reportback items, to do that I made a few changes...

- Return proper status code for Laravel auth failure 
- Created a Reaction Controller for interfacing client --> phoenix
- Fixed a typo in the Campaign Controller for grabbing the legacy id
- Also grabbed reportbacks with the actual logged in user applied in the Campaign Controller 
- Added reaction methods to the Phoenix Legacy provider
- Created a 'Reaction' component that can be used anywhere
- Created a Reportback Item reaction to contain the data / logic for the reaction


app/Http/Controllers/Api/ReactionController.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        $reportbackItemId = $request->input('reportback_item_id');
+        $termId = $request->input('term_id');
+        $reactionId = $request->input('reaction_id');
+        $userId = Auth::user()->northstar_id;
+
+        if ($request->input('value')) {
+            return response()->json($this->phoenixLegacy->storeReaction($reportbackItemId, $termId, $userId));
+        } else {
+            return response()->json($this->phoenixLegacy->deleteReaction($reactionId));
```
DFurnes: This doesn't need to block merging this PR, but it feels like this should be a separate `DELETE /reactions/:id` endpoint, rather than a conditional in the POST action.

resources/assets/components/ReportbackBlock/index.js
```diff
@@ -1,43 +1,70 @@
...
+      total: this.state.total + (newReactionState ? 1 : -1),
+    });
+
+    this.phoenix.post('api/v1/reactions', {
+      'reportback_item_id': this.props.itemId,
+      'term_id': this.props.reactions.term.id,
+      'value': newReactionState,
+      'reaction_id': this.state.reactionId,
+    })
+    .then((res) => {
```
DFurnes: Let's call this either `response` or `json`.

resources/assets/components/ReportbackBlock/index.js
```diff
@@ -1,43 +1,70 @@
...
+      console.log(res);
```
DFurnes: 👾

resources/assets/components/ReportbackBlock/index.js
deadlybutter: Sorry habit of using Express haha, i know Diego mentioned it in the other PR 👍

app/Http/Controllers/Api/ReactionController.php
deadlybutter: I'm in partial to either as they're both equally "complex". I went this route to minimize the code required in the JS component but I can make the swap if you think it will look better

weerd: I'd vote for the separate endpoints and methods in general. Feels more useful for future stuff :man_shrugging:

deadlybutter: Okie dokes, I'll make the swap then!

app/Providers/RouteServiceProvider.php
```diff
@@ -69,7 +69,7 @@ protected function mapWebRoutes()
     protected function mapApiRoutes()
     {
         Route::group([
-            'middleware' => 'api',
+            'middleware' => 'web',
```
weerd: why did this get changed to `web`?

deadlybutter: @weerd me and dave spent quite a while trying to debug why the Laravel auth wasn't working API endpoints, realized the api middleware was missing everything we needed for auth to work. 

I think the consensus was in The Future ™️  we will have a external API & that will use the middleware group, but for internal api endpoints like Reactions we'll use the web middleware group. @DFurnes If I didn't explain this right feel free to correct me

weerd: Would it maybe make sense to add a new item to the `$middlewareGroups` array, that is basically the api one, but with the additional items needed from the `web` one? I feel there might be a few items in the `web` group that aren't actually needed to be instantiated when it's an internal API request? Thoughts? Maybe a future update or something. I was just curious cause that just seemed odd to me.

deadlybutter: @DFurnes Any thoughts on this?

DFurnes: Yeah, I think it actually might make sense to just move these all these endpoints from `routes/api.php` to `routes/web.php` since they're really "web" routes, just ones that specifically return JSON. They're not stateless API endpoints like our internal APIs, since they need the session, cookie, and CSRF token middleware.

deadlybutter:
@weerd @DFurnes Updated! Endpoints are in the `web` group now and the endpoint was split into two

deadlybutter:
In another PR ill tackle the cache issue (Unable to see Kudos changes after refresh)

app/Http/Controllers/ReactionController.php
```diff
@@ -0,0 +1,52 @@
...
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request)
+    {
+        $this->validate($request, [
+            'reportback_item_id' => 'required|string',
+            'term_id' => 'required|string',
+        ]);
+
+        return response()->json($this->phoenixLegacy->storeReaction($request->input('reportback_item_id'),
```
DFurnes: This is nitpicky, but can you reformat this so this line is a little easier to read?

```php
$kudo = $this->phoenixLegacy->storeReaction(
    $request->input('reportback_item_id'),
    $request->input('term_id'),
    Auth::user()->northstar_id)
);

return response()->json($kudo);
```

app/Http/Controllers/ReactionController.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * Delete a reaction.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  int  $id
+     * @return \Illuminate\Http\Response
+     */
+    public function delete(Request $request, $id) {
```
DFurnes: Function brackets should be on their own line:

```php
public function delete(Request $request, $id)
{
  // ...
```

app/Http/Controllers/ReactionController.php
deadlybutter: Sure thing!

app/Http/Controllers/ReactionController.php
deadlybutter: D'OH

deadlybutter:
Cool, def. looks better now 😄

Issue #46, opened 2017-02-28
Reportback Uploader


weerd:
_FEATURE OVERVIEW_
-------------------
#### User Story
> As a user, I want to be able to reportback on Phoenix-next (hold on Cropping though) so that users can RB on Phoenix-next

#### Why This Matters
We want to allow users to easily upload their Reportbacks in the new campaign experience.

#### Tasks
- [x] Create `MediaUploader` component.
- [x] Create `ReportbackUploader` component.
- [x] Send uploaded reportback to Phoenix-Ashes.
- [x] Display submitted reportback in RB Uploader interface within submission gallery.

weerd:
![](https://github.trello.services/images/mini-trello-icon.png) [(13) As a user, I want to be able to reportback on Phoenix-next (hold on Cropping though) so that users can RB on Phoenix-next](https://trello.com/c/uHyDi2RL/264-13-as-a-user-i-want-to-be-able-to-reportback-on-phoenix-next-hold-on-cropping-though-so-that-users-can-rb-on-phoenix-next)

PR #52, opened 2017-03-02
Reportback uploader component

weerd:
This PR add the `ReportbackUploader` component which currently includes the Reportback form and will also include the upcoming Reportback Submissions gallery. It incorporates the use of the `MediaUploader` component for uploading files as well as the required form inputs.

It also sets up a new property called `submission` in the Redux store in preparation for receiving submissions via the RB Uploader.

Currently, it allows you to add a photo and accompanying text input in the fields and submit the reportback but only outputs the submitted data to the console. The next steps will include actually submitting the reportback data to Phoenix-Ashes and then updating the Redux store if the submission was successful.

The styles for the Reportback Uploader are pretty basic for the time being since the designs are not finalized yet, but it's responsive (enough).

Refs #46

### DIFF
--- a/resources/assets/components/MediaUploader/index.js
+++ b/resources/assets/components/MediaUploader/index.js
@@ -7,11 +7,6 @@ class MediaUploader extends React.Component {
     super(props);
 
     this.handleChange = this.handleChange.bind(this);
-
-    this.state = {
-      file: null,
-      filePreviewUrl: null
-    };
   }
 
   handleChange(event) {
@@ -30,7 +25,7 @@ class MediaUploader extends React.Component {
       try {
         blob = processFile(fileReader.result);
 
-        this.setState({
+        this.props.setMediaState({
           file: blob,
           filePreviewUrl: URL.createObjectURL(blob)
         });
@@ -43,7 +38,7 @@ class MediaUploader extends React.Component {
   }
 
   render() {
-    let { filePreviewUrl } = this.state;
+    let { filePreviewUrl } = this.props.media;
     let content = null;
 
     if (filePreviewUrl) {
--- a/resources/assets/components/MediaUploader/index.js
+++ b/resources/assets/components/MediaUploader/index.js
@@ -25,7 +25,7 @@ class MediaUploader extends React.Component {
       try {
         blob = processFile(fileReader.result);
 
-        this.props.setMediaState({
+        this.props.onChange({
           file: blob,
           filePreviewUrl: URL.createObjectURL(blob)
         });
@@ -61,7 +61,7 @@ class MediaUploader extends React.Component {
 MediaUploader.propTypes = {
   label: React.PropTypes.string,
   media: React.PropTypes.object,
-  setMediaState: React.PropTypes.func
+  onChange: React.PropTypes.func
 };
 
 MediaUploader.defaultProps = {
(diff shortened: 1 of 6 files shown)


resources/assets/components/MediaUploader/index.js
```diff
@@ -64,11 +59,17 @@ class MediaUploader extends React.Component {
 }
 
 MediaUploader.propTypes = {
-  label: React.PropTypes.string
+  label: React.PropTypes.string,
+  media: React.PropTypes.object,
+  setMediaState: React.PropTypes.func
```
DFurnes: I think the `setMediaState` prop should be renamed `onChange` to reflect that it's an event listener.

resources/assets/components/ReportbackUploader/reportback-uploader.scss
```diff
@@ -0,0 +1,44 @@
+@import '~@dosomething/forge/scss/toolkit';
+
+$media-uploader-size: 300;
```
DFurnes: This variable looks unused.

resources/assets/components/ReportbackUploader/index.js
```diff
@@ -0,0 +1,85 @@
...
+    event.preventDefault();
+
+    const reportback = {
+      photo: this.state.media,
+      caption: this.caption.value,
+      impact: this.impact.value,
+      why_participated: this.why_participated.value
+    };
+
+    console.log(reportback);
```
deadlybutter: 👀

resources/assets/components/ReportbackUploader/index.js
```diff
@@ -0,0 +1,85 @@
...
+  setMediaState(media) {
+    this.setState({ media })
+  }
+
+  storeReportback(event) {
+    event.preventDefault();
+
+    const reportback = {
+      photo: this.state.media,
+      caption: this.caption.value,
```
deadlybutter: Is this correct? Shouldn't it be `this.state.`?

resources/assets/components/MediaUploader/index.js
weerd: I can change. I was worried it might get confusing since there's already an `onChange` on the file input in the `MediaUploader`, but probs not a big concern.

resources/assets/components/ReportbackUploader/reportback-uploader.scss
weerd: Yeah, was an attempt at something I didn't end up going with. Good catch :)

resources/assets/components/ReportbackUploader/index.js
weerd: Oh, this is grabbing the items directly from the form and not from the state. I think it _does_ highlight that maybe I _only_ need the `media` property in the RB Uploader state for now.

resources/assets/components/ReportbackUploader/index.js
weerd: On purpose 😉  see description

deadlybutter: D'OH

resources/assets/components/ReportbackUploader/index.js
deadlybutter: Ah I see, makes sense

weerd: @deadlybutter I'll see if I end up needing to keep those other items in the RB Uploader state as I progress farther with the rest of the steps involved. If not I'll remove and only keep the `media` prop since it's meant to be kind of a "fleeting" state ;)

PR #53, opened 2017-03-02
Login prompt for reactions

deadlybutter:
This PR makes it so, If an unauthenticated user tries to react on a reportback, they are taken to the Northstar login page.

There is an annoying complication / workaround I had to do in the JS you'll see commented below, other than that everything is pretty straightforward. Also I added the Northstar ID, even though it's not used now we will need it later for event tracking.

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -26,7 +26,7 @@ function scriptifyCurrentUser()
     return scriptify([
         'authenticated' => $authenticated,
         'northstarId' => $authenticated ? Auth::user()->northstar_id : null,
-    ], 'user');
+    ], 'USER');
 }
 
 /**
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -20,6 +20,15 @@ function scriptify($json = [], $store = 'STATE')
     return new HtmlString('<script type="text/javascript">window.'.$store.' = '.json_encode($json).'</script>');
 }
 
+function scriptifyCurrentUser()
+{
+    $authenticated = Auth::check();
+    return scriptify([
+        'authenticated' => $authenticated,
+        'northstarId' => $authenticated ? Auth::user()->northstar_id : null,
+    ], 'user');
+}
+
 /**
  * Get an item from the cache, or store the default value.
  *
--- a/resources/assets/components/ReportbackBlock/index.js
+++ b/resources/assets/components/ReportbackBlock/index.js
@@ -8,10 +8,10 @@ class ReportbackReaction extends React.Component {
   constructor(props) {
     super(props);
 
-    this.onReact = this.onReact.bind(this);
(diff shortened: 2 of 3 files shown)


deadlybutter:
The review app actually works for this PR haha 🎉  🌮

resources/assets/components/ReportbackBlock/index.js
```diff
@@ -11,16 +11,25 @@ class ReportbackReaction extends React.Component {
     this.onReact = this.onReact.bind(this);
     this.phoenix = new Phoenix();
 
+    // This user might the current user or the Drupal API User
```
weerd: small typo "might be"

resources/assets/components/ReportbackBlock/index.js
```diff
@@ -11,16 +11,25 @@ class ReportbackReaction extends React.Component {
...
 
     this.state = {
-      active: currentUser ? currentUser.reacted : false,
+      active: reacted,
       total: this.props.reactions.term.total,
       reactionId: currentUser ? currentUser.kudos_id : '',
     }
   }
 
   onReact() {
```
weerd: lol. maybe we call this `onReaction`?

app/helpers.php
```diff
@@ -20,6 +20,15 @@ function scriptify($json = [], $store = 'STATE')
     return new HtmlString('<script type="text/javascript">window.'.$store.' = '.json_encode($json).'</script>');
 }
 
+function scriptifyCurrentUser()
+{
+    $authenticated = Auth::check();
+    return scriptify([
+        'authenticated' => $authenticated,
```
DFurnes: This should be a part of the state tree in `window.STATE`, rather than it's own separate global.

resources/assets/components/ReportbackBlock/index.js
DFurnes: Hah, yeah. Or `onClick` is fine until we have more than one click handler in the component.

deadlybutter: haters gonna hate :rofl:

app/helpers.php
deadlybutter: Just updating here -- Chatted w/ Dave about this, I'm going to work on redux-ifying _(thats a word now)_ the reaction stuff. in the meantime we will roll with this to merge the PR.

deadlybutter:
[(8) As a developer, if an unauthenticated user tries to do something on the feed, I want them to be prompted to login or create an account so that they are an authenticated user](https://trello.com/c/1MWyBYyG/316-8-as-a-developer-if-an-unauthenticated-user-tries-to-do-something-on-the-feed-i-want-them-to-be-prompted-to-login-or-create-an-a)

PR #54, opened 2017-03-02
Embed API

DFurnes:
#### Changes
This adds an `/embed?url=<url>` endpoint to grab embed information from a given URL. This uses the [`embed/embed`](https://github.com/oscarotero/Embed) package to determine metadata for the URL and an appropriate embed code.

I went this route after looking into [Embed.ly](http://embed.ly), which didn't seem like an excellent fit due to a kinda aggressive pricing model (pay per embed pageview) and a comically large script dependency (70kb… including a full copy of Underscore). I also looked into [Iframely](https://iframely.com) which had a slightly better pricing model (pay per embed, but only once per hour) & a much smaller embed code (14kb). But this is free. 💸

Some examples:

youtube

times

📬

### DIFF
--- a/app/Http/Controllers/EmbedController.php
+++ b/app/Http/Controllers/EmbedController.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use Embed\Embed;
+use Illuminate\Http\Request;
+use Illuminate\Http\JsonResponse;
+
+class EmbedController extends Controller
+{
+    /**
+     * Get details for embedding a link.
+     * GET /embed?url=<url>
+     *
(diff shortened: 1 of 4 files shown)


weerd:
Actually, quick question... if a provided site url doesn't have the embed info, does the package figure stuff out and provide reasonable info to fill out the properties?

DFurnes:
> Actually, quick question... if a provided site url doesn't have the embed info, does the package figure stuff out and provide reasonable info to fill out the properties?

Yes! It'll grab the URL and title but leave the embed code empty:

screen shot 2017-03-02 at 12 15 37 pm

PR #58, opened 2017-03-03
Pre-work for adding additional pages.

DFurnes:
#### Changes
This pull request includes some refactoring I did while working on adding additional pages to the feed. I've pushed it up as a separate PR here to keep things a bit more manageable.

📁 Makes separate `actions/` and `reducers` directories for actions and reducers. Also switches things up to export the actual reducers themselves (rather than the combined root reducer), so we can add additional ones in the store later.

📦 Extracts a "Feed Container" presentation component with it's associated styles from the feed.

🔀 Moves more stuff out of `app.js` into either the `App` or `CampaignFeed` container components.

❓ Removes jQuery (how'd that get in here?!) and Forge's `forge.js` since neither are used.

⬇️ Moves the main `app.js` script below the inlined `window.STATE` object so we don't need to wait on `document.ready` to start putting that stuff into the store.

There are no functional changes, so a quick spot-check of the code and [the review app](https://phoenix-next-pr-58.herokuapp.com/campaigns/teens-for-jeans) is probably good enough. 👀


package.json
```diff
@@ -11,6 +11,9 @@
     "presets": [
       "es2015",
       "react"
+    ],
+    "plugins": [
+      "transform-object-rest-spread"
```
weerd: I encountered this today. Alternatively, could add the `stage-2` preset which includes it :man_shrugging: But might be best not to include extras we don't need.

DFurnes: Yeah, I'm a teeny bit wary of adding too many "in progress" language features, since there's no guarantee they won't change or be abandoned before making it into the spec. But from what I can tell object rest spread is pretty much a sure thing.

DFurnes:
Okay, rebased and ready to merge. Looks like a lot of the same changes happened in other pull requests on Monday so we're all on the same page. 😄

resources/assets/components/FeedContainer/index.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import './feed-container.scss';
+
+const FeedContainer = ({children}) => (
```
deadlybutter: That's a nifty trick!

PR #59, opened 2017-03-06
Updating structure of reducers for Redux pattern.

weerd:
This PR splits up the reducers into their own corresponding files within a new `/reducers` directory, and includes an `index.js` file that collects them all to be combined to provide the exportable `rootReducer` to make things more manageable.

### DIFF
--- a/resources/assets/reducers/submissions.js
+++ b/resources/assets/reducers/submissions.js
@@ -1,0 +1,16 @@
+import { STORED_REPORTBACK_SUBMISSION } from '../actions';
+
+/**
+ * Submissions reducer:
+ */
+const submissions = (state = {}, action) => {
+  switch (action.type) {
+    case STORED_REPORTBACK_SUBMISSION:
+      return state;
+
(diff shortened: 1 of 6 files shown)


resources/assets/reducers/submissions.js
```diff
@@ -0,0 +1,16 @@
+import { STORED_REPORTBACK_SUBMISSION } from '../actions';
+
+/**
+ * Submissions reducer:
+ */
+const submissions = (state = {}, action) => {
+  switch (action.type) {
+    case STORED_REPORTBACK_SUBMISSION:
```
deadlybutter: I'm assuming the next PR is adding this part in?

weerd: Yeah. This is setup for that, but I wanted to get the reducers stuff in so it's coming in smaller bits. I'm not even sure if that's the official action I'll be using moving forward; something I'll be learning as I move to the next bits.

PR #60, opened 2017-03-06
Cap the amount of blocks that can be displayed

deadlybutter:
This PR introduces a few changes to how the feed is rendered

## points

Every block now has a `point` value based on the columns it takes up. This is the reasoning behind using points (Stolen from slack convo w/ Diego)

```
the more i think about it, i think points are easier…

if we have 30 blocks in the campaign, and hard cut at lets say 10, that could leave half or a third of a row empty.

but if the points always add up correctly, you could calculate how many rows you have filled regardless of the blocks. eg: `1/3 = 1 point, 2/4 = 1.5 points, 3/3 = 3 points`. that means if you do `total % 3` you’ll get the total filled rows
```

You can see that in principle ^ in this PR, and it seems to work quite well. 

## Campaign Feed / Feed

In this PR I introduced the *CampaignFeed*, which acts as a data layer for parsing out the blocks, modifying them, etc before being sent out to the *Feed* for rendering.

### DIFF
--- a/resources/assets/components/CampaignFeed/index.js
+++ b/resources/assets/components/CampaignFeed/index.js
@@ -12,9 +12,9 @@ class CampaignFeed extends React.Component {
    */
   mapDisplayToPoints(displayOption) {
     switch (displayOption[0]) {
-      case 'full': return 3;
       case 'one-third': return 1;
       case 'two-thirds': return 2;
+      case 'full': return 3;
       default: return 0;
     }
   }
@@ -49,6 +49,11 @@ class CampaignFeed extends React.Component {
     }
   }
 
+  /**
+   * Get the blocks fully formatted for rendering in the feed.
+   *
+   * @return Array feed
+   */
   formulateFeed() {
     const feed = [];
     const reportbacks = this.props.reportbacks.data.slice(0);
--- a/resources/assets/components/CampaignFeed/index.js
+++ b/resources/assets/components/CampaignFeed/index.js
@@ -58,7 +58,7 @@ class CampaignFeed extends React.Component {
       const displayOptions = block.fields.displayOptions;
       blockPoints += this.mapDisplayToPoints(displayOptions);
 
-      if (blockPoints % 3 > this.props.rowsPerPage) {
+      if (blockPoints / 3 > this.props.rowsPerPage) {
         return true;
       }
 
--- a/resources/assets/components/CampaignFeed/index.js
+++ b/resources/assets/components/CampaignFeed/index.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import Feed from '../Feed';
 
-class Portal extends React.Component {
+class CampaignFeed extends React.Component {
 
   /**
    * Map the given display option to a
@@ -83,7 +83,7 @@ class Portal extends React.Component {
 
 }
 
-Portal.defaultProps = {
+CampaignFeed.defaultProps = {
   rowsPerPage: 3,
   campaign: {
     activityFeed: [],
@@ -93,4 +93,4 @@ Portal.defaultProps = {
   },
 };
 
-export default Portal;
+export default CampaignFeed;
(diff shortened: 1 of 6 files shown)


resources/assets/components/CampaignFeed/index.js
```diff
@@ -0,0 +1,96 @@
...
+   * Map the given display option to a
+   * numeric point value.
+   *
+   * @param array displayOption
+   * @return int
+   */
+  mapDisplayToPoints(displayOption) {
+    switch (displayOption[0]) {
+      case 'full': return 3;
+      case 'one-third': return 1;
```
deadlybutter: I think for neatness I should probably re-order this to read 1, 2, 3

resources/assets/components/CampaignFeed/index.js
```diff
@@ -0,0 +1,96 @@
...
+      for (let i = 0; i < count; i++) {
+        let reportback = reportbacks.shift();
+        if (reportback) {
+          block.reportbacks.push(reportback);
+        }
+      }
+    }
+  }
+
+  formulateFeed() {
```
deadlybutter: forgot a comment

resources/assets/components/CampaignFeed/index.js
```diff
@@ -0,0 +1,101 @@
...
+    });
+
+    return feed;
+  }
+
+  render() {
+    return (
+      <div className="feed-container">
+        <div className="wrapper">
+          <Feed blocks={this.formulateFeed()} />
```
weerd: Does it need the parentheses here `formulateFeed()`? Doesn't this invoke it immediately?

deadlybutter: Good catch!

deadlybutter: I do this pretty often in react, ill learn one day...

deadlybutter: @weerd Oh wait actually no, in this specific case it should be calling the function, because I want it to pass an array

deadlybutter: As it results in this 
screen shot 2017-03-06 at 4 36 45 pm

weerd: hmmm, does it need to get passed as a function then?

`<Feed blocks={() => this.formulateFeed(theArray)} />`?

Not 100% sure, btw... lol

deadlybutter: Well, doing that ^ is no different than what I currently have, both are function calls?

weerd: I thought the former gets invoked immediately while the second doesn't, as an anonymous function... :man_shrugging:

deadlybutter: Oh I think you're right about that, but It doesn't _seem_ to matter (as in, it works, so I assume React was smart enough to handle the immediate invocation?)

deadlybutter: I have no problem changing it if you think its less confusing

weerd: Nah, if it works go for it. We can check on it later if it ends up causing any issues :)

deadlybutter: okie dokes - PR good to go then?

DFurnes: Yeah, I think I'd prefer if we pulled out the `formulateFeed` function so it's clear it's not a typo for a handler func.

```js
const blocks = this.formulateFeed();

return (
    <div className="feed-container">
        <div className="wrapper">
            <Feed blocks={blocks} />
        </div>
    </div>
);
```

deadlybutter: yeah this is already changed, to something like that, will be in the next PR!

DFurnes: Great!

PR #61, opened 2017-03-07
View more button

deadlybutter:
This PR adds a "view more" button to the bottom of the feed and lets you paginate through content. Further enforcing that the campaign feed acts as the state controller, and the Feed acts acts as the pure display. There might be a more redux-y way of doing this?

Additionally, I also did some small refactoring (@DFurnes included that change you suggested as well)


resources/assets/components/Feed/index.js
```diff
@@ -36,16 +36,17 @@ class Feed extends React.Component {
       <Flex>
         {this.props.blocks.map((block, index) => this.renderFeedItem(block, index))}
 
-        <FlexCell key="reportback_uploader">
-          <ReportbackUploader/>
```
DFurnes: Let's keep this here for now since @weerd is still working on this component.

resources/assets/components/Feed/index.js
```diff
@@ -36,16 +36,17 @@ class Feed extends React.Component {
...
-        </FlexCell>
+        <div className="flex_cell -full">
```
DFurnes: This should be using the `<FlexCell width="full">` component.

resources/assets/components/CampaignFeed/index.js
```diff
@@ -50,37 +65,38 @@ class CampaignFeed extends React.Component {
...
       this.setType(block);
-      this.appendReportbacks(block, reportbacks);
-
-      feed.push(block);
+      this.appendReportbacks(block);
+      blocks.push(block);
     });
 
-    return feed;
+    this.setState({
```
DFurnes: This doesn't block merging this PR, but these should both be kept in the Redux store & then this component should be dispatching an action to mutate the store each time the user presses the "View More" button.

resources/assets/components/Feed/index.js
deadlybutter: Ah of course, sorry about that

resources/assets/components/Feed/index.js
deadlybutter: 👍

resources/assets/components/CampaignFeed/index.js
deadlybutter: I can make a go at this in the next PR

deadlybutter:
Sweet, should be good to go now

PR #63, opened 2017-03-07
Reportback uploader redux

weerd:
This PR updates the ReportbackUploader to use the container & presentation pattern common in Redux, and also makes sure to pass all RB `submissions` related data to the Redux store to help manage the state.

The state related to grabbing a file int he browser is still maintained in the ReportbackUploader component since it's not technically a finalized state. The record of the RB only gets passed to the Redux store once the RB submission is successfully accepted.

There is a placeholder for the API request to PhoenixNext to store a newly submitted RB (which needs to hit PhoenixLegacy) that is part of an upcoming PR. This also does not flesh out the markup for the submissions gallery or style it.

### DIFF
--- a/resources/assets/components/ReportbackSubmissions/index.js
+++ b/resources/assets/components/ReportbackSubmissions/index.js
@@ -1,0 +1,22 @@
+import React from 'react';
+
+class ReportbackSubmissions extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  renderSubmission(reportback, index) {
+    let key = `submission-${index}`;
+
+    // @TODO: need to flesh out the markup.
+    return <li key={key}>{reportback.caption}</li>;
+  }
+
+  render() {
+    let submissions = this.props.submissions.data || null;
+
+    return submissions ? <ul>{submissions.map(this.renderSubmission)}</ul> : null;
+  }
+}
+
+export default ReportbackSubmissions;
--- a/resources/assets/components/ReportbackSubmissions/index.js
+++ b/resources/assets/components/ReportbackSubmissions/index.js
@@ -6,10 +6,8 @@ class ReportbackSubmissions extends React.Component {
   }
 
   renderSubmission(reportback, index) {
-    let key = `submission-${index}`;
-
     // @TODO: need to flesh out the markup.
-    return <li key={key}>{reportback.caption}</li>;
+    return <li key={index}>{reportback.caption}</li>;
   }
 
   render() {
--- a/resources/assets/reducers/submissions.js
+++ b/resources/assets/reducers/submissions.js
@@ -1,12 +1,31 @@
-import { STORED_REPORTBACK_SUBMISSION } from '../actions';
+import {
+  STORE_REPORTBACK_PENDING,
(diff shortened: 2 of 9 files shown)


weerd:
Working on resolving merge conflicts....

resources/assets/components/ReportbackSubmissions/index.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+class ReportbackSubmissions extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  renderSubmission(reportback, index) {
+    let key = `submission-${index}`;
```
DFurnes: You can probably just set the `key` to `index` here (since it only needs to be unique among it's siblings).

resources/assets/components/ReportbackSubmissions/index.js
```diff
@@ -0,0 +1,22 @@
...
+
+    // @TODO: need to flesh out the markup.
+    return <li key={key}>{reportback.caption}</li>;
+  }
+
+  render() {
+    let submissions = this.props.submissions.data || null;
```
DFurnes: Should the fallback here be an empty array so you can still call `map` on it if it's not set? And would it make sense to set this as a default prop for the component?

resources/assets/reducers/submissions.js
```diff
@@ -1,12 +1,31 @@
...
 
 /**
  * Submissions reducer:
  */
 const submissions = (state = {}, action) => {
   switch (action.type) {
-    case STORED_REPORTBACK_SUBMISSION:
-      return state;
+    case STORE_REPORTBACK_PENDING:
+      return Object.assign({}, state, {
```
DFurnes: If you want, this could be reformatted to use object spread properties since we added that in an earlier PR! :)

```js
switch (action.type) {
  case STORE_REPORTBACK_PENDING:
    return {...state, isStoring: true };

  case STORE_REPORTBACK_SUCESSFUL:
    return {...state, isStoring: false };

  // ...
```

resources/assets/components/ReportbackSubmissions/index.js
weerd: okie doke. Yeah, wasn't sure and was playing it safe.

resources/assets/components/ReportbackSubmissions/index.js
weerd: but it would output the empty `<ul>` then right?

resources/assets/reducers/submissions.js
weerd: oh jes! I forgot you added that. okie dokes

weerd: it was pre-rebasing. needed to NPM install it!

weerd:
Updated based on your suggestions! Thanks @DFurnes

resources/assets/components/ReportbackSubmissions/index.js
DFurnes: Oh, yep! I totally didn't notice the ternary below. 👍

PR #65, opened 2017-03-08
Generate blocks after a redux action is made

deadlybutter:
woop woop, i think i got redux to work for the feed.

this PR adds a new CREATED_BLOCKS action, which signifies new blocks were generated for the feed to render & a reducer to update the state. all of the relevant components were also refactored to use this new format.


resources/assets/containers/Activity.js
```diff
@@ -1,19 +1,22 @@
 import React from 'react';
 import { connect } from 'react-redux'
-import { toggleTodo } from '../actions'
+import { createdBlocks, toggleTodo } from '../actions';
 import CampaignFeed from '../components/CampaignFeed'
```
weerd: can we add some semicolons to these import statements that are missing them. Didn't realize they were missing them before.

deadlybutter: Oh yeah sure thing!

resources/assets/reducers/blocks.js
```diff
@@ -0,0 +1,18 @@
+import { CREATED_BLOCKS } from '../actions';
+
+/**
+ * Block reducer:
+ */
+const blocks = (state = {}, action) => {
+  switch (action.type) {
+    case CREATED_BLOCKS:
+      return Object.assign({}, state, {
+        data: (state.data || []).concat(action.blocks),
```
weerd: My last, quick PR yesterday added a way we can set defaults in the [**store.js**](https://github.com/DoSomething/phoenix-next/blob/27103d48a42b2e90f8d8c606f7c14fad94772a84/resources/assets/store.js#L6-L15) file. You can likely set the data to an empty array in there so it's set by default from the get go 💃

deadlybutter: oh sweet

resources/assets/components/CampaignFeed/index.js
```diff
@@ -86,15 +82,12 @@ class CampaignFeed extends React.Component {
       blocks.push(block);
     });
 
-    this.setState({
-      blockIndex: this.state.blockIndex + blocks.length,
-      blocks: this.state.blocks.concat(blocks),
-    });
+    this.props.createdBlocks(blocks);
```
DFurnes: Hmm, it actually feels a little weird to be passing the blocks back up into the store from here, since we just end up duplicating existing data that's already in there. Sorry, I think that was bad advice on my part.

It might make more sense to have the action _just_ push the new page offset up to the store & continue filtering blocks in here based on that. That way we always know the state of the app from the Redux store (e.g. we have X feed items and X page offset), but the feed component is responsible for actually rendering based on that.

```js
// Action
export const CLICKED_VIEW_MORE = 'CLICKED_VIEW_MORE';

export function clickedViewMore() {
  return { type: CLICKED_VIEW_MORE };
}

// Reducer
const blocks = (state = {}, action) => {
  switch (action.type) {
    case CLICKED_VIEW_MORE:
      return {...state, offset: state.offset + 1 };

    default:
      return state;
  }
}

// Component
class CampaignFeed extends React.Component {
  // ...

  render() {
    const blocks = this.formulateFeed(this.props.offset);

    return <Feed blocks={blocks} viewMore={this.clickedViewMore} />
  }
```

weerd:
@deadlybutter is this ready for re-review?

deadlybutter:
@weerd Nope!

deadlybutter:
@weerd Now it is!

resources/assets/components/CampaignFeed/index.js
```diff
@@ -55,29 +50,32 @@ class CampaignFeed extends React.Component {
...
+  generateFeed() {
+    this.reportbackIndex = 0;
     let blockPoints = 0;
     const blocks = [];
 
-    const feed = this.props.campaign.activityFeed.slice(this.state.blockIndex);
-    feed.some((block) => {
-      const displayOptions = block.fields.displayOptions;
-      blockPoints += this.mapDisplayToPoints(displayOptions);
+    this.props.campaign.activityFeed.some((block) => {
```
DFurnes: I'm a bit confused by the use of `[].some` here, since the result isn't used... should this be a filter instead?

```js
return this.props.campaign.activityFeed.filter(block => {
      blockPoints += this.mapDisplayToPoints(block.fields.displayOptions);

      const totalRows = blockPoints / BLOCKS_PER_ROW;
      const rowTarget = this.props.blocks.offset * this.props.rowsPerPage;

      if (totalRows > rowTarget) {
        return false;
      }

      this.setType(block);
      this.appendReportbacks(block);
      
      return block;
});
```

resources/assets/reducers/blocks.js
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Block reducer:
+ */
+const blocks = (state = {}, action) => {
+  switch (action.type) {
+    case CLICKED_VIEW_MORE:
+      return Object.assign({}, state, {
+        offset: state.offset + 1,
+      });
```
weerd: Haha, dang it. @DFurnes github didn't update to tell me you already commented on this :\

DFurnes: Too fast!! 🏇

resources/assets/components/CampaignFeed/index.js
deadlybutter: @DFurnes I think semantically you'd be right, but from a "performance" standpoint I picked this specifically because returning `false` in `.some` kills the callback loop. whereas with filter we have to iterate for N blocks before the loop finishes.

resources/assets/reducers/blocks.js
```diff
@@ -0,0 +1,18 @@
+import { CLICKED_VIEW_MORE } from '../actions';
+
+/**
+ * Block reducer:
+ */
+const blocks = (state = {}, action) => {
+  switch (action.type) {
+    case CLICKED_VIEW_MORE:
+      return Object.assign({}, state, {
```
deadlybutter: Oh gotcha - i saw the Object.assign being used so wasn't sure. Easy fix

resources/assets/components/CampaignFeed/index.js
DFurnes: We can do some performance instrumenting later on, but my guess is the extra loops are negligible. There aren't that many blocks that we're iterating over.

deadlybutter: Would you prefer I make the swap then?

DFurnes: Yeah, it's more important that the code be readable than ultra-fast right now.

resources/assets/reducers/blocks.js
weerd: the flash

deadlybutter:
@DFurnes Seems like everything still works with those changes!

DFurnes:
Woo! Ship it! ⛵️

deadlybutter:
🚂 🚋 🚋

deadlybutter:
[(10) As a user, I want to have a "view more" option at the bottom of the Affiliated feed so that it's not infinite scroll in the feed](https://trello.com/c/E1T2FSyV/325-10-as-a-user-i-want-to-have-a-view-more-option-at-the-bottom-of-the-affiliated-feed-so-that-it-s-not-infinite-scroll-in-the-feed)

PR #66, opened 2017-03-09
Endpoint restructuring

weerd:
This PR does some quick updates to the endpoints for `/reportbacks` and `/signups` that are needed for some upcoming work with the `ReportbackUploader` and the signup components.

*NOTE:* I tried doing a bit of testing but it was difficult because I was having issues with reportbacks on Thor and with signups I wasn't sure how to test that via Postman without being auth'd 😬 

I'll be testing further as I use these endpoints via the app to wrap up the components I'm working on.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -23,9 +23,21 @@ $router->get('campaigns/{slug}/{page?}', 'CampaignController@show');
 // Waiting List: for collecting user emails pre-launch.
 $router->post('waitinglist', 'WaitingListController@store');
 
+// Embeds
+$router->get('embed', 'EmbedController@index');
+
+
+/**
+ * The following are API Routes that are currently using the web middleware,
+ * until the implementation of JWT tokens.
+ */
+
 // Reactions
 $router->post('reactions', 'ReactionController@store');
 $router->delete('reactions/{id}', 'ReactionController@delete');
 
-// Embeds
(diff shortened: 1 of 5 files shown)


routes/web.php
```diff
@@ -23,9 +23,21 @@
...
+ */
+
 // Reactions
 $router->post('reactions', 'ReactionController@store');
 $router->delete('reactions/{id}', 'ReactionController@delete');
 
-// Embeds
-$router->get('embed', 'EmbedController@index');
+// Reportbacks
+$router->resource('reportbacks', 'ReportbackController', ['except' => ['create', 'edit', 'destroy']]);
```
deadlybutter: We haven't used these anywhere in the JS right? I know Dave made that Redux demo with it, but I'm assuming we don't have to maintain that atm?

deadlybutter: oh my god im an idiot you did fix the JS stuff IGNORE ME

weerd: Yeah, I need to update the `fetchReportbacks()` method. It's using the helper from Gateway, but that won't work until we setup the JWT token shiz in the future. Good call.

weerd:
@deadlybutter just FYI, made the update to that `fetchReportbacks()` endpoint 😉

deadlybutter:
👍

PR #62, opened 2017-03-07
Pages

DFurnes:
#### Changes
This pull request adds client-side routing, an example page, and a little navigation component for switching between them. You can check it out on the [review app](https://phoenix-next-pr-62.herokuapp.com/campaigns/teens-for-jeans).

#### How should this be reviewed?
I separated out the changes into a couple of smaller commits, so feel free to page through those one-by-one for more "bite sized" pieces. I'll make another PR for loading the additional page(s) & their content from Contentful.

PR #67, opened 2017-03-09
Load additional page content from Contentful.

DFurnes:
#### Changes
This pull request is a follow-up to #62, with real page content being fetched from Contentful for the relevant URL slug. So for example, these two "content pages" are now shown in the nav:

screen shot 2017-03-09 at 4 34 07 pm

screen recording 2017-03-09 at 04 28 pm

Check out the two example campaigns: [with custom pages](http://phoenix-next-pr-67.herokuapp.com/campaigns/teens-for-jeans/) and [without](https://phoenix-next-pr-67.herokuapp.com/campaigns/treats-troops).


resources/assets/components/Markdown/index.js
```diff
@@ -0,0 +1,13 @@
...
+const Markdown = ({className = null, children}) => (
+  <div className={className} dangerouslySetInnerHTML={markdown(children)} />
+);
+
+Markdown.propTypes = {
+  children: React.PropTypes.string.isRequired,
+  className: React.PropTypes.string,
+};
+
+export default Markdown;
```
weerd: coooool :)

resources/assets/containers/ContentPage.js
```diff
@@ -0,0 +1,20 @@
...
+
+const ContentPage = (props) => (
+  <Flex>
+    <FlexCell>
+      <ConnectedContentPage route={props.params} />
+    </FlexCell>
+  </Flex>
+);
+
+export default ContentPage;
```
weerd: Did you want to append `Container` to the name of the file to continue w/ that pattern @DFurnes?

DFurnes: Yes! Good call. 🖊

PR #68, opened 2017-03-09
Please affiliate

deadlybutter:
This PR adds a new Revealer component which is used for masking the bottom of the feed as either a view more button or a signup button.

Additionally, I added authentication status to the redux store. *NOTE*: I left the `scriptifyAuth` thing in the helpers still because if I removed that I would have had to refactor all the reaction stuff to use Redux (that card is assigned to me / coming up soon)

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -73,7 +73,9 @@ class CampaignController extends Controller
             'reportbacks' => [
                 'data' => $reportbacks
             ],
-            'user' => Auth::check() ? Auth::user()->northstar_id : null,
+            'user' => [
+                'id' => Auth::check() ? Auth::user()->northstar_id : null,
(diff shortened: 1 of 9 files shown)


deadlybutter:
Also if the review app fails here as well, 

screen shot 2017-03-09 at 4 07 55 pm
screen shot 2017-03-09 at 4 08 29 pm
screen shot 2017-03-09 at 4 09 16 pm

deadlybutter:
[(5) As a developer, I want to limit some of the Unaffiliated view from users so they have to join the campaign to get full access](https://trello.com/c/DQ4JXIho/339-5-as-a-developer-i-want-to-limit-some-of-the-unaffiliated-view-from-users-so-they-have-to-join-the-campaign-to-get-full-access)

deadlybutter:
sweet 🍨

PR #69, opened 2017-03-10
Refactoring reactions to use Redux

deadlybutter:
This PR refactors the reactions components to use Redux

- Adds Reportback Container to connect it with Redux
- Changes up the Reportback Block to reflect the new props structure. 
  - This got quite large, and might be worth breaking up. Thoughts?
- Added some reaction specific actions & reducers to manage that state.
- Updated the reaction component to work with this new structure.


deadlybutter:
@weerd Ready for a review!

resources/assets/actions/index.js
```diff
@@ -10,6 +10,9 @@ export const STORE_REPORTBACK_PENDING = 'STORE_REPORTBACK_PENDING';
 export const STORE_REPORTBACK_SUCESSFUL = 'STORE_REPORTBACK_SUCESSFUL';
 export const ADD_TO_SUBMISSIONS_LIST = 'ADD_TO_SUBMISSIONS_LIST';
 export const CLICKED_VIEW_MORE = 'CLICKED_VIEW_MORE';
+export const USER_LIKED_REPORTBACK = 'USER_LIKED_REPORTBACK';
+export const USER_UNLIKED_REPORTBACK = 'USER_UNLIKED_REPORTBACK';
+export const REACTION_COMPLETE = 'REACTION_COMPLETE';
```
weerd: Just a thought, could we rename these to `USER_TOGGLED_REACTION_ON` and `USER_TOGGLED_REACTION_OFF`? or something that isn't so specific to the idea of a "like" (the heart)?

resources/assets/components/Reaction/index.js
```diff
@@ -5,7 +5,8 @@ import classnames from 'classnames';
 const Reaction = (props) => {
   const active = props.active || false;
   const total = props.total || 0;
-  const onClick = props.onClick || null;
+
+  const onClick = () => active ? props.onUnlike() : props.onLike();
```
weerd: Similar to my comment above about naming to be more generic and not so specific to "like". Thoughts?

resources/assets/actions/index.js
deadlybutter: 👍  sounds good to me

resources/assets/components/Reaction/index.js
deadlybutter: Yeah Ill carry the `toggle` naming convention down to here as well

resources/assets/reducers/reactions.js
```diff
@@ -0,0 +1,49 @@
+import {
+  USER_TOGGLED_REACTION_ON,
+  USER_TOGGLED_REACTION_OFF,
+  REACTION_COMPLETE,
+} from '../actions';
+const update = require('react/lib/update');
```
weerd: not using `import`?

deadlybutter: I copied it from a SO answer, but could make it an import to make it fall in line with everything else for sure

weerd: also why are you calling `update()` directly in each return?

deadlybutter: I can stash it in a variable but seemed redundant?

deadlybutter:
@DFurnes Refactored to use async actions!

resources/assets/actions/index.js
```diff
@@ -10,6 +10,10 @@ export const STORE_REPORTBACK_PENDING = 'STORE_REPORTBACK_PENDING';
 export const STORE_REPORTBACK_SUCESSFUL = 'STORE_REPORTBACK_SUCESSFUL';
 export const ADD_TO_SUBMISSIONS_LIST = 'ADD_TO_SUBMISSIONS_LIST';
 export const CLICKED_VIEW_MORE = 'CLICKED_VIEW_MORE';
+export const USER_TOGGLED_REACTION = 'USER_TOGGLED_REACTION';
+export const USER_TOGGLED_REACTION_ON = 'USER_TOGGLED_REACTION_ON';
```
DFurnes: These two constants (`USER_TOGGLED_REACTION_ON` and `USER_TOGGLED_REACTION_OFF`) can be removed, yeah?

deadlybutter: Oh yep, nice catch

PR #73, opened 2017-03-13
Quick refactors to clean some code up.

weerd:
This PR just does some quick refactors for a few lines of code. 😄 

_Note:_ the change in **app/Repositories/CampaignRepository.php** was made because when the new `Campaign` is created, the extended `Entity` class calls `setLocal()` within the constructor; didn't think it made sense to call it on the `DynamicEntity` and then again when the `Entity` class is instantiated.

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -58,14 +58,15 @@ class CampaignController extends Controller
         $campaign = $this->campaignRepository->findBySlug($slug);
         $phoenixNid = data_get($campaign, 'legacyCampaignId', '1144');
 
-        $userId = Auth::check() ? Auth::user()->northstar_id : '';
+        $userId = Auth::check() ? Auth::user()->northstar_id : null;
 
         $response = $this->phoenixLegacy->getAllReportbacks([
             'campaigns' => $phoenixNid,
             'status' => 'promoted',
             'load_user' => true,
-            'as_user' => $userId
+            'as_user' => $userId ?: ''
         ]);
+
         $reportbacks = $response['data'];
 
         $state = [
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/CampaignController.php
```diff
@@ -58,14 +58,15 @@ public function show($slug)
         $campaign = $this->campaignRepository->findBySlug($slug);
         $phoenixNid = data_get($campaign, 'legacyCampaignId', '1144');
 
-        $userId = Auth::check() ? Auth::user()->northstar_id : '';
+        $userId = Auth::check() ? Auth::user()->northstar_id : null;
```
DFurnes: This could be replaced with `Auth::id()` and inlined in the two places it's used.

DFurnes: (Although for that, we may need to override `getAuthIdentifierName` on the user.)

weerd: I think in an old PR I tried changing it to `Auth::id()` before and it doesn't default to the northstar_id. 

It needs to be an empty string in the query sent to PhoenixAshes which is why I did the change to make it `null` and then use the shorthand ternary to change it to empty string in the query stuff.

Random, we could use the helper method instead of the facade `auth()->user()`...

deadlybutter: I once fiddled with changing the `Auth::id()` in a PR here but i ended up causing unintended bugs, just used `->northstarId` like Diego has here

DFurnes: Right, `Auth::id()` will return the default `id` column unless you override that method on the User model. Just something to consider in the future! 😄

PR #75, opened 2017-03-13
Generating the form data to send via Fetch request.

weerd:
This PR builds out the `FormData` object needed to send the submitted Reportback data via a fetch() request.

The `FormData` object is built dynamically since we don't just grab the entire HTML form and toss it into the method.

### DIFF
--- a/resources/assets/components/ReportbackUploader/index.js
+++ b/resources/assets/components/ReportbackUploader/index.js
@@ -38,16 +38,38 @@ class ReportbackUploader extends React.Component {
       photo: this.state.media,
       caption: this.caption.value,
       impact: this.impact.value,
-      why_participated: this.why_participated.value
+      whyParticipated: this.why_participated.value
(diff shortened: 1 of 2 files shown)


resources/assets/components/ReportbackUploader/index.js
```diff
@@ -50,6 +50,28 @@ class ReportbackUploader extends React.Component {
     });
   }
 
+  setFormData(reportback) {
+    let formData = new FormData;
+
+    Object.keys(reportback).map((item) => {
```
deadlybutter: Are we worried about validating what people can append to this request? I think as long as we have server side we should be fine for now...?

weerd: Yeah, validation is going to be coming later. Server side first and then eventually @DFurnes and I want to try and add the updated DS Validation package :)

PR #76, opened 2017-03-14
Fix login error when running on PHP 7.1.

DFurnes:
#### Changes
This fixes an issue with logging in to [next.dosomething.org](https://next.dosomething.org), because of an issue in one of the dependencies of the [`league/oauth2-client`](https://github.com/thephpleague/oauth2-client) which Gateway relies on.

I've also commented out the `drupal_id` sync code for now, since we want to avoid relying on a local copy of that field if possible & I was running into some lil' issues with getting it to save on auth.

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -31,7 +31,7 @@ class AuthController extends Controller
      */
     public function getLogin(ServerRequestInterface $request, ResponseInterface $response)
     {
-        return gateway('northstar')->usingGrant('authorization_code')->authorize($request, $response, $this->redirectTo);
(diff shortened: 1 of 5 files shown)


DFurnes:
This has been also been tested on [next.dosomething.org](https://next.dosomething.org) by [deploying this branch](https://envoyer.io/projects/17799/deployments/1414420).

DFurnes:
[(3) 🐛 As a developer, I want to look into why login/logout isn't working on http://52.3.246.166/campaigns/teens-for-jeans so that users can login/logout. 🐛](https://trello.com/c/AUJ0Dja4/373-3-as-a-developer-i-want-to-look-into-why-login-logout-isn-t-working-on-http-52-3-246-166-campaigns-teens-for-jeans-so-that-users)

PR #77, opened 2017-03-14
Creating signups actions/reducers and connecting it to the feed

deadlybutter:
Before I went ahead and did _everything_ involved with signups, I figured I would lay out my current idea for how it could work & get the feedback early on. (so **not ready for merge**)

_assumptions_
* we wanted to avoid having phoenix signups endpoint block the page from loading
* it would be great if we could cache the sign ups somewhere to avoid repeating the sign ups request every page load
 * (IMO) we shouldn't be competing with Rogue and replicating all of this data in our database?

_solution_
* cache all the of the users signups in local storage _(TODO: Make the key user based)_
* on page load set the signup status based on local storage
* if the campaign isnt cached locally make an API request to check the status
* when a user clicks signup, update the local storage

If you check out the CampaignFeed component you'll be able to see where these onLoad actions are fired


DFurnes:
Yeah, agreed on minimizing our reliance on legacy Phoenix & not duplicating Rogue functionality. This seems like a totally sound approach to me. It might be worthwhile to look into something like [redux-localstorage](https://github.com/elgerlambert/redux-localstorage) to handle the details of persisting state, but definitely not required for an initial merge.

deadlybutter:
Cool, will totally bookmark that one to look into!

deadlybutter:
Something to also note, the `/api/v1/signups` index is totally borked on Phoenix, copying from Slack yesterday

```
RDS DB load is really heavy right now. I'm about to check.
@joe There's some long-running queries I'm analyzing
Seeing this query taking about 30-45 seconds to run:
EXPLAIN  
SELECT s.sid AS sid, s.uid AS uid, s.nid AS nid, s.run_nid AS run_nid, s.timestamp AS timestamp, rb.rbid AS rbid, ns.field_northstar_id_value AS field_northstar_id_value
FROM 
dosomething_signup s
LEFT OUTER JOIN dosomething_reportback rb ON s.uid = rb.uid and s.nid = rb.nid and s.run_nid = rb.run_nid
LEFT OUTER JOIN field_data_field_northstar_id ns ON ns.entity_id = s.uid
ORDER BY sid DESC;
```

Its also happening on Thor, but its _not_ happening on staging. I'm not sure why it differs between environments besides the underlying data.

Regardless, Phoenix is blocking this work & will need to do either an investigation of why its borked or make a temp workaround (And since Rogue is going be online soon..(?) I lean towards the workaround option)

deadlybutter:
Think I found a good workaround for now actually, the user activity endpoint

https://www.dosomething.org/api/v1/users/2632038/activity

Will need to restructure the code in this PR _a tad_ but should look ok

deadlybutter:
Oh, I need to update it to support NS Id's tho

resources/assets/components/CampaignFeed/index.js
```diff
@@ -87,13 +98,17 @@ class CampaignFeed extends React.Component {
...
     const authenticated = this.props.user.id !== null;
-    const title = authenticated ? 'view more' : 'sign up';
-    const callToAction = authenticated ? '' : this.props.campaign.callToAction;
-    const onReveal = authenticated ? this.props.clickedViewMore : () => window.location.href = '/login';
+    const signedUp = this.props.signups.status;
+
+    const title = authenticated && signedUp ? 'view more' : 'sign up';
+    const callToAction = authenticated && signedUp ? '' : this.props.campaign.callToAction;
+
+    const onReveal = authenticated ? (
```
deadlybutter: I'm thinking there might be a cleaner way of doing this. I know [one of the upcoming cards for CTA blocks](https://trello.com/c/uJEgoIBB/272-5-as-a-developer-i-want-to-build-the-three-cta-blocks-that-have-been-selected-for-the-v1-launch-so-that-we-can-prompt-people-to-) will involve this logic again, so maybe we can talk more about it there?

resources/assets/components/Revealer/revealer.scss
```diff
@@ -15,11 +15,12 @@
   &::before {
     content: '';
     display: block;
-    background: linear-gradient(to bottom, rgba(221, 221, 221, 0.25), rgba(34, 34, 34, 0.9));
+    background: linear-gradient(to bottom, rgba(150, 150, 150, 0), rgba(0, 0, 0, 0.6));
```
deadlybutter: These style changes have nothing to do with the PR, but i felt like fiddling with this more haha

app/Services/PhoenixLegacy.php
```diff
@@ -156,7 +157,25 @@ public function storeReaction($reportback_item_id, $term_id, $user_id) {
...
+     * @param  string $user_id      User Northstar ID
+     * @param  string $campaign_id  Legacy drupal campaign id
+     * @return array               API response
+     */
+    public function getActivity($user_id, $campaign_id)
+    {
+        $path = 'v1/users/' . $user_id . '/activity';
+        $query = ['nid' => $campaign_id];
+
+        return remember(make_cache_key('legacy-'.$path, $query), $this->cacheExpiration, function() use ($path, $query) {
```
deadlybutter: I also debated whether this one should be cached, the more I think about it the more I lean no, but maybe there is a good reason for doing so?

deadlybutter:
@weerd @DFurnes Sorry this got so large, ready for a look!

resources/assets/components/CampaignFeed/index.js
```diff
@@ -14,6 +14,17 @@ class CampaignFeed extends React.Component {
     this.reportbackIndex = 0;
   }
 
+  componentDidMount() {
+    const campaignId = this.props.campaign.legacyCampaignId.toString();
```
DFurnes: I'm not super excited about the `.toString()` call here. Is that because we store campaign IDs in different formats when returned from different API responses? If so, can we update the Contentful field to match Dr00ps?

resources/assets/components/CampaignFeed/index.js
```diff
@@ -14,6 +14,17 @@ class CampaignFeed extends React.Component {
...
+    const existingSignups = this.props.signups.data;
+
+    if (existingSignups.indexOf(campaignId) > -1) {
```
DFurnes: Not a blocker for merging this pull request, but we should try to be careful about keeping "data" logic in reducers or "container" components, and have these only be responsible for filtering data for display.

resources/assets/actions/index.js
```diff
@@ -115,3 +117,46 @@ export function fetchReportbacks(node, page) {
...
+  })
+  .then(response => {
+    if (!response || !response[0]) return;
+
+    dispatch({
+      type: SIGNUP_COMPLETE,
+      campaignId,
+    });
+
+    dispatch({
```
DFurnes: Are these two actions always dispatched together? If so, would it make sense to combine?

resources/assets/actions/index.js
```diff
@@ -115,3 +117,46 @@ export function fetchReportbacks(node, page) {
...
+// Action: set whether the user is signed up for this campaign
+export function setCurrentlySignedUp(status) {
+  return { type: SET_CURRENTLY_SIGNED_UP, status };
+}
+
+// Async Action: check if user already signed up for the campaign
+export function checkForSignup(campaignId) {
+  return dispatch => (new Phoenix).get(`activity/${campaignId}`)
+    .then(response => {
+      if (!response || !response.sid) return;
```
deadlybutter: Yeah I think we should probably make a card for some kind of error notification system, so every component can just push error messages up to the store and have some system for displaying them

resources/assets/actions/index.js
deadlybutter: Yeah, if you look in CampaignFeed you'll see `this.props.setCurrentlySignedUp(true);` is called by itself. https://github.com/DoSomething/phoenix-next/pull/77/files#diff-6a1450808ef7073816a941fa205f5915R127

Maybe there is a way to combine the logic for just the action functions here?

resources/assets/components/CampaignFeed/index.js
deadlybutter: Yeah contentful is returning an int if i remember right, Ill go ahead and change that

resources/assets/components/CampaignFeed/index.js
deadlybutter: Gotcha- wasn't sure of a better place to put this

resources/assets/actions/index.js
deadlybutter: @DFurnes given ^, thoughts on something like this?

```js
function postSignup(dispatch, campaignId) {
  dispatch({
    type: SIGNUP_COMPLETE,
    campaignId,
  });

  dispatch({
    type: SET_CURRENTLY_SIGNED_UP,
    status: true,
  });
}

// Async Action: check if user already signed up for the campaign
export function checkForSignup(campaignId) {
  return dispatch => (new Phoenix).get(`activity/${campaignId}`)
    .then(response => {
      if (!response || !response.sid) return;

      postSignup(dispatch, campaignId);
    });
}

// Async Action: send signup to phoenix.
export function clickedSignUp(campaignId) {
  return dispatch => (new Phoenix).post('signups', {
    campaignId,
  })
  .then(response => {
    if (!response || !response[0]) return;

    postSignup(dispatch, campaignId);
  });
}
```

resources/assets/actions/index.js
DFurnes: That makes sense... I think we may also want to have action-specific `isRequesting` state for things like this, so we can show appropriate spinners in the interface when a Phoenix request takes a long time. ⌛️

deadlybutter: Oh yep i can add that

resources/assets/actions/index.js
DFurnes: If the events need to be separate, I think it's fine to keep them that way so we're not adding any unnecessary levels of abstraction. One thing you can do to reduce some noise is use the action creators for each of those events:

```js
function signupComplete(campaignId) {
  return {type: SIGNUP_COMPLETE, campaignId};
}

// Async Action: check if user already signed up for the campaign
export function checkForSignup(campaignId) {
  return dispatch => (new Phoenix).get(`activity/${campaignId}`)
    .then(response => {
      if (!response || !response.sid) return;

      dispatch(signupComplete(campaignId));
      dispatch(setCurrentlySignedUp(true));
    });
}

// Async Action: send signup to phoenix.
export function clickedSignUp(campaignId) {
  return dispatch => (new Phoenix).post('signups', {campaignId})
    .then(response => {
      if (!response || !response[0]) return;

      dispatch(signupComplete(campaignId));
      dispatch(setCurrentlySignedUp(true));
    });
}

```

deadlybutter:
Here is the new `pending` property in action (We can do the spinners and stuff later I presume)

screen shot 2017-03-15 at 1 58 06 pm

deadlybutter:
🚂 🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋

CHOO CHOO, PR COMIN THROUGH

PR #78, opened 2017-03-14
RB Upload form submission.

weerd:
This PR finalizes the process of submitting a reportback via the ReportbackUploader component.

The process involved is as follows:

- user uploads photo via the MediaUploader.
- user enters data in additional fields.
  - the additional fields are current all required using server side validation; the `response` status is checked and if there's a validation error from missing fields, a redux action is dispatched, but not really anything else (need to include client-side validation and/or handling this action).
- user submits the RB form.
- all the data is passed to Phoenix-Next which acts as a proxy to send it off to Phoenix-Ashes.
- prior to proxying the data, PN will generate a Data URI for the uploaded image as well as generate a filename with a timestamp.
- if the submission is successful, an action is dispatched to indicate it was successful as well as another action to add the RB to the submission gallery below the uploader (this functionality was already done in a prior PR).

### Things to note
I tested submitting more than one RB in the session and seems to work fine. We do seem to have some errors on submitting the RB photo as a super long Data URI via JSON to Phoenix-Ashes; the Data URI (if it's a big enough image) will get cut-off so the image is broken when it arrives at PA. Something we'll need to address and @DFurnes and I have chatted about it a bit today.

Another thing to note it that the upload process will fail if the image is larger than a certain "safe" size (it might just be it's failing at being too big to convert to a Data URI, but not 100% sure).

[Gateway-JS](https://github.com/DoSomething/gateway-js) currently doesn't support overriding the headers for the client, so I had to pull back and use `fetch()` directly. Once Gateway-JS is updated I can update the `fetch()` code to help simplify things :)

https://trello.com/c/uHyDi2RL/264-13-as-a-user-i-want-to-be-able-to-reportback-on-phoenix-next-hold-on-cropping-though-so-that-users-can-rb-on-phoenix-next

### DIFF
--- a/app/Http/Controllers/ReportbackController.php
+++ b/app/Http/Controllers/ReportbackController.php
@@ -38,9 +38,27 @@ class ReportbackController extends Controller
      */
     public function store(Request $request)
     {
-        $data = $request->input('caption');
+        // $this->validate($request, [
+        //     'photo' => 'required',
+        //     'caption' => 'required',
+        //     'impact' => 'required',
+        //     'whyParticipated' => 'required',
+        // ]);
 
-        return response()->json(['data' => $data]);
+        $reportbackPhoto = $request->file('photo');
+
+        return $this->phoenixLegacy->storeReportback(
+            auth()->user()->legacy_id,
+            $request->input('campaignId'),
+            [
+                'file' => make_data_uri($reportbackPhoto->getPathname(), $reportbackPhoto->getMimeType()),
+                'filename' => time().'_reportback_photo.'.$reportbackPhoto->guessClientExtension(),
+                'caption' => $request->input('caption'),
+                'quantity' => $request->input('impact'),
+                'why_participated' => $request->input('whyParticipated'),
+                'source' => 'phoenix-next',
+            ]
+        );
     }
 
     /**
--- a/app/Http/Controllers/ReportbackController.php
+++ b/app/Http/Controllers/ReportbackController.php
@@ -38,12 +38,12 @@ class ReportbackController extends Controller
      */
     public function store(Request $request)
     {
-        // $this->validate($request, [
-        //     'photo' => 'required',
-        //     'caption' => 'required',
-        //     'impact' => 'required',
-        //     'whyParticipated' => 'required',
-        // ]);
+        $this->validate($request, [
+            'photo' => 'required',
+            'caption' => 'required',
+            'impact' => 'required',
+            'whyParticipated' => 'required',
+        ]);
 
         $reportbackPhoto = $request->file('photo');
 
--- a/app/Http/Controllers/ReportbackController.php
+++ b/app/Http/Controllers/ReportbackController.php
@@ -16,6 +16,8 @@ class ReportbackController extends Controller
     public function __construct(PhoenixLegacy $phoenixLegacy)
     {
         $this->phoenixLegacy = $phoenixLegacy;
+
+        $this->middleware('auth', ['only' => ['store']]);
     }
 
     /**
@@ -36,7 +38,9 @@ class ReportbackController extends Controller
      */
     public function store(Request $request)
     {
-        return response()->json('temp response here!');
+        $data = $request->input('caption');
+
+        return response()->json(['data' => $data]);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/ReportbackController.php
```diff
@@ -36,7 +38,27 @@ public function index(Request $request)
...
+            'photo' => 'required',
+            'caption' => 'required',
+            'impact' => 'required',
+            'whyParticipated' => 'required',
+        ]);
+
+        $reportbackPhoto = $request->file('photo');
+
+        return $this->phoenixLegacy->storeReportback(
+            auth()->user()->legacy_id,
```
DFurnes: This doesn't need to happen before merging this pull request, but let's add a `@TODO` to update the Phoenix endpoint to accept a Northstar ID here like the other ones.

resources/assets/actions/index.js
```diff
@@ -94,13 +95,31 @@ export function submitReportback(reportback) {
   return dispatch => {
     dispatch(storeReportback(reportback));
 
-    return (new Phoenix).post('reportbacks', reportback)
-      .then(dispatch({
-        type: STORE_REPORTBACK_SUCESSFUL
-      }))
+    const url = `${services.PHOENIX_URL}/reportbacks`;
```
DFurnes: This should be safe to use a relative URL for I think. `window.fetch('/reportbacks', { ...`

resources/assets/actions/index.js
```diff
@@ -94,13 +95,31 @@ export function submitReportback(reportback) {
...
+    })
       .then((response) => {
-        dispatch(addToSubmissionsList(reportback));
-      });
+        if (response.status >= 300) {
+          dispatch({ type: STORE_REPORTBACK_FAILURE })
+          // @TODO: implement showing validation error.
+        }
+        else {
+          dispatch({ type: STORE_REPORTBACK_SUCESSFUL })
```
DFurnes: Only just noticed this, but "SUCESSFUL" is missing a "C" here! ✏️

DFurnes: I'm also tempted to say we should always make action creators for dispatching events like this so things look consistent. For example:

```js
// Action Creator: Stored a reportback successfully
function storeReportbackSuccessful() {
  return { type: STORE_REPORTBACK_SUCESSFUL };
}

// ...
dispatch(storeReportbackSuccessful());
dispatch(addToSubmissionsList(reportback));


```

resources/assets/components/ReportbackUploader/index.js
```diff
@@ -55,23 +56,18 @@ class ReportbackUploader extends React.Component {
 
     Object.keys(reportback).map((item) => {
       if (item === 'photo') {
-        formData.append(item, reportback[item].file, this.makeFileNameForBlob(reportback[item].file));
+        formData.append(item, reportback[item].file);
```
DFurnes: I wonder if you can keep the `filePreviewUrl` internal to the `MediaUploader` so we don't need a conditional here. Or maybe it lives outside the `reportback` that's being iterated? Either way, definitely not necessary for merge!

app/Http/Controllers/ReportbackController.php
deadlybutter: I think i might have done that already, let me double check

deadlybutter: Yep! https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc#L351

weerd: I'll add in a `@TODO` @DFurnes and then make an update to use the NS ID in upcoming PR @deadlybutter so I have time to test but meanwhile merge this in :)

weerd: I intended to add the `@TODO` and then forgot in midst of all the other issues I encountered with this upload process @_@

resources/assets/actions/index.js
weerd: gah!

weerd: As for recommendation that sounds good to me. Will update :)

resources/assets/components/ReportbackUploader/index.js
weerd: hmmm, yeah, I think we could do something like that or at minimum have maybe 2 separate properties? The `reportback` object gets passed around a bit to then get tossed into the submission list. The preview URL is need for the submission gallery but the blob is needed for the upload to PA. :man_shrugging: I can think about a better approach for a future PR. But makes sense and would defs simplify this.

PR #79, opened 2017-03-14
First pass at the static block.

DFurnes:
#### Changes
This pull request includes a first pass at the "static block", which displays a title, markdown content, and an optional "source" at the bottom. For now, I did some negative margin hackery & hard-coded that wonderful red from the mocks – it'll makes sense to clean that up once we have a clearer creative direction in mind next week. 😄

screen shot 2017-03-14 at 5 14 10 pm

🦊

### DIFF
--- a/resources/assets/components/StaticBlock/static-block.scss
+++ b/resources/assets/components/StaticBlock/static-block.scss
@@ -1,0 +1,22 @@
+@import '~@dosomething/forge/scss/toolkit';
+
+.static-block__title {
+  color: #fff;
+  background-color: #e8273e; // red, the color of desire ♬
+  text-transform: uppercase;
+  margin: (- $base-spacing / 2) (- $base-spacing / 2) $base-spacing;
+  padding: $base-spacing / 2;
+}
+
+.static-block__citation {
+  margin-top: $base-spacing;
+
+  &::before {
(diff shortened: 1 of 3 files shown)


DFurnes:
Unrelated thought – we do `$base-spacing / 2` everywhere. Let's think of a name for that.

DFurnes:
Not quite sure where the "had a problem deploying" comes from – here's the [review app](https://phoenix-next-pr-79.herokuapp.com/campaigns/teens-for-jeans).

resources/assets/components/StaticBlock/static-block.scss
```diff
@@ -0,0 +1,22 @@
+@import '~@dosomething/forge/scss/toolkit';
+
+.static-block__title {
+  color: #fff;
+  background-color: #e8273e; // red, the color of desire ♬
+  text-transform: uppercase;
+  margin: (- $base-spacing / 2) (- $base-spacing / 2) $base-spacing;
```
deadlybutter: Did we ever think of a name for this?

DFurnes: My fav so far is `$half-spacing`, but I don't think we ever settled on a choice.

PR #81, opened 2017-03-15
Refactoring file and directory naming

weerd:
This PR does some quick file renaming and file/directory shuffling to be more consistent with the following standard pattern we are starting to follow:

- Redux related _container_ components are housed in the `/resources/assets/containers` directory and should be suffixed with _SomeNameHere**Container**.js_.
- Components that don't behave as containers are all housed in the `/resources/assets/components` directory.

The above standard is based on our approach/recommended pattern for Redux.

Might be useful if reviewed commit by commit. It mostly goes in some sort of logical order :)


deadlybutter:
Cool w/ me!

PR #84, opened 2017-03-17
Redirect to expected page after login or logout.

DFurnes:
#### Changes
__This pull request updates the `/login` route to redirect back to the expected page after a user logs in to their account.__ This is either the "intended" page (if they are sent to log in via the `auth` middleware on a protected route) or the previous page (if they just hit "Log In" in nav).

This should probably be moved into Gateway in the future.

Test [logging in from the campaign page](https://phoenix-next-pr-84.herokuapp.com/campaigns/teens-for-jeans)! Are there any other use-cases I'm missing?

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -2,14 +2,14 @@
 
 namespace App\Http\Controllers;
 
-use App\Http\Controllers\Controller;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
 class AuthController extends Controller
 {
     /**
-     * Where to redirect users after login.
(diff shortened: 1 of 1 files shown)


DFurnes:
[(3) As a user, when I login via Northstar on Phoenix-next, I want it it to refresh to that URL so that I'm looking at the URL I was expecting to look at.](https://trello.com/c/tcbAA2BL/360-3-as-a-user-when-i-login-via-northstar-on-phoenix-next-i-want-it-it-to-refresh-to-that-url-so-that-i-m-looking-at-the-url-i-was-)

DFurnes:
Another thing we should consider before launch: when a user logs out of Phoenix Next, they'll still have an active Phoenix Ashes session. We should make sure that we remove both sessions so a user doesn't accidentally stay logged in.

PR #85, opened 2017-03-17
Metrics

deadlybutter:
- Adds Google Analytics + KeenIO libraries
- Adds DS-Analytics library
- Removes old auth scriptify code 

TODO
- Should probably only load these analytics libraries if its prod
- Need to provide a way of attaching phoenix-next meta data to all events

OPEN THOUGHTS/QUESTIONS
- All Redux actions trigger an event (idea courtesy of Dave, Im a fan). Not sure in practice what this looks like
- Parts of the store (User info, Block index, Signup status, etc) should probably be a part of the default metadata. Need to think an intuitive way of linking everything together


resources/assets/actions/metrics.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+export const POST_EVENT = 'POST_EVENT';
+
+/**
+ * Action Creators: these functions create actions, which describe changes
+ * to the state tree (either as a result of application logic or user input).
+ */
+
+ // Async Action: post the given data to the collection
+ export function postEvent(collection, data) {
```
deadlybutter: not sure i want to keep it / implement this way, just playing with the idea of it

deadlybutter:
ah, build failed because i forgot to actually publish ds-analytics 2.0, i have it npm linked still...

deadlybutter:
@DFurnes Removed some of the temp / playground code from this PR so we can merge it in, hopefully at sprint planning today we confirm what events / work we'll be doing for event tracking & Ill be able to make a better approach.

Should hopefully be a quick review!

resources/assets/app.js
```diff
@@ -15,6 +15,13 @@ import { ready } from './helpers';
 // WHATWG Fetch Polyfill
 import 'whatwg-fetch';
 
+import { init } from '@dosomething/analytics';
+//TODO: Only enable analytics on prod?
```
DFurnes: I think it's a good idea to enable the analytics package everywhere, but only add the relevant snippets to "connect" it to external services in production, since it's nice to see events firing in the console.

resources/assets/app.js
```diff
@@ -15,6 +15,13 @@ import { ready } from './helpers';
...
+init('track', true, {
+  projectId: '58cbf99954532c1f35f4331f',
```
DFurnes: It feels like these should be coming from environment variables set during the build in case we want to use separate project IDs per environment (for example, so we don't get "bad" data from testing pull requests on Heroku).

resources/views/layouts/master.blade.php
```diff
@@ -44,8 +44,18 @@
     </div>
 </div>
 
+<script>
```
DFurnes: I would add the `UA-XXXXXX-1` string as a config setting so we can customize that between environments, and then conditionally include this block based on whether or not that environment variable is provided. Take a look at [voting-app](https://github.com/DoSomething/voting-app/blob/a1a794b0ba35673d699be151b195171237fb02ef/resources/views/partials/analytics.blade.php#L1-L20) as an example.

resources/views/layouts/master.blade.php
```diff
@@ -44,8 +44,18 @@
...
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+  ga('create', 'UA-493209-1', 'auto');
+  ga('send', 'pageview');
```
DFurnes: We may actually not want to trigger this here since all of our routing for the new campaign template happens within the user's browser. Instead, try using adding a listener to the [router history](https://github.com/DoSomething/phoenix-next/blob/916e0ab8713869bf733447b2af5c3174de37cdf9/resources/assets/components/App.js#L20):

```js
history.listen(location => {
    if (! window.ga) {
        return;
    }

    window.ga('send', 'pageview', location.pathname);
});
```

resources/views/partials/analytics.blade.php
```diff
@@ -0,0 +1,15 @@
+@if(env('GOOGLE_ANALYTICS_ID'))
```
DFurnes: These should be stored in [`config/services.php`](https://github.com/DoSomething/phoenix-next/blob/dev/config/services.php). Calling `env()` outside of config files [can cause issues](https://laravel.com/docs/5.4/configuration#configuration-caching).

deadlybutter: Oh interesting, 👍

package.json
```diff
@@ -18,6 +18,7 @@
     ]
   },
   "dependencies": {
+    "@dosomething/analytics": "^2.0.0",
```
weerd: `2.0.0` 🎉

deadlybutter:
Just need to move that page event to the proper place for redux, otherwise good to go

deadlybutter:
Should be good to go for another 👀

PR #86, opened 2017-03-20
CTA block updates

weerd:
This PR does a few updates to the `CallToActionBlock` so it can accommodate more variations based on the content entered in Contentful by the Campaign Lead:

image

One thing to note, I originally coded the component to use the new `ContentfulImage` component that inserts an `` tag with proper link to a modified version of the cover image for the campaign using URL params. However, I ended up deciding that for the `CallToActionBlock` component, it would work better (responsively for small/medium/large screens) if the image was added as a custom `style` attribute. I ended up leaving the new `ContentfulImage` component because it could come in handy in the future 🔮 

Reference: https://trello.com/c/uJEgoIBB/272-5-as-a-developer-i-want-to-build-the-three-cta-blocks-that-have-been-selected-for-the-v1-launch-so-that-we-can-prompt-people-to-

### DIFF
--- a/resources/assets/components/CallToActionBlock/index.js
+++ b/resources/assets/components/CallToActionBlock/index.js
@@ -1,21 +1,51 @@
 import React from 'react';
+import classnames from 'classnames';
+import { contentfulImageUrl } from '../../helpers';
 import './cta.scss';
 
 const CallToActionBlock = (props) => {
+  const additionalContent = props.fields.additionalContent || {};
+
   //TODO: This should probably be editable in contentful
-  const cta = props.signups.thisCampaign ? 'Reportback' : 'Get Involved';
+  const buttonText = props.signups.thisCampaign ? 'Reportback' : 'Get Involved';
 
   //TODO: This should take you to the action page if you're signed up
   const onClick = () => props.clickedSignUp(props.campaign.legacyCampaignId);
 
+  const backgroundImageStyle = {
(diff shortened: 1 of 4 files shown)


resources/assets/components/CallToActionBlock/index.js
```diff
@@ -1,21 +1,51 @@
...
+  const buttonText = props.signups.thisCampaign ? 'Reportback' : 'Get Involved';
 
   //TODO: This should take you to the action page if you're signed up
   const onClick = () => props.clickedSignUp(props.campaign.legacyCampaignId);
 
+  const backgroundImageStyle = {
+    backgroundImage: `url(${contentfulImageUrl(props.campaign.coverImage.url, '400', '400', 'fill')})`,
+  };
+
+  const impactContent = () => {
```
DFurnes: I think this might make more sense as a nullable variable than a function.

```js
let impactContent = null;
if (additionalContent.impactNumber) {
    impactContent = (
        <div className="cta__block cta__impact">
            {additionalContent.impactPrefix ? <span className="cta__impact-prefix">{additionalContent.impactPrefix}</span> : null}
            {additionalContent.impactNumber ? <span className="cta__impact-number">{additionalContent.impactNumber}</span> : null}
            {additionalContent.impactMessage ? <span className="cta__impact-message">{additionalContent.impactMessage}</span> : null}
        </div>
    );
}
```

weerd: sounds good :)

PR #87, opened 2017-03-20
Clean up containers & feed component.

DFurnes:
#### Changes
This pull request just includes a wee little tidying up to reduce the amount of boilerplate we need to write in our Redux container components & fixes up some docblocks in the CampaignFeed.

🍃


resources/assets/components/CampaignFeed/index.js
```diff
@@ -114,15 +113,13 @@ class CampaignFeed extends React.Component {
   }
 
   render() {
-    console.log(this.props.signups)
```
weerd: hehe

resources/assets/components/ContentPage/index.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import Block from '../Block';
+import Markdown from '../Markdown';
+import { Flex, FlexCell } from '../Flex';
+
+const ContentPage = ({ pages, route }) => {
```
weerd: Nice, yeah... I guess it didn't make sense for these to be called `PageBlock`s, when not in the feed!

resources/assets/components/CampaignFeed/index.js
deadlybutter: What, who would leave that in the render function, that idi-

Oh hey that was me...

resources/assets/containers/ActivityContainer.js
```diff
@@ -1,13 +1,15 @@
-import React from 'react';
 import { connect } from 'react-redux';
+import CampaignFeed from '../components/CampaignFeed';
```
weerd: to help things match up better, should we either rename this file to **CampaignFeedContainer.js** or rename the component to **ActivityFeed.js**? thoughts?

DFurnes: Yeah, that's a good thought. I'm also wondering if we pull that component out into the containers directory, since it's responsibility is getting the data into the right structure to render the feed, and not really presentational...

DFurnes: I'm going to think more about that while I work on infini-reportbacks.

weerd: pulling out the component sounds reasonable. either way, cool to hold off and make decision while working on infinite-RBs like you mentioned :)

PR #88, opened 2017-03-21
Rearranging actions

weerd:
This PR separates out Reportback related actions into its own file for easier organization. Not sure if this is ideal, but wanted to take a crack at it and see if it looks like it could help keep actions a bit more manageable!

I tested submitting a Reportback and all seemed to work fine, which is a good sign.

### DIFF
--- a/resources/assets/actions/reportbackActions.js
+++ b/resources/assets/actions/reportbackActions.js
@@ -1,13 +1,11 @@
-/*
- * Action names: import these constants to dispatch an event
- * without having hardcoded strings all about.
- */
-export const REQUESTED_REPORTBACKS = 'REQUESTED_REPORTBACKS';
-export const RECEIVED_REPORTBACKS = 'RECEIVED_REPORTBACKS';
-export const STORE_REPORTBACK_PENDING = 'STORE_REPORTBACK_PENDING';
-export const STORE_REPORTBACK_FAILED = 'STORE_REPORTBACK_FAILED';
-export const STORE_REPORTBACK_SUCCESSFUL = 'STORE_REPORTBACK_SUCCESSFUL';
-export const ADD_TO_SUBMISSIONS_LIST = 'ADD_TO_SUBMISSIONS_LIST';
+import {
+  REQUESTED_REPORTBACKS,
+  RECEIVED_REPORTBACKS,
+  STORE_REPORTBACK_PENDING,
+  STORE_REPORTBACK_FAILED,
+  STORE_REPORTBACK_SUCCESSFUL,
+  ADD_TO_SUBMISSIONS_LIST
+} from '../actions';
 
 // Action: reportback fetch initiated.
 export function requestingReportbacks(node) {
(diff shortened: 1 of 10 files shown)


resources/assets/containers/ReportbackUploaderContainer.js
```diff
@@ -1,6 +1,6 @@
 import { connect } from 'react-redux';
 import ReportbackUploader from '../components/ReportbackUploader';
-import { submitReportback, addToSubmissionsList } from '../actions';
+import { submitReportback, addToSubmissionsList } from '../actions/reportbackActions';
```
deadlybutter: Can you `export * from './reportbackActions';` inside the `actions/index` -- this way we only ever have to require a single file for actions

resources/assets/actions/reportbackActions.js
```diff
@@ -0,0 +1,92 @@
+/*
+ * Action names: import these constants to dispatch an event
+ * without having hardcoded strings all about.
+ */
+export const REQUESTED_REPORTBACKS = 'REQUESTED_REPORTBACKS';
```
DFurnes: I wonder if we should still have one file for constants so we can be ensure we don't have any overlapping names.

resources/assets/containers/ReportbackUploaderContainer.js
weerd: I had that originally but I think you have to give it an alias, like:

```
export * as ReportbackActions from './reportbackActions';
```
Which means when importing they'd be props on a `ReportbackActions` object, right? But I may be confused...

resources/assets/actions/reportbackActions.js
weerd: hmmm, yeah. Maybe that's what the `index.js` file is for, holding all the constants? :man_shrugging:

resources/assets/containers/ReportbackUploaderContainer.js
deadlybutter: Try it? Not sure

deadlybutter: @weerd If that doesn't work, maybe something like this

```js
import * as reportbackActions from './reportbackActions';
export reportbackActions;
```

DFurnes: Yeah, I think you should be able to do multiple `export * from`s. I wonder what the error messaging looks like if you end up with naming collisions there.

resources/assets/actions/reportbackActions.js
DFurnes: Yeah, that seems like a smart idea. `index.js` is the sorta like the table of contents then.

weerd:
K! Thanks for the input. I'll rearrange a bit more based on your suggestions and see how it goes. I'll keep it to just the RB related actions for now.

weerd:
@DFurnes @deadlybutter thanks for the suggestions. I was able to switch it to export all the functions from `reportbackActions.js` via `actions/index.js`, which makes it nicer to just needing to import the single `./actions` like you mentioned @deadlybutter

weerd:
Lemme know if this looks cool and I can update the rest as well in following commits. :)

deadlybutter:
Oh shoot thats what your comment was haha

weerd:
Yeah, I can do that in this PR, just wanted to pick a direction first :)

PR #90, opened 2017-03-21
Add "lazy loaded" and "base figure" components.

DFurnes:
#### Changes
This is a little chunk of work towards getting the infini-reportbacks view implement. It adds a "base figure" class so we can appropriately use that pattern to build interface elements, and a "lazy loaded" class that allows content to be swapped out based on an `isFetching` state.

A lil' peek at the (currently unused) "reportback placeholder":

screen shot 2017-03-21 at 1 57 03 pm

🐇


resources/assets/components/ReportbackBlock/index.js
```diff
@@ -1,23 +1,42 @@
 import React from 'react';
 import Block from '../Block';
 import { FlexCell } from '../Flex';
-import { Phoenix } from '@dosomething/gateway';
+import { Figure, BaseFigure } from '../Figure';
 import Reaction from '../Reaction';
+import LazyLoaded from '../LazyLoaded';
+
+const ReportbackItemPlaceholder = () => {
```
deadlybutter: Hm, couldnt this just be a `ReportbackItemContents` with those default properties?

DFurnes: Oh, that's an interesting idea. I'll give that a try!

PR #91, opened 2017-03-21
Affirmation!

deadlybutter:
_HEY, YOU DID IT! HIGH FIVE!_

This PR adds the affirmation component, which shows up after you signup for a campaign. 

mar-21-2017 14-07-57

The placement, style, fields, etc are *not* final, we are waiting on further direction from design/creative.

### DIFF
--- a/resources/assets/components/Affirmation/affirmation.scss
+++ b/resources/assets/components/Affirmation/affirmation.scss
@@ -1,1 +1,22 @@
 @import '~@dosomething/forge/scss/toolkit';
+
+.affirmation {
+  width: 100%;
+  text-align: center;
+  padding: $base-spacing;
+  background: $off-black;
+  color: $white;
+  position: relative;
+
+  a {
+    color: $white;
+    position: absolute;
+    top: 0;
+    right: 0;
+    padding: $base-spacing / 2;
+
+    &:hover {
+      text-decoration: none;
+    }
+  }
+}
--- a/resources/assets/components/Affirmation/affirmation.scss
+++ b/resources/assets/components/Affirmation/affirmation.scss
@@ -1,0 +1,1 @@
+@import '~@dosomething/forge/scss/toolkit';
--- a/resources/assets/components/Affirmation/affirmation.scss
+++ b/resources/assets/components/Affirmation/affirmation.scss
@@ -14,6 +14,7 @@
     top: 0;
     right: 0;
     padding: $base-spacing / 2;
+    font-size: 26px;
 
     &:hover {
       text-decoration: none;
--- a/resources/assets/components/Affirmation/index.js
+++ b/resources/assets/components/Affirmation/index.js
@@ -25,7 +25,7 @@ class Affirmation extends React.Component {
         <div className="affirmation">
           <h1>{ this.props.title }</h1>
           <p>{ this.props.description }</p>
-          <a href="#" onClick={this.onClick}>X</a>
+          <a href="#" onClick={this.onClick}>x</a>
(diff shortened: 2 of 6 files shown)


resources/assets/components/Affirmation/affirmation.scss
```diff
@@ -0,0 +1,22 @@
...
+
+.affirmation {
+  width: 100%;
+  text-align: center;
+  padding: $base-spacing;
+  background: $off-black;
+  color: $white;
+  position: relative;
+
+  a {
```
deadlybutter: I know we could probably use an "X" icon eventually if we wanted to, but figured this was fine for now

deadlybutter:
[(5) As a campaign lead, I want an 'Affirmation Component' outside of the feed that I can edit to congratulate and thank users for affiliating with the campaign so that they are affirmed.](https://trello.com/c/REYVcrtA/390-5-as-a-campaign-lead-i-want-an-affirmation-component-outside-of-the-feed-that-i-can-edit-to-congratulate-and-thank-users-for-aff)

resources/assets/components/Affirmation/affirmation.scss
weerd: Quick fender trick, use the `&times;` symbol (multiplication x symbol), it's a symbol that isn't affected by the chosen font, so the "x" is more symmetrical and uniform w/ consistent line-height 😉

resources/assets/components/Affirmation/index.js
```diff
@@ -0,0 +1,40 @@
...
+
+  render() {
+    if (!this.state.open) return null;
+
+    return (
+      <FlexCell width="full">
+        <div className="affirmation">
+          <h1>{ this.props.title }</h1>
+          <p>{ this.props.description }</p>
+          <a href="#" onClick={this.onClick}>x</a>
```
weerd: try the `&times;` symbol here, instead of "x" 💃

deadlybutter:
@weerd Looks great, thanks for the trick!

screen shot 2017-03-21 at 4 02 35 pm

deadlybutter:
Tempted to maybe bump the font size up a notch tho

deadlybutter:
screen shot 2017-03-21 at 4 05 02 pm

yeah this looks better, ill push the changes up

resources/assets/store.js
```diff
@@ -45,6 +45,7 @@ export default function(reducers, preloadedState = {}) {
     signups: {
       data: (localStorage.getItem('signups') || '').split(','),
       thisCampaign: false,
+      thisSession: false,
```
weerd: Is there a reason we can't just update the `thisCampaign` property to be true on signup? On full page (server) reload it'll be `true` moving forward anyhoo. The `thisSession` property just sounds s bit confusing, and it'll be false while `thisCampaign` will be true post-signup and page reload which seems a bit weird to me.

deadlybutter: Well that was the intent

*signups.thisCampaign* will persist for the life of time once you signup
*signups.thisSession* will only persist for the session in which you signed up

This is incredibly useful to have if we want post signup features to trigger

deadlybutter: Otherwise we'll need some super weird logic to check if you've signed up in a previous session, which im not even sure how we'd do.

deadlybutter: Also thisCampaign is set to true on signup, see the reducer

weerd: Ah, gotcha. I was misreading it a bit. Makes sense :)

PR #92, opened 2017-03-22
Reportback component re-organization.

DFurnes:
#### Changes
My goal with this pull request was to re-organize how data is passed to reportback blocks (since we want to be able to add the blocks to the layout and then "fill them in" later on with real data). I'm going to follow this up with another pull request that normalizes the stores, but wanted to keep this one from getting too overwhelming.

Based on a suggestion from @deadlybutter in #90, I removed the `LazyLoaded` component and instead just modify the appearance of the `ReportbackPhoto` based on it's `isFetching` prop.

Everything should still function correctly on the review app.


resources/assets/components/ReportbackBlock/index.js
```diff
@@ -1,76 +1,28 @@
...
+import './reportback-block.scss';
 
 const ReportbackBlock = props => {
   const items = props.reportbacks.map(reportback => {
-    const item = reportback.reportback_items.data[0];
+    const reportbackItem = reportback.reportback_items.data[0];
     const data = {
-      image: item.media.uri,
-      firstName: reportback.user.first_name,
+      isAuthenticated: props.user.id !== null,
```
deadlybutter: I feel like we've done this in a few places, is it worth adding a shorthand property in the store for this?

resources/assets/actions/index.js
```diff
@@ -21,7 +21,7 @@ export const CLICKED_VIEW_MORE = 'CLICKED_VIEW_MORE';
 export * from './feed';
 
 // Reaction Action Names & Creators
-export const USER_TOGGLED_REACTION = 'USER_TOGGLED_REACTION';
+export const REACTION_CHANGED = 'REACTION_CHANGED';
```
deadlybutter: cool, works for me!

resources/assets/components/ReportbackPhoto/index.js
```diff
@@ -0,0 +1,28 @@
...
+import Reaction from '../Reaction';
+import { ensureAuth } from '../../helpers';
+import './reportback-photo.scss';
+
+const ReportbackPhoto = ({data = {}, isFetching = false, toggleReactionOn, toggleReactionOff}) => {
+  const image = isFetching ? 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7' : data.url;
+  const firstName = isFetching ? '…' : data.firstName;
+  const impact = isFetching ? '…' : `${data.quantity} jeans`;
+
+  const reaction = isFetching ? null : (
```
deadlybutter: Is there a reason we moved this to `ReportbackPhoto` specifically? It just feels odd since it implies (to me) this reaction is specific to photos, when it's specific to the report back

resources/assets/components/ReportbackPhoto/index.js
```diff
@@ -0,0 +1,28 @@
...
+      <Reaction active={data.reaction.reacted} total={data.reaction.total}
+                onToggleOn={() => ensureAuth(data.isAuthenticated) && toggleReactionOn(data.id, data.reaction.termId)}
```
deadlybutter: I do like the simplification here though!

resources/assets/components/ReportbackBlock/index.js
weerd: Could we update the store to set it to `null` instead of `false`? Seems more appropriate for it to be `null` rather than a boolean.

deadlybutter: It is null already! I'm just suggesting literally putting `isAuthenticated: props.user.id !== null,` on the store so we don't have to rewrite this conditional everywhere

(also totally not necessary for this PR)

weerd: Sorry I meant make this `null` by default: https://github.com/DoSomething/phoenix-next/blob/749ebb37c15778c1b83d9b1fe78652866e21596b/resources/assets/store.js#L40

Forgot to add link!

deadlybutter: Agh, yep that should be null. The server always defaults to null so i didnt even consider the actual store file default haha

resources/assets/components/ReportbackPhoto/index.js
weerd: @deadlybutter Wait, aren't reactions specific to the RB item and not an RB that contains numerous items? They can each have a reaction...

deadlybutter: In that case this component should be `ReportbackItem` then?

resources/assets/components/ReportbackBlock/index.js
DFurnes: Agreed on `null` as the default! I didn't realize it wasn't. 😅 I'm planning on pulling out the things in this `data` array into a reducer in my next PR, so will tidy that up then.

resources/assets/components/ReportbackPhoto/index.js
DFurnes: Yeah I think the reaction is to a specific photo/item. I'm up for renaming to `ReportbackItem`.

DFurnes: Updated in b015153!

PR #94, opened 2017-03-23
Normalize data before passing into components.

DFurnes:
#### Changes
__This pull request updates how we pass data to the feed & reportback components from the store.__ Instead of grabbing the whole "raw" JSON blob and attaching it to the feed props, we decide which IDs to include in the `CampaignFeed` and then grab the data for each individual gallery item in the `ReportbackContainer`. This better matches Redux convention & allows us to minimize re-renders by only passing the data necessary to render individual parts of the interface.

I've also pulled the contents of `CampaignFeed` into the container component, since it's all concerned with processing data based on the application state (not rendering things to the user).

Sorry in advance for such a hefty & chaotic pull request. 😬


DFurnes:
@weerd @deadlybutter This is ready for review! 👓

deadlybutter:
@DFurnes Have to run for a 1:1 w/ Matt but will def. look when I get back!

resources/assets/actions/index.js
```diff
@@ -26,7 +26,8 @@ export const REACTION_COMPLETE = 'REACTION_COMPLETE';
 export * from './reaction';
 
 // Signup Action Names & Creators
-export const SIGNUP_COMPLETE = 'SIGNUP_COMPLETE';
+export const SIGNUP_CREATED = 'SIGNUP_CREATED';
+export const SIGNUP_EXISTS = 'SIGNUP_EXISTS';
 export const SIGNUP_PENDING = 'SIGNUP_PENDING';
-export const SET_CURRENTLY_SIGNED_UP = 'SET_CURRENTLY_SIGNED_UP';
+export const SIGNUP_FINISHED = 'SIGNUP_FINISHED';
```
weerd: Do we want to stick w/ a standard naming convention? Seems like we're doing `...pending`, `..successful`, `..failure`.

resources/assets/components/App.js
```diff
@@ -8,7 +8,7 @@ import createBrowserHistory from 'history/lib/createBrowserHistory';
 import { syncHistoryWithStore, routerReducer } from 'react-router-redux';
 
 import Chrome from './Chrome';
-import ActivityContainer from '../containers/ActivityContainer';
+import FeedContainer from '../containers/FeedContainer';
```
weerd: 👏

resources/assets/actions/index.js
DFurnes: That's a good idea, although I'm a little concerned that in this case both "finished" & "successful" implies something occurred, which may not be the case with the initial check on page load.

I'm thinking we should just handle changing the `isPending` status as  `SIGNUP_NOT_FOUND`, like we do in `SIGNUP_CREATED` and `SIGNUP_EXISTS`.

weerd: Yeah, that makes sense. Initially, what was the diff between `SIGNUP_CREATED` and `SIGNUP_FINISHED`?

DFurnes: Yeah yeah, that's the confusing part. Definitely makes sense to change the name.

`SIGNUP_CREATED` is when an actual record is made, `SIGNUP_EXISTS` is when an existing record is found (e.g. if you visit campaign on a new computer & were already signed up), and `SIGNUP_FINISHED` was just the catch-all (e.g. if you visit campaign on a new computer & weren't signed up). Probably `SIGNUP_NOT_FOUND` makes more sense.

resources/assets/components/ReportbackItem/index.js
```diff
@@ -1,25 +1,21 @@
...
+              onToggleOn={() => ensureAuth(isAuthenticated) && toggleReactionOn(id, reaction.termId)}
+              onToggleOff={() => toggleReactionOff(id, reaction.id)} />
+  );
 
   return (
-    <Figure className="reportback-item" image={image} imageClassName={isFetching ? 'is-fetching' : null}>
-      <BaseFigure media={reaction} alignment="right" className="padded">
-        <h4>{firstName}</h4>
-        <p className="footnote">{impact}</p>
+    <Figure className="reportback-item" image={isFetching ? EMPTY_IMAGE : url} imageClassName={isFetching ? 'is-fetching' : null}>
```
weerd: Nice. I might update the stuff in my PR to do it this way later, haha.

resources/assets/containers/ReportbackItemContainer.js
```diff
@@ -0,0 +1,37 @@
...
+
+  return {
+    isFetching: false,
+    id: reportbackItem.id,
+    url: reportbackItem.media.uri,
+    quantity: reportback.quantity,
+    firstName: reportback.user.first_name,
+    reaction: state.reactions.data[reportbackItem.id],
+    isAuthenticated: state.user.id !== null,
+  };
```
weerd: Nice. I might start doing more of this so it's less chaining to get at the props you need. I like it :)

resources/assets/store.js
```diff
@@ -37,7 +37,7 @@ export default function(reducers, preloadedState = {}) {
       offset: 1,
     },
     user: {
-      id: false,
+      id: null,
```
weerd: 😉

resources/assets/actions/index.js
weerd: cool, thanks for clarifying. Yeah, I'd vote for `SIGNUP_NOT_FOUND`. Seems clearer to me :)

deadlybutter: Hm, I'm not sure about this rename. I think `SET_CURRENTLY_SIGNED_UP` was called both async & sync, and `FINISHED/SUCCESSFUL` makes me think its the result of an async

deadlybutter: Oh I see, was this renamed to `SIGNUP_CREATED`

DFurnes: Yep, that's part of the change here – we now only dispatch actions when things change.

DFurnes: @weerd Renamed `SIGNUP_EXISTS` to `SIGNUP_FOUND` and `SIGNUP_FINISHED` to `SIGNUP_NOT_FOUND`... reads a lot better now. 👏

resources/assets/containers/FeedContainer.js
```diff
@@ -0,0 +1,103 @@
...
+import Feed from '../components/Feed';
+import {
+  clickedViewMore,
+  clickedSignUp,
+  checkForSignup,
+  setCurrentlySignedUp,
+} from '../actions';
+
+const BLOCKS_PER_ROW = 3;
+const ROWS_PER_PAGE = 3;
```
deadlybutter: I think I had this as a prop before in case we wanted affiliated/unaffiliated feeds to be different length (or A/B tested)

DFurnes: That makes sense. We could potentially move this into the state in that case.

PR #95, opened 2017-03-23
DATA!

deadlybutter:
**still need to finish some things & also add some requests from data team, no mergey**

----

This PR adds a few critical pieces of code that will let us do a ton of automated data tracking for phoenix-next.

- **all actions produce events**. this makes _time to data_ effectively zero for most of our tracking needs.
- **every event has the application state attached**. It goes through a transformer first to clean it up and remove data we dont need tracked.
- **unique session ids**. These ID's are effective pre- & post-auth, allowing us to "link" your events together. 
- **unique device id**. Suggestion from Jeff, could be useful to see how many times a device comes to site before signing up & what activity they do. They could link these id's together later-on as well if they wanted.
- **custom metadata**. My solution for now was to just attach all of the properties sent along with the action, eg:
- **connect to keen.io**. *duh*.

```json
"meta": {
   "action": "REACTION_CHANGED",
   "data": {
      "reportbackItemId": "172193",
      "value": false
   }
}
```


deadlybutter:
@DFurnes @weerd PR isn't ready for merge but feel free to take a look & let me what you think so far!

deadlybutter:
Currently looking at everything in Keen, but feel free to take a look @DFurnes @weerd -- should be good for a review!

resources/assets/analytics.js
```diff
@@ -0,0 +1,97 @@
...
+  history.listen(({ basename, pathname }) => {
+    let path = basename;
+    if (pathname) path += pathname;
+
+    pageView(path);
+  });
+  pageView(window.location.pathname);
+
+  // Track state changes for Keen.io
+  store.subscribe(() => stateChanged(store.getState()));
```
DFurnes: This feels like a perfect use case for [Redux middleware](http://redux.js.org/docs/advanced/Middleware.html), especially since that lets us include action names.

resources/assets/helpers.js
```diff
@@ -190,3 +190,67 @@ function stripExifData(image, dataView = null) {
...
   }
 }
+
+/**
+ * Generate a random unique id based on the current time
+ * and a random 5 digit number.
+ *
+ * @return {String}
+ */
+export function generateUniqueId() {
```
DFurnes: It looks like this might already be something the [Keen client handles for us](https://github.com/keen/keen-tracking.js/blob/627c9e198d2a99bab9adf42bfe43ee214a8d4d37/README.md#unique-id).

resources/assets/reducers/analytics.js
```diff
@@ -0,0 +1,20 @@
+import {
+  CLICKED_VIEW_MORE,
+} from '../actions';
+
+/**
+ * Analytics reducer:
+ * WARNING: Do NOT rely on this for any components. This is purley for data tracking.
```
DFurnes: This feels dangerous. Let's use a middleware so we don't need to store this in the state.

resources/assets/store.js
```diff
@@ -48,6 +48,10 @@ export default function(reducers, preloadedState = {}) {
       thisSession: false,
       pending: false,
     },
+    analytics: { // Don't use this in application logic!
```
DFurnes: ⚠️

resources/assets/analytics.js
deadlybutter: Never heard of this, will read up!

resources/assets/helpers.js
deadlybutter: Ah yeah we can use this. Originally I was using the timestamp in the ID to calculate lastUpdated, but later figured it was cleaner to just use two variables.

deadlybutter: Actually... I'm not sure we should -- If the Keen client isn't loaded the Analytics framework in this PR will still execute regardless (good for debugging like rn while im working on it).

resources/assets/analytics.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/**
+ * Redux middleware for tracking state changes.
+ *
+ * @param  {Object} store Application store
+ * @return {Object}
+ */
+export const observerMiddleware = store => next => action => {
+  const result = next(action);
```
DFurnes: Can this variable be inlined?

```js
export const observerMiddleware = store => next => action => {
    stateChanged(action, store.getState());

    return next(action);
};
```

resources/assets/analytics.js
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * @param  {Object} action Action that fired
+ * @param  {Object} state  Application state
+ */
+function stateChanged(action, state) {
+  updateSession();
+  const transformation = transformState(action, state);
+
+  analyze('action', transformation);
+  console.log(transformation);
```
DFurnes: Unrelated to this pull request, but we should handle logging this within the analytics package!

resources/assets/components/App.js
```diff
@@ -19,19 +19,8 @@ const store = configureStore({...reducers, routing: routerReducer}, window.STATE
...
-}
-
-pageView(window.location.pathname);
-history.listen(({ basename, pathname }) => {
-  let path = basename;
-  if (pathname) path += pathname;
-
-  pageView(path);
-});
+import observe from '../analytics';
```
DFurnes: Let's pull this up with the other imports.

resources/assets/analytics.js
deadlybutter: Ah, probably would have been a better to do it there!

resources/assets/analytics.js
deadlybutter: Unfortunately not (I think), the Redux examples imply `next` has to be called if you want the updated state

```js
  let result = next(action)
  console.log('next state', store.getState())
```

DFurnes: Ah, that makes sense.

weerd:
Haha, maybe update the description 😉 

> still need to finish some things & also add some requests from data team, no mergey

PR #96, opened 2017-03-23
Submission gallery preload user submissions

weerd:
This PR provides an update that allows the ReportbackUploader to request prior RB submissions from the user so that it may show them in the RB gallery. The gallery submissions are viewed from newest to oldest.

Future optimization: avoid re-rendering entire submissions gallery when a user uploads and adds a new Reportback.


resources/assets/actions/reportback.js
```diff
@@ -83,6 +98,31 @@ export function submitReportback(reportback) {
...
+
+  return dispatch => {
+    dispatch(requestingUserReportbacks());
+
+    return (new Phoenix).get('signups', { campaigns: campaignId, users: userId })
+      .then(json => {
+        dispatch(receivedUserReportbacks());
+
+        if (json.data.length) {
+          let reportbackItems = json.data.shift().reportback.reportback_items.data;
```
DFurnes: Ooh, this is a neat little trick. 🎩

resources/assets/actions/reportback.js
```diff
@@ -83,6 +98,31 @@ export function submitReportback(reportback) {
...
+
+          reportbackItems.map(reportbackItem => {
```
DFurnes: Since this isn't returning anything, it should be a `forEach`.

resources/assets/components/ReportbackSubmissions/index.js
```diff
@@ -6,10 +6,14 @@ class ReportbackSubmissions extends React.Component {
   }
 
   renderSubmission(reportback, index) {
+    // let uri = reportback.media.filePreviewUrl || reportback.media.uri;
+    // @TODO: should be the above line, but currently, images submitted (and thus returned) from Thor are broken.
+    let uri = reportback.media.filePreviewUrl || 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7';
```
DFurnes: I moved this into the "helpers" file in #94, so you can import the constant from there after merge! 😄

resources/assets/containers/ReportbackUploaderContainer.js
```diff
@@ -1,14 +1,18 @@
...
 
 /**
  * Provide state from the Redux store as props for this component.
  */
 const mapStateToProps = (state) => {
   return {
-    campaign: state.campaign,
-    submissions: state.submissions
+    campaign: {
+      id: state.campaign.id,
```
DFurnes: I don't think it's a bad idea to send each of these as a top-level `campaignId` and `legacyCampaignId` prop.

resources/assets/actions/reportback.js
weerd: 💫

resources/assets/actions/reportback.js
weerd: Oh ya, true.

resources/assets/components/ReportbackSubmissions/index.js
weerd: haha, I actually added a constant initially and then removed it cause I wasn't sure I'd keep it, but yeah, I like the idea of having it as a constant 👍

resources/assets/containers/ReportbackUploaderContainer.js
weerd: ok, sounds good.

resources/assets/reducers/submissions.js
```diff
@@ -9,16 +12,24 @@ import {
...
       return {...state, isStoring: true};
 
     case STORE_REPORTBACK_SUCCESSFUL:
       return {...state, isStoring: false};
 
     case ADD_TO_SUBMISSIONS_LIST:
-      return Object.assign({}, state, {
-        data: [...state.data, action.reportback]
-      })
+      return {...state, data: [...state.data, action.reportback].reverse()}
```
DFurnes: Hmm, will this reverse the submissions gallery each time we add a new item?

Should we do `[action.reportback, ...state.data]` instead?

weerd: I can play with it in the following PR when I do the rendering of the RB items. I wasn't sure since they come from API as oldest -> newest. I was reversing it earlier, but needed some place to do it both when it adds items via API and/or when adding uploads via the uploader.

DFurnes: Hmm, okay. Something to keep an eye on!

weerd: Re-tested and you're right. I'll defs fix in next PR ;)

PR #99, opened 2017-03-24
Collapse Redux logger for easier reading.

DFurnes:
## Changes
This collapses the Redux log statement so they aren't so noisy. 🔇

#### Before
screen shot 2017-03-24 at 11 35 16 am

#### After
screen shot 2017-03-24 at 11 35 36 am


deadlybutter:
wait how dow we even turn the redux logging on? 😅

DFurnes:
@deadlybutter It's on by default whenever you're making dev builds with `npm start`!

PR #100, opened 2017-03-24
Normalized application state, lazy-loaded reportbacks & "infinite" pagination.

DFurnes:
### Changes
__This pull request adds an MVP for "infinite" reportback pagination.__ It allows the feed to be rendered independently of reportbacks (displaying placeholder cards at first), and loads more reportbacks into the store whenever we need more. Work in progress. 🚧 

#### Normalized State
One of the big parts of this pull request is "normalizing" data before putting it into the Redux state, so we're not adding these big nested objects with duplicated data. Right now I'm doing this manually in `normalizers.js` but we can probably handle this with [normalizr](https://github.com/paularmstrong/normalizr) later.

Here's the gist:
```diff
reportbacks: {
  isFetching: false,
+  ids: ['1321412', '1321413', '1321414'],
+  entities: {
+    '1321412': { /* ... */, reportback_items: ['22215', '22216'] },
+    '1321413': { /* ... */ },
+    '1321414': { /* ... */ },
+  },
+  itemEntities: {
+    '22215': { /* ... */ },
+    '22216': { /* ... */ },
+ },
-  data: [
-    {
-      id: '1321412',
-      reportback_items: [
-        {id: '22215', /* ... */},
-        {id: '22216', /* ... */},
-      ],
-    },
-    { /* ... */ },
-    { /* ... */ },
-  ],
}
```

A nice side-effect of this is that we no longer need to loop over the entire collection when looking for particular records since they're keyed by their ID. I've also pulled `reactions` into the "reportback items" in the state tree, because we do not currently support multiple "types" of reactions in Rogue and so shouldn't need this as a separate entity for a long while.

#### Lazy-Loaded Reportbacks
We don't want Phoenix Next to be bottle-necked on slow Phoenix-Ashes performance, so we're no longer pre-loading reportbacks in the Laravel controller. Instead, the feed handles fetching the first page of reportbacks on initial mount, and displays empty "loading" blocks in the meantime.
 
#### Infinite Reportbacks
Right now, this is handled by creating "fake" reportback blocks on demand whenever the page offset exceeds the number of available blocks. I think this can be done more cleanly, but it gets the job done!

📸


DFurnes:
This is (another) real monster of a reportback, but ready for review! 🙃 

Things that aren't done yet:
 - We currently load another reportback page each time a user clicks "view more", which is way overkill. I'll tone this down in a follow-up pull request.
 - We don't yet hide the "view more" button once all reportbacks are shown.

deadlybutter:
> I've also pulled reactions into the "reportback items" in the state tree, because we do not currently support multiple "types" of reactions in Rogue and so shouldn't need this as a separate entity for a long while.

I moved this originally because reaching the `reactions` object tree was a YUGE pain in the ass but this seems to have resolved that 👏

resources/assets/actions/reaction.js
```diff
@@ -1,50 +0,0 @@
-import { Phoenix } from '@dosomething/gateway';
```
deadlybutter: Not sure I understand why I merged the reaction actions & reportback actions into one file?

resources/assets/containers/FeedContainer.js
```diff
@@ -32,15 +32,33 @@ const mapDisplayToPoints = (displayOption) => {
...
-    return totalRows < rowTarget;
+    return totalPoints < pointTarget;
   });
+
+  // If we weren't able to fill enough rows with blocks, add
+  // enough additional rows of reportbacks.
+  while (totalPoints < pointTarget) {
+    // @TODO: There's gotta be a better way of doing this.
+    filteredBlocks.push({
+      id: String(Math.ceil(Math.random() * 100000)),
```
deadlybutter: Im a little worried about using random id's here once we start using them for analytics purposes - can we potentially use a static string?

resources/assets/actions/reaction.js
DFurnes: Since I merged the reaction state into the "reportback item" entity, those state changes have to happen in the reportbacks reducer (since each reducer only modifies it's own slice of the state). Based on that it seemed to make sense to do the same with actions.

It feels like a reasonable separation of concerns would be to have one set of actions/reducer for reportback gallery concerns, and another for reportback submission concerns (which I left as a `@TODO` suggestion).

deadlybutter: Oh right right - that makes sense now

resources/assets/containers/FeedContainer.js
DFurnes: Agreed. We need a unique string because this is used as the `key` when rendering out the feed interface, but there's definitely a better solution here. A quick solution would be prefixing the "dynamic block" IDs, or maybe making individual blocks per reportback across the board (rather than adding rows and then filling them in)...

resources/assets/actions/index.js
```diff
@@ -10,24 +10,25 @@
...
+// Submissions
 export const STORE_REPORTBACK_PENDING = 'STORE_REPORTBACK_PENDING';
 export const STORE_REPORTBACK_FAILED = 'STORE_REPORTBACK_FAILED';
 export const STORE_REPORTBACK_SUCCESSFUL = 'STORE_REPORTBACK_SUCCESSFUL';
 export const ADD_TO_SUBMISSIONS_LIST = 'ADD_TO_SUBMISSIONS_LIST';
 export const REQUESTED_USER_SUBMISSIONS = 'REQUESTED_USER_SUBMISSIONS';
 export const REQUESTED_USER_SUBMISSIONS_FAILED = 'REQUESTED_USER_SUBMISSIONS_FAILED';
 export const RECEIVED_USER_SUBMISSIONS = 'RECEIVED_USER_SUBMISSIONS';
-export * from './reportback';
+// @TODO: Split submission actions & reducers into separate file.
```
weerd: Sounds good.

resources/assets/actions/reportback.js
```diff
@@ -64,6 +81,32 @@ export function addToSubmissionsList(reportback) {
...
+      });
+  }
+}
+
+// Async Action: user un-reacted to a photo.
+export function toggleReactionOff(reportbackItemId, reactionId) {
+  return dispatch => {
+    dispatch(reactionChanged(reportbackItemId, false));
+
+    (new Phoenix).delete(`reactions/${reactionId}`);
```
weerd: don't think I recall from before, but why doesn't this dispatch a `reactionComplete()`? Does Phoenix Ashes not return anything useful when removing a reactions?

resources/assets/components/ReportbackBlock/index.js
```diff
@@ -5,15 +5,20 @@ import ReportbackItemContainer from '../../containers/ReportbackItemContainer';
...
-  const items = props.reportbacks.map(id => {
-    return (
-      <FlexCell key={id}>
+  let items = [];
+
+  for (let i = 0; i < props.fields.additionalContent.count; i++) {
+    const id = props.reportbacks[i];
+
+    items.push(
+      <FlexCell key={id || `null-${i}`}>
```
weerd: 🤔

DFurnes: I know, I know! I'm going to try to make this cleaner in a follow-up PR.

resources/assets/actions/reportback.js
DFurnes: Yeah, we should definitely look into this to unset the old reaction ID from the state tree. Also we should think about adding a `REACTION_FAILED` action and state property.

resources/assets/normalizers.js
```diff
@@ -0,0 +1,31 @@
...
+    reportback.reportback_items = reportback.reportback_items.data.map((item) => {
+      const currentUser = item.kudos.data[0] ? item.kudos.data[0].current_user : false;
+
+      item.reaction = {
+        id: currentUser ? currentUser.kudos_id : null,
+        reacted: !!(currentUser && currentUser.kudos_id),
+        total: item.kudos.data[0] ? item.kudos.data[0].term.total : 0,
+        termId: item.kudos.data[0] ? item.kudos.data[0].term.id : '1274', // This is a hardcoded default because phoenix-ashes is bugged.
+      };
+      delete item.kudos;
```
weerd: how did I not know `delete` was a thing in JS? @_@

DFurnes: Hehe, it's a secret! 🙊

PR #102, opened 2017-03-28
Social tag overrides

deadlybutter:
This PR lets campaign admins override share fields, eg: this is the new Facebook share for Teens for Jeans

screen shot 2017-03-28 at 11 06 30 am

---

- Created a new content model called "Share Overrides" in contentful. I was originally using JSON but we needed to also support custom image, so figured it was easier to just make the content model.

- Added some new helper functions to get the share field data & chose between base fields and override fields. Note: I did _not_ use `isset` here, as Diego & myself discovered, Contentful is using magic methods and they dont play well with `isset`.

- Refactored the partial to use these new vars.

For testing, check both the override fields (Above ^) but also check a campaign with no override fields, eg: 

screen shot 2017-03-28 at 11 15 33 am


deadlybutter:
totally unrelated - how long have these heroku deployments been borked? 👀

DFurnes:
@deadlybutter Hmm, the Heroku deployment looks fine to me! What's wrong with it? 👀

deadlybutter:
Oh weird, i opened it so fast the post deploy steps didnt run yet

PR #104, opened 2017-03-28
Adding facebook app id to share properties

deadlybutter:
Don't need this for open graph technically but I think its useful for tracking, and I definitely need it for the share dialog

screen shot 2017-03-28 at 12 41 17 pm

(Created a Test App for our prod app in the FB dev tools, if you want the APP_ID for it LMK)


deadlybutter:
@weerd or @DFurnes very quick one!

PR #106, opened 2017-03-28
Adds share component

deadlybutter:
Sharing PR! 

This PR adds a few things

- Facebook SDK. This loads async & doesn't block the page from rendering. Seems better than trying to hack it a nonofficial way, also gives us potential to leverage the FB api down the road.
- Adds some "sharing" actions & reducers. This could have all fit in 1 component neatly, but I wanted everything to be tracked when people try to share.
- Temp. adds the share button in the update component. This will obviously move somewhere else in the final design (or stay, who knows!)

This pr does NOT

- Style the button. Next up!
- Add user id's to the share url (Suggestion by Matt). I can play around with that and add it in the next PR if it works!

### DIFF
--- a/resources/assets/components/CampaignUpdateBlock/index.js
+++ b/resources/assets/components/CampaignUpdateBlock/index.js
@@ -7,6 +7,9 @@ import Embed from '../Embed';
 import DEFAULT_AVATAR from './default-avatar.png';
 import './campaign-update.scss';
 
+// TEMP - purley for demo purposes till it has an official home.
+import Share from '../../containers/ShareContainer';
+
 const Byline = ({author, jobTitle = 'DoSomething.org Staff', avatar = DEFAULT_AVATAR}) => (
   <Figure size="small" alignment="left" verticalAlignment="center" image={avatar} imageClassName="avatar">
     <strong>{author}</strong><br/>
@@ -28,6 +31,7 @@ const CampaignUpdateBlock = (props) => {
       </Markdown>
       { link ? <Embed url={link} /> : null }
       { author ? <Byline author={author} jobTitle={jobTitle} avatar={avatar} />: null}
+      <Share />
     </Block>
   );
 };
--- a/resources/assets/components/CampaignUpdateBlock/index.js
+++ b/resources/assets/components/CampaignUpdateBlock/index.js
@@ -8,7 +8,7 @@ import DEFAULT_AVATAR from './default-avatar.png';
 import './campaign-update.scss';
 
 // TEMP - purley for demo purposes till it has an official home.
-import Share from '../../containers/ShareContainer';
+import ShareContainer from '../../containers/ShareContainer';
 
 const Byline = ({author, jobTitle = 'DoSomething.org Staff', avatar = DEFAULT_AVATAR}) => (
   <Figure size="small" alignment="left" verticalAlignment="center" image={avatar} imageClassName="avatar">
@@ -31,7 +31,7 @@ const CampaignUpdateBlock = (props) => {
       </Markdown>
       { link ? <Embed url={link} /> : null }
       { author ? <Byline author={author} jobTitle={jobTitle} avatar={avatar} />: null}
-      <Share />
+      <ShareContainer />
     </Block>
   );
 };
--- a/resources/assets/containers/ShareContainer.js
+++ b/resources/assets/containers/ShareContainer.js
@@ -1,0 +1,21 @@
+import { connect } from 'react-redux';
+import Share from '../components/Share';
+import { clickedShare } from '../actions';
+
+const mapStateToProps = (state) => {
+  return {
+    social: status.user,
+    user: state.user,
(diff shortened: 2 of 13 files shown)


deadlybutter:
[(5) As a campaign lead, I want to be able to drop the Facebook share button into the Affirmation component and maybe elsewhere](https://trello.com/c/Z4SOtMMn/396-5-as-a-campaign-lead-i-want-to-be-able-to-drop-the-facebook-share-button-into-the-affirmation-component-and-maybe-elsewhere)

resources/assets/containers/ShareContainer.js
```diff
@@ -0,0 +1,21 @@
+import { connect } from 'react-redux';
+import Share from '../components/Share';
+import { clickedShare } from '../actions';
+
+const mapStateToProps = (state) => {
+  return {
+    social: status.user,
```
DFurnes: What is `status` here? Should this be `state.status`?

resources/views/partials/social.blade.php
```diff
@@ -10,3 +10,23 @@
 <meta property="og:image" content="{{ $shareFields['coverImage'] }}" />
 
 <meta property="fb:app_id" content="{{ $shareFields['facebookAppId'] }}" />
+
+{{-- This is a non-blocking script --}}
+<script>
+  window.fbAsyncInit = function() {
+    FB.init({
```
DFurnes: I'm a little wary of including the full Facebook SDK on our page, since it's a big ol' 196kb and we don't really use much of it. Can we handle this with an [old fashioned popup](https://github.com/DoSomething/voting-app/blob/a1a794b0ba35673d699be151b195171237fb02ef/resources/assets/js/utilities/share-link.js#L15-L24) like we have in other applications?

resources/assets/components/CampaignUpdateBlock/index.js
```diff
@@ -7,6 +7,9 @@ import Embed from '../Embed';
 import DEFAULT_AVATAR from './default-avatar.png';
 import './campaign-update.scss';
 
+// TEMP - purley for demo purposes till it has an official home.
+import Share from '../../containers/ShareContainer';
```
DFurnes: I know this is temporary, but let's always name our imports based on the class, so `ShareContainer` here! Otherwise it could be easy to think we're importing the unwrapped `Share` presentational component.

resources/views/partials/social.blade.php
deadlybutter: We could but we'd lose out on grabbing the share metrics - I also think we might eventually want to leverage the FB API (These are your friends doing the campaign, Total shares, Login, etc). Coupled with the fact this doesn't block initial page load / render I'm not that worried about loading the full sdk

resources/assets/containers/ShareContainer.js
deadlybutter: Yep, bad copy/paste. The variable isn't actually used so it didn't throw any errors

resources/assets/components/CampaignUpdateBlock/index.js
deadlybutter: Oh of course!

resources/assets/containers/ShareContainer.js
deadlybutter: The whole line is outdated actually, it should be `share` instead of `social`

resources/views/partials/social.blade.php
DFurnes: Ehh, I still think this a pretty heavy implementation since the only one of those features that's currently on the roadmap is displaying a share modal. I'm not _that_ strongly opposed to merging, but I think we should be thoughtful when adding third-party scripts like this.

DFurnes: I do appreciate the value of knowing if a share successfully submitted though... I guess let's merge for now but maybe add a backlog task to consider more bandwidth-friendly options.

deadlybutter: We can also force the client to cache this file (even though the browser should be doing that automatically?)

weerd: Hmmm... In terms of caching, we could set it up so that webpack adds a vendor directory with third-party libraries that change less frequently so that we can cache that a bit better than our app code which is going to be changing more frequently.

Just a side thought... 🤔

DFurnes: Yeah, not arguing that it won't be cached (or that many visitors may not already have it cached from visiting other sites that use the Facebook SDK). It just seems like a little bit of a bummer that we're more than doubling our total JavaScript payload just to display a popup.

deadlybutter: it'd be cool if Facebook didn't ship their entire api in a single package, i can always look into pulling out the parts we don't need and just having our own (I assume its on github somewhere)

DFurnes: Yeah, that's fine by me.

PR #108, opened 2017-03-29
Submissions gallery rendering

weerd:
This PR updates the Reportback Submission Gallery to render out ReportbackItems collected from prior submissions via the `/signups` endpoint, as well as new submissions added via the ReportbackUploader.

image
Didn't add much styling. Figured that can come later, but for now using the basic Forge triad gallery. Will be converted to a flexbox based gallery when implementing the finalized design.

I decided to separate out the actions for adding RB submission metadata (like `whyParticipated` and the `quantity/impact`) from the actual list of submission items. I had them in same action before and worked great for the most part when retrieving data and updating the store from the `/signups` endpoint call, but ended up running into some trouble when submitting individual items via the uploader and then updating data via this alternative representation of a Reportback and ReportbackItem. It's something we could try and _normalize_ better in the future.

_NOTE:_ I had to add a slight override to the PhoenixLegacy `getAllSignups()` method so that if a query param for `users` and `campaigns` is passed then we should not cache it, otherwise users coming back to the gallery after a page refresh would not see their latest submissions for 30 minutes (cache expiration time). I considered instead checking for maybe a `clear` or `cache` query param that could be passed instead and if set to the proper boolean than not cache or flush the cache, but wasn't sure on the most ideal approach. It's something we can come back to, but wasn't vital to execute pristinely ✨ for this PR.

Refs https://trello.com/c/zkIhhT01/381-5-as-a-user-when-i-submit-a-rb-in-the-uploader-i-want-it-to-show-up-under-the-uploader-looking-like-a-rb-item-so-that-i-can-see-


resources/assets/components/Gallery/index.js
```diff
@@ -0,0 +1,24 @@
...
+import classnames from 'classnames';
+import { modifiers } from '../../helpers';
+
+const renderGalleryItem = (child, index) => {
+  return <li key={`submission-${index}`} className="">{child}</li>;
+};
+
+const Gallery = ({isFetching, type = null, children}) => {
+  if (isFetching) {
+    return <div>loading…</div>;
```
deadlybutter: super minor and not a blocker, but we have a spinner thing! 

https://github.com/DoSomething/phoenix-next/blob/b35836bb86c860bed95ae75ee297881e1a75ca1d/resources/assets/components/Embed/index.js#L21

weerd: Yep yep. Figured I'd update that stuff when I went back in to do the styling more thoroughly. :)

resources/assets/helpers.js
```diff
@@ -83,7 +83,11 @@ export function markdown(source) {
 export function modifiers(...classes) {
   if (! Array.isArray(classes)) classes = [classes];
 
-  return classes.map(className => `-${className}`);
+  return classes.map(className => {
```
DFurnes: This is kinda subjective, but I'd love to avoid the conditional here. Plus no undefineds in the result. 🤓

```js
return classes.filter(className => className).map(className => `-${className}`);
```

resources/assets/reducers/submissions.js
```diff
@@ -28,8 +29,19 @@ const submissions = (state = {}, action) => {
     case STORE_REPORTBACK_SUCCESSFUL:
       return {...state, isStoring: false};
 
-    case ADD_TO_SUBMISSIONS_LIST:
-      return {...state, data: [action.reportback, ...state.data]}
+    case ADD_SUBMISSION_METADATA:
+      return {
+        ...state,
+        reportback: {
+          id: action.id || action.reportback.id,
```
DFurnes: Not merge blocking, but I think it makes sense to normalize this in the action (so it always explicitly asks for a particular structure), rather than in the reducer.

app/Services/PhoenixLegacy.php
```diff
@@ -37,6 +37,12 @@ public function getAllSignups(array $query = [])
     {
         $path = 'v1/signups';
 
+        // Avoid caching when requesting signups for sepcific user and campaign.
```
weerd: Sounds good. Will remove in upcoming PR!

resources/assets/components/Gallery/index.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+  else {
```
weerd: ah true.

resources/assets/components/ReportbackUploader/index.js
```diff
@@ -107,8 +108,9 @@ class ReportbackUploader extends React.Component {
             <button className="button" type="submit">Submit a new photo</button>
           </form>
         </div>
-
-        <ReportbackSubmissions submissions={this.props.submissions}/>
+          <Gallery isFetching={this.props.submissions.isFetching} type="triad">
+            {this.props.submissions.items.map((submission, index) => <ReportbackItem key={Date.now() + index} {...submission} url={submission.media.uri || submission.media.filePreviewUrl} reaction={null} />)}
```
weerd: Yeah, I wasn't super keen on this, but part of the issue was that it's the ID on the ReportbackItem and not on the "parent" reportback... which means on a submission via the ReportbackUploader, we don't have a RBItem ID just yet, so I kinda needed to make something up. I had an `||` statement w/ using the `submission.id || something-else` before...

resources/assets/helpers.js
weerd: Yeah, that works. I can update in future PR or feel free to do it if you get to it before me :)

resources/assets/reducers/submissions.js
weerd: Yeah, I wasn't sure on this where the ideal place to normalize it would be... since a lot of the actions are so nice and clean, haha. I'll switch it up in upcoming update.

PR #111, opened 2017-03-30
User share

deadlybutter:
screen shot 2017-03-30 at 3 33 56 pm

Quotes!....

This PR,

- adds the NS user id to the share URL if available, suggestion by @mshmsh5000 
- adds "quotes" as another thing campaign leads can specify in the FB post, looks like this:

screen shot 2017-03-30 at 3 18 23 pm


deadlybutter:
[(5) As a campaign lead, I want to be able to drop the Facebook share button into the Affirmation component and maybe elsewhere](https://trello.com/c/Z4SOtMMn/396-5-as-a-campaign-lead-i-want-to-be-able-to-drop-the-facebook-share-button-into-the-affirmation-component-and-maybe-elsewhere)

resources/assets/actions/share.js
```diff
@@ -21,12 +21,19 @@ export function facebookShareCancelled() {
 
 // Action: user clicked a share button.
 export function clickedShare() {
-  return dispatch => {
+  return (dispatch, getState) => {
     dispatch(requestedFacebookShare());
 
+    const user = getState().user;
+    const id = user.id ? `?ns=${user.id}` : '';
+    const href = `${window.location.href}${id}`;
```
DFurnes: This will break if we have any query strings on the page URL when someone clicks to share (like UTM params from a marketing email, or a user who came from someone else's share link).

deadlybutter: ...Would we want to keep that across shares though?

DFurnes: Imagine if I click your share link and then share the page myself, my share URL would be:

[`https://next.dosomething.org/campaigns/teens-for-jeans/?ns=5564c93d469c6415048b8670?ns=5543dfd6469c64ec7d8b46b3`](https://next.dosomething.org/campaigns/teens-for-jeans/?ns=5564c93d469c6415048b8670?ns=5543dfd6469c64ec7d8b46b3)

...that's probably not what we want.

deadlybutter: Good point! Ill go ahead & fix that

deadlybutter: After further debugging, it seems like appending query params breaks the Facebook crawler (Facebook was probably using a cached page yesterday which is why it still worked...?)

deadlybutter: Regardless, not a launcher blocker, was just a fun idea we wanted to explore. Can revisit at a later time, pushing up a version that should be good for merge

PR #112, opened 2017-03-31
User based local storage mechanism, signups refactoring

deadlybutter:
Woot woot, this PR does a few things

- Adds *Storage Helpers*, this is for managing all user-level data we want to store in local storage.
- Refactors the signup actions+redux to 1. use this new mechanism 2. use the sign ups index endpoint
- adds localStorage load function to the store so we have the proper default state on startup

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -13,12 +13,28 @@ import {
 
 // Action: a new signup was created for a campaign.
 export function signupCreated(campaignId) {
-  return { type: SIGNUP_CREATED, campaignId };
+  return (dispatch, getState) => {
+    const { user } = getState();
+
+    dispatch({
+      type: SIGNUP_CREATED,
+      campaignId,
+      userId: user.id,
+    });
+  }
 }
 
 // Action: an existing signup was found for a campaign.
 export function signupFound(campaignId) {
-  return { type: SIGNUP_FOUND, campaignId };
+  return (dispatch, getState) => {
+    const { user } = getState();
+
+    dispatch({
+      type: SIGNUP_FOUND,
+      campaignId,
+      userId: user.id,
+    });
+  }
 }
 
 // Action: no existing signup was found for the campaign.
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -54,12 +54,14 @@ export function checkForSignup(campaignId) {
       campaigns: campaignId,
       user: getState().user.id,
     }).then(response => {
-      console.log(response);
       if (!response || !response.data || !response.data[0]) {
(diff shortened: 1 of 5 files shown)


resources/assets/actions/signup.js
```diff
@@ -33,11 +49,13 @@ export function signupPending() {
...
-    dispatch(signupPending());
-
-    (new Phoenix).get(`activity/${campaignId}`).then(response => {
-      if (!response || !response.sid) {
+  return (dispatch, getState) => {
+    (new Phoenix).get('signups', {
+      campaigns: campaignId,
+      user: getState().user.id,
+    }).then(response => {
+      console.log(response);
```
DFurnes: 🌴

resources/assets/actions/signup.js
```diff
@@ -33,11 +49,13 @@ export function signupPending() {
...
+      if (!response || !response.data || !response.data[0]) {
         return dispatch(signupNotFound());
```
DFurnes: Not necessary for merge, but let's `throw` in cases like this and then dispatch the failure action in the `catch` handler, since that way we'll get failure handling on network errors as well.

resources/assets/store.js
```diff
@@ -63,9 +64,11 @@ export default function(reducers, preloadedState = {}) {
   // If React DevTools are available, use instrumented compose function.
   const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
 
+  const transformedState = loadStorage(initialState, preloadedState);
```
DFurnes: This is nitpicky, but it's a little weird that we use `initialState` and `preloadedState` but only return a mutated version of `initialState`... and then merge that with `preloadedState` separately below.

I feel like it'd make a bit more sense for this to return an object with _just_ the part of the state that was in local storage, and then merge that with the rest below.

```js
const localState = loadStorage(initialState, preloadedState);

return createStore(
  combineReducers(reducers),
  merge(initialState, preloadedState, localState),
  composeEnhancers(applyMiddleware(...middleware))
);
```

Nothing that needs to be done for merge, but worth a little thought later on.

resources/assets/actions/signup.js
deadlybutter: Oh sure thing, i can swap that out

resources/assets/store.js
deadlybutter: Oh I just assumed `merge` accepted two parameters, I agree it certainly makes more sense to do all 3 like this.

deadlybutter: Ill add a TODO for later!

deadlybutter:
Cleaned up & retested for that change, thanks for hopping on this @DFurnes !

PR #113, opened 2017-03-31
JS Event queue

deadlybutter:
This PR adds an event queue for actions that need to run post authentication. For example, being signed up for a campaign. This is important because we don't want users clicking sign up, making an account, and being asked to sign up again.

### DIFF
--- a/resources/assets/actions/event.js
+++ b/resources/assets/actions/event.js
@@ -15,7 +15,7 @@ import {
 
 // Action: remove completed event from storage.
 export function completedEvent(index) {
-  return { type: COMPLETED_EVENT, index };
+  return { type: COMPLETED_EVENT, index, deviceId: getDeviceId() };
 }
 
 // Action: run through all of the events in the queue.
@@ -24,15 +24,11 @@ export function startQueue() {
     const queue = storageGet(getDeviceId(), EVENT_STORAGE_KEY);
 
     queue.forEach((event, index) => {
-      //TODO: Check if action is valid
-      dispatch(event.action);
+      // Check if the event is over 30 min old.
+      if (isTimestampValid(event.createdAt, (30 * 60 * 1000))) dispatch(event.action);
+
       dispatch(completedEvent(index));
     });
-
-    // Get events
-    // If still valid
-    // Dispatch action
-    // Dispatch completed event
   }
 }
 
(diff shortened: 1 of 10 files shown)


deadlybutter:
[(5) As a user, when I Affiliate with the campaign and get sent to Northstar and then sent back to the campaign page, I want to be signed into the campaign and not have to Affiliate again.](https://trello.com/c/8yWVvlV8/418-5-as-a-user-when-i-affiliate-with-the-campaign-and-get-sent-to-northstar-and-then-sent-back-to-the-campaign-page-i-want-to-be-si)

deadlybutter:
Ok steps left here

- Change `ensureAuth` to be an action that fires. Replace `isAuthenticated` with the action being fired.
- Change `queueEvent` to accept action type as param 1, and all `...arguments` afterwards are treated as function params to use later.
- Figure out how we're going to actually dispatch actions. I think we need to map these specific actions back to functions somewhere, and `.apply` the params or something when we call the function.
- Implement this in the Feed/Revealer stuff, CTA, Reactions... anywhere else??
- Remove test code from feed

deadlybutter:
alrighty @DFurnes @weerd this monster is ready for a look. I can do the reactions part in another PR, didn't want to add another bajillion files to this one 😬

resources/assets/actions/event.js
```diff
@@ -0,0 +1,60 @@
...
+
+    queue.forEach((event, index) => {
+      // Check if the event is over 30 min old before dispatching.
+      if (isTimestampValid(event.createdAt, (30 * 60 * 1000))) {
+        // Match the action creator from the saved name, load parameters to apply.
+        const action = allActions[event.action.creatorName];
+        const args = event.action.args || [];
+
+        // If the creator was found, dispatch the action.
+        if (action) dispatch(action.call(action.prototype, ...args));
```
DFurnes: This is tiny, but you should be able to simplify this to `dispatch(action(...args));`.

resources/assets/actions/event.js
```diff
@@ -0,0 +1,60 @@
...
+      // Always remove the event from storge.
+      dispatch(completedEvent(index));
+    });
+  }
+}
+
+// Action: add an event to the queue.
+export function queueEvent(actionCreatorName) {
+  const args = [...arguments];
+  args.shift();
```
DFurnes: This could be handled with a spread parameter in the function:

```js
export function queueEvent(actionCreatorName, ...args) {
  // ...
}
```

PR #114, opened 2017-03-31
More efficient pagination & hide "view more" button when at end of feed.

DFurnes:
#### Changes
__This pull request wraps up the "infinite" reportback gallery (for now).__ It will now only makes requests to load more reportbacks when there are empty "slots" in the feed, and will hide the "view more" button when the user reaches the end of the feed.

The calculations needed to render the feed happen in a new "selectors" directory. Right now this is just symbolic, but we can probably use something like [reselect](https://github.com/reactjs/reselect) to make this extra zippy in the future.

I've added an `initializeStore` export from the store which gets fired on a `onEnter` transition hook to the router. This gives us a convenient place that to dispatch actions that need to happen on application load (since we can't ensure that a user will always start on the community feed).

### DIFF
--- a/resources/assets/components/Feed/index.js
+++ b/resources/assets/components/Feed/index.js
@@ -39,11 +39,12 @@ const renderFeedItem = (block, index) => {
  *
  * @returns {XML}
  */
-const Feed = ({ blocks, callToAction, signedUp, hasNewSignup, hasPendingSignup, isAuthenticated, clickedViewMore, clickedSignUp }) => {
+const Feed = ({ blocks, callToAction, signedUp, hasNewSignup, hasPendingSignup, isAuthenticated, canLoadMorePages, clickedViewMore, clickedSignUp }) => {
   const viewMoreOrSignup = signedUp ? clickedViewMore : clickedSignUp;
   const revealer = <Revealer title={signedUp ? 'view more' : 'sign up'}
                              callToAction={signedUp ? '' : callToAction}
                              isLoading={hasPendingSignup}
+                             isVisible={(isAuthenticated && !signedUp) || canLoadMorePages}
                              onReveal={() => ensureAuth(isAuthenticated) && viewMoreOrSignup()} />;
 
   return (
--- a/resources/assets/components/Feed/index.js
+++ b/resources/assets/components/Feed/index.js
@@ -12,64 +12,51 @@ import Revealer from '../Revealer';
 import StaticBlock from '../StaticBlock';
 import { Flex, FlexCell } from '../Flex';
 
-class Feed extends React.Component {
-  /**
-   * Perform actions immediately after mounting component.
-   */
-  componentDidMount() {
-    // If we don't already have a signup, check.
-    if (! this.props.signedUp) {
-      this.props.checkForSignup(this.props.legacyCampaignId);
-    }
+/**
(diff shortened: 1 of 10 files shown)


resources/assets/components/Feed/index.js
```diff
@@ -12,64 +12,52 @@ import Revealer from '../Revealer';
...
+ * Render the feed.
+ *
+ * @returns {XML}
+ */
+const Feed = ({ blocks, callToAction, signedUp, hasNewSignup, hasPendingSignup, isAuthenticated, canLoadMorePages, clickedViewMore, clickedSignUp }) => {
+  const viewMoreOrSignup = signedUp ? clickedViewMore : clickedSignUp;
+  const revealer = <Revealer title={signedUp ? 'view more' : 'sign up'}
+                             callToAction={signedUp ? '' : callToAction}
+                             isLoading={hasPendingSignup}
+                             isVisible={(isAuthenticated && !signedUp) || canLoadMorePages}
```
deadlybutter: I'm assuming `canLoadMorePages` is true even for unauthenticated users

resources/assets/selectors/feed.js
```diff
@@ -0,0 +1,116 @@
...
+ * @returns {number}
+ */
+export const getBlockOffset = (state) => state.blocks.offset * BLOCKS_PER_ROW * ROWS_PER_PAGE;
+
+/**
+ * Get the number of blocks that are visible in the feed.
+ * @param state
+ * @returns {number}
+ */
+export const getMaximumOffset = (state) => totalBlocksInFeed(state) + (state.reportbacks.total - totalReportbackBlocksInFeed(state));
```
deadlybutter: Wait, are reportback blocks not counted in `totalBlocksInFeed`?

resources/assets/store.js
```diff
@@ -64,4 +66,28 @@ export default function(reducers, preloadedState = {}) {
...
-};
+}
+
+/**
+ * Dispatch any actions to lazy-load application state. This
+ * is done exactly once on initial page load.
+ *
+ * @param {Store<S>} store
+ */
+export function initializeStore(store) {
```
deadlybutter: Did we remove the `loadStorage` function? `https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/store.js#L68`

weerd:
Is there no PR branch deploy to check?

resources/assets/components/Feed/index.js
DFurnes: Yep, should be!

resources/assets/selectors/feed.js
DFurnes: For this, we're looking to get the total block offset – so the total number of blocks defined on the campaign in Contentful + the total number of "overflow" reportbacks. Reportback blocks subtract from that reportback total, since they display some of the reportbacks that would otherwise be appended to the end of the feed.

resources/assets/store.js
DFurnes: Nope, I haven't rebased this after you merged your PR. It'd be within the `configureStore` function above.

resources/assets/selectors/feed.js
deadlybutter: Oh I see, that makes sense

PR #115, opened 2017-04-03
Component tracking

deadlybutter:
This PR does some refactoring around the analytics implementation to make the files organized a bit cleaner and also allow for more in-depth tracking for specific actions we really care about. As an example, I did the reactions toggle functions.


deadlybutter:
[(5) As a developer, I want to add data tracking to my components](https://trello.com/c/VAFhXxWE/441-5-as-a-developer-i-want-to-add-data-tracking-to-my-components)

resources/assets/components/ReportbackItem/index.js
```diff
@@ -30,8 +31,8 @@ const ReportbackItem = ({
 
   const reactionElement = reaction && !isFetching ? (
     <Reaction active={reaction.reacted} total={reaction.total}
-              onToggleOn={() => ensureAuth(isAuthenticated) && toggleReactionOn(id, reaction.termId)}
-              onToggleOff={() => toggleReactionOff(id, reaction.id)} />
+              onToggleOn={() => ensureAuth(isAuthenticated) && toggleReactionOn(id, reaction.termId, props)}
```
DFurnes: We can change our mind later, but I think it makes sense to pass an explicit identifier here, like `item-${id}`. Otherwise a change in element props could unexpectedly impact our analytics data.

resources/assets/actions/reportback.js
```diff
@@ -89,8 +90,9 @@ export function addSubmissionItemToList(reportbackItem) {
 }
 
 // Async Action: user reacted to a photo.
-export function toggleReactionOn(reportbackItemId, termId) {
+export function toggleReactionOn(reportbackItemId, termId, component) {
   return dispatch => {
+    dispatch(trackEvent('reaction toggled on', component));
```
DFurnes: I'd still really love to try filtering analytics in the middleware, but that can be a later experiment I guess.

deadlybutter: I can add a quick array.indexOf check, should only take a second

resources/assets/components/ReportbackItem/index.js
deadlybutter: hm, not sure im following.

DFurnes: Oy, sorry not sure how that didn't send when I wrote earlier. Fine to merge for now.

PR #116, opened 2017-04-03
Adds authentication+queue to reactions

deadlybutter:
This PR requires users to authenticate if they try to react, and will post the reaction upon returning to the page. Followup of this PR: https://github.com/DoSomething/phoenix-next/pull/113


deadlybutter:
[(5) As a user, when I Affiliate with the campaign and get sent to Northstar and then sent back to the campaign page, I want to be signed into the campaign and not have to Affiliate again.](https://trello.com/c/8yWVvlV8/418-5-as-a-user-when-i-affiliate-with-the-campaign-and-get-sent-to-northstar-and-then-sent-back-to-the-campaign-page-i-want-to-be-si)

PR #117, opened 2017-04-03
Add the action page.

DFurnes:
#### Changes
__This pull request adds the action page!__ This involved extracting the Affirmation component to the Chrome so it'd stay between pages, adding the new action page & container (with functioning revealer & some fake placeholder content), and moving the reportback uploader to the new page.

Check it out [on the review branch](https://phoenix-next-pr-117.herokuapp.com/campaigns/teens-for-jeans).

🍹

### DIFF
--- a/resources/assets/components/ActionPage/index.js
+++ b/resources/assets/components/ActionPage/index.js
@@ -1,18 +1,9 @@
-/**
- * Render a single feed item.
- *
- * @param block
- * @param index
- * @returns {XML}
- */
 import React from 'react';
-import { ensureAuth } from '../../helpers';
-
-import Revealer from '../Revealer';
-import { Flex, FlexCell } from '../Flex';
-import ReportbackUploaderContainer from '../../containers/ReportbackUploaderContainer';
 import Block from '../Block';
 import Markdown from '../Markdown';
+import ReportbackUploaderContainer from '../../containers/ReportbackUploaderContainer';
+import Revealer from '../Revealer';
+import { Flex, FlexCell } from '../Flex';
 
 /**
  * Render the feed.
--- a/resources/assets/components/ActionPage/index.js
+++ b/resources/assets/components/ActionPage/index.js
@@ -1,0 +1,45 @@
+/**
+ * Render a single feed item.
+ *
+ * @param block
+ * @param index
+ * @returns {XML}
+ */
+import React from 'react';
+import { ensureAuth } from '../../helpers';
+
+import Revealer from '../Revealer';
+import { Flex, FlexCell } from '../Flex';
+import Block from '../Block';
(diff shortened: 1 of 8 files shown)


resources/assets/containers/NavigationContainer.js
```diff
@@ -16,7 +16,8 @@ const NavigationContainer = (props) => {
 
   return (
     <Navigation>
-      <NavigationLink to="/">Feed</NavigationLink>
+      <NavigationLink to="/">Community</NavigationLink>
```
deadlybutter: Not a huge deal, just wondering if we're worried about not using a const to represent the pathnames for these routes since they're being used in two places

resources/assets/components/ActionPage/index.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * Render a single feed item.
+ *
+ * @param block
+ * @param index
+ * @returns {XML}
+ */
+import React from 'react';
+import { ensureAuth } from '../../helpers';
```
DFurnes: Oh, huh. Didn't realize that was removed. 😕 Regardless, not used in this file so I'll remove.

resources/assets/components/App.js
```diff
@@ -26,8 +27,9 @@ observe(history, store);
 const App = (props) => (
   <Provider store={store}>
     <Router history={history}>
-      <Route component={Chrome} onEnter={initializeStore(store)}>
+      <Route component={ChromeContainer} onEnter={initializeStore(store)}>
         <Route path="/" component={FeedContainer}/>
+        <Route path="/action" component={ActionPageContainer}/>
```
DFurnes: The prefix is for differentiating "dynamic" pages set on the campaign in Contentful.

resources/assets/containers/NavigationContainer.js
DFurnes: Yeah, it could be interesting to pull those out. Worth keeping an eye on.

resources/assets/components/ActionPage/index.js
deadlybutter: We could re-add it if we run into a scenario where we want people to login but it has nothing to do with completing an action on the page, but right now all of the use cases for it got replaced by the event queue stuffs

ngjo:
[(3) As a campaign lead, I want Unauthenticated/Authenticated, Unaffiliated users to see an Action tab that Fades and prompts to Affiliate so that they must Affiliate to see the rest of the Action page.](https://trello.com/c/bz2e1Ojd/388-3-as-a-campaign-lead-i-want-unauthenticated-authenticated-unaffiliated-users-to-see-an-action-tab-that-fades-and-prompts-to-affi)

PR #119, opened 2017-04-04
Provide local URLs to uploaded files to send in API POST requests.

weerd:
This PR switches up posting Reportbacks to use the `file_url` to provide a URI for a file instead of sending the file as a DataURI. 

The issue was when converting the images to a DataURI to send in the API request to Phoenix-Ashes, the DataURI would be created properly, but JSON values have a string character limit and if the image was not "ant-sized" then the string would get cut off. The resulting image DataURI Phoenix-Ashes would get would thus be broken.

The current solution is to store the uploaded file in a `/uploads/images` directory on Phoenix-Next that is publicly accessible. In the API request we then send the URL to this file for Phoenix-Ashes to upload it itself and store it on its end. Once a response is received, then the locally stored file is deleted.

This PR also removed the caching of signups, so we can receive and up-to-date representation of user submissions.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -6,6 +6,7 @@ use Exception;
 use Illuminate\Auth\AuthenticationException;
 use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
 use Illuminate\Validation\ValidationException;
+use App\Exceptions\InvalidFileUploadException;
 
 class Handler extends ExceptionHandler
 {
@@ -86,6 +87,9 @@ class Handler extends ExceptionHandler
         }
         elseif ($exception instanceof AuthenticationException) {
             $code = 401;
+        }
+        elseif ($exception instanceof InvalidFileUploadException) {
+            $code = $exception->getStatusCode();
         }
         else {
             $code = 500;
--- a/app/Http/Controllers/ReportbackController.php
+++ b/app/Http/Controllers/ReportbackController.php
@@ -5,7 +5,6 @@ namespace App\Http\Controllers;
 use App\Exceptions\InvalidFileUploadException;
 use App\Services\PhoenixLegacy;
 use Illuminate\Http\Request;
-use Illuminate\Support\Facades\File;
 
 class ReportbackController extends Controller
 {
--- a/app/Http/Controllers/ReportbackController.php
+++ b/app/Http/Controllers/ReportbackController.php
@@ -2,8 +2,10 @@
 
 namespace App\Http\Controllers;
 
+use App\Exceptions\InvalidFileUploadException;
 use App\Services\PhoenixLegacy;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\File;
 
 class ReportbackController extends Controller
 {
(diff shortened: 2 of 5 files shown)


app/Http/Controllers/ReportbackController.php
```diff
@@ -47,18 +48,29 @@ public function store(Request $request)
...
         $reportbackPhoto = $request->file('media');
 
-        return $this->phoenixLegacy->storeReportback(
-            auth()->user()->legacy_id,
+        if (!$reportbackPhoto->isValid()) {
+            throw new InvalidFileUploadException;
+        }
+
+        // Store the uploaded file.
+        $path = '/uploads/'.$reportbackPhoto->store('images', 'uploads');
```
deadlybutter: I assume at some point we will introduce a script to remove these uploads?

app/Http/Controllers/ReportbackController.php
```diff
@@ -47,18 +48,29 @@ public function store(Request $request)
...
+
+        $response = $this->phoenixLegacy->storeReportback(
+            auth()->id(),
             $request->input('campaignId'),
             [
-                'file' => make_data_uri($reportbackPhoto->getPathname(), $reportbackPhoto->getMimeType()),
-                'filename' => time().'_reportback_photo.'.$reportbackPhoto->guessClientExtension(),
+                'file_url' => env('APP_ENV') !== 'local' ?  asset($path) : 'https://placeimg.com/1000/768/animals',
```
deadlybutter: Hm, I think it's probably better we remove this before merge unless it's still needed?

app/Http/Controllers/ReportbackController.php
deadlybutter: Oh I see we're doing that below, nevermind!

app/Http/Controllers/ReportbackController.php
weerd: Yeah, I'm leaving it in on purpose otherwise submitting locally leads to issues/broken images. You can't provide Thor with an image URL to a file on your local computer.

deadlybutter: gotcha!

app/Exceptions/Handler.php
```diff
@@ -87,6 +88,9 @@ protected function buildJsonResponse($request, Exception $exception) {
         elseif ($exception instanceof AuthenticationException) {
             $code = 401;
         }
+        elseif ($exception instanceof InvalidFileUploadException) {
```
DFurnes: Not a big deal, but this should be handled by the `$exception instanceof HttpException` check above, since the `InvalidFileUploadException` extends that.

weerd: I tried and it didn't seem to handle it. Would switch to `500`, but from the end of the ifelse conditional.

app/Http/Controllers/ReportbackController.php
DFurnes: We should use `config('app.env')` here since environment variables [get cached in production](https://laravel.com/docs/5.4/configuration#configuration-caching).

weerd: Cool. I can update that.

DFurnes:
Post-merge 👍!

app/Exceptions/Handler.php
DFurnes: Hmm, weird. Something to look into later on.

weerd: @DFurnes we might need to use [`is_a`](http://php.net/manual/en/function.is-a.php) instead of `instanceof`...

PR #121, opened 2017-04-05
Changing how we format data for custom events

deadlybutter:
This PR tries to make the data we send in custom events more standard so our queries (hopefully) won't break if we change component hierarchy, property names, etc.

I updated the work from yesterday (reactions) and also did sign ups to show the different ways it could be used.

### DIFF
--- a/resources/assets/actions/reportback.js
+++ b/resources/assets/actions/reportback.js
@@ -91,7 +91,7 @@ export function addSubmissionItemToList(reportbackItem) {
 }
 
 // Async Action: user reacted to a photo.
-export function toggleReactionOn(reportbackItemId, termId, component) {
+export function toggleReactionOn(reportbackItemId, termId, metadata) {
   return (dispatch, getState) => {
     // If the user is not logged in, handle this action later.
     if (! getState().user.id) {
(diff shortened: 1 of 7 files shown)


deadlybutter:
Here is what the CTA block looks like

screen shot 2017-04-05 at 11 09 57 am

deadlybutter:
So we could have a dashboard now that compares the signup rates from different CTA block types!

I'm also considering if we should add the Contentful block ID into the metadata as well, so we can compare specific ones. Jeff asked about this and I don't think it would be too much of a change, could probably just inject it [here](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/Feed/index.js#L31).

resources/assets/actions/reportback.js
```diff
@@ -91,37 +91,40 @@ export function addSubmissionItemToList(reportbackItem) {
...
+export function toggleReactionOff(reportbackItemId, reactionId, metadata) {
   return dispatch => {
-    dispatch(trackEvent('reaction toggled off', component));
     dispatch(reactionChanged(reportbackItemId, false));
 
     (new Phoenix).delete(`reactions/${reactionId}`)
-      .then(json => dispatch(reactionComplete(reportbackItemId, null)))
+      .then(json => {
+        dispatch(reactionComplete(reportbackItemId, null));
+        dispatch(trackEvent('reaction toggled off', metadata));
```
DFurnes: It still feels awkward to me that we're dispatching an action that doesn't impact the store. Definitely not merge-blocking (let's get this thing launched!!) but something to revisit later on.

deadlybutter: Good point, could probably just be a pure function call. I'll make a card in the "TODO POST LUNCH" thing

PR #122, opened 2017-04-05
Removing extra affirmation

deadlybutter:
When we moved the affirmation from the Feed to the Chrome we accidentally left it in the Feed (see [here](https://dosomething.slack.com/archives/C3ASB4204/p1491324454860194) for commits)

screen shot 2017-04-05 at 11 16 56 am

Easy fix!


deadlybutter:
retested after all that merge fixin, display seems okay, data in console seems good

screen shot 2017-04-05 at 11 25 21 am

PR #127, opened 2017-04-06
Action after signup

deadlybutter:
This PR does a few things!

1) After you signup, you're taken to the action page. To do this I had to make some changes to how the `History` object is used + exposed in our app.

2) I moved the `Affirmation` component to its correct position on the page and gave it a container so we could automatically track when people close it, and the state of it in general.

3) I needed to re-use the `.wrapper` class that's inside the `.feed-enclosure`. I abstracted the class out so multiple components could use it, if anyone has any thoughts / suggestions please throw em out!

### DIFF
--- a/resources/assets/components/Affirmation/affirmation.scss
+++ b/resources/assets/components/Affirmation/affirmation.scss
@@ -70,6 +70,7 @@
     line-height: 18px;
 
     &:hover {
+      cursor: pointer;
       text-decoration: none;
     }
   }
--- a/resources/assets/components/Affirmation/affirmation.scss
+++ b/resources/assets/components/Affirmation/affirmation.scss
@@ -7,10 +7,6 @@
   color: $white;
   position: relative;
 
-  .wrapper {
-    @extend .feed-wrapper;
-  }
-
   .affirmation__section {
     margin-bottom: $base-spacing;
     width: 100%;
--- a/resources/assets/components/Affirmation/affirmation.scss
+++ b/resources/assets/components/Affirmation/affirmation.scss
@@ -7,9 +7,8 @@
   color: $white;
   position: relative;
 
-  .affirmation__container {
-    max-width: 720px;
-    margin: auto;
+  .wrapper {
+    @extend .feed-wrapper;
   }
 
   .affirmation__section {
--- a/resources/assets/components/Affirmation/index.js
+++ b/resources/assets/components/Affirmation/index.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { FlexCell } from '../Flex';
 import Highlight from '../Highlight';
 import { Figure } from '../Figure';
+import { Wrapper } from '../Wrapper';
 import ShareContainer from '../../containers/ShareContainer';
 import { EMPTY_IMAGE } from '../../helpers';
 import './affirmation.scss';
@@ -12,7 +13,7 @@ const Affirmation = (props) => {
   return (
     <FlexCell width="full">
       <div className="affirmation">
-        <div className="wrapper">
+        <Wrapper width="feed">
           <div className="affirmation__section affirmation__section-heading">
             <Highlight>{ props.header }</Highlight>
           </div>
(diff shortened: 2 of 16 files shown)


deadlybutter:
1 million deploys later this PR is ready for review!

resources/assets/components/Affirmation/index.js
```diff
@@ -6,52 +6,36 @@ import ShareContainer from '../../containers/ShareContainer';
...
-            <div className="affirmation__section affirmation__section-heading">
-              <Highlight>{ this.props.header }</Highlight>
-            </div>
-            <div className="affirmation__section affirmation__section-quote">
-              <Figure image={this.props.photo} alt={this.props.author} alignment="left">
-                <p>{ this.props.quote }</p>
-                <span>- { this.props.author }</span>
-              </Figure>
+const Affirmation = (props) => {
+  if (! props.showAffirmation) return null;
```
DFurnes: Why aren't we handling this within the Chrome like we did before?

resources/assets/reducers/signups.js
```diff
@@ -26,6 +27,7 @@ const signupReducer = (state = {}, action) => {
         data: signups,
         isPending: false,
         thisSession: true,
```
DFurnes: Lets remove `thisSession` since it's been replaced by `showAffirmation`.

resources/assets/scss/next-toolbox.scss
```diff
@@ -17,3 +17,10 @@ $light-background-color: #F6F6F6;
 $dark-background-color: #000000;
 
 $half-spacing: $base-spacing / 2;
+
+.feed-wrapper {
```
DFurnes: We shouldn't output any CSS classes from the toolbox file, since they'll be duplicated each time we include this `next-toolbox.scss` from a stylesheet.

resources/assets/components/Affirmation/affirmation.scss
```diff
@@ -7,9 +7,8 @@
   color: $white;
   position: relative;
 
-  .affirmation__container {
-    max-width: 720px;
-    margin: auto;
+  .wrapper {
+    @extend .feed-wrapper;
```
DFurnes: This is a confusing abstraction to me... can we make a concrete class that's responsible for this layout pattern instead? I don't think I would otherwise guess that a "feed wrapper" applied to the affirmation block.

DFurnes: I would also avoid leaning on SCSS's `@extend` if possible since it makes it a lot harder to reason about what actual CSS is being outputted.

resources/assets/history.js
```diff
@@ -0,0 +1,29 @@
...
+import { syncHistoryWithStore } from 'react-router-redux';
+
+let history = null;
+
+/**
+ * Get the history object for this app
+ * @return {History}
+ */
+export function get() {
+  return history;
```
DFurnes: What's the benefit of this function over just exporting the `history` variable directly?

resources/assets/components/Affirmation/index.js
deadlybutter: Figured now that the affirmation component had the data it could determine that for itself, I think we do this in a few places?

resources/assets/reducers/signups.js
deadlybutter: I think we still want that for data purposes, would be good to know if someone did an action in the session they signed up. showAffirmation would turn false however if they close it out

resources/assets/components/Affirmation/affirmation.scss
deadlybutter: Sure!

resources/assets/history.js
deadlybutter: Not functional difference, but the `history` object isn't a const in this file so wanted to make it use a getter function

deadlybutter:
@DFurnes Updated this with a new `Wrapper` component, placed it in all of the places I know it was being used

deadlybutter:
[(3) As a Phoenix-next visitor that's Affiliating, I want to land on the Action tab so that I know how to take Action.](https://trello.com/c/AfH0yPzZ/440-3-as-a-phoenix-next-visitor-that-s-affiliating-i-want-to-land-on-the-action-tab-so-that-i-know-how-to-take-action)

PR #131, opened 2017-04-07
Preventing infinite authentication loop

deadlybutter:
Currently, if a user tries to back out of authenticating after clicking signup, they'll be sent right back to Northstar. This is because the event queue isn't handling whether they actually logged in or not. This PR fixes that!

Bug:
loop

Fix:
fixedd


deadlybutter:
[🐛 (5) Prevent background client tasks from running twice 🐛](https://trello.com/c/pyuGUKaE/496-5-prevent-background-client-tasks-from-running-twice)

PR #134, opened 2017-04-07
Add titles to Campaign Update blocks... and a bit more styling!

DFurnes:
#### Changes
I started this pull request just intending to add titles to the Campaign Update blocks and got a little carried away! This is totally not pixel-perfect, but gets the site looking much closer to the mocks so we'll look real nice for today's demo. 🎨

screen shot 2017-04-07 at 3 27 40 pm

### DIFF
--- a/resources/assets/app.js
+++ b/resources/assets/app.js
@@ -16,6 +16,7 @@ import { ready } from './helpers';
 import 'whatwg-fetch';
 
 // Style Components
+import './scss/chrome.scss';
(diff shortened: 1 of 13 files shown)


deadlybutter:
Not a merge blocker but the revealer could probably use a small style tweak to make it more readable against the new background

deadlybutter:
Same with the nav hover color ^

DFurnes:
@deadlybutter Oh yeah, good catch!

PR #138, opened 2017-04-09
Connecting action page to Contentful & styling the page

deadlybutter:
Felt like wrapping up Fridays work & figured might as well style it while I have the extra time:

action-small

action-large

In Contentful you can add up to 10 steps on the Campaign 
screen shot 2017-04-08 at 11 16 05 pm

Each step is a Campaign Action Step
screen shot 2017-04-08 at 11 18 42 pm
 

The supporting photos are positioned on the page based on the display option. Photos are not required to publish.


resources/assets/helpers/index.js
```diff
@@ -219,3 +219,24 @@ export function generateUniqueId() {
 export function isTimestampValid(timestamp, maxTime) {
   return (timestamp + maxTime) > Date.now();
 }
+
+/**
+ * Convert an int to a string. (Supports 0-10)
+ * @param  {int} number
+ * @return {String}
+ */
+export function convertDigitToWord(number) {
```
weerd: Maybe rename to `convertNumberToWord()` since the param passed is named `number`?

deadlybutter: Ah, yeah that would make more sense!

deadlybutter:
[(3) As a campaign lead, I want the content I put in Contentful to show up on the Action page so that users know how to complete the action.](https://trello.com/c/6m5TwrdG/429-3-as-a-campaign-lead-i-want-the-content-i-put-in-contentful-to-show-up-on-the-action-page-so-that-users-know-how-to-complete-the)

resources/assets/components/ActionPage/actionPage.scss
```diff
@@ -0,0 +1,57 @@
+@import '../../scss/next-toolbox.scss';
+
+.action-step {
+  margin-bottom: $base-spacing;
+
+  > .flex {
```
DFurnes: I'd like to be careful with overriding classes... not a merge blocker but let's think about ways we could avoid this, since it adds a less explicit dependency on how that "Flex" component works internally.

resources/assets/helpers/index.js
```diff
@@ -219,3 +219,24 @@ export function generateUniqueId() {
 export function isTimestampValid(timestamp, maxTime) {
   return (timestamp + maxTime) > Date.now();
 }
+
+/**
+ * Convert an int to a string. (Supports 0-10)
```
DFurnes: Let's throw an exception here so this doesn't fail silently if misused.

resources/assets/components/ActionPage/actionPage.scss
deadlybutter: Yeah I didn't enjoy making this one, I was pondering how we could make the Flex component accept custom styling options but figured that was a problem for later

resources/assets/helpers/index.js
deadlybutter: Sure!

PR #139, opened 2017-04-10
Adds signup errors

deadlybutter:
error

many-errors

This PR displays errors on the top of the page when you fail to signup. It has an underlying "notification" system which lets you push notifications to the user. (Currently, these are just errors, but flexible enough for future expansion). 

Also I've hard coded the error string for now, @ngjo can we confirm what the error text should be?


deadlybutter:
I'll go with Add/Remove - totally agree!

deadlybutter:
@weerd All re-tested - LMK if that works better!

PR #140, opened 2017-04-10
Lede banner updates woffing it up

weerd:
This PR updates the `LedeBanner` component to enable the affiliation button along with doing some cleanup and updates to the styles.

It also does a bit of code cleanup in a few other spots from recommendations in past PR which I didn't get a chance to get to before.

This PR also adds the new font files for Phoenix-Next. I've only really applied the font to the LedeBanner so it wouldn't get too crazy, but we might want to do a bit of thinking about the general SASS structure at some point.


resources/assets/init.js
```diff
@@ -27,6 +27,7 @@ import './scss/chrome.scss';
 import './scss/construction.scss';
 import './scss/container.scss';
 import './scss/navigation.scss';
+import './scss/fonts.scss';
```
weerd: I wasn't keen on importing this here, but I was getting lots of path issues (maybe from the url-loader?) when `@importing` in the toolbox file.

resources/assets/containers/ChromeContainer.js
```diff
@@ -1,14 +1,15 @@
...
 /**
  * Provide state from the Redux store as props for this component.
  */
 const mapStateToProps = (state, props) => {
   return {
     children: props.children,
-    hasNewSignup: state.signups.thisSession,
-    isSignedUp: state.signups.thisCampagin,
+    legacyCampaignId: state.campaign.legacyCampaignId,
+    isAffiliated: state.signups.thisSession || state.signups.thisCampaign,
```
DFurnes: I think this can just check `state.signups.thisCampaign`.

resources/assets/init.js
DFurnes: No this is perfect! Toolbox file should not contain anything that outputs CSS.

weerd: oh true. But it felt weird that the imports above it are all for general components for the most part.

resources/assets/components/LedeBanner/index.js
```diff
@@ -4,11 +4,13 @@ import { contentfulImageUrl } from '../../helpers';
 
 import './lede-banner.scss';
 
-const LedeBanner = ({title, subtitle, blurb, coverImage}) => {
+const LedeBanner = ({title, subtitle, blurb, coverImage, isAffiliated, legacyCampaignId, clickedSignUp}) => {
   const backgroundImageStyle = {
     backgroundImage: `url(${contentfulImageUrl(coverImage.url, '800', '600', 'fill')})`,
   };
 
+  const onClick = () => clickedSignUp(legacyCampaignId, {}); // @TODO: No metadata for now.
```
deadlybutter: Is this in the next PR? You can just add `{source: 'lede banner'}` for now

resources/assets/containers/ChromeContainer.js
weerd: Shouldn't it remove the button if they sign up for the campaign in the current session though?

resources/assets/components/LedeBanner/index.js
weerd: Yeah, I just wasn't sure what to add, since I haven't done any of the analytics stuff. Figured you'd jump in there. I can add that for now though.

resources/assets/scss/next-toolbox.scss
```diff
@@ -25,3 +25,10 @@ $dark-background-contrast-color: $white;
 
 // spacing
 $half-spacing: $base-spacing / 2;
+
+// Font Stacks
+$font-source-sans: "Source Sans Pro", "Helvetica Neue", "Helvetica", "Arial", sans-serif;
```
deadlybutter: Nothing related to merging this PR, but I think we need a card for using these everywhere in the existing codebase? Maybe override the base defaults?

resources/assets/containers/ChromeContainer.js
deadlybutter: @weerd thisCampaign should be true if thisSession is true

resources/assets/scss/next-toolbox.scss
weerd: Yeah, felt a bit hefty for this PR so I only modified what was in the LedeBanner for now. Something we should address soon.

resources/assets/init.js
DFurnes: I could see it making sense to put this in a new "base" section above the components.

PR #143, opened 2017-04-10
Gracefully handle missing fields in Contentful.

DFurnes:
We weren't checking if a `Asset` existed for an image before calling `->getFile()->getUrl()` or passing it to the `get_image_url` helper method. This pull request fixes a few of those issues.

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -101,7 +101,9 @@ function markdown($source)
  */
 function get_image_url($asset, $crop = 'landscape')
 {
-    if (! $asset) return null;
+    if (! $asset) {
+        return null;
+    }
 
     $options = [];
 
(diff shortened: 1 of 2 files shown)


app/helpers.php
```diff
@@ -95,10 +97,12 @@ function markdown($source)
  *
  * @param Asset $asset
  * @param  string $crop
- * @return string
+ * @return string|null
  */
-function get_image_url(Asset $asset, $crop = 'landscape')
+function get_image_url($asset, $crop = 'landscape')
 {
+    if (! $asset) return null;
```
weerd: brackets? Do we not have a rule for all conditionals to still use brackets even if only a single line? If not no worries, just don't remember offhand.

DFurnes: We do! Fixed.

PR #146, opened 2017-04-10
Cleanup and update to add hashing for React keys.

weerd:
This PR tries out adding a `makeHash()` helper method to turn provided strings into hashes. Did a little _post-work_ playing around with this because I didn't like the solution in place on RB Submission Gallery with how the keys were being assigned. It was actually not a useful method at all since on each re-render the key would end up being different based on how it was produced, which defeats the purpose of said React `key`s.

So this approach attempts to grab a string common and unique among both types of RBs (prior API requested submissions and new uploaded submission), takes the string, chops off a piece and makes a hash from it. The new key is the same after each re-render which will help React optimize how it renders the submissions gallery.

I'm still not super keen on the long prop name passed and use of the `||` for both the `key` and `url` props... but it works nicely otherwise and left a comment to ideally normalize this data elsewhere if possible :)

This PR also fixes a missing `/` for the profile link the Navigation template!

### DIFF
--- a/resources/assets/components/ReportbackUploader/index.js
+++ b/resources/assets/components/ReportbackUploader/index.js
@@ -4,6 +4,7 @@ import { Flex } from '../Flex';
 import MediaUploader from '../MediaUploader';
 import Gallery from '../Gallery';
 import ReportbackItem from '../ReportbackItem';
+import { makeHash } from '../../helpers';
 import './reportback-uploader.scss';
 
 class ReportbackUploader extends React.Component {
@@ -109,7 +110,8 @@ class ReportbackUploader extends React.Component {
           </form>
         </div>
           <Gallery isFetching={this.props.submissions.isFetching} type="triad">
-            {this.props.submissions.items.map((submission, index) => <ReportbackItem key={Date.now() + index} {...submission} url={submission.media.uri || submission.media.filePreviewUrl} reaction={null} />)}
+            {/* @TODO: Need to normalize data for uploaded RBs vs API retrieved RBs earlier in process if possible... */}
+            {this.props.submissions.items.map((submission, index) => <ReportbackItem key={makeHash((submission.media.uri || submission.media.filePreviewUrl).slice(-20))} {...submission} url={submission.media.uri || submission.media.filePreviewUrl} reaction={null} />)}
           </Gallery>
       </Block>
     );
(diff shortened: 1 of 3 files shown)


resources/assets/components/ReportbackUploader/index.js
```diff
@@ -109,7 +110,8 @@ class ReportbackUploader extends React.Component {
           </form>
         </div>
           <Gallery isFetching={this.props.submissions.isFetching} type="triad">
-            {this.props.submissions.items.map((submission, index) => <ReportbackItem key={Date.now() + index} {...submission} url={submission.media.uri || submission.media.filePreviewUrl} reaction={null} />)}
+            {/* @TODO: Need to normalize data for uploaded RBs vs API retrieved RBs earlier in process if possible... */}
+            {this.props.submissions.items.map((submission, index) => <ReportbackItem key={makeHash((submission.media.uri || submission.media.filePreviewUrl).slice(-20))} {...submission} url={submission.media.uri || submission.media.filePreviewUrl} reaction={null} />)}
```
deadlybutter: I feel like there is quite a lot happening here, maybe pull it out into a function?

resources/views/partials/navigation.blade.php
```diff
@@ -21,7 +21,7 @@
                 @if (Auth::user())
                     <a id="js-account-toggle" class="navigation__dropdown-toggle">My Profile</a>
                     <ul>
-                        <li><a href="{{ config('services.phoenix-legacy.url') . 'northstar/' . Auth::user()->northstar_id }}">Profile</a></li>
+                        <li><a href="{{ config('services.phoenix-legacy.url') . '/northstar/' . Auth::user()->northstar_id }}">Profile</a></li>
```
deadlybutter: I left the slashes out of these links intentionally since they were present in the last ones I copied from. On our current setup this would take you to `//northstar`. 

In retrospect, I don't like the fact we're basing app logic on whether the url has this trailing slash but not sure of a better idea, maybe a helper function? There are a ton of links I'll have to go back and fix so I wouldn't even bother with that in this pr.

weerd: @deadlybutter I think you're misreading what this change was. It's supposed to take you to your profile on Thor, not Northstar based on what I saw in the link. It was just a quick update cause the link was totally broken:

```
before:
https://thor.dosomething.orgnorthstar/58someidhere32342342

after
https://thor.dosomething.org/northstar/58someidhere32342342
```

If this is supposed to go to the profile on Northstar then that's a different story. I was just fixing the broken link, but didn't originally work on this so not sure where it's supposed to officially go!

deadlybutter:
👍 ⚡️

resources/views/partials/navigation.blade.php
weerd: TLDR; After much confusion, it turns out we have different values in on ENV files for our URLs. Some end in a `/` and some don't which is "no bueno". @deadlybutter suggested maybe having a safety function to wrap config values to deal w/ missing or extra `/` before outputting which seems like a good idea to me. For now, we'll just make sure all ENV values have a standard format: No trailing backslashes!

PR #147, opened 2017-04-11
Updates for new routing rules & cache issues.

DFurnes:
#### Changes

This pull request includes some changes for the updated Fastly routing rules.

🌐 Hosts campaigns at `/us/campaigns/{slug}`, since that allows us to seamlessly host this "within" the existing Drupal app. Also updates the client-side router to expect the extra prefix.

↪️ Adds a convenience redirect from `/campaigns/{path}` to `/us/campaigns/{path}`.

📜 Adds a hash to file names (e.g. `app-cad3da13eaff74eca3e4.js`) based on their contents, and creates an asset manifest so that Laravel can output the correct paths in our views. This should fix issues where Fastly is still serving old assets after a deploy.

References DoSomething/devops#240.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,5 +1,7 @@
 var webpack = require('webpack');
 var dotenv = require('dotenv').config();
+var ManifestPlugin = require('webpack-manifest-plugin');
+var ExtractTextPlugin = require('extract-text-webpack-plugin');
 var configurator = require('@dosomething/webpack-config');
 
 var config = configurator({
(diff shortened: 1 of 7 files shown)


webpack.config.js
```diff
@@ -9,6 +11,7 @@ var config = configurator({
 });
 
 // Override output path.
+config.output.filename = '[name]-[hash].js';
```
weerd: 👏

DFurnes:
[Unplanned Work: Fix app issues with new Fastly routing rules.](https://trello.com/c/WljXjHrz/507-unplanned-work-fix-app-issues-with-new-fastly-routing-rules)

PR #148, opened 2017-04-11
Adding signups total

deadlybutter:
Still WIP - The caching had unintended consequences, need to make a separate index function that caches the response & leave the original untouched

This PR grabs the total signups from phoenix-ashes and adds it to the client store. Additionally it adds some caching on the sign ups index since it tends to take a few seconds to load. This PR also accounts for updating the total when a user signs up.

This pr _does not_ display anywhere, only visible in the dev console

screen shot 2017-04-11 at 11 30 54 am

(Note the number change below)

screen shot 2017-04-11 at 11 37 54 am

### DIFF
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -38,7 +38,7 @@ class SignupController extends Controller
      */
     public function total($campaignId)
     {
-        return response()->json($this->phoenixLegacy->getAllSignups(['campaign' => $campaignId]));
+        return response()->json($this->phoenixLegacy->getAllSignupsCached(['campaign' => $campaignId]));
     }
 
     /**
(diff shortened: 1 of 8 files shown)


deadlybutter:
Need to resolve conflicts but other than that this is ready for review!

deadlybutter:
Also accidentally undid using the /total in the fender, ignore that for a sec, will be in next commit

deadlybutter:
Ok, now this PR is ready 😅

app/Http/Controllers/SignupController.php
```diff
@@ -31,6 +31,18 @@ public function index(Request $request)
...
 
     /**
+     * Get the total signups for this campaign.
+     *
+     * @param  int $campaignId
+     * @return \Illuminate\Http\Response
+     */
+    public function total($campaignId)
+    {
+        // Count = 1 is a hack to make the API actually return something. Don't ask.
```
DFurnes: "Don't ask". The official tagline of Phoenix Ashes.

PR #149, opened 2017-04-11
Put the correct global prefix in the right place, oy.

DFurnes:
#### Changes

I'm a silly boy and set campaigns to be at `/campaigns/us/*` rather than `/us/campaigns/*`. Oops!

🌐 ⬅️ 🇺🇸


DFurnes:
[Unplanned Work: Fix app issues with new Fastly routing rules.](https://trello.com/c/WljXjHrz/507-unplanned-work-fix-app-issues-with-new-fastly-routing-rules)

PR #151, opened 2017-04-11
Set `thisCampaign` on signups storage load

deadlybutter:
`thisCampaign` is false when the store is initiated, even if the signup is cached. this is blocking @weerd


deadlybutter:
screen shot 2017-04-11 at 12 48 19 pm

example of it working, see the beforeState & the thisCampaign values

PR #152, opened 2017-04-11
changing break all to break word

deadlybutter:
### before
pasted image at 2017_04_11 01_02 pm

### after
screen shot 2017-04-11 at 1 05 53 pm

### DIFF
--- a/resources/assets/components/Markdown/markdown.scss
+++ b/resources/assets/components/Markdown/markdown.scss
@@ -1,8 +1,6 @@
 @import '../../scss/next-toolbox.scss';
 
 .markdown {
-  word-break: break-word;
-
   blockquote {
     position: relative;
     margin: $base-spacing / 2 0;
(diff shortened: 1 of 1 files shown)


resources/assets/components/Markdown/markdown.scss
```diff
@@ -1,7 +1,7 @@
 @import '../../scss/next-toolbox.scss';
 
 .markdown {
-  word-break: break-all;
+  word-break: break-word;
```
weerd: Curious, but why is this even necessary?

weerd: Also, I don't think `break-word` is a valid value. I think it's probably just ignoring this rule and defaulting to normal. <https://developer.mozilla.org/en-US/docs/Web/CSS/word-break>

weerd: I think we're fine just removing this rule altogether 🙂

PR #153, opened 2017-04-11
Lazy load the action page step backgrounds.

DFurnes:
#### Changes
Noticed this while working on slow coffee shop wi-fi. This pull request updates step backgrounds to lazy-load, applies a CSS3 grayscale filter for the fun of it, and sets a gray background until then (or if a background wasn't provided).

screen shot 2017-04-11 at 4 10 25 pm

There's also a few updates to improve rendering of action page content:

screen shot 2017-04-11 at 4 31 54 pm

### DIFF
--- a/resources/assets/components/ActionPage/actionPage.scss
+++ b/resources/assets/components/ActionPage/actionPage.scss
@@ -17,14 +17,24 @@
   }
 
   .action-step__header {
+    position: relative;
     width: 100%;
-    background-size: cover;
-    background-position: center center;
+    background: $dark-gray;
     text-align: center;
     text-transform: uppercase;
     padding: $base-spacing;
+    overflow: hidden;
+
+    img {
+      width: 100%;
+      position: absolute;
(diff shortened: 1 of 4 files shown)


resources/assets/components/ActionPage/actionPage.scss
```diff
@@ -17,21 +17,32 @@
...
     text-transform: uppercase;
     padding: $base-spacing;
+    overflow: hidden;
+
+    img {
+      width: 100%;
+      position: absolute;
+      top: 50%; left: 0;
+      transform: translateY(-50%);
+      filter: grayscale(100%);
```
deadlybutter: I believe Luke/Creative are going to edit the photos as well, but i dont think a filter greyscale will have any visual impact on a greyscale image right?

DFurnes: Yeah, I think you're right. As far as I know, it shouldn't!

PR #155, opened 2017-04-11
Improved Northstar login flow.

DFurnes:
#### Changes
This pull request adds user-friendly "destination" names to the Northstar login flow. It also pins Contentful packages now that they have tagged releases.

References DoSomething/gateway#72.

🔒 🎨


DFurnes:
[(2) As a campaign lead, I want the Northstar page to be more clear that the user is signing into DoSomething's Sincerely, Us campaign and not Phoenix next so that it's not as confusing.](https://trello.com/c/6mZLTJ9U/469-2-as-a-campaign-lead-i-want-the-northstar-page-to-be-more-clear-that-the-user-is-signing-into-dosomething-s-sincerely-us-campaig)

DFurnes:
Just added 2323ca6 to fix a 404 when being redirected to login via the client-side app.

resources/views/campaigns/index.blade.php
```diff
@@ -1,28 +1,23 @@
...
         <div class="wrapper">
             <div class="container__block">
-                <p>Check out these sample campaigns, powered by Contentful!</p>
+                <h1>Campaigns</h1>
+                <p>Here's all the campaigns available on <strong>Phoenix Next</strong>, the new DoSomething.org
+                web experience.</p>
             </div>
             <ul class="gallery -quartet">
-                @foreach($campaigns as $campaign)
+                @foreach($campaigns as $link)
```
weerd: Just curious why you wanted to name this `$link` instead?

routes/web.php
```diff
@@ -13,7 +13,8 @@
 });
 
 // Authentication
-$router->get('next/login', 'AuthController@getLogin');
+$router->get('next/login', 'AuthController@getLogin')->name('login');
```
deadlybutter: super tiny / whatever, why did we rename login but not logout?

resources/views/campaigns/index.blade.php
deadlybutter: Yeah I think $campaign made more sense?

DFurnes: PHP (and Blade by extension) doesn't scope the iterator variables the way you'd expect, so `$campaign` was accessible outside the scope of the loop. This was causing the check for `$campaign` in the navigation partial to pull in the information for the last campaign.

deadlybutter: Thats... yikes. makes sense then 👍

weerd: Ah, ok. Maybe name it `$item` then? Link seems weird to me. :man_shrugging:

DFurnes: Ah, yeah `$item` is good by me.

routes/web.php
DFurnes: Wanted to use `route()` helper for login link, which relies on a named route. Doesn't hurt to add one for logout too, but didn't need it at the time.

resources/views/campaigns/index.blade.php
DFurnes: Updated in 0ce71ce.

routes/web.php
DFurnes: Added in 3ea8cbe.

PR #156, opened 2017-04-12
Wercker deploy

blisteringherb:
This PR includes the supporting code for deploying to phoenix-next servers via Wercker. https://app.wercker.com/dosomething/phoenix-next/runs

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -1,0 +1,55 @@
+# config/deploy.rb file
+require 'bundler/capistrano'
+
+set :application, "phoenix-next"
+set :deploy_to, ENV["DEPLOY_PATH"]
+role :app, ENV["SERVER_NAME"]
+role :web, ENV["SERVER_NAME"]
+server  ENV["SERVER_NAME"], :app, :web
+
+gateway = ENV["GATEWAY"]
+unless gateway.nil?
+  set :gateway, ENV["GATEWAY"]
+end
+
+set :user, "dosomething"
+set :group, "dosomething"
+set :use_sudo, false
+
+set :repository, "."
+set :scm, :none
+set :deploy_via, :copy
+set :keep_releases, 1
+
+ssh_options[:keys] = [ENV["CAP_PRIVATE_KEY"]]
+
+default_run_options[:shell] = '/bin/bash'
+
+namespace :deploy do
+  folders = %w{logs dumps system}
+
+  task :link_folders do
+    run "ln -nfs #{shared_path}/.env #{release_path}/"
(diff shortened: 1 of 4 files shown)


config/deploy.rb
```diff
@@ -0,0 +1,55 @@
...
+end
+
+set :user, "dosomething"
+set :group, "dosomething"
+set :use_sudo, false
+
+set :repository, "."
+set :scm, :none
+set :deploy_via, :copy
+set :keep_releases, 1
```
DFurnes: Do we want to set this to keep 5 releases [to avoid the "No input file specified" issue](https://github.com/DoSomething/devops/issues/82)?

config/deploy.rb
```diff
@@ -0,0 +1,55 @@
...
+
+ssh_options[:keys] = [ENV["CAP_PRIVATE_KEY"]]
+
+default_run_options[:shell] = '/bin/bash'
+
+namespace :deploy do
+  folders = %w{logs dumps system}
```
DFurnes: What are these `dumps` and `system` folders used for?

config/deploy.rb
```diff
@@ -0,0 +1,55 @@
...
+  task :artisan_migrate do
+    run "cd #{release_path} && php artisan migrate --force"
+  end
+
+  task :artisan_cache_clear do
+    run "cd #{release_path} && php artisan cache:clear"
+  end
+
+  task :restart_queue_worker, :on_error => :continue do
+    run "ps -ef | grep 'queue:work' | awk '{print $2}' | xargs sudo kill -9"
```
DFurnes: There's no queue worker on this application, so not sure if this is necessary.

wercker.yml
```diff
@@ -0,0 +1,31 @@
...
+          name: start mysql
+          code: sudo service mysql start
+      - leipert/composer-install@0.9.1
+      - wercker/bundle-install@1.1.1
+      - script:
+          name: npm install
+          code: |-
+            mkdir -p $WERCKER_CACHE_DIR/wercker/npm
+            npm config set cache $WERCKER_CACHE_DIR/wercker/npm
+            sudo npm install
```
DFurnes: Is `sudo` necessary here?

wercker.yml
```diff
@@ -0,0 +1,31 @@
...
+
```
DFurnes: I think we need to add an extra step here to build the application assets:

```yaml
    - script:
        name: build front-end assets
        code: npm run build
```

pnasrat:
👍  as I think this plus my #159 will fix the build

PR #157, opened 2017-04-12
Connecting Dashboard to contentful, dynamic variables

deadlybutter:
This PR grabs content from Contentful to fill the dashboard.

screen shot 2017-04-12 at 10 50 06 am

Additionally, it adds some dynamic variables which let the campaign lead replace text with stuff like the **end date** or the **total signups**. This is what that looks like in Contentful

screen shot 2017-04-12 at 10 51 21 am

### DIFF
--- a/resources/assets/components/Dashboard/index.js
+++ b/resources/assets/components/Dashboard/index.js
@@ -24,12 +24,12 @@ const Dashboard = (props) => {
       <FlexCell width='full'>
         <div className='dashboard'>
           <div className='dashboard__block -quarter'>
-            <h1>{ replaceTemplateVars(props.content.fields.leftValue) }</h1>
-            <span>{ replaceTemplateVars(props.content.fields.leftDescription) }</span>
+            <h1>{ replaceTemplateVars(props.content.fields.firstValue) }</h1>
+            <span>{ replaceTemplateVars(props.content.fields.firstDescription) }</span>
           </div>
           <div className='dashboard__block -quarter'>
-            <h1>{ replaceTemplateVars(props.content.fields.rightValue) }</h1>
-            <span>{ replaceTemplateVars(props.content.fields.rightDescription) }</span>
+            <h1>{ replaceTemplateVars(props.content.fields.secondValue) }</h1>
+            <span>{ replaceTemplateVars(props.content.fields.secondDescription) }</span>
           </div>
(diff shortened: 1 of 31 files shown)


weerd:
@deadlybutter I saw in Contentful that the end date is required. Are we sure that we want it required? Are there any "evergreen" campaigns we'll have on here? Not actually sure...

resources/assets/components/Dashboard/index.js
```diff
@@ -2,26 +2,40 @@ import React from 'react';
...
+
   return (
     <Flex>
       <FlexCell width='full'>
         <div className='dashboard'>
           <div className='dashboard__block -quarter'>
-            <h1>45 days</h1>
-            <span>until campaign closes</span>
+            <h1>{ replaceTemplateVars(props.content.fields.leftValue) }</h1>
+            <span>{ replaceTemplateVars(props.content.fields.leftDescription) }</span>
```
weerd: Not super keen on these fields being named based on visual location...

deadlybutter:
@weerd I can un-require it, good point!

resources/assets/components/Dashboard/index.js
deadlybutter: Any suggestions? Was drawing a blank on this one

weerd: So one of them typically holds "timeline" value and the other holds "member count" related data? Ideally something generic. Any suggestions @DFurnes. I could just see a scenario where we A/B test switching order around on dashboard stuff and that'll all just get confusing if one is name `Left Value` but in future shows up in the center or something (if) as the dashboard changes in design.

DFurnes: I think I'd be more in favor of `first` and `second` than left/right, but I can't really think of anything better than that. The problem is we don't really want to lock ourselves in with the semantic content here either, so I'm not sure there's a great name to be found. 😕

weerd: "Timeline" & "MemberTally"

weerd: @DFurnes hmmm, true. 🤔  Yeah, maybe `first` and `second` could work better...

deadlybutter:
@weerd @DFurnes Renamed!

PR #158, opened 2017-04-12
Redirect to action page if already signed up.

DFurnes:
#### Changes
This sends the user to the action page if they click a CTA block when already signed up.

🉑

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -111,8 +111,11 @@ export function clickedSignUp(campaignId, metadata) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


DFurnes:
[(3) As a Phoenix-next user, when I click on the post-Affiliation CTA button, I want to be taken to the Action page so that I can take action.](https://trello.com/c/K3vkGeCZ/494-3-as-a-phoenix-next-user-when-i-click-on-the-post-affiliation-cta-button-i-want-to-be-taken-to-the-action-page-so-that-i-can-tak)

PR #159, opened 2017-04-12
Add explicitly pinned extract-text-webpack-plugin

pnasrat:


DFurnes:
It's weird that this is needed since we pin this in [our webpack config](https://github.com/DoSomething/webpack-config/blob/03de03a1c4d1b588c3116cb3d953f5d65857cefa/package.json#L21), but whatever gets the job done!

⛵️

DFurnes:
@pnasrat Should we be merging #156 so we can see this with the build?

pnasrat:
I'll rebase when #156 is merged

pnasrat:
@blisteringherb or @DFurnes can you approve this plus the merged #156 fixes wercker and prod deploys.

PR #160, opened 2017-04-12
Fix links to not depend on trailing slash in the env

deadlybutter:
If your env file doesn't have a trailing slash, it shouldn't break all of the links surrounding the page

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -170,3 +170,19 @@ function getShareFields($campaign, $shareOverrides)
         'quote' => $shareOverrides ? $shareOverrides->quote : null,
     ];
 }
+
+/**
+ * Generate a link to Phoenix Ashes.
+ *
+ * @param  String $path path/to/something
+ * @return String
+ */
+function phoenixLink($path)
+{
+    $base = config('services.phoenix-legacy.url');
+    if (substr($base, -1) !== '/') {
+        $base .= '/';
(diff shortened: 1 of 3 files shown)


resources/views/partials/navigation.blade.php
```diff
@@ -21,7 +21,7 @@
                 @if (Auth::user())
                     <a id="js-account-toggle" class="navigation__dropdown-toggle">My Profile</a>
                     <ul>
-                        <li><a href="{{ config('services.phoenix-legacy.url') . '/northstar/' . Auth::user()->northstar_id }}">Profile</a></li>
+                        <li><a href="{{ phoenixLink('/northstar/' . Auth::user()->northstar_id) }}">Profile</a></li>
```
deadlybutter: Whoops, forgot an extra slash

app/helpers.php
```diff
@@ -170,3 +170,19 @@ function getShareFields($campaign, $shareOverrides)
...
     ];
 }
+
+/**
+ * Generate a link to Phoenix Ashes.
+ *
+ * @param  String $path path/to/something
+ * @return String
+ */
+function phoenixLink($path)
```
weerd: You could maybe universalize it for any DS related link... :man_shrugging: But not crucial for merge and something we could do later :)

PR #163, opened 2017-04-12
Fixing a few things with signups

deadlybutter:
This should _hopefully_ be the end of all the bugginess. 

- Dispatched the `completed event` before the action itself, this is because if we have redirects it will prevent the completed event from firing and thus turning into an infinite loop.
- Properly created a new redux state instead of modifying it
- Set more variables in the signups reducer that was missing before
- Changed the event logic to be more clear

### DIFF
--- a/resources/assets/actions/event.js
+++ b/resources/assets/actions/event.js
@@ -24,13 +24,19 @@ export function startQueue() {
     const queue = getArray(getDeviceId(), EVENT_STORAGE_KEY);
 
     queue.forEach((event, index) => {
+      // Always remove the event from storge.
+      dispatch(completedEvent(index));
+
       // Check if the event is over 30 min old before dispatching.
       const isValidTimestamp = isTimestampValid(event.createdAt, (30 * 60 * 1000));
 
       // Check if the user successfully authenticated
       const isAuthenticated = typeof getState().user.id !== 'undefined';
 
-      if (!isValidTimestamp || (event.requiresAuth && !isAuthenticated)) {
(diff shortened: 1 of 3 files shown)


resources/assets/actions/event.js
```diff
@@ -24,23 +24,26 @@ export function startQueue() {
     const queue = getArray(getDeviceId(), EVENT_STORAGE_KEY);
 
     queue.forEach((event, index) => {
+      // Always remove the event from storge.
```
weerd: `storge` typo ;)

deadlybutter: Its HTML6!

resources/assets/actions/event.js
```diff
@@ -24,23 +24,26 @@ export function startQueue() {
...
+
       // Check if the event is over 30 min old before dispatching.
       const isValidTimestamp = isTimestampValid(event.createdAt, (30 * 60 * 1000));
 
       // Check if the user successfully authenticated
       const isAuthenticated = typeof getState().user.id !== 'undefined';
 
-      if (!isValidTimestamp || (event.requiresAuth && !isAuthenticated)) {
+      let shouldFireEvent = isValidTimestamp;
+      if (shouldFireEvent && event.requiresAuth) shouldFireEvent = isAuthenticated;
```
weerd: can we stick w/ `{...}` brackets :)

resources/assets/actions/event.js
weerd: 🤔

PR #164, opened 2017-04-12
Campaign content page updates

weerd:
This PR updates the campaign content page structure and styles. I wanted to break this up a bit before it got too involved. My aim was not to mess around too much with the block styles which didn't work as great for these content pages, so I implemented a slightly different structure. Some of the styles will get updated as the CTAs get updated styles in the next PR @_@

### DIFF
--- a/resources/assets/components/Block/block.scss
+++ b/resources/assets/components/Block/block.scss
@@ -1,12 +1,9 @@
 @import '../../scss/next-toolbox';
 
 .block {
+  @extend %block;
+  margin: $half-spacing;
   width: 100%;
-  background: #fff;
-  margin: $base-spacing / 2;
-  padding: $base-spacing / 2;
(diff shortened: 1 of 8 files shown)


resources/assets/components/Block/block.scss
```diff
@@ -1,19 +1,16 @@
 @import '../../scss/next-toolbox';
 
 .block {
+  @extend %block;
```
DFurnes: I don't really think we need a placeholder selector for this, but not merge-blocking – let's revisit post launch.

weerd: I'm thinking of calling it something else, but seemed potentially useful for mimicking the same styles in both a "block" and in the content pages enclosing div or whatever. Could be some other approach too. We can defs chat about it post launch :)

PR #171, opened 2017-04-13
Quick debugging tool

deadlybutter:
This should make it easier for admins to peak at the Analytics/Signup storage stuff and also clear local storage when trying to test

screen shot 2017-04-13 at 11 27 34 am


deadlybutter:
@weerd @DFurnes No rush on this - but might be helpful for debugging & testing today + tomorrow

weerd:
Just a suggestion but maybe name component `Debug` or `Debugger`since you're not actually configuring anything in there, right?

resources/assets/components/Config/index.js
```diff
@@ -0,0 +1,38 @@
...
+import {
+  getArray as storageGetArray,
+  remove as storageRemove,
+  SIGNUP_STORAGE_KEY
+} from '../../helpers/storage';
+import { getSession } from '../../helpers/analytics';
+import './config.scss';
+
+const Config = (props) => {
+  if (props.user.role !== 'admin') return null;
```
weerd: ye ole' brackets `{ ... }` :)

weerd: even for one liners, when we eventually add some eslinting, it'll yell at us based on our DS eslint config.

resources/assets/components/Config/index.js
```diff
@@ -0,0 +1,38 @@
...
+
+  return (
+    <div className="config">
+      <Flex>
+        <FlexCell width="one-third">
+          <div>
+            <h1>analytics debug</h1>
```
weerd: all lowercase for an `<h1>`. bold sir.

deadlybutter: cc: NYTCopyDesk

deadlybutter:
@weerd I think this reads better! Seems to still work

deadlybutter:
ALSO PSA

This delete signup button *DOES NOT* update the store (Trying to timebox this..). So you have to refresh the page to see any updates. (And remember to delete your signup on Phoenix)

PR #172, opened 2017-04-13
Cleaning up styles and output for the campaign content pages.

weerd:
This PR does some additional tweaks, responsive fixes, etc to wrap up the Campaign content pages for now, with the two hardcoded CTAs.

Refs https://trello.com/c/eCW7M75Z/394-3-as-a-developer-i-want-to-be-able-to-add-cta-blocks-to-the-faq-and-scholarship-pages-so-that-i-can-push-users-to-affiliate-or-t

### DIFF
--- a/resources/assets/components/CallToActionBlock/cta.scss
+++ b/resources/assets/components/CallToActionBlock/cta.scss
@@ -8,6 +8,12 @@
   margin: 12px;
   width: 100%;
 
+  &.-transparent {
+    background-color: transparent;
+    border: 0 none;
+    color: $off-black;
+  }
+
   &.has-photo {
     @include media($medium) {
       display: flex;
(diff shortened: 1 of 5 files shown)


resources/assets/components/Markdown/index.js
```diff
@@ -3,7 +3,7 @@ import classnames from 'classnames';
 import { markdown } from "../../helpers";
 import './markdown.scss';
 
-const Markdown = ({className = 'markdown', children}) => (
+const Markdown = ({className = null, children}) => (
```
weerd: We already have `markdown` always being output, so this was outputting `markdown` as a class twice. `classnames` will ignore outputting `null`, so this should be safe.

resources/assets/components/CallToActionBlock/cta.scss
```diff
@@ -26,44 +32,51 @@
...
+  .cta__title {
+    font-size: $font-medium;
+    font-weight: $weight-sbold;
+  }
 
-.cta__message {
-  font-size: $font-regular;
-  font-weight: $weight-normal;
-}
+  .cta__message {
```
DFurnes: Do these need to be nested? Not a big deal, but feels like a specificity problem waiting to happen.

resources/assets/components/Markdown/index.js
DFurnes: Good catch!

resources/assets/components/CallToActionBlock/cta.scss
weerd: Yep, the specificity problem came from Forge, so I had to nest them 😞

DFurnes: Ahhh, makes total sense. 👍

PR #176, opened 2017-04-13
Removes min-height from action page photos

lkpttn:
This PR just removes min-height from action steps photos. 

This is the easiest way to handle the image stretching on mobile and desktop sizes. Later on, we should make these into squares with proper overflows and such.

### DIFF
--- a/resources/assets/components/ActionPage/actionPage.scss
+++ b/resources/assets/components/ActionPage/actionPage.scss
@@ -62,7 +62,7 @@
       padding: $base-spacing / 2;
 
       img {
-        min-height: 200px;
+/*         min-height: 200px; */
       }
     }
   }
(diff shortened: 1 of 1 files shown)


resources/assets/components/ActionPage/actionPage.scss
```diff
@@ -62,7 +62,7 @@
       padding: $base-spacing / 2;
 
       img {
-        min-height: 200px;
+/*         min-height: 200px; */
```
deadlybutter: I would just delete the entire selector

weerd: ya, don't comment it out if removing it. just delete it! 🗑

lkpttn: Do we want to keep if we know we'll need to target it again when we actually fix this?

deadlybutter: @lkpttn Leave a `TODO: ...`

PR #177, opened 2017-04-13
Some style/layout fixes

deadlybutter:
## Increases the length of the community tab to 5 rows

screen shot 2017-04-13 at 1 32 23 pm

## Cuts the action page step off sooner

screen shot 2017-04-13 at 1 27 07 pm
screen shot 2017-04-13 at 1 26 59 pm

## Fixes the affirmation header on mobile

screen shot 2017-04-13 at 1 26 18 pm

### DIFF
--- a/resources/assets/components/ActionPage/actionPage.scss
+++ b/resources/assets/components/ActionPage/actionPage.scss
@@ -12,7 +12,7 @@
     }
   }
 
-  &.-hide {
+  &.-truncate {
     overflow-y: hidden;
     max-height: 600px;
 
--- a/resources/assets/components/ActionPage/actionPage.scss
+++ b/resources/assets/components/ActionPage/actionPage.scss
@@ -12,6 +12,15 @@
     }
   }
 
+  &.-hide {
+    overflow-y: hidden;
+    max-height: 600px;
+
+    @include media($medium) {
+      max-height: 400px;
+    }
+  }
+
   h1, h2, h3, p {
     color: $black;
     font-family: $primary-font-family;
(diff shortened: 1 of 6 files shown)


deadlybutter:
@lkpttn Should that header in the affirmation be smaller?

resources/assets/components/ActionPage/actionPage.scss
```diff
@@ -12,6 +12,15 @@
     }
   }
 
+  &.-hide {
```
DFurnes: Would "truncate" make more sense? Was confused what this was doing at first glance.

resources/assets/components/ActionPage/index.js
```diff
@@ -56,6 +58,7 @@ const renderStep = (step, index) => {
 const ActionPage = ({ steps, callToAction, campaignId, signedUp, hasPendingSignup, isAuthenticated, clickedSignUp }) => {
   if (! signedUp) {
     steps = steps.slice(0, 2);
+    steps[steps.length - 1].hide = true;
```
DFurnes: We should check if this index exists before modifying it. Otherwise, we'd get a ⚠️ on new campaigns.

resources/assets/components/ActionPage/actionPage.scss
deadlybutter: Yeah I think that sounds nicer!

lkpttn:
@deadlybutter I think I would rather lose the emoji before making the text smaller.

deadlybutter:
@DFurnes Renamed, reads better!

PR #183, opened 2017-04-13
Hopefully fixing reportback uploads

deadlybutter:
This forces the RB upload to use the app url & a /next prefix

### DIFF
--- a/app/Http/Controllers/ReportbackController.php
+++ b/app/Http/Controllers/ReportbackController.php
@@ -59,7 +59,7 @@ class ReportbackController extends Controller
             auth()->id(),
             $request->input('campaignId'),
             [
-                'file_url' => config('app.env') !== 'local' ?  config('url').'/next'.$path : 'https://placeimg.com/1000/768/animals',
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/ReportbackController.php
```diff
@@ -59,7 +59,7 @@ public function store(Request $request)
             auth()->id(),
             $request->input('campaignId'),
             [
-                'file_url' => config('app.env') !== 'local' ?  asset($path) : 'https://placeimg.com/1000/768/animals',
+                'file_url' => config('app.env') !== 'local' ?  config('url').'/next'.$path : 'https://placeimg.com/1000/768/animals',
```
weerd: is it `config('app.url')`?

deadlybutter: I didnt see an app prefix in the file but maybe thats applied because its in the app file... let me double check

DFurnes: Should this be `config('app.url')`?

DFurnes: Oh, too slow! 😆

PR #184, opened 2017-04-13
Cards not jeans!

deadlybutter:
One more quick copy fix, we'll make this dynamic later


deadlybutter:
Accidentally borked this branch

PR #189, opened 2017-04-14
Reportback uploader validation messages

weerd:
This PR adds form validation errors displaying inside the Reportback Uploader form so a user can understand what is wrong if an error occurs when submitting a reportback. It sets up a system for showcasing messaging of different kinds in the form using a `FormMessage` component, so we can show other info (submission was fine but something happened on the receiving server) or show a success message.

For the next upcoming PR I'll be adding the success message and also persisting the field data in the form if an error occurs so that the user can amend what they entered. Currently, if there are errors all the fields will be wiped, but this is expected... for now.

### DIFF
--- a/resources/assets/actions/reportback.js
+++ b/resources/assets/actions/reportback.js
@@ -52,9 +52,12 @@ export function storeReportback(reportback) {
   };
 }
 
-// Action: storeing new user submitted reportback failed.
-export function storeReportbackFailed(reportback) {
-  return { type: STORE_REPORTBACK_FAILED };
+// Action: storing new user submitted reportback failed.
+export function storeReportbackFailed(error) {
+  return {
+    type: STORE_REPORTBACK_FAILED,
+    error
+  };
 }
 
 // Action: storing new user submitted reportback was successful.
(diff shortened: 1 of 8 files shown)


resources/assets/actions/reportback.js
```diff
@@ -152,8 +155,9 @@ export function submitReportback(reportback) {
     })
       .then(response => {
         if (response.status >= 300) {
-          dispatch(storeReportbackFailed());
-          // @TODO: implement showing validation error.
+          response.json().then(json => {
+            dispatch(storeReportbackFailed(json));
```
DFurnes: Do we want to dispatch this error in the `.catch()` handler down below as well?

weerd: Potentially. It could be something I update, but I remember wanting to specifically deal w/ status codes greater than `300` separately from other errors that were caught. I don't remember offhand why that was though... I need to write more things down 😢

ngjo:
[(5) As an Affiliated member trying to reportback, I want some kind of error notification so that I know reportingback has failed or success notification if it has succeeded](https://trello.com/c/0mGPzVdv/389-5-as-an-affiliated-member-trying-to-reportback-i-want-some-kind-of-error-notification-so-that-i-know-reportingback-has-failed-or)

PR #194, opened 2017-04-14
Luke fixes more photos

lkpttn:
Adds a float to photos inside a campaign update block, so they don't just get stuck on top of each other.

### DIFF
--- a/resources/assets/components/CampaignUpdateBlock/campaign-update.scss
+++ b/resources/assets/components/CampaignUpdateBlock/campaign-update.scss
@@ -8,7 +8,7 @@
   font-size: $font-large;
 }
 
-.campaign-update__content. p img {
+.campaign-update__content p img {
   float: left;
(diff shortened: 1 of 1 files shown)


resources/assets/components/CampaignUpdateBlock/campaign-update.scss
```diff
@@ -7,3 +7,9 @@
 .campaign-update__content.-tweet p {
   font-size: $font-large;
 }
+
+.campaign-update__content p img {
+  float: left;
+  margin-right: $base-spacing/2;
+  margin-bottom $base-spacing;
```
DFurnes: @lkpttn You're missing a `:` here between the property name and value!

DFurnes:
I'm still not sure this is working as expected.

screen shot 2017-04-18 at 1 16 10 pm

PR #200, opened 2017-04-14
Reportback uploader messaging updates

weerd:
Hokay! This PR adds success messaging, disables the submit button while the uploader is "thinking" 🤔 

We're holding off on resetting the form only after successful submit, once we move the RB Uploader state to the Redux store. It was proving a bit hairy to reset the form since there's some state change updates and React gets mad if that's done in the wrong place.


deadlybutter:
kool ⚡️

PR #201, opened 2017-04-14
Show entire step post signup

deadlybutter:
Currently, we modify the store which breaks the action page post signup. This makes a copy of the array so it works on re-render.


resources/assets/components/ActionPage/index.js
```diff
@@ -56,11 +56,13 @@ const renderStep = (step, index) => {
  * @returns {XML}
  */
 const ActionPage = ({ steps, callToAction, campaignId, signedUp, hasPendingSignup, isAuthenticated, clickedSignUp }) => {
+  let actionSteps = JSON.parse(JSON.stringify(steps));
```
DFurnes: 🤔

DFurnes: Can you try this with Lodash's [`cloneDeep`](https://lodash.com/docs/#cloneDeep)?

```js
import cloneDeep from 'lodash/cloneDeep';

let actionSteps = cloneDeep(steps);
```

deadlybutter:
@DFurnes Lodash worked!

ngjo:
[🐛 Parts of Step 2 in the Action page are hidden even after Affiliation 🐛](https://trello.com/c/dTMBPsVl/539-parts-of-step-2-in-the-action-page-are-hidden-even-after-affiliation)

PR #205, opened 2017-04-14
Quick fix for a visual bug with the plus sign on the uploader.

weerd:
This PR makes a quick fix for when uploading images and the plus sign stays put in the media preview! 🌵


ngjo:
[As a user, I want it to be more clear where I upload my photo](https://trello.com/c/Q4pHOZIN/525-as-a-user-i-want-it-to-be-more-clear-where-i-upload-my-photo)

PR #206, opened 2017-04-17
Adding eslint via DS eslint config.

weerd:
This PR adds eslint using our [DS ESLint config](https://github.com/DoSomething/eslint-config/) and cleaning up some low-hanging fruit! 🍎🍐🍊🍌🍉 

From 200+ linting errors we're down to about 60, which are ones that related to the rules specified below or I needed to spend a little bit more time on.

The following are some rules that produced errors when linting, but maybe something we should discuss if we want to change how the linter respects these rules and based on how we write our code:

- `no-use-before-define`
- `max-len`
- `no-case-declarations`
- `arrow-body-style`
- `space-unary-ops`
- `no-bitwise`
- `react/jsx-filename-extension`
- `brace-style`

Currently, I did not hook up the linting to webpack to avoid a lot of noise. Once we discuss the above rules and then wrap up fixing the last of the linting errors we can hook webpack up to lint on the file as we code :)

### DIFF
--- a/resources/assets/actions/reportback.js
+++ b/resources/assets/actions/reportback.js
@@ -1,3 +1,5 @@
+/* global window, document */
+
 import { Phoenix } from '@dosomething/gateway';
 import has from 'lodash/has';
 import get from 'lodash/get';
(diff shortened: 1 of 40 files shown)


resources/assets/actions/reportback.js
```diff
@@ -1,7 +1,9 @@
+/* global window, document */
```
deadlybutter: whats this?

resources/assets/actions/reportback.js
```diff
@@ -82,60 +84,60 @@ export function addSubmissionMetadata(reportback, id) {
...
+    reportbackItem,
+  };
 }
 
 // Async Action: user reacted to a photo.
 export function toggleReactionOn(reportbackItemId, termId, metadata) {
   return (dispatch, getState) => {
     // If the user is not logged in, handle this action later.
-    if (! getState().user.id) {
+    if (!getState().user.id) {
```
deadlybutter: Just as I started getting used to having a space! darn.

resources/assets/actions/signup.js
```diff
@@ -88,25 +88,25 @@ export function setTotalSignups(total) {
...
-    (new Phoenix).get(`next/signups/total/${campaignId}`).then(response => {
+    (new Phoenix()).get(`next/signups/total/${campaignId}`).then((response) => {
       if (!response || !response.meta || !response.meta.pagination) {
         throw new Error('no signup metadata found');
       }
 
       let total = response.meta.pagination.total;
       // TODO: This isn't ideal, but the browser doesn't know if this is an old cached response or not.
-      if (getState().signups.thisCampaign) total++;
+      if (getState().signups.thisCampaign) total += 1;
```
deadlybutter: oh come on linter!!!!

resources/assets/actions/reportback.js
weerd: Tells ESLint to ignore these variables which are globally defined, otherwise it throws an error since they're not defined in the file.

resources/assets/actions/signup.js
weerd: Apparently it's a bad thing to use `++`. Can lead to errors and it's just a better practice to be more explicit with these.

resources/assets/actions/reportback.js
DFurnes: Yeah I've really grown fond of this weird style rule. 😆

weerd: Yeah, we can change the ruling on this. It's something we can discuss if we want to.

resources/assets/components/Block/index.js
```diff
@@ -10,10 +10,8 @@ BlockTitle.PropTypes = {
...
 };
 
-export const Block = (props) => (
+export default Block = (props) => (
   <div className={classNames('block', props.className)}>
     {props.children}
   </div>
 );
-
-export default Block;
```
DFurnes: Aww I also really liked this define & then export pattern. Damn you, AirBnB!

weerd: We can chat about it, lol.

deadlybutter:
:shipit:

weerd:
@DFurnes Regarding https://github.com/DoSomething/phoenix-next/pull/206#pullrequestreview-33059373 turns out I need to leave it as is for now. While eslint still yells, the app was not working because it could not find the export @_@ So leaving as is for now. It relates to the following rule: [import/no-named-as-default](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-named-as-default.md)

PR #208, opened 2017-04-17
Heroku prod

deadlybutter:
This PR makes two quick fixes so

1) Heroku prod deploy works
2) Users can back out of Northstar auth

### DIFF
--- a/bootstrap/setup.sh
+++ b/bootstrap/setup.sh
@@ -1,4 +1,5 @@
 ## Used by Heroku review apps to setup the application
 
 php artisan migrate --seed --force
+npm install --only=dev
 npm run build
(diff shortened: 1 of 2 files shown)


bootstrap/setup.sh
```diff
@@ -1,4 +1,5 @@
 ## Used by Heroku review apps to setup the application
 
 php artisan migrate --seed --force
+npm install --only=dev
```
weerd: ?

deadlybutter: on prod the NODE_ENV is production, and moderniz & whatnot isnt installed which breaks the build. so this just makes sure that all gets installed

deadlybutter: a npm install is run by the Heroku buildpack already, so we already have the normal dependencies at this point

weerd: 🤔

Issue #207, opened 2017-04-17
Users are being given the wrong session.


DFurnes:
_BUG_
-------------------
#### Current Behavior
Currently, most (all?) of the users on Phoenix Next production are getting the session for [the same user](https://aurora.dosomething.org/users/5544d41e469c64ec7d8b48f9). This means we're all seeing her reportbacks, and reporting back on her behalf.

#### Desired Behavior
That shouldn't be happening.

#### Steps to Replicate
It seems like everyone who goes to [the campaign](https://www.dosomething.org/us/campaigns/sincerely-us/action) has this issue.

You can see which Northstar ID you have by dumping out the preloaded state in `window.STATE.user`:

<img width="374" alt="screen shot 2017-04-17 at 11 16 09 am" src="https://cloud.githubusercontent.com/assets/583202/25093224/458899a0-235f-11e7-998e-5c98a94666ac.png">

#### Why This Matters
We want users to be able to log in as _themselves_!

#### Countries Affected (optional)
…

#### OS/Browser (optional)
…

#### Relevant Screenshots + Links
- [Phoenix Next production](https://www.dosomething.org/us/campaigns/sincerely-us/action)
- [Patient 0's profile](https://aurora.dosomething.org/users/5544d41e469c64ec7d8b48f9)

deadlybutter:
<img width="960" alt="screen shot 2017-04-17 at 11 17 36 am" src="https://cloud.githubusercontent.com/assets/897368/25093268/7eae6f2a-235f-11e7-9819-e29461a846a3.png">

Worked for me on login just now

deadlybutter:
<img width="365" alt="screen shot 2017-04-17 at 11 18 28 am" src="https://cloud.githubusercontent.com/assets/897368/25093292/99175fa2-235f-11e7-8564-277907dd68cf.png">

DFurnes:
Phoenix Next authenticates using [Gateway](https://github.com/DoSomething/gateway) and Laravel's built-in authentication & Redis session driver. We have seen some odd behavior with [mismatched state tokens](https://github.com/DoSomething/phoenix-next/issues/188) on this app, but otherwise this is all the exact same authentication flow we use on Aurora, Gladiator, Rogue, Voting App, and Celebs Gone Good (and the same Northstar endpoint as we use w/ Drupal module in Phoenix Ashes).

DFurnes:
So far we've tried:
[...]

PR #210, opened 2017-04-17
Fix issue where we would fetch wrong ID column.

DFurnes:
#### Changes
Oh my _god_. This fixes the great "Chloe debacle" of 2017. References #207.

😩

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -13,6 +13,20 @@ class User extends Model implements AuthenticatableContract, NorthstarUserContra
     use Authenticatable, HasNorthstarToken;
 
     /**
+     * The primary key for the model.
+     *
+     * @var string
+     */
+    protected $primaryKey = 'northstar_id';
+
+    /**
+     * Indicates if the IDs are auto-incrementing.
+     *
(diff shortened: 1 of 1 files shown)


deadlybutter:
wow

deadlybutter:
are we worried changing the primary key will break anything?

weerd:
Why did nothing show up regarding this in the past though?

DFurnes:
> are we worried changing the primary key will break anything?

It shouldn't! In most cases we were specifically asking for this column.

DFurnes:
> Why did nothing show up regarding this in the past though?

Wrote up a little bit in the linked issue and we should definitely do a retrospective, but the gist is that it was trying to find a integer user `id` based on the string `northstar_id` (which was always `58f...`), and until we had 58 users it was always returning `null`.

As soon as any environment had 58 users (so, Friday afternoon 🙃) it would break in the same way.

weerd:
Ah, one of those _fun_ 🐛s. I feel like something similar to this has happened to us before... Thanks MongoIDs! lol

PR #217, opened 2017-04-19
Session fix

deadlybutter:


DFurnes:
What keeps happening with all these "Fixing merge" commits?

PR #218, opened 2017-04-19
Run linter on Wercker builds.

DFurnes:
#### Changes
This pull request adds a linting step that will fail Wercker builds if there are style problems. It also fixes a ton more style issues (womp womp) that we found when fixing the lint command syntax.

This also updates `React.PropTypes` to reference the `prop-types` package instead via the [code mod](https://github.com/reactjs/react-codemod#react-proptypes-to-prop-types).

✏️ 💯

### DIFF
--- a/resources/assets/components/ActionPage/index.js
+++ b/resources/assets/components/ActionPage/index.js
@@ -1,15 +1,16 @@
 import React from 'react';
 import classnames from 'classnames';
+import { cloneDeep } from 'lodash';
+
+import LazyImage from '../LazyImage';
 import Markdown from '../Markdown';
 import ReportbackUploaderContainer from '../../containers/ReportbackUploaderContainer';
 import Revealer from '../Revealer';
-import LazyImage from '../LazyImage';
 import { Flex, FlexCell } from '../Flex';
 import { convertNumberToWord } from '../../helpers';
-import cloneDeep from 'lodash/cloneDeep';
 import './actionPage.scss';
 
-const Stepheader = ({ title, step, background }) => (
+const StepHeader = ({ title, step, background }) => (
   <FlexCell width="full">
     <div className="action-step__header">
       <LazyImage src={background} />
@@ -19,12 +20,32 @@ const Stepheader = ({ title, step, background }) => (
   </FlexCell>
 );
 
+StepHeader.propTypes = {
+  title: React.PropTypes.string.isRequired,
+  step: React.PropTypes.number.isRequired,
+  background: React.PropTypes.string.isRequired,
+};
+
+/**
+ * Render a photo on the action page.
+ *
+ * @param step
+ * @param index
+ * @returns {XML}
+ */
 const renderPhoto = (photo, index) => (
   <div className="action-step__photo" key={index}>
     <img src={photo} />
   </div>
 );
 
+/**
+ * Render a single step on the action page.
+ *
+ * @param step
+ * @param index
+ * @returns {XML}
+ */
 const renderStep = (step, index) => {
   const title = step.title;
   const background = step.background;
@@ -34,9 +55,9 @@ const renderStep = (step, index) => {
 
   return (
     <FlexCell width="full" key={index}>
-      <div className={classnames('action-step', {'-truncate': shouldTruncate})}>
+      <div className={classnames('action-step', { '-truncate': shouldTruncate })}>
         <Flex>
-          <Stepheader title={title} step={index + 1} background={background} />
+          <StepHeader title={title} step={index + 1} background={background} />
           <FlexCell width="two-thirds">
             <Markdown>{ step.content }</Markdown>
           </FlexCell>
@@ -48,17 +69,22 @@ const renderStep = (step, index) => {
         </Flex>
       </div>
     </FlexCell>
-  )
-}
+  );
+};
 
 /**
  * Render the feed.
  *
  * @returns {XML}
  */
-const ActionPage = ({ steps, callToAction, campaignId, signedUp, hasPendingSignup, isAuthenticated, clickedSignUp }) => {
-  let actionSteps = cloneDeep(steps);
+const ActionPage = (props) => {
+  const {
+    steps, callToAction, campaignId, isAuthenticated,
+    signedUp, hasPendingSignup, clickedSignUp,
+  } = props;
 
+  // Truncate steps if user isn't signed up.
+  let actionSteps = cloneDeep(steps);
   if (! signedUp) {
     actionSteps = actionSteps.slice(0, 2);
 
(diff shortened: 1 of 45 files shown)


resources/assets/components/ActionPage/index.js
```diff
@@ -86,7 +117,23 @@ const ActionPage = ({ steps, callToAction, campaignId, signedUp, hasPendingSignu
...
+  callToAction: React.PropTypes.string.isRequired,
+  campaignId: React.PropTypes.string.isRequired,
+  hasPendingSignup: React.PropTypes.bool.isRequired,
+  isAuthenticated: React.PropTypes.bool.isRequired,
+  signedUp: React.PropTypes.bool.isRequired,
+  clickedSignUp: React.PropTypes.func.isRequired,
+};
+
 ActionPage.defaultMetadata = {
+  steps: [],
```
deadlybutter: whats this?

resources/assets/components/CallToActionBlock/index.js
```diff
@@ -19,46 +19,61 @@ const renderImpactContent = (data) => {
...
+  imageUrl: React.PropTypes.string.isRequired,
+  campaignId: React.PropTypes.string.isRequired,
+  clickedSignUp: React.PropTypes.func.isRequired,
+  modifierClasses: React.PropTypes.string, // @TODO: Should this be an array?
+};
+
 CallToActionBlock.defaultMetadata = {
   source: 'call to action block',
-}
+  modifierClasses: [],
```
deadlybutter: 🚫  as well -- unless we're putting data on this (which i dont see anywhere)

resources/assets/components/ContentPage/index.js
```diff
@@ -1,17 +1,15 @@
...
 import './content-page.scss';
 
 const ContentPage = ({ pages, route }) => {
-  const page = pages.find(page => page.fields.slug === route.page);
+  const page = pages.find(item => item.fields.slug === route.page);
 
   // @TODO: temporary variables until these CTAs are no longer hardcoded.
-  const ctaText1 = 'Help us send letters of support to every mosque in the US.\n\n__Join hundreds of members members making cards!__';
-  const ctaText2 = 'Help us send letters of support to every mosque in the United States.';
+  const ctaText1 = { content: 'aHelp us send letters of support to every mosque in the US.\n\n__Join hundreds of members members making cards!__' };
```
deadlybutter: `aHelp`

resources/assets/components/Dashboard/index.js
```diff
@@ -4,39 +4,45 @@ import ShareContainer from '../../containers/ShareContainer';
...
+/**
+ * Render the dashboard.
+ * @param props
+ * @returns {XML}
+ * @constructor
+ */
 const Dashboard = (props) => {
   /**
    * Replace the given text with variables from the props.
+   * @TODO: This should not be defined in the render function.
```
deadlybutter: we could move it out & pass the props var but this seemed easier ¯\_(ツ)_/¯

resources/assets/components/Notification/index.js
```diff
@@ -1,34 +1,48 @@
...
   style: 'error',
 };
 
 const NotificationList = ({ notifications, removeNotification }) => (
-  <div className='notification-list'>
+  <div className="notification-list">
     {notifications.map(({ message, style }, index) => (
       <Notification
-        key={index}
+        key={makeHash(message)}
```
deadlybutter: Why did we change this to a hash?

resources/assets/components/ReportbackItem/index.js
```diff
@@ -39,27 +49,49 @@ const ReportbackItem = (props) => {
...
+    termId: React.PropTypes.string,
+    total: React.PropTypes.number,
+  }),
+  toggleReactionOff: React.PropTypes.func,
+  toggleReactionOn: React.PropTypes.func,
+  url: React.PropTypes.string,
+};
+
+ReportbackItem.defaultProps = {
+  id: undefined,
```
deadlybutter: Is there a reason we used undefined over null here? Id prefer to stay consistent

resources/assets/components/ActionPage/index.js
deadlybutter: this got marked as outdated for some reason but its still here

resources/assets/components/ContentPage/index.js
DFurnes: Fixed in 7553246!

resources/assets/components/Dashboard/index.js
DFurnes: It feels uncomfortable to have a function that relies on surrounding context like this... ideally we want functions to be pure `params => output` so that they're easier to test.

resources/assets/components/CallToActionBlock/index.js
DFurnes: Confused what you're suggesting here. If a prop type isn't required, we should have a default.

deadlybutter: This is under `defaultMetadata`!

resources/assets/components/Notification/index.js
weerd: So ESLint throws an error if we use the array index as a key. It's not a good idea to do so, and thus it yells at you for it. I wasn't 100% sure on what else to use, which is why I used the `makeHash()` helper that I specifically created for not needing to use an array index as a key. It'll return a random number string based on the content, so it's the same each time, regardless of where in the array the item might fall, or if the array is reordered.

resources/assets/components/CallToActionBlock/index.js
DFurnes: Oh doy, those are frustratingly similar names. 😖 Fixed in 109b0e8.

resources/assets/components/Notification/index.js
deadlybutter: @weerd Ah I see, it looks like React updated their docs as well, this used to have `index` as the default Key example https://facebook.github.io/react/docs/lists-and-keys.html

resources/assets/components/ActionPage/index.js
deadlybutter: @DFurnes You forgot this one as well in https://github.com/DoSomething/phoenix-next/commit/109b0e81901ed04a46a6b686e0e34c1fe9c17b28

DFurnes: Fixed in 4d7afe1.

resources/assets/components/ReportbackItem/index.js
weerd: Yeah, this one was a little weird. Something with how the RB items get rendered or checked while LazyLoading. I was getting errors if it was null, but using `undefined` seemed to work. Maybe something to look into...

PR #220, opened 2017-04-21
Changing session length

deadlybutter:
We discussed earlier with #data making this 30 minutes, almost forgot!


DFurnes:
I'm a little confused what this is used for, would love to chat it over at some point. But totally fine to merge! :smile:

PR #222, opened 2017-04-25
Initial work on hooking up Phoeni-Next with Gladiator.

weerd:
This PR adds a bunch of the setup to get Phoenix Next™ (PN) talking with Gladiator™.

It adds a couple new routes for PN to be able to request user info from Gladiator and to store a user in Gladiator if they signup for a contest. Since Gladiator needs to be aware of not only the legacy campaign id, but also the legacy campaign run id, this PR also adds a helper method to make it possible to easily retrieve the legacy campaign run id. The run id is also passed to the Redux store so it can be used with fetch requests.

image

_Note:_ I haven't fully tested the API calls to Gladiator because that's waiting on some updates to Gladiator for that to work well 😬


deadlybutter:
Looks good to me! 🌮

app/Services/Gladiator.php
```diff
@@ -0,0 +1,69 @@
...
+     * Get an index of (optionally filtered) users from Gladiator.
+     *
+     * @param  array  $query
+     * @return array - JSON response
+     */
+    public function getAllUsers(array $query = [])
+    {
+        $path = 'v1/users';
+
+        if (isset($query['current_user'])) {
```
DFurnes: Hmm, I wonder if we would ever want user traffic to be able to specify ID here.

Could we always force `auth()->id()`?

weerd: Not exactly sure what you mean... 🤔 It's only forcing `auth()->id()` if we want to pass along the current user, otherwise it's technically just the index endpoint. Mind clarifying :\

DFurnes: I might be forgetting some functionality, which is why I asked! I'm just trying to figure out if we would want users to be able to see an index for anything other than their own records... my hunch is that we wouldn't.

weerd: Yeah, not necessarily for now, but could be cool if in the future we could show who's in a contest with a particular user (esp if in the future we have a way to know if they're a friend) 😄  Also just felt weird to restrict it when it's an index list.

PR #223, opened 2017-04-25
Adds competition block

deadlybutter:
Apparently, I changed like 20 files, _SORRY_...

competition_prompt

This PR adds the frontend for the competition block. I left the actions fairly empty, but once we hook them up to the api everything should work!


app/Entities/Campaign.php
```diff
@@ -44,6 +44,8 @@ public function jsonSerialize()
                 'displayOptions' => $step->displayOptions,
                 'background' => $step->background ? $step->background->getFile()->getUrl() : '',
                 'photos' => $photos,
+                'customType' => $step->customType,
```
weerd: @deadlybutter curious, but feels weird to be adding a checkbox for the sake of differentiating this step block as slightly different from the other step blocks... Since the multiple ref fields can have a any content model (or a specified subset) was there a reason we couldn't use the more generic custom block and just allow the 'Action Steps' multiple reference to reference either action steps or custom blocks? Feels like it'd be a bit more versatile than having to create other steps and each one have a `customType`. Thoughts?  cc: @DFurnes

deadlybutter: Yeah I was trying to avoid polluting different types atm cause I wasn't sure what kind of side effects that would bring

weerd: yeah... 🤔 It also feels like there's a bunch on unnecessary fields for the competition prompt y using the action steps content type.

weerd: :man_shrugging:

weerd: Probs fine for now, but maybe a conversation we wanna have. If we end up putting other items as "steps" but require different fields, we would end up having to add extra fields to the action step content type versus just adding different block type to the list of steps referenced... Just something to think about 😉

DFurnes: Yeah, I'd really love if we could just use "custom blocks" here as well. Doesn't need to block merging, but maybe something to look into cleaning up later this week or early next week.

resources/assets/components/ActionPage/index.js
```diff
@@ -47,30 +48,53 @@ const renderPhoto = (photo, index) => (
...
+    const key = makeHash(title);
+    const type = step.customType[0];
+    const background = step.background;
+    const stepWidth = step.displayOptions[0];
+    const photoWidth = stepWidth === 'full' ? 'full' : 'one-third';
+    const shouldTruncate = step.truncate;
+    const additionalContent = step.additionalContent;
+
+    // Handle custom steps
+    // @TODO: I think it would make sense to handle the Reportback Uploader here as well?
```
DFurnes: Yeah! I think that's a great idea.

resources/assets/components/ActionPage/index.js
```diff
@@ -47,30 +48,53 @@ const renderPhoto = (photo, index) => (
...
+    if (type === 'competition') {
```
DFurnes: I'd love to handle this the [same way we handle blocks in the feed](https://github.com/DoSomething/phoenix-next/blob/3ca51ff4ca76892d6c1dce8bdae9b2c3e2898184/resources/assets/components/Feed/index.js#L22-L31).

resources/assets/components/ActionPage/index.js
```diff
@@ -81,17 +105,22 @@ const renderStep = (step, index) => {
...
-    actionSteps = actionSteps.slice(0, 2);
+    // Truncate steps if user isn't signed up & remove any custom steps.
+    actionSteps = actionSteps.filter(step => ! step.customType[0]).slice(0, 2);
 
     if (actionSteps[actionSteps.length - 1]) {
       actionSteps[actionSteps.length - 1].truncate = true;
     }
+  } else if (! showCompetition) {
+    // Filter out any steps that have a competition type.
+    actionSteps = actionSteps.filter(step => step.customType[0] !== 'competition');
```
DFurnes: This feels like something the competition block should be able to handle for itself, so that we don't end up with a ton of special cases in the `ActionPage` component for each of the possible types of steps.

resources/assets/components/CampaignUpdateBlock/index.js
```diff
@@ -8,7 +8,7 @@ import Embed from '../Embed';
 import DEFAULT_AVATAR from './default-avatar.png';
 import './campaign-update.scss';
 
-const Byline = ({ author, jobTitle, avatar }) => (
+export const Byline = ({ author, jobTitle, avatar }) => (
```
DFurnes: We should export this to a standalone component in `./components/Byline/index.js` now that we're using it elsewhere. Not necessary for merging this, but maybe another small todo for later this week.

resources/assets/components/CompetitionBlock/competitionBlock.scss
```diff
@@ -0,0 +1,32 @@
+@import '../../scss/next-toolbox';
+
+.competition-block {
+  flex-direction: column-reverse;
```
DFurnes: Is this for re-arranging the photo on mobile? If so, could we use a `<Figure align="right">` here?

resources/assets/components/CompetitionBlock/competitionBlock.scss
```diff
@@ -0,0 +1,32 @@
...
+
+  @include media($medium) {
+    flex-direction: row;
+  }
+
+  .competition-block__content {
+    &.is-confirmation > .markdown {
+      margin-bottom: $base-spacing;
+
+      h1, p {
```
DFurnes: Not something that needs to be fixed for this PR, but this is getting quite deeply nested. Ideally we could just have a `.success-color` helper class so we don't need our styles to be so tightly bound to the component structure.

resources/assets/components/CompetitionBlock/index.js
```diff
@@ -0,0 +1,57 @@
...
+import LazyImage from '../LazyImage';
+import './competitionBlock.scss';
+
+const CompetitionBlock = (props) => {
+  const { content, photo, byline, joinCompetition, hasJoinedCompetition,
+    hasPendingJoin, showConfirmation, campaignId } = props;
+
+  if (! showConfirmation && hasJoinedCompetition) return null;
+
+  const confirmation = '# You\'re signed up!\nYou should recieve an email shortly with more instructions. You can keep working on the campaign for now. I\'m so excited to have you onboard!';
```
DFurnes: Would a template string work here? If so, might be more readable to split onto a few lines.

app/Entities/Campaign.php
deadlybutter: sure thing!

resources/assets/components/Markdown/markdown.scss
```diff
@@ -1,6 +1,10 @@
 @import '../../scss/next-toolbox.scss';
 
 .markdown {
+  h1, h2, h3, p {
```
DFurnes: Is this a specific rule for Markdown text headings, or something we want for all headings in the future?

resources/assets/components/ActionPage/index.js
deadlybutter: It actually does handle this on it's own, so this code is redundant!

resources/assets/components/CampaignUpdateBlock/index.js
deadlybutter: I can add a TODO for that

resources/assets/components/CompetitionBlock/competitionBlock.scss
deadlybutter: I was originally using Figures for this but was having alignment issues, I might have been doing it wrong?

resources/assets/components/CompetitionBlock/competitionBlock.scss
deadlybutter: Yeah I didn't feel great about this either, helper class seems fine to me

resources/assets/components/CompetitionBlock/index.js
deadlybutter: I got yelled at by the linter for using a template string here

resources/assets/components/Markdown/markdown.scss
deadlybutter: Currently Forge is making purple headings, I think generally yeah we want this rule applied across all headings

resources/assets/actions/competition.js
```diff
@@ -0,0 +1,33 @@
...
+ * Action Creators: these functions create actions, which describe changes
+ * to the state tree (either as a result of application logic or user input).
+ */
+
+// Action: join the competition for the given campaign id
+export function joinCompetition(campaignId) {
+  return (dispatch, getState) => {
+    dispatch({ type: COMPETITION_PENDING });
+
+    // TODO: Obviously, replace this with an api call.
```
weerd: @deadlybutter for future ref, can you use add the `@` to `@TODO` so it's easier to search for these. I know this is temporary once the API stuff is ready 😄

deadlybutter: Agh ive been trying to, will fix!

resources/assets/components/CompetitionBlock/index.js
weerd: For using a template string or for it being too long? what was the eslint error? curious if it's something we might want to override or not...

deadlybutter: I think it had to do with the fact it's not actually a template string (in that it didn't have a variable in it)

weerd: Ah, gotcha. ok.

DFurnes: @joe You should be able to use a template string here... the linter will only complain if you use one unnecessarily (where a normal `'` would have worked just as well).

```js
const confirmation = `
# You're signed up!
You should receive an email shortly with more instructions. You can keep
working on the campaign for now. I'm so excited to have you onboard!`;
```

resources/assets/components/CompetitionBlock/competitionBlock.scss
DFurnes: Okie dokie, doesn't block merging but it'd be nice to figure out what's going on.

deadlybutter:
@DFurnes @weerd Updated a bunch based on feedback!

- [Changed](https://github.com/DoSomething/phoenix-next/pull/223/files#diff-81db49fe9a082ccf5c4c3208fee17d56R99) the renderStep function to use a switch instead of the if block. @DFurnes I decided to not use the `get` function from Lodash because 1) different components didn't need all of the props 2) the `stepIndex` logic needed to run somewhere. I think part of the action page refactoring could be simplyfing all of this.
- Due to ^, I [refactored](https://github.com/DoSomething/phoenix-next/pull/223/files#diff-81db49fe9a082ccf5c4c3208fee17d56R44) the ActionStep into a component. The file is starting to get big, another part of the refactoring might be worth exploring multiple files within this directory?
- [Added]https://github.com/DoSomething/phoenix-next/pull/223/files#diff-bd4b1f6479cef8a0b8411a9a15ec04eeR11) the `TODO` for the byline
- @weerd [added](https://github.com/DoSomething/phoenix-next/pull/223/commits/3ea0ba3dc684ff2ba295d940b40f906cb07a5070#diff-beffefc64fde686a8120427a971eb89aR17) the `@` 😉 
- [Enforced the default color](https://github.com/DoSomething/phoenix-next/pull/223/files#diff-96426e7dde62551c26e4b19018087095R7) to be black, and [added](https://github.com/DoSomething/phoenix-next/pull/223/files#diff-98429901da85f2a38e51b2aab685ee17R9) the success helper

resources/assets/components/CompetitionBlock/index.js
```diff
@@ -9,13 +9,18 @@ import { Byline } from '../CampaignUpdateBlock';
 import LazyImage from '../LazyImage';
 import './competitionBlock.scss';
 
+const DEFAULT_CONFIRMATION = `
+# You're signed up!
+
+You should recieve an email shortly with more instructions. You can keep working on the campaign for now. I'm so excited to have you onboard!
```
DFurnes: I know this is nitpicky, but why not wrap this to multiple lines for better readability?

resources/assets/components/Markdown/markdown.scss
```diff
@@ -1,8 +1,13 @@
 @import '../../scss/next-toolbox.scss';
 
 .markdown {
+
   h1, h2, h3, p {
-    color: $black;
+    color: inherit;
```
DFurnes: Is `inherit` not the default value?

resources/assets/scss/base.scss
```diff
@@ -3,4 +3,8 @@
 body, html {
   font-family: $primary-font-family;
   line-height: 1.4;
+
+  h1, h2, h3, p {
```
DFurnes: This doesn't need to be nested under the `body, html`, and makes a pretty horrendous selector if it is:

```
body h1, body h2, body h3, body p, html h1, html h2, html h3, html p {
  color: #222;
}
```

DFurnes: Nesting is the root of all evil in SCSS. Best to avoid unless you _really_ know you need it. 👹

resources/assets/components/Markdown/markdown.scss
deadlybutter: I believe that would be `initial`, which breaks the success helper

deadlybutter:
💥  i think its good now

deadlybutter:
@DFurnes Mind taking a look if you have a sec?

deadlybutter:
🚂 🚋 🚋

MERGE TRAIN COMIN THRU

Issue #188, opened 2017-04-14
Unexpected CSRF token error on uploading photos.


DFurnes:
#### Current Behavior
Submitting a photo on [Sincerely Us](https://www.dosomething.org/us/campaigns/sincerely-us/action) intermittently results in a 500 error.

```json
{"error":{"code":500,"message":"Looks like something went wrong. We've noted the problem and will try to get it fixed!"}}
```

I wasn't [able to find anything in the application logs](https://cloud.githubusercontent.com/assets/583202/25045227/362d45a4-20f9-11e7-98b8-ae0588668bc2.png), but temporarily enabling the  `APP_DEBUG` flag to display more detailed errors within JSON responses gave this location from the compiled source:

```json
{"error":{"code":500,"message":""},"debug":{"file":"\/var\/www\/next.dosomething.org\/releases\/20170413213115\/bootstrap\/cache\/compiled.php","line":3314}}
```

This looks like it corresponds to a TokenMismatchException:

```php
    public function handle($request, Closure $next)
    {
        if ($this->isReading($request) || $this->runningUnitTests() || $this->shouldPassThrough($request) || $this->tokensMatch($request)) {
            return $this->addCookieToResponse($request, $next($request));
        }
        throw new TokenMismatchException(); // <--- line 3314
    }
    protected function shouldPassThrough($request)
```

#### Desired Behavior
A user should be able to upload photos on Phoenix Next, while also being protected from CSRF attacks.

#### Steps to Reproduce
I haven't fully tracked it down but it seems like tokens are being expired too early...?

#### Why This Matters
We want everyone to be able to upload their photos without problems!

DFurnes:
From trying to reproduce again this morning, it seemed like I was able to successfully upload a _ton_ of photos, but as soon as I refreshed the page every upload failed (as described above).

DFurnes:
We temporarily removed the CSRF middleware from the reportback & signup routes in #191. This isn't ideal, but there's nothing super dangerous that can happen in the meantime.

DFurnes:
Let's try re-enabling this in #226 now that we haven't gotten these errors in a while.

PR #226, opened 2017-04-26
Re-enable CSRF middleware on all routes.

DFurnes:
We haven't had any login CSRF token errors since we fixed the session issue in #207... let's try re-enabling this for those two routes. ✌️ 🔀

References #188.

### DIFF
--- a/resources/assets/normalizers.js
+++ b/resources/assets/normalizers.js
@@ -43,8 +43,8 @@ export function normalizeReportbackItemResponse(data) {
   const reportbacks = {};
   const reportbackItems = {};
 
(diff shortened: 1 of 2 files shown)


resources/assets/normalizers.js
```diff
@@ -43,8 +43,8 @@ export function normalizeReportbackItemResponse(data) {
   const reportbacks = {};
   const reportbackItems = {};
 
-  data.forEach((reprotbackItemRecord) => {
-    const reportbackItem = reprotbackItemRecord;
+  data.forEach((reportbackItemRecord) => {
```
deadlybutter: it took me way to long to see what changed here 😅

PR #227, opened 2017-04-26
Adding some tests for helpers.

weerd:
This PR adds some quick tests for a few of the helper methods to play around with adding more Jest tests!

### DIFF
--- a/resources/assets/helpers/test.js
+++ b/resources/assets/helpers/test.js
@@ -5,7 +5,19 @@ import {
 } from './index';
 
 // Test contentfulImageUrl()
-test('generate contentful image url with added parameters', () => {
+test('generate contentful image url with defaults', () => {
+  const contentfulImage = contentfulImageUrl('//images.contentful.com/somecrazystring.jpg');
+
+  expect(contentfulImage).toBe('//images.contentful.com/somecrazystring.jpg');
+});
+
+test('generate contentful image url with specified fit', () => {
+  const contentfulImage = contentfulImageUrl('//images.contentful.com/somecrazystring.jpg', undefined, undefined, 'fill');
+
+  expect(contentfulImage).toBe('//images.contentful.com/somecrazystring.jpg?fit=fill');
+});
+
+test('generate contentful image url with all specified parameters', () => {
   const contentfulImage = contentfulImageUrl('//images.contentful.com/somecrazystring.jpg', 800, 600, 'fill');
 
   expect(contentfulImage).toBe('//images.contentful.com/somecrazystring.jpg?w=800&h=600&fit=fill');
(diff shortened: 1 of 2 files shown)


resources/assets/helpers/test.js
```diff
@@ -0,0 +1,26 @@
...
+// Test contentfulImageUrl()
+test('generate contentful image url with added parameters', () => {
+  const contentfulImage = contentfulImageUrl('//images.contentful.com/somecrazystring.jpg', 800, 600, 'fill');
+
+  expect(contentfulImage).toBe('//images.contentful.com/somecrazystring.jpg?w=800&h=600&fit=fill');
+});
+
+// Test makeHash()
+test('makes hash from a string', () => {
+  expect(makeHash('some arbitrary string here')).toBe(988941499);
```
DFurnes: It could be nice to decouple this a bit from the actual hash implementation – like "hashing the same string gives the same result", "hashing an integer works", "hashing null errors", etc.

resources/assets/helpers/test.js
```diff
@@ -0,0 +1,26 @@
+import {
+  makeHash,
+  modifiers,
+  contentfulImageUrl,
+} from './index';
+
+// Test contentfulImageUrl()
+test('generate contentful image url with added parameters', () => {
+  const contentfulImage = contentfulImageUrl('//images.contentful.com/somecrazystring.jpg', 800, 600, 'fill');
```
DFurnes: Should we add some assertions here for other combinations of options (like omitting a size or fit)?

resources/assets/helpers/test.js
```diff
@@ -0,0 +1,26 @@
...
+});
+
+// Test makeHash()
+test('makes hash from a string', () => {
+  expect(makeHash('some arbitrary string here')).toBe(988941499);
+});
+
+// Test modifiers()
+test('prefix a class name', () => {
+  expect(modifiers('danger')).toEqual(expect.arrayContaining(['-danger']));
```
DFurnes: Let's add an assertion that it skips over values like `null` or `undefined`, here (especially since I think it might fail w/ the current implementation!)

resources/assets/helpers/test.js
weerd: Yeah, I was thinking about that. Need to review and see how Jest recommends doing that... whether just adding it as another item right below or whatnot.

resources/assets/helpers/test.js
weerd: Ah, yeah. Good call.

weerd: Works fine :) I had updated the `modifiers()` method to handle `undefined || null` since it was actually something I encountered while working on a component!

DFurnes: Awesome! 👏

PR #232, opened 2017-04-27
Update Contentful dependencies and fix some fatal errors.

DFurnes:
#### Changes
This pull request updates our Contentful dependencies to their official 1.0 releases (woooo), and also fixes some scenarios where an empty image field would cause a huge fatal error. Yowza.

📷 💥 🚳


app/helpers.php
```diff
@@ -125,13 +136,11 @@ function get_image_url($asset, $crop = 'landscape')
         ->setHeight(50)
         ->setResizeFit('scale');
 
-    if (! array_key_exists($crop, $options)) {
-        throw new \InvalidArgumentException('The specified cover image type of '.$crop.' is not available.');
-    }
-
-    $locale = app()->getLocale();
+    // Provide cropped image if specified, or just the original image by default.
+    $crop = isset($options[$style]) ? $options[$style] : null;
```
weerd: Is there any worry with passing an invalid or crop style or just a typo in it?

weerd: Oh wait, this does determine that as well, just doesn't throw an error for it. ok!

DFurnes: Yeah, I wasn't really sure the best way to handle this. Maybe if it's not null we do the strict checking that we'd done before... either way, we'll want a way to make an "unstyled" URL since we do it client-side in a few places.

weerd: Ya... :man_shrugging: I think this is fine :)

PR #233, opened 2017-04-28
Updating affirmation

deadlybutter:
This PR updates the affirmation component to connect with contentful data & does some general improvements / cleanup to the component!

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -69,6 +69,14 @@ class Campaign extends Entity implements JsonSerializable
             'activityFeed' => $this->activity_feed,
             'actionSteps' => $actionSteps,
             'dashboard' => $this->dashboard,
+            'affirmation' => [
+                'header' => $this->affirmation->header,
+                'photo' => get_image_url($this->affirmation->photo),
+                'quote' => $this->affirmation->quote,
+                'author' => $this->affirmation->author,
+                'callToActionHeader' => $this->affirmation->callToActionHeader,
+                'callToActionDescription'=> $this->affirmation->callToActionDescription,
+            ],
             'pages' => $this->pages,
             'additionalContent' => $this->additionalContent,
         ];
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -71,7 +71,7 @@ class Campaign extends Entity implements JsonSerializable
             'dashboard' => $this->dashboard,
             'affirmation' => [
                 'header' => $this->affirmation->header,
-                'photo' => get_image_url($this->affirmation->photo),
+                'photo' => get_image_url($this->affirmation->photo, 'square'),
                 'quote' => $this->affirmation->quote,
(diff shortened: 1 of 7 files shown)


deadlybutter:
Note: This only works on Teens for Jeans atm, need to add affirmation content to Sincerely us (doing now)

weerd:
> Note: This only works on Teens for Jeans atm, need to add affirmation content to Sincerely us (doing now)

If Affirmation content isn't added does it fail?

deadlybutter:
@weerd thats the case with most of our stuff, eg the campaign update bug that happened in slack earlier. i was just putting it there as a further disclaimer for people looking at the review app (The content is live now tho)

app/Entities/Campaign.php
```diff
@@ -69,6 +69,14 @@ public function jsonSerialize()
             'activityFeed' => $this->activity_feed,
             'actionSteps' => $actionSteps,
             'dashboard' => $this->dashboard,
+            'affirmation' => [
```
DFurnes: Should the Affirmation field be required in Contentful?

app/Entities/Campaign.php
```diff
@@ -69,6 +69,14 @@ public function jsonSerialize()
...
+                'header' => $this->affirmation->header,
+                'photo' => get_image_url($this->affirmation->photo),
```
DFurnes: Should we pick a crop here? If not, we should do it on the client-side.

resources/assets/reducers/signups.js
```diff
@@ -33,7 +33,7 @@ const signupReducer = (state = {}, action) => {
         isPending: false,
         thisSession: true,
         thisCampaign: true,
-        showAffirmation: true,
+        shouldShowAffirmation: true,
```
DFurnes: This makes much more sense. 👌

resources/assets/scss/base.scss
```diff
@@ -8,3 +8,9 @@ body, html {
 h1, h2, h3, p {
   color: $black;
 }
+
+.-inverted {
```
DFurnes: This isn't a modifier for anything is it? If not, it should just be `.inverted`.

app/Entities/Campaign.php
deadlybutter: Good call! It should technically be a square

app/Entities/Campaign.php
deadlybutter: Yeah I can make it required

resources/assets/scss/base.scss
deadlybutter: 👍

deadlybutter:
@DFurnes Marked the field as required & updated the PR, everything still seems to look good 👍

ngjo:
[As a developer, I want to add the Affirmation Component to Contentful](https://trello.com/c/paYlRk97/602-as-a-developer-i-want-to-add-the-affirmation-component-to-contentful)

PR #234, opened 2017-04-28
Allow setting custom nouns for the impact item.

### LINKED ISSUE #230: Editing Text in the Reportback Uploader
Right now, all of the text in the RB uploader are hardcoded: 

screen shot 2017-04-27 at 3 23 15 pm

The line "How many cards are in this photo?" will change campaign to campaign (the noun). Two campaigns launching on May 1 will need this changed for them.

Mirror Messages: How many notes are in this photo?
Uncover the Truth: How many signed pledges are in this photo?

It's not exactly following the "noun" "verb" format that we've used on Phoenix-ashes which was "Total # of nouns verbed" (https://www.dosomething.org/us/node/7/active)

DFurnes:
#### Changes
This pull request allows campaign leads to customize the "impact noun" (for example, "cards" or "jeans") using the Additional Content JSON field in Contentful. If not set, it will default to "items".

Check it out on [Mirror Messages](https://phoenix-next-stage-pr-234.herokuapp.com/us/campaigns/mirror-messages). Closes #230.

### DIFF
--- a/resources/assets/components/ReportbackItem/index.js
+++ b/resources/assets/components/ReportbackItem/index.js
@@ -16,17 +16,8 @@ function getMetadataFromProps(props) {
 }
 
 const ReportbackItem = (props) => {
-  const {
-    id,
-    url,
-    quantity,
-    caption,
-    firstName,
-    reaction = null,
-    isFetching = false,
-    toggleReactionOn,
-    toggleReactionOff,
-  } = props;
+  const { id, url, quantity, noun, caption, firstName, reaction = null,
+    isFetching = false, toggleReactionOn, toggleReactionOff } = props;
 
   const metadata = mergeMetadata(ReportbackItem.defaultMetadata, getMetadataFromProps(props));
 
@@ -50,12 +41,11 @@ const ReportbackItem = (props) => {
     );
   }
 
-  // TODO: Don't hardcode cards
   return (
     <Figure className="reportback-item" image={url} alt={`${firstName}'s photo`}>
       <BaseFigure media={reactionElement} alignment="right" className="padded">
         {firstName ? <h4>{firstName}</h4> : null }
-        {quantity ? <p className="footnote">{quantity} cards</p> : null }
+        {quantity ? <p className="footnote">{quantity} {noun}</p> : null }
         {caption ? <p>{caption}</p> : null }
       </BaseFigure>
     </Figure>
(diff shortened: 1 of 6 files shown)


weerd:
@DFurnes the link in the description for "Mirror Messages" goes to this same PR 😬

DFurnes:
@weerd Too quick! Waiting on the review branch to be created so I can link. 😉

resources/assets/components/ReportbackItem/index.js
```diff
@@ -50,12 +41,11 @@ const ReportbackItem = (props) => {
     );
   }
 
-  // TODO: Don't hardcode cards
   return (
     <Figure className="reportback-item" image={url} alt={`${firstName}'s photo`}>
       <BaseFigure media={reactionElement} alignment="right" className="padded">
         {firstName ? <h4>{firstName}</h4> : null }
-        {quantity ? <p className="footnote">{quantity} cards</p> : null }
+        {quantity ? <p className="footnote">{quantity} {noun}</p> : null }
```
weerd: Not needed for merge but could be something we update to check if if the `quantity` is `1` then use `noun.singular` otherwise use `noun.plural`?

DFurnes: Oh doy. Yeah, somehow that didn't even register as a possibility to me.

weerd: But then we'd want to pass the entire `noun`object in the container and not just the `noun.plural` :)

resources/assets/components/ReportbackItem/index.js
```diff
@@ -88,6 +79,7 @@ ReportbackItem.defaultProps = {
   caption: undefined,
   firstName: undefined,
   isFetching: false,
+  noun: 'items',
```
weerd: If we roll with passing `noun` containing `singular` and `plural` versions, then maybe this default should be something like:
```
noun: {
  singular: 'item',
  plural: 'items',
}
```
thoughts?

resources/assets/components/ReportbackItem/index.js
DFurnes: Updated in 1a40058!

resources/assets/components/ReportbackItem/index.js
DFurnes: Yeah, this makes sense to me! 👍

resources/assets/helpers/index.js
```diff
@@ -91,6 +91,17 @@ export function modifiers(...names) {
 }
 
 /**
+ * Pluralize a noun.
+ *
+ * @param {Number} quantity
+ * @param {String} singular
+ * @param {String} plural
+ */
+export function pluralize(quantity, singular, plural) {
```
weerd: very helpful!

PR #236, opened 2017-04-28
Connecting competition block to gladiator & some other fixes

deadlybutter:
- Connected the competition actions to gladiator
- Fixed ActionTypes props for when a competition is present (We handle Step proptypes elsewhere now)
- Updated competition block logic to account for the various states
- Fixed a bug with the notification component throwing an error when trying to hash a null key

### DIFF
--- a/resources/assets/actions/competition.js
+++ b/resources/assets/actions/competition.js
@@ -24,7 +24,6 @@ export function joinCompetition(campaignId, campaignRunId) {
       if (! response) throw new Error('competition signup failed');
       if (response.data) dispatch({ type: JOINED_COMPETITION, campaignId, userId });
     }).catch((err) => {
-      console.error(err);
       dispatch(addNotification('error'));
     });
   };
@@ -42,7 +41,6 @@ export function checkForCompetition(campaignId, campaignRunId) {
       if (! response) throw new Error('competition get failed');
       if (response.data) dispatch({ type: COMPETITION_FOUND, campaignId, userId });
     }).catch((err) => {
-      console.error(err);
       dispatch(addNotification('error'));
     });
   };
--- a/resources/assets/actions/competition.js
+++ b/resources/assets/actions/competition.js
@@ -23,7 +23,7 @@ export function joinCompetition(campaignId, campaignRunId) {
     }).then((response) => {
       if (! response) throw new Error('competition signup failed');
       if (response.data) dispatch({ type: JOINED_COMPETITION, campaignId, userId });
-    }).catch((err) => {
+    }).catch(() => {
       dispatch(addNotification('error'));
     });
   };
@@ -34,13 +34,13 @@ export function checkForCompetition(campaignId, campaignRunId) {
   return (dispatch, getState) => {
     const userId = getState().user.id;
 
-    (new Phoenix()).post('next/contests/users', {
-      legacyCampaignId: campaignId,
-      legacyCampaignRunId: campaignRunId,
+    (new Phoenix()).get('next/contests/users', {
+      campaign_id: campaignId,
+      campaign_run_id: campaignRunId,
     }).then((response) => {
       if (! response) throw new Error('competition get failed');
       if (response.data) dispatch({ type: COMPETITION_FOUND, campaignId, userId });
-    }).catch((err) => {
+    }).catch(() => {
       dispatch(addNotification('error'));
     });
   };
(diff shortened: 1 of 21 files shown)


resources/assets/components/ActionPage/index.js
```diff
@@ -172,12 +172,7 @@ const ActionPage = (props) => {
 };
 
 ActionPage.propTypes = {
-  steps: PropTypes.arrayOf(PropTypes.shape({
-    title: PropTypes.string.isRequired,
-    content: PropTypes.string.isRequired,
-    background: PropTypes.string.isRequired,
-    photos: PropTypes.array,
-  })),
+  steps: PropTypes.array.isRequired, // eslint-disable-line react/forbid-prop-types
```
DFurnes: This doesn't need to block merging, but I think this is a sign we're sending too arbitrary of a prop here.

resources/assets/actions/competition.js
```diff
@@ -1,33 +1,47 @@
...
-    // @TODO: Obviously, replace this with an api call.
-    setTimeout(() => {
-      const userId = getState().user.id;
-      dispatch({ type: JOINED_COMPETITION, campaignId, userId });
-    }, 2000);
+    (new Phoenix()).post('next/contests/users', {
+      legacyCampaignId: campaignId,
+      legacyCampaignRunId: campaignRunId,
+    }).then((response) => {
+      if (! response) throw new Error('competition signup failed');
```
weerd: Not needed for this merge but we could update to dispatch an action if contest signup failed and handle it somehow :)

resources/assets/actions/competition.js
```diff
@@ -1,33 +1,47 @@
...
-// Action: check if the user joined the competition for the given campaign id
-export function checkForCompetition(campaignId) {
+// Action: check if the user joined the competition for the given campaign & run id
+export function checkForCompetition(campaignId, campaignRunId) {
   return (dispatch, getState) => {
     const userId = getState().user.id;
 
-    // @TODO: make api call
-    dispatch({ type: COMPETITION_FOUND, campaignId, userId });
+    (new Phoenix()).post('next/contests/users', {
```
weerd: This should be a `POST` request and also pass the `current_user=true` param.

resources/assets/actions/competition.js
deadlybutter: @weerd Look below in the catch!

resources/assets/actions/competition.js
deadlybutter: @weerd Thats why its returning the response for every new user i make... Thank you for spotting that

DFurnes: @weerd We removed `current_user=true` in #235!

resources/assets/actions/competition.js
weerd: Oh gotcha, this is then caught below. Maybe then pass the message from this to the notification?

resources/assets/actions/competition.js
weerd: oh yeah, duh! thanks @DFurnes I had a brain fart.

resources/assets/actions/competition.js
deadlybutter: If we're adding more messaging to the site id wanna run it by Jen, the default error msg is a good catch all for now i think

resources/assets/actions/competition.js
deadlybutter: So thats not it, back to the drawing board!

DFurnes: Sorry, @weerd's point about this needing to be a `GET` request is still true! Just commenting on the param.

deadlybutter: Agh - Bad copy paste, totally slipped by me there

deadlybutter:
Just went through the entire flow again, seems like all is well! I think this is good for merge

deadlybutter:
screen shot 2017-04-28 at 3 46 26 pm

double checking the waiting room end as well, seems good here

resources/assets/components/Notification/index.js
```diff
@@ -28,7 +27,7 @@ const NotificationList = ({ notifications, removeNotification }) => (
   <div className="notification-list">
     {notifications.map(({ message, style }, index) => (
       <Notification
-        key={makeHash(message)}
+        key={index} // eslint-disable-line react/no-array-index-key
```
weerd: we'll eventually need to figure out a better solution for keys to avoid using indexes :\

deadlybutter: Agreed, but wanted to get around the error in the meantime!

PR #237, opened 2017-04-28
Removes waiting rooms & moves campaign list to homepage.

DFurnes:
#### Changes
This will make testing pull request apps & unreleased campaigns a lil' easier. 🍃


DFurnes:
Yeah I'm not sure! I left it just to be safe, but I assume we've already messaged people if wanted to...?

PR #238, opened 2017-05-01
Update to allow for more custom CTAs on content pages.

weerd:
This PR makes an update to allow having slightly more customized CTAs on the content pages for a campaign. It tries to be smart enough to handle a few scenarios, but to customize it we need to pass a custom JSON object for the campaign via Contentful as follows:

```
{
  "tagline": "Help us send letters of support to every mosque in the United States.",
  "verb": {
    "singular": "make a",
    "plural": "make"
    },
  "noun": {
    "singular": "card",
    "plural": "cards"
  }
}
```

### DIFF
--- a/resources/assets/components/ContentPage/index.js
+++ b/resources/assets/components/ContentPage/index.js
@@ -5,12 +5,8 @@ import CallToActionContainer from '../../containers/CallToActionContainer';
 
 import './content-page.scss';
 
-const ContentPage = ({ pages, route }) => {
+const ContentPage = ({ pages, route, tagline, noun, verb }) => {
   const page = pages.find(item => item.fields.slug === route.page);
-
-  // @TODO: temporary variables until these CTAs are no longer hardcoded.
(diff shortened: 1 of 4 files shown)


resources/assets/components/ContentPage/index.js
```diff
@@ -21,10 +17,10 @@ const ContentPage = ({ pages, route }) => {
         </article>
       </div>
       <div className="secondary">
-        <CallToActionContainer fields={ctaText1} />
+        <CallToActionContainer fields={{ content: `${tagline}\n\n__Join hundreds of members and ${verb.plural} ${noun.plural}!__` }} />
```
DFurnes: This feels sorta awkward to me. Can we pass in as a `content` prop instead?

weerd: Yeah, it's not ideal, but has to be in `field` cause on regular CTAs added in feed, it comes in via the fields prop.

PR #239, opened 2017-05-01
Default to a first name of Doer.

weerd:
This PR lets us default to a first name of **A Doer** if no first name is available when rendering out Reportback Items. It was a little tricky because in the submission gallery, we prefer not to display any name, even if a default is provided. Since both reportback item blocks in the feed and submission gallery use the same RB Item component, I added a new `basicDisplay` boolean prop that can be passed to make sure only a very basic display of the RB item is rendered.

Refs https://trello.com/c/ehcE3Vay/606-1-as-a-developer-i-want-the-user-name-to-default-doer-if-the-account-has-no-name

### DIFF
--- a/resources/assets/components/ReportbackItem/index.js
+++ b/resources/assets/components/ReportbackItem/index.js
@@ -17,10 +17,8 @@ function getMetadataFromProps(props) {
 }
 
 const ReportbackItem = (props) => {
-  const { id, url, quantity, noun, caption, reaction = null, isFetching = false,
+  const { id, url, quantity, noun, caption, firstName, reaction = null, isFetching = false,
     basicDisplay = false, toggleReactionOn, toggleReactionOff } = props;
-
-  const firstName = props.firstName || 'A Doer';
 
   const metadata = mergeMetadata(ReportbackItem.defaultMetadata, getMetadataFromProps(props));
 
@@ -85,7 +83,7 @@ ReportbackItem.defaultProps = {
   id: undefined,
   basicDisplay: false,
   caption: undefined,
-  firstName: undefined,
+  firstName: 'A Doer',
   isFetching: false,
   noun: {
     singular: 'item',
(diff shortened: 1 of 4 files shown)


deadlybutter:
Disclaimer - I can't verify the jest part, maybe @DFurnes wants to weigh in?

weerd:
> May the force be with you (Did I do that right? I haven't actually seen it)

ugh, @deadlybutter we need to remedy this.

resources/assets/components/ReportbackItem/index.js
```diff
@@ -17,8 +17,10 @@ function getMetadataFromProps(props) {
 }
 
 const ReportbackItem = (props) => {
-  const { id, url, quantity, noun, caption, firstName, reaction = null,
-    isFetching = false, toggleReactionOn, toggleReactionOff } = props;
+  const { id, url, quantity, noun, caption, reaction = null, isFetching = false,
+    basicDisplay = false, toggleReactionOn, toggleReactionOff } = props;
+
+  const firstName = props.firstName || 'A Doer';
```
DFurnes: Can we make this a default prop?

weerd: Oh true. I think that should work. Vestige from old approach I was trying...

DFurnes:
> Disclaimer - I can't verify the jest part, maybe @DFurnes wants to weigh in?

I'm still learning the best practices here like everyone else, but I think we might want to look into adding more assertion-style tests for important elements (e.g. "expect this element to exist", "expect this text to be rendered") since they indicate intent. Otherwise I could see us getting in the habit of just blindly accepting snapshot tests whenever they change.

Either way, definitely not something to block merging this PR.

weerd:
@DFurnes yeah, agreed. I'm still learning that stuff, but wanted to at least add _a_ test in the component to get the ball rolling, and then move further along with it.

DFurnes:
Totally! Ship it! ⛵️ 💨

PR #240, opened 2017-05-01
Multi competition fix

deadlybutter:
This pr

- Cleans up competitions reducer
- Fixes multi competition signup
- Makes photo not required for prompt

### DIFF
--- a/resources/assets/actions/competition.js
+++ b/resources/assets/actions/competition.js
@@ -39,7 +39,11 @@ export function checkForCompetition(campaignId, campaignRunId) {
       campaign_run_id: campaignRunId,
     }).then((response) => {
       if (! response) throw new Error('competition get failed');
-      if (response.data) dispatch({ type: COMPETITION_FOUND, campaignId, userId });
+
+      const joinedCompetition = response.data &&
+        (response.data.waitingRoom || response.data.competition);
+
+      if (joinedCompetition) dispatch({ type: COMPETITION_FOUND, campaignId, userId });
     }).catch(() => {
       dispatch(addNotification('error'));
     });
--- a/resources/assets/reducers/competitions.js
+++ b/resources/assets/reducers/competitions.js
@@ -13,12 +13,11 @@ import {
  * Competitions reducer:
  */
 const competitions = (state = {}, action) => {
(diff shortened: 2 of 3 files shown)


resources/assets/actions/competition.js
```diff
@@ -39,7 +39,11 @@ export function checkForCompetition(campaignId, campaignRunId) {
       campaign_run_id: campaignRunId,
     }).then((response) => {
       if (! response) throw new Error('competition get failed');
-      if (response.data) dispatch({ type: COMPETITION_FOUND, campaignId, userId });
+
+      const joinedCompetition = response.data &&
+        (response.data.waitingRoom || response.data.competition);
```
weerd: Do we need the `||` here? I don't remember offhand but would `response.data` be an empty array if user not in contest? If so, can probs just check for `response.data` right?

weerd: I'm assuming it's why you had to change it, but I'm curious...

weerd: confirmed! yeah, that's a bug in Gladiator we'll need to eventually fix.

resources/assets/reducers/competitions.js
```diff
@@ -13,15 +13,10 @@ import {
  * Competitions reducer:
  */
 const competitions = (state = {}, action) => {
-  let joinedCompetitions = [];
+  const joinedCompetitions = [...(state.data || []), action.campaignId];
```
weerd: This reads a little weird...

weerd: If we end up with `COMPETITION_NOT_FOUND` or `COMPETITION_FAILED` action types to handle those scenarios, seems weird to have the above sit outside of the cases? Thoughts?

deadlybutter: Agreed but we're not there - I can revert this change & unblock the merge, since it was just a styling thing

deadlybutter: @weerd reverted this change back to normal, should be good to go

resources/assets/actions/competition.js
deadlybutter: just updating the PR here, seems like we're considering this intended functionality / not a bug

PR #241, opened 2017-05-02
Quick fix for the CTA color issue, and something when Thor is down.

weerd:
This PR makes a quick fix for the CTAs and also a small remedy to allow the page to render for local development when the Thor DS service is down.

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -155,6 +155,9 @@ function get_legacy_campaign_data($id, $key = null)
     try {
         $campaign = (new PhoenixLegacy)->getCampaign($id);
     } catch (\Exception $error) {
+        $handler = app(\App\Exceptions\Handler::class);
+        $handler->report($error);
+
         return null;
     }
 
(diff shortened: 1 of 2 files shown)


app/helpers.php
```diff
@@ -152,7 +152,11 @@ function get_image_url($asset, $style = null)
  */
 function get_legacy_campaign_data($id, $key = null)
 {
-    $campaign = (new PhoenixLegacy)->getCampaign($id);
+    try {
+        $campaign = (new PhoenixLegacy)->getCampaign($id);
```
DFurnes: Smort. We should make sure we still report errors to New Relic if this happens though.

resources/assets/components/CallToActionBlock/cta.scss
```diff
@@ -12,6 +12,10 @@
     background-color: transparent;
     border: 0 none;
     color: $off-black;
+
+    .cta__title {
+      color: $off-black;
```
DFurnes: tumblr_li9cks2toh1qzxu0ko1_500

deadlybutter: giphy 1

PR #242, opened 2017-05-03
Adding Laravel Debugbar.

weerd:
This PR activates the use of Laravel Debugbar during debug mode development! Will be very useful for debugging issues (for example there's a whole panel on session info which would have been ideal for the Chloe bug methinks!).

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
         "erusev/parsedown": "^1.6",
         "predis/predis": "^1.1",
         "fideloper/proxy": "^3.2",
-        "embed/embed": "^3.0"
+        "embed/embed": "^3.0",
+        "barryvdh/laravel-debugbar": "^2.3"
     },
     "require-dev": {
         "fzaninotto/faker": "~1.4",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -13,7 +13,8 @@
         "erusev/parsedown": "^1.6",
         "predis/predis": "^1.1",
         "fideloper/proxy": "^3.2",
-        "embed/embed": "^3.0"
+        "embed/embed": "^3.0",
+        "barryvdh/laravel-debugbar": "^2.3"
```
weerd: The instructions seemed to imply popping this in the `require` dep and not the `require-dev`. Thinking about moving it into `require-dev`. Thoughts?

DFurnes: I think that'd cause issues when trying to register the `Barryvdh\Debugbar\ServiceProvider` provider in production (since the production vs. debug check is done inside there).

weerd: Oh ya, good point.

weerd:
Yeah, I mostly wanted to play around with it and see how useful it is. If we deem it not that great, we can try something else. Might also help inform us whether we want to use it on other Laravel projects that do more sans-AJAX requests 😃

PR #243, opened 2017-05-04
Updates to the CampaignUpdateBlock and splitting out the Byline as a component.

weerd:
This PR fixes a bug where if the CampaignUpdate block on Contentful does not have the byline data added to the JSON object field, the page would break when trying to render out the component. I think that for the time being, in areas where we use the JSON Object field for supplying additional data for a component, we can not specify any of that data as being required in the React component and instead need to supply some defaults. Once we convert these components to their own content types on contentful and can have "required" fields, then we can update the React component to actually require that data since we'll have more confidence that the data should exists after retrieval from Contentful. 

There is also now a new `Byline` component that can be used outside of the CampaignUpdateBlock.

It also adds some _basic_ snapshot test to the `CampaignUpdateBlock` and new `Byline` components.


resources/assets/components/Byline/index.js
```diff
@@ -0,0 +1,28 @@
...
+  </Figure>
+);
+
+Byline.propTypes = {
+  author: PropTypes.string,
+  jobTitle: PropTypes.string,
+  avatar: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.object,
+  ]),
```
weerd: There was something odd happening when testing with Jest where the default avatar image is pulled in as an object and not as a data uri string, so I had to add the above prop type definition to account for both... Something to look into 🤔

Issue #248, opened 2017-05-09
Competitions Prompt not rendering on iOS mobile web


lkpttn:
# MAJOR BUG
On iOS mobile web (Safari and Chrome) the competition prompt is not rendering correctly and users can't read/join competitions. Not able to replicate on Android devices or desktop browsers with small windows. The title should also read "Go Above and Beyond", maybe indicative of a botched deploy?

**Desired Behavior**
The block should render appropriately and allow users to join competitions. Should display appropriate post-joining copy.

![screen shot 2017-05-09 at 4 23 29 pm](https://cloud.githubusercontent.com/assets/1865372/25870950/e59bae16-34d3-11e7-996f-db6498a67550.png)

**Steps to Replicate**
1 - Load the campaign on an iOS device
2- Join the campaign and move to the action tab
3 - The competitions prompt will not display correctly.

![image uploaded from ios](https://cloud.githubusercontent.com/assets/1865372/25870711/10f8daee-34d3-11e7-8c8e-eaf05b2186d7.jpg)

ngjo:
![](https://github.trello.services/images/mini-trello-icon.png) [UNPLANNED WORK: 🐛 Competition block isn't showing up on mobile 🐛](https://trello.com/c/Q7DzyS8E/632-unplanned-work-competition-block-isn-t-showing-up-on-mobile)

PR #249, opened 2017-05-09
Hot fixing the flex cell for competitions

deadlybutter:
See this issue for details of the bug: https://github.com/DoSomething/phoenix-next/issues/248

TLDR: Competitions block is shrunk on IOS

This PR is a shitty hotfix I found by playing around in the inspector. The actual flex cell value is `0 0 100%;` which seems to work fine everywhere else. @DFurnes & @weerd can you weigh in here with any thoughts / ideas?

### DIFF
--- a/resources/assets/components/CompetitionBlock/competitionBlock.scss
+++ b/resources/assets/components/CompetitionBlock/competitionBlock.scss
@@ -6,6 +6,10 @@
   @include media($medium) {
     flex-direction: row;
   }
+  
+  > .flex__cell {
+    flex: 0 0 auto;  
+  }
 
   .competition-block__content {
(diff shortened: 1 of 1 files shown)


DFurnes:
Build is passing after DoSomething/webpack-config#8. ✅

DFurnes:
I'm good with merging this as a quick fix. I also think this is a good reminder to ourselves not to use flexbox if we don't need it, since there's still plenty of weird lil' browser quirks that you don't run into with good old fashioned floats.

deadlybutter:
Just waiting to make sure it looks good on Heroku as well, will merge in a sec!

deadlybutter:
coolio seems good

PR #252, opened 2017-05-10
Update to fix up styles for the competition block.

weerd:
This PR make a few markup and style fixes for the competition block and removes the use of flexbox which was a bit of overkill for what the content holds. Updates also include responsive fixes.

image

### DIFF
--- a/resources/assets/components/CompetitionBlock/index.js
+++ b/resources/assets/components/CompetitionBlock/index.js
@@ -4,7 +4,6 @@ import classnames from 'classnames';
 
 import Markdown from '../Markdown';
 import Block, { BlockTitle } from '../Block';
-import { Flex, FlexCell } from '../Flex';
 import Byline from '../Byline';
 import LazyImage from '../LazyImage';
 import './competitionBlock.scss';
@@ -32,7 +31,7 @@ const CompetitionBlock = (props) => {
     checkForCompetition(campaignId, campaignRunId);
   }
 
-  const button = showConfirmation ? null : (
+  const renderButton = showConfirmation ? null : (
     <button
       disabled={hasPendingJoin}
       className={classnames('button', { 'is-loading': hasPendingJoin })}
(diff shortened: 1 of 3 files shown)


resources/assets/components/CompetitionBlock/index.js
```diff
@@ -32,31 +31,31 @@ const CompetitionBlock = (props) => {
...
-  const button = showConfirmation ? null : (
+  const renderButton = showConfirmation ? null : (
     <button
       disabled={hasPendingJoin}
       className={classnames('button', { 'is-loading': hasPendingJoin })}
       onClick={() => joinCompetition(campaignId, campaignRunId)}
     >join competition</button>
   );
 
+  const renderPhoto = photo && ! showConfirmation ? (
```
deadlybutter: Is the `! showConfirmation` check done intentionally?

resources/assets/components/CompetitionBlock/index.js
```diff
@@ -32,31 +31,31 @@ const CompetitionBlock = (props) => {
     checkForCompetition(campaignId, campaignRunId);
   }
 
-  const button = showConfirmation ? null : (
+  const renderButton = showConfirmation ? null : (
```
DFurnes: This naming feels weird to me, since it's a rendered "thing" rather than a function…

resources/assets/components/Markdown/markdown.scss
```diff
@@ -6,6 +6,18 @@
     color: inherit;
   }
 
+  h1 + p,
```
DFurnes: ➕ ➕ ➕

deadlybutter: Well, that's an interesting selector... I guess that makes sense

resources/assets/components/CompetitionBlock/index.js
weerd: Yeah, I was under the assumption if a photo is added we don't want it showing w/ the success message.

resources/assets/components/CompetitionBlock/index.js
weerd: oh true. I'll update `renderPhoto` as well since it's a similar situation.

resources/assets/components/Markdown/markdown.scss
weerd: lol

weerd: yep. targeting `<p>` tags that are siblings of a `<h1>` tags :)

PR #255, opened 2017-05-15
Sixpack powering up

weerd:
This PR adds a simple Sixpack A/B test to the campaign, switching out the some of the styling for the Competition block. This is mostly a test to enable a first test and confirm that Phoenix-Next-Thor is connecting properly to Sixpack-Thor.

This setup is not ideal since adding a test requires modifying too many different locations. I'm considering some alternatives for how best to structure A/B test so it will require modifying less files to execute on a test. However, these are the "pieces" needed to perform a full A/B test.


resources/assets/components/CompetitionBlock/competitionBlock.scss
```diff
@@ -28,3 +28,16 @@
...
+
+.block {
+  &.-colorful {
+    background-color: $purple;
+    border-radius: 4px;
+  }
+}
+
+.competition-block {
+  .-colorful & {
```
DFurnes: Absolutely not the point of this pull request, but this feels awkward – both modifying `.block` from the "competition block" file, and having to account for `.-colorful` modifier on both.

resources/assets/components/CompetitionBlock/index.js
```diff
@@ -19,6 +20,20 @@ const CompetitionBlock = (props) => {
   const { content, photo, byline, joinCompetition, hasJoinedCompetition,
     hasPendingJoin, showConfirmation, campaignId, campaignRunId, checkForCompetition } = props;
 
+  // @TEST: Sixpack Experiments
```
DFurnes: I like this! `@TEST:` 💡

resources/assets/components/CompetitionBlock/competitionBlock.scss
weerd: Yeah, this is by no means ideal. It's what I mention in the description with "This setup is not ideal..." 😉 I just wanted to get something up to test Sixpack on Thor and confirm it's working. Definitely need to spend some time on considering a better approach overall.

DFurnes: I know, I know! Definitely not something to deal with in this PR. 😄

PR #256, opened 2017-05-16
Adding experiments to analytics data

deadlybutter:
This pr adds experiments to the analytics data we send off to Keen

_Why are you transforming the data this way?_

From a Keen.IO perspective, we'll be able to query using the [in](https://keen.io/docs/api/#filters) operator. This way we can ask questions such as "What was the conversion rate for people in this variant A" or "What was the conversion rate for people in A and B".  I don't expect we'll ever need the `tests` array, but its there if we do.

### DIFF
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -24,6 +24,11 @@ export function getSession() {
  * @return {Object}        Object to send
  */
 export function transformState(action, state) {
+  const experiments = {
+    tests: Object.keys(state.experiments),
+    variants: Object.keys(state.experiments).reduce((acc, key) => [...acc, key], []),
+  };
+
   const transformation = {
     feed: {
       page: state.blocks.offset,
@@ -40,6 +45,7 @@ export function transformState(action, state) {
       session: getSession(),
       ...state.user,
     },
+    experiments,
     action,
   };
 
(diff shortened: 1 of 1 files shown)


resources/assets/helpers/analytics.js
```diff
@@ -24,6 +24,12 @@ export function getSession() {
  * @return {Object}        Object to send
  */
 export function transformState(action, state) {
+  const experimentKeys = state.experiments ? Object.keys(state.experiments) : [];
+  const experiments = {
+    tests: experimentKeys,
+    variants: experimentKeys.reduce((acc, key) => [...acc, key], []),
```
weerd: What's `acc`?

DFurnes: I'm confused what this does... don't we just end up reducing to the same array we started with...?

deadlybutter: Accumulator, its a common abbreviation for it

deadlybutter: @DFurnes yeah it looks that was supposed to be state.experiments[key] :man_facepalming:

deadlybutter: screen shot 2017-05-16 at 11 24 14 am

Here's what it looks like with the reducer function fixed

DFurnes: Can we use `Object.values` for this?

deadlybutter: That's a thing!?!

deadlybutter:
this latest commit is appropriately named "using daves wizard magic"

PR #258, opened 2017-05-16
Get user screen size

deadlybutter:
This pr...

- adds a helper function to get user screen size
- add the screen size to the keen payload
- fixes a proptype warning

screen shot 2017-05-16 at 1 13 30 pm
screen shot 2017-05-16 at 1 13 12 pm
screen shot 2017-05-16 at 1 13 07 pm

### DIFF
--- a/resources/assets/components/CompetitionBlock/index.js
+++ b/resources/assets/components/CompetitionBlock/index.js
@@ -83,7 +83,7 @@ CompetitionBlock.propTypes = {
     jobTitle: PropTypes.string.isRequired,
     avatar: PropTypes.string.isRequired,
   }).isRequired,
-  experiments: PropTypes.string,
+  experiments: PropTypes.object, // eslint-disable-line react/forbid-prop-types
   joinCompetition: PropTypes.func.isRequired,
   checkForCompetition: PropTypes.func.isRequired,
   hasJoinedCompetition: PropTypes.bool.isRequired,
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -46,7 +46,7 @@ export function transformState(action, state) {
       ...state.user,
     },
     experiments,
-    browser: { // Just in case we ever need a store item named "browser", prefixing with _
+    browser: {
       size: getFormattedScreenSize(),
     },
     action,
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -1,7 +1,7 @@
 /* global document, location, localStorage */
 
 import { analyze } from '@dosomething/analytics';
-import { generateUniqueId, isTimestampValid } from '../helpers';
+import { generateUniqueId, isTimestampValid, getFormattedScreenSize } from '../helpers';
 
 const DEVICE_ID = 'DEVICE_ID';
 const SESSION_ID = 'SESSION_ID';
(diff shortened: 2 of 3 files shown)


resources/assets/helpers/index.js
```diff
@@ -292,3 +292,29 @@ export function getDaysBetween(dateOne, dateTwo) {
...
+ * of the current display size.
+ *
+ * @param  {int} width - Defaults to the screen width
+ * @return {String}
+ */
+export function getFormattedScreenSize(screenWidth = window.innerWidth) {
+  const breakpoints = [
+    {
+      name: 'small',
+      test: width => width <= 759,
```
weerd: this looks so weird, lol

deadlybutter: Javascript!

DFurnes: Oh that's real funny looking. 😆

resources/assets/components/CompetitionBlock/index.js
```diff
@@ -83,7 +83,7 @@ CompetitionBlock.propTypes = {
     jobTitle: PropTypes.string.isRequired,
     avatar: PropTypes.string.isRequired,
   }).isRequired,
-  experiments: PropTypes.string,
+  experiments: PropTypes.object, // eslint-disable-line react/forbid-prop-types
```
DFurnes: Not something to block this pull request, but I think this is an indicator that we should be rethinking our approach here if we're constantly disabling the same linting rule. (Or rethinking the rule!)

resources/assets/helpers/analytics.js
```diff
@@ -46,6 +46,9 @@ export function transformState(action, state) {
       ...state.user,
     },
     experiments,
+    _browser: { // Just in case we ever need a store item named "browser", prefixing with _
```
weerd: Isn't this the case for it? Or do you see some other issue for needing to prefix?

DFurnes: I don't know how likely this is to happen... I'd rather we deal with that once it occurs, if so.

resources/assets/helpers/index.js
DFurnes: This might read a bit better with parenthesis around the expression?

```js
{
 test: width => (width <= 759),
 // ...
 test: width => (width >= 760 && width <= 959),
```

resources/assets/components/CompetitionBlock/index.js
deadlybutter: I think in this case we can't reasonably specify the shape for every container, so it's intentionally left as just "object". I guess it's a "use discretion" rule? (so maybe just make it a warning?)

resources/assets/helpers/analytics.js
deadlybutter: Ok, ill rename it to just browser then!

resources/assets/helpers/index.js
weerd: I think the linter would likely recommend the parenthesis as well ;)

deadlybutter:
cool I think it reads a bit nicer now

resources/assets/helpers/analytics.js
```diff
@@ -46,6 +46,9 @@ export function transformState(action, state) {
       ...state.user,
     },
     experiments,
+    browser: { // Just in case we ever need a store item named "browser", prefixing with _
```
DFurnes: Can you remove the comment now that this has been renamed?

deadlybutter:
comment removed!

PR #260, opened 2017-05-16
Add doctrine/dbal package for changing columns.

DFurnes:
I'm a dummy, and didn't include `doctrine/dbal` package so the migration added in #259 was failing.


weerd:
@DFurnes any of the other PRs that involved a similar change in other repos missing this package?

PR #261, opened 2017-05-16
Route improvements

deadlybutter:
This PR introduces some better routing logic into the Keen data payload. Currently, we're doing some hacky querying in Keen to figure out what tab you're on, and it doesn't seem like a great long-term idea.

To resolve this,

1) Standardized the paths in an object multiple places can reference 
2) Created a function that can intake a path and figure out the display name for it
3) Updated the Keen payload with this new data


deadlybutter:
Note: @ngjo Once this is live on prod we won't be able to get the latest conversion/return/etc rates until 

1) I update `next-data` to use these variables (easy)
2) I retroactively fix all of our existing events to use this new & better data format. (more complex)

I'm hoping it doesn't take more than a day though

deadlybutter:
Note to self: Update router definitions in the App as well

deadlybutter:
screen shot 2017-05-17 at 10 06 03 am

here's what it looks like after all the changes, seems like everything still works as intended and the data is showing correctly

resources/assets/components/Chrome.js
```diff
@@ -56,7 +56,7 @@ Chrome.propTypes = {
     timezone: PropTypes.string,
     timezone_type: PropTypes.number,
   }).isRequired,
-  children: PropTypes.array.isRequired, // eslint-disable-line react/forbid-prop-types
+  children: PropTypes.node.isRequired, // eslint-disable-line react/forbid-prop-types
```
weerd: I think you can remove the comment here now that this changed to `node`...

resources/assets/helpers/analytics.js
```diff
@@ -24,28 +25,28 @@ export function getSession() {
...
     signups: state.signups,
     submissions: state.submissions,
     user: {
       session: getSession(),
       ...state.user,
     },
-    experiments,
+    experiments: {
+      tests: state.experiments ? Object.keys(state.experiments) : [],
+      variants: state.experiments ? Object.values(state.experiments) : [],
```
weerd: @deadlybutter any chance we could change `variants` to `alternatives` which is the terminology Sixpack uses and I've been using elsewhere? I know this is from a previous PR 😬

deadlybutter: Sure!

deadlybutter:
@weerd should be good to go!

resources/assets/helpers/navigation.js
```diff
@@ -1,5 +1,40 @@
...
+
+  // Check if /pages/faq starts with /pages/.
+  const match = pathValues.find(path => route.startsWith(path));
+  if (! match) return 'undefined route';
+
+  // Find the display name for the matched path value.
+  // This is a bit crazy because we need to find the index
+  // in the `paths` object and the `pathValues` array is in
+  // a different order.
+  let name = Object.keys(paths)[Object.values(paths).findIndex(path => path === match)];
```
DFurnes: Lodash's `findKey` method seems like a good fit for this.

resources/assets/helpers/navigation.js
```diff
@@ -1,5 +1,40 @@
...
+  pages: '/pages/',
+};
+
+/**
+ * Get the display name of the given route,
+ *
+ * @param  {String} route
+ * @return {String}
+ */
+export function getRouteName(route) {
```
DFurnes: If it's not too much trouble, can you add some unit tests to demonstrate expected behavior for this?

deadlybutter: @DFurnes Would it be ok if I do this later? trying to minimize the data blackout

resources/assets/helpers/navigation.js
deadlybutter: I can swap this one out tho

resources/assets/helpers/navigation.js
DFurnes: Sure. Let's try to get in the habit of writing tests when we make changes in the future though.

deadlybutter:
Here's the output using the lodash find key, seems like everything still works 
screen shot 2017-05-17 at 1 52 15 pm

PR #262, opened 2017-05-16
Updating how we setup experiments to hopefully make things easier.

weerd:
This PR updates how we implement Sixpack A/B tests within Phoenix-Next. To begin with, instead of having to specify experiments (name and list of alternatives) within the `CampaignController`, we can just update a JSON file. We can then call a function `get_experiment_alternatives_selection()` which will create (if not already created in Sixpack) and participate the current client in the experiment, as well as return an array of all the selected alternatives the current client has been assigned to.

```javascript
// experiments.json example experiments definitions
{
  "competitions_prompt_style": {
    "a": "default_block",
    "b": "colorful_block"
  },
  "lede_banner_number_of_buttons": {
    "a": "one_button",
    "b": "two_buttons",
    "c": "three_buttons"
  }
}
```
The return value from calling the `get_experiment_alternatives_selection()` function:

```php
// experiments and selected alternatives for current client
[
  "competitions_prompt_style" => "default_block"
  "lede_banner_number_of_buttons" => "two_buttons"
]
```

The selected alternative array above is what gets passed to the redux store, which contains the same established structure:

image

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -128,7 +128,7 @@ function get_experiment_alternatives_selection()
 
     $experiments = get_experiments();
 
-    return collect($experiments)->map(function($alternatives, $name) use ($sixpack) {
+    return collect($experiments)->map(function ($alternatives, $name) use ($sixpack) {
         return $data[$name] = $sixpack->participate($name, array_values($alternatives))->getAlternative();
     })->toArray();
 }
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -6,6 +6,7 @@ use Contentful\Delivery\Asset;
 use App\Services\PhoenixLegacy;
 use Illuminate\Support\HtmlString;
 use Contentful\Delivery\DynamicEntry;
+use Illuminate\Support\Facades\Storage;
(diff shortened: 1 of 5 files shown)


app/helpers.php
```diff
@@ -94,6 +95,46 @@ function markdown($source)
...
 
 /**
+ * Get all Sixpack experiments from experiments definition file.
+ *
+ * @return array
+ */
+function get_experiments()
+{
+    try {
+        $experiments = file_get_contents(resource_path('assets/experiments.json'));
```
DFurnes: Can we use Laravel's `Storage::get` for this?

weerd: Oh, jes! Good call :)

deadlybutter:
Actually something I just thought of, does this mean we had to read from disk on every request? (to load the json file)

weerd:
@deadlybutter yes, for the time being. I'd like to improve all of this over time to make it better. I'm also concerned with shooting all those requests to sixpack to participate in experiments, but something we can keep an eye on and improve if as we use it more it becomes less efficient.

deadlybutter:
@weerd gotcha

PR #263, opened 2017-05-17
Fixing fixing fixing

deadlybutter:
The subfunction in our initialize function wasn't being called, so nothing was booting up. This caused numerous errors, including the data downtime and the signup number being 0.


resources/assets/helpers/analytics.js
```diff
@@ -42,8 +42,8 @@ export function transformState(action, state) {
     },
     routing: {
       referer: document.referrer,
-      path: state.routing.location.pathname,
-      page: getRouteName(state.routing.location.pathname),
+      path: state.routing.location ? state.routing.location.pathname : '',
+      page: state.routing.location ? getRouteName(state.routing.location.pathname) : '',
```
deadlybutter: We didn't spot this error before because this only happens for one of the initial actions on startup

DFurnes:
Looks like there are some linter errors, but otherwise looks a-okay to me. 👌

deadlybutter:
@DFurnes For phoenix next this is the only error,

```
WARNING in ./resources/assets/components/LazyImage/index.js

/Users/josephkent/code/dosomething/homestead/phoenix-next/resources/assets/components/LazyImage/index.js
  32:7  warning  img elements must have an alt prop, either with meaningful text, or an empty string for decorative images  jsx-a11y/img-has-alt
```

I assume you meant the ds-analytics error?

DFurnes:
Was in the previous build, looks like your merge commit fixed things up.

PR #264, opened 2017-05-17
adding app init component

deadlybutter:
Unfortunately, another unexpected problem from the react router upgrade, we hit a snag here where the routing information is not in the store when the app boots up, so our APP_INIT data is currently returning blank paths.

This PR is literally a hack to get around this, I don't like it at all but had no better ideas.


deadlybutter:
Here's what it looks like in console with the fix 
screen shot 2017-05-17 at 3 36 32 pm

deadlybutter:
Works on thor, deploying again to mitigate the mini crisis 

screen shot 2017-05-17 at 3 41 57 pm

PR #265, opened 2017-05-17
Sixpacking first experiments

weerd:
This PR adds Sixpack experiments for the LedeBanner and the Competition Prompt.

Experiment alternative that has 2 buttons in the lede banner:
image

Experiment alternative that updates the style of the competition prompt:
screen shot 2017-05-17 at 4 36 42 pm

Some of the setup for these experiments isn't ideal, but we're working to find solutions to make adding these tests easier in the future.

### DIFF
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -4,6 +4,7 @@ namespace App\Http\Controllers;
 
 use Illuminate\Http\Request;
 use App\Services\PhoenixLegacy;
+use SeatGeek\Sixpack\Session\Base as Sixpack;
 
 class SignupController extends Controller
 {
@@ -53,6 +54,10 @@ class SignupController extends Controller
         $this->validate($request, [
             'campaignId' => 'required',
         ]);
+
+        // @TEST 2017-05-17 lede_banner_number_of_buttons
+        $sixpack = app(Sixpack::class);
+        $sixpack->convert('lede_banner_number_of_buttons');
 
         return $this->phoenixLegacy->storeSignup(
             auth()->id(),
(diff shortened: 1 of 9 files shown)


resources/assets/components/LedeBanner/index.js
```diff
@@ -13,12 +14,33 @@ const LedeBanner = ({
...
+
+  // @TEST 2017-05-17 lede_banner_number_of_buttons
+  const submissionActions = () => {
+    const buttonCount = get(experiments, 'lede_banner_number_of_buttons', null);
+
+    if (buttonCount === 'two_buttons') {
+      return (
+        <ul className="button-group">
+          <li><button className="button" onClick={() => onClick({ source: 'lede banner|text: Support the cause' })}>Support the cause</button></li>
+          <li><button className="button" onClick={() => onClick({ source: `lede banner|text: ${verb.plural} ${noun.plural}` })}>{verb.plural} {noun.plural}</button></li>
```
deadlybutter: Can we make this just `source: lede banner|text: noun` but as a static string? This will be hard to query across all campaigns

app/Http/Controllers/SignupController.php
```diff
@@ -54,6 +55,10 @@ public function store(Request $request)
             'campaignId' => 'required',
         ]);
 
+        // @TEST 2017-05-17 lede_banner_number_of_buttons
+        $sixpack = app(Sixpack::class);
```
DFurnes: Totally not needed for this PR, but could be nice to make a `sixpack()` helper since we ask for this a lot!

weerd: oh jes! I can do that for upcoming tests 👍

PR #266, opened 2017-05-18
Add the Scroll Concierge™.

DFurnes:
#### Changes
This pull request includes the __Scroll Concierge™__, which can guide the user to a particular area on the page upon load. It also includes a small fix to `FlexCell` prop types & an incorrect docblock.

I'm currently figuring out some occasional bugs, so not ready to merge yet. 🚧

PR #269, opened 2017-05-19
Add some extra redirects & bug fixes.

DFurnes:
### Changes
This pull request re-configures some redirects for Uncle Mike – the homepage ([next.dosomething.org](https://next.dosomething.org)) now sends visitors directly to Sincerely Us, and the "next" campaign list can be viewed at `/campaigns`. It also includes a few bug fixes from #266.

👋

### DIFF
--- a/resources/assets/components/LedeBanner/index.js
+++ b/resources/assets/components/LedeBanner/index.js
@@ -68,7 +68,7 @@ LedeBanner.propTypes = {
     description: PropTypes.string,
     url: PropTypes.string,
   }).isRequired,
-  experiments: PropTypes.array, // eslint-disable-line react/forbid-prop-types
+  experiments: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
   isAffiliated: PropTypes.bool.isRequired,
   legacyCampaignId: PropTypes.string.isRequired,
   noun: PropTypes.shape({
--- a/resources/assets/components/LedeBanner/index.js
+++ b/resources/assets/components/LedeBanner/index.js
@@ -68,7 +68,7 @@ LedeBanner.propTypes = {
     description: PropTypes.string,
     url: PropTypes.string,
   }).isRequired,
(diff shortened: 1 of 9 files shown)


resources/assets/components/LedeBanner/index.js
```diff
@@ -68,7 +68,7 @@ LedeBanner.propTypes = {
     description: PropTypes.string,
     url: PropTypes.string,
   }).isRequired,
-  experiments: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+  experiments: PropTypes.array, // eslint-disable-line react/forbid-prop-types
```
deadlybutter: Wait, did the data structure change?

weerd: ??

deadlybutter: In the latest dev this is an object, 

screen shot 2017-05-19 at 2 24 45 pm

This prop is getting experiments from [here](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/Chrome.js#L26), and that comes from [this](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/containers/ChromeContainer.js#L26) container which is just grabbing the object

DFurnes: Ah yep, but it gives an array (:interrobang:) if Sixpack is disabled. Updated to accept either.

resources/assets/components/LedeBanner/index.js
```diff
@@ -68,7 +68,7 @@ LedeBanner.propTypes = {
     description: PropTypes.string,
     url: PropTypes.string,
   }).isRequired,
-  experiments: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+  experiments: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
```
weerd: :dancing_men:

PR #271, opened 2017-05-23
Fetch env vars on demand

deadlybutter:
We're seeing problems getting new env vars to populate in the webpack build, so this can act as a workaround for now

Here are some successful Keen requests 

screen shot 2017-05-23 at 12 56 29 pm
screen shot 2017-05-23 at 12 56 24 pm

### DIFF
--- a/config/services.php
+++ b/config/services.php
@@ -17,6 +17,7 @@ return [
     'analytics' => [
         'google_id' => env('GOOGLE_ANALYTICS_ID'),
         'keen_id' => env('KEEN_PROJECT_ID'),
+        'keen_key' => env('KEEN_WRITE_KEY'),
         'facebook_id' => env('FACEBOOK_APP_ID'),
     ],
 
--- a/resources/assets/middleware/analytics.js
+++ b/resources/assets/middleware/analytics.js
@@ -47,7 +47,7 @@ export function start() {
   checkSession();
 
   // Initialize Analytics
-  const env = window.env || {};
+  const env = window.ENV || {};
   init('track', true, env.KEEN_PROJECT_ID ? {
     projectId: env.KEEN_PROJECT_ID,
     writeKey: env.KEEN_WRITE_KEY,
--- a/resources/assets/middleware/analytics.js
+++ b/resources/assets/middleware/analytics.js
@@ -47,9 +47,10 @@ export function start() {
   checkSession();
 
   // Initialize Analytics
-  init('track', true, services.KEEN_PROJECT_ID ? {
-    projectId: services.KEEN_PROJECT_ID,
-    writeKey: services.KEEN_WRITE_KEY,
+  const env = window.env || {};
+  init('track', true, env.KEEN_PROJECT_ID ? {
+    projectId: env.KEEN_PROJECT_ID,
+    writeKey: env.KEEN_WRITE_KEY,
   } : null);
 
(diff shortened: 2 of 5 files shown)


config/services.php
```diff
@@ -17,6 +17,7 @@
     'analytics' => [
         'google_id' => env('GOOGLE_ANALYTICS_ID'),
         'keen_id' => env('KEEN_PROJECT_ID'),
+        'keen_key' => env('KEEN_WRITE_KEY'),
```
weerd: Were we missing this [before](https://github.com/DoSomething/phoenix-next/pull/271/files#diff-e3c99df1d0ac9a756f950539c98e36acL52)?

deadlybutter: We were missing it on the PHP side because it wasn't needed, that key was coming from the webpack define plugin

deadlybutter: Here is the specific line in the webpack config https://github.com/DoSomething/phoenix-next/blob/dev/webpack.config.js#L33

resources/assets/middleware/analytics.js
```diff
@@ -47,9 +47,10 @@ export function start() {
   checkSession();
 
   // Initialize Analytics
-  init('track', true, services.KEEN_PROJECT_ID ? {
-    projectId: services.KEEN_PROJECT_ID,
-    writeKey: services.KEEN_WRITE_KEY,
+  const env = window.env || {};
+  init('track', true, env.KEEN_PROJECT_ID ? {
+    projectId: env.KEEN_PROJECT_ID,
+    writeKey: env.KEEN_WRITE_KEY,
```
weerd: If it's an empty object, do these just result to `null` or does it get angry?

deadlybutter: `env.KEEN_PROJECT_ID ? {` note the question mark! And if there was any errors its handled on the ds-analytics side of things

resources/views/layouts/master.blade.php
```diff
@@ -29,6 +29,7 @@
 
     @include('partials.analytics')
     {{ isset($state) ? scriptify($state) : scriptify() }}
+    {{ isset($env) ? scriptify($env, 'env') : scriptify() }}
```
weerd: Should we capitalize `ENV`...? I think that's a convention that we follow.

deadlybutter: @weerd Are you suggesting we capitalize the php var or the `window.env` key? (Or both?)

config/services.php
weerd: Ah, jes. ok gotcha!

resources/assets/middleware/analytics.js
weerd: Ah cool. Yeah, missed the `?` :)

resources/views/layouts/master.blade.php
weerd: `window.ENV`, kinda like we do w/ `window.STATE`. Not the PHP var... sorry. Should have been more specific!

deadlybutter: Fine with me! One sec

deadlybutter:
@weerd updated & retested!

ngjo:
[UNPLANNED WORK: Fixing the env variables](https://trello.com/c/VOKvzlUC/667-unplanned-work-fixing-the-env-variables)

PR #273, opened 2017-05-24
Sixpack with a sprinkling of react

weerd:
Hokay, here we go. Maybe grab a nice ☕️ or 🍵 to sit back with and drink whilst reviewing this PR 😬 Sorry for the length!

To begin with this PR turns off the Sixpack integration via the PHP backend. No longer are experiments executed for all users on loading a campaign.

This PR enables Sixpack A/B experiments via JavaScript and our React-based front-end.

Please refer to the [wiki documentation](https://github.com/DoSomething/phoenix-next/wiki/Sixpack-AB-Testing) for a more detailed explanation!


app/Http/Controllers/CampaignController.php
```diff
@@ -57,15 +57,14 @@ public function show($slug)
     {
         $campaign = $this->campaignRepository->findBySlug($slug);
         $shareFields = getShareFields($campaign, $campaign->socialOverrides);
-        $env = getClientEnvironmentVars();
+        $env = get_client_environment_vars();
```
weerd: We should be using snake_case variable names for helper functions. Missed this in prior PR!

deadlybutter: Yeah we started using a mix, seems like thats mostly my fault 😬

resources/assets/helpers/sixpack.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Participate current client to specified experiment.
+ *
+ * @param  {String} name
+ * @return {Promise}
+ */
+export function participate(name) {
+  return new Promise((resolve, reject) => {
```
weerd: Debating whether to add a `Promise` polyfill... seems unsupported in IE but supported on Edge... :man_shrugging: Maybe just see during testing? For future reference, needed to wrap these calls in a promise because Sixpack-client has its own methods that make requests to the Sixpack server for experiment participation and conversion. The promise helped deal with proceeding with logic after response received.

app/Http/Controllers/CampaignController.php
weerd: No worries. We can just fix up over time :)

resources/assets/middleware/experiments.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param  {String} condition
+ * @param  {Object} state
+ * @return {Boolean}
+ */
+function assertConditionPasses(condition, state) {
+  if (condition === 'unaffiliated') {
+    return ! state.signups.thisCampaign;
+  }
+
+  // @TODO: Add additional conditions. Maybe break this function out into other file.
```
deadlybutter: Yeah maybe this also goes in the experiments helper i mentioned earlier ^

resources/assets/helpers/sixpack.js
deadlybutter: We covered that already! https://babeljs.io/docs/usage/polyfill/

weerd: Ah perfect. Must have been in a late night haze 😴

resources/assets/actions/index.js
```diff
@@ -8,6 +8,12 @@
  */
 import { LOCATION_CHANGE } from 'react-router-redux';
 
+export const PARTICIPATE_IN_EXPERIMENT = 'PARTICIPATE_IN_EXPERIMENT';
+export const ADD_TO_EXPERIMENTS_STORE = 'ADD_TO_EXPERIMENTS_STORE';
+export const UPDATE_EXPERIMENTS_STORE = 'UPDATE_EXPERIMENTS_STORE';
+export const CONVERT_EXPERIMENT = 'CONVERT_EXPERIMENT';
```
weerd: Yeah, I was going to leave the Keen updates/additions to you since you're way more familiar with that stuff. 

On an additional note, I aimed to not have to change the `experiments` store property if I didn't have to. The only thing is that it isn't auto-populated immediately since these experiments (and thus populating the store with data) are now conditionally fired.

deadlybutter: No worries! In the future if you want to track an action just add it to [this](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/actions/index.js#L68) array and you're done!

resources/assets/components/Experiment/index.js
```diff
@@ -0,0 +1,51 @@
...
+class Experiment extends React.Component {
+  componentWillMount() {
+    this.props.participateInExperiment(this.props.name);
+  }
+
+  render() {
+    const env = window.ENV || {};
+
+    // Return default control component if Sixpack is not enabled.
+    if (typeof env.SIXPACK_ENABLED === 'undefined' || ! env.SIXPACK_ENABLED) {
```
weerd: Yeah, I was debating that. Isn't there some edge cases or something where that doesn't work and it's best to be more explicit. I couldn't remember offhand, but according to [this}(http://2ality.com/2013/04/check-undefined.html) (point #3) you're defs right and should be fine. I'll update!

resources/assets/middleware/experiments.js
weerd: Makes sense :)

resources/assets/components/Chrome.js
```diff
@@ -4,27 +4,48 @@ import AppInit from './AppInit';
...
-import LedeBanner from './LedeBanner';
+import LedeBanner from './LedeBanner/LedeBanner';
+import LedeBannerAltB from './LedeBanner/LedeBannerAltB';
+import ExperimentContainer from '../containers/ExperimentContainer';
 import NavigationContainer from '../containers/NavigationContainer';
 import AffirmationContainer from '../containers/AffirmationContainer';
 import NotificationContainer from '../containers/NotificationContainer';
 
+// @TODO: Might be useful to have a separate file of constants?
+const LEDE_BANNER_NUMBER_OF_BUTTONS = 'lede_banner_number_of_buttons';
```
weerd: Main reason I was hesitant with this is that it's very specific to the component alternatives, and will change back and forth as experiments are created. Initial thought was that if set in the file, then it'd be easier to deal with and remove once winning alternative selected? :man_shrugging:

resources/assets/actions/index.js
weerd: Hokay! I can add :)

resources/assets/components/Experiment/index.js
deadlybutter: One of the cases that will trip me up is arrays. EG: 

```
const arr = [];

arr.length // 0
if (arr.length) // false
if (arr.indexOf('test')) // true?! --> because this is only false when its 0.
```

resources/assets/components/Chrome.js
deadlybutter: Oh you meant the JSON file... hmm

weerd: Relates to the JSON file, but yeah. Ideal is just not having to modify multiple files... maybe something to come back :)

resources/assets/components/Experiment/index.js
weerd: Good to know!

deadlybutter:
🚂 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋

weerd:
image

ngjo:
[(5) As a developer, I want to make a better system for setting up AB tests in 6pack.](https://trello.com/c/ZunjmNox/653-5-as-a-developer-i-want-to-make-a-better-system-for-setting-up-ab-tests-in-6pack)

ngjo:
[(5) As a developer, I want to revise the two A/B tests (competition prompt and lede banner) to be more accurate.](https://trello.com/c/XrjLKhzS/662-5-as-a-developer-i-want-to-revise-the-two-a-b-tests-competition-prompt-and-lede-banner-to-be-more-accurate)

PR #275, opened 2017-05-24
Overlay component

deadlybutter:
This pull request adds an Overlay component. This component will present Notifications, Modals and anything else we want floating above the screen.


resources/assets/init.js
```diff
@@ -42,7 +42,7 @@ import { observerMiddleware } from './middleware/analytics';
 const history = historyInit();
 const middleware = [thunk, routerMiddleware(history), observerMiddleware];
 const store = configureStore({ ...reducers, routing: routerReducer }, middleware, window.STATE);
-
+store.dispatch({ type: 'ADD_NOTIFICATION' });
```
deadlybutter: delete this!

deadlybutter:
(Also: Take note at how i named the new component files, let me know whatcha think)

weerd:
mind popping up an example screenshot? Helps w/ visualizing some of the css code :)

weerd:
🤦‍♂️ did it again

deadlybutter:
@weerd you're a monster, and yeah sure thing, one sec

deadlybutter:
@weerd So the first screenshot is the overlay itself,

screen shot 2017-06-06 at 10 12 49 am

And this one is the individual components within the overlay

screen shot 2017-06-06 at 10 13 09 am

weerd:
So, there can be multiple items inside the overlay right? Does each of them have their own close `X` button? If so, then is there an additional close `X` button for the overlay; otherwise how do you close it?

deadlybutter:
Yeah, the close functionality is handled outside of the overlay. Modals for example will behave differently than the notifications, because clicking the black background is supposed to close it.

weerd:
Does it just auto-close when all things inside it have been closed?

deadlybutter:
@weerd There technically is no "close" here, if there are no children it just doesn't render anything

PR #276, opened 2017-05-24
Implementation of Competitions prompt experiment with new Sixpack setup.

weerd:
This PR uses the new React/Sixpack combo setup to implement a test for the Competitions Prompt. It showcases another example (now along w/ the `LedeBanner`) for how these tests are structured. 

It also shows an example where we need to work with a "container" and not the component directly. This made it slightly more complicated, and maybe highlights that a container isn't necessarily needed for the competition block, or maybe that we need to figure out an alternative structure for containers and their components? (@deadlybutter has some thoughts on this 👋). The core concept of how to implement the tests is still the same though: it just involved having a `CompetitionContainerAltB` AND corresponding `CompetitionBlockAltB` as alternatives.

Feel free to amend or merge this in if all looks good @deadlybutter


ngjo:
[(5) As a developer, I want to revise the two A/B tests (competition prompt and lede banner) to be more accurate.](https://trello.com/c/XrjLKhzS/662-5-as-a-developer-i-want-to-revise-the-two-a-b-tests-competition-prompt-and-lede-banner-to-be-more-accurate)

PR #277, opened 2017-05-24
Adding pre-test logic

deadlybutter:
This PR adds some "pre-test" logic to make sure we only run A/B tests when the Redux state meets certain requirements. 

Additionally, I did a few little cleanup things,

- Removed the second experiments file, was confusing having two.
- Cleaned up the LedeBanner props spacing
-


resources/assets/experiments.json
```diff
@@ -1,10 +1,27 @@
 {
   "competitions_prompt_style": {
-    "a": "default_block",
-    "b": "colorful_block"
+    "meta": {
+      "preTest": {
+        "signups.thisCampaignn": true,
```
katiecrane: typo! 🔍

PR #283, opened 2017-05-26
Attempting to fix wercker build

deadlybutter:
See here for why i think this fixes it: https://github.com/nodesource/distributions/issues/53#issuecomment-70392701


deadlybutter:
this build failed, but the other just passed. and according to the third law of computers anything that doesn't make sense at all is probably a network error

deadlybutter:
i totally made that up but it sounds real

PR #285, opened 2017-05-26
committing idea for improving containers

deadlybutter:
I made a quick & tiny [package](https://github.com/deadlybutter/easy-container/blob/master/index.js) for making redux containers a lot faster. This PR shows how that could be used for the affirmation component as an example, and also plays with the idea of switching around file naming conventions to be easier for searching within code editors.


deadlybutter:
A part of me also thinks the index file should not even do the connect w/ Redux, as in theory any component should be able to connect itself now. The downside will be testing, we might need to export two different versions of the component?

deadlybutter:
I'll dig into this later since not pressing, but build is failing on Wercker twice now with same error

```
ERROR in app-1135825b8144654737ac.js from UglifyJs
Invalid assignment [app-1135825b8144654737ac.js:35474,39]
```

resources/assets/components/Affirmation/index.js
```diff
@@ -1,68 +1,5 @@
...
-    photo: '',
-    author: '',
-    quote: '',
-    callToActionHeader: '',
-    callToActionDescription: '',
-  },
-};
-
-export default Affirmation;
+export default Container(connect, Affirmation);
```
DFurnes: This seems like an awkward abstraction to me. What's the advantage over idiomatic `connect` here?

DFurnes:
> The downside will be testing, we might need to export two different versions of the component?

Yeah, we could probably handle that by making the default export the connected component, and then have a separate named export for the "base" component to be used for testing.

resources/assets/components/Affirmation/index.js
deadlybutter: The (very tiny) package just abstracts out the `mapStateToProps` and `actionCreators` automatically. I would have made it a helper but I plan on using this for my side projects as well. If its not intuitive I can keep fiddling with it.

DFurnes: I'm just wary of creating our own special format for this since it doesn't seem to save any typing over standard `connect()` usage. What problem are we solving here?

deadlybutter: Fair enough, we can type it manually!

PR #286, opened 2017-05-30
"Pitch page" A/B Test

deadlybutter:
I made a thing this weekend. Still some edits to make based on feedback but it's nearly ready for a dev look


deadlybutter:
context I've sent to devs:

heya! ok so the pitch page is good for “launch”. just some notes before I ask you to look at it (I think Ill cross post in this in the PR as well)

- started building this with the intent to have some “modular” components we could one day assemble in like Contentful, so this is another “tool in the toolbox”. however, that adds complexity and we dont even know if this works, so I started fine tuning certain styles to be specific to this campaign & hardcoded all of the content in a separate file

- if this test fails, we can delete the `/pitch` folder and move on with our lives. If it works, assume a lot of this will be re-worked…

-_REGARDLESS_, we also now have the ability to run more full page experiments in the future. so most important review feedback is actually on the changes to how we render the app itself (the experience component, App.js)

Also NOTE: it also doesn’t have signup functionality, final copy, or the A/B test. I will take care of that in a followup PR

resources/assets/components/Pitch/pitchToolbox.scss
```diff
@@ -0,0 +1,3 @@
+@import '../../scss/next-toolbox';
+
+// This is just for saving me the extra ../../ in all of the files.
```
DFurnes: 🤔

deadlybutter: @DFurnes This is my peak laziness

deadlybutter: In retrospect its probably more characters to do it this way, I'll get rid of it later if we want to continue with all of this

weerd:
Post-merge 👍 Looks cool and should make for a good experiment. If it works out defs stuff to consider refactoring as you mentioned.

PR #288, opened 2017-06-01
Update to remove signup requirement for experiment conversion.

weerd:
This PR provides a quick update to remove the the post-signup requirement to convert on any Sixpack experiments.

### DIFF
--- a/resources/assets/actions/experiments.js
+++ b/resources/assets/actions/experiments.js
@@ -3,16 +3,10 @@ import {
   ADD_TO_EXPERIMENTS_STORE,
   UPDATE_EXPERIMENTS_STORE,
   CONVERT_EXPERIMENT,
-  queueEvent,
 } from '../actions';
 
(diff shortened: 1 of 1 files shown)


resources/assets/actions/experiments.js
```diff
@@ -3,16 +3,10 @@ import {
...
 } from '../actions';
 
 export function convertExperiment(name) {
-  return (dispatch, getState) => {
-    if (! getState().user.id) {
-      dispatch(queueEvent('convertExperiment', name));
-      return;
-    }
-
+  return (dispatch) => {
```
DFurnes: Want to un-thunk these while you're in this file?

weerd: I was planning on doing that in an update in upcoming PR, because I wanted to test it and make sure all works ok. At least I'm more confident this works as is, since it's just back to the way it was before.

PR #289, opened 2017-06-01
Updating how history state is pushed.

weerd:
This PR fixes an issue with the tabbed navigation bar not indicating the correct page after a redirect post signup.

### DIFF
--- a/resources/assets/containers/NavigationContainer.js
+++ b/resources/assets/containers/NavigationContainer.js
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { connect } from 'react-redux';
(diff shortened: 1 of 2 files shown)


resources/assets/containers/NavigationContainer.js
```diff
@@ -9,7 +10,9 @@ const mapStateToProps = state => ({
   pathname: state.routing.location.pathname,
 });
 
-const NavigationContainer = ({ pages }) => {
+const NavigationContainer = (props) => {
+  const { pages } = props;
```
DFurnes: Oh, totally unimportant but can change this back if you want. ✏️

weerd: ah jes.

PR #290, opened 2017-06-01
Scroll Concierge, Part II

DFurnes:
### Changes
This pull request includes a slew of little fixes, mainly focused around the Scroll Concierge™:

🚧 Allows the Laravel debug bar to be toggled off with `APP_DEBUGBAR`. I found it was getting in the way when testing mobile layout stuff, so nice to be able to "shush" it from time to time.

🏃 Removes the `ready` event handler from the main entry point, since the script is already loading at the very bottom of the body & the DOM will be ready regardless.

⌛️ Tell the Scroll Concierge to wait for the browser to finish layout & reflowing the page, so we can better ensure the scroll offset is accurate. This one is pretty funky, but interesting!

✋ Fixes an edge case where we could accidentally scroll past the desired offset depending on rendering performance. Now we always stop at the exact `offset` rather than right after it.

:left_right_arrow: Fixes issue with navigation scrollbars appearing on desktop with `overflow-x: auto`, and another issue where the `bottom-border` was getting cut off by adding a lil' padding.


weerd:
> 🚧 Allows the Laravel debug bar to be toggled off with APP_DEBUGBAR. I found it was getting in the way when testing mobile layout stuff, so nice to be able to "shush" it from time to time.

Nice. Yeah, it does minimize to a tiny button but have been toggling it on occasion. Sounds like a good addition :)

resources/assets/components/ScrollConcierge/index.js
```diff
@@ -4,6 +4,20 @@ import React from 'react';
...
+ * so we don't read the wrong values for layout.
+ *
+ * @see https://stackoverflow.com/a/34999925/811624
+ */
+const waitForLayout = (callback) => {
+  // Wait for the call stack to clear.
+  setTimeout(() => {
+    // Then, wait until we've rendered a frame.
+    window.requestAnimationFrame(callback);
+  }, 0);
```
weerd: Interesting 🤔

PR #291, opened 2017-06-01
Show a 404 when a user goes to a bad "custom page" slug.

DFurnes:
### Changes
This is a quick fix for when users go to a slug that doesn't exist, like `/campaigns/sincerely-us/pages/skooolershooops` or something very silly. I'd like to refactor how we handle this in the future to move it back into the router, but this way we give people a better experience than a blown-up page.

I also fixed the padding between the content page and footer, since it was real cozy.

### DIFF
--- a/resources/assets/components/ContentPage/content-page.scss
+++ b/resources/assets/components/ContentPage/content-page.scss
@@ -1,10 +1,9 @@
 @import '../../scss/next-toolbox';
 
 .content-page {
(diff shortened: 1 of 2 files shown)


ngjo:
[As a developer, I want to add error handling for routes w/ slugs](https://trello.com/c/4FIjfhEn/657-as-a-developer-i-want-to-add-error-handling-for-routes-w-slugs)

PR #292, opened 2017-06-01
Pitch fixes and AB test

deadlybutter:
This pr fixes a few things and introduces the pitch (BACKGROUND, whatever) A/B test.

This test will split unaffiliated traffic to the community tab 50/50 and is checking for any signup on either experience.

It also adds some style fixes and signup functionality


resources/assets/helpers/navigation.js
```diff
@@ -6,7 +6,7 @@ export const paths = {
   community: '/',
   action: '/action',
   pages: '/pages/',
-  pitch: '/pitch',
+  background: '/background',
```
weerd: Curious... why'd you rename this to `background`?

resources/assets/components/App.js
```diff
@@ -24,18 +25,38 @@ const chrome = component => wrap(ChromeContainer, component);
 const App = ({ store, history }) => {
   initializeStore(store);
 
+  const PitchTest = (props) => {
+    const PitchRedirect = () => {
+      history.push(`${paths.background}`);
```
weerd: Should we be using the `react-router-redux` method here:
```js
dispatch(push(`${paths.background}`));
```

deadlybutter: I mean I know from testing it functionally works - philosophically speaking, not sure. I think given that the history is intentionally passed to the App component for router usage, its safe to use it here?

resources/assets/components/Pitch/index.js
```diff
@@ -3,10 +3,10 @@ import PropTypes from 'prop-types';
 
 import PitchSincerelyUs from './PitchCampaigns/SincerelyUs';
 
-const Pitch = ({ campaignId }) => {
-  if (campaignId !== '7656') return null;
+const Pitch = (props) => {
+  if (props.campaignId !== '7656') return null;
```
weerd: Hmm, not super keen on this being hard-coded. I think it should be fine for this experiment, but we should be careful with this. Is this the Production campaign ID or the Thor campaign ID?? Seems like a point of confusion.

deadlybutter: thankfully those should always be the same, what's a better way of doing this in the future? by title?

resources/assets/components/App.js
weerd: In prior PR I've updated the calls to `push` in the `/actions/signups.js` and would be better to be consistent. I went off of what was suggested in the usage section [here](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux#usage).

resources/assets/components/Pitch/index.js
deadlybutter: Oh I bet we could use the contentful id as well....

deadlybutter: This thread made me also realize, we should include this id check in the experiments.json !!

weerd: Maybe ENV vars... not sure. Maybe having a prop in the Contenful JSON object that says whether to show pitch page?

deadlybutter: @weerd I think the contentful id would be a lot simpler, no? otherwise we'd have to update the env after the first thor/prod deploy

deadlybutter:
@weerd did some updates & re-testing with the contentful id, I think this should be better!

resources/assets/components/Pitch/index.js
DFurnes: Is this something we can add as a feature flag in the JSON on the campaign?

DFurnes:
Sorry, still catching up – what's the advantage of hard-coding a campaign ID over treating this as a feature flag on the campaign?

deadlybutter:
@DFurnes I'm just trying to minimize the number of changes that need to be reverted in the event the test isn't successful. Using ID just seems like a simple way to test this

resources/assets/components/Pitch/index.js
weerd: I feel adding a boolean to the contentful JSON object would be the ideal approach @deadlybutter Rather not rely on specific IDs (even if it is the contentful id).

weerd:
@deadlybutter I'd prefer if we moved to using the JSON object on the contentful campaign entry. It'll help also set a precedent for how we approach these kinds of things in the future; stuff that needs to apply to a specific campaign is defined in the JSON object which is easy to edit/remove at a later date. It'll also be really easy to see what specific stuff is set for a campaign, instead of trying to find it in code. We look at the JSON object in contentful and see, "oh this campaign has a pitch page defined, needs an address entry form, and other things" if that makes sense. What do you think?

DFurnes:
Yeah, this is a perfect example of what that JSON field was added for.

deadlybutter:
Sorry grabbed early lunch - I updated & retested using custom json property!

deadlybutter:
@DFurnes Excuse me sir, it's "Let's do this!" Have you not read the brand book?

🚂 🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋

PR #293, opened 2017-06-01
Reportback uploader quantity label renaming

ngjo:
We changed "How many cards are in the photo?" to "Total number of cards made." because there was some confusion from competitors looking at the leaderboard (leaderboard shows most recent quantity, not a sum of the submitted RB quantities). 

Small caveat - 'made' is hardcoded into that field so when we have other campaigns with a different 'verb' action, we'll need to revisit this. This shouldn't be a big problem for the next two campaigns we're launching.


weerd:
@ngjo Let's be sure to add an issue or icebox trello card to remember to come back to this! 🤔

PR #294, opened 2017-06-02
Sticky tabbed navigation

weerd:
This PR updates the campaign page section navigation so that it can behave as a "sticky nav" as the user scrolls the page, allowing them to access the other sections of the campaign easily while on medium or large devices.

_Note_: IE 11/Edge do not support CSS `sticky` so this will not work on those browsers and _should_ just behave as it did before; scrolling with rest of page. We were considering using a polyfill, but the one we use on PhoenixAshes [is deprecated](https://github.com/filamentgroup/fixed-sticky#-this-plugin-is-deprecated-) (as of 2 hours ago 😱 ). Microsoft currently has this CSS feature under development and should appear on the Edge browser in the future, so that seems good enough 🤷‍♂️ 

### Sticky Nav in action
tabbed nav

### Updated mobile nav version
The tabbed navigation on mobile also now adds a subtle fade to better indicate there's more stuff to the right. We can adjust the amount of fade if desired.
image

This PR also does a little refactoring of the Navigation components to better match what navigation they are (`TabbedNavigation` instead of being misunderstood as a main navigation component), moves a couple of those files around and is trying out moving constants to their own dedicated directory.

Refs: https://trello.com/c/UP1Nbund/363-3-as-a-large-and-medium-screen-sized-user-i-want-a-sticky-navigation-to-follow-me-down-the-page-so-that-i-can-jump-to-section-or


resources/assets/components/Navigation/NavigationLink.js
```diff
@@ -0,0 +1,25 @@
...
+  <NavLink
+    {...props}
+    className={classnames('nav-link', props.className)}
+    activeClassName="is-active"
+  >
+    { props.children }
+  </NavLink>
+);
+
+export default NavigationLink;
```
DFurnes: This is kinda subjective, but want to move this default export to the bottom like we do elsewhere?

resources/assets/components/Navigation/TabbedNavigation.js
```diff
@@ -0,0 +1,90 @@
...
+import './tabbed-navigation.scss';
+
+class TabbedNavigation extends React.Component {
+  /**
+   * Class contructor
+   * @param  {Object} props
+   */
+  constructor(props) {
+    super(props);
+    console.log(props);
```
DFurnes: 👾

resources/assets/components/Navigation/TabbedNavigation.js
```diff
@@ -0,0 +1,90 @@
...
+
+    this.tabbedNav = null;
```
DFurnes: Can we rename to `this.node` to make it more clear it's a reference to the "real" DOM node?

resources/assets/components/Navigation/TabbedNavigation.js
```diff
@@ -0,0 +1,90 @@
...
+    this.state = {
+      isStuck: false,
+    };
+  }
+
+  /**
+   * React lifecycle method called after render() method runs.
+   */
+  componentDidMount() {
+    this.tabbedNav = document.getElementById('tabbed-navigation');
```
DFurnes: Want to use a `ref` for this like we do [in the concierge](https://github.com/DoSomething/phoenix-next/blob/b767ec7a3a6b82bdabdee43eeb0614f51ab9eea3/resources/assets/components/ScrollConcierge/index.js#L73)?

resources/assets/components/Navigation/tabbed-navigation.scss
```diff
@@ -0,0 +1,67 @@
...
+      border-bottom: 4px solid rgba($off-black, 0.5);
+    }
+
+    &.is-active {
+      color: $black;
+      border-bottom: 4px solid $off-black;
+    }
+  }
+
+  &.is-stuck {
```
DFurnes: Let's move this up above the child selectors so it's more discoverable.

resources/assets/components/Navigation/TabbedNavigation.js
weerd: gah!

resources/assets/components/Navigation/TabbedNavigation.js
weerd: sure!

PR #297, opened 2017-06-07
Fixes for some quirky behavior with the sticky nav.

weerd:
This PR adds a few quick adjustments to the sticky nav so it works better and doesn't leave any visual bugs, or _seemingly_ visual bugs. I removed the transition animation for the background color since it seemed like a bug if you were scrolled a bit down and switched to the action page or community page and you see a slight flash of content behind the nav as the page repaints and the background color transitions in.

This also updates the visual offset on the Scroll Concierge™ so it'll scroll to a better location that triggers the sticky nav 😄

### DIFF
--- a/resources/assets/components/ScrollConcierge/index.js
+++ b/resources/assets/components/ScrollConcierge/index.js
@@ -62,7 +62,7 @@ class ScrollConcierge extends React.Component {
     // Wait for headline font to load so we don't scroll to
     // the wrong place when the page reflows & offset changes.
     font.load().then(() => {
-      const VISUAL_OFFSET = 100;
+      const VISUAL_OFFSET = 95;
 
       // Wait until the page has finished layout, then scroll.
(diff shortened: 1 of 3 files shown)


resources/assets/components/ScrollConcierge/index.js
```diff
@@ -62,7 +62,7 @@ class ScrollConcierge extends React.Component {
     // Wait for headline font to load so we don't scroll to
     // the wrong place when the page reflows & offset changes.
     font.load().then(() => {
-      const VISUAL_OFFSET = 100;
+      const VISUAL_OFFSET = 95;
```
DFurnes: Heh. 💯

weerd: 😬

weerd: 💯  - 5

PR #298, opened 2017-06-08
Fixes issue loading the site in IE 10.

### LINKED ISSUE #57: URL Polyfill?
> I'm not sure if it's already included in the fetch polyfill that we're using in Phoenix Next, but we may also need to include a [separate polyfill](https://github.com/webcomponents/URL) for the URL Standard.

References DoSomething/gateway-js#5.

DFurnes:
#### Changes
This pull request fixes an issue loading Phoenix Next campaigns in IE 10/11 by adding polyfills for `URL` and `window.location.origin` so we don't run into problems with Gateway. I moved the polyfills into a separate file so they're not taking up too much space in the entry file. Fixes #57.

I also updated the readme to include directions for building the front-end code. 🔨

👀

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "dosomething-modal": "^0.3.0",
     "enzyme": "^2.8.2",
     "enzyme-to-json": "^1.5.1",
-    "fontfaceobserver": "^2.0.9",
+    "fontfaceobserver": "^2.0.13",
     "history": "^4.5.1",
     "lodash": "^4.17.4",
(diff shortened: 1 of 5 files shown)


DFurnes:
@deadlybutter @weerd A new version of our Docker box [was published](https://github.com/DoSomething/ds-docker-php/pull/1) and it seems our overrides broke when applied to that. Make sure to rebase against this branch once it's merged to get the fix!

PR #299, opened 2017-06-08
Modal.

deadlybutter:
_Ok_, this one is a giant, and it could have been way larger but I'm stopping myself!

At the user level, this PR...

- Adds a post signup modal containing the affirmation and competition prompt.
- Removes the Overlay component and changes how notifications are displayed instead.

Here is a gif:

competition

-----

To help alleviate some of the reviewing I'm going to leave a few comments, so standby for that


resources/assets/components/Chrome.js
```diff
@@ -1,28 +1,27 @@
...
 import TabbedNavigationContainer from '../containers/TabbedNavigationContainer';
-import AffirmationContainer from '../containers/AffirmationContainer';
 import NotificationContainer from '../containers/NotificationContainer';
 
 // @TODO: Might be useful to have a separate file of constants?
 const LEDE_BANNER_QUAD_BUTTONS = 'lede_banner_quad_buttons';
 
 const Chrome = props => (
-  <div>
+  <div className={classnames({ '-lock': props.shouldShowModal })}>
```
deadlybutter: This is to lock the background so only the modal scrolls.

resources/assets/components/Modal/Modal.js
```diff
@@ -0,0 +1,74 @@
...
+    photo: PropTypes.string,
+    additionalContent: PropTypes.shape({
+      author: PropTypes.string.isRequired,
+      jobTitle: PropTypes.string.isRequired,
+      avatar: PropTypes.string.isRequired,
+    }).isRequired,
+  }).isRequired,
+};
+
+Modal.defaultProps = {
```
deadlybutter: You can see here I'm trying a new approach with component/container relationships. More background on the approach here https://github.com/DoSomething/phoenix-next/pull/285#discussion_r120394349

resources/assets/components/Modal/Modal.js
```diff
@@ -0,0 +1,74 @@
...
+  competitionStep: {
+    photo: null,
+  },
+};
+
+Modal.mapStateToProps = state => ({
```
deadlybutter: You can see here I'm trying a new approach with component/container relationships. More background on the approach here https://github.com/DoSomething/phoenix-next/pull/285#discussion_r120394349

resources/assets/components/Modal/RenderModalInBody.js
```diff
@@ -0,0 +1,58 @@
...
+import PropTypes from 'prop-types';
+import { Provider } from 'react-redux';
+
+/**
+ * Renders the modal outside the normal React tree so scroll behavior is correct.
+ * Referenced from here: http://jamesknelson.com/rendering-react-components-to-the-document-body/
+ * Did some renaming and some of the function usage was deprecated.
+ */
+
+class RenderModalInBody extends React.Component {
```
deadlybutter: This approach was inspired originally by a React modal package Dave sent me, along with referencing an old blog post listed above in the comment. The critical change/difference here is we're wrapping everything in a new Provider so Redux continues to work.

resources/assets/components/Modal/modal.scss
```diff
@@ -0,0 +1,39 @@
...
+    // Override any child components
+    > div {
+      margin: 0;
+    }
+  }
+
+  .modal__exit {
+    position: absolute;
+    top: $base-spacing;
+    left: $base-spacing; // TODO: Move this to the right.
```
deadlybutter: This TODO will be completed when we move the Header/Footer out of the php template to React (second PR later)

resources/assets/components/Notification/notification.scss
```diff
@@ -1,8 +1,19 @@
 @import '../../scss/next-toolbox.scss';
 
+.notification-list {
```
deadlybutter: Here's what these look like now.

screen shot 2017-06-08 at 12 19 07 pm

I get into why this changed below with my comment on the Overlay

resources/assets/components/Overlay/Overlay.js
```diff
@@ -1,20 +0,0 @@
-import PropTypes from 'prop-types';
```
deadlybutter: I removed this component because it was overcomplicating the layout, and the modal ended up not being used here anyway.

resources/assets/init.js
```diff
@@ -41,6 +41,7 @@ import { observerMiddleware } from './middleware/analytics';
 const history = historyInit();
 const middleware = [thunk, routerMiddleware(history), observerMiddleware];
 const store = configureStore({ ...reducers, routing: routerReducer }, middleware, window.STATE);
+window.reduxStore = store; // This doesn't feel right, importing the init.js also felt weird...
```
deadlybutter: Here is the larger conversation I had with Dave about attaching the store to the window like this,

screen shot 2017-06-08 at 12 21 16 pm

deadlybutter:
@DFurnes @weerd Alrighty this is ready for a look 👁 -- I tried to leave a lot of context since it's such a large change.

resources/assets/components/Affirmation/affirmation.scss
```diff
@@ -3,8 +3,8 @@
 .affirmation {
   width: 100%;
   padding: $base-spacing;
-  background: $dark-background-color;
-  color: $white;
+  color: $primary-text-color;
+  background: $light-background-color;
```
weerd: small thing. If we're only setting the background color here, than use the `background-color` property otherwise it defaults the rest of the background declarations and makes it harder to override. Not a huge deal in general, but want to get into this practice as a team 😃

deadlybutter: Yeah I ran into that somewhere else the other day, was throwing me off for sure. I'll update that now!

resources/assets/components/Modal/Modal.js
```diff
@@ -0,0 +1,74 @@
...
+
+  /* eslint-disable jsx-a11y/no-static-element-interactions */
+  return (
+    <RenderModalInBody shouldShowModal={shouldShowModal}>
+      <div id={MODAL_ID} className="modal" onClick={onBackgroundClick}>
+        <div className="modal__container">
+          <div className="modal__slide">
+            <AffirmationContainer />
+          </div>
+          <div className="modal__slide">
```
weerd: If there isn't a competition on a campaign, then does this output an empty `<div>`? I'm guessing because above we're not pulling these components dynamically yet...

deadlybutter: Good point- this will probably error out since it required the competition content. I will add a thing to account for this

resources/assets/components/Notification/notification.scss
```diff
@@ -1,8 +1,19 @@
...
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  width: 25%;
+  min-width: 300px;
+  display: flex;
```
weerd: Do we really need `flex` for these? Isn't it just a stacked list of `notifications`? Feels like overkill.

deadlybutter: I've been getting used to Flex'ing so I guess it doesn't feel overkill to me anymore

resources/assets/containers/AffirmationContainer.js
```diff
@@ -1,22 +1,18 @@
...
 });
 
 /**
  * Provide pre-bound functions that allow the component to dispatch
  * actions to the Redux store as props for this component.
  */
-const actionCreators = {
-  hideAffirmation,
-};
+const actionCreators = { /* ... */ };
```
weerd: Should we just remove if there are no dispatch methods to bind?

deadlybutter: 🤔  Can we pass just the mapStateToProps to `connect()`? If so then fine by me!

resources/assets/components/Notification/notification.scss
DFurnes: I'd say not to use flexbox unless you're doing something that requires it, since we've run into a fair number of cross-browser quirks.

weerd: Until there are issues, haha. 😉

resources/assets/containers/AffirmationContainer.js
DFurnes: Yeah, that's one of the reasons I'm wary of building our own abstractions over this. Here's a link to the [API documentation](https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options) if you're interested.

deadlybutter: @DFurnes This is the generic container file we've been using for everything else, the component=container idea was only on the modal!

weerd:
> removing flexbox :(

LOL. it'll be ok. there's still plenty of flexbox happening on the page 😉

PR #302, opened 2017-06-09
Changing up the background test

deadlybutter:
Demo: https://gfycat.com/DeliciousYoungAmericanindianhorse

(Or visit `/us/campaigns/sincerely-us?sixpack-force-pitch_page_connected=pitch` in the review app)

**Theory** -- We might be able to hook people not convinced by the background page by showing them the community, action, etc tabs.

This PR starts a new test and updates the layout, it additionally does some small fixes
- halfContainer height is now actually half height
- the scroll hint doesn't block the conversion button

### DIFF
--- a/resources/assets/components/Navigation/TabbedNavigation.js
+++ b/resources/assets/components/Navigation/TabbedNavigation.js
@@ -35,6 +35,10 @@ class TabbedNavigation extends React.Component {
     window.addEventListener('scroll', this.onScroll, false);
   }
 
+  componentWillUnmount() {
+    window.removeEventListener('scroll', this.onScroll);
+  }
+
   /**
    * Method called when the page is scrolled.
    */
@@ -62,8 +66,6 @@ class TabbedNavigation extends React.Component {
    */
   updateState() {
     this.isAnimatingFrame = false;
-
-    if (! this.node) return;
 
     const tabbedNavRect = this.node.getBoundingClientRect();
 
--- a/resources/assets/components/Navigation/TabbedNavigation.js
+++ b/resources/assets/components/Navigation/TabbedNavigation.js
@@ -63,6 +63,8 @@ class TabbedNavigation extends React.Component {
   updateState() {
     this.isAnimatingFrame = false;
 
+    if (! this.node) return;
+
     const tabbedNavRect = this.node.getBoundingClientRect();
 
     this.setState({ isStuck: tabbedNavRect.top <= 0 });
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -39,7 +39,7 @@ export function transformState(action, state) {
       ...state.user,
     },
     experiments: {
-      tests: Object.keys(state.experiments),
+      tests: state.experiments ? Object.keys(state.experiments) : [],
     },
     routing: {
       referer: document.referrer,
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -39,8 +39,7 @@ export function transformState(action, state) {
       ...state.user,
     },
     experiments: {
-      tests: state.experiments ? Object.keys(state.experiments) : [],
-      alternatives: state.experiments ? Object.values(state.experiments) : [],
+      tests: Object.keys(state.experiments),
     },
     routing: {
       referer: document.referrer,
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -40,6 +40,7 @@ export function transformState(action, state) {
     },
     experiments: {
       tests: state.experiments ? Object.keys(state.experiments) : [],
+      alternatives: state.experiments ? Object.values(state.experiments) : [],
     },
     routing: {
       referer: document.referrer,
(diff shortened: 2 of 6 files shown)


resources/assets/components/Navigation/TabbedNavigation.js
```diff
@@ -63,6 +63,8 @@ class TabbedNavigation extends React.Component {
   updateState() {
     this.isAnimatingFrame = false;
 
+    if (! this.node) return;
```
DFurnes: Is this to fix the scroll errors? If so, a better solution would be to remove the event listener on unmount.

resources/assets/helpers/analytics.js
```diff
@@ -40,7 +40,6 @@ export function transformState(action, state) {
     },
     experiments: {
       tests: state.experiments ? Object.keys(state.experiments) : [],
-      alternatives: state.experiments ? Object.values(state.experiments) : [],
```
DFurnes: Did you mean to remove this?

deadlybutter: Yep, I realized we're not storing these anymore in the client (from what I can tell, @weerd feel free to weigh in if that's intentional I guess?)

resources/assets/components/Navigation/TabbedNavigation.js
deadlybutter: @DFurnes 👍 I can do that now

weerd:
@deadlybutter if clicking on the "Take a Stand" or "Get Started" buttons on the pitch background, where do they take you?

deadlybutter:
@weerd Originally I was thinking they should go to the community page, but Luke pointed out we'd then have a triple conversion funnel, so they currently point to Northstar

resources/assets/components/Navigation/TabbedNavigation.js
weerd: What's this error?

deadlybutter:
Also, I updated the tabbed navigation to use the life cycle function, seems like everything still works as intended

resources/assets/components/Navigation/TabbedNavigation.js
```diff
@@ -35,6 +35,10 @@ class TabbedNavigation extends React.Component {
     window.addEventListener('scroll', this.onScroll, false);
   }
 
+  componentWillUnmount() {
+    window.removeEventListener('scroll', this.onScroll);
+  }
+
```
weerd: Just curious what error you were seeing?

deadlybutter: @weerd https://dosomething.slack.com/archives/C3ASB4204/p1496936970281685

weerd: Oh meant more how/why you were seeing it? Steps to recreate :)

deadlybutter: Oh just opening the page with the pitch alternative would spam the console with that

deadlybutter: I'm guesing it has something to do with the fact its not being rendered first like it normally is, and something with the DOM isn't there yet?

resources/assets/helpers/analytics.js
weerd: Not sure what you mean here. We store the experiment name and selected alternative for a user in the Redux store. Do we not want to send the selected alt to Keen? I dunno.

deadlybutter: Oh I'm an idiot, I was looking at the first action which doesn't include the alternative.

screen shot 2017-06-12 at 10 57 57 am

but there is another one after which does add the alternative

resources/assets/components/Navigation/TabbedNavigation.js
DFurnes: Yeah, I don't think this is an issue anymore since we keep the rest of the page at the bottom but it _would've_ been before when the tabbed navigation was first mounted on load and then removed from the DOM when the user got opted in to the "pitch" experiment.

deadlybutter: @DFurnes Nope, this is definitely happening when its all on the same page :/

weerd: Hmmm, that's weird. So it mounts but doesn't really, so it throws those errors cause it can't find the tabbed nav?

weerd: > Yeah, I don't think this is an issue anymore since we keep the rest of the page at the bottom but it would've been before when the tabbed navigation was first mounted on load and then removed from the DOM when the user got opted in to the "pitch" experiment.

Yeah, this makes sense to me as to why it would happen. But now with everything else at the bottom it should still be mounted... 🤔

deadlybutter: That's my guess, the virtual dom makes it but the real dom doesn't (yet)

DFurnes: @deadlybutter Really? Huh. It'd be interesting to figure out why it's getting unmounted here, but regardless I think having the `componentWillUnmount` handler is a good thing™. 👌

deadlybutter:
Oh in re-adding the alternative I also added the package lock file, @DFurnes you said you were ok with that though right?

DFurnes:
@deadlybutter Yes! I'm on board. 👍

PR #303, opened 2017-06-12
Fix hamburger in modal & some refactoring.

DFurnes:
#### Changes
This pull request includes a few fixes to the modal implementation:

🌀  Uses [react-portal](https://github.com/tajo/react-portal) instead of our own custom implementation. (Redux store is passed down through context so there's no need to add another provider here.) 44fbf2b

🎨  Uses `button` element for modal's close, rather than a clickable `div`. Adds default styles to our base stylesheet so these don't look wonky. 8949808

✏️  Updates to the latest version of our ESLint config, which pulls in upstream updates & merges in the custom overrides we had made here. 14b6de7

⬆️  Pushes the modal above the hamburger navigation like we do with Forge's old modals. 1409518


resources/assets/components/Modal/modal.scss
```diff
@@ -8,6 +8,7 @@
   top: 0;
   left: 0;
   background-color: $dark-background-color-transparent;
+  z-index: 9999;
```
deadlybutter: Something I mentioned to Diego I think was adding all of our z-indexes in one place (eg `next-toolbox`) so its easy to visually see what the layers are

DFurnes: Yeah, that's a good idea! 👍

package.json
```diff
@@ -52,7 +53,7 @@
   },
   "devDependencies": {
     "@dosomething/babel-config": "^1.0",
-    "@dosomething/eslint-config": "^2.0.0",
+    "@dosomething/eslint-config": "^3.0.0",
```
weerd: 👏

resources/assets/components/Modal/Modal.js
```diff
@@ -1,48 +1,52 @@
...
+    super();
 
-const MODAL_ID = 'modal';
+    this.handleOverlayClick = this.handleOverlayClick.bind(this);
+  }
 
-const Modal = (props) => {
-  const { shouldShowModal, competitionStep } = props;
+  handleOverlayClick(event) {
+    if (event.target !== this.node) return;
```
weerd: Just curious, we've done this a bit in this codebase, but I thought we ideally do no single-line if conditionals, preferring to use brackets, etc. I don't remember well enough tho. There's no eslint rule that calls this out, but I thought it was our preference in the past. 🤷‍♂️

DFurnes: Yeah, I'm up for doing that.

PR #304, opened 2017-06-12
Sticky nav now with a CTA button.

weerd:
This PR adds a new `<Button />` component that is then used in the new sticky nav as a CTA button that animates in when the sticky nav is activated.

screen recording 2017-06-12 at 04 45 pm


resources/assets/components/Button/Button.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import './button.scss';
+
+const Button = ({
+  callback,
+  callbackArgument,
```
DFurnes: I'm a little curious why these are separate props instead of just passing a single `onClick` function prop. It feels like we're locking ourselves into a very specific usage scenario.

package.json
```diff
@@ -44,6 +44,7 @@
     "react-router": "^4.1.1",
     "react-router-dom": "^4.0.0-beta.6",
     "react-router-redux": "5.0.0-alpha.6",
+    "react-test-renderer": "^15.5.4",
```
DFurnes: Was this intentionally moved out of dev dependencies?

weerd: Yeah, it was recommended in the ESLint based on the new tests I wrote for the `<Button />` component where I didn't use `enzyme` and `shallow` rendering since it doesn't have child components.

resources/assets/components/Button/Button.js
weerd: Hmmm, so are you saying the `callbackArgument` and `callbackMetadata` should be referenced all within the callback that's passed as the `onClick`?

DFurnes: Right. This is locking us into a scenario where the on-click handler always has exactly two arguments. Is there an advantage over just passing a single `onClick` prop that can have any format? For example:

```js
<Button classNames="-modifier" onClick={() => handler(1234, { source: 'jest test' })} />
```

weerd: Yeah, I think that makes sense to me and would be much more versatile. 👍

resources/assets/components/Button/Button.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import './button.scss';
+
+const Button = ({ onClick, classNames, text }) => {
+  const handleOnClick = () => {
```
DFurnes: Is this `handleOnClick` still necessary?

weerd: yeah, probs not 😬

PR #307, opened 2017-06-13
Adding github templates.

weerd:
### What does this PR do?
This PR adds some templates for creating issues and pull-requests in Github. It also adds a contributing file, but we should discuss this further and see what it should contain.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,3 +1,4 @@
 @TODO: Add contributing guidelines here...
 
 <https://help.github.com/articles/setting-guidelines-for-repository-contributors/>
+
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,3 @@
+@TODO: Add contributing guidelines here...
+
+<https://help.github.com/articles/setting-guidelines-for-repository-contributors/>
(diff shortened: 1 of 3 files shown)


weerd:
I popped these in a .github folder so it keeps the root directory a bit more organized 💃

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,4 @@
+@TODO: Add contributing guidelines here...
```
DFurnes: Do we need this?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
+### What does this PR do?
+...
+
+
+### Any background context you want to provide?
+...
+
+
+### What are the relevant tickets/cards?
+Fixes #
```
DFurnes: Should we remove this since we don't generally have issues for work?

weerd: I added `/cards` to help remind us to associate a trello card to the PR if we're not using a ticket. Maybe just different wording would suffice?

.github/CONTRIBUTING.md
weerd: 🤷‍♂️

.github/PULL_REQUEST_TEMPLATE.md
DFurnes: Okay, just assuming it's something we'll be deleting more often than not.

weerd: We should be popping links to the trello cards though. I often forget and this'll be a friendly reminder.

PR #311, opened 2017-06-15
improving x button

deadlybutter:
screen shot 2017-06-15 at 11 41 44 am
screen shot 2017-06-15 at 11 41 24 am

### DIFF
--- a/resources/assets/components/Modal/modal.scss
+++ b/resources/assets/components/Modal/modal.scss
@@ -30,8 +30,8 @@
 
   .modal__exit {
     position: absolute;
-    top: 2%;
-    right: 4%;
+    top: $half-spacing;
+    right: $half-spacing;
     font-size: 34px;
     line-height: 34px;
     color: $white;
@@ -41,7 +41,7 @@
     }
 
     @include media($medium) {
-      right: 2%;
+      right: $base-spacing;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


resources/assets/components/Modal/modal.scss
```diff
@@ -30,11 +30,18 @@
 
   .modal__exit {
     position: absolute;
-    top: $base-spacing;
-    left: $base-spacing; // TODO: Move this to the right.
+    top: 2%;
```
weerd: Just curious but why `%` for the `top` and `right` and not a fixed value like `$half-spacing` or `$base-spacing`?

deadlybutter: @weerd originally i did percents to solve it looking weird on mobile, but i ended up adding that media query so i could go back to `$base-spacing`

deadlybutter:
@weerd updated w/ static values, looks fine for me still!

resources/assets/components/Modal/modal.scss
weerd: Yeah, I'd rather stay to a know value and be more consistent w/ the spacing across the page :)

ngjo:
[🐛 Move the X into the modal 🐛](https://trello.com/c/kIonVAcS/697-%F0%9F%90%9B-move-the-x-into-the-modal-%F0%9F%90%9B)

PR #319, opened 2017-06-26
Sticky cta

weerd:
### What does this PR do?
This PR adds a sticky CTA for small screens. I considered a few different approaches for this feature. Eventually I had thought it would be a good idea to repurpose the `Revealer` component to also behave as the sticky CTA on small screens, since it contained a CTA inside of it. However, this proved to be difficult in the end because the sticky positioned element is only sticky relative to the its immediate parent element. I had considered that maybe if the parent had no `position` defined, then it would work up the hierarchy to the next ancestor element, similar to how `position: absolute` works (this turned out _not_ to be the case).

sticky cta

While this PR completes the feature, I started a branch with a potential alternative where the `Revealer` does not contain a CTA so it's more set to a single purpose and simplified. This could potentially allow for then adding a separate CTA below the Revealer which on small screens behaves as a sticky CTA.

Additionally, I started the work for refactoring the CTAs. It felt odd that you couldn't easily output a simple CTA as a default. So now there is both a `CallToAction` and `CallToActionBlock` component, with the intent that the block version will be refactored into the `CallToAction`, so it's not specific to just the feed.

This was a bit of a whirlwind of my own doing 😥 

Fixes https://trello.com/c/Hdz3Igkd/367-3-as-a-small-screen-sized-user-i-want-a-cta-prompt-to-pop-up-from-the-bottom-of-the-screen-and-stay-there

### Checklist


.env.example
```diff
@@ -25,7 +25,6 @@ GLADIATOR_API_KEY=
 
 NORTHSTAR_URL=
 NORTHSTAR_CLIENT_ID=
-NORTHSTAR_CLIENT_SECRET=
```
deadlybutter: @weerd 😬

resources/assets/containers/CallToActionContainer.js
```diff
@@ -1,27 +1,23 @@
 import { connect } from 'react-redux';
-import { get } from 'lodash';
-import CallToActionBlock from '../components/CallToActionBlock';
-import { clickedSignUp, convertExperiment } from '../actions';
+import CallToAction from '../components/CallToAction/CallToAction';
```
deadlybutter: If you were trying to adopt the pattern Dave started under the Block/ directory then this should technically change- 

The "CallToAction" file is fine, but it should be exported as the default in the index.js file, along with the container file. This way you don't have to type `CallToAction` twice here and it's all in the same place.

.env.example
weerd: We don't use this one anymore.

deadlybutter: Oh right we use the authorization one now

resources/assets/containers/CallToActionContainer.js
weerd: Oh right. I forgot that got changed. Ok, I can update.

resources/assets/components/CallToAction/CallToAction.js
```diff
@@ -0,0 +1,26 @@
...
+
+const CallToAction = ({ clickedSignUp, legacyCampaignId, classNames = '' }: CallToActionProps) => (
+  <div className={classnames('call-to-action', classNames)}>
+    <Button onClick={() => clickedSignUp(legacyCampaignId, { source: 'small screen sticky cta|text: Join us' })} />
+  </div>
+);
+
+export default CallToAction;
+
+// @TODO: ESLint kept yelling at me that I needed to define this somewhere and does not
```
DFurnes: Hmm, yeah looks like this is [a known thing](https://git.io/vQGEy).

weerd: from this [comment](https://github.com/yannickcr/eslint-plugin-react/issues/1184#issuecomment-300635989) should I remove my `@TODO` and leave as is?

DFurnes: Yeah, I think that's a fine call.

weerd:
@DFurnes @deadlybutter made a quick fix along with the folder restructure so that I don't accidentally hide the `view more` button when I hide the CTA button on small screen to show the sticky CTA!

resources/assets/components/Block/Block.js
```diff
@@ -5,7 +5,7 @@ import CampaignUpdateBlock from '../CampaignUpdateBlock';
 import PlaceholderBlock from '../PlaceholderBlock';
 import ReportbackBlock from '../ReportbackBlock';
 import StaticBlock from '../StaticBlock';
-import CallToActionContainer from '../../containers/CallToActionContainer';
+import CallToActionBlockContainer from '../../containers/CallToActionBlockContainer';
```
deadlybutter: Old import ⚠️

resources/assets/components/Block/Block.test.js
```diff
@@ -4,7 +4,7 @@ import Block from './Block';
 
 // Mock Redux containers so we don't need Provider context.
 jest.mock('./BlockContainer', () => 'BlockContainer');
-jest.mock('../../containers/CallToActionContainer', () => 'CallToActionContainer');
+jest.mock('../../containers/CallToActionBlockContainer', () => 'CallToActionBlockContainer');
```
deadlybutter: Also old import here ⚠️

resources/assets/components/ContentPage/index.js
```diff
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import Markdown from '../Markdown';
 import ScrollConcierge from '../ScrollConcierge';
-import CallToActionContainer from '../../containers/CallToActionContainer';
+import CallToActionBlockContainer from '../../containers/CallToActionBlockContainer';
```
deadlybutter: Old import here as well ⚠️

resources/assets/components/Block/Block.js
weerd: Nope! those are for the CTA Block and not the CTA! Currently there are 2 until I refactor and mash them together.

ngjo:
[(3) As a small screen sized user, I want a CTA prompt to pop up from the bottom of the screen and stay there](https://trello.com/c/Hdz3Igkd/367-3-as-a-small-screen-sized-user-i-want-a-cta-prompt-to-pop-up-from-the-bottom-of-the-screen-and-stay-there)

PR #320, opened 2017-06-27
Action page refactoring, some cleanup

deadlybutter:
### What does this PR do?
- Refactoring action page a bit (mostly file structure)
- Action page improvements (RB Uploader is now a custom action step, I could see Blocks being the next thing we add). Action page also hides when the campaign closes.
- Removed the pitch experiment stuff. Was in the routing logic so figured why not. (**That's what the bulk of the files are**) 
- Fixed a bug in the modal where the children weren't being assigned a key


deadlybutter:
Weird the build error here is totally valid but not getting it locally 🤔  -- oh well easy fix

resources/assets/components/ActionPage/ActionHelpers.js
```diff
@@ -0,0 +1,93 @@
...
+import Revealer from '../Revealer';
+import { FlexCell } from '../Flex';
+import { makeHash } from '../../helpers';
+import CompetitionContainer from '../../containers/CompetitionContainer';
+import ReportbackUploaderContainer from '../../containers/ReportbackUploaderContainer';
+
+/**
+ * Render a photo on the action page.
+ */
+export const renderPhoto = (photo, index) => (
```
DFurnes: Why is this a helper method? We should make it into a component if it's used in more than one place.

resources/assets/components/ActionPage/ActionHelpers.js
```diff
@@ -0,0 +1,93 @@
...
+  <div className="action-step__photo" key={index}>
+    <img src={photo} alt="action step example" />
+  </div>
+);
+
+/**
+ * Render a single step on the action page.
+ */
+export const renderSteps = (steps, props) => {
```
DFurnes: Same here.

resources/assets/components/ActionPage/ActionHelpers.js
```diff
@@ -0,0 +1,93 @@
...
+      isAuthenticated={isAuthenticated}
+    />
+  );
+
+  const RevealerOrUploader = isAuthenticated ? Uploader : ActionRevealer;
+
+  let stepIndex = 0;
+  let appendUploader = true; // TODO: Remove this after content updates.
+
+  const map = steps.map((step) => {
```
DFurnes: Can we name this something more descriptive? Perhaps `stepComponents`?

resources/assets/components/ActionPage/ActionPage.js
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <Flex>
+      { renderSteps(actionSteps, props) }
+    </Flex>
+  );
+};
+
+ActionPage.propTypes = {
+  steps: PropTypes.array.isRequired, // eslint-disable-line react/forbid-prop-types
+  /* eslint-disable  react/no-unused-prop-types */
```
DFurnes: I think this is a good indication we're doing something awkward here.

resources/assets/components/App.js
```diff
@@ -7,68 +7,42 @@ import { initializeStore } from '../store';
...
   <Container><Component {...props} /></Container>
 );
 
-const experience = component => wrap(Experience, component);
 const chrome = component => wrap(ChromeContainer, component);
 
 const App = ({ store, history }) => {
   initializeStore(store);
 
-  const PitchTest = (props) => {
```
DFurnes: 🔥

resources/assets/containers/TabbedNavigationContainer.js
```diff
@@ -34,7 +40,7 @@ const TabbedNavigationContainer = (props) => {
     <TabbedNavigation>
       <div className="nav-items">
         <NavigationLink to={paths.community} exact>Community</NavigationLink>
-        <NavigationLink to={paths.action}>Action</NavigationLink>
+        { ActionLink }
```
DFurnes: I'd be tempted to negate the boolean and inline this in both cases (here & the app routing):

```js
<div className="nav-items">
  <NavigationLink to={paths.community} exact>Community</NavigationLink>
  {isOpen ? <NavigationLink to={paths.action}>Action</NavigationLink> : null }
  { additionalPages }
</div>
```

resources/assets/store.js
```diff
@@ -91,6 +92,9 @@ export function configureStore(reducers, middleware, preloadedState = {}) {
   // If React DevTools are available, use instrumented compose function.
   const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose; // eslint-disable-line
 
+  // Set initial variables based on the data given to us.
+  initialState.campaign.isClosed = new Date(preloadedState.campaign.endDate.date) - new Date() < 0;
```
DFurnes: I don't think this is a good idea - we shouldn't be storing derived data in store. Can we make a selector to grab this and calculate from the `campaign.endDate.date` instead?

resources/assets/components/ActionPage/ActionHelpers.js
```diff
@@ -0,0 +1,93 @@
...
+  </div>
+);
+
+/**
+ * Render a single step on the action page.
+ */
+export const renderSteps = (steps, props) => {
+  const { isAuthenticated, clickedSignUp } = props;
+
+  const Uploader = (
```
DFurnes: Variables should  be `camelCase` unless they're class constructors.

resources/assets/components/ActionPage/ActionHelpers.js
```diff
@@ -0,0 +1,93 @@
...
+    <FlexCell key="reportback_uploader" width="full">
+      <ReportbackUploaderContainer />
+    </FlexCell>
+  );
+
+  const ActionRevealer = (
```
DFurnes: 🐫

resources/assets/components/ActionPage/ActionHelpers.js
```diff
@@ -0,0 +1,93 @@
...
+      key="revealer"
+      title="Join Us"
+      callToAction={props.callToAction}
+      isLoading={props.hasPendingSignup}
+      onReveal={() => clickedSignUp(props.campaignId, { source: 'action page revealer' })}
+      isAuthenticated={isAuthenticated}
+    />
+  );
+
+  const RevealerOrUploader = isAuthenticated ? Uploader : ActionRevealer;
```
DFurnes: 🐫

resources/assets/components/ActionPage/ActionHelpers.js
deadlybutter: They're technically only used in one place, I just decided to move all the custom map functions to one place. Maybe just remove `helpers` from the file name and call it something else?

resources/assets/components/ActionPage/ActionPage.js
deadlybutter: Let me see if I can re-work this a bit

resources/assets/containers/TabbedNavigationContainer.js
deadlybutter: That's what I originally did in the `App.js` but the statement gets quite long, I think your suggestion looks better for the TabbedNavigation though 👍

resources/assets/components/ActionPage/ActionHelpers.js
deadlybutter: Renamed this file to `ActionRenders`, so the import now looks like this `import { renderSteps } from './ActionRenders';`. Let me know if that makes more sense.

resources/assets/helpers/index.js
```diff
@@ -320,3 +320,13 @@ export function getFormattedScreenSize(screenWidth = window.innerWidth) {
...
   return breakpoints.find(breakpoint => breakpoint.test(screenWidth)).name;
 }
+
+/**
+ * Check if the given campaign is closed.
+ *
+ * @param  {Object}  campaign
+ * @return {Boolean}
+ */
+export function isCampaignClosed(campaign) {
```
weerd: Debating if it's better that we just pass a date to this function instead of the entire `campaign` object.

```js
export function isCampaignClosed(campaignEndDate) {
  return new Date(campaignDate) - new Date() < 0;
}
```

Then in the tabbed nav container, we only need to grab the campaign date

```js
 ...
campaignEndDate: state.campaign.endDate.date,
...
```
Thoughts?

deadlybutter: Fine w/ me!

deadlybutter: @weerd Cool updated & retested the helper function to just be date based

resources/assets/containers/TabbedNavigationContainer.js
```diff
@@ -7,20 +7,24 @@ import NavigationLink from '../components/Navigation/NavigationLink';
...
 import { clickedSignUp } from '../actions';
 import { paths } from '../helpers/navigation';
+import { isCampaignClosed } from '../helpers';
 
 const mapStateToProps = state => ({
   isAffiliated: state.signups.thisCampaign,
   legacyCampaignId: state.campaign.legacyCampaignId,
   pages: state.campaign.pages,
   pathname: state.routing.location.pathname,
+  campaign: state.campaign,
```
weerd: I think also change this, so we're only specifically grabbing the data. If we ever need more from the campaign object then we can update later, but I think it'd be best to aim to always grab the minimum instead of grabbing everything in one shot. Easier to see what we actually _need_ from the campaign object:

```js
campaignEndDate: state.campaignendDate.date,
````

deadlybutter: Oh yeah sorry should have updated that one as well. _ALSO_ It's technically good for performance, because it will only re-render the components who are subscribed to the endDate vs the entire campaign (I think)

deadlybutter: Fixed, much better! 👍

deadlybutter:
giphy 15

ngjo:
[(5) As a campaign lead, I want Phoenix-next to have a campaign closed state](https://trello.com/c/imNd1bs4/701-5-as-a-campaign-lead-i-want-phoenix-next-to-have-a-campaign-closed-state)

PR #324, opened 2017-06-29
Only redirect signups to action page if the campaign is open

deadlybutter:
### What does this PR do?
In order to ship the action page changes, we should also make sure people who are signing up on a closed campaign are not taken to the action page post signup. This change ensures they still get the modal experience, but the redirect is dependent on campaign status.

We need to close Sincerely Us like now

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -1,5 +1,6 @@
 import { Phoenix } from '@dosomething/gateway';
 import { push } from 'react-router-redux';
+import { isCampaignClosed } from '../helpers';
 import {
   SIGNUP_CREATED,
   SIGNUP_FOUND,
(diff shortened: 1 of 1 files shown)


deadlybutter:
Tested on Sincerely Us (Open) and Teens for Jeans (Closed)

resources/assets/actions/signup.js
```diff
@@ -128,8 +129,11 @@ export function clickedSignUp(campaignId, metadata) {
         // If Drupal denied our signup request, check if we already had a signup.
         dispatch(checkForSignup(campaignId));
       } else {
-        // Otherwise, mark the signup as a success & take user to the action page.
-        dispatch(push('/action'));
+        // Otherwise, mark the signup as a success and
+        // take user to the action page if campaign is open.
+        if (! isCampaignClosed(getState().campaign.endDate.date)) {
+          dispatch(push('/action'));
+        }
```
weerd: If the signup is a success and the campaign _is_ closed, where does the user end up getting taken?

deadlybutter: You'll stay on whatever page you signed up- We could enforce community if we really wanted but I wasn't given any direction on that

weerd: Gotcha. No worries, I just wanted to confirm :)

ngjo:
[(5) As a campaign lead, I want Phoenix-next to have a campaign closed state](https://trello.com/c/imNd1bs4/701-5-as-a-campaign-lead-i-want-phoenix-next-to-have-a-campaign-closed-state)

PR #330, opened 2017-07-07
Fixing a couple bugs quickly.

weerd:
### What does this PR do?
This PR fixes two quick bugs I ran into. 

First, now that the RB uploader can be added as a step on the action page, the component should not make the background required.

Second, the `addToSubmissionsList` action was renamed, but the `ReportbackUploaderContainer` did not update the references to the new name of `addSubmissionItemToList`.


deadlybutter:
> the component should not make the background required.

Weird, I thought I removed that. Oh, wait no, that was content. 👍 

Ill quick check the codez

PR #331, opened 2017-07-10
Reportback Submission gallery separation from RB Uploader Component

weerd:
### What does this PR do?
This PR continues the process for the updates and refactoring of the RB Uploader. It specifically works on separating out the Reportback Submission Gallery _out_ of the Reportback Uploader component, so they can be placed in different locations if desired.

image

This is what the gallery looks like outside of the RB uploader. Next steps will include creating a `Card` component that combines with the `Block` to give the RB submission items the correct visual appearance, along with also some updates on the Contentful side for how the gallery gets included into the Action Steps.

Tackling the RB Uploader refactor in digestible chunks. 

### Checklist

### DIFF
--- a/resources/assets/components/ActionPage/ActionStepsWrapper.js
+++ b/resources/assets/components/ActionPage/ActionStepsWrapper.js
@@ -8,15 +8,22 @@ import { Flex, FlexCell } from '../Flex';
 import { makeHash } from '../../helpers';
 import CompetitionContainer from '../../containers/CompetitionContainer';
 import { ReportbackUploaderContainer } from '../ReportbackUploader';
+import { SubmissionGalleryContainer } from '../Gallery';
 import { clickedSignUp as clickedSignUpAction } from '../../actions';
 
 const ActionStepsWrapper = (props) => {
-  const { callToAction, campaignId, hasPendingSignup,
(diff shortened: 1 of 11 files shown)


resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -56,9 +66,15 @@ const ActionStepsWrapper = (props) => {
         );
 
       case 'photo-uploader':
-        appendUploader = false; // TODO: Remove this flag after content updates post deploy.
+        appendPhotoUploader = false; // TODO: Remove this flag after contentful updates post deploy.
+
         return revealerOrUploader;
 
+      case 'submission-gallery':
+        appendSubmissionGallery = false; // TODO: Remove this flag after contentful updates post deploy.
```
weerd: I'm thinking we can create a new content type on Contentful that is a counterpart to the `CampaignActionStep` called `ComponentActionStep` when it's sole purpose is to declare inserting a particular component within the action steps reference object list.

weerd: Right now, we're repurposing the `CampaignActionStep` content type to do this and it comes with a bunch of baggage for other fields to add content to, which is gross.

PR #333, opened 2017-07-11
Twerking the action steps

weerd:
### What does this PR do?
This PR makes a few updates to allow the use of _both_ `CampaignActionSteps` and `ComponentActionSteps`. The latter are simpler content types for the Action Steps that can be used to easily define a specific component to insert into the flow of the steps. Currently the available component steps to include are the `photo-uploader` and the `submission-gallery`, but we now have a system to easily allow adding more which we will be tackling in the near future when adding instant action steps.

What's nice, is that for more robust steps, like the `default` ones, which include photos, a background image, etc., they can still have variation types within them, such as the `competition` type.

image

When enabling the `ComponentActionStep` it revealed a sort of bug where the campaign would no longer render because we _used_ to depend on each action step having a `photos` property, even if it wasn't used. The new content type doesn't have this field, thus the site would break. The updates to the `app/Entities/Campaign.php` specifically address this new bug and resolve it.

This initially started as part of the work for the RB Uploader refactoring, but ended up also benefitting for upcoming instant actions and so it was approach in a way that could afford adding instant action in the future much simpler 😄 

This also introduces a potential approach to per-campaign feature flags via the `additionalContent` :)

### Checklist

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -37,13 +37,9 @@ class Campaign extends Entity implements JsonSerializable
      */
     public function parseActionStepPhotos($photos)
     {
-        $output = [];
-
-        foreach ($photos as $photo) {
-            $output[] = get_image_url($photo, 'landscape');
-        }
-
-        return $output;
+        return collect($photos)->map(function ($photo) {
+            return get_image_url($photo, 'landscape');
+        });
     }
 
     /**
@@ -54,25 +50,20 @@ class Campaign extends Entity implements JsonSerializable
      */
     public function parseActionSteps($actionSteps)
     {
-        $output = [];
-
-        foreach ($actionSteps as $step) {
+        return collect($actionSteps)->map(function ($step) {
             $data = [];
 
             $data['title'] = $step->title;
-            $data['displayOptions'] = $step->displayOptions->shift();
+            $data['displayOptions'] = $step->displayOptions->first();
 
             $step->content ? $data['content'] = $step->content : null;
             $step->background ? $data['background'] = get_image_url($step->background, 'landscape') : null;
             $step->photos ? $data['photos'] = $this->parseActionStepPhotos($step->photos) : null;
-            $step->type ? $data['type'] = $step->type->shift() : null;
-            $step->customType ? $data['customType'] = $step->customType->shift() : null; // @TODO deprecate.
+            $step->customType ? $data['customType'] = $step->customType->first() : null;
             $step->additionalContent ? $data['additionalContent'] = $step->additionalContent : null;
 
-            $output[] = $data;
-        }
-
-        return $output;
+            return $data;
+        });
     }
 
     public function jsonSerialize()
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -29,26 +29,54 @@ class Campaign extends Entity implements JsonSerializable
         return $this->active;
     }
 
-    public function jsonSerialize()
+    /**
+     * Parse and extract photo data for action steps.
+     *
+     * @param  array $photos
+     * @return array
+     */
+    public function parseActionStepPhotos($photos)
     {
-        $actionSteps = [];
-        foreach ($this->actionSteps as $step) {
-            $photos = [];
-            foreach ($step->photos as $photo) {
-                $photos[] = get_image_url($photo, 'landscape');
-            }
+        $output = [];
 
-            $actionSteps[] = [
-                'title' => $step->title,
-                'content' => $step->content,
-                'displayOptions' => $step->displayOptions,
-                'background' => get_image_url($step->background, 'landscape'),
-                'photos' => $photos,
-                'customType' => $step->customType,
-                'additionalContent' => $step->additionalContent,
-            ];
+        foreach ($photos as $photo) {
+            $output[] = get_image_url($photo, 'landscape');
         }
 
+        return $output;
+    }
+
+    /**
+     * Parse and extract data for action steps.
+     *
+     * @param  array $actionSteps
+     * @return array
+     */
+    public function parseActionSteps($actionSteps)
+    {
+        $output = [];
+
+        foreach ($actionSteps as $step) {
+            $data = [];
+
+            $data['title'] = $step->title;
+            $data['displayOptions'] = $step->displayOptions->shift();
+
+            $step->content ? $data['content'] = $step->content : null;
+            $step->background ? $data['background'] = get_image_url($step->background, 'landscape') : null;
+            $step->photos ? $data['photos'] = $this->parseActionStepPhotos($step->photos) : null;
+            $step->type ? $data['type'] = $step->type->shift() : null;
+            $step->customType ? $data['customType'] = $step->customType->shift() : null; // @TODO deprecate.
+            $step->additionalContent ? $data['additionalContent'] = $step->additionalContent : null;
+
+            $output[] = $data;
+        }
+
+        return $output;
+    }
+
+    public function jsonSerialize()
+    {
         return [
             'id' => $this->entry->getId(),
             'legacyCampaignId' => $this->legacyCampaignId,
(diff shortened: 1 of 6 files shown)


app/Entities/Campaign.php
```diff
@@ -29,26 +29,54 @@ public function isActive()
...
-                'photos' => $photos,
-                'customType' => $step->customType,
-                'additionalContent' => $step->additionalContent,
-            ];
+            $step->content ? $data['content'] = $step->content : null;
+            $step->background ? $data['background'] = get_image_url($step->background, 'landscape') : null;
+            $step->photos ? $data['photos'] = $this->parseActionStepPhotos($step->photos) : null;
+            $step->type ? $data['type'] = $step->type->shift() : null;
+            $step->customType ? $data['customType'] = $step->customType->shift() : null; // @TODO deprecate.
+            $step->additionalContent ? $data['additionalContent'] = $step->additionalContent : null;
```
weerd: While the above approach looks maybe a little weird, it works great! I preferred it over a bunch of `if () { ... }` statements. Lemme know if you know of a better way 😬

app/Entities/Campaign.php
```diff
@@ -29,26 +29,54 @@ public function isActive()
...
-    public function jsonSerialize()
+    /**
+     * Parse and extract photo data for action steps.
+     *
+     * @param  array $photos
+     * @return array
+     */
+    public function parseActionStepPhotos($photos)
+    {
+        $output = [];
```
DFurnes: Totes subjective, but this would be a nice opportunity to map over the array:

```php
    public function parseActionStepPhotos($photos)
    {
        return collect($photos)->map(function ($photo) {
            return get_image_url($photo, 'landscape');
        });
    }
```

app/Entities/Campaign.php
```diff
@@ -29,26 +29,54 @@ public function isActive()
...
     {
-        $actionSteps = [];
-        foreach ($this->actionSteps as $step) {
-            $photos = [];
-            foreach ($step->photos as $photo) {
-                $photos[] = get_image_url($photo, 'landscape');
-            }
+        $output = [];
+
+        foreach ($actionSteps as $step) {
```
DFurnes: Same here, if you think it aids readability.

app/Entities/Campaign.php
```diff
@@ -29,26 +29,54 @@ public function isActive()
...
+            $data = [];
+
+            $data['title'] = $step->title;
+            $data['displayOptions'] = $step->displayOptions->shift();
```
DFurnes: Would `->first()` be more readable here since we don't need to mutate the original array?

app/Entities/Campaign.php
```diff
@@ -29,26 +29,54 @@ public function isActive()
...
-                'displayOptions' => $step->displayOptions,
-                'background' => get_image_url($step->background, 'landscape'),
-                'photos' => $photos,
-                'customType' => $step->customType,
-                'additionalContent' => $step->additionalContent,
-            ];
+            $step->content ? $data['content'] = $step->content : null;
+            $step->background ? $data['background'] = get_image_url($step->background, 'landscape') : null;
+            $step->photos ? $data['photos'] = $this->parseActionStepPhotos($step->photos) : null;
+            $step->type ? $data['type'] = $step->type->shift() : null;
```
DFurnes: Same here?

app/Entities/Campaign.php
```diff
@@ -29,26 +29,54 @@ public function isActive()
...
+            $step->customType ? $data['customType'] = $step->customType->shift() : null; // @TODO deprecate.
```
DFurnes: I think I'm still confused why this is being removed, can we talk it over again IRL?

app/Entities/Campaign.php
deadlybutter: I totally forgot PHP had mapping like this 💡

app/Entities/Campaign.php
DFurnes: Oh woah, I don't think I read this correctly at first. Is there a reason we can't pull the variable it's being assigned to outside of the ternary?

weerd: Well, part of the idea is not assigning anything if there's nothing in the property. If I pull it out then the variable would always exist and we'd have a `$data['photos']` set to `null` on an action step that would never have photos since the field doesn't exist on the `ComponentActionStep`. It's partly why I made the initial comment... ;)

deadlybutter: Not sure I follow, do you mean to put something like this?

```
$additionalContent = $step->additionalContent;
if ($additionalContent) {
   $data['additionalContent'] = $additionalContent;
}
```

app/Entities/Campaign.php
weerd: Yeah, I suspected that at initially and had intended to use `->first()` but when checking the [documentation](https://laravel.com/docs/5.4/collections#method-first) I realized that it doesn't do exactly what you would suspect it does...

app/Entities/Campaign.php
weerd: Oh I'm favoring just `type` over `customType`. Just a preferred naming thing. In my opinion, custom type might confuse a bit, whereas type can apply and be interpreted within the context of the action step it is used in. I've updated the Contenful models to have `type` available in both `CampaignActionStep` and `ComponentActionStep`.

app/Entities/Campaign.php
weerd: Defs. I forget to use `.map()` more 😬  It's so useful! I will update 💃

app/Entities/Campaign.php
deadlybutter: Yeah the customType was a naming thing I invented because in the Old World ™️ you were hacking the Action Step to be something else

app/Entities/Campaign.php
DFurnes: I think I'm missing something, is the second example not what we're going for?

weerd: ha! I didn't read far enough in the doc... 🤦‍♂️

weerd:
@DFurnes @deadlybutter  made the updates to the `app/Entities/Campaign.php` to be a bit cleaner/clearer, as well as popping out the `ActionStepsWrapper` redux connect stuff into a dedicated container.

resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -111,19 +111,21 @@ ActionStepsWrapper.defaultProps = {
...
 };
 
-ActionStepsWrapper.mapStateToProps = state => ({
-  campaignId: state.campaign.legacyCampaignId,
-  callToAction: state.campaign.callToAction,
-  hasPendingSignup: state.signups.isPending,
-  isSignedUp: state.signups.thisCampaign,
-  isAuthenticated: state.user.id !== null,
-});
+// ActionStepsWrapper.mapStateToProps = state => ({
```
deadlybutter: @weerd Did you intend to leave this commented?

weerd: Doh! nope 😬

resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
-import { connect } from 'react-redux';
+// import { connect } from 'react-redux';
```
DFurnes: 👾

resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -9,7 +10,7 @@ import { makeHash } from '../../helpers';
 import CompetitionContainer from '../../containers/CompetitionContainer';
 import { ReportbackUploaderContainer } from '../ReportbackUploader';
 import { SubmissionGalleryContainer } from '../Gallery';
-import { clickedSignUp as clickedSignUpAction } from '../../actions';
+// import { clickedSignUp as clickedSignUpAction } from '../../actions';
```
DFurnes: 👾

resources/assets/components/ActionPage/ActionStepsWrapperContainer.js
```diff
@@ -0,0 +1,27 @@
+import { connect } from 'react-redux';
+import { get } from 'lodash';
+import ActionStepsWrapper from './ActionStepsWrapper';
+import { clickedSignUp as clickedSignUpAction } from '../../actions';
```
DFurnes: Any particular reason to rename this?

weerd: I kept this as it was in the `ActionStepsWrapper` originally, but doesn't need to be. Changing it up!

PR #336, opened 2017-07-13
Action image

deadlybutter:
### What does this PR do?
This pr fixes the images that display on the action page to be more responsive/square and cleans up some other layout bugs 

**Column 1/3** 
screen shot 2017-07-13 at 11 31 40 am

**Row full**
screen shot 2017-07-13 at 11 32 01 am

**Mobile (full width, column)**
screen shot 2017-07-13 at 11 31 50 am

I left a few comments on some of the specific css additions but otherwise fairly straight forward


weerd:
Just a thought that square crop on mobile could be a little large. Maybe we do two columns on mobile? Thoughts? cc: @lkpttn

resources/assets/components/ActionPage/ActionStep.js
```diff
@@ -7,9 +7,11 @@ import Markdown from '../Markdown';
...
 
 const renderPhoto = (photo, index) => (
-  <div className="action-step__photo" key={index}>
-    <img src={photo} alt="action step example" />
-  </div>
+  <div
+    className="action-step__photo"
+    key={index}
+    style={{ backgroundImage: `url(${photo})` }}
+  />
```
weerd: Oh, so wait, the only solution for this is to not include the images as content but as background images on a div? :diegosadface:

deadlybutter:
@weerd Agreed that could look better, I'll play around with it & talk to Luke more!

resources/assets/components/ActionPage/ActionStep.js
deadlybutter: Going this route has made it way easier to do the responsive scaling

deadlybutter:
@weerd Opened the page on my phone to see how it would actually look, the browser is a bit deceiving! It's pretty great on mobile, waiting on Luke to check it out as well

resources/assets/components/ActionPage/ActionStep.js
weerd: Hmmm, it's not ideal. We'd want these to be inserted as content images and not empty div containers. Mind elaborating on how this solved responsive issues that having it be an inserted `` that scales couldn't address?

deadlybutter: The image tag can't have child elements so the `:after` trick won't work on it. And because I'm trying to let the width of the image be fluid rather than fixed, it's necessary to use this trick to ensure the height matches.

DFurnes: Yeah, I'd love to just keep this as a standard image tag if possible.

lkpttn:
I approve

resources/assets/components/ActionPage/ActionStep.js
deadlybutter: @weerd explanation attempt number 2! 😬 

the action step photos do not have a fixed width, the width will resize until it's at either the mobile breakpoint or the feed width. since we wanted these images to be square, I can't just apply a height to them, since it will lose aspect ratio when the width changes. 

to solve this, I used the `:after { padding-bottom: 100%` [trick](https://github.com/DoSomething/phoenix-next/pull/336/files#diff-48d5724756ccb0982c40c36b24d6883bR81), which ensures the div is always a square, regardless of width. (Longer blog post [here](https://spin.atomicobject.com/2015/07/14/css-responsive-square/) is where I found it)

this would not work on an image element because image elements can't have a child tag like `:after`. so i created the div with a background as a simple alternative. 

lmk if that makes more sense!

weerd: Based on that post, couldn't you do the `` wrapped in a `<div>` tag which would have the `:after` hack or whatever?

deadlybutter: 🤔 I can give it a try!

weerd: @joe follow the example at the very end in the "Adding Content" section :)

resources/assets/components/ActionPage/ActionStep.js
```diff
@@ -8,7 +8,7 @@ import { Flex, FlexCell } from '../Flex';
 
 const renderPhoto = (photo, index) => (
   <div className="action-step__photo" key={index}>
-    <img src={photo} alt="action step example" />
+    <img src={photo} />
```
weerd: can we toss in an `alt` from the image object if it has one otherwise default to something like `action step photo` or whatever so ESLint is happier? :)

resources/assets/components/ActionPage/actionPage.scss
```diff
@@ -64,18 +64,34 @@
...
+    width: 100%;
 
     .action-step__photo {
-      padding: $base-spacing / 2;
-      // TODO: We'll want to make these square at some point, and get a proper overflow for the action step photos.
+      margin-bottom: $base-spacing;
+      width: 100%;
+      height: 100%;
+      background-position: center center;
+      background-size: 100% 100%;
```
weerd: Can probs remove all this background image stuff right?

deadlybutter: Oh yep, that ain't needed anymore

resources/assets/components/ActionPage/ActionStep.js
```diff
@@ -8,7 +8,7 @@ import { Flex, FlexCell } from '../Flex';
 
 const renderPhoto = (photo, index) => (
   <div className="action-step__photo" key={index}>
-    <img src={photo} alt="action step example" />
+    <img alt="Action step" src={photo} />
```
weerd: Maybe add a comment here to eventually pass along the image's description (which can be entered in Contentful as part of the default fields on images). So that there's a legit `alt` value instead of the same thing for each one. It can default to `Action step` but would be nice to include actual description if it's available ;)

resources/assets/components/ActionPage/actionPage.scss
```diff
@@ -64,18 +64,32 @@
 
   .action-step__photos {
     display: flex;
```
weerd: @deadlybutter Just a future reference side note; part of the complications you might have been seeing with the layout could have been due to the use of flexbox here. I _think_ this would have worked fine and be a bit simpler if we had originally just used simple two div floats (one for markdown content and the other for photos); floated when on desktop and not floated on moblile.

PR #337, opened 2017-07-13
Action page bug fix

weerd:
### What does this PR do?
This PR fixes a bug I accidentally introduced in prior Action Page updates. Since now we return a string instead of an array of one item for the `customType` property in an action step, we needed to update how we check against it when deciding whether or not to remove customTypes for unaffiliated users.

It also reintroduces the Revealer in a nicer fashion and separates the dependency between it and the RB Photo Uploader 🎉 

### DIFF
--- a/resources/assets/components/ActionPage/ActionStepsWrapper.js
+++ b/resources/assets/components/ActionPage/ActionStepsWrapper.js
@@ -37,9 +37,6 @@ const ActionStepsWrapper = (props) => {
     />
   );
 
-  const renderPhotoUploader = isSignedUp ? photoUploader : null;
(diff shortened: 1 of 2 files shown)


resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -37,7 +37,7 @@ const ActionStepsWrapper = (props) => {
     />
   );
 
-  const renderPhotoUploader = isSignedUp ? photoUploader : actionRevealer;
+  const renderPhotoUploader = isSignedUp ? photoUploader : null;
```
DFurnes: WAIT I DO HAVE AN OPINION™: can we name this as a noun `photoUploader` rather than a verb?

PR #338, opened 2017-07-13
Fixing action header responsiveness

deadlybutter:
### What does this PR do?
Fixes the action header being weird on mobile not taking the full container height

screen shot 2017-07-13 at 3 14 45 pm
screen shot 2017-07-13 at 3 14 40 pm

It's a bit stretched on mobile, not sure of a great solution without alternating between landscape / square photos. I think if the effect was more dramatic I would work on that but seems low priority and not very noticeable.

### DIFF
--- a/resources/assets/components/ActionPage/actionPage.scss
+++ b/resources/assets/components/ActionPage/actionPage.scss
@@ -37,10 +37,16 @@
 
     img {
       width: 100%;
+      height: 100%;
       position: absolute;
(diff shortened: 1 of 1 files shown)


weerd:
So this should totally be a CSS background image! It's definitely not content like the images _inside_ the steps container. As a background image you can defs use `background-size: cover;` or use the `background` shorthand to call `cover` and `center` it :)

Issue #341, opened 2017-07-14
[Feature] Refactor & Implement Redesign of RB Photo Upload Experience


weerd:
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want to do some cleanup on the RB uploader and make some enhancements too!

#### Additional Information (optional)
Refs https://trello.com/c/CifTBgbt/682-8-as-a-developer-i-want-to-do-some-cleanup-on-the-rb-uploader-and-make-some-enhancements-too

#### Tasks
This feature includes the following tasks:
- [x] Update the directory structure for the RB Photo Uploader to help facilitate upcoming refactors and start using current structure [#325](https://github.com/DoSomething/phoenix-next/pull/325)
- [x] Separate the User's RB Submission Gallery component from the RB Photo Uploader component to allow them to be included independently [#331](https://github.com/DoSomething/phoenix-next/pull/331)
- [x] Update the action steps to allow adding RB Photo Uploader and Submission Gallery as independent, sortable items in the list of Action Steps [#333](https://github.com/DoSomething/phoenix-next/pull/333), [#337](https://github.com/DoSomething/phoenix-next/pull/337)
- [ ] Create new `Card` component to style the unstyled RBs in the Submission Gallery.
- [ ] Update the dispatch methods and aim to pull logic out and into middleware.
- [ ] Update the RB Photo Uploader to showcase the User's submitted photo based on new designs.
- ...

itsjoekent:
@weerd I think you did the last two items in this issue right? safe to close?

weerd:
Nope.

weerd:
Coming back to that sooooon.

PR #343, opened 2017-07-18
Fixing submission messaging styles and cleaning out feature flag.

weerd:
### What does this PR do?
This PR adds a quick fix to some (mysteriously) overridden styles for the messaging on successful submission of a Reportback photo in the Photo Uploader component.

It also cleans up and removes the temporary feature flag for the `useComponentActions` which is now officially on the Production environment!

🔮 

Refs #341 
Pivotal https://www.pivotaltracker.com/story/show/149099235

### DIFF
--- a/resources/assets/components/FormMessage/form-message.scss
+++ b/resources/assets/components/FormMessage/form-message.scss
@@ -6,7 +6,7 @@
   &.-error {
     color: red;
 
-    & h3 {
+    h3, p {
       color: red;
     }
   }
@@ -14,7 +14,7 @@
   &.-success {
     color: green;
 
-    & h3 {
+    h3, p {
       color: green;
     }
   }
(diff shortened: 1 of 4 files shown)


resources/assets/components/FormMessage/form-message.scss
```diff
@@ -6,15 +6,15 @@
...
     color: red;
 
-    & h3 {
+    h3, p {
       color: red;
     }
   }
 
   &.-success {
     color: green;
```
deadlybutter: Actually, do we need both here? (Color at the class level and the element level)

weerd: Yeah. The element level was needed because of the setting [here](https://github.com/DoSomething/phoenix-next/pull/223/files#diff-96426e7dde62551c26e4b19018087095R9) and at the class level because on validation error we show a list of items to fix, all of which are in red. So the specific tags override the base settings and the class level just colors in any other tags in there.

deadlybutter: okie 🌮

PR #345, opened 2017-07-31
Adding sanctioned entities

weerd:
### What does this PR do?
This PR updates the entities in the system to add a few more `CampaignUpdate`, `Staff` and `CustomBlock` which all handle transforming the Contentful data to make it more usable. Defining the `CampaignUpdate` and `Staff` entities also signifies that we are trying to formalize their data structure both in the application and as formal content types on the Contentful side.

This also helps update entities so they follow a similar structure in the data they transform. It follows this pattern:

```javascript
{
  id: string
  type: string
  fields: {}
}
```

These updates aim to improve how we approach transforming data from Contentful but it's also specifically related to the update to add a `CampaignUpdate` component specifically for the new `campaignUpdate` content type on Contentful.

Refs # https://www.pivotaltracker.com/story/show/148973337

### Checklist

### DIFF
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -48,6 +48,7 @@ class CampaignRepository
         $query = (new Query)
             ->setContentType('campaign')
             ->where('fields.slug', $slug)
+            ->setInclude(3)
             ->setLimit(1);
 
         $campaigns = $this->makeRequest($query);
--- a/resources/assets/components/CampaignUpdate/CampaignUpdate.js
+++ b/resources/assets/components/CampaignUpdate/CampaignUpdate.js
@@ -1,0 +1,7 @@
+// import React from 'react';
+// import Card from '../Card';
+// import Markdown from '../Markdown';
+
+const CampaignUpdate = () => (null);
+
+export default CampaignUpdate;
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -43,6 +43,28 @@ class Campaign extends Entity implements JsonSerializable
     }
 
     /**
+     * Parse and extract activity feed item data based on content type.
+     *
+     * @param  array $activityItems
+     * @return array
+     */
+    public function parseActivityFeed($activityItems)
+    {
+        return collect($activityItems)->map(function ($item) {
+            switch ($item->getContentType()) {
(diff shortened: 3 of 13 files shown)


resources/assets/components/CampaignUpdate/CampaignUpdate.js
```diff
@@ -0,0 +1,7 @@
+// import React from 'react';
+// import Card from '../Card';
+// import Markdown from '../Markdown';
+
+const CampaignUpdate = () => (null);
+
+export default CampaignUpdate;
```
weerd: Shell component for now. I have it more fleshed out in upcoming PR!

weerd:
Sorry for the double-digit number of file changes, but a few of them are just updating `block.fields.type` to `block.type`. If you'd like you can go [commit by commit](https://github.com/DoSomething/phoenix-next/pull/345/commits)!

app/Repositories/CampaignRepository.php
```diff
@@ -48,6 +48,7 @@ public function findBySlug($slug)
         $query = (new Query)
             ->setContentType('campaign')
             ->where('fields.slug', $slug)
+            ->setInclude(3)
```
deadlybutter: Whats this?

resources/assets/components/CampaignUpdate/CampaignUpdate.js
deadlybutter: I was getting confused where the old one went, but I noticed the name went from underscore to camel case. I guess its not a huge deal but might be good to leave a note

app/Repositories/CampaignRepository.php
weerd: This helps so that when we grab data from Contentful, it includes a certain number of nested data retrievals, so we don't have to do it ourselves later with another API call. So in the `CampaignUpdate` where we now have  the `author` property, which uses the new `Staff` content type, instead of it giving us meta data about the content type so we can go and do another call to retrieve it, the initial query will grab it for us. It's ideal not to get too crazy with the amount of includes retrieved, but `3` seemed to get me all the data I needed for now :)

deadlybutter: Yeah I feel like it could get a bit crazy if we went beyond this

PR #346, opened 2017-08-01
Updating to add some defaults to SubmissionGallery and help with jest tests.

weerd:
### What does this PR do?
This PR removes the `required`-ness from a few of the properties for the `SubmissionGallery` component. These are properties that are passed in from the container, and were affecting the basic snapshot test since they weren't getting passed in. Might update the test in the future once I figure out how to properly supply props from a component's container within the tests themselves... 🤔  However, this works fine for now.

Also did a little bit of cleanup in the component.

Nope 🚫 

Refs https://www.pivotaltracker.com/story/show/149011395

### Checklist

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -69,8 +69,7 @@ export function checkForSignup(campaignId) {
       }
 
       dispatch(signupFound(campaignId));
-    })
-    .catch(() => {
+    }).catch(() => {
       dispatch(signupNotFound());
     });
   };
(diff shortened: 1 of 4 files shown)


deadlybutter:
Just out of curiosity, why was the `userId` not able to be a required parameter?

weerd:
Nevermind. I was being dumb and was passing in the props incorrectly in the test 🤦‍♂️ Updating coming up!

weerd:
@deadlybutter ok! Updated! 👯‍♂️

PR #347, opened 2017-08-02
Missed updating this displayOption variable to be string instead of array.

weerd:
### What does this PR do?
This PR makes a quick update to fix a reference to a variable whose expected type was changed for the sake of simplicity, but forced the need to update it in numerous places and I missed two references in the prior PR.

It also updates to indicate a specific type for `reportbacks` blocks.

🍑


deadlybutter:
nice catch!

PR #348, opened 2017-08-03
Adds some customization options for campaign admins

deadlybutter:
### What does this PR do?
To make the new Hunt campaign a better experience we're adding some new functionality for campaign admins to hide things on the action page

- Hide the "Step One", "Step Two", etc on the action steps
- Hide the quantity input field and submit a specific number as the impact (eg: 1)

### hide steps 
screen shot 2017-08-03 at 2 47 23 pm

### hide input, test submission
screen shot 2017-08-03 at 3 52 18 pm
screen shot 2017-08-03 at 3 52 11 pm

### DIFF
--- a/resources/assets/components/ActionPage/ActionStepsWrapper.js
+++ b/resources/assets/components/ActionPage/ActionStepsWrapper.js
@@ -13,7 +13,7 @@ const ActionStepsWrapper = (props) => {
   const { actionSteps, callToAction, campaignId, clickedSignUp,
     hasPendingSignup, isAuthenticated, isSignedUp } = props;
 
-  const photoUploader = photoUploaderProps => (
+  const renderPhotoUploader = photoUploaderProps => (
     <FlexCell key="reportback_uploader" width="full">
       <ReportbackUploaderContainer {...photoUploaderProps} />
     </FlexCell>
@@ -57,7 +57,7 @@ const ActionStepsWrapper = (props) => {
         );
 
       case 'photo-uploader':
-        return isSignedUp ? photoUploader({
+        return isSignedUp ? renderPhotoUploader({
           quantityOverride: additionalContent.quantityOverride || null,
         }) : null;
 
--- a/resources/assets/components/ActionPage/ActionStepsWrapper.js
+++ b/resources/assets/components/ActionPage/ActionStepsWrapper.js
@@ -73,6 +73,7 @@ const ActionStepsWrapper = (props) => {
             background={step.background}
             photos={step.photos}
             photoWidth={step.displayOptions === 'full' ? 'full' : 'one-third'}
+            hideStepNumber={step.hideStepNumber}
             shouldTruncate={step.truncate}
           />
         );
--- a/resources/assets/components/ActionPage/ActionStepsWrapper.js
+++ b/resources/assets/components/ActionPage/ActionStepsWrapper.js
@@ -42,7 +42,7 @@ const ActionStepsWrapper = (props) => {
     const type = step.customType || 'default';
     const title = step.title;
     const content = step.content || null;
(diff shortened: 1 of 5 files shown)


resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -73,6 +76,7 @@ const ActionStepsWrapper = (props) => {
             background={step.background}
             photos={step.photos}
             photoWidth={step.displayOptions === 'full' ? 'full' : 'one-third'}
+            hideStepNumber={step.hideStepNumber}
```
deadlybutter: In retrospect I could have probably done the hideStepNumber in the additional content as well

deadlybutter:
screen shot 2017-08-03 at 3 59 04 pm

Also here's a test without the input override

resources/assets/components/ActionPage/ActionStepsWrapper.js
weerd: Was it added as a field?

deadlybutter: Yeah I added it on the action step

weerd: K. For stuff like this lets use the `additionalContent` since we don't really know yet the full extent of how this will shape up and don't want to be adding/deleting fields unless we're sure it's a more permanent thing :)

deadlybutter:
Updated to remove the field on the action step model, it now lives in the additional content json.

screen shot 2017-08-04 at 12 46 46 pm

screen shot 2017-08-04 at 12 54 31 pm

resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -51,12 +52,14 @@ const ActionStepsWrapper = (props) => {
...
             content={content}
             photo={step.photos[0]}
-            byline={step.additionalContent}
+            byline={additionalContent}
           />
         );
 
       case 'photo-uploader':
-        return isSignedUp ? photoUploader : null;
+        return isSignedUp ? photoUploader({
```
DFurnes: Let's rename this to `renderPhotoUploader` now that it's a function.

resources/assets/components/ReportbackUploader/ReportbackUploader.js
```diff
@@ -56,7 +56,7 @@ class ReportbackUploader extends React.Component {
     const reportback = {
       media: this.state.media,
       caption: this.caption.value,
-      impact: this.impact.value,
+      impact: this.props.quantityOverride || this.impact.value,
```
DFurnes: What's the deal here? Is this to sidestep server-side validation?

deadlybutter: Yeah, we didn't want to cause any downstream problems with having 0's / null's being thrown around. Adam is going to just use 1 here.

resources/assets/components/ActionPage/ActionStepsWrapper.js
deadlybutter: sure thing!

resources/assets/components/ReportbackUploader/ReportbackUploader.js
DFurnes: Oh, I see what's going on. This seems fine.

DFurnes: Yep yep, that seems good for now.

deadlybutter:
Naming updated!

DFurnes:
Thanks! 👍 from me, let's get @weerd's feedback on Monday and then merge it in.

PR #349, opened 2017-08-07
Layout updates and separating rbs

weerd:
### What does this PR do?
This PR updates layout styles to shift towards using more `padding` around items instead of `margins`, whose behavior and collapsing properties, while great for formatting body copy and other media, are less ideal and predictable for layout (except for maybe very large regions in how we use it with Susy).

It also adds a temporary solution for dealing with Reportbacks feed blocks a bit better™ (see below).

Previously, we'd output a `full`, `two-thirds` or `one-third` container to then place `one-third` Reportback items inside that container. Lots of nesting (the RBs are themselves nested 3 levels deep in `flex__cell` divs, so that's a story for another time 📖 ). It also didn't match how we would insert Reportbacks in the view more section... each row was just 3 `one-third` items which makes the most sense. So, this PR updates how these `reportbacks` `custom_block` items are processed to separate them out into rows of however many specified, but each as a separate item. This also helped with the new `padding` approach to layout!

Refs https://www.pivotaltracker.com/story/show/148973337


app/Entities/Campaign.php
```diff
@@ -20,13 +20,44 @@
...
+            default:
+                $count = 3;
+        }
+
+        for ($i = 0; $i < $count; $i++) {
+            $posts[] = [
+                'id' => str_random(22),
+                'type' => 'reportbacks', // @TODO: reporbackPost?
+                'fields' => [
+                    'displayOptions' => 'one-third',
```
deadlybutter: Should this be `$entry->getType()`?

deadlybutter: Oh wait I see what's going on

weerd: Oh it could be but I already check the type in the `case` that ends up calling this function. I was also thinking about renaming it to `reportbackPost` to go more in line w/ the Rogue naming but wasn't 100% sure.

resources/assets/scss/base.scss
```diff
@@ -32,3 +32,8 @@ h1, h2, h3, p {
     width: 100%;
   }
 }
+
+// Utility classes.
+.padded {
```
deadlybutter: Hm, I saw we removed the margin from that call to action class. Should this padding be applied to that element as well?

weerd: Nope. The CTA needed margins to help align it with the other items. Not a problem anymore.

weerd: image

app/Entities/Campaign.php
DFurnes: Yeah, at some point it'd be nice to switch over to Rogue naming – "reportback" is 100% out. It's just signups and posts now. 😈

PR #350, opened 2017-08-08
Updating the gateway version

deadlybutter:
Update to use latest version of Gateway-JS

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   "dependencies": {
     "@dosomething/analytics": "^2.0.2",
     "@dosomething/forge": "^6.7.10",
-    "@dosomething/gateway": "^1.0.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
   "dependencies": {
     "@dosomething/analytics": "^2.0.2",
     "@dosomething/forge": "^6.7.10",
-    "@dosomething/gateway": "^1.0.1",
+    "@dosomething/gateway": "^1.0.3",
```
weerd: Wrong version: https://github.com/DoSomething/gateway-js/blob/master/package.json#L3

`1.1.3` it wasn't working cause the carrot only grabs updates to patches, but we haven't updated this since 1.0!

DFurnes: Caret should grab any `1.x.x`, but still nice to update!

PR #351, opened 2017-08-09
CampaignUpdate component and Card component/pattern

weerd:
### What does this PR do?
This PR fleshes out the new `CampaignUpdate` component and adds the new `Card` component that will be taking over the duties that the `BlockWrapper` was doing. Moving forward the `Block` component will behave mostly as a container component but _not_ provide any visual styling. The role of visual styling will be provided by the `Card` component or some other more specific visual component if needed.

This also begins our initial experimentation with the idea of _utility classes_ or _functional classes_ for use within the markup of our components. It adds some initial classes used with newly added components.

The following are some examples as classes are added to the `Card` component:

Bare-bones `Card`:
screen shot 2017-08-08 at 8 12 55 pm

`Card` with `padded` and `bordered` classes:
screen shot 2017-08-08 at 8 14 25 pm

`Card` with `rounded` class:
screen shot 2017-08-08 at 8 14 57 pm

Passing `title` prop to the `Card` component:
screen shot 2017-08-08 at 8 16 04 pm

I did not include the "share" component to the `CampaignUpdate`. It will come in a follow-up PR since it deals with the permalink stuff, which I need to do some testing on and making sure the system can decide when to render a `CampaignUpdate` or a `CustomBlock` of type `campaign_update` which are currently 2 different things until we fully switch to the new component.

Refs https://www.pivotaltracker.com/story/show/149011365
Refs https://www.pivotaltracker.com/story/show/148973337

### Checklist


weerd:
screen shot 2017-08-09 at 9 44 07 am

Example visual of the new `CampaignUpdate` component which should look exactly like the old one, but is now more versatile ;)

resources/assets/components/CampaignUpdate/CampaignUpdate.js
```diff
@@ -1,7 +1,38 @@
...
+  return (
+    <Card id={id} classNames="rounded bordered" title="Campaign Update">
+      <Markdown classNames="padded">
+        {content}
+      </Markdown>
+
+      <footer className="padded">
+        <Byline
+          author={name}
+          avatar={avatar || undefined}
```
deadlybutter: `null`?

weerd: The value for `avatar` might come in as `null` and if so we want to pass `undefined` so that the default value for `avatar` in the `Byline` is used instead. If you pass `undefined` when a param has a default value, then it is used, if `null` is passed then it thinks that's the value you want and uses `null` instead of the default value (which in this case is the default-avatar.jpg).

weerd: image

End result is this... my staff entry on Contentful doesn't have a pic added which is something I was specifically testing for, but in Adam's case it does.

deadlybutter: Gotcha!

resources/assets/components/CampaignUpdate/CampaignUpdate.test.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { shallowToJson } from 'enzyme-to-json';
+import CampaignUpdate from './index';
+
+const author = {
+  fields: {
+    avatar: 'http://example.com/avatar-aang.jpg',
+    jobTitle: 'The Last Airbender',
+    name: 'Aang'
```
DFurnes: :arrow_up: 👏

weerd:
@deadlybutter @DFurnes Added tests for the [`CampaignUpdate`](https://github.com/DoSomething/phoenix-next/pull/351/files#diff-df5669d97608855f4e2b5c36a89ee6d0) and [`Card`](https://github.com/DoSomething/phoenix-next/pull/351/files#diff-40e1bf1236f5b9e9912be5e530ee2ee9) components 💃

deadlybutter:
Seems like the linter is angry

weerd:
Yeah, at the new test files. I forgot I have to manually lint them.

PR #352, opened 2017-08-09
Quiz page

deadlybutter:
### What does this PR do?
This PR adds the scaffolding for the quiz page.
- Generic /quiz page that will list all of the quizzes on the campaign
- Specific /quiz/:id page that will display a specific quiz.

**No quizzes**
screen shot 2017-08-09 at 1 32 19 pm

**Fake quiz exists**
screen shot 2017-08-09 at 1 49 27 pm

**Visiting fake quiz**
screen shot 2017-08-09 at 1 50 06 pm

----

Also tested that Keen.io is recognizing the path/page relationship correctly when sending events ✅

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,4 +1,5 @@
 {
+  "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
     "@dosomething/analytics": {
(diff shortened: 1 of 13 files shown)


deadlybutter:
also warning: lots of files but they're either empty or almost empty

weerd:
Wouldn't the Quiz be part of the Action page? Maybe I missed the discussion on this...

deadlybutter:
Product brief calls for experimenting with it in a few places, but me & Ashley decided to try it as a seperate page first

deadlybutter:
For reference if anyone stumbles onto this, we're looking at Buzzfeed as a good example for a starting point we'd like to reach. And they seem to have a lot of success with full page quizzes.

ngjo:
Not sure if this is still an issue, but if this tab would go on campaigns w/ lots of other tabs, it might look wonky on smaller screens. I think @weerd flagged this as an issue in the past when we were doing tab work

deadlybutter:
Yeah I was thinking about that as well. For now it's only all-we-win which _should_ be fine

weerd:
>  which should be fine

Famous last words 🗣

deadlybutter:
😄

ashleybaldwin:
yeah, i think the multi tab wonkiness was just on mobile or tablet if i remember the card correctly, but pretty sure it is still in the backlog. I think its fine for now.

weerd:
@ashleybaldwin Is there a reason we're not taking this step by step? So first doing a single quiz in the Actions, testing how that goes and moving from there. Feels like a jump to decide on a dedicated tab page with potential for multiple quizes, etc...

ashleybaldwin:
@weerd i don't have a problem with that as long as we can still get proper event tracking in keen/if just putting this in the actions section is easier as an initial prototype. @deadlybutter thoughts?

deadlybutter:
I don't think it's that big of a step, and it arguably makes the code simpler because we don't have to build out a carousel or anything fancy to fit in a card

deadlybutter:
If quizzes are a whopping failure it's not _that_ hard to remove a page

weerd:
@deadlybutter wouldn't the first step just be building a `Quiz` component that could be added anywhere? And make sure it functions as intended?

>If quizzes are a whopping failure it's not that hard to remove a page

It's even easier just removing a single component 😉

DFurnes:
+1 to just making this a single component.

deadlybutter:
Maybe I'm alone on this but sticking 10 questions on the action page would look incredibly weird unless we built a carousel or some way to hide them & progress through it. I would rather stick to the Buzzfeed model, which we know works, and iterate from that

DFurnes:
Did we ever get around to allowing arbitrary components to be added on a page?

deadlybutter:
No, but that would have the call to action floating next to it as well, which would require a workaround to disable that as well. It seems like a lot of hacking when this really doesn't seem like a lot to me.

DFurnes:
Yeah, I'm not saying it needs to block merging this since you've already done all the work, but this is a perfect example of why we would've wanted to make that more flexible.

deadlybutter:
True, but also we'd still need the logic to determine if the page should display or not based on the quizzes. And something to handle linking to the quiz. And at that point it would feel like over engineering the custom type to be more than the markdown space it was intended to be .

deadlybutter:
Another thing to note: There's no reason we can't test Quizzes elsewhere, and I could definitely see us making multiple Quiz layouts to test in different places. This is just the first place we wanted to test it, because we're referencing a proven model we have high confidence in.

weerd:
> This is just the first place we wanted to test it, because we're referencing a proven model we have high confidence in.

Still it's a proven model with a different goal in mind than ours so we can't be necessarily sure of that.

PR #353, opened 2017-08-09
Updating Markdown component prop name.

weerd:
### What does this PR do?
This PR updates the prop name for the `Markdown` component which was changed up in [#351](https://github.com/DoSomething/phoenix-next/pull/351/files#diff-9b8604dbf9a929a73f72e97b2dbb03c2)

Without the change, we were getting some broken styles!

### Checklist


weerd:
I've contemplated maybe changing `classNames` to `classes` across our components since that extra `s` in `classNames` is too close to the official `className` prop.

weerd:
Nevermind.

Refs #355

PR #355, opened 2017-08-10
Classname prop updates

weerd:
### What does this PR do?
This PR updates all the components we had that were using `classNames` attribute instead of `className` which is what we should be using. In a prior PR ([#353](https://github.com/DoSomething/phoenix-next/pull/353)) I had updated some items to use the latter instead of the former since we had already been using `classNames` in a few other places. But after discussion with @DFurnes and @deadlybutter expressing concern (in the form of 🤔 ) it became apparent we should only be using `className`.

Refs #353


weerd:
@deadlybutter yar, nice catch!

PR #358, opened 2017-08-11
Adding social share link to campaign update component

weerd:
### What does this PR do?
This PR makes a few updates to the Campaign Update component(s), changes the structure for how the Byline and Share components are output in the Campaign Updates. This also properly adds the shareable facebook link to the new version of the `CampaignUpdate` component.

Sorry for the large number of files. 5 or so are just test related, and I split the commits into logical sections so might be easier going commit by commit
- [Updating the components](https://github.com/DoSomething/phoenix-next/pull/358/commits/436f72c05ebb84edc1233cb4f32fe54db162e4cd)
- [Making sure Block component knows to use the container instead](https://github.com/DoSomething/phoenix-next/pull/358/commits/51136647c775b65649f754d313cb43288758f18d)
- [Updates to the Byline and Share components](https://github.com/DoSomething/phoenix-next/pull/358/commits/d14f417c40d576a16191fc049525ff35369012d1)
- [Updating tests!](https://github.com/DoSomething/phoenix-next/pull/358/commits/f7278ec1669b0c528ced009d2da8600c7a946518)

Next up will be updating the `Card` component so that when an ID is passed, it makes the header (yellow bar text) a permalink to that content.

Refs https://www.pivotaltracker.com/story/show/150140471

### Checklist

### DIFF
--- a/resources/assets/helpers/index.js
+++ b/resources/assets/helpers/index.js
@@ -282,6 +282,24 @@ export function makeHash(string) {
 }
 
 /**
+ * Make a shareable link to a content item.
+ *
+ * @param  {Object} state
+ * @param  {String} key  An id or a slug for the content.
+ * @param  {String} type
+ * @return {String}
+ */
+export function makeShareLink(type, options = {}) {
+  switch (type) {
+    case 'campaigns':
+      return `${options.domain}/us/campaigns/${options.slug}/blocks/${options.key}`;
+
+    default:
+      throw new Error('Please provide an expected section type for generating the link.');
+  }
+}
+
+/**
  * Get the days between two Date objects
  * @see  http://stackoverflow.com/questions/2627473/how-to-calculate-the-number-of-days-between-two-dates-using-javascript
  *
--- a/resources/assets/helpers/index.js
+++ b/resources/assets/helpers/index.js
@@ -288,8 +288,9 @@ export function makeHash(string) {
  * @param  {String} key  An id or a slug for the content.
  * @param  {String} type
  * @return {String}
+ * @flow
  */
-export function makeShareLink(type, options = {}) {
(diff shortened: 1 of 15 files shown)


resources/assets/helpers/index.js
```diff
@@ -282,6 +282,24 @@ export function makeHash(string) {
...
 
 /**
+ * Make a shareable link to a content item.
+ *
+ * @param  {Object} state
+ * @param  {String} key  An id or a slug for the content.
+ * @param  {String} type
+ * @return {String}
+ */
+export function makeShareLink(type, options = {}) {
```
DFurnes: It might be nice to add type-checking here to enforce the `{domain: string, slug: string, key: string}` format.

resources/assets/scss/base.scss
```diff
@@ -38,6 +38,30 @@ h1, h2, h3, p {
   border: 1px solid $primary-border-color;
 }
 
+.cl-both {
```
DFurnes: This is subjective, but what do you think about spelling these out? e.g. `.clear-both` or `.float-left`

weerd: Yeah, I was debating internally whether or not to. I think I prefer spelling them out, but was going by what I saw in other systems. I'll switch it up; prefer it that way as well :)

resources/assets/helpers/index.js
weerd: I can give that a whirl ;)

weerd:
@DFurnes when I added `/* @flow */` to top of the helpers file it blew up with a ton of errors flagged by Flow. I didn't realize we can just add `@flow` in the docblock and it'll only check that one method. Pretty cool and very helpful to work on that file bit by bit to add type checking!

DFurnes:
Oh very neat! Had no idea you could do that.

PR #360, opened 2017-08-14
Adding Quiz routing

deadlybutter:
### What does this PR do?
This PR adds the quiz routing and some quick fields to the Quiz Beta contentful model so we can find+display a quiz block placeholder. 

screen shot 2017-08-14 at 2 01 15 pm

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -124,6 +124,24 @@ class Campaign extends Entity implements JsonSerializable
     }
 
     /**
+     * Parse and extract data for quizzes.
+     *
+     * @param  array $quizzes
+     * @return array
+     */
+    public function parseQuizzes($quizzes)
+    {
+        return collect($quizzes)->map(function ($quiz) {
+            $data = [];
+
+            $data['title'] = $quiz->title;
+            $data['slug'] = $quiz->slug;
+
+            return $data;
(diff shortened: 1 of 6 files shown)


app/Entities/Campaign.php
```diff
@@ -124,6 +124,24 @@ public function parseActionSteps($actionSteps)
...
+     */
+    public function parseQuizzes($quizzes)
+    {
+        return collect($quizzes)->map(function ($quiz) {
+            $data = [];
+
+            $data['title'] = $quiz->title;
+            $data['slug'] = $quiz->slug;
+
+            return $data;
```
weerd: Since we know we'll definitely be having quizes, might make sense to go ahead and make a `Quiz.php` entity similar to the `CampaignUpdate.php` and `Staff.php` entities. This way the setup is there even once we decide to use an official content type on Contentful for the quizes (instead of the beta content type), we'd end up taking this approach anyway.

weerd: Not vital for merging though... could be a quick update in the future.

deadlybutter: Yeah, I was def. thinking about that once the questions & whatnot are getting pulled in.

weerd:
👍  post-merge

PR #362, opened 2017-08-14
Make sure cards fill their container regardless of amount of content provided.

weerd:
### What does this PR do?
This PR fixes a small styling bug with the `Card` component where it does not fill its parent container if there is not enough content.

Example of the bug:
image


weerd:
No promises @deadlybutter :trollface:

PR #363, opened 2017-08-14
Quiz fields

deadlybutter:
### What does this PR do?
This PR continues to define the structure of the quiz and starts breaking out the fields required 

TODO: Test case (I didn't forget!)

### DIFF
--- a/app/Entities/Quiz.php
+++ b/app/Entities/Quiz.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace App\Entities;
+
+use JsonSerializable;
+
+class Quiz extends Entity implements JsonSerializable
+{
+    /**
+     * Convert the object into something JSON serializable.
+     *
+     * @return array
+     */
+    public function jsonSerialize()
+    {
+        return [
+            'title' => $this->title,
+            'slug' => $this->slug,
+        ];
+    }
+}
--- a/app/Entities/Quiz.php
+++ b/app/Entities/Quiz.php
@@ -16,6 +16,7 @@ class Quiz extends Entity implements JsonSerializable
         return [
             'title' => $this->title,
             'slug' => $this->slug,
+            'json' => $this->json,
         ];
     }
 }
--- a/app/Entities/Quiz.php
+++ b/app/Entities/Quiz.php
@@ -19,6 +19,8 @@ class Quiz extends Entity implements JsonSerializable
             'fields' => [
                 'title' => $this->title,
                 'slug' => $this->slug,
+                'introduction' => $this->introduction,
+                'conclusion' => $this->conclusion,
                 'json' => $this->json,
             ],
         ];
--- a/app/Entities/Quiz.php
+++ b/app/Entities/Quiz.php
@@ -14,9 +14,13 @@ class Quiz extends Entity implements JsonSerializable
     public function jsonSerialize()
     {
         return [
-            'title' => $this->title,
-            'slug' => $this->slug,
-            'json' => $this->json,
+            'id' => $this->entry->getId(),
+            'type' => $this->getContentType(),
+            'fields' => [
+                'title' => $this->title,
+                'slug' => $this->slug,
+                'json' => $this->json,
+            ],
         ];
     }
 }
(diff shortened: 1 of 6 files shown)


deadlybutter:
Added a test, ready for 👁  @weerd

app/Entities/Quiz.php
```diff
@@ -0,0 +1,22 @@
...
+     *
+     * @return array
+     */
+    public function jsonSerialize()
+    {
+        return [
+            'title' => $this->title,
+            'slug' => $this->slug,
+            'json' => $this->json,
+        ];
```
weerd: Nice! One thing that @DFurnes and I talked about and I forgot to mention in last PR, especially since the entity wasn't made yet, that we want to universalize how we return the data so it's consistent across entity types. With that we want to have this structure:

```php
return [
    'fields' => [
        // all the rest of the fields...
    ],
];
```
Take a look at the [`CampaignUpdate.php`](https://github.com/DoSomething/phoenix-next/blob/dev/app/Entities/CampaignUpdate.php) and [`Staff.php`](https://github.com/DoSomething/phoenix-next/blob/dev/app/Entities/Staff.php) for ref.

resources/assets/components/Quiz/Question.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Question = ({ question }) => (
+  <div className="question">
+    <h1>{ question.title }</h1>
```
weerd: Not a huge deal but the title of the quiz is likely the `h1` in the content, and so each question's title is probs an `h2`. :)

app/Entities/Quiz.php
deadlybutter: Seems ok to me!

resources/assets/components/Quiz/Question.js
deadlybutter: Yeah that would make sense

resources/assets/components/Quiz/Quiz.js
```diff
@@ -1,19 +1,26 @@
...
-    <div className="quiz">
-      <h1>{content.title}</h1>
-    </div>
-  </Card>
+const Quiz = ({ quiz }) => (
+  <div className="quiz">
+    <h1 className="quiz__title">{quiz.title}</h1>
+    { quiz.introduction ? <Markdown>{quiz.introduction}</Markdown> : null }
+    { quiz.json.questions.map(question => (
+      <Question key={question.title} question={question} />
```
weerd: maybe use the `makeHash()` helper on the `question.title` so it doesn't add spaces? Not sure if that matters...

DFurnes: Each quiz should also have an ID, which is an ideal key! 🔑

deadlybutter: The questions are going to be defined in JSON since Diego wanted to keep things simple for now. I'll just hash the question title for now!

weerd: @deadlybutter I believe @DFurnes meant the actual quiz containing the questions could have a key (esp since you're mapping through potentially more than one quiz) and add the quiz's id as the key.

deadlybutter: Nope, there is only one quiz here...?

resources/assets/components/Quiz/QuizContainer.js
```diff
@@ -7,9 +7,9 @@ import Quiz from './Quiz';
  */
 const mapStateToProps = (state, ownProps) => {
   const { slug } = ownProps.match.params;
-  const content = find(state.campaign.quizzes, { slug });
+  const quiz = find(state.campaign.quizzes, { fields: { slug } });
 
-  return { content };
+  return { quiz: quiz.fields };
```
weerd: maybe just return from the second line:

```js
return find(state.campaign.quizzes, { fields: { slug } });
```

This way the object you pass has the quiz id _and_ the fields, right?

deadlybutter: @weerd Trying to avoid making the statements in the Quiz component longer by adding another var in the chain. Check out the change I just made, might be a good compromise

PR #364, opened 2017-08-15
Refactoring action header to photo header

deadlybutter:
### What does this PR do?
Going to use the same styling for the quiz mvp so figured I would just break this out into a new component. Also, solves the issues with https://github.com/DoSomething/phoenix-next/pull/338

Still works on the action page!
screen shot 2017-08-15 at 11 06 08 am
screen shot 2017-08-15 at 11 06 00 am

### DIFF
--- a/resources/assets/components/ActionPage/StepHeader.js
+++ b/resources/assets/components/ActionPage/StepHeader.js
@@ -2,16 +2,17 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { FlexCell } from '../Flex';
-import LazyImage from '../LazyImage';
+import PhotoHeader from '../PhotoHeader';
 import { convertNumberToWord } from '../../helpers';
 
 const StepHeader = ({ title, step, background, hideStepNumber }) => (
   <FlexCell width="full">
-    <div className="action-step__header">
-      { background ? <LazyImage src={background} /> : null }
-      { hideStepNumber ? null : <span>step { convertNumberToWord(step) }</span> }
-      <h1>{ title }</h1>
-    </div>
+    <PhotoHeader backgroundImage={background}>
+      <div className="action-step__header">
+        { hideStepNumber ? null : <span>step { convertNumberToWord(step) }</span> }
+        <h1>{ title }</h1>
+      </div>
+    </PhotoHeader>
   </FlexCell>
 );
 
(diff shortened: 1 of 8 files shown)


deadlybutter:
Going to add tests now 😉

deadlybutter:
and snapshot tests added

resources/assets/components/ActionPage/StepHeader.js
```diff
@@ -2,16 +2,17 @@ import React from 'react';
...
 import { convertNumberToWord } from '../../helpers';
 
 const StepHeader = ({ title, step, background, hideStepNumber }) => (
   <FlexCell width="full">
-    <div className="action-step__header">
-      { background ? <LazyImage src={background} /> : null }
-      { hideStepNumber ? null : <span>step { convertNumberToWord(step) }</span> }
-      <h1>{ title }</h1>
-    </div>
+    <PhotoHeader backgroundImage={background}>
```
weerd: Not sure I get why there was a need to componentize the background image setting?

why not just add it to the `action-step__header` div instead of adding more nesting?

```html
<div className="action-step__header" style={{ backgroundImage: `url(${backgroundImage})` }}>
...
</div>
```
And toss the styles in the `action-step__header`.

thoughts?

deadlybutter: @weerd See the PR description!

weerd: Oh, than why not just use the class and apply it to each?

```html
<div className="action-step__header photo-background" style={{ backgroundImage: `url(${background})` }}>
...
</div>
```

deadlybutter: I guess it's a style preference between cluttering up the base.scss and containing it to a specific new file

deadlybutter: Like it would be one thing if this was just a background-image: cover, but there are a few additional styles on this. So to me it's a presentation component

resources/assets/scss/base.scss
```diff
@@ -62,10 +62,24 @@ h1, h2, h3, p {
   float: right;
 }
 
+// TODO: Remove this.
```
deadlybutter: @weerd I left this in otherwise the PR would grow by a lot of files... I can make an issue to remove it later.

weerd: cool cool.

PR #365, opened 2017-08-15
More field work on quizzes

deadlybutter:
### What does this PR do?
This pr does more work around the quiz fields and specifically does a bunch of work around adding answers to the questions.

I did some refactoring to make working with the json a bit safer. Also added ID's in the php side of things like previously suggested. 

Also, there is a new `award` variable. This is what we'll use to determine what we assign points to when people answer questions.

Here is the new json structure on the `questions` property in contentful 

```
{
    "items": [
        {
            "title": "test question",
            "answers": [
                {
                    "title": "this is a title",
					"award": "winner"
                },
                {
                    "title": "this is a second title",
					"award": "loser"
                }
            ]
        },
        {
            "title": "test second question",
            "answers": [
                {
                    "title": "this is a title",
					"award": "winner"
                },
                {
                    "title": "this is a second title",
					"award": "loser"
                },
                {
                    "title": "this is a third title",
					"award": "loser"
                }
            ]
        }
    ]
}
```

### DIFF
--- a/app/Entities/Quiz.php
+++ b/app/Entities/Quiz.php
@@ -18,9 +18,9 @@ class Quiz extends Entity implements JsonSerializable
             return [];
         }
 
-        return collect($question['answers'])->map(function ($answer) {
-            $data = copyFieldIfSet(['title', 'award'], $answer);
-            $data['id'] = uniqid();
+        return collect($question['answers'])->map(function ($answer, $index) {
+            $data = array_only($answer, ['title', 'award']);
+            $data['id'] = $index;
 
             return $data;
         });
@@ -38,9 +38,9 @@ class Quiz extends Entity implements JsonSerializable
             return [];
         }
 
-        return collect($questions)->map(function ($question) {
-            $data = copyFieldIfSet(['title', 'background'], $question);
-            $data['id'] = uniqid();
+        return collect($questions)->map(function ($question, $index) {
+            $data = array_only($question, ['title', 'background']);
+            $data['id'] = $index;
             $data['answers'] = $this->parseAnswersFromQuestion($question);
 
             return $data;
--- a/app/Entities/Quiz.php
+++ b/app/Entities/Quiz.php
@@ -29,7 +29,7 @@ class Quiz extends Entity implements JsonSerializable
     /**
      * Parse out the questions from the json data.
      *
-     * @return Array
+     * @return array
      */
     private function parseQuestionsFromJson()
     {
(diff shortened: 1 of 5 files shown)


app/Entities/Quiz.php
```diff
@@ -7,6 +7,47 @@
...
+     * @return array
+     */
+    private function parseAnswersFromQuestion($question)
+    {
+        if (! isset($question['answers'])) {
+            return [];
+        }
+
+        return collect($question['answers'])->map(function ($answer) {
+            $data = copyFieldIfSet(['title', 'award'], $answer);
```
DFurnes: Can we use Laravel's [`array_only`](https://laravel.com/docs/5.4/helpers#method-array-only) for this?

app/Entities/Quiz.php
```diff
@@ -7,6 +7,47 @@
...
+            $data['id'] = uniqid();
```
DFurnes: Could we use the array offset for this? It sounds like `uniqid()` [might not guarantee a unique value](http://php.net/manual/en/function.uniqid.php).

resources/assets/components/Quiz/Answer.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Answer = ({ answer }) => (
+  <div className="answer">
+    <p>{ answer.title }</p>
```
DFurnes: Let's pass an explicit prop for this, not an object.

app/Entities/Quiz.php
deadlybutter: Sure!

app/Entities/Quiz.php
deadlybutter: Of course Laravel already had a helper for that 😎

resources/assets/components/Quiz/Question.js
```diff
@@ -1,16 +1,19 @@
...
 import PropTypes from 'prop-types';
+import Answer from './Answer';
 
-const Question = ({ question }) => (
+const Question = ({ title, answers }) => (
   <div className="question">
-    <h2>{ question.title }</h2>
+    <h2>{ title }</h2>
+    {answers.map(answer => (
+      <Answer key={answer.id} answer={...answer} />
```
DFurnes: Oh, one thing: remove the `answer=` so this is just `{...answer}`.

PR #366, opened 2017-08-15
Campaign update links and tweets

weerd:
### What does this PR do?
This PR updates the `CampaignUpdate` component to allow the last items of functionality that the old custom block type allowed; adding embedded links and option to show content as a tweet-style message if the content is less than 144 characters.

It cleans up the `Embed` component a bunch. I noticed that the embed component had a bunch of flexbox styles for it just to render the spinner in the center and then immediately overrode them once the content was loaded. So I reversed it a bit and now it uses a `.flex-center-xy` class that it toggles based on whether the content is loaded or not. I made a few other utility class usage improvements and maybe a [questionable one (due to the content being in `<p>` tags inside the Markdown component)](https://github.com/DoSomething/phoenix-next/pull/366/files#diff-96426e7dde62551c26e4b19018087095R66), but feel free to modify if you don't agree with it!

I only updated the existing tests, but didn't add new ones that could test for tweet style and passing a an embed link because I ran out of time before I'm off. I can add when I get back or feel free to add as well!

?

### DIFF
--- a/resources/assets/components/Embed/index.js
+++ b/resources/assets/components/Embed/index.js
@@ -1,3 +1,5 @@
+/* eslint-disable react/no-danger */
+
 import PropTypes from 'prop-types';
 import React from 'react';
 import classnames from 'classnames';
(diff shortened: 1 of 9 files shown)


resources/assets/components/Embed/index.js
```diff
@@ -38,15 +38,27 @@ class Embed extends React.Component {
     }
 
     return (
-      <div className={classnames('embed', { '-loaded': this.state.title })}>
-        {embed}
+      <div className={classnames('embed', this.props.className)}>
+        <div className={classnames('wrapper', {
+          'flex-center-xy': ! this.state.title,
+          'bordered': this.state.title, // eslint-disable-line quote-props
```
DFurnes: We should consider disabling this linter, we use it quite a bit.

resources/assets/scss/base.scss
```diff
@@ -34,6 +34,12 @@ h1, h2, h3, p {
 }
 
 // Utility classes.
+// @TODO: Move over to Forge 7 soon!
```
DFurnes: ⛵️ 🔜 7️⃣

DFurnes:
One little piece of design QA, we should add a bottom margin to the embed:

screen shot 2017-08-16 at 3 15 42 pm

weerd:
@DFurnes the bottom margin only affects the old `customBlock` component of `campaign_update` not the new `campaignUpdate` component:

image

I can try and see if I can add a special class or something for when campaign updates are rendered in the old custom block vs the new campaign update, because if I add it universally, it'll add too much spacing with the new component.

resources/assets/components/Embed/index.js
weerd: Yeah, I meant to ask in a comment in the PR whether we wanna disable this linter. I can add an issue to the eslint-config repo ;)

weerd:
@DFurnes updated and added a div with utility class to add the padding on the old custom block campaign updates!

image

DFurnes:
Looks perfect! 👍

PR #369, opened 2017-08-17
Adding actual signup to the quiz

deadlybutter:
### What does this PR do?
This PR ensures that the quiz completion makes an actual signup on the campaign, and doesn't result in any changes to the user's experience afterwards like a traditional signup

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -10,7 +10,6 @@ import {
   queueEvent,
   trackEvent,
   addNotification,
-  convertExperiment,
   openModal,
 } from '../actions';
 
@@ -23,8 +22,6 @@ import {
 export function signupCreated(campaignId) {
   return (dispatch, getState) => {
     const { user } = getState();
-
-    dispatch(openModal());
 
     dispatch({
       type: SIGNUP_CREATED,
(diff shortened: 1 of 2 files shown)


resources/assets/actions/signup.js
```diff
@@ -99,28 +96,23 @@ export function getTotalSignups(campaignId) {
...
     if (getState().signups.data.includes(campaignId)) {
-      dispatch(push('/action'));
-      return;
+      return shouldUpdateExperience ? dispatch(push('/action')) : null;
     }
 
     dispatch(signupPending());
 
-    (new Phoenix()).post('next/signups', { campaignId }).then((response) => {
+    return (new Phoenix()).post('next/signups', { campaignId }).then((response) => {
```
deadlybutter: Linter was yelling at me to add a return here since I made the other dispatches a `return`

resources/assets/actions/signup.js
```diff
@@ -99,28 +96,23 @@ export function getTotalSignups(campaignId) {
...
-      dispatch(queueEvent('clickedSignUp', campaignId, metadata));
-      return;
+      return dispatch(queueEvent('clickedSignUp', campaignId, metadata));
     }
 
     // If we already have a signup, just go to the action page.
     if (getState().signups.data.includes(campaignId)) {
-      dispatch(push('/action'));
-      return;
+      return shouldUpdateExperience ? dispatch(push('/action')) : null;
```
DFurnes: Can we rename this to `shouldRedirectToActionTab` (or something else that's more descriptive)?

deadlybutter: @DFurnes I originally did that but then I had to handle the Modal as well, so I went more generic intentionally

DFurnes: Hmm, okay. I don't really know what "update experience" means when skimming the code, though. Even with the modal, isn't this still effectively showing the user the action page?

deadlybutter: True, okay, I'll make the switch!

PR #372, opened 2017-08-18
Adding quiz comparison

deadlybutter:
### What does this PR do?
This PR adds the quiz comparison feature to finalize the conversion work. It also did the string replacing and winner logic to present that information to the user. Also did some quick visual fixes.

### DIFF
--- a/resources/assets/components/Quiz/QuizContainer.js
+++ b/resources/assets/components/Quiz/QuizContainer.js
@@ -2,6 +2,11 @@ import { connect } from 'react-redux';
 import { find } from 'lodash';
 import Quiz from './Quiz';
 import {
+  pickWinner,
+  replaceStringWithWinner,
+  replaceStringWithPercent,
+} from './QuizHelpers';
+import {
   pickQuizAnswer,
   compareQuizAnswer,
   viewQuizResult,
@@ -16,10 +21,22 @@ const mapStateToProps = (state, ownProps) => {
   const quizContent = find(state.campaign.quizzes, { fields: { slug } });
   const quizId = quizContent.id;
   const quizData = state.quiz[quizId];
+  let quizFields = quizContent.fields;
+
+  const winner = pickWinner(quizData ? quizData.questions : {}, quizFields.questions);
+  if (winner) {
+    quizFields = {
+      ...quizFields,
+      conclusion: replaceStringWithWinner(quizFields.conclusion, winner),
+      comparison: replaceStringWithPercent(
+        replaceStringWithWinner(quizFields.comparison, winner),
+      ),
+    };
+  }
 
   return {
     id: quizId,
-    fields: quizContent.fields,
+    fields: quizFields,
     data: quizData,
   };
 };
--- a/resources/assets/components/Quiz/QuizContainer.js
+++ b/resources/assets/components/Quiz/QuizContainer.js
@@ -5,7 +5,7 @@ import {
   pickWinner,
   replaceStringWithWinner,
   replaceStringWithPercent,
-} from './QuizHelpers';
+} from './helpers';
 import {
   pickQuizAnswer,
   compareQuizAnswer,
--- a/resources/assets/components/Quiz/QuizHelpers.js
+++ b/resources/assets/components/Quiz/QuizHelpers.js
@@ -23,7 +23,7 @@ export const pickWinner = (responses, questions) => {
     const newTallies = { ...currentTallies };
 
     awards.forEach(award => (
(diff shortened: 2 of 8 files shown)


resources/assets/components/Quiz/QuizHelpers.js
```diff
@@ -0,0 +1,46 @@
+import { findById, getRandomInt } from '../../helpers';
+
+/* eslint-disable no-restricted-syntax */
+/* eslint-disable no-continue */
```
deadlybutter: Added these because of the multiple `for` and `continue` statements that eslint was mad about

weerd: Can you write this in a different way so you don't have to disable them. While we have to `eslint-disable` in a few areas in our codebase, we should aim to not have to use them just because it's more convenient.

resources/assets/helpers/index.js
```diff
@@ -349,3 +349,26 @@ export function getFormattedScreenSize(screenWidth = window.innerWidth) {
...
+/**
+ * Get a random number in a range.
+ * @see https://stackoverflow.com/a/24152886/2129670
+ *
+ * @param  {int} min
+ * @param  {int} max
+ * @return {int}
+ */
+export function getRandomInt(min, max) {
+  return Math.floor(Math.random() * (max - min + 1)) + min; // eslint-disable-line
```
weerd: What was eslint complaining about here?

deadlybutter: something about `+` and `-` in the same line if I remember right. And if I added the specific rule in the `eslint-disable` it complained the line was too long. (sigh.)

DFurnes: Let's just replace this with Lodash's [`random`](https://lodash.com/docs/#random).

resources/assets/components/Quiz/QuizHelpers.js
```diff
@@ -1,39 +1,36 @@
...
-      if (! tallies[value]) {
-        tallies[value] = 0;
-      }
+    const newTallies = { ...currentTallies };
 
-      tallies[value] += 1;
-    }
-  }
+    awards.forEach(award => (
+      newTallies[award] = newTallies[award] ? newTallies[award] + 1 : 1
```
deadlybutter: Actually, I think I can make this simpler...

deadlybutter: [see this commit](https://github.com/DoSomething/phoenix-next/pull/372/commits/b1c67e03a89fb7a9051fa14b7f3d62e5f2705c8a)

resources/assets/components/Quiz/QuizContainer.js
```diff
@@ -2,6 +2,11 @@ import { connect } from 'react-redux';
 import { find } from 'lodash';
 import Quiz from './Quiz';
 import {
+  pickWinner,
+  replaceStringWithWinner,
+  replaceStringWithPercent,
+} from './QuizHelpers';
```
weerd: Not vital for merge, but you can probs name this file just `helpers.js` since it's namespaced to the Quiz directory. And we can follow that naming pattern with helpers in component directories.

resources/assets/components/Quiz/QuizHelpers.js
```diff
@@ -0,0 +1,43 @@
...
+    finalTallies[alpha] - finalTallies[beta]
+  ))[0];
+};
+
+export const replaceStringWithWinner = (string, winner) => (
+  string.replace(/{{winner}}/g, winner)
+);
+
+export const replaceStringWithPercent = string => (
+  string.replace(/{{percent}}/g, `${random(25, 34)}%`)
```
weerd: 😼

resources/assets/helpers/index.js
```diff
@@ -349,3 +349,14 @@ export function getFormattedScreenSize(screenWidth = window.innerWidth) {
...
+
+/**
+ * Search an array of objects for the given id.
+ *
+ * @param  {Array}  array
+ * @param  {String} compareId
+ * @return {Object}
+ */
+export function findById(array, compareId) {
+  return array.find(({ id }) => id === compareId);
```
weerd: Is there no equivalent LoDash method? If not no worries...

deadlybutter: Swapped it with the lodash `find`, seems to work fine. I just tend to be a DIY person for easy stuff like this so didn't even check

weerd: Sounds good. Yeah, same here... I'm a DIYer as well and have to actively remember to reference LoDash docs. Only was reminded after Dave mentioned using `random()` from LoDash ;)

PR #373, opened 2017-08-22
Update to not require end date field.

weerd:
### What does this PR do?
This PR updates the code to not require an "end date" for a campaign created in Contentful. The Contentful campaign content type does not indicate that an end date is required, but the code _did_ require it for the campaign to render. Considering Legacy campaigns using the "classic" template don't require and end date (especially for database campaigns), we shouldn't require it in the Contentful campaign content type so that it can be more versatile.

Working towards moving Legacy campaigns into Contentful.

Refs https://www.pivotaltracker.com/story/show/149026419

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -1,5 +1,6 @@
-import { Phoenix } from '@dosomething/gateway';
+import { get } from 'lodash';
 import { push } from 'react-router-redux';
+import { Phoenix } from '@dosomething/gateway';
 import { isCampaignClosed } from '../helpers';
 import {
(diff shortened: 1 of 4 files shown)


weerd:
Isn't that what I'm doing here: https://github.com/DoSomething/phoenix-next/pull/373/files#diff-927b208f806cb78d14c0f0587928cb8bR350

deadlybutter:
oh crap i was looking at the other commit :man_facepalming:

weerd:
@deadlybutter You might not have seen the check before commenting... it's in the last file in the commit ;)

PR #374, opened 2017-08-22
Fixing lint errors that suddenly blocked builds from working

deadlybutter:
### What does this PR do?
As the title describes! Here is an example 
https://app.wercker.com/dosomething/phoenix-next/runs/build/599c5c1e01f3bd0001adc3e5?step=599c5c5891990800012d9976

### DIFF
--- a/resources/assets/components/Gallery/Gallery.js
+++ b/resources/assets/components/Gallery/Gallery.js
@@ -5,13 +5,18 @@ import { modifiers } from '../../helpers';
 
 const renderGalleryItem = (child, index) => <li key={`submission-${index}`}>{child}</li>;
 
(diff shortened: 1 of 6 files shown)


resources/assets/components/Gallery/Gallery.js
```diff
@@ -14,4 +14,9 @@ Gallery.propTypes = {
   type: PropTypes.oneOf(['triad', 'quartet']),
 };
 
+Gallery.defaultProps = {
+  children: [],
```
deadlybutter: I was going to make this one null but I noticed the code tries to access `.length` so a blank array felt safer

weerd: Yeah, hokay.

PR #375, opened 2017-08-22
Improving answer styling based on Luke mock

deadlybutter:
screen shot 2017-08-22 at 2 45 41 pm
screen shot 2017-08-22 at 2 45 30 pm

this pr adds a few styling changes to the answer component,

- third sized column for now
- bold & centered text
- opacity for non selected answers
- larger border width for selected answer 
- fixing the width being incorrect on mobile 

screen shot 2017-08-22 at 2 48 02 pm

### DIFF
--- a/resources/assets/components/Quiz/Answer.js
+++ b/resources/assets/components/Quiz/Answer.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import Card from '../Card';
 
-const Answer = ({ id, title, quizId, questionId, pickQuizAnswer, isActive, shouldHide }) => (
+const Answer = ({ id, title, quizId, questionId, pickQuizAnswer, isActive, shouldFade }) => (
   <a
     className="answer"
     onClick={() => pickQuizAnswer(quizId, questionId, id)}
@@ -13,7 +13,7 @@ const Answer = ({ id, title, quizId, questionId, pickQuizAnswer, isActive, shoul
     <Card
       className={classnames('bordered rounded padding-lg', {
         '-active': isActive,
-        '-hide': shouldHide,
+        fade: shouldFade,
       })}
     >
       <p>{ title }</p>
@@ -28,7 +28,7 @@ Answer.propTypes = {
   questionId: PropTypes.string.isRequired,
   pickQuizAnswer: PropTypes.func.isRequired,
   isActive: PropTypes.bool.isRequired,
-  shouldHide: PropTypes.bool.isRequired,
+  shouldFade: PropTypes.bool.isRequired,
 };
 
 export default Answer;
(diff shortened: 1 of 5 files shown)


weerd:
image

Might be useful to actually use some real sample answers here. Styling could get weird if one answer happens to be way longer than another and you have a weird long column answer. Is there a content rule that answers need to be a certain length?

resources/assets/components/Quiz/quiz.scss
```diff
@@ -38,6 +40,11 @@
 
     .-active {
       border-color: $blue;
+      border-width: 2px;
+    }
+
+    .-hide {
```
weerd: can we change the instances of "hide" to "fade" cause this rule isn't really hiding the answer option, just fading it ;)

deadlybutter: True! Great idea

weerd: maybe try creating a utility class instead... something like `.faded` and the default is 50% opacity? thoughts?

deadlybutter: That sounds fine to me

deadlybutter:
@weerd updated! here's a screenshot with actual content as well with the latest changes

screen shot 2017-08-24 at 1 10 36 pm

weerd:
Much better with some actual sample content! thanks @deadlybutter 💃

resources/assets/components/Quiz/Answer.js
```diff
@@ -3,14 +3,19 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import Card from '../Card';
 
-const Answer = ({ id, title, quizId, questionId, pickQuizAnswer, active }) => (
+const Answer = ({ id, title, quizId, questionId, pickQuizAnswer, isActive, shouldFade }) => (
```
weerd: Small note, but I think `isActive` and `isFaded` would read better.

deadlybutter: Ah yeah, I had `should` before because `isHide` wouldn't make any sense, but `isFaded` def works

resources/assets/scss/base.scss
```diff
@@ -106,3 +106,7 @@ h1, h2, h3, p {
 .rounded {
   border-radius: $lg-border-radius;
 }
+
+.fade {
```
weerd: stay with the pattern like `padded` `rounded` `bordered` should be `faded` 😬

resources/assets/components/Quiz/Answer.js
weerd: gotcha. ok!

resources/assets/scss/base.scss
deadlybutter: 🌮

deadlybutter:
@weerd Check both commits I just pushed ^, one addresses the naming, the other is another quick bug I caught (I kept getting North Carolina and thought it was rigged!)

PR #377, opened 2017-08-23
Legacy compatible fixes part 3

weerd:
### What does this PR do?
This PR exposes a new field now available on the Campaign content type on Contentful for specifying a `template`, and makes it available within the Redux state for future use.

Initial groundwork for helping create the "classic" template on Phoenix-Next 😉 

_Note: I did not make the `template` field required for the time being in Contentful, because I didn't want to potentially break any campaigns. We can update current campaigns and eventually make the field required. For the meantime, I provide a default of `community` to the template field if the field was not filled out._

We might rename the `community` template to some other name but this works for now. Renaming since we've had talks where we might want to do a PN campaign without the community tab as an option in the future for very specific campaigns...

Refs https://www.pivotaltracker.com/story/show/149026419


weerd:
@deadlybutter 

> We might rename the community template to some other name but this works for now. Renaming since we've had talks where we might want to do a PN campaign without the community tab as an option in the future for very specific campaigns...

PR #380, opened 2017-08-24
Quiz share button

deadlybutter:
### What does this PR do?
- Adds a share button to the quiz
- Fixes some reference errors if a user tried submitting the quiz before actually picking any answers

screen shot 2017-08-24 at 1 04 29 pm

Totally random thought, I think at some point we should rename `answer` to `choice`. Also, the `width: inherit;` on the buttons is to override the `width: 100%` on the social button.


weerd:
> we should rename `answer` to `choice`

ya, this seems way more intuitive. I've been kinda confused by the use of `answer` but wasn't sure I was missing something is the logic that made that name make more sense 😖

deadlybutter:
Yeah I was thinking the same as well re: refactoring after we define the content type

PR #382, opened 2017-08-29
Dual wielding campaigns

weerd:
### What does this PR do?
This PR is the start of trying to enable components to have templates that can be otherwise specified to use to visually render it differently depending on the template chosen. This first example is with the `LedeBanner` component and will allow rendering it for either the `legacy` template style or the `mosaic` template style.

#### Legacy template style for LedeBanner
image

#### Mosaic template style for LedeBanner
image

### DIFF
--- a/resources/assets/components/LedeBanner/LedeBanner.js
+++ b/resources/assets/components/LedeBanner/LedeBanner.js
@@ -1,55 +1,34 @@
-import PropTypes from 'prop-types';
 import React from 'react';
-import Markdown from '../Markdown';
+import PropTypes from 'prop-types';
+
+import LegacyTemplate from './templates/LegacyTemplate';
+import MosaicTemplate from './templates/MosaicTemplate';
 import { contentfulImageUrl } from '../../helpers';
 
 import './lede-banner.scss';
 
 const LedeBanner = (props) => {
-  const {
-    title,
-    subtitle,
-    blurb,
-    coverImage,
-    isAffiliated,
-    legacyCampaignId,
-    clickedSignUp,
-  } = props;
+  const { coverImage, template } = props;
 
   const backgroundImageStyle = {
     backgroundImage: `url(${contentfulImageUrl(coverImage.url, '800', '600', 'fill')})`,
   };
 
-  return (
-    <header role="banner" className="lede-banner">
-      <div className="lede-banner__image" style={backgroundImageStyle} />
-      <div className="lede-banner__content">
-        <div className="wrapper">
-          <div className="lede-banner__headline">
-            <h1 className="lede-banner__headline-title">{title}</h1>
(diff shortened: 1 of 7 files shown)


resources/assets/components/LedeBanner/LedeBanner.js
```diff
@@ -1,55 +1,36 @@
...
-          <Markdown className="lede-banner__blurb">{blurb}</Markdown>
-
-          { isAffiliated ? null : <button className="button" onClick={() => clickedSignUp(legacyCampaignId, { source: 'lede banner|text: Join us' })}>Join us</button> }
-        </div>
-      </div>
-    </header>
-  );
+  const { coverImage, template } = props;
+
+  // const backgroundImageStyle = {
```
deadlybutter: Were you keeping this around intentionally?

weerd: Crap. I removed it and forgot to push up the PR. Hold please!

resources/assets/components/LedeBanner/templates/LegacyTemplate.js
```diff
@@ -0,0 +1,51 @@
...
+  const backgroundImageStyle = {
+    backgroundImage: `url(${backgroundImageUrl})`,
+  };
+
+  return (
+    <header role="banner" className="header -hero header--action has-promotions" style={backgroundImageStyle}>
+      <div className="wrapper">
+        <h1 className="header__title">{title}</h1>
+        <p className="header__subtitle">{subtitle}</p>
+        { endDate ? <p className="header__date">Ends {moment(endDate.date).format('MMMM Do')}</p> : null }
```
DFurnes: @weerd Moment.js is [rather huge](https://user-images.githubusercontent.com/583202/30333810-dd47e294-97ab-11e7-9e7e-0934de7f8d6c.png) for usage on the front-end (it accounts for 20% of our total build now, 162kb gzipped). I'd recommend checking out [date-fns](https://date-fns.org) if we want more fancy date handling logic.

weerd: Oh shiza! Ok, good to know. Forgot to consider that. I can do an update. I'll add a card to Pivotal to replace the lib w/ `date-fns` 💃

deadlybutter: Wait didn't we already have something to make the Date look pretty? What happened to that?

DFurnes: Not sure! If so, could definitely use that.

weerd: It wasn't the same and only allowed a single format I believe. I'd rather have a standardized package (like we do with `Carbon` on the PHP side) to deal with all things dates.

PR #383, opened 2017-08-29
Quiz convert 2

deadlybutter:
### What does this PR do?
This is doing some more updates to the MVP and quick fixes

- Comparison will be an A/B test we do later, so the conversion is now happening when you get your quiz results
- Renamed some functions
- One more style change

- In the actions I basically just moved some stuff around, not a lot of code changes
- On the quiz component I had to wrap those two things in a div because of the React sibling stuff, otherwise everything should be straight forward


deadlybutter:
Thanks!

PR #386, opened 2017-08-29
(NOT 100% TESTED DONT MERGE) Queue refactor

itsjoekent:
### What does this PR do?
- Cut out a some extra stuff from the events we're not going to use, makes the code a bit nicer
- Added different types of auth events (Generic & Facebook)
- Added example Facebook signup action (Not used anywhere yet)

We want to run experiments that let people authenticate via Facebook with a 1 click experience. This work allows us to run these experiments. 

For example in the Quiz action,

```
export function quizConvert(quizId) {
  return ((dispatch, getState) => {
    if (! getState().user.id) {
      const quizData = getState().quiz[quizId];
      set(quizId, QUIZ_STORAGE_KEY, quizData.questions);

      return dispatch(queueFacebookAuthEvent('quizConvert', quizId));
    }
```

Note the last return!

### DIFF
--- a/resources/assets/actions/event.js
+++ b/resources/assets/actions/event.js
@@ -21,15 +21,9 @@ export function startQueue() {
       // Check if the event is over 30 min old before dispatching.
       const isValidTimestamp = isTimestampValid(event.createdAt, (30 * 60 * 1000));
 
-      // Check if the user successfully authenticated
       const isAuthenticated = getState().user.id !== null;
 
-      let shouldFireEvent = isValidTimestamp;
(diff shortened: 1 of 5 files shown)


weerd:
Does this work allow for running the experiments or for recording the data behind the events related to 1-click auth?

deadlybutter:
That reminds me, technically I should update the action array so these two are being watched. As for experiments, yes! See the PR for an example with the quiz

weerd:
@itsjoekent also this one... can we close it?

weerd:
Closing since PR is stale and has merge conflicts.

PR #387, opened 2017-08-30
Fixing the /campaigns page.

weerd:
### What does this PR do?
This PR fixes some logic that was added to send campaign info to Northstar for when a user logs in. It accounted for when a user logs in via a specific campaign page, but not when logging in from any other page (which granted, we don't have right now unless you're an admin), like the `/campaigns` page.

I approached this solution keeping in mind that in the future, when we have static pages, etc on the Phoenix-Next platform, those pages won't have any campaign specific info to pass along to Northstar, so it only considers trying to do that if a `$campaigns` variable is even available.

### DIFF
--- a/resources/views/partials/navigation.blade.php
+++ b/resources/views/partials/navigation.blade.php
@@ -25,7 +25,11 @@
                         <li><a href="{{ route('logout') }}" class="secondary-nav-item" id="link--logout">Log Out</a></li>
                     </ul>
                 @else
-                    <a href="{{ route('login', get_login_query($campaign)) }}">Log In</a>
+                    @if (isset($campaign))
+                        <a href="{{ route('login', get_login_query($campaign)) }}">Log In</a>
+                    @else
+                        <a href="{{ route('login') }}">Log In</a>
+                    @endif
                 @endif
             </li>
         </ul>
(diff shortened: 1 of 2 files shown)


resources/views/partials/navigation.blade.php
```diff
@@ -25,7 +25,11 @@
                         <li><a href="{{ route('logout') }}" class="secondary-nav-item" id="link--logout">Log Out</a></li>
                     </ul>
                 @else
-                    <a href="{{ route('login', get_login_query($campaign)) }}">Log In</a>
+                    @if (isset($campaign))
+                        <a href="{{ route('login', get_login_query($campaign)) }}">Log In</a>
+                    @else
+                        <a href="{{ route('login') }}">Log In</a>
+                    @endif
```
weerd: Not a big fan of doing this within an `if/else` but helps limit to only when a `$campaign` variable is present.

deadlybutter: I think it's fine, alternative could be a ternary in the `get_login_query` but :man_shrugging:

weerd: Yeah, I considered a ternary, except you can't use `isset()` inside a call to a function. PHP gets angry.

PR #389, opened 2017-08-31
Quiz makeover

deadlybutter:
### What does this PR do?
quiz demo

This is a total makeover of the experience based on the mocks Luke provided. Images, colors, layout, etc.

### DIFF
--- a/resources/assets/components/Quiz/Quiz.js
+++ b/resources/assets/components/Quiz/Quiz.js
@@ -7,10 +7,13 @@ import './quiz.scss';
 
 const Quiz = ({ id, fields, data, completeQuiz, pickQuizAnswer }) => (
   <div className="quiz">
-    <h1 className="quiz__title">{fields.title}</h1>
-    {data.shouldSeeResult ? null : (
-      <Markdown>{fields.introduction}</Markdown>
-    )}
+    <div className="quiz__introduction">
+      <h1 className="quiz__title">Quiz</h1>
+      <h2 className="quiz__subtitle">{fields.title}</h2>
+      {data.shouldSeeResult ? null : (
+        <Markdown className="quiz__description">{fields.introduction}</Markdown>
+      )}
+    </div>
     {data.shouldSeeResult ? null : (fields.questions).map(question => (
       <Question
         key={question.id}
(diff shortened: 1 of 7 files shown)


resources/assets/components/Quiz/Quiz.js
```diff
@@ -70,6 +76,7 @@ Quiz.defaultProps = {
     questions: [],
     conclusion: '',
     comparison: '',
+    callToAction: '',
```
weerd: not `null` for these?

resources/assets/components/Quiz/quiz.scss
```diff
@@ -43,4 +66,28 @@
       border-width: 2px;
     }
   }
+
+  .conclusion {
+    margin-bottom: $base-spacing;
+
+    .conclusion__item {
+      display: inline-block;
+      float: left;
```
weerd: you shouldn't use both `display: inline-block` and `float: left` 🤔

weerd:
Where can I find the quiz in the deployed branch on Heroku? What's the url?

weerd:
@deadlybutter The quiz page in the example branch deploy loads rather slowly @_@ Try adding url params to the image urls so that contentful is returning smaller, more appropriate sizes!

resources/assets/components/Quiz/Quiz.js
deadlybutter: The empty string felt safer since I wasn't sure how all the Markdown functions would react to a null value, I can give it a try

resources/assets/components/Quiz/quiz.scss
deadlybutter: Yeah this was a prior thing I tried, should actually be `vertical-align: middle;`

resources/assets/components/Quiz/Quiz.js
deadlybutter: Yeah passing `null` to Markdown causes the entire thing to break ❌

weerd: ah, gotcha. just wanted to make sure!

resources/assets/scss/next-toolbox.scss
```diff
@@ -31,6 +31,7 @@ $half-spacing: $base-spacing / 2;
 // Font Stacks
 $font-source-sans: "Source Sans Pro", "Helvetica Neue", "Helvetica", "Arial", sans-serif;
 $font-league-gothic: "League Gothic", "Impact", "Franklin Gothic Bold", "Arial Black", sans-serif;
+$gothic-letter-spacing: 4px;
```
weerd: 😓

PR #391, opened 2017-09-01
Quiz design QA

deadlybutter:
### What does this PR do?
- Changes font weight & size of the question
screen shot 2017-09-01 at 10 14 31 am

### DIFF
--- a/resources/assets/components/Quiz/quiz.scss
+++ b/resources/assets/components/Quiz/quiz.scss
@@ -11,7 +11,6 @@
     color: $primary-color;
     font-size: $font-large;
     font-weight: $weight-bold;
-    font-weight: normal;
     text-transform: uppercase;
   }
 
@@ -34,6 +33,7 @@
       color: $white;
       font-family: $secondary-font-family;
       font-size: $font-larger;
+      font-weight: normal;
       letter-spacing: 2px;
       padding: $base-spacing;
 
--- a/resources/assets/components/Quiz/quiz.scss
+++ b/resources/assets/components/Quiz/quiz.scss
@@ -9,8 +9,9 @@
 
   .quiz__title {
     color: $primary-color;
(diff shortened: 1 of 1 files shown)


resources/assets/components/Quiz/quiz.scss
```diff
@@ -9,8 +9,9 @@
 
   .quiz__title {
     color: $primary-color;
-    font-size: $font-large;
+    font-size: $font-hero;
```
weerd: Speaking of font-sizes... do we not want to have a smaller size for mobile and then increase it after a breakpoint? I saw on Ashley's phone yesterday when she was going through the quiz on mobile and the title's seemed a little "large" on mobile... or keep the size if that's preferred but reduce the padding on around it on mobile?

deadlybutter: Hm it looked fine for me but yeah we can break point this now that it jumped up a lot

deadlybutter:
@weerd 👁

resources/assets/components/Quiz/quiz.scss
```diff
@@ -33,8 +33,13 @@
       color: $white;
       font-family: $secondary-font-family;
       font-size: $font-larger;
+      font-weight: normal;
       letter-spacing: 2px;
       padding: $base-spacing;
```
weerd: maybe `$half-spacing` for small screen and `$base-spacing` for the medium breakpoint?

image

Looks like a lot of spacing around question header on small screen. Thoughts?

deadlybutter: or actually I can just remove the title padding on mobile

deadlybutter: screen shot 2017-09-01 at 10 42 11 am

PR #393, opened 2017-09-14
Action step header legacy styling

mendelB:
### What does this PR do?
Adds ability to render Action Step Headers with legacy styling.

*My first pull request!!*
Basing a lot of this off of the logic @weerd implemented for the [`LedeBanner`](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/LedeBanner/LedeBanner.js). Using a templates folder and `switch` logic checking which `campaign.template` is being rendered.

https://www.pivotaltracker.com/story/show/150968514

### Checklist

### DIFF
--- a/resources/assets/components/ActionPage/StepHeader.js
+++ b/resources/assets/components/ActionPage/StepHeader.js
@@ -7,17 +7,21 @@ import MosaicStepHeaderTemplate from './templates/MosaicStepHeaderTemplate';
 const StepHeader = ({ title, step, background, hideStepNumber, template }) => {
   switch (template) {
     case 'legacy':
-      return <LegacyStepHeaderTemplate
-              title={title}
-              step={step}
-            />;
+      return (
+        <LegacyStepHeaderTemplate
+          title={title}
+          step={step}
+        />
+      );
     default:
-      return <MosaicStepHeaderTemplate
-              title={title}
-              step={step}
-              background={background}
-              hideStepNumber={hideStepNumber}
-            />;
+      return (
+        <MosaicStepHeaderTemplate
+          title={title}
+          step={step}
+          background={background}
+          hideStepNumber={hideStepNumber}
+        />
+      );
   }
 };
 
--- a/resources/assets/components/ActionPage/StepHeader.js
+++ b/resources/assets/components/ActionPage/StepHeader.js
@@ -11,6 +11,7 @@ const StepHeader = ({ title, step, background, hideStepNumber, template }) => {
         <LegacyStepHeaderTemplate
           title={title}
           step={step}
+          hideStepNumber={hideStepNumber}
         />
       );
     default:
--- a/resources/assets/components/ActionPage/StepHeader.js
+++ b/resources/assets/components/ActionPage/StepHeader.js
@@ -1,24 +1,32 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { FlexCell } from '../Flex';
-import PhotoHeader from '../PhotoHeader';
-import { convertNumberToWord } from '../../helpers';
+import LegacyStepHeaderTemplate from './templates/LegacyStepHeaderTemplate';
+import MosaicStepHeaderTemplate from './templates/MosaicStepHeaderTemplate';
 
-const StepHeader = ({ title, step, background, hideStepNumber }) => (
-  <FlexCell width="full">
-    <PhotoHeader className="action-step__header" backgroundImage={background}>
-      { hideStepNumber ? null : <span>step { convertNumberToWord(step) }</span> }
-      <h1>{ title }</h1>
-    </PhotoHeader>
-  </FlexCell>
-);
+const StepHeader = ({ title, step, background, hideStepNumber, template }) => {
+  switch (template) {
+    case 'legacy':
+      return <LegacyStepHeaderTemplate
+              title={title}
+              step={step}
+            />;
+    default:
+      return <MosaicStepHeaderTemplate
+              title={title}
+              step={step}
+              background={background}
+              hideStepNumber={hideStepNumber}
+            />;
+  }
+};
 
 StepHeader.propTypes = {
   title: PropTypes.string.isRequired,
   step: PropTypes.number.isRequired,
   hideStepNumber: PropTypes.bool.isRequired,
   background: PropTypes.string,
+  template: PropTypes.string.isRequired,
 };
 
 StepHeader.defaultProps = {
--- a/resources/assets/components/ActionPage/actionPage.scss
+++ b/resources/assets/components/ActionPage/actionPage.scss
@@ -62,10 +62,7 @@
     }
   }
 
-  .heading.-banner.legacy-step-header span {
-    font-family: $primary-font-family;
-    padding: $base-spacing 0px 0px;
-    margin-left: 0px;
+  .legacy-step-header {
     width: 100%;
   }
 
--- a/resources/assets/components/ActionPage/actionPage.scss
+++ b/resources/assets/components/ActionPage/actionPage.scss
@@ -62,14 +62,11 @@
     }
   }
 
-  .heading.-banner.legacy-step-header {
-
-    span {
-      font-family: Proxima Nova;
-      padding: 24px 0px 0px;
-      margin-left: 0px;
-      width: 100%;
-    }
+  .heading.-banner.legacy-step-header span {
+    font-family: $font-proxima-nova;
+    padding: 24px 0px 0px;
+    margin-left: 0px;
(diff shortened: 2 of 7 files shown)


resources/assets/components/ActionPage/actionPage.scss
```diff
@@ -62,6 +62,18 @@
     }
   }
 
+  .heading.-banner.legacy-step-header {
```
deadlybutter: Hm let's remove the extra linebreak here. For the selector I think we can just make this `.legacy-step-header span`. As for the styling,

- replace `proxima-nova` with [this variable]($font-proxima-nova: "Proxima Nova", "Helvetica Neue", "Helvetica", "Arial", sans-serif;)
- if you need that specific padding format, use `$base-spacing` instead of 24px. However Im not sure you need this rule, is this overriding something or can you just use `padding-top`? Same for margin left, is that overriding something?

resources/assets/components/ActionPage/templates/LegacyStepHeaderTemplate.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FlexCell } from '../../Flex';
+
+const LegacyStepHeaderTemplate = ({ title, step }) => (
+  <FlexCell width="full">
+    <h2 className="heading -banner legacy-step-header">
+      <span>
+        Step { String(step) }: {title}
```
deadlybutter: Hm, not crucial but this would probably read better a template literal

deadlybutter: Also action steps don't always have a step number, there is a Contentful setting to disable that, so make sure it's not a required field and youre handling that if it's null. Checkout the Mosaic version which does this. Also did the mock want the actual step number or did it want the word version like the Mosaic? I don't remember from the mock

resources/assets/components/ActionPage/templates/MosaicStepHeaderTemplate.js
```diff
@@ -0,0 +1,28 @@
...
+    <PhotoHeader className="action-step__header" backgroundImage={background}>
+      { hideStepNumber ? null : <span>step { convertNumberToWord(step) }</span> }
+      <h1>{ title }</h1>
+    </PhotoHeader>
+  </FlexCell>
+);
+
+MosaicStepHeaderTemplate.propTypes = {
+  title: PropTypes.string.isRequired,
+  step: PropTypes.number.isRequired,
```
deadlybutter: Again i wouldn't explicitly require step number here either, also `hideStepNumber` shouldnt be required, and make the default prop false.

deadlybutter:
I also just took a peak at the build failures. I would grab a plugin for your IDE that supports editor config so you're not worrying about the spacing & indents!

resources/assets/components/ActionPage/actionPage.scss
mendelB: The `padding`, `margin-left`, and `width` are overriding some styling set for `.heading.-banner span`s (I think from forge)

resources/assets/components/ActionPage/templates/LegacyStepHeaderTemplate.js
mendelB: I know that the Drupal site shows the actual number, not sure about the mock though

mendelB:
Thanks, @deadlybutter for the feedback! changed whatever I could based on your suggestions (save for the styling overriding the forge stuff), also added the editorConfig plugin to sublime, so hopefully, that'll prevent the styling snafus.

resources/assets/components/ActionPage/actionPage.scss
deadlybutter: Gotcha, ok! Also just confirmed with Luke in Slack, let's not use Proxima Nova here. You can use `$primary-font-family` which comes from the [Phoenix next toolbox](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/scss/next-toolbox.scss). Feel free to ping Luke or Ashley in the future if you have a design or product question!

deadlybutter: Also even though it's an override I would still replace the `24px` with `$base-spacing`

deadlybutter:
I think besides those quick style changes everything is good to go, but let's make sure @weerd gets a chance to weigh in as well

weerd:
> also added the editorConfig plugin to sublime, so hopefully, that'll prevent the styling snafus.

Yeah, I found that sometimes if editorconfig is added later it doesn't fix the files already updated, but will moving forward!

mendelB:
> Yeah, I found that sometimes if editorconfig is added later it doesn't fix the files already updated, but will moving forward!

Ahaaa, that would explain why it was only affecting certain files for me which was confusing.

resources/assets/components/ActionPage/StepHeader.js
```diff
@@ -1,24 +1,37 @@
...
-const StepHeader = ({ title, step, background, hideStepNumber }) => (
-  <FlexCell width="full">
-    <PhotoHeader className="action-step__header" backgroundImage={background}>
-      { hideStepNumber ? null : <span>step { convertNumberToWord(step) }</span> }
-      <h1>{ title }</h1>
-    </PhotoHeader>
-  </FlexCell>
-);
+const StepHeader = ({ title, step, background, hideStepNumber, template }) => {
+  switch (template) {
```
weerd: 🎉

resources/assets/components/ActionPage/actionPage.scss
```diff
@@ -62,6 +62,15 @@
     }
   }
 
+  .heading.-banner.legacy-step-header span {
+    font-family: $primary-font-family;
+    padding: $base-spacing 0px 0px;
```
weerd: no need for unit specification when `0` is used as the value. `0` is unit-less 💃

resources/assets/components/ActionPage/templates/LegacyStepHeaderTemplate.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FlexCell } from '../../Flex';
+
+const LegacyStepHeaderTemplate = ({ title, step, hideStepNumber }) => (
+  <FlexCell width="full">
+    <h2 className="heading -banner legacy-step-header">
```
weerd: we've deprecated the use of `-banner`. Use `-emphasized` instead! http://forge.dosomething.org/#heading :)

resources/assets/components/ActionPage/actionPage.scss
```diff
@@ -62,6 +62,15 @@
     }
   }
 
+  .heading.-banner.legacy-step-header span {
```
weerd: It's a better practice not to overly specify a selector in CSS, because if we later need to override this, it makes it a lot more difficult! I think you'd be fine with just `.legacy-step-header span` right?

resources/assets/components/ActionPage/templates/LegacyStepHeaderTemplate.js
mendelB: Oh nice! That seems to eliminate the need for any of the custom CSS on the `span`, save for the fact that the underline recedes due to a `width` reduction:
image
adding `width: 100%;` to the `h3` brings it back though, and I could use a better selector as per your comment below?
```
.legacy-step-header {
    width: 100%;
  }
```
resulting in:
image

resources/assets/components/ActionPage/actionPage.scss
deadlybutter: Oh yeah I left that in one of the other comments as well

weerd: Also, I believe you should be able to just add this styling to the `h2` instead of specifying the `span`. In the legacy Phoenix-Ashes site we wrapped the text in a `span` for safe measure since adding the `::after` pseudo element for the yellow bar 😀

resources/assets/components/ActionPage/templates/LegacyStepHeaderTemplate.js
weerd: yeah, much preferred!

weerd:
@mendelB just confirming, does it look ok on smaller screens?

mendelB:
I toggled around on developer console through the various breakpoints and it seems to play nicely with all of em.

mendelB:
Thanks @weerd @deadlybutter for all the help and feedback!!! 

Am I meant to do the merge?

weerd:
Go for it @mendelB!

PR #395, opened 2017-09-18
Enabled all routes and adds LandingPage component

weerd:
### What does this PR do?
This PR does a bit of refactoring for the general routing of the client-side app to allow for non-campaign pages which will be coming in the near future 🔮! It also sets up the ability to allow providing a Pitch Landing Page for campaigns if one is provided and a user is not affiliated with the campaign.

I apologize ahead of time for the large number of files 😬, but it was a bit unavoidable due to the nature of having to change up the routing logic a bunch.

The content and styling for the pitch landing page will be coming in a separate PR. There are also a few components like the `Chrome`, `Wrapper` and `FeedEnclosure` which can now be deleted, but waiting to do so after this is merged and a few other PRs are merged in as well to avoid conflicts.

Fixes https://www.pivotaltracker.com/story/show/150984393

### Checklist


resources/assets/components/App.js
```diff
@@ -1,52 +1,28 @@
...
-          <Route path={`${paths.blocks}:id`} component={chrome(BlockContainer)} />
-          <Route path={`${paths.quiz}:slug`} component={chrome(QuizContainer)} />
-          {/* * */}
-
-          {/* Custom user experiences... */}
-
-          {/* 404 default */}
-          <Route component={chrome(NotFound)} />
-          {/* * */}
+          { (! isAffiliated && useLandingPage) ?
```
itsjoekent: As I mentioned in Slack we should move this <switch> to a component that's wrapped with a container. Otherwise `isAffiliated` might be false initially but turn true after a few seconds (when it hits phoenix ashes for data)

DFurnes:
screen shot 2017-09-18 at 2 54 54 pm

Look what you've become. 👹

resources/assets/scss/base.scss
```diff
@@ -111,10 +115,33 @@ h1, h2, h3, p {
...
+
 .rounded {
   border-radius: $lg-border-radius;
 }
 
 .faded {
   opacity: 0.5;
 }
+
+.col-default {
```
weerd: Wasn't sure if this made sense... Just wanted the default main grid containment with the responsive breakpoints. Wasn't entirely sure what to call it...

resources/assets/components/Campaign/CampaignContainer.js
```diff
@@ -0,0 +1,10 @@
+import { connect } from 'react-redux';
+
+import Campaign from './Campaign';
+
+const mapStateToProps = state => ({
+  isAffiliated: state.signups.thisCampaign,
+  useLandingPage: state.campaign.landingPage !== null,
```
itsjoekent: This feels weird comparing it to null?

weerd:
@DFurnes I've turned into you :trollface: mwahahahaha

resources/assets/components/Campaign/CampaignContainer.js
itsjoekent: This feels weird comparing it to null?

resources/assets/components/Enclosure.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+const Enclosure = ({ children, className = '' }) => (
```
itsjoekent: I sat on it for a bit, I guess this makes sense. makes things slightly less tedious

DFurnes: I like this name. 📜👏

resources/assets/components/Page/LandingPage/LandingPageContainer.js
```diff
@@ -0,0 +1,33 @@
+import { connect } from 'react-redux';
+
+import LandingPage from './LandingPage';
+import { clickedSignUp, convertExperiment } from '../../../actions';
```
itsjoekent: It doesn't seem like we are using `convertExperiment` anywhere

itsjoekent: This import is on the other page container as well

resources/assets/components/Page/CampaignSubPage/CampaignSubPage.js
```diff
@@ -0,0 +1,81 @@
...
+import PropTypes from 'prop-types';
+
+import NotFound from '../../NotFound';
+import Markdown from '../../Markdown';
+import ScrollConcierge from '../../ScrollConcierge';
+import CallToActionBlockContainer from '../../../containers/CallToActionBlockContainer';
+
+import './campaign-subpage.scss';
+
+const CampaignSubPage = (props) => {
```
DFurnes: submarine0113 pngbe41a9a5-7c6b-4654-818b-35fcf8cfed2coriginal

resources/assets/components/Page/Page.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const Page = () => (
+  <div>Generic Page</div>
```
DFurnes: Is this still needed, or just for testing?

resources/assets/containers/TabbedNavigationContainer.js
```diff
@@ -28,9 +28,11 @@ const TabbedNavigationContainer = (props) => {
 
   const isClosed = isCampaignClosed(campaignEndDate);
 
+  const campaignSlug = props.campaignSlug;
+
   // Create links for additional "content" pages on this campaign in Contentful.
   const additionalPages = pages.map((page) => {
-    const path = `${paths.pages}${page.fields.slug}`;
+    const path = `/us/campaigns/${campaignSlug}${campaignPaths.pages}${page.fields.slug}`;
```
DFurnes: I'm thinking we should start using `path.join` here, so we don't have to worry about `/`s lining up:

```js
import { join } from 'path';

// ...

const path = join('/us/campaigns', campaignSlug, campaignPaths.pages, page.fields.slug);
```

resources/assets/components/Enclosure.js
weerd: Yeah, we were using `Wrapper` and `FeedEnclosure` together to achieve the same result as just `Enclosure`. Plus, we were using `FeedEnclosure` on ALL pages, like the FAQ and Scholarships, not just the Feed 😱

resources/assets/components/Page/LandingPage/LandingPageContainer.js
weerd: Oh nice. Yeah, I was copy/pasting some from old ones.

weerd: Actually... wasn't sure whether to leave these since we'll be doing experiments 🤔

resources/assets/history.js
```diff
@@ -18,9 +16,7 @@ export function get() {
  * @return {History}
  */
 export function init() {
-  // Set the application "base name" to /us/campaigns/:slug so all pages are relative to that.
-  const basename = window.location.pathname.split('/').slice(0, 4).join('/');
```
DFurnes: Just want to double-check that we're still getting routed to server-rendered pages with this change, yeah?

resources/assets/components/Campaign/CampaignContainer.js
weerd: Yeah, I just wanted to turn it into a boolean, so when used in component logic it worked a bit better than comparing agains either `null` or a large object, which felt cleaner.

weerd: for context, the value of `landingPage` is either `null` or a typical Contentful object.

resources/assets/scss/base.scss
DFurnes: I'm into it. I kinda want to name it as it's own standalone thing though, so that `.col-` can be a prefix for different column classes in Forge 7. Maybe just `.page-container` or `.default-container` or something like that?

resources/assets/components/Page/Page.js
weerd: Potentially for static/generic pages. Not sure yet. Figured I'd leave in for now... 🔮

resources/assets/history.js
weerd: Which server rendered pages? `/us/campaigns` is the only one right?

resources/assets/components/Campaign/CampaignContainer.js
itsjoekent: I see now, I think this makes sense

resources/assets/history.js
DFurnes: Not when this is running on www.dosomething.org though - we want to make sure people can still follow links (like in the header or footer) to the homepage, non-Next campaigns, or static pages.

weerd: Oh yeah, I believe so. I checked out a bunch and all seemed to work fine, but lemme know if there are specific ones to double-check!

PR #397, opened 2017-09-19
Updates to remove no longer used components.

weerd:
### What does this PR do?
This PR removes a series of components that are no longer in use after the updates that happened in #395.


itsjoekent:
Can we merge this after the Nuke PR? I think this one is a lot easier to rebase

PR #399, opened 2017-09-19
Campaign footer additions

mendelB:
### What does this PR do?
- Adds `CampaignFooter` component to `CampaignPage`
- Using utility style classes for styling `CampaignFooter` instead of `wrapper` class
- Using Campaign Lead name and/or email for the 'Contact Us' section in `CampaignFooter`

(Plase let me know if I should pull the 'Contact Us' stuff out into its own component. I wasn't positive it warranted that in this case. I'm also not positive that I had the right approach regarding the `defaultProps` spiel for the `campaignLead`)

Additions to previous pull request for Campaign Footer.

https://www.pivotaltracker.com/story/show/151131919

### DIFF
--- a/resources/assets/components/CampaignFooter/CampaignFooter.js
+++ b/resources/assets/components/CampaignFooter/CampaignFooter.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import AffiliateCredit from '../AffiliateCredit';
 
-const CampaignFooter = ({ affiliateSponsors, affiliatePartners }) => {
+const CampaignFooter = ({ affiliateSponsors, affiliatePartners, campaignLead }) => {
   // @TODO Either deal with potentially supporting multiple affiliates per campaign
   // as opposed to just grabbing the first one
   // or change fields in Contentful to reflect single affiliates per campaign.
@@ -23,7 +23,7 @@ const CampaignFooter = ({ affiliateSponsors, affiliatePartners }) => {
             /> : null
         }
         <div className="info-bar__secondary">
-          Questions? <a href="mailto:help@dosomething.org">Contact Us</a>
+          Questions? <a href={`mailto:${campaignLead.email}`}>Contact {campaignLead.name}</a>
         </div>
       </div>
     </footer>
@@ -33,6 +33,17 @@ const CampaignFooter = ({ affiliateSponsors, affiliatePartners }) => {
 CampaignFooter.propTypes = {
   affiliateSponsors: PropTypes.array.isRequired, // eslint-disable-line react/forbid-prop-types
   affiliatePartners: PropTypes.array.isRequired, // eslint-disable-line react/forbid-prop-types
+  campaignLead: PropTypes.shape({
+    name: PropTypes.string,
+    email: PropTypes.string,
+  }),
+};
+
+CampaignFooter.defaultProps = {
+  campaignLead: {
+    name: 'Us',
+    email: 'help@dosomething.org',
+  },
 };
 
 
--- a/resources/assets/components/CampaignFooter/CampaignFooter.js
+++ b/resources/assets/components/CampaignFooter/CampaignFooter.js
@@ -14,7 +14,7 @@ const CampaignFooter = ({ affiliateSponsors, affiliatePartners }) => {
 
   return (
     <footer className="info-bar">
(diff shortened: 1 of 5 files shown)


resources/assets/components/CampaignFooter/CampaignFooter.js
```diff
@@ -33,6 +33,17 @@ const CampaignFooter = ({ affiliateSponsors, affiliatePartners }) => {
 CampaignFooter.propTypes = {
   affiliateSponsors: PropTypes.array.isRequired, // eslint-disable-line react/forbid-prop-types
   affiliatePartners: PropTypes.array.isRequired, // eslint-disable-line react/forbid-prop-types
+  campaignLead: PropTypes.shape({
```
itsjoekent: This seems fine to me!

resources/assets/components/Page/CampaignPage/CampaignPage.js
```diff
@@ -100,6 +114,7 @@ CampaignPage.defaultProps = {
   endDate: null,
   isAffiliated: false,
   totalCampaignSignups: 0,
+  campaignLead: undefined,
```
itsjoekent: Make this `null`

mendelB: I did that so that the `defaultProps` within the `CampaignFooter` would kick in in case the `CampaignPage` doesn't receive a `campaignLead`. I believe `null` doesn't trigger that?

itsjoekent: `null` will indeed still trigger the default props 😄

mendelB: oh 😳 hmm.... I was bumping into an error but must have been something else. I'll triple check that I didn't mess anything up

weerd: `null` doesn't trigger the default props....

resources/assets/components/Page/CampaignPage/CampaignPage.js
```diff
@@ -13,11 +13,13 @@ import LedeBanner from '../../LedeBanner/LedeBanner';
 import { ActionPageContainer } from '../../ActionPage';
 import { CampaignSubPageContainer } from '../CampaignSubPage';
 import TabbedNavigationContainer from '../../../containers/TabbedNavigationContainer';
+import CampaignFooter from '../../CampaignFooter';
 
 const CampaignPage = (props) => {
   const {
     blurb, clickedSignUp, coverImage, dashboard, endDate, isAffiliated,
     legacyCampaignId, match, slug, subtitle, template, title, totalCampaignSignups,
+    affiliateSponsors, affiliatePartners, campaignLead,
```
weerd: Mind alphabetizing the newly added ones. Found it makes it easier to visually locate if in alpha order and knowing where to find them :)

mendelB: My disorganized mind completely missed that!

resources/assets/components/Page/CampaignPage/CampaignPage.js
weerd: That being said, you aren't setting a default property in the component:

```js
const CampaignFooter = ({ affiliateSponsors, affiliatePartners, campaignLead }) => { ... }
```

Setting it as `undefined` would work if you do something like this:
```js
const CampaignFooter = ({ affiliateSponsors, affiliatePartners, campaignLead = 'Puppet Slot' }) => { ... }
```

itsjoekent: My bad! Diego is totally write here

mendelB: For some reason, unless I have `undefined` as the `defaultProp` in `CampaignPage`, the `CampaignFooter` blows up when a `null` value is passed as a `prop` from the `CampaignPageContainer`?

mendelB: Spoke with Diego. His professional opinion was to stick with `undefined` for now to deal with missing `campaignLead` info - to trigger the `defaultProps`. (After taking care of potential missing `additionalContent` prop with `get`)

weerd:
post-merge 👍

PR #400, opened 2017-09-19
Installing Puck

itsjoekent:
### What does this PR do?
This PR installs Puck, adds the required config and sets up the provider so it can start tracking page-related events.

screen shot 2017-09-19 at 2 54 15 pm

The next step is going to be tracking signups + shares.


resources/assets/components/App.js
```diff
@@ -12,9 +13,15 @@ const App = ({ store, history }) => {
 
   return (
     <Provider store={store}>
-      <ConnectedRouter history={history}>
-        <Route path="/us/campaigns/:slug" component={CampaignContainer} />
-      </ConnectedRouter>
+      <PuckProvider
+        source="phoenix-next"
+        getUser={() => store.getState().user.id}
+        getHistory={() => history}
```
DFurnes: What's the advantage of the callback here (rather than `history={history}`?

resources/assets/components/App.js
```diff
@@ -12,9 +13,15 @@ const App = ({ store, history }) => {
 
   return (
     <Provider store={store}>
-      <ConnectedRouter history={history}>
-        <Route path="/us/campaigns/:slug" component={CampaignContainer} />
-      </ConnectedRouter>
+      <PuckProvider
+        source="phoenix-next"
+        getUser={() => store.getState().user.id}
```
DFurnes: Not necessary for this PR, but it'd be nice to make a selector function for this since we do it in a bunch of places (so something like `getUserId(store.getState())` from a module in `selectors/`.

app/helpers.php
```diff
@@ -273,12 +273,11 @@ function phoenixLink($path)
 function get_client_environment_vars()
 {
     return [
-        'KEEN_PROJECT_ID' => config('services.analytics.keen_id'),
-        'KEEN_WRITE_KEY' => config('services.analytics.keen_key'),
         'SIXPACK_ENABLED' => config('services.sixpack.enabled'),
         'SIXPACK_BASE_URL' => config('services.sixpack.url'),
         'SIXPACK_COOKIE_PREFIX' => config('services.sixpack.prefix'),
         'SIXPACK_TIMEOUT' => config('services.sixpack.timeout'),
+        'PUCK_URL' => config('services.puck_url'),
```
DFurnes: What's the behavior like if this server is not available? (e.g. on local dev, or in case of downtime)

itsjoekent: Just a `console.log`

resources/assets/components/App.js
itsjoekent: I can do that now, not a big deal

resources/assets/components/App.js
itsjoekent: True yeah the history probably doesn't need a getter

resources/assets/helpers/index.js
```diff
@@ -364,3 +364,13 @@ export function isCampaignClosed(endDate) {
...
   return array.find(({ id }) => id === compareId);
 }
+
+/**
+ * Get a value from the window.ENV
+ *
+ * @param  {String} key
+ * @return {String}
+ */
+export function env(key) {
```
DFurnes: Nice! 👏

.env.example
```diff
@@ -56,3 +56,4 @@ PUSHER_SECRET=
 GOOGLE_ANALYTICS_ID=
 FACEBOOK_APP_ID=
 CUSTOMER_IO_ID=
+PUCK_URL=
```
weerd: It's happening!

itsjoekent:
latest updates pushed up, here is the console output when you're clicking between pages

screen shot 2017-09-19 at 3 48 33 pm

itsjoekent:
And when you start a new session (for this example, setting the heartbeat ticker to 0), you also get the view event before anything else

screen shot 2017-09-19 at 3 50 54 pm

resources/assets/components/App.js
```diff
@@ -3,18 +3,28 @@ import PropTypes from 'prop-types';
...
   return (
     <Provider store={store}>
-      <ConnectedRouter history={history}>
-        <Route path="/us/campaigns/:slug" component={CampaignContainer} />
-      </ConnectedRouter>
+      <PuckProvider
+        source="phoenix-next"
+        getUser={() => getUserId(store.getState())}
+        getHistory={() => history}
+        puckUrl={env('keen_url')}
```
DFurnes: Should this be `env('puck_url')`?

resources/assets/components/App.js
```diff
@@ -3,18 +3,28 @@ import PropTypes from 'prop-types';
...
   initializeStore(store);
 
   return (
     <Provider store={store}>
-      <ConnectedRouter history={history}>
-        <Route path="/us/campaigns/:slug" component={CampaignContainer} />
-      </ConnectedRouter>
+      <PuckProvider
+        source="phoenix-next"
+        getUser={() => getUserId(store.getState())}
```
weerd: If viewing page anonymously, is `getUser` fine as `null`? Also, can we call it `user` 😬

resources/assets/components/App.js
```diff
@@ -3,18 +3,28 @@ import PropTypes from 'prop-types';
...
+        getHistory={() => history}
```
weerd: > What's the advantage of the callback here (rather than history={history}? from @DFurnes 

The comment got hidden but the change wasn't made 🤔

resources/assets/components/App.js
DFurnes: If it needs to be a callback, I think having that `get` prefix makes it a bit more clear. Ideally it'd be nice to just pass a value though, especially since I don't think we'll ever be doing logins without a page reload.

weerd: oh sorry, I meant change the prop name from `getUser` to `user` not the function name @DFurnes!

DFurnes: I know! I mean that naming the prop `getUser` makes it more clear to me that it's asking for a function that lets it "get the user", rather than a specific user string or object.

resources/assets/selectors/user.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Get the user id from the state.
+ *
+ * @param state
+ * @returns {String}
+ */
+export function getUserId(state) { // eslint-disable-line import/prefer-default-export
```
weerd: why is the comment needed? not sure I understand why it's mad. We have helper files that don't have this eslint comment after each function 🤔

itsjoekent: It's the only function in the file

weerd: oh so until you add another it's angry, haha

itsjoekent: yarp. 🙄

resources/assets/components/App.js
weerd: Hmmm, oh ok, yeah since it's returning an anonymous function and not the actual ID I guess that makes sense. Still feels weird but not a merge blocker!

itsjoekent:
Everything seems to be piping through correctly in Puck staging (It's in the Heroku team for anyone curious)

PR #401, opened 2017-09-19
Fix for broken campaign update links that use old component.

weerd:
### What does this PR do?
This PR fixes an issue where the link generated by the `BlockWrapper` used by the old `CampaignUpdateBlock` component would not be the full link with the campaign and campaign slug, thus the specific block page would end up being broken.

Fixes item 11 on https://docs.google.com/spreadsheets/d/1pgnbX3eQDjWnvDPe1Afie4xyEGgwMdW7eXMXyNNs3IQ/edit#gid=0

### DIFF
--- a/resources/assets/components/CampaignUpdateBlock/CampaignUpdateBlock.js
+++ b/resources/assets/components/CampaignUpdateBlock/CampaignUpdateBlock.js
@@ -14,7 +14,7 @@ const CampaignUpdateBlock = (props) => {
   const isTweet = content.length < 144;
 
   return (
-    <BlockWrapper title="Campaign Update" id={props.id} shareLink={props.shareLink} >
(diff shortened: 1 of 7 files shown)


resources/assets/components/CampaignUpdateBlock/CampaignUpdateBlock.js
```diff
@@ -14,7 +14,7 @@ const CampaignUpdateBlock = (props) => {
   const isTweet = content.length < 144;
 
   return (
-    <BlockWrapper title="Campaign Update" id={props.id}>
+    <BlockWrapper title="Campaign Update" id={props.id} shareLink={props.shareLink} >
```
itsjoekent: Did you want to remove ID here as well?

weerd: oh, good catch. yep!

PR #402, opened 2017-09-20
Path bug fix

weerd:
### What does this PR do?
This PR fixes the post-signup return to the action page to use the correct full campaign path which includes the campaign slug.

It also updates the `ActionSteps` to _not_ require the `content` property and _not_ output anything if the Campaign Lead adds no content. This field is also _not_ set to be required in Contentful so it didn't make sense for it to be required in the component if the field could be left empty.

https://docs.google.com/spreadsheets/d/1pgnbX3eQDjWnvDPe1Afie4xyEGgwMdW7eXMXyNNs3IQ/edit#gid=0

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -1,5 +1,7 @@
+import { join } from 'path';
 import { get } from 'lodash';
 import { push } from 'react-router-redux';
+
 import { Phoenix } from '@dosomething/gateway';
 import { isCampaignClosed } from '../helpers';
 import {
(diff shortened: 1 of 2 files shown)


resources/assets/actions/signup.js
```diff
@@ -104,14 +106,16 @@ export function getTotalSignups(campaignId) {
 // check if the user is logged in or has an existing signup.
 export function clickedSignUp(campaignId, shouldRedirectToActionTab = true) {
   return (dispatch, getState) => {
+    const campaignActionUrl = join('/us/campaigns', getState().campaign.slug, '/action');
```
itsjoekent: Should this be another selector?

weerd: oh it could be... can be a post update in future PR.

PR #403, opened 2017-09-20
Sponsor logo in lede banner

mendelB:
### What does this PR do?
- Adds CampaignSponsorPromotion component to credit campaign sponsor
- CampaignSponsorPromotion component rendered in `LedeBanner` if sponsor is present
- CampaignSponsorComponent will be rendered only in `LegacyTemplate`
- Adds `Affiliate` entity to define props needed for rendering sponsor logos, along with other the presently consumed properties by the react app. This entity is used by both affiliateSponsor and affiliatePartner props

The sponsor promo is styled by importing styling used in ashes
https://github.com/DoSomething/phoenix/blob/dev/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_promotion.scss 

 As a sponsor, I want to be able to see my logo on the ledebanner of a legacy campaign

https://www.pivotaltracker.com/story/show/151129347

### DIFF
--- a/app/Entities/Affiliate.php
+++ b/app/Entities/Affiliate.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace App\Entities;
+
+use JsonSerializable;
+
+class Affiliate extends Entity implements JsonSerializable
+{
+    /**
+     * Convert the object into something JSON serializable.
+     *
+     * @return array
+     */
+    public function jsonSerialize()
+    {
+        return [
+            'id' => $this->entry->getId(),
+            'type' => $this->getContentType(),
+            'fields' => [
+                'link' => $this->link,
+                'title' => $this->title,
+                'logo' => [
+                    'url' => get_image_url($this->logo, 'logo'),
+                    'title' => $this->logo->getTitle(),
+                ],
+            ],
+        ];
+    }
+}
--- a/resources/assets/components/CampaignSponsorPromotion/CampaignSponsorPromotion.js
+++ b/resources/assets/components/CampaignSponsorPromotion/CampaignSponsorPromotion.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './CampaignSponsorPromotion.scss';
+
+const CampaignSponsorPromotion = ({ imgUrl, title }) => (
+  <div className="promotions">
+    <div className="promotion promotion--sponsor">
+      <div className="wrapper">
+        <p className="__copy">Powered by</p>
+        <div className="__image">
+          <img src={imgUrl} alt={title} />
+        </div>
+      </div>
+    </div>
+  </div>
+);
+
+CampaignSponsorPromotion.propTypes = {
+  title: PropTypes.string,
+  imgUrl: PropTypes.string.isRequired,
+};
+
+CampaignSponsorPromotion.defaultProps = {
+  title: 'Campaign Sponsor Logo',
+};
+
+export default CampaignSponsorPromotion;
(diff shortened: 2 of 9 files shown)


resources/assets/components/CampaignSponsorPromotion/CampaignSponsorPromotion.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './CampaignSponsorPromotion.scss';
+
+const CampaignSponsorPromotion = ({ imgUrl, title }) => (
+  <div className="promotions">
+    <div className="promotion promotion--sponsor">
+      <div className="wrapper">
+        <p className="__copy">Powered by</p>
```
itsjoekent: I know this was the old markup, but I'm not convinced all of this is necessary. But also not super important or merge blocking

mendelB: Yeah fair enough. I kept it to jive with the styling which was based on this layout. I could technically pour through it and try and reorganize if you think it's a good idea?

resources/assets/components/CampaignSponsorPromotion/CampaignSponsorPromotion.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './CampaignSponsorPromotion.scss';
+
+const CampaignSponsorPromotion = ({ imgUrl, title }) => (
```
weerd: I'm debating whether we just call this `SponsorPromotion` and it just happens to be in the campaign. So it affords eventually adding sponsor promotions elsewhere... thoughts?

mendelB: I think that def makes sense.

resources/assets/components/CampaignSponsorPromotion/CampaignSponsorPromotion.js
weerd: @joe defs good point, but something to focus on post getting Legacy stuff moved over. There's going to be a whole revamp of styles with Forge 7, so those updates can be scooped up with that.

app/Entities/Affiliate.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace App\Entities;
+
+use JsonSerializable;
+
+class Affiliate extends Entity implements JsonSerializable
```
weerd: 🎉

resources/assets/components/LedeBanner/templates/LegacyTemplate.js
```diff
@@ -11,12 +13,15 @@ const LegacyTemplate = (props) => {
...
     clickedSignUp,
     endDate,
+    affiliateSponsors,
   } = props;
 
   const backgroundImageStyle = {
     backgroundImage: `url(${backgroundImageUrl})`,
   };
 
+  const sponsor = affiliateSponsors[0];
```
weerd: Mind adding a quick
```
// @TODO: consider whether there can be more than one affiliate, or 
// whether to grab single entry when transforming in PHP

mendelB: Will do!

Oh and you know what - I just found this:
https://thor.dosomething.org/us/campaigns/give-spit-about-cancer?source=campaigns

so def worth revisiting and adding support for at least two?

weerd: Aha! interesting. Good find!

PR #404, opened 2017-09-21
Modal rework

itsjoekent:
### What does this PR do?
Stopping myself before this gets bigger 😬 

This PR will let us do a few custom modal types without making the Contentful build extreme. This PR itself was just making the modal more flexible, the next one will be rendering a given contentful ID from the `modal/${id}` route.

Basically, we now have a `Modal Switch` component that determines what components are injected into the `Modal` component. I added the PostSignup version so we retain that functionality, and the next option will be for rendering the given Contentful ID.

### DIFF
--- a/resources/assets/components/Modal/Modal.js
+++ b/resources/assets/components/Modal/Modal.js
@@ -1,11 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Portal from 'react-portal';
-import { closeModal } from '../../actions';
 import './modal.scss';
-
-import { AffirmationContainer } from '../Affirmation';
-import CompetitionContainer from '../../containers/CompetitionContainer';
 
 class Modal extends React.Component {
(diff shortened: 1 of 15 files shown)


resources/assets/components/Modal/Modal.js
```diff
@@ -23,28 +19,13 @@ class Modal extends React.Component {
...
-      children.push(
-        <CompetitionContainer
-          key="competition"
-          content={competitionStep.content}
-          photo={competitionStep.photos[0]}
-          byline={competitionStep.additionalContent}
-        />,
-      );
-    }
+    const { shouldShowModal, children } = this.props;
```
weerd: nice! much cleaner 💃

itsjoekent:
@weerd I pushed an update to the campaign update block test, it was breaking the test suite. Not sure why it decided to kick in now /shrug

PR #405, opened 2017-09-21
Adding ability to display a page in a modal

itsjoekent:
### What does this PR do?
screen shot 2017-09-21 at 3 43 10 pm

This PR lets you route to `/us/campaigns/test-teens-for-jeans/modal/24tJwKV9WgMoSmIcg2IgcE` which is the contentful id for the FAQ page of that campaign. 

As I noted in the comments if you link to a modal from a page other than the root it's going to annoy users because the page under the modal will be the root page. For the legacy experience, this should be fine, however, since we don't have tabs.

I also renamed `blockId` to `contentfulId` since it's not specific to blocks or pages technically

### DIFF
--- a/resources/assets/components/Modal/ModalSwitch.js
+++ b/resources/assets/components/Modal/ModalSwitch.js
@@ -1,6 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Modal, PostSignupModal, POST_SIGNUP_MODAL } from '../Modal';
+import {
+  Modal, PostSignupModal, PageModal,
+  POST_SIGNUP_MODAL, PAGE_MODAL,
+} from '../Modal';
 
 const ModalSwitch = (props) => {
   const { modalType } = props;
@@ -8,7 +11,7 @@ const ModalSwitch = (props) => {
 
   switch (modalType) {
     case POST_SIGNUP_MODAL: children = <PostSignupModal />; break;
-    // TODO: <Page Modal>
+    case PAGE_MODAL: children = <PageModal />; break;
     default: break;
   }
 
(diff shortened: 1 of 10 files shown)


resources/assets/components/Modal/ModalSwitch.js
```diff
@@ -1,14 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Modal, PostSignupModal, POST_SIGNUP_MODAL } from '../Modal';
+import {
+  Modal, PostSignupModal, PageModal,
```
weerd: I like that we can break these out into respective `*Modal` components :)

resources/assets/components/Modal/configurations/PageModal.js
```diff
@@ -0,0 +1,20 @@
...
+import Markdown from '../../Markdown';
+
+const PageModal = ({ content }) => (
+  <Card className="bordered padded modal__slide">
+    <Markdown>{ content }</Markdown>
+  </Card>
+);
+
+PageModal.propTypes = {
+  content: PropTypes.string,
```
weerd: Should this be required? If there's no `content` not really anything to show and have an empty `Card` 😬

itsjoekent: The modal won't show if it can't find the contentful ID, this is specific to what gets injected in the Markdown component

weerd: Yeah, but since `content` is the only thing we're showing, we don't really want to show an empty modal right? Likely not to happen, but this way we enforce to pass actually value in content instead of default to empty string. 🤔

resources/assets/containers/CompetitionContainer.js
```diff
@@ -1,6 +1,6 @@
 import { connect } from 'react-redux';
 import CompetitionBlock from '../components/CompetitionBlock';
-import { joinCompetition, checkForCompetition } from '../actions';
+import { joinCompetition, checkForCompetition } from '../actions/competition';
```
weerd: Ya, this is weird why it changed... 🤔

PR #406, opened 2017-09-22
Hide on legacy

itsjoekent:
### What does this PR do?
This PR 

- Hides the tabbed navigation on legacy
- Shows action page instead of community on legacy

screen shot 2017-09-22 at 10 11 12 am

Not really


resources/assets/components/Page/CampaignPage/CampaignPage.js
```diff
@@ -52,7 +52,15 @@ const CampaignPage = (props) => {
 
         <Enclosure className="margin-top-lg margin-bottom-lg default-container">
           <Switch>
-            <Route path={`${match.url}`} exact component={FeedContainer} />
+            <Route
```
weerd: Nice 💃

PR #409, opened 2017-09-22
Signup tracking

itsjoekent:
### What does this PR do?
This PR introduces a new HOC for wrapping signup buttons. This ensures we have uniform functionality & tracking across all our conversion points. [Here is the file for that](https://github.com/DoSomething/phoenix-next/compare/signup-tracking?expand=1#diff-c8160e3f6f48e96c64cb8c090de3d2c3)

Everything else is removing `clickedSignupUp` implementations where necessary and updating with this new component.

The only places that haven't been updated yet for this functionality are very custom implementations of clickedSignup such the Revealer component and the Quiz. I'll have to tackle that next.

The HOC lets us pass in whatever markup and keeps all of the underlying data stable. It's a few extra lines of code for a lot of added benefit! 

screen shot 2017-09-22 at 11 29 11 am


weerd:
> The HOC lets us pass in whatever markup and keeps all of the underlying data stable. It's a few extra lines of code for a lot of added benefit! ™

resources/assets/components/SignupButton/SignupButton.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { PuckConnector } from '@dosomething/puck-client';
+import SignupButtonContainer from './SignupButtonContainer';
+
+/**
+ * HOC for wrapping signup buttons. Handles tracking
+ * and dispatching the signup action.
```
weerd: so cool!

resources/assets/components/SignupButton/SignupButtonContainer.js
```diff
@@ -0,0 +1,8 @@
+import { connect } from 'react-redux';
+import { clickedSignUp } from '../../actions/signup';
+
+const actionCreators = {
+  clickedSignUp,
+};
+
+export default Component => connect(null, actionCreators)(Component);
```
weerd: Mind if we continue with the pattern of `mapActionsToProps` for the name instead of `actionCreators`?

I've been slowly renaming other containers that don't follow the pattern for `mapStateToProps` and `mapActionsToProps`, since it sounds a bit more straight-forward.

itsjoekent: Ah yes, totes fine!

DFurnes: This is maybe nitpicky, but I feel like `mapActionsToProps` implies that it's a function. Since we're using the action creator object shorthand in most places, it seems appropriate to follow [the example in the docs](https://github.com/reactjs/react-redux/blob/master/docs/api.md#inject-all-action-creators--addtodo-completetodo--without-subscribing-to-the-store).

weerd: oh interesting. I didn't realize it was like that in the docs. I think I had seen a diff example elsewhere maybe? The reasoning makes sense to me. We can keep it `actionCreators` and then update other containers. 😬

itsjoekent:
@weerd Were you still looking at this PR?

weerd:
```javascript
const SignupButton = SignupButtonFactory(({ clickedSignUp }) => (
  <Button className="-inline nav-button" onClick={() => clickedSignUp(legacyCampaignId)} />
), 'tabbed navigation', { text: 'join us' });
```

Not for this PR, but this seems like it's a common pattern block of code in multiple places. Maybe it's something we can consolidate into a helper function or something? Not really sure but a suggestion to consider in a future update :)

itsjoekent:
Thanks @weerd ! Sorry for the monster

PR #410, opened 2017-09-22
Adding CTA on Pitch Landing Page.

weerd:
### What does this PR do?
This PR adds a CTA to the Pitch Landing Page which was missing before.

I did some playing around with the `CallToAction` component which will be replacing the existing `CallToActionBlock` component. Not ready for prime time yet, so defaulted to the older one for now.

Fixes https://www.pivotaltracker.com/story/show/151265915

### DIFF
--- a/resources/assets/containers/CallToActionBlockContainer.js
+++ b/resources/assets/containers/CallToActionBlockContainer.js
@@ -23,7 +23,5 @@ const actionCreators = {
   clickedSignUp,
 };
 
-console.log(actionCreators);
-
 // Export the container component.
 export default connect(mapStateToProps, actionCreators)(CallToActionBlock);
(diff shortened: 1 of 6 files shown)


resources/assets/containers/CallToActionBlockContainer.js
```diff
@@ -1,7 +1,7 @@
 import { connect } from 'react-redux';
 import { get } from 'lodash';
 import CallToActionBlock from '../components/CallToActionBlock';
-import { clickedSignUp } from '../actions';
+import { clickedSignUp } from '../actions/signup';
```
weerd: @DFurnes this is weird. It stopped importing as it did before. @itsjoekent ran into this too on a separate file... not sure what changed 🤔

DFurnes: So strange that this has started happening! 🤔

PR #411, opened 2017-09-25
Campaign footer changes

mendelB:
### What does this PR do?
- Affiliate links in CampaignFooter will open in new tab
- Adding ability to render multiple affiliates which might occur when there is a partner as well as a sponsor e.g. [here](https://thor.dosomething.org/us/campaigns/give-spit-about-cancer?source=campaigns) (Check the footer).

This changes the setup a bit for the `AffiliateCredit` component. Now we're exporting a sort of wrapper to deal with the different eventualities (no affiliate | sponsor or partner | sponsor and partner). Please let me know if this is too much - or if what I'm doing makes no sense.

https://www.pivotaltracker.com/story/show/151295853


resources/assets/components/AffiliateCredits/AffiliateCredits.js
```diff
@@ -0,0 +1,31 @@
...
+import AffiliateCredit from './AffiliateCredit';
+
+const AffiliateCredits = ({ affiliatePartners, affiliateSponsors }) => {
+  const sponsor = affiliateSponsors[0];
+  const partner = affiliatePartners[0];
+
+  const sponsorCredit = sponsor ?
+    <AffiliateCredit link={sponsor.fields.link} title={sponsor.fields.title} />
+    : null;
+  const partnerCredit = (partner ?
```
DFurnes: Kinda nitpicky: we should be consistent on whether we're wrapping these ternaries in parens or not.

mendelB:
> If we don't think it'll be used elsewhere, I could also see it making sense to inline the AffiliateCredit component within AffiliateCredits but that's relatively subjective.

I can go ahead and inline it. Worse comes to worse we can always extract it?

DFurnes:
Sounds good to me!

resources/assets/components/AffiliateCredits/AffiliateCredit.js
```diff
@@ -2,9 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const AffiliateCredit = ({ title, link }) => {
-  const affiliate = link ? <a href={link}>{title}</a> : title;
+  const affiliate = link ? <a href={link} target="blank">{title}</a> : title;
```
weerd: need to be `_blank` I believe.

resources/assets/components/AffiliateCredits/AffiliateCredits.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const AffiliateCredit = ({ title, link }) => {
+  const affiliate = link ? <a href={link} target="blank">{title}</a> : title;
```
weerd: I think my comment got eaten up in last update, but I was saying I believe this needs to be `_blank`.

mendelB: That's creepy foreals! I most definitely put in that underscore 😨 
For some reason, it still seems to open in a new tab without the underscore!

weerd: https://stackoverflow.com/questions/35703005/what-is-the-difference-between-target-blank-and-target-blank interesting 🤔

mendelB: iiiinteresting
https://superuser.com/a/707774

mendelB: lol just saw your comment @weerd 'Great minds...' 🍾

resources/assets/components/AffiliateCredits/AffiliateCredits.js
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    (sponsorCredit || partnerCredit) ?
+      <span>In partnership with {sponsorCredit} {multipleCredits} {partnerCredit}</span>
+      : null
+  );
+};
+
+AffiliateCredits.propTypes = {
+  affiliateSponsors: PropTypes.array.isRequired, // eslint-disable-line react/forbid-prop-types
```
weerd: try using `PropTypes.arrayOf` here and indicate what the prop types the array should contain instead of using the `eslint-disable`

weerd: Example from FB:

```javascript
// An array of a certain type
optionalArrayOf: PropTypes.arrayOf(PropTypes.number),
```

mendelB: How specific should I be with that - would say `PropTypes.arrayOf(PropTypes.object)` be sufficient?

weerd: Yeah, that works :)

PR #412, opened 2017-09-25
fixing call to action background

itsjoekent:
### What does this PR do?
screen shot 2017-09-25 at 3 31 31 pm

the background was being overridden by Forge
screen shot 2017-09-25 at 3 31 41 pm

### DIFF
--- a/resources/assets/components/CallToActionBlock/cta.scss
+++ b/resources/assets/components/CallToActionBlock/cta.scss
@@ -2,7 +2,7 @@
 
 .cta {
   color: $white;
-  background: $black;
+  background: $black !important; // Forge override
   border: 1px solid $primary-border-color;
   border-radius: $lg-border-radius;
 


resources/assets/components/CallToActionBlock/cta.scss
```diff
@@ -2,7 +2,7 @@
 
 .cta {
   color: $white;
-  background: $black;
+  background: $black !important; // Forge override
```
weerd: This should be overriding Forge since it follows after Forge is loaded. wtf is going on with our build and these weird issues lately...

itsjoekent: :man_shrugging:

weerd: So I think this will end up forcing a background color in the CTA's that have a class of `transparent` passed to them. This is not an ideal solution, but not sure what's going on with our build 😢

itsjoekent: Just tried it locally yeah it does... Where is transparent being used?

weerd: in bottom of page CTAs, on main Campaign and Landing Pages. Can you maybe look into _why_ Forge is loading forge CTA class _after_ PN CTA class?

weerd: It feels weird that first we have those import issues w/ JS functions and now dealing with weird ordering with CSS classes 🤔

PR #413, opened 2017-09-25
reversing for activity feed ordering

mendelB:
### What does this PR do?
By default, a campaign's `activityFeed` order will be reversed to avoid dragging in Contentful. 
To cancel the reversing, the `reverseActivityFeedOrder` field in `additionalContent` should be set to `false`. (This'd probably apply to old campaigns where the order has already been manually reversed.)

https://www.pivotaltracker.com/story/show/151126984

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -182,7 +182,7 @@ class Campaign extends Entity implements JsonSerializable
             // @TODO: Why is it 'activity_feed' oy? ;/
             'activityFeed' => $this->parseActivityFeed(
                 $this->activity_feed,
-                array_get($this->additionalContent, 'reverseUpdateOrder', true)
+                array_get($this->additionalContent, 'reverseActivityFeedOrder', true)
             ),
             'actionSteps' => $this->parseActionSteps($this->actionSteps),
             'quizzes' => $this->parseQuizzes($this->quizzes),
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -79,9 +79,9 @@ class Campaign extends Entity implements JsonSerializable
      * @param  array $activityItems
      * @return array
      */
-    public function parseActivityFeed($activityItems)
+    public function parseActivityFeed($activityItems, $reverseActivityFeedOrder)
     {
-        return collect($activityItems)->map(function ($item) {
+        $parsedActivityItems = collect($activityItems)->map(function ($item) {
             switch ($item->getContentType()) {
                 case 'campaignUpdate':
                     return new CampaignUpdate($item->entry);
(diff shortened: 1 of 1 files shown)


app/Entities/Campaign.php
```diff
@@ -174,7 +180,10 @@ public function jsonSerialize()
             'affiliateSponsors' => $this->parseAffiliates($this->affiliateSponsors),
             'affiliatePartners' => $this->parseAffiliates($this->affiliatePartners),
             // @TODO: Why is it 'activity_feed' oy? ;/
-            'activityFeed' => $this->parseActivityFeed($this->activity_feed),
+            'activityFeed' => $this->parseActivityFeed(
+                $this->activity_feed,
+                array_get($this->additionalContent, 'reverseActivityFeedOrder', true)
```
itsjoekent: Have we set `reverseActivityFeedOrder : false` on all of the existing Contentful campaigns?

weerd: Good note. I don't believe so, but we would have to before this is sent to production.

app/Entities/Campaign.php
```diff
@@ -79,9 +79,9 @@ public function parseActionStepPhotos($photos)
      * @param  array $activityItems
      * @return array
      */
-    public function parseActivityFeed($activityItems)
+    public function parseActivityFeed($activityItems, $reverseActivityFeedOrder)
```
weerd: can you set the parameter to default to true? `$reverseActivityFeedOrder = true`

itsjoekent: Isn't this supposed to be the new default functionality though?

itsjoekent: Check the array_get call he made

app/Entities/Campaign.php
weerd: if the `additionalContent` property doesn't exist on the Campaign entity, does this still work out fine? Defaults to `true`?

mendelB: Yeah, I think this method is similar to the `lodash` function you were showing me - I think it was `get`?

mendelB: I'll be sure to do so on Contentful -- this should be done for *all* campaigns there?

weerd: Gotcha. But that depends on the initial parameter (the array) existing right? So if `$this->additionalContent` doesn't exist then it has no array to look inside of. Just something to double check :)

weerd: Might just mean changing it to something like:

```php
array_get($this, 'additionalContent.reverseActivityFeedOrder', true)
```

But I could very well be wrong! I haven't tested it just an initial gut reaction on it :)

mendelB: ooh I'll check into that!

mendelB:
Will do. Thanks, @weerd @itsjoekent!

PR #414, opened 2017-09-26
Competition block style fix

itsjoekent:
### What does this PR do?
screen shot 2017-09-26 at 10 11 18 am

Fixed the spacing between the button and the byline

### DIFF
--- a/resources/assets/components/CompetitionBlock/competitionBlock.scss
+++ b/resources/assets/components/CompetitionBlock/competitionBlock.scss
@@ -10,11 +10,11 @@
     @include media($large) {
       float: left;
       padding-right: $base-spacing;
-      width: 66.6666666666666%;
+      width: 66.6%;
     }
   }
 
-  > .figure {
+  .byline {
     margin-top: $base-spacing;
   }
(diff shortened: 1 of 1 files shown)


resources/assets/components/CompetitionBlock/competitionBlock.scss
```diff
@@ -10,11 +10,11 @@
     @include media($large) {
       float: left;
       padding-right: $base-spacing;
-      width: 66.6666666666666%;
+      width: 66.6%;
```
weerd: I can't remember if you need at least a couple decimal places depending on browser... @DFurnes do you know offhand?

DFurnes: Yep, different browsers [truncate to different decimal points](http://cruft.io/posts/percentage-calculations-in-ie/). Was there a specific reason for removing them in this case?

itsjoekent: Oh interesting I didn't know that, I thought it was an accident!

PR #416, opened 2017-09-26
Action page sources

mendelB:
### What does this PR do?
- Allows sources to be added via markdown in an action page, and I guess anywhere else that we use the `Markdown` component.
- Replaces the `marked` package we were using for markdown rendering with the [`markdown-it`](https://github.com/markdown-it/markdown-it) package along with the [`markdown-it-footnote`](https://github.com/markdown-it/markdown-it-footnote) plugin. This is in order to support footnotes
- This - naturally - works in modals as well!

The preference seems to be to support adding sources through markdown. Unfortunately the `marked` package we were using does [not support](https://github.com/chjj/marked/issues/27) footnotes syntax such as https://rephrase.net/box/word/footnotes/. However `markdown-it` does through the `markdown-it-footnotes` plugin.

It'll take markdown like this:

image

 and output the following HTML:

```HTML
<h2>The Problem</h2>
<p>We all know texting and driving is dangerous and dumb. Yet, everyone has a friend or family member who does it anyway. In fact, 2 in 5 high school students text behind the wheel</p>
<p>Otherwise, all is quiet on the Western front<sup class="footnote-ref"><a href="#fn1" id="fnref1">[1]</a></sup>, I know this because I asked, and they answered with silence. Then in 1943, I traveled to Bermuda to plant an apple tree<sup class="footnote-ref"><a href="#fn2" id="fnref2">[2]</a></sup>, the tree is my best friend.</p>
<h2>The Solution</h2>
<p>You’ve got the power to keep your friends safer. Research shows that warnings from parents and scare tactics from ad campaigns don't work. Know what does? Reminders from friends!
Use our easy guide below to make two pairs of Thumb Socks, one to keep and one to give to a friend. These little coverings for your thumbs make it impossible to text!</p>
<hr class="footnotes-sep">
<section class="footnotes">
<ol class="footnotes-list">
<li id="fn1" class="footnote-item"><p>All Is Not Quiet On The Western Front, by Lebron James. [online] Available at: https:😕/www.lebronjames.com <a href="#fnref1" class="footnote-backref">↩︎</a></p>
</li>
<li id="fn2" class="footnote-item"><p>I haven't the faintest. https://www.google.com <a href="#fnref2" class="footnote-backref">↩︎</a></p>
</li>
</ol>
</section>
```

Which with a touch of CSS will look like this in a campaign:

image

One issue may be that clicking on those links to the footnote (not the link *in* the footnote but rather the internal link to the id of the footnote as shown in HTML), and the other way around -- seem to fire off a puck / reportback / analytics event. Not sure what the implications are.

image

https://www.pivotaltracker.com/story/show/150828675


resources/assets/components/Markdown/markdown.scss
```diff
@@ -92,3 +92,10 @@
     background-color: #f2f2f2;
   }
 }
+
+.footnotes {
+  ol, .footnote-item p {
```
weerd: use the class for `.footnotes-list` instead of `ol` here :)

resources/assets/helpers/index.js
```diff
@@ -68,13 +69,11 @@ export function ready(fn) {
...
-  const options = {
-    sanitize: true,
-  };
+  const markdownIt = new MarkdownIt();
+  markdownIt.use(markdownItFootnote);
 
   return {
-    __html: marked(source, options),
+    __html: markdownIt.render(source),
   };
```
weerd: Do we know if MarkdownIt sanitizes the output? Since it was an option in prior package and we were explicitly setting it to `true` 🤔

mendelB: They talk about security here
https://github.com/markdown-it/markdown-it/blob/master/docs/security.md
I deduced from this that we're safe here but could be wrong

mendelB: Otherwise, I guess we'd have to add a sanitizing package?

weerd: @DFurnes do you recall anything specific that we had to ensure we were sanitizing from markdown in Contentful?

DFurnes: We're still disallowing arbitrary HTML markup here, right? If so, 👍 from me!

mendelB: Yeah the package disallows HTML by default
"Don't enable HTML. Extend markup features with plugins. We think it's the best choice and use it by default."

DFurnes:
One quick note - this new Markdown package is 151kb (54kb gzipped), versus 28kb (10kb gzipped) for the package we were using before. Not a big deal, but good to keep aware of the performance impact of third party code we're bundling.

weerd:
Yeah, I foresee this number continuing to go up. We might consider using a PHP markdown package in the future and doing the processing on the server side for the markdown and just returning the HTML for the frontend to use 🤔  cc: @DFurnes @mendelB

mendelB:
I had that same thought initially when trying to figure out a good package to use - I think that's a cool idea. I totally didn't consider size though in this!

weerd:
@mendelB added potential idea to the refactor check list ^^

PR #417, opened 2017-09-27
Updates to the post gallery.

weerd:
### What does this PR do?
This PR updates the styles for the RB posts in the gallery, wrapping them in `Card` components. It also adds the section header and styles for the "User Gallery".

image

I did a bit of reorganizing of the `Gallery` component and its corresponding sub-components.


resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -5,9 +5,10 @@ import ActionStep from './ActionStep';
 import Revealer from '../Revealer';
 import { makeHash } from '../../helpers';
 import { Flex, FlexCell } from '../Flex';
+import { PostGalleryContainer } from '../Gallery/PostGallery';
 import { ReportbackUploaderContainer } from '../ReportbackUploader';
 import CompetitionContainer from '../../containers/CompetitionContainer';
-import { PostGalleryContainer, SubmissionGalleryContainer } from '../Gallery';
+import { SubmissionGalleryContainer } from '../Gallery/SubmissionGallery';
```
weerd: Ran into this issue again with having to specify the component within the directory further down, even though they are getting exported from their specific directory into the Gallery directory and the `Gallery/index.js` file properly.

resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -20,7 +21,7 @@ const ActionStepsWrapper = (props) => {
   );
 
   const postGallery = (
-    <PostGalleryContainer key="post_gallery" />
+    <PostGalleryContainer key="post_gallery" type="reportback" />
```
weerd: Added a `type` for future use 🔮

resources/assets/components/ActionPage/ActionStepsWrapper.js
itsjoekent: Agghhhhh. Okay, maybe I should set aside time to look at this, and make an issue over at the Babel repo to get some Expert ™️ help

resources/assets/components/Gallery/index.js
```diff
@@ -1,9 +1,7 @@
-export PostGallery from './PostGallery';
+export { PostGallery } from './PostGallery';
```
itsjoekent: I don't think this is correct syntax?

weerd: very possible 😬

weerd:
```
export { name1, name2, …, nameN } from …;
```
Not the same? From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export)

itsjoekent: The export is fine but I assumed `./PostGallery` was only exporting a default component, in which case you shouldn't have to use the brackets

weerd: oh very true! These imports/exports keep messing me up 😢

PR #419, opened 2017-09-27
Design qa

itsjoekent:
### What does this PR do?
1. Fixes the reportback item going extra wide if the figure body is one continuous word (eg: a youtube link)

screen shot 2017-09-27 at 10 19 05 am

2. Fixes the byline so it displays inline with the share

screen shot 2017-09-27 at 10 19 24 am

3. Fixes the Share hover effect

### DIFF
--- a/resources/assets/components/Share/share.scss
+++ b/resources/assets/components/Share/share.scss
@@ -20,7 +20,7 @@
     border: none;
     color: $light-gray;
 
-    &:hover {
+    &:hover, &:active, &:focus {
       color: $med-gray;
       cursor: pointer;
       background: none;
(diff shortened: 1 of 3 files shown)


itsjoekent:
oh shiz wait, wrong word break haha

itsjoekent:
better 
screen shot 2017-09-27 at 10 24 54 am

resources/assets/components/Share/share.scss
```diff
@@ -23,6 +23,7 @@
     &:hover {
       color: $med-gray;
       cursor: pointer;
+      background: none;
```
weerd: why was this removed? just curious

itsjoekent: hover effect without it

screen shot 2017-09-27 at 10 29 49 am

speaking of which, i should remove it for the other hover states as well

PR #420, opened 2017-09-27
custom type is a string not array

itsjoekent:
### What does this PR do?
parsing the action steps wrong, its preventing the competition from showing in the modal. also updated the test that for some reason didn't break in my prior PR...

### DIFF
--- a/resources/assets/components/Modal/containers/PostSignupModalContainer.js
+++ b/resources/assets/components/Modal/containers/PostSignupModalContainer.js
@@ -3,7 +3,7 @@ import PostSignupModal from '../configurations/PostSignupModal';
 
 const mapStateToProps = state => ({
   competitionStep: state.campaign.actionSteps.find(step => (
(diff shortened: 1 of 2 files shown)


resources/assets/components/Modal/containers/PostSignupModalContainer.js
```diff
@@ -3,7 +3,7 @@ import PostSignupModal from '../configurations/PostSignupModal';
 
 const mapStateToProps = state => ({
   competitionStep: state.campaign.actionSteps.find(step => (
-    step.customType && step.customType[0] === 'competition'
+    step.customType && step.customType === 'competition'
```
weerd: #_#

itsjoekent: I think it used to be an array in its older life and I had a gut reaction to make it an array haha

PR #423, opened 2017-09-28
Refactoring Share, tracking with Puck

itsjoekent:
### What does this PR do?
- Moves the FB UI logic from the action functions to the FB Share component
- Creates a helper to handle calling the FB UI function 
- Adds ability to share a quote in the post (Fun thing people can use, /shrug)
- Adds tracking with Puck so we're collecting all the good stuffs.

The UI logic shouldn't have been in the actions to begin with, I remember Dave or someone suggesting that in the original PR.

----

### clicking campaign update share
screen shot 2017-09-28 at 10 27 44 am

### closing a share prompt
screen shot 2017-09-28 at 10 28 14 am


itsjoekent:
ooo I should add the quote to the tracking data as well

resources/assets/components/Share/Share.js
```diff
@@ -3,19 +3,38 @@
...
 
+  const trackingData = { parentSource, variant, link, quote };
+
+  const onClick = () => {
+    requestedFacebookShare();
+    trackEvent('clicked facebook share', trackingData);
+
+    showFacebookSharePrompt({ link, quote }, (response) => {
+      if (response) {
+        facebookShareCompleted();
```
DFurnes: Do we still need these redux actions & reducer? They look like no-ops to me.

itsjoekent: Yeah atm they are, I figured it was useful to keep for down the line we need to tell Rogue a user shared something

DFurnes: I'd still kinda like to clean them up now, if it's not too much work. We'd only want to keep them in the Redux store if we wanted to update the interface based on share status, which I don't think we have any plans to do.

itsjoekent: _rips chainsaw cord_

did someone say delete some code

PR #424, opened 2017-09-28
Replace moment with date fns

mendelB:
### What does this PR do?
Replaces the impossibly hefty `moment.js` package with the slim and trim `date-fns`

A global search seemed to indicate that we were only using `moment` in the `legacy-template` in `LedeBanner` so I adjusted that accordingly. Please lemme know if I'm missing something important!

https://www.pivotaltracker.com/story/show/151012324


weerd:
> Please lemme know if I'm missing something important!

Only place so far! But I figured it's something we would need to use in more places in the future which is why I wanted a standardized, easy to use lib for this 😄

PR #426, opened 2017-09-28
Moving component containers

mendelB:
### What does this PR do?
- Moving all component containers from `components` directory into their respective component directories.
- where applicable, exporting container by default from `index.js` in component directories, and moving component logic into a separate file when found in `index.js`.
- adjusting import paths to reflect changes

Yes, this looks scary. I mean, 27 files changed goddam! But it's mostly just porting over files from one place to another. No major changes or updates save for file paths and such stuff.

https://www.pivotaltracker.com/story/show/150264425


resources/assets/components/CompetitionBlock/index.js
```diff
@@ -1,97 +1,7 @@
...
-import Markdown from '../Markdown';
-import { BlockWrapper } from '../Block';
-import Byline from '../Byline';
-import LazyImage from '../LazyImage';
-import './competitionBlock.scss';
+export CompetitionContainer from './CompetitionContainer';
 
-const DEFAULT_CONFIRMATION = `
-# You're signed up!
+export CompetitionBlockAltB from './CompetitionBlockAltB';
```
itsjoekent: We can delete these alternates, no point in maintaining an old test

resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -7,7 +7,7 @@ import { makeHash } from '../../helpers';
 import { Flex, FlexCell } from '../Flex';
 import { PostGalleryContainer } from '../Gallery/PostGallery';
 import { ReportbackUploaderContainer } from '../ReportbackUploader';
-import CompetitionContainer from '../../containers/CompetitionContainer';
+import { CompetitionContainer } from '../CompetitionBlock';
```
weerd: Can we rename the `CompetitionContainer` to reflect the name of the component, so `CompetitionBlockContainer`. Once we formalize this as _not_ a custom block and a formalized component, we'll have both for a period of time until we can remove the older "block" version.

resources/assets/components/CompetitionBlock/CompetitionBlock.js
```diff
@@ -49,30 +56,29 @@ const CompetitionBlockAltB = (props) => {
   ) : null;
 
   return (
-    <Block className={classnames('-colorful')}>
-      <BlockTitle>Go above and beyond!</BlockTitle>
+    <BlockWrapper title="Go above and beyond!" className="-default">
```
weerd: I believe we want the colorful version here, unless that was decided against at a later date 🤔

weerd: Oh you know what, I think the GitHub diff showing changes is throwing me off. Since there were two alternates for the `CompetitionBlock` during experimentation. The colorful one was the one that I think ultimately did better so I think we want that one AltB to be the current version.

mendelB:
@weerd replaced `CompetitionBlock` with the `CompetitionBlockAltB` as well as the containers. Please let me know if I got this right!

PR #428, opened 2017-09-29
Open graph refactoring + adding custom share props to campaign updates

itsjoekent:
### What does this PR do?
- Refactors the open graph system so the naming is more consistent, properly parsed by the Campaign class, etc
- Adds logic for the Campaign Update to accept the Social Override

##### Campaign with no custom share override
screen shot 2017-09-29 at 2 29 28 pm

##### Campaign with custom share override
screen shot 2017-09-29 at 2 26 07 pm

##### Campaign Update with no custom share override
screen shot 2017-09-29 at 2 36 10 pm

##### Campaign Update with custom share override
screen shot 2017-09-29 at 2 28 18 pm

i write great content


app/Http/Controllers/CampaignController.php
```diff
@@ -56,16 +57,19 @@ public function index()
     public function show($slug)
     {
         $campaign = $this->campaignRepository->findBySlug($slug);
-        $shareFields = getShareFields($campaign, $campaign->socialOverrides);
         $env = get_client_environment_vars();
 
+        // The slug argument will not contain `/blocks` or other path extensions, hence `::url()`.
+        // $socialFields = get_social_fields($campaign->jsonSerialize(), Request::url());
```
itsjoekent: whoops i should get rid of that

itsjoekent:
Used this for the previews in the screenshots, super useful! https://chrome.google.com/webstore/detail/open-graph-preview/ehaigphokkgebnmdiicabhjhddkaekgh?hl=en

app/helpers.php
```diff
@@ -224,28 +224,49 @@ function useOverrideIfSet($field, $campaign, $override)
...
+            return $value->id === $blockId;
+        });
+
+        if ($block && isset($block->fields->socialOverride)) {
+            $socialOverride = $block->fields->socialOverride->fields;
+        }
+    }
+
+    $coverImage = useOverrideIfSet('coverImage', $campaignFlattened, $socialOverride);
+    if (gettype($coverImage) === 'object') {
```
DFurnes: What is `$coverImage` in this case? Is there more specific type-checking we can do?

itsjoekent: Here's an example,

screen shot 2017-09-29 at 3 15 21 pm

And if you run the following you simply get `'object'`

```
dd(gettype($coverImage));
```

----

The reason this logic happens is because of the line above, depending on where the `coverImage` is pulled from it either outputs a string or an object.

itsjoekent: It's not _perfect_ but as we chatted over yesterday, I don't think this is the long-term solution anyway

itsjoekent: I think it makes sense to add a comment here quickly explaining this actually

DFurnes: What happens if you `dd(get_class($coverImage))`? Not arguing this doesn't work in this scenario, but I'm wary that we might later have a different case where an object is passed here that doesn't have a `url` property.

itsjoekent: @DFurnes I just get `stdobject`. Maybe we do an additional check for if the object also has the URL property?

DFurnes: Blah, we must just be casting it as an `(object)` somewhere. Oh well.

DFurnes: Let's merge this, can always revisit later on if it becomes an issue!

itsjoekent: Alrighty!

PR #429, opened 2017-09-29
Fix dashboard

itsjoekent:
### What does this PR do?
This PR guarantees we will no longer get any possible errors from the dashboard totalSignups being null or the prop value given being null.

(Also note: the default value in the store is 0)


weerd:
@itsjoekent any clue as to what was changing it to `null`?

itsjoekent:
No idea, probably a bad API response?

PR #430, opened 2017-09-29
rendering `a` tag when shareLink prop is absolute

mendelB:
### What does this PR do?
Temp fix for an issue with campaigns using the old campaign update block in activity feed. The link is a full absolute link that gets rendered in a React `Link`, that appends the given URL to the current URL, causing agony and grief.

image

This is a bit explicit and inelegant but perhaps a sufficient temporary fix for an issue we'd like to do away with. To quote the great Diego:

> "I’m debating if we should just transfer the old updates to the new style and call it a day. then we can ideally fully deprecate the old campaign update and remove it as an option for campaign leads to use :thinking_face:"

### DIFF
--- a/resources/assets/components/Block/BlockWrapper.js
+++ b/resources/assets/components/Block/BlockWrapper.js
@@ -5,16 +5,21 @@ import classnames from 'classnames';
 import { Link } from 'react-router-dom';
 import './block-wrapper.scss';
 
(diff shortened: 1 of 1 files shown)


weerd:
> To quote the great Diego

Oh man... too much pressure 😱

PR #432, opened 2017-09-29
Arrow signup blurb - Legacy

mendelB:
### What does this PR do?
- Adds a new CampaignSignupArrow component that can render the awesome arrow next to a signup button (a bit specific to the LedeBanner signup button in legacy for now) based on an additionalContent field. 

Scholarships + Awesome Arrow == Signups!

Before Awesome Arrow:
image

:(

After Awesome Arrow:
image

:)

https://www.pivotaltracker.com/story/show/151132567

### DIFF
--- a/resources/assets/components/Page/CampaignPage/CampaignPageContainer.js
+++ b/resources/assets/components/Page/CampaignPage/CampaignPageContainer.js
@@ -17,6 +17,7 @@ const mapStateToProps = state => ({
   affiliatePartners: state.campaign.affiliatePartners,
   campaignLead: get(state.campaign.additionalContent, 'campaignLead'),
   legacyCampaignId: state.campaign.legacyCampaignId,
+  signupArrowContent: get(state.campaign.additionalContent, 'signupArrowContent', null),
   slug: state.campaign.slug,
   subtitle: state.campaign.callToAction,
   template: state.campaign.template,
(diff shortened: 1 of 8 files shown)


itsjoekent:
looks good on the heroku app!

resources/assets/components/Page/CampaignPage/CampaignPageContainer.js
```diff
@@ -17,6 +17,7 @@ const mapStateToProps = state => ({
   affiliatePartners: state.campaign.affiliatePartners,
   campaignLead: get(state.campaign.additionalContent, 'campaignLead'),
   legacyCampaignId: state.campaign.legacyCampaignId,
+  signupArrowContent: get(state.campaign.additionalContent, 'signupArrowContent', null),
```
weerd: Isn't it only really needed on the LandingPage?

mendelB: Ohh right. The `CampaignPage` won't get rendered if they're not affiliated right?

ashleybaldwin: def just the landingpage for the arrow

mendelB: Aight will remove

PR #436, opened 2017-10-03
Sticky tab and scroll concierge™ fix

mendelB:
### What does this PR do?
- Adds parent div offset in `ScrollConcierge` to account for the `div` in `CampaignPage` that encloses the `TabbedNavigation`
- Adds `clearfix` class to that new `div` to force wrap all content to make sticky tab work

The Tabbed Navigation and Scroll Concierge were out of order

and-justice-for-all-movie-misquote-captioned

due to a new parent element in `CampaignPage` that was messing with the Scroll Concierge[™]'s `offset` calculation (being that the node now has a new `offsetParent`) and since this new element wasn't wrapping all content properly, it was causing the `sticky` rule to fail on tabbed navigation.
But that all works now!!!!
(HT Dave and Diego!)

https://www.pivotaltracker.com/story/show/151553928

### Checklist

### DIFF
--- a/resources/assets/components/ScrollConcierge/index.js
+++ b/resources/assets/components/ScrollConcierge/index.js
@@ -64,7 +64,8 @@ class ScrollConcierge extends React.Component {
     // Wait for headline font to load so we don't scroll to
     // the wrong place when the page reflows & offset changes.
     font.load().then(() => {
-      // Get node offset from top of viewport combined with scroll distance to obtain desired scroll target
+      // Get node offset from top of viewport combined with
+      // scroll distance to obtain desired scroll target
       const viewportOffset = this.node.getBoundingClientRect().top;
       const pageOffset = window.scrollY + viewportOffset;
 
--- a/resources/assets/components/ScrollConcierge/index.js
+++ b/resources/assets/components/ScrollConcierge/index.js
@@ -64,13 +64,14 @@ class ScrollConcierge extends React.Component {
     // Wait for headline font to load so we don't scroll to
     // the wrong place when the page reflows & offset changes.
     font.load().then(() => {
-      const offsetParent = this.node.offsetParent;
-      const offsetTop = offsetParent.offsetTop + this.node.offsetTop;
+      // Get node offset from top of viewport combined with scroll distance to obtain desired scroll target
+      const viewportOffset = this.node.getBoundingClientRect().top;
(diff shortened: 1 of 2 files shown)


resources/assets/components/ScrollConcierge/index.js
```diff
@@ -64,10 +64,13 @@ class ScrollConcierge extends React.Component {
     // Wait for headline font to load so we don't scroll to
     // the wrong place when the page reflows & offset changes.
     font.load().then(() => {
+      const offsetParent = this.node.offsetParent;
```
DFurnes: So is this needed because we added a new wrapping `position: relative` container to the page markup? If so, would adding another break this again?

mendelB: I believe so 😬 
This doesn't feel like the most elegant solution...

DFurnes: Hey, it works! 😄 If you want to experiment a bit more before merging, it looks like [`getBoundingClientRect`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) (which we use [in the sticky nav](https://github.com/DoSomething/phoenix-next/blob/8dde8c84961956ad60a5d23e7dc6a39af91c1e84/resources/assets/components/Navigation/TabbedNavigation.js#L73)) might be a bit closer to what we're looking for (paired with `window.scrollY` to get the top offset, not just viewport offset).

mendelB: Will do!

mendelB: @DFurnes This totally works! Imma push the change

resources/assets/components/ScrollConcierge/index.js
```diff
@@ -64,10 +64,15 @@ class ScrollConcierge extends React.Component {
     // Wait for headline font to load so we don't scroll to
     // the wrong place when the page reflows & offset changes.
     font.load().then(() => {
+      // Get node offset from top of viewport combined with
+      // scroll distance to obtain desired scroll target
+      const viewportOffset = this.node.getBoundingClientRect().top;
```
weerd: oh yeah! I forgot I used `getBoundingClientRect()` in the sticky nav. It's pretty sweet :D

PR #437, opened 2017-10-03
Update default database to match setup guide.

DFurnes:
### What does this PR do?
In our [Homestead setup guide](https://github.com/DoSomething/communal-docs/tree/master/Homestead), we create a `phoenix` database but in our default `.env` we just connect to `homestead`. We should use app-specific databases by default so we don't run into conflicts when working on more than one application. 🔨

This came up while some of the Team Bleed folks were setting up environments.

N/A

### Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,7 +7,7 @@ APP_URL=http://localhost
 DB_CONNECTION=mysql
 DB_HOST=127.0.0.1
 DB_PORT=3306
-DB_DATABASE=homestead
+DB_DATABASE=phoenix
 DB_USERNAME=homestead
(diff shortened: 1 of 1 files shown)


.env.example
```diff
@@ -7,7 +7,7 @@ APP_URL=http://localhost
 DB_CONNECTION=mysql
 DB_HOST=127.0.0.1
 DB_PORT=3306
-DB_DATABASE=homestead
+DB_DATABASE=phoenix
```
weerd: haha, good call! I think we figured people would change it and possibly name the DB what they wanted, but this makes more sense 👍

PR #438, opened 2017-10-03
removing margin-bottom on `article.figure`

mendelB:
### What does this PR 
sets `margin-bottom` on `article.figure` elements to `0`

On on *Mobile* or a *small screen*, there's a mess of whitespace below any `article.figure` that seems out of place. 
The ticket only called for fixing the `byline` whitespace but we're using `Figure` which renders the `article.figure` in several places, and as far as I can tell, the whitespace issue exists in all places.

CampaignUpdate Byline:
image

CampaignUpdate ArticleEmbed:
image

This also happens in `Affirmation` and `Reportback` but you'll have to take my word for it :)

After:
image

https://www.pivotaltracker.com/story/show/151524638

### Checklist

### DIFF
--- a/resources/assets/components/Page/CampaignPage/CampaignPage.js
+++ b/resources/assets/components/Page/CampaignPage/CampaignPage.js
@@ -42,7 +42,7 @@ const CampaignPage = (props) => {
         affiliateSponsors={affiliateSponsors}
       />
 
-      <div className="main">
+      <div className="main clearfix">
         { dashboard ?
           <Dashboard
             totalCampaignSignups={totalCampaignSignups}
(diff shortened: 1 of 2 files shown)


resources/assets/components/Figure/figure.scss
```diff
@@ -12,7 +12,7 @@
   img { width: 100% }
 }
 
-.figure.-left.-center.-small {
+.figure.-left.-center.-small, article.figure {
```
weerd: Could we maybe use utility classes here? Worried this might have unintended consequences since it's not as specific... try creating a class that we can add to the components affected? Thoughts?

mendelB: Yeah that makes sense! I'm just trying to circumvent a rule from - I assume - legacy that adds that margin on small screens. I'll try adding a class to address this instead

PR #439, opened 2017-10-04
removing margin-bottom on various `.figure` elements

mendelB:
### What does this PR 
sets `margin-bottom` on various `.figure` elements to `0`

On on *Mobile* or a *small screen*, there's a mess of whitespace below some of the `.figure` elements that seem out of place. 
The ticket only called for fixing the `byline` whitespace but we're using `Figure` which renders the `article.figure` in several places, and as far as I can tell, the whitespace issue exists in all places.

- This CSS rule comes from Forge and seems to be intended for the https://www.dosomething.org/us/about/our-people page for example to give breathing space to some vertically aligned elements on small screens. Diego recommended adjusting this in the specific areas within their respective components as opposed to a general removal of the margin on the `Figure` component itself. See #438 

CampaignUpdate Byline:
image

CampaignUpdate ArticleEmbed:
image

This also happens in `Affirmation` and `Reportback` but you'll have to take my word for it :)

After:
image

https://www.pivotaltracker.com/story/show/151524638

### Checklist

### DIFF
--- a/resources/assets/components/Affirmation/affirmation.scss
+++ b/resources/assets/components/Affirmation/affirmation.scss
@@ -61,8 +61,4 @@
       color: $primary-color;
     }
   }
-
-  .figure {
-    margin-bottom: 0;
-  }
 }
--- a/resources/assets/components/Affirmation/affirmation.scss
+++ b/resources/assets/components/Affirmation/affirmation.scss
@@ -61,4 +61,8 @@
       color: $primary-color;
     }
   }
+
+  .figure {
+    margin-bottom: 0;
+  }
 }
--- a/resources/assets/components/ReportbackItem/reportback-item.scss
+++ b/resources/assets/components/ReportbackItem/reportback-item.scss
@@ -23,6 +23,10 @@
   > .figure__body p {
     word-break: break-word;
   }
+
+  &.figure, .figure {
+    margin-bottom: 0;
+  }
 }
 
 .padded {
--- a/resources/assets/components/ReportbackItem/reportback-item.scss
+++ b/resources/assets/components/ReportbackItem/reportback-item.scss
@@ -23,10 +23,6 @@
   > .figure__body p {
     word-break: break-word;
   }
-
-  &.figure, .figure {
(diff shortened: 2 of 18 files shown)


resources/assets/components/Affirmation/affirmation.scss
```diff
@@ -61,4 +61,8 @@
       color: $primary-color;
     }
   }
+
+  .figure {
```
DFurnes: Can we handle this with a utility class instead of CSS overrides?

resources/assets/components/ReportbackItem/reportback-item.scss
```diff
@@ -23,6 +23,10 @@
   > .figure__body p {
     word-break: break-word;
   }
+
+  &.figure, .figure {
```
DFurnes: This is a pretty scary looking selector. 👻 If we don't decide to use a utility class here, let's at least figure out what's causing us to need this.

mendelB: In this case it was to combat a nested component of the same kind, causing two places CSS tries enforcing that `margin-bottom` rule. this had me spinning in circles

resources/assets/components/Affirmation/affirmation.scss
mendelB: I think so, Diego agrees.

mendelB:
This is how it might look with a `margin-bottom-none` utility class. The issue is that the rule is only applied using `!important` otherwise, it's overridden by the old rule.

DFurnes:
I'm very okay with `!important` on utility classes.

mendelB:
Question though (source Diego) is if there should be some naming convention for rules with `!important` in them for clarity sake? perhaps in this case `overide__margin-bottom-none` or some such clever bout of naming genius?

DFurnes:
I think that's an inherent feature of utility classes - they always act as overrides. Longer-term, we can probably fix this by reducing our reliance on Sass nesting since that causes 90% of our specificity woes.

PR #442, opened 2017-10-04
Post gallery show hide view more

weerd:
### What does this PR do?
This PR updates the `reportbacks` reducer to more clearly account for the current "page" of reportbacks and also pass along the total pages of reportbacks returned from the API.

This allows doing a check to then decide whether to show or hide the view more button for loading more RB post items depending on whether or not we've seen all the pages of items.

### DIFF
--- a/resources/assets/actions/reportback.js
+++ b/resources/assets/actions/reportback.js
@@ -30,8 +30,25 @@ export function requestedReportbacks(node) {
 }
 
 // Action: new reportback data received.
-export function receivedReportbacks({ reportbacks, reportbackItems, reactions }, page, total) {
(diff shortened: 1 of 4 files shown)


resources/assets/actions/reportback.js
```diff
@@ -30,8 +30,25 @@ export function requestedReportbacks(node) {
...
+) {
+  return {
+    type: RECEIVED_REPORTBACKS,
+    reportbacks,
+    reportbackItems,
+    reactions,
+    currentPage,
+    totalPages,
+    total,
+  };
```
itsjoekent: im pretty sure javascript is simply a philosophical idea at this point

PR #443, opened 2017-10-04
Avoid empty items in row of posts 3 or 2 items across.

weerd:
### What does this PR do?
This PR provides a simple (albeit hard coded for now) solution to ensure that we get items by page in a count divisible by both 2, 3, and 4 which is useful for how we display items in a variety of galleries.

Avoid this:
image

Aiming for this:
image

### DIFF
--- a/resources/assets/actions/reportback.js
+++ b/resources/assets/actions/reportback.js
@@ -221,7 +221,7 @@ export function fetchUserReportbacks(userId, campaignId) {
 // Async Action: fetch another page of reportbacks.
 export function fetchReportbacks() {
   return (dispatch, getState) => {
-    const count = 24;
+    const count = 24; // Count of items divisible by 2, 3, and 4,for full gallery rows!
     const node = getState().campaign.legacyCampaignId;
     const page = getState().reportbacks.currentPage + 1;
 
--- a/resources/assets/actions/reportback.js
+++ b/resources/assets/actions/reportback.js
@@ -221,12 +221,13 @@ export function fetchUserReportbacks(userId, campaignId) {
 // Async Action: fetch another page of reportbacks.
 export function fetchReportbacks() {
   return (dispatch, getState) => {
+    const count = 24;
     const node = getState().campaign.legacyCampaignId;
     const page = getState().reportbacks.currentPage + 1;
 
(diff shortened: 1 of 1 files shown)


resources/assets/actions/reportback.js
```diff
@@ -221,12 +221,13 @@ export function fetchUserReportbacks(userId, campaignId) {
 // Async Action: fetch another page of reportbacks.
 export function fetchReportbacks() {
```
weerd: I tried doing something like:

```js
export function fetchReportbacks(count = 24) {
...
}
```
to default to 24 and potentially allow passing a custom value... but apparently the way `fetchReportbacks` in React components, it seems to pass a value of undefined for count if not specified... 🤷‍♂️

weerd:
> Can you add a comment restating the 2, 3, 4 thing?

In the code right? sure!

itsjoekent:
sweets!

resources/assets/actions/reportback.js
```diff
@@ -221,12 +221,13 @@ export function fetchUserReportbacks(userId, campaignId) {
...
   return (dispatch, getState) => {
+    const count = 24; // Count of items divisible by 2, 3, and 4,for full gallery rows!
```
DFurnes: 💯

weerd:
Actually, was looking at Missing In History campaign, and this doesn't actually help that well. It only does if everyone submitted only one item per RB. If they submitted more and then we consolidate to the most recent, you still end up at times without a fully filled row since we're asking for 24 RB items instead of 24 RBs. So I'm going to need to execute on the other idea of using a bit more logic to check on the count of RBs retrieved instead. 🤷‍♂️

DFurnes:
Mmmmm, tricky. I wonder if we can make that easier to query from Rogue as well. 🤔

weerd:
Yeah. Ideally, and I don't know if this exists already in Rogue, but the aim would be to be able to request posts of only the latest post for a user. cc @sbsmith86

I'll see if I can figure out a quick fix if possible, otherwise I'll leave as is for now and wait for Team Bleed to update with requests to Rogue directly.

sbsmith86:
@weerd I don't believe we currently do that, but we certainly could!

PR #444, opened 2017-10-05
Add affiliate option under signup button

sbsmith86:
### What does this PR do?
This PR adds an `AffiliateOption` component that can be used to opt-ins to third party messaging when a user is signing up for a campaign. 

All this does is add the component to the page. 

I haven't done any work on the contentful side to allow for custom configuration, so everything that i believe should be configured I have added to the `store`. And I should note, I set the option to show or not show this component as false by default so it shouldn't pop up anywhere while we continue to work on it. @katiecrane should be picking up from where I left off to hook that piece up. 

This also doesn't send the opt-in information along with the signup request. Hoping @katiecrane or @chloealee will be able to hop on to that part to get more familiar with the codebase as well. 

This is set up to mimic the third party optin option we set up on phoenix next: 

image

Addresses https://www.pivotaltracker.com/story/show/151270798

### Checklist

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -58,11 +58,6 @@ export function signupPending() {
   return { type: SIGNUP_PENDING };
 }
 
-// Action: reportback fetch initiated.
-export function signupOptOut() {
-  return { type: SIGNUP_CLICKED_OPT_OUT };
-}
-
 // Async Action: check if user already signed up for the campaign
 export function checkForSignup(campaignId) {
   return (dispatch, getState) => {
@@ -152,7 +147,5 @@ export function clickedSignUp(campaignId, shouldRedirectToActionTab = true) {
 
 // Action: sends whether the user opted out of affiliate messaging.
 export function clickedOptOut() {
-  return (dispatch) => {
-    dispatch(signupOptOut());
-  };
+  return { type: SIGNUP_CLICKED_OPT_OUT };
 }
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -10,6 +10,7 @@ import {
   SIGNUP_FOUND,
   SIGNUP_NOT_FOUND,
   SIGNUP_PENDING,
+  SIGNUP_CLICKED_OPT_OUT,
   SET_TOTAL_SIGNUPS,
   queueEvent,
   addNotification,
@@ -55,6 +56,11 @@ export function signupNotFound() {
 // Action: waiting on a signup response.
 export function signupPending() {
   return { type: SIGNUP_PENDING };
+}
+
+// Action: reportback fetch initiated.
+export function signupOptOut(node) {
+  return { type: SIGNUP_CLICKED_OPT_OUT, node };
 }
 
 // Async Action: check if user already signed up for the campaign
@@ -141,5 +147,12 @@ export function clickedSignUp(campaignId, shouldRedirectToActionTab = true) {
         }
       }
     });
+  };
+}
+
+// Action: sends whether the user opted out of affiliate messaging.
+export function clickedOptOut() {
+  return (dispatch) => {
+    dispatch(signupOptOut());
   };
 }
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -59,8 +59,8 @@ export function signupPending() {
 }
 
 // Action: reportback fetch initiated.
-export function signupOptOut(node) {
-  return { type: SIGNUP_CLICKED_OPT_OUT, node };
+export function signupOptOut() {
+  return { type: SIGNUP_CLICKED_OPT_OUT };
 }
 
 // Async Action: check if user already signed up for the campaign
--- a/resources/assets/components/AffiliateOption/AffiliateOption.js
+++ b/resources/assets/components/AffiliateOption/AffiliateOption.js
@@ -25,7 +25,7 @@ const AffiliateOption = (props) => {
   return (
     <div className="form-wrapper affiliate-option">
       <label className="option -checkbox" htmlFor="affiliate_opt_in">
-        <input type="checkbox" id="opt_in" name="affiliate_opt_in" value={optedOut} className="form-checkbox" onClick={clickedOptOut} />
+        <input type="checkbox" id="opt_in" name="affiliate_opt_in" value={optedOut} checked={!optedOut} className="form-checkbox" onClick={clickedOptOut} />
         <span className="option__indicator" />
         {affiliateOptionLabel}
       </label>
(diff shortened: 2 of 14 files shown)


resources/assets/components/AffiliateOption/AffiliateOption.js
```diff
@@ -0,0 +1,48 @@
...
+    clickedOptOut,
+    affiliateOptionLabel,
+    moreInformationLabel,
+    moreInformationMessage,
+  } = props;
+
+  const handleClick = (event) => {
+    event.preventDefault();
+
+    const details = document.querySelector('.footnote-details');
```
DFurnes: Directly modifying React-rendered DOM elements is a little dangerous, since React may re-render them at any moment and this change would be lost. Instead, let's make this into a "stateful" class component:

```js
class AffiliateOption {
    constructor(props) {
        super(props);

        this.state = { expanded: false };
        this.handleClick = this.handleClick.bind(this);
    }

    handleClick(event) {
        this.setState(state => { expanded: ! state.expanded });
    }

    render() {
        // ...
        <div className={classNames('footnote-details', {'is-expanded': this.state.expanded})}>{moreInformationMessage}</div>
    }
}

```

resources/assets/components/AffiliateOption/AffiliateOptionContainer.js
```diff
@@ -0,0 +1,19 @@
...
+  moreInformationLabel: state.signups.affilitateConfiguration.moreInformationLabel,
+  moreInformationMessage: state.signups.affilitateConfiguration.moreInformationMessage,
+});
+
+const actionCreators = {
+  clickedOptOut,
+};
+
+// Probably need a opt-out boolean, whether the user has already opted in/out
+// Probably need a show boolean to let us know if we should or should not show.
```
DFurnes: Can remove these comments!

resources/assets/store.js
```diff
@@ -40,6 +40,14 @@ const initialState = {
     thisSession: false,
     isPending: false,
     total: 0,
+    optedOut: false,
+    // @TODO: this configuration should come from contentful. Should this be in the campaign object?
```
DFurnes: Yes! We tend to keep things like this in the `additionalContent` JSON blob on the campaign (until we decide it's something that we're confident in and using a lot), so we could make a `displayAffiliateOptOut` "feature flag" that a campaign lead or PM could add there.

resources/assets/components/AffiliateOption/AffiliateOption.js
sbsmith86: omg haha yes.

resources/assets/actions/signup.js
```diff
@@ -143,3 +149,10 @@ export function clickedSignUp(campaignId, shouldRedirectToActionTab = true) {
     });
   };
 }
+
+// Action: sends whether the user opted out of affiliate messaging.
+export function clickedOptOut() {
```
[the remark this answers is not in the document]
sbsmith86: Hmm I was getting a weird error when I tried not using an async action here, where it seemed to be making some cross domain request to an iframe. Very strange. I asked @weerd about it and I tried this approach and it cleared up that error. 

However, i'm playing around with it now and it seems to work the way you suggested. I wish I could find the original error I was seeing but when I was googling it, it did seem related to redux-dev tools???? 🤷‍♀️ Anyway, I will just go with your suggestion lol.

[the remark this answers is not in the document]
DFurnes: Woah, spooky! 👻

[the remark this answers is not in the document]
weerd: I can attest to seeing said error. It was real weird.

resources/assets/store.js
sbsmith86: @DFurnes ok cool, so just change that to `displayAffilitateOptOut` or do I have to do something special to specify it should be in the `additionalContent` JSON?

weerd: I'll try and find an example where we're using the `additionalContent` field in JSON to set a flag. But that would be a value obtained in the container (likely there) which then sets the value of this prop in the store.

sbsmith86:
@DFurnes @weerd Made updates based on your comments! Also it looks like I had forgot to commit a stylesheet, so you might want to take a look at that.

resources/assets/store.js
weerd: I would maybe suggest calling this something more obvious though, like `receiveAffiliateMessaging` so it's more obvious what it's referring to. `optedOut` could end up being confusing regarding who or what was optedOut.

weerd:
@sbsmith86 you were too quick and by the time I added another comment you pushed an update and my comment collapsed. Not sure if you saw [this one](https://github.com/DoSomething/phoenix-next/pull/444#discussion_r143013436).

weerd:
@sbsmith86 here's an [example](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/Page/LandingPage/LandingPageContainer.js#L18) where we set the pitch page content in the additional content JSON field in Contentful. Then we grab it within a container file so the value gets sent as a prop within the component it's rendering. You'd probably be grabbing it from within an action to send value to reducer.

resources/assets/store.js
```diff
@@ -17,6 +17,13 @@ const initialState = {
   },
   campaign: {
     activityFeed: [],
+    // @TODO: this configuration should come from contentful?
+    displayAffilitateOptOut: {
```
weerd: I don't think @DFurnes meant to rename this prop to `displayAffilitateOptOut`since that sounds like it would have just a boolean value and not include the label, etc. Maybe rename this to just `AffiliateOptOut` or `AffiliateMessaging` or something.

sbsmith86: ok!

sbsmith86: Sorry I think I am still a little confused by this. Is this what the campaign object should look like in the store?

```
{
  campaign: {
    activityFeed: [],
    AffiliateMessaging: get(state.campaign.landingPage.fields.additionalContent, 'affiliateMessaging'),
    displayAffilitateOptOut: get(state.campaign.landingPage.fields.additionalContent, 'displayAffilitateOptOut'),
  },

}
```

sbsmith86:
I updated this PR to use configurations from the `additionalContent` field in this format. 

image

resources/assets/components/AffiliateOption/AffiliateOption.js
```diff
@@ -0,0 +1,45 @@
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import React from 'react';
+import './affiliateOption.scss';
+
+
```
weerd: 👾

sbsmith86: 😬

PR #446, opened 2017-10-10
Send signups to rogue

sbsmith86:
### What does this PR do?

Sends signups directly to Rogue instead of using the Phoenix-ashes proxy. Will also send over affiliate opt-in/out value to Rogue, if we were collecting it. 

The Phoenix-ashes signups endpoint did support sending affiliate option information, so we decided to hook this up directly to Rogue. 

This PR only adds support for sending a signup over to Rogue. PN will still use the PA proxy endpoints for any other work it was doing with signups and RBs. There will be future work to fully connect with Rogue for all things. 

NOTES: Rogue doesn't currently support contentful campaign IDs, that is coming in the next sprint, so we need to send over legacy campaign and campaign run IDs for now. 

Fixes https://www.pivotaltracker.com/story/show/151767351

and 

Fixes https://www.pivotaltracker.com/story/show/151270798

### Checklist


weerd:
> Rogue doesn't currently support contentful campaign IDs, that is coming in the next sprint, so we need to send over legacy campaign and campaign run IDs for now.

👍

PR #447, opened 2017-10-11
Fixing quiz actions import

itsjoekent:
Stupid babel bug strikes again.


itsjoekent:
thanks

PR #448, opened 2017-10-11
Modal improvements

itsjoekent:
### What does this PR do?
TLDR: Allows you to link to campaign updates in a modal. Try it `/us/campaigns/do-something-las-vegas/modal/69vW8NITMk4U8MwYKcAQwy`

#### Campaign Update
- Updates Container to import all of the necessary fields based on the given ID
- Adds a `bordered` property to determine if the bordered class is applied

#### Modal
- Renames PageModal to ContentModal
- Updates ContentModal container to fetch from a larger pool of objects (Pages, activity feed)
- Allows the ContentModal to either use a generic Card or a Campaign Update. (It's a switch, could add more content types here)

#### CampaignPage
- Just a quick to-do to solve a problem we currently have with the modal

**STILL TO DO IN NEXT PR**: Reportback uploader 

https://www.pivotaltracker.com/story/show/151765277

### DIFF
--- a/resources/assets/components/CampaignUpdate/CampaignUpdate.js
+++ b/resources/assets/components/CampaignUpdate/CampaignUpdate.js
@@ -8,13 +8,13 @@ import Byline from '../Byline';
 import Markdown from '../Markdown';
 import { ShareContainer } from '../Share';
 
-const CampaignUpdate = ({ id, author, content, link, shareLink }) => {
+const CampaignUpdate = ({ id, author, content, link, shareLink, bordered }) => {
   const { avatar, jobTitle, name } = author.fields;
 
   const isTweet = content.length < 144;
 
   return (
-    <Card id={id} className="rounded bordered" link={shareLink} title="Campaign Update">
+    <Card id={id} className={classnames('rounded', { bordered })} link={shareLink} title="Campaign Update">
       <Markdown className={classnames('padded', { 'font-size-lg': isTweet })}>
         {content}
       </Markdown>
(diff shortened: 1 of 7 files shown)


resources/assets/components/CampaignUpdate/CampaignUpdate.js
```diff
@@ -49,10 +49,12 @@ CampaignUpdate.propTypes = {
   content: PropTypes.string.isRequired,
   link: PropTypes.string,
   shareLink: PropTypes.string.isRequired,
+  bordered: PropTypes.bool,
 };
 
 CampaignUpdate.defaultProps = {
   link: null,
+  bordered: true,
```
weerd: I'd like to eventually consider a better way to handle styles when context dependent, but this is ok for now 🤔

resources/assets/components/Modal/ModalSwitch.js
```diff
@@ -11,7 +11,7 @@ const ModalSwitch = (props) => {
 
   switch (modalType) {
     case POST_SIGNUP_MODAL: children = <PostSignupModal />; break;
-    case PAGE_MODAL: children = <PageModal />; break;
+    case CONTENT_MODAL: children = <ContentModal />; break;
```
weerd: can you write these as multiple lines so it's easier to read?

resources/assets/components/Modal/configurations/ContentModal.js
```diff
@@ -0,0 +1,39 @@
...
+
+  const card = (
+    <Card title={title} className="modal__slide">
+      <Markdown className="padded">{ content }</Markdown>
+    </Card>
+  );
+
+  switch (type) {
+    case 'campaignUpdate': return campaignUpdate;
+    default: return card;
```
weerd: can you write these switch cases as multiple lines so it's easier to read?

resources/assets/components/CampaignUpdate/CampaignUpdate.js
itsjoekent: ya was kind of a quick hack

PR #449, opened 2017-10-11
Pass opt out option to queued event

sbsmith86:
### What does this PR do?

If a user is unauthenticated, they click to signup to a campaign and we redirect them to northstar. We then direct them back to PN where we re-initialize the store and lose any state we updated before the redirect

It looks like the signup action gets queued when that happens so I added support for an `options` parameter that we can pass in any options that got set before the redirect. That way if a user opts-out, then signs up, then signs in on northstar to authenticate, we caption the opt out value in the signup request. 

Fixes a bug found where we were not capturing opt-in/opt-out options of unauthenticated users during campaign signup.

Fixes https://www.pivotaltracker.com/story/show/151767351

### Checklist

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -106,7 +106,7 @@ export function getTotalSignups(campaignId) {
 
 // Async Action: send signup to phoenix and
 // check if the user is logged in or has an existing signup.
-export function clickedSignUp(campaignId, shouldRedirectToActionTab = true) {
+export function clickedSignUp(campaignId, options = null, shouldRedirectToActionTab = true) {
   return (dispatch, getState) => {
     const campaignActionUrl = join('/us/campaigns', getState().campaign.slug, '/action');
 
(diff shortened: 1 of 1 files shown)


sbsmith86:
@itsjoekent Sorry, I don't think I understand what that would replace? I do need `details` to be a custom string not just `true` or `false` but I'm not sure if you meant that should replace something specific.

itsjoekent:
The ternary will return the value if its truthy, quick mock of it

```
state = { campaign: { additionalContent: { displayAffilitateOptOut: 'test string' } } }

function getState() {
  return state;
}

function clickedSignup(options) {
	return options || getState().campaign.additionalContent.displayAffilitateOptOut;
}

console.log(clickedSignup('this is an options string'));
console.log(clickedSignup())
```

produces

```
this is an options string
test string
```

its a pretty minor thing, not merge blocking!

sbsmith86:
I think the confusion is that we decided that 

`state = { campaign: { additionalContent: { displayAffilitateOptOut: 'test string' } } }`

would actually be

`state = { campaign: { additionalContent: { displayAffilitateOptOut: true } } }`

Because, semantically, that seems to be a boolean. So I would need to set the custom string somewhere. I could also grab that from some other value in the `additionalContent` json, i guess, but wanted to keep it simple so campaign leads didn't have to think about that.

itsjoekent:
Oh I understand now, alright that makes sense!

PR #451, opened 2017-10-13
Updating to add separate ThirdPartyAction component and new Actions directory.

weerd:
### What does this PR do?
This PR adds a new `ThirdPartyAction` component which will _eventually_ be a custom type for allowing more flexibility when we have external actions that can be done in a campaign or otherwise (for now just in campaign). It's very similar to the `ActionStep` component but I decided to split it as its own component to help with future work on having actions be their own components and not necessarily be forced to be a "step" as well as allow for more flexibility in more unique action items.

This particular component is using the `customBlock` content type but renders as the `ThirdPartyAction` component and retrieves the user's Northstar ID to append to a custom external link that needed the id as a url param. URL params for now can be passed as a series of strings in an array in the `additionalContent` field.

Refs https://www.pivotaltracker.com/story/show/151885515

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -125,6 +125,10 @@ class Campaign extends Entity implements JsonSerializable
             $step->customType ? $data['customType'] = $step->customType->first() : null;
             $step->additionalContent ? $data['additionalContent'] = $step->additionalContent : null;
 
+            if ($step->type) {
+                $data['customType'] = $step->type;
+            }
+
             return $data;
         });
     }
--- a/resources/assets/components/ActionPage/ActionStepsWrapper.js
+++ b/resources/assets/components/ActionPage/ActionStepsWrapper.js
@@ -6,6 +6,7 @@ import Revealer from '../Revealer';
 import { makeHash } from '../../helpers';
 import { Flex, FlexCell } from '../Flex';
 import { PostGalleryContainer } from '../Gallery/PostGallery';
+import { ThirdPartyActionContainer } from '../Actions/ThirdPartyAction';
 import { ReportbackUploaderContainer } from '../ReportbackUploader';
 import { CompetitionBlockContainer } from '../CompetitionBlock';
 import { SubmissionGalleryContainer } from '../Gallery/SubmissionGallery';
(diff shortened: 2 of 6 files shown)


app/Entities/Campaign.php
```diff
@@ -125,6 +125,10 @@ public function parseActionSteps($actionSteps)
             $step->customType ? $data['customType'] = $step->customType->first() : null;
             $step->additionalContent ? $data['additionalContent'] = $step->additionalContent : null;
 
+            if ($step->type) {
```
itsjoekent: Not sure I understand why this is needed, the line above handles this no?

```
$step->customType ? $data['customType'] = $step->customType->first() : null;
```

resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -69,6 +70,22 @@ const ActionStepsWrapper = (props) => {
       case 'submission-gallery':
         return isSignedUp ? submissionGallery : null;
 
+      case 'third-party-action':
```
DFurnes: A thought for the future: it'd be nice if we could offload most of this work of finding and creating compoents to `Block` and just deal with step indexes (and anything else unique to the action page) here.

weerd:
Forgot to mention that on the Contentful side, in the content field, you designate what to replace for the dynamic link using `:::dynamicLink:::`. So as an example:

```
[Click here](:::dynamicLink:::) and it will go to external location and add info.
```

resources/assets/components/ActionPage/ActionStepsWrapper.js
weerd: Yeah, I definitely want to change how that works... Ideally we eventually won't need the `ActionsStepsWrapper` anymore in the future.

resources/assets/components/Actions/ThirdPartyAction/ThirdPartyAction.js
```diff
@@ -0,0 +1,70 @@
...
+import StepHeader from '../../ActionPage/StepHeader';
+
+const ThirdParyAction = (props) => {
+  const {
+    content, dynamicLink, dynamicUrlParams, hideStepNumber, stepIndex,
+    template, title, userId,
+  } = props;
+
+  const urlParams = dynamicUrlParams.map((param) => {
+    if (param === 'northstarId') {
```
itsjoekent: This is a really minor point but it wasn't clear at first why this `if` was needed.

Maybe something like this?

```
const value = param === 'northstarId' ? userId : props[param];
return `${param}=${value}`;
```

resources/assets/components/Actions/index.js
```diff
@@ -0,0 +1 @@
+export * from './ThirdPartyAction';
```
itsjoekent: I don't think file exists yet?

resources/assets/components/Actions/ThirdPartyAction/ThirdPartyAction.js
weerd: Noice! Yeah, I like that. It's mostly because we told the third party it'd be `northstarId` and while I'd prefer to have the url param be `userId` I didn't want to change things up last minute for them.

resources/assets/components/Actions/index.js
weerd: this [one](https://github.com/DoSomething/phoenix-next/pull/451/files#diff-da92ed64739ebba9258635650f4dd574)

PR #452, opened 2017-10-13
Enforcing landscape for shares

itsjoekent:
### What does this PR do?
We're currently hitting the 8mb photo limit on Facebook for one of our campaigns, this PR ensures we always send a standard sized landscape image in our open graph tags.

Tested with a normal campaign and a social override

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -176,7 +176,7 @@ class Campaign extends Entity implements JsonSerializable
             'coverImage' => [
                 'description' => $this->coverImage ? $this->coverImage->getDescription() : '',
                 'url' => get_image_url($this->coverImage),
-                'landscape_url' => get_image_url($this->coverImage, 'landscape'),
+                'landscapeUrl' => get_image_url($this->coverImage, 'landscape'),
             ],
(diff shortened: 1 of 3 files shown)


itsjoekent:
screen shot 2017-10-13 at 5 09 54 pm

app/Entities/Campaign.php
```diff
@@ -176,6 +176,7 @@ public function jsonSerialize()
             'coverImage' => [
                 'description' => $this->coverImage ? $this->coverImage->getDescription() : '',
                 'url' => get_image_url($this->coverImage),
+                'landscape_url' => get_image_url($this->coverImage, 'landscape'),
```
weerd: why do you have to set this here? Doesn't just specifying a `type` in the `get_image_url()` in the social override work?

itsjoekent: That works if it has a social override, which almost nothing does currently

weerd: ah, jes. ok 🤔

itsjoekent: also this should be camel case, whoops

PR #453, opened 2017-10-16
Slideshow

itsjoekent:
### What does this PR do?
- Adds a Slideshow component. (See section below for implementation details)
- Implements the Slideshow in the Post Signup Modal

See it in action: 

In order to use the Slideshow, simply drop in the component and each child will be treated as a slide. You can have multiple slide shows, because each has its own counter for the current slide. The slideshow container takes care of identifying the current slide and slicing it from the children (This sentence sounds horrifying).

Things that could be a later TODO
- Supply a class name, like how we do for the Card component
- Support a "back" button
- Automatically advance the slide when you join a competition? (@product)
- More styling options for the navigation?

https://www.pivotaltracker.com/story/show/151721587

### DIFF
--- a/resources/assets/actions/slideshow.js
+++ b/resources/assets/actions/slideshow.js
@@ -1,5 +1,5 @@
 import { NEXT_SLIDE } from '../actions';
 
-export function nextSlide(slideshowId) {
+export function nextSlide(slideshowId) { // eslint-disable-line import/prefer-default-export
   return { type: NEXT_SLIDE, slideshowId };
 }
--- a/resources/assets/actions/slideshow.js
+++ b/resources/assets/actions/slideshow.js
@@ -1,0 +1,5 @@
+import { NEXT_SLIDE } from '../actions';
+
+export function nextSlide(slideshowId) {
+  return { type: NEXT_SLIDE, slideshowId };
+}
--- a/resources/assets/components/Modal/configurations/PostSignupModal.js
+++ b/resources/assets/components/Modal/configurations/PostSignupModal.js
@@ -2,11 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { AffirmationContainer } from '../../Affirmation';
 import { CompetitionBlockContainer } from '../../CompetitionBlock';
-import Slideshow from '../../Slideshow';
+import SlideshowContainer from '../../Slideshow';
 
-const PostSignupModal = ({ competitionStep }) => (
+const PostSignupModal = ({ competitionStep, closeModal }) => (
   <div className="modal__slide">
-    <Slideshow slideshowId="post signup modal">
+    <SlideshowContainer slideshowId="post signup modal" onComplete={closeModal}>
       { competitionStep ? (
         <CompetitionBlockContainer
           content={competitionStep.content}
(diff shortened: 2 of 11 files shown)


resources/assets/actions/slideshow.js
```diff
@@ -0,0 +1,5 @@
+import { NEXT_SLIDE } from '../actions';
+
+export function nextSlide(slideshowId) { // eslint-disable-line import/prefer-default-export
```
mendelB: Just out of curiosity: Is this an old-style '`function`' for a specific reason? Also, I imagine the reason for the `eslint` ignore is bc we anticipate more actions for Slideshow in the future?

weerd:
>  takes care of identifying the current slide and slicing it from the children 😱

resources/assets/actions/slideshow.js
itsjoekent: Indeed! Check out the other action files, its just the general style used for actions. As for the lint ignore, yep!

resources/assets/components/Modal/configurations/PostSignupModal.js
```diff
@@ -2,25 +2,25 @@ import React from 'react';
...
 
-const PostSignupModal = ({ competitionStep }) => (
-  <article>
-    <div className="modal__slide">
-      <AffirmationContainer />
-    </div>
-    <div className="modal__slide">
+const PostSignupModal = ({ competitionStep, closeModal }) => (
+  <div className="modal__slide">
+    <SlideshowContainer slideshowId="post signup modal" onComplete={closeModal}>
```
weerd: can we hyphenate the `slideshowId` or something? It was real weird reading it and I totally was confusing them for some sort of series of class names or something 🤔

itsjoekent: Like `slideshowTitle`?

weerd: oh sorry, I meant hyphenate the id value `post-signup-modal` cause it seems otherwise like a series of class names. Not used to seeing a space in between items in an attribute unless they're class names. What are you using this ID for? Puck data tracking?

itsjoekent: ohhhhhh ok. yeah sure that's an easy change, the code just cares about how unique it is.

and currently there is no data tracking, the ID is for allowing multiple slideshows to co-exist.

resources/assets/components/Modal/configurations/PostSignupModal.js
```diff
@@ -2,25 +2,25 @@ import React from 'react';
...
 
-const PostSignupModal = ({ competitionStep }) => (
-  <article>
-    <div className="modal__slide">
-      <AffirmationContainer />
-    </div>
-    <div className="modal__slide">
+const PostSignupModal = ({ competitionStep, closeModal }) => (
+  <div className="modal__slide">
+    <SlideshowContainer slideshowId="post-signup-modal" onComplete={closeModal}>
```
weerd: 💃

itsjoekent:
🚂 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋 🚋

PR #455, opened 2017-10-17
Making sure additional content is an object not null

itsjoekent:
Currently helping-hands-puerto-rico has a null additionalContent field. It's not clear why but it's preventing people from signing up because we're accessing a property of null. This is a temp fix to quickly move to prod.


DFurnes:
In the future, I think it probably makes sense to use Lodash's [`get`](https://lodash.com/docs/#get) helper wherever we are grabbing from `additionalContent` so we get more predictable error handling.

itsjoekent:
@DFurnes for sure!

PR #457, opened 2017-10-17
Sticky join button mobile

mendelB:
### What does this PR do?
Adds sticky "Join Us" button to bottom of campaign page on mobile.

sticky nav
 

It's just a matter of placing the bare bones `CallToAction` component at the end of `CampaignPage`. The button in the revealer at the bottom of most pages disappears on small screens and the cta button appears and is sticky af. Pages that have a `CallToActionBlock` at the bottom such as `ComponentSubPage` will still show their own Join Us button, and if that's a problem we can work on some fix.

https://www.pivotaltracker.com/story/show/151861749

### Checklist

### DIFF
--- a/resources/assets/components/CallToAction/CallToAction.js
+++ b/resources/assets/components/CallToAction/CallToAction.js
@@ -11,13 +11,17 @@ type CallToActionProps = {
   className: ?string,
 };
 
-const CallToAction = ({ legacyCampaignId, className }: CallToActionProps) => (
-  <div className={classnames('call-to-action', className)}>
-    { SignupButtonFactory(({ clickedSignUp }) => (
-      <Button onClick={() => clickedSignUp(legacyCampaignId)} />
-    ), 'call to action', { text: 'join us' })}
-  </div>
-);
+const CallToAction = ({ legacyCampaignId, className }: CallToActionProps) => {
+  const SignupButton = SignupButtonFactory(({ clickedSignUp }) => (
+    <Button onClick={() => clickedSignUp(legacyCampaignId)} />
+  ), 'call to action', { text: 'join us' });
+
+  return (
+    <div className={classnames('call-to-action', className)}>
+      <SignupButton />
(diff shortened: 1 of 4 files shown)


resources/assets/components/ActionPage/ActionStepsWrapper.js
```diff
@@ -38,10 +39,14 @@ const ActionStepsWrapper = (props) => {
       callToAction={callToAction}
       isLoading={hasPendingSignup}
       onReveal={() => clickedSignUp(campaignId, { source: 'action page revealer' })}
-      isAuthenticated={isAuthenticated}
+      isAuthenticated={isAuthenticated && isSignedUp}
```
itsjoekent: I think this could probably just be `isSignedUp`

weerd: Wait, why was `isSignedUp` added to this?

mendelB: Otherwise, if the user is authenticated but not signed up the button would not get the `is-cta` class and wouldn't be removed on mobile to make place for the sticky button

weerd:
Could we actually move it to [here](https://github.com/DoSomething/phoenix-next/blob/ff2b4fae779128af8efceabcacce7b5117a1ddab/resources/assets/components/Page/CampaignPage/CampaignPage.js#L100), before the closing `</div>` and then it applies to all campaign pages (except the landing page) and toss it in as:

```
<div className="main clearfix">
  // ...stuff before this inside of "main" div...
  { isAffiliated ? null : <CallToActionContainer className="-sticky" /> }
</div>
```

mendelB:
Updated to implement what @weerd mentioned. This simplifies things and ensures that the sticky CTA is omnipresent across campaign pages. I also removed the `isAuthenticated` prop value altogether as it defaults to `false` anyway which solves our problems

PR #458, opened 2017-10-18
Revealer button shenanigans

mendelB:
### What does this PR do?
Replaces the `isAuthenticated` prop in the `Revealer` component with `isSignedUp`. 
Adds hide button logic to `CallToActionBlock` for non-affiliated user.

The revealer adds an `is-cta` class to its button if the user is not authenticated. This class gets rid of the button for mobile, intended to make way for the sticky CTA button. 
This resulted in some collision though when the user is authenticated but not signed up for the campaign, so we still want to display the sticky CTA, but the Revealer's button - having no `is-cta` class - remains which results in two buttons. Using `isSignedUp` instead fixes this.

Adding this same logic to the `CallToActionBlock` allows for solving the same issue.

image

image

#457 

### Checklist

### DIFF
--- a/resources/assets/components/CallToActionBlock/cta.scss
+++ b/resources/assets/components/CallToActionBlock/cta.scss
@@ -89,4 +89,12 @@
   .cta__impact-number {
     font-size: $font-hero;
   }
+
+  .button.join-us {
+    display: none;
+
+    @include media($medium) {
+      display: inline-block;
+    }
+  }
 }
--- a/resources/assets/components/CallToActionBlock/cta.scss
+++ b/resources/assets/components/CallToActionBlock/cta.scss
@@ -90,7 +90,7 @@
     font-size: $font-hero;
   }
 
-  .button.join-us {
+  .last-cta.is-not-affiliated {
     display: none;
 
     @include media($medium) {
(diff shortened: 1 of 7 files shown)


weerd:
> Replaces the isAuthenticated prop in the Revealer component with isSignedUp.

@mendelB so `isAuthenticated` is different from `isSignedUp` though... just want to make sure we're using the correct, intended logic.

mendelB:
Right, my thought was: If we anyways don't want to show the non-signed-up user the 'View More' button, but rather the 'Join Us' button, then why not use *that* logic to determine whether to add the `is-cta` class? (In `Feed`, we only show 'view more' *if* they're signed up)

resources/assets/components/CallToActionBlock/cta.scss
```diff
@@ -89,4 +89,12 @@
   .cta__impact-number {
     font-size: $font-hero;
   }
+
+  .button.join-us {
```
weerd: can probs just use `.join-us` here, instead of the more specific `.button.join-us`... however, on an additional note could you rename `.join-us` to a more flexible class name... it'll be confusing if the button text changes and says "Sign Up" but the class is `.join-us`... just something to consider when naming classes... don't go by what the text of the button would necessarily say in a single scenario.

PR #462, opened 2017-10-19
adding background to footnote details for legibility

mendelB:
### What does this PR do?
Adds background to footnote details based for better legibility.

Styling is copied from what we have on legacy

image

image

https://www.pivotaltracker.com/n/projects/2019313/stories/151953742

### Checklist

### DIFF
--- a/resources/assets/components/AffiliateOption/affiliateOption.scss
+++ b/resources/assets/components/AffiliateOption/affiliateOption.scss
@@ -21,8 +21,8 @@
   .footnote-details {
     display: none;
     background-color: rgba(0,0,0,.7);
-    border-radius: 4px;
-    padding: 5px;
+    border-radius: $lg-border-radius;
+    padding: $half-spacing / 2;
 
     &.is-expanded {
       display: block;
--- a/resources/assets/components/AffiliateOption/affiliateOption.scss
+++ b/resources/assets/components/AffiliateOption/affiliateOption.scss
@@ -20,6 +20,9 @@
 
   .footnote-details {
     display: none;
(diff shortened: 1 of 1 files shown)


resources/assets/components/AffiliateOption/affiliateOption.scss
```diff
@@ -20,6 +20,9 @@
 
   .footnote-details {
     display: none;
+    background-color: rgba(0,0,0,.7);
+    border-radius: 4px;
```
itsjoekent: `$lg-border-radius: 4px;`

https://github.com/DoSomething/forge/blob/dev/scss/_utilities/_variables.scss#L103

resources/assets/components/AffiliateOption/affiliateOption.scss
```diff
@@ -20,6 +20,9 @@
...
+    padding: 5px;
```
itsjoekent: Why 5px? I think we should stick with our standard sizing rules to keep things consistent

mendelB: I was just copying the code from legacy but yeah that makes sense! I'll apply the forge variables though I don't see one for `padding`?

itsjoekent: Try this one? not sure how it will look https://github.com/DoSomething/phoenix-next/blob/c0bf6b264086f3ec9b2453696a3ba5483aaebd36/resources/assets/scss/base.scss#L125

mendelB:
@joe I used `half-spacing / 2` which is six, so not terribly more than before.  Does that work? 

image

PR #465, opened 2017-10-20
Progress on adding CTA entity.

weerd:
### What does this PR do?
This PR does some initial work for the upcoming `CallToAction` component updates. I took an approach that will end up being more versatile and not require us to make sure we update all the old instances of the CTA as a "Custom Block" with `join_cta` as the type. This approach takes both in and transforms the resulting `CallToAction` entity so regardless of what content type is used (moving forward with `callToAction` and still allowing/not-breaking the old `customBlock`).

CTAs rendering (top is the new content type, bottom is old content type):
screen shot 2017-10-20 at 5 00 26 pm

The two items in Contentful:
image

This will help setup work for the updates to the `CallToAction` component.

Refs https://www.pivotaltracker.com/story/show/151174012

### DIFF
--- a/app/Entities/CallToAction.php
+++ b/app/Entities/CallToAction.php
@@ -6,19 +6,42 @@ use JsonSerializable;
 
 class CallToAction extends Entity implements JsonSerializable
 {
+    /**
+     * Convert the object into something JSON serializable.
+     *
+     * @return array
+     */
     public function jsonSerialize()
     {
-        // dd($this);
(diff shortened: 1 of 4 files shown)


weerd:
Not sure if a [`switch`](https://github.com/DoSomething/phoenix-next/pull/465/files#diff-0c32a1733800aaaefd914f70ded4abc3R16) was necessarily the best approach here, but seems to work well and also reads pretty easily.

PR #466, opened 2017-10-23
Shifting to a single CTA component.

weerd:
### What does this PR do?
This PR established an official CTA component that eliminates the need for the custom block component as cta (code for old approach will be deleted in upcoming PR). It allows still supporting old style of cta entered on the Contentful side, but all CTAs now use the new component when rendering onto the page. It also makes a few other updates to clean things up relating around CTAs.

There are a few fields that have not been added to contentful so they have only been partially included in the CTA component, but will be getting added/updated in an upcoming PR.

Light version (with varying content):
image

Dark version (with varying content and as sidebar):
image

Transparent version (with varying content):
image

Refs: https://www.pivotaltracker.com/story/show/151174012

### Checklist

### DIFF
--- a/app/Entities/CallToAction.php
+++ b/app/Entities/CallToAction.php
@@ -35,13 +35,16 @@ class CallToAction extends Entity implements JsonSerializable
             'id' => $this->entry->getId(),
             'type' => $type,
             'fields' => [
+                'buttonText' => 'Join us',
                 'content' => $content,
-                'displayOptions' =>$this->displayOptions->first(),
+                'displayOptions' => $this->displayOptions->first(),
                 'impactPrefix' => $impactPrefix,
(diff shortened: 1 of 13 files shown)


app/Entities/CallToAction.php
```diff
@@ -35,13 +35,16 @@ public function jsonSerialize()
             'id' => $this->entry->getId(),
             'type' => $type,
             'fields' => [
+                'buttonText' => 'Join us',
```
itsjoekent: Why is this hardcoded?

weerd: > There are a few fields that have not been added to contentful so they have only been partially included in the CTA component, but will be getting added/updated in an upcoming PR.

Haven't added to Contentful yet ;)

PR #467, opened 2017-10-23
formatting contentful image urls within provided merkdown string

mendelB:
### What does this PR do?
Formats any contentful image url within the string of markdown provided to the `Markdown` component, adding formatting through the `contentfulImageUrl` helper.

We're loading yuuuuge images by default which is killing our google score and performance speeds.

I somewhat arbitrarily decided on the measurements here (`w=1000, h=700`) Please let me know if you have an opinion on this. It's based on messing around with this campaign [https://www.dosomething.org/us/campaigns/sincerely-us](https://www.dosomething.org/us/campaigns/sincerely-us), and working to replace some images in the community tab campaign updates with a smaller version without reducing their size on the actual page:

*Original*:
image

*If formatted with `500/500`*:
image

*With current formatting* (actual image is smaller than original though):
image

Raw Image URL:
http://images.contentful.com/81iqaqpfd8fy/4TGGGR6PC8mMmMgiySK8Kq/9e98f4ed5cbcc1c5c3de28d4f9d17dd5/4V4A0326.jpg

https://www.pivotaltracker.com/story/show/152201463

### Checklist

### DIFF
--- a/resources/assets/components/Markdown/index.js
+++ b/resources/assets/components/Markdown/index.js
@@ -5,9 +5,9 @@ import { markdown, contentfulImageUrl } from '../../helpers';
 
 import './markdown.scss';
 
-const urlToFormat = /\/\/images\.contentful\.com.+\.(jpg|png)/g;
-const formatImageUrl = url => (contentfulImageUrl(url, '1000', '700', 'fill'));
-const formatImageUrls = string => (string.replace(urlToFormat, formatImageUrl));
+const pattern = /\/\/images\.contentful\.com.+\.(jpg|png)/g;
+const contentfulImageFormat = url => (contentfulImageUrl(url, '1000', '700', 'fill'));
+const formatImageUrls = string => (string.replace(pattern, contentfulImageFormat));
 
 const Markdown = ({ className = null, children }) => (
   <div className={classnames('markdown', 'with-lists', className)} dangerouslySetInnerHTML={markdown(formatImageUrls(children))} /> // eslint-disable-line react/no-danger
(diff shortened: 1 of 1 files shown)


resources/assets/components/Markdown/index.js
```diff
@@ -1,12 +1,15 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import classnames from 'classnames';
-import { markdown } from '../../helpers';
+import { markdown, contentfulImageUrl } from '../../helpers';
 
 import './markdown.scss';
 
+const formatImageUrl = url => (contentfulImageUrl(url, '1000', '700', 'fill'));
+const formatImageUrls = string => (string.replace(/\/\/images\.contentful\.com.+\.jpg/g, formatImageUrl));
```
weerd: do we want to allow for jpg/png? 🤷‍♂️

mendelB: ooh good point! Are there any other formats we might want to include?

weerd: not that I can think of offhand, but at least adding an additional one leaves it open in the regex formula to easily add more if needed

resources/assets/components/Markdown/index.js
```diff
@@ -1,12 +1,16 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import classnames from 'classnames';
-import { markdown } from '../../helpers';
+import { markdown, contentfulImageUrl } from '../../helpers';
 
 import './markdown.scss';
 
+const urlToFormat = /\/\/images\.contentful\.com.+\.(jpg|png)/g;
```
weerd: call this the `pattern`

resources/assets/components/Markdown/index.js
```diff
@@ -1,12 +1,16 @@
...
+const formatImageUrl = url => (contentfulImageUrl(url, '1000', '700', 'fill'));
```
weerd: update the name of these two... just sounds a little weird one just being the plural of the other :\

PR #468, opened 2017-10-24
Community post signup

itsjoekent:
### What does this PR do?

1. *NEW* The community tab now shows for affiliated users on the legacy template, along with the tabbed navigation.
screen shot 2017-10-24 at 10 32 59 am

2. this prompted us to realize we need the ability to potentially hide additional pages from the nav. so i made a quick boolean value in contentful
screen shot 2017-10-24 at 10 45 02 am
screen shot 2017-10-24 at 10 44 51 am

https://www.pivotaltracker.com/story/show/152151797

### DIFF
--- a/resources/assets/components/Navigation/TabbedNavigationContainer.js
+++ b/resources/assets/components/Navigation/TabbedNavigationContainer.js
@@ -13,6 +13,7 @@ import { isCampaignClosed } from '../../helpers';
 import SignupButtonFactory from '../SignupButton';
 
 const mapStateToProps = state => ({
+  hasActivityFeed: Boolean(state.campaign.activityFeed.length),
   isAffiliated: state.signups.thisCampaign,
   legacyCampaignId: state.campaign.legacyCampaignId,
   pages: state.campaign.pages,
@@ -22,7 +23,10 @@ const mapStateToProps = state => ({
 });
 
 const TabbedNavigationContainer = (props) => {
-  const { isAffiliated, legacyCampaignId, pages, campaignEndDate, template } = props;
+  const {
+    hasActivityFeed, isAffiliated, legacyCampaignId,
+    pages, campaignEndDate, template,
+  } = props;
 
   if (template === 'legacy' && ! isAffiliated) {
     return null;
@@ -49,10 +53,12 @@ const TabbedNavigationContainer = (props) => {
     <Button className="-inline nav-button" onClick={() => clickedSignUp(legacyCampaignId)} />
   ), 'tabbed navigation', { text: 'join us' });
 
+  const hideCommunity = (template === 'legacy') && ! hasActivityFeed;
+
   return (
     <TabbedNavigation>
       <div className="nav-items">
-        <NavigationLink to={join('/us/campaigns', campaignSlug, campaignPaths.community)} exact>Community</NavigationLink>
+        { hideCommunity ? null : <NavigationLink to={join('/us/campaigns', campaignSlug, campaignPaths.community)} exact>Community</NavigationLink> }
         { isClosed ? null : <NavigationLink to={join('/us/campaigns', campaignSlug, campaignPaths.action)}>Action</NavigationLink> }
         { additionalPages }
       </div>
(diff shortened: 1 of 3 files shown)


resources/assets/components/Navigation/TabbedNavigationContainer.js
```diff
@@ -34,6 +34,10 @@ const TabbedNavigationContainer = (props) => {
 
   // Create links for additional "content" pages on this campaign in Contentful.
   const additionalPages = pages.map((page) => {
+    if (page.fields.hideFromNavigation) {
```
weerd: Something to keep in mind is that we'll likely end up using the same `Page` content type eventually for static pages as well, so we should be mindful of what fields we add. That being said, I could see this field helping in other scenarios for static pages as well 🤔

resources/assets/components/Page/CampaignPage/CampaignPage.js
```diff
@@ -62,11 +62,13 @@ const CampaignPage = (props) => {
...
-                :
-                <FeedContainer />
-              )}
+              render={() => {
+                if (template === 'legacy') {
+                  return isAffiliated ? <FeedContainer /> : <ActionPageContainer />;
+                }
+
+                return <FeedContainer />;
+              }}
```
weerd: @itsjoekent see convo in Slack, but we'd prefer to only show the community feed _if_ there's items in the reference field. So we should change this logic up a bit.

itsjoekent:
screen shot 2017-10-24 at 1 32 58 pm

itsjoekent:
screen shot 2017-10-24 at 1 36 27 pm

resources/assets/components/Page/CampaignPage/CampaignPage.js
```diff
@@ -64,7 +64,8 @@ const CampaignPage = (props) => {
               exact
               render={() => {
                 if (template === 'legacy') {
-                  return isAffiliated ? <FeedContainer /> : <ActionPageContainer />;
+                  const displayFeed = (isAffiliated && hasActivityFeed);
```
itsjoekent: hm, I guess this could be `shouldDisplayFeed`

resources/assets/components/Page/CampaignPage/CampaignPage.js
```diff
@@ -62,11 +62,14 @@ const CampaignPage = (props) => {
...
               exact
-              render={() => (template === 'legacy' ?
-                <ActionPageContainer />
-                :
-                <FeedContainer />
-              )}
+              render={() => {
+                if (template === 'legacy') {
+                  const shouldDisplayFeed = (isAffiliated && hasActivityFeed);
+                  return shouldDisplayFeed ? <FeedContainer /> : <Redirect to={`${match.url}/action`} />;
```
weerd: I think this might be incorrect... 🤔

weerd: I don't think you need to bother with `isAffiliated`. If it's the legacy experience then they won't make it this far if they're un-affiliated, right?

weerd: also, I think we do need to use the `<ActionPageContainer />` if `hasActivityFeed` is false.

itsjoekent: re: `isAffiliatted` -- Right I guess since they have the pitch page it doesn't technically matter, I would call it unnecessary over incorrect though. regardless, I'll get rid of that bit!

re: using `<ActionPageContainer />` -- I intentionally made this swap to the redirect for two reasons. 1. When users are able to see the tabs, you could be on the root `/` and not have any tabs highlighted, because the action tab only checks if you're on `/action`. 2. It's very confusing for me whenever I'm on `missing-history` or `missing-history/action` because it's two different url's for the exact same page

weerd:
[Replying to this message](https://github.com/DoSomething/phoenix-next/pull/468#discussion_r146642068)

Yeah, but if you're on a legacy template campaign _without_ a campaign activity feed, we just want to see the action page stuff at `/missing-history` which does what we want and mimics the legacy experience.

Eventually we want it all to just merge into one experience, but not the case now.

itsjoekent:
@weerd If you went to the `/missing-history` link you would still see the action page, it would just append `/action` to the url? I'll just add the container back though

weerd:
image

There might be some confusion there. I get that `/action` would go to the action page, but right now on legacy template campaigns that don't have any tabbed navigation, etc... going to the root url for the campaign (if signed up) takes you to the action page without `/action` in the url which is what we were mimicking with the legacy experience. see the screenshot above.

PR #470, opened 2017-10-24
Add "phoenix:setup" command.

DFurnes:
### What does this PR do?
Surprise! Added a `phoenix:setup` command.

screen shot 2017-10-24 at 3 55 39 pm

Following the same format used in our other apps, but also asking for Contentful vars.

🌱

### Checklist


weerd:
@DFurnes quick question... this doesn't conflict w/ the `post-create-project-cmd` in composer.json right?

DFurnes:
> @DFurnes quick question... this doesn't conflict w/ the `post-create-project-cmd` in composer.json right? or the `post-root-package-install`?

Oh wow, I didn't know `post-root-package-install` existed! I'm partial to leaving `php artisan phoenix:setup` as a manual step since that makes it clear how to run it again if you ever wanted to reset your environment, but not totally opposed to adding it there.

Either way shouldn't cause an issue, the install script works happily with or without an existing `.env` file. And definitely no conflict with `post-create-project-cmd` since that only runs after the Laravel `laravel new` command is run.

PR #471, opened 2017-10-24
Quick cache campaign™

mendelB:
### What does this PR do?
- Implements a bare-bones caching strategy for Campaigns to prevent a trazillion calls to Contentful API

Using the `Cache` class in the `CampaignRepository` repository to stash the campaign for 15-minute increments.
The annoying thing was that once you serialize the `campaign` data, contentful cries when you try and create a `Campaign` entity with that data, for it requires a `DynamicEntry` argument to the constructor. 
A little 🔍 revealed a `reviveJson` method in our Contentful API client, allowing us to use the cached JSON to create a campaign and help make the world a better place!

https://www.pivotaltracker.com/story/show/152202846

### Checklist


app/Repositories/CampaignRepository.php
```diff
@@ -45,19 +48,37 @@ public function getAll()
...
     {
-        $query = (new Query)
+        if (Cache::has($slug)) {
+            $cachedCampaignJson = Cache::get($slug);
+
+            // This turns the JSON into a Contentful Dynamic Entry
+            // so we can create a campaign entity out of the cached data
+            $revivedContentfulEntry = $this->client->reviveJson($cachedCampaignJson);
+
+            $campaign = new Campaign($revivedContentfulEntry);
```
weerd: Line 58 and 78 are pretty much the same. We can likely take them out of the conditional and place right before the `return` call, and then just rename the var to something that's the same in both conditions... I'll roll over IRL :)

app/Repositories/CampaignRepository.php
```diff
@@ -45,19 +48,35 @@ public function getAll()
...
-        $campaigns = $this->makeRequest($query);
+            $campaigns = $this->makeRequest($query);
+
+            if (! $campaigns->count()) {
+                throw new ModelNotFoundException;
+            }
+
+            $campaignEntry = $campaigns[0];
+
+            if (! App::environment(['local', 'staging'])) {
```
weerd: @mendelB @itsjoekent I'm debating if we would prefer to go by our environments, or potentially set a new env variable like `APP_CACHE=true` so it's something we can easily toggle on/off if we _want_ to see caching working locally or on staging etc... thoughts?

itsjoekent: Yeah I think that could be fine as well

mendelB: Cool yeah that feels like a good idea!

mendelB: @weerd So we'd be assuming that the `APP_CACHE` is set to `true` or `false` and basing caching on that? meaning if the variable does *not* exist it wouldn't cache?

itsjoekent:
@mendelB Can you use the config/services file https://github.com/DoSomething/phoenix-next/blob/79ca93cc5b40e8587e995e7cf29aaa8e3fb1f37c/config/services.php#L19 and reference it with the config function? https://github.com/DoSomething/phoenix-next/blob/cefffd6df344e1fe70f62e82acab44592e379b26/app/helpers.php#L270

app/Repositories/CampaignRepository.php
```diff
@@ -45,19 +47,35 @@ public function getAll()
...
-        $campaigns = $this->makeRequest($query);
+            $campaigns = $this->makeRequest($query);
+
+            if (! $campaigns->count()) {
+                throw new ModelNotFoundException;
+            }
+
+            $campaignEntry = $campaigns[0];
+
+            if (env('CONTENTFUL_CACHE')) {
```
weerd: Yeah use `config()` instead since that uses the cached config info

mendelB:
Which function in `helpers.php` would I be adding this to? Or create a new function there for this?

itsjoekent:
@mendelB Was just showing an example of the `config` function

mendelB:
Gotcha gotcha! was confused for a sec

config/services.php
```diff
@@ -55,4 +55,8 @@
         'prefix' => env('SIXPACK_COOKIE_PREFIX'),
         'timeout' => env('SIXPACK_TIMEOUT'),
     ],
+
+    'contentful' => [
+        'cache' => env('CONTENTFUL_CACHE'),
```
weerd: Maybe default this to `true`.

mendelB:
@weerd Should it also be set to `true` in the .env example?

PR #473, opened 2017-10-27
Contentful cache

itsjoekent:
### What does this PR do?
Ohboi.

- Adds a Campaign, Link model, and a pivot table to create a many to many relationship.
- Adds a webhook route to accept requests from contentful
-  Adds a /refresh route to manually break the cache for a given slug
- Adds logic to parse through a Campaign and extract id's from it
- Adds logic to parse through a Campaign model and update references in the database
- Adds several new Entity classes & updates relevant JS when needed
- Refactors the getBySlug function to have an additional parameter and integrate with the new cache system.

So, picture a simplified graph of the campaign model.

screen shot 2017-10-27 at 10 48 43 am

Notice here there are multiple problems to solve,

1) How do we know what campaigns ladder up from Author?
2) How do we handle a shared resource such as an Author or Affirmation?

That is what this PR set out to solve. 

To do this, I've made a system that maintains a reference of every contentful ID that relates back to a Campaign. Using the same data from the graph above, think of it like this,

screen shot 2017-10-27 at 10 52 25 am

This way, when `Author mnas90gsm` is updated, we know to only update `Campaign mkdioqwsa42` and `Campaign uicv8sdzdwe`.

One problem this doesn't cover is when we create something totally brand new, that might require different logic in the webhook controller.

There is also another issue Mendel realized with our dummy cache solution, that renders this entire PR useless, but I figured let's solve this one step at a time and get this out there.


weerd:
@itsjoekent doing some additional thinking on this. Mind outlining what data the Contentful webhooks can, say for a `CampaignUpdate` to PN?

weerd:
@itsjoekent mind adding some notes regarding what we were discussing with a more individual entity cache and the caveats/complications involved?

itsjoekent:
to recap what ive said to diego / dave & outlined in this PR,

Keeping a local copy of all contentful entries, each being there own DB record, would require a serious engineering effort to fully replicate all of the contentful logic and manage nested entry relationships. On top of the work to manage all of these webhook requests and mapping them back to DB models, we'd also have to rewrite how we "build" the campaign object to join a large series of DB models together.

This approach is a hybrid between a "dummy" cache (in that, it doesn't update until it hits a certain time limit or manually cleared), and a full-on Contentful re-creation in Phoenix Next. This PR lets us maintain a dead simple mapping of entries to campaigns, and thus intelligently break the cache in a smart manner but still leverage the contentful api in the way we have been.

PR #474, opened 2017-10-27
Temp caching solution fixes

mendelB:
### What does this PR do?
*Actually* caches all campaign data.

Previously in #471, we were caching only the raw campaign data. However most of the children of the campaign merely existed as `Links` without the content, and these Links would only be converted to full `DynamincEntry`'s with content (presumabely by always making yet more API calls to contentful) when the `json_encode` method was called on the campaign data in `helpers.php` - `get_social_fields`. (We were also calling it in `scriptify` which probably was triggering yet more calls.)

Now, we can have the `CampaignRepository`'s `findBySlug`, create the `Campaign`, call `json_encode` forcing all the children to be populated fully, and cache all that to begin with, saving all those extra API calls.

The only change is that now `CampaignRepository` is returning a std object of all campaign data, as opposed to an actual `CampaignEntity`. So `get_social_fields` had to be slightly adjusted to work with that.

https://www.pivotaltracker.com/story/show/152202846

### Checklist

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -56,19 +56,19 @@ class CampaignController extends Controller
      */
     public function show($slug)
     {
-        $campaignFlattened = $this->campaignRepository->findBySlug($slug);
+        $flattenedCampaign = $this->campaignRepository->findBySlug($slug);
 
         $env = get_client_environment_vars();
 
         // The slug argument will not contain `/blocks` or other path extensions, hence `::url()`.
-        $socialFields = get_social_fields($campaignFlattened, Request::url());
+        $socialFields = get_social_fields($flattenedCampaign, Request::url());
 
         return view('campaigns.show', [
-            'campaign' => $campaignFlattened,
+            'campaign' => $flattenedCampaign,
             'socialFields' => $socialFields,
             'env' => $env,
         ])->with('state', [
-            'campaign' => $campaignFlattened,
+            'campaign' => $flattenedCampaign,
             'share' => $socialFields,
             'user' => [
                 'id' => auth()->id(),
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -56,18 +56,19 @@ class CampaignController extends Controller
      */
     public function show($slug)
     {
-        $campaign = $this->campaignRepository->findBySlug($slug);
+        $campaignFlattened = $this->campaignRepository->findBySlug($slug);
+
         $env = get_client_environment_vars();
 
         // The slug argument will not contain `/blocks` or other path extensions, hence `::url()`.
-        $socialFields = get_social_fields($campaign, Request::url());
+        $socialFields = get_social_fields($campaignFlattened, Request::url());
 
         return view('campaigns.show', [
-            'campaign' => $campaign,
+            'campaign' => $campaignFlattened,
             'socialFields' => $socialFields,
             'env' => $env,
         ])->with('state', [
-            'campaign' => $campaign,
+            'campaign' => $campaignFlattened,
             'share' => $socialFields,
             'user' => [
                 'id' => auth()->id(),
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -56,19 +56,19 @@ class CampaignController extends Controller
      */
     public function show($slug)
     {
-        $flattenedCampaign = $this->campaignRepository->findBySlug($slug);
+        $campaign = $this->campaignRepository->findBySlug($slug);
 
         $env = get_client_environment_vars();
 
         // The slug argument will not contain `/blocks` or other path extensions, hence `::url()`.
-        $socialFields = get_social_fields($flattenedCampaign, Request::url());
+        $socialFields = get_social_fields($campaign, Request::url());
 
         return view('campaigns.show', [
-            'campaign' => $flattenedCampaign,
+            'campaign' => $campaign,
             'socialFields' => $socialFields,
             'env' => $env,
         ])->with('state', [
-            'campaign' => $flattenedCampaign,
+            'campaign' => $campaign,
             'share' => $socialFields,
             'user' => [
                 'id' => auth()->id(),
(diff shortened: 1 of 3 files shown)


app/Repositories/CampaignRepository.php
```diff
@@ -68,14 +64,18 @@ public function findBySlug($slug)
 
             $campaignEntry = $campaigns[0];
 
+            $campaign = new Campaign($campaignEntry);
+
+            $flattenedCampaign = json_decode(json_encode($campaign));
```
itsjoekent: really minor, let's use either `campaignFlattened` or `flattenedCampaign` but not both. i kinda like the second one more.

app/Http/Controllers/CampaignController.php
```diff
@@ -56,18 +56,19 @@ public function index()
      */
     public function show($slug)
     {
-        $campaign = $this->campaignRepository->findBySlug($slug);
+        $flattenedCampaign = $this->campaignRepository->findBySlug($slug);
```
weerd: I'm debating if it's cool that we just call this `$campaign`, and just understand that when it's coming from the repository it's a flattened version of the campaign.

itsjoekent: 👍

mendelB: Yeah that makes sense. What might we write in the param comment?
`* @param  Campaign $campaign`

weerd: param where?

mendelB: Oh my bad - I was referring to the `get_social_links` in `/helpers.php`
we use `$flattenedCampaign` a bunch there too

weerd: Just call it `$campaign` as well :)

app/helpers.php
```diff
@@ -219,7 +219,12 @@ function get_legacy_campaign_data($id, $key = null)
  */
 function useOverrideIfSet($field, $campaign, $override)
 {
-    $base = $campaign->{$field};
+    if (! isset($campaign->{$field})) {
```
weerd: I did a quick find in project, and do we actually use this `useOverrideIfSet ()` function anywhere? It seems like it is declared but never used... although may have missed it, but I did the search a few times 🤔

weerd: Nooope. Nevermind. Found where else it's used!

app/Repositories/CampaignRepository.php
```diff
@@ -48,11 +48,7 @@ public function getAll()
     public function findBySlug($slug)
     {
         if (Cache::has($slug)) {
-            $cachedCampaignJson = Cache::get($slug);
-
-            // This turns the JSON into a Contentful Dynamic Entry
-            // so we can create a campaign entity out of the cached data
-            $campaignEntry = $this->client->reviveJson($cachedCampaignJson);
+            $flattenedCampaign = Cache::get($slug);
```
DFurnes: Can we replace this with [`remember`](https://github.com/DoSomething/phoenix-next/blob/db6aca605e01b78d9f6a8d5a95d522cb5483ce0c/app/helpers.php#L28-L39) to simplify the conditional?

```php

$flattenedCampaign = remember('campaign_'.$slug, 15, function () {
    $query = (new Query)
        ->setContentType('campaign')
        ->where('fields.slug', $slug)
        ->setInclude(3)
        ->setLimit(1);

    $campaigns = $this->makeRequest($query);

    // ...

    return json_encode($campaignEntry);
});

return json_decode($flattenedCampaign);
```

DFurnes: Let's also pull the `ModelNotFoundException` outside of the cache function so can cache that as well (since otherwise every time someone requests a slug that 404'd it'd cause a new request to Contentful).

app/helpers.php
```diff
@@ -317,11 +321,18 @@ function get_login_query($campaign = null)
         return [];
     }
 
+    // This deals with when the $campaign is the flattenedCampaign std object
```
DFurnes: Is there a scenario where it wouldn't be now that we're flattening & caching these requests?

app/Repositories/CampaignRepository.php
DFurnes: Another quick thought, I assume the `json_encode` and `json_decode` are used to make sure we get a "serialized" version of the campaign entry to cache. If that's right, mind adding a quick comment? (Otherwise I could see us trying to refactor it to remove those in the future.)

app/helpers.php
weerd: So the `$campaign` is a general StdObject at this point, right? Can we update the docblock to indicate that, so we know it's not a `Campaign` object?

app/helpers.php
mendelB: Hmm I don't think so actually. Should I change this to only be dealing with the `flattenedCampaign`?

DFurnes: Yeah, I'd like to keep the type consistent wherever possible. 👍

app/Repositories/CampaignRepository.php
mendelB: @DFurnes regarding using the `remember` method, how would we deal with when the CONTENTFUL_CACHE is set to false? Could we possibly have a variable for the time that sets it to zero in that scenario?

DFurnes: Hmm, I'm into adding another `$shouldCache` parameter to `remember` to optionally skip caching (and instead just run the provided function every time). 

Also curious if we'd be safe with leaving caching enabled on local and just hitting `php artisan cache:clear` when necessary (since that keeps us closer to prod behavior as long as we're not constantly editing campaigns).

mendelB:
@DFurnes regarding:

> Hmm, I'm into adding another $shouldCache parameter to remember to optionally skip caching (and instead just run the provided function every time).

>  Also curious if we'd be safe with leaving caching enabled on local and just hitting php artisan cache:clear when necessary (since that keeps us closer to prod behavior as long as we're not constantly editing campaigns).

Are we cool with going with the latter idea (and I guess meaning not updating that `remember` method with a `$shouldCache` param)?

DFurnes:
@itsjoekent [brought up some concerns](https://dosomething.slack.com/archives/C3ASB4204/p1509386727000125). I don't have strong feelings, so it's whatever y'all decide. 👍

mendelB:
Gotcha, just saw that thread. I'll add that skip option in.

weerd:
Post-merge 👍 !

PR #476, opened 2017-10-30
Flex cell fixes.

weerd:
### What does this PR do?
This PR removes the `display: flex` style on the `.flex__cell` class which turns each flex item into a flex container as well. While this can be useful in some scenarios it was causing some unwanted issues with sidebar block sitting next to tall campaign updates for example, and forcing the sidebar block to stretch the full length of the campaign update:

image

This update now lets sidebar block only span the size they need to:

image

You can restore the old behavior of stretching vertically by now adding the `.display-flex` class to `FlexCell` components as needed.

This PR also removes the multiple nested flex__cell divs that were being wrapped around `ReportbackBlock` items, which are no longer necessary ever since the PR that started rendering them as individual items from a while back.

This PR relates to work on the CTA components for an upcoming PR.

Refs https://www.pivotaltracker.com/story/show/151174012

### DIFF
--- a/resources/assets/components/Feed/Feed.js
+++ b/resources/assets/components/Feed/Feed.js
@@ -15,18 +15,17 @@ import './feed.scss';
  * @param index
  * @returns {XML}
  */
-const renderFeedItem = (block, index) => {
-  console.log(block);
-
-
-  return (
-    <FlexCell key={`${block.id}-${index}`} width={block.fields.displayOptions} className={classnames('padded', {
+const renderFeedItem = (block, index) => (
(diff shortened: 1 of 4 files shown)


weerd:
There's an example in the [code](https://github.com/DoSomething/phoenix-next/pull/476/files#diff-1666018dacf606775d07f2b3913755d5R23) where I add the `display-flex` class to RB items... ideally we'll want to decide that if an RB is in a row with other RBs vs a single RB next to a campaign update, but we can tackle that in a future update.

PR #481, opened 2017-11-03
changing social share link to be modal instead of block

mendelB:
### What does this PR do?
- Changes the link generated by `makeShareLink` for `campaign` types, to be linking to the modal instead of the block. 
-  the facebook share button in `CampaignUpdate`s  and `CampaignUpdateBlock`s which make use of the `helpers#makeShareLink` to generate their social links, are now linking to modals as a result
- Another side effect of this is that the primary link atop a campaign update 
screen shot 2017-11-03 at 2 39 37 pm
(in both blocks and regular) now also link to the modal. (Since the link used is the same as the social link). (This was approved by Ashley/Adam).

Older `CampaignUpdateBlocks` don't render so nicely in modals yet as for now they're being returned plainly in a `Card`. I'll work on another pull request to support them in models.

https://www.pivotaltracker.com/story/show/152351435

### Checklist

### DIFF
--- a/resources/assets/helpers/index.js
+++ b/resources/assets/helpers/index.js
@@ -292,7 +292,7 @@ export function makeHash(string) {
 export function makeShareLink(type, options: { domain: string, slug?: string, key: string }) {
   switch (type) {
     case 'campaigns':
-      return `${options.domain}/us/campaigns/${options.slug}/blocks/${options.key}`;
+      return `${options.domain}/us/campaigns/${options.slug}/modal/${options.key}`;
 
     default:
       throw new Error('Please provide an expected section type for generating the link.');
--- a/resources/assets/helpers/test.js
+++ b/resources/assets/helpers/test.js
@@ -53,7 +53,7 @@ test('it makes the expected share link for a content item and app section type',
     key: '123',
   };
 
-  expect(makeShareLink('campaigns', options)).toBe('http://awesome.com/us/campaigns/seriously-awesome-campaign/blocks/123');
+  expect(makeShareLink('campaigns', options)).toBe('http://awesome.com/us/campaigns/seriously-awesome-campaign/modal/123');
 });
 
 test('it cannot make share link with an unknown app section type', () => {


itsjoekent:
Are we worried about this change from a product perspective, @lkpttn @ashleybaldwin ?

ashleybaldwin:
I'm not necessarily concerned by making the "campaign update" text link out to the modal. I chatted with adam about it and from other conversations with the campaigns team, they like the modal more for making links and i think giving them the ability to make that link easily by clicking the campaign update is actually nice.

mendelB:
Truth is though that clicking it wouldn't really give them the modal link since the URL changes back to the regular campaign URL almost instantaneously

mendelB:
image

ashleybaldwin:
right click > copy url for the win!!

DFurnes:
Gah yeah, we really need to get modals to use standard routing. I do kinda think it might make more sense to ask the campaigns team to manually swap `pages` to `modals` in the URL when sharing for now so we don't send users to pages they can't link to.

itsjoekent:
@DFurnes What do you mean by standard routing?

DFurnes:
Modals should be a normal React Router route (where whatever we're seeing on the page is reflected in the URL in the address bar), not a "redirect" to a Redux state.

itsjoekent:
That was intentional though so we could have "background" and not just a blank page behind the modal.

DFurnes:
We can do that with a normal route too, though.

itsjoekent:
True I guess you could still make the redux dispatch while also returning the template component?

DFurnes:
I don't think it should use Redux for this at all – this is a page routing concern, not an application state concern. We should be able to do this by including a `<Route>` component directly in the modal container that only pops content in there when on this route (and then the rest of the page still renders to the campaign based on the route prefix).

Part of [the philosophy](https://reacttraining.com/react-router/web/guides/philosophy) behind React Router v4 is that you can throw your routes wherever they make sense in the application, not just in a central "routes" component. So it's totally reasonable to have the `Modal` component handle whether or not to show this modal, while the `CampaignPage` renders what displays behind the modal.

Let me know if you want to chat IRL to scope out this work for a future sprint.

PR #485, opened 2017-11-06
adding twitter intent share button for campaignUpdates

mendelB:
### What does this PR do?
Adds a twitter icon with sharing capabilities (a link to a twitter intent pre-filled with update URL) to campaign updates (or any ShareContainer rendered with a `variant="icon"`)

- Unfortunately -along with the other faults of the tweet intent link- twitter won't give us a callback or trigger an event when the user actually tweets the tweet rather only when they click the link to tweet the tweet.
["Note that Web Intent events fire when the interaction occurs, not after the action completes."](https://dev.twitter.com/web/javascript/events)
So although there is a 'Twitter For Websites' JS plugin, for now, I don't believe there's even a reason to include it, as we can track the clicking ourselves thank you very much.

- One thing we could gain by using their plugin is that clicking the share button will pop up a modal, that is size formatted behind the scenes which feels a bit better than opening a new tab.
image
Not sure if that makes it worthwhile?

- The tweet button has kinda been lumped in with the facebook sharing button in the `Share` component. I wasn't sure if now was the time to start creating separate social share components for each service as the twitter stuff doesn't add much complexity as of yet. If I'm wrong about that please do let me know.

- @lkpttn If you have any design thoughts on this. Here's how it looks as of now:
image

https://www.pivotaltracker.com/story/show/152418373

### Checklist

### DIFF
--- a/resources/assets/components/Share/Share.js
+++ b/resources/assets/components/Share/Share.js
@@ -15,7 +15,7 @@ const Share = (props) => {
 
   const trackingData = { parentSource, variant, link, quote };
 
-  const onClick = () => {
+  const onFacebookClick = () => {
     trackEvent('clicked facebook share', trackingData);
 
     showFacebookSharePrompt({ href: link, quote }, (response) => {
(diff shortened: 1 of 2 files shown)


lkpttn:
These are in the wrong order, it should always be the FB f first, then the twitter bird. 

I'd also reduce the amount of space between the icons, it shouldn't be more than a half-spacer (12px).

resources/assets/components/Share/Share.js
```diff
@@ -27,14 +27,30 @@ const Share = (props) => {
...
-      onClick={onClick}
-    >
-      {variant === 'icon' ? null : 'share on'}
-      <i className="social-icon -facebook"><span>Facebook</span></i>
-    </button>
+    <div>
+      { variant === 'icon' ? (
+        <a href={`https://twitter.com/intent/tweet?url=${link}`} target="_blank" onClick={onTwitterClick}>
+          <button className={buttonClassName}>
+            {variant === 'icon' ? null : 'share on'}
```
itsjoekent: This will always equal 'icon' right?

mendelB: Unless it's rendered by an `Affirmation` or the `Dashboard`

itsjoekent: No, I meant,

L40: If this is an 'icon', use this component
L43: If this is a 'icon', use this text

But L43 will always equal icon?

mendelB: Oh yikes yes yes. indeed!

mendelB:
Added the twitter share prompt function from https://github.com/DoSomething/voting-app/blob/dev/resources/assets/js/utilities/share-link.js

resources/assets/helpers/index.js
```diff
@@ -390,3 +390,27 @@ export function showFacebookSharePrompt(share, callback) {
...
   }, callback);
 }
+
+/**
+ * Share a link by generating a Twitter intent share prompt.
+ *
+ * @param  {Object}   share
+ */
+export function showTwitterSharePrompt(share) {
+  const { href, quote } = share;
```
DFurnes: Can we have these as two distinct parameters on the function, rather than an object?

mendelB: I copied that pattern from the `showFacebookSharePrompt` function right above. Should I change it there as well?

DFurnes: Oh hmm, got it! Yeah I like that change, let's consider updating the other function too when we have some spare time.

PR #487, opened 2017-11-07
Add script to clear the cache when deploying on production.

weerd:
### What does this PR do?
Since we have a 15 minute cache, there could be a possible scenario where we make updates to Contentful, and make changes to code that address these updates, yet after a deploy to Production, the prod website is still referring to different content structure based on content that is cached and could lead to trouble. Thus, we should be clearing the application cache whenever we deploy to Production.

I haven't messed much with Wercker before, so please tell me if I this is correct or not!

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -21,6 +21,9 @@ deploy:
   steps:
     - bundle-install
     - script:
+        name: clear all cache
+        code: php artisan cache:clear
+    - script:
         name: write private key env var
         code: |-
           export CAP_PRIVATE_KEY=`mktemp`


DFurnes:
We want to add this to the [Capistrano config](https://github.com/DoSomething/phoenix-next/blob/dev/config/deploy.rb), which runs on the actual servers. Wercker is a separate "sandbox" style environment that we just use for builds & running tests.

blisteringherb:
To clear cache on the server itself, we'd want to put the cache clearing logic inside the deploy.rb file  so that it gets run through Capistrano

weerd:
hokay!

weerd:
Looks like the Capistrano file already does this: https://github.com/DoSomething/phoenix-next/blob/dev/config/deploy.rb#L44

PR #488, opened 2017-11-07
Cache refresh endpoint

mendelB:
### What does this PR do?
Adds a route/controller action (based on Joe's work in #473) to refresh the cache for specified campaign.

- The idea is for this to be triggered by a button in the soon to come Admin Dashboard™ so campaign leads can clear a specific campaigns cache and lead happier lives. The front end will be added in an upcoming pull request.
- I wasn't sure if the idea here was to 'refresh' or merely to 'clear' the cache. If our only intention here is to clear the cache then perhaps we might want a dedicated `forget` helper method that we could call from the controller and then redirect as opposed to putting this burden on the `CampaignRepository`?

### Checklist

### DIFF
--- a/app/Http/Controllers/CacheController.php
+++ b/app/Http/Controllers/CacheController.php
@@ -2,6 +2,8 @@
 
 namespace App\Http\Controllers;
 
+use URL;
+use Auth;
 use Illuminate\Http\Request;
 
 class CacheController extends Controller
@@ -15,10 +17,15 @@ class CacheController extends Controller
      */
     public function __invoke(Request $request, $cacheId)
     {
-        $redirectUrl = $request->query('redirect');
+        $redirectUrl = $request->query('redirect') ?: URL::previous();
 
-        app('cache')->forget($cacheId);
+        if (auth()->user() && auth()->user()->isStaff()) {
+            app('cache')->forget($cacheId);
+            $message = $cacheId.' has been cleared from the cache';
+        } else {
+            $message = 'Hey, staff only please!';
+        }
 
-        return $redirectUrl ? redirect($redirectUrl) : back();
+        return redirect($redirectUrl)->with('flash_message', ['class' => 'messages', 'text' => $message]);
     }
 }
--- a/app/Http/Controllers/CacheController.php
+++ b/app/Http/Controllers/CacheController.php
@@ -2,8 +2,6 @@
 
 namespace App\Http\Controllers;
 
-use URL;
-use Auth;
 use Illuminate\Http\Request;
 
 class CacheController extends Controller
@@ -17,10 +15,10 @@ class CacheController extends Controller
      */
     public function __invoke(Request $request, $cacheId)
     {
-        $redirectUrl = $request->query('redirect') ?: URL::previous();
+        $redirectUrl = $request->query('redirect') ?: url()->previous();
 
         if (auth()->user() && auth()->user()->isStaff()) {
-            app('cache')->forget($cacheId);
+            cache()->forget($cacheId);
             $message = $cacheId.' has been cleared from the cache';
         } else {
             $message = 'Hey, staff only please!';
--- a/app/Http/Controllers/CacheController.php
+++ b/app/Http/Controllers/CacheController.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use Illuminate\Http\Request;
+
+class CacheController extends Controller
+{
+    /**
+     * Clear the cached data for specified cacheId.
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @param  string $cacheId
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function __invoke(Request $request, $cacheId)
+    {
+        $redirectUrl = $request->query('redirect');
+
+        app('cache')->forget($cacheId);
+
+        return $redirectUrl ? redirect($redirectUrl) : back();
+    }
+}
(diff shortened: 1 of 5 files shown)


app/Repositories/CampaignRepository.php
```diff
@@ -43,8 +43,12 @@ public function getAll()
      * @return stdClass
      * @throws \Illuminate\Database\Eloquent\ModelNotFoundException
      */
-    public function findBySlug($slug)
+    public function findBySlug($slug, $clearFromCache = false)
```
weerd: I don't think this should be a param passed to this function and instead the cache is cleared in the `refresh()`.

mendelB: Gotcha. Should we still be making the call to `findBySlug` to 'refresh' the cache, or leave that for when the `show` action is hit in the redirect?

weerd:
Ok, how about we try this:

```php
$router->get('next/cache/{cacheId}', 'CacheController');
```
The button that the admin would click would include the id of the cached campaign as `campaign_teens-for-jeans` or whatever.

The `CacheController` just cares about clearing the cache... so all you do is pass it a `cacheId`. For the time being we keep this controller simple and not even specify an action. Instead use the `__invoke()` method on the class to automatically invoke the default method:

```php
class CacheController extends Controller
{
  public function __invoke($id)
  {
    // 1. take the ID find the item in cache if it exists and clear it.
    // 2. redirect back to the campaign page causing the campaign to reload and recache
  }
}
```

I really think the `findBySlug()` should not be in charge of clearing the cache... it should just care about finding the content by the specified slug.

app/Repositories/CampaignRepository.php
weerd: Just commented in main [discussion area](https://github.com/DoSomething/phoenix-next/pull/488#issuecomment-342639412)

itsjoekent:
@weerd I'm fine with isolating the CacheController to just clearing cache, my suggestion in Slack was that we simply call `findBySlug` again so it repopulates the data, no changes necessary to the function

weerd:
Sure, but redirecting back to the Campaign would result in the normal function of how we deal with loading campaigns and cause it to get cached again right? So for example if we add this "clear cache" button for campaign leads to also clear static pages on the site, we just redirect back to the page they were on prior to clicking the button and causes that page to get cached again. It'll agnostic of whatever method was used to find the previously requested item... whether it was by some slug or maybe a contentful id.

itsjoekent:
@weerd did you see my edit? I'm in agreement!

weerd:
LOL. Yeah, but you mention needing to call `findBySlug` again and we don't really need to if we just redirect to the same page. :)

weerd:
oh, sorry! Github didn't refresh for me 😬  Just seeing it now!

mendelB:
`php artisan cach:clear --github_comments`

mendelB:
One possible issue:
I was finding that redirecting `back()` was strangely redirecting to Embed or Reportback routes. Could this be because of the calls being made from our front end?

mendelB:
That was the reason I went for an explicit redirect to `campaigns/{slug}` in the `refresh` action instead of using `back()`
Short term could we technically do the same in the `CacheController` or is there a better alternative?

weerd:
🤔  hokay, what if we pass a url param like `?redirect="/path/to/redirect/to` and use it if present otherwise use `back()`... just aiming to make it more versatile since we wanna get static pages in here soon as well and this will all apply to them too.

mendelB:
Nice yeah that sounds like a good idea!

mendelB:
Ok, added the new controller and action for caching. Kept it mega simple. 
Question about two things I'd added for the previous attempt at this:
- Do we want authentication for this action (e.g. staff only)?
- Do we want flash messages indicating success/failure?

weerd:
> Do we want authentication for this action (e.g. staff only)?
Yeah, I would think so and probs add it to the `CacheController` logic... ?

> Do we want flash messages indicating success/failure?
Yeah, defs. We want to let the admin user know cache was cleared. You can `redirect()` or head `back()` with the flash message :)

app/Http/Controllers/CacheController.php
```diff
@@ -0,0 +1,31 @@
...
+     * @param  \Illuminate\Http\Request $request
+     * @param  string $cacheId
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function __invoke(Request $request, $cacheId)
+    {
+        $redirectUrl = $request->query('redirect') ?: URL::previous();
+
+        if (auth()->user() && auth()->user()->isStaff()) {
+            app('cache')->forget($cacheId);
```
itsjoekent: should probably use the Cache facade here i think https://laravel.com/docs/5.5/facades#introduction

mendelB: Is that generally the better convention. I was just reading through some documentation and it seems to be the recommended route. By the same token should we use the `Auth` facade as well?

itsjoekent: I'm definitely not the guru on Laravel conventions which is why I wanted @weerd to weigh in

weerd: you rang? ☎️ 

👁  now :)

DFurnes: They're interchangeable, in terms of what happens behind the scenes. I tend to prefer helper functions like `auth()` or `cache()` to facades since there's less "magic" involved.

weerd: Yeah, we can use the facades... @DFurnes might prefer to weigh in but I'm fine either way. You do already bring in the `Auth` class up to but don't use it in favor of the `auth()` helper method. So defs decide on which way to go with that.

weerd: oh haha... github didn't update in time... @DFurnes DID weigh in!

mendelB:
OK, now using helpers throughout the CacheController 🎛

PR #493, opened 2017-11-09
Admin dashboard

mendelB:
### What does this PR do?
- Adds an AdminDashboard component containing -as of now- a 'Clear Cache' link to clear the current campaign from the cache.
- AdminDashboard is rendered from the `CampaignPage` is the users role is `admin`.

I imagine eventually we'll want to abstract away the clear cache button into its own component and make it more item agnostic (such as when we import the static pages and want to apply cache clearing for those).

### Checklist


resources/assets/components/Campaign/Campaign.js
```diff
@@ -4,12 +4,15 @@ import PropTypes from 'prop-types';
 import ModalSwitch from '../Modal';
 import { CampaignPageContainer, LandingPageContainer } from '../Page';
 import NotificationContainer from '../Notification';
+import AdminDashboardContainer from '../AdminDashboard';
 
 const Campaign = (props) => {
-  const { isAffiliated, useLandingPage } = props;
+  const { isAffiliated, useLandingPage, userRole, location } = props;
```
weerd: where is location coming from? it's a prop?

weerd:
> I imagine eventually we'll want to abstract away the clear cache button into its own component and make it more item agnostic (such as when we import the static pages and want to apply cache clearing for those).

Is there a reason we can't be doing this now? Grabbing the `cacheId` in the `AdminDashboardContainer` that's specific to campaigns seems like it's something we could just account for now and not be so specific...

resources/assets/components/Campaign/Campaign.js
mendelB: I believe that's coming from ReactRouter https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/location.md

mendelB:
Do you mean to instead pass it through from `Campaign` as a prop instead of hardcoding it to be `campaign_...` in the admin container?

resources/assets/components/Campaign/Campaign.js
itsjoekent: This isn't a huge deal to me, but I think it would be a lot simpler to just remove the React Router `location` usage & `redirectPath` entirely. Just use the Window location (`location.pathname`) inside the Admin Dashboard as I don't expect anyone would want to customize the redirect path here.

mendelB: That does feel way simpler

weerd:
For example, you call the `AdminDashboard` as:

```js
{ ! isAdmin ? 
null
:
<AdminDashboardContainer>
  <a className="button -secondary" href={`/next/cache/campaign_${slug}${location.pathname}`}>
    Clear Cache
  </a>
</AdminDashboardContainer>
}
```

This allows us to pass whatever `children` we want into the AdminDashboard apart from maybe having some default things in the admin dashboard as well ;)

mendelB:
iiinteresting. So the idea is a totally agnostic dashboard that we can render based on whatever logic we want, plus insert into it whichever children we want.

weerd:
Ya! 🎉

resources/assets/components/Campaign/Campaign.js
```diff
@@ -1,15 +1,26 @@
...
 const Campaign = (props) => {
-  const { isAffiliated, useLandingPage } = props;
+  const { isAffiliated, useLandingPage, isAdmin, slug } = props;
 
   return (
     <div>
+      { ! isAdmin ?
+        null :
+        <AdminDashboard>
+          <a className="button -secondary" href={`/next/cache/campaign_${slug}?redirect=${window.location.pathname}`}>
```
itsjoekent: Sorry I didn't add this yesterday but wasn't following the convo. This feels weird to me @weerd - not totally sure I understand the reasoning behind all of the changes.

- Moving the `isAdmin` check out of the `AdminDashboard`?
- Moving the clear cache button out of the admin dashboard? Accepting custom children??

I hear the need for one day having different types of admin dashboards, but I don't think we should make the admin dashboard more complicated for problems we don't have now. If we're truly set on passing children to the admin dashboard, I suggest we at least use the approach from the [modals](https://github.com/DoSomething/phoenix-next/tree/dev/resources/assets/components/Modal/configurations) so we keep the child components dry and re-usable.

Again though, this seems like a lot of work to future-proof the code and Mendels previous iteration seemed perfectly fine

weerd: It's not really more complicated. Prior to this we were grabbing the `cacheId` in the `AdminDashboardContainer` and assigning it specifically for just campaigns... so when we add this `AdminDashboard` to a page (near future) we'd have to rethink how it works anyways. This keeps the `AdminDashboard` only caring about rendering the dashboard and any _universal_ items we'd need, and allows passing custom children that are specific to a campaign or page, etc. The only thing that should care about the `cacheId` is the button that is going to call the endpoint and pass the ID... the `AdminDashboard` shouldn't care about that.

I _do_ like the idea of moving the `isAdmin` check inside the `AdminDashboard` so in components we just add the `AdminDashboard` without needing the logic in there as well since it'll happen in the component.

itsjoekent: alrighty - I can live with that

weerd: phew! lol 😂

mendelB:
Ok, here's how it looks as of now:
image

PR #494, opened 2017-11-09
ensuring we dont push blank reportbacks as blocks when insufficient reportbacks

mendelB:
### What does this PR do?
When we don't have enough reportbacks in state to fill the specified block offset amount, this will set the amount to fit with however many reportbacks we have.

We fetch a static amount of 24 reportbacksItems from the API, we then normalize the data to parse out unique reportbacks. For missing-history this ended up being only 7 unique reportbacks. 
The issue is that we'd be trying to fill the offset block amount of 15 with only 7 reportbacks, leading to a bunch of blank reportbacks on the page:
image

This PR will ensure that if we have less reportbacks available then the offset wants, instead of churning out blank reportbacks until we achieve that number, it will just set the number to be however many RB we have.

image

(This is a somewhat rushed attempt at a fix. The logic running the Blocks and RBs is fairly complicated so this took me a bit. Plese LMK if I'm very off base with this / if there's a better way to go about this. (one suggestion might be to either up the count of RB items we're requesting from Rogue to begin with which will probably prevent this from ever happening. Or to keep requesting more RBs until we have enough unique ones to fill the offset)) Anyways I'm rambling here so 🤐 

https://www.pivotaltracker.com/story/show/152723398

### Checklist

### DIFF
--- a/resources/assets/selectors/feed.js
+++ b/resources/assets/selectors/feed.js
@@ -80,9 +80,13 @@ export function getVisibleBlocks(state) {
     return totalPoints <= blockOffset;
   });
 
+  // In case we don't have enough reportbacks in our state to fill the
+  // target amount, we'll shave the target number to fit with what we
+  // actually have so we don't churn out blank RBs.
   if ((state.reportbacks.ids.length + totalPoints) < blockOffset) {
     blockOffset = state.reportbacks.ids.length + totalPoints;
   }
+
   // If we weren't able to fill enough rows with blocks, add
   // additional reportback blocks until we hit the target.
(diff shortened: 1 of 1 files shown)


resources/assets/selectors/feed.js
```diff
@@ -79,6 +80,9 @@ export function getVisibleBlocks(state) {
     return totalPoints <= blockOffset;
   });
 
+  if ((state.reportbacks.ids.length + totalPoints) < blockOffset) {
```
DFurnes: Let's add a comment explaining what this fixes for future reference.

PR #499, opened 2017-11-10
Affirmation trigger button on admin dashboard

mendelB:
### What does this PR do?
adds Show Affirmation button to the admin dashboard to enable admins to preview the post signup affirmation conveniently.

- Added a special action to the `signup` actions to make this work as the regular way is within the whole mess of clickedSignup logic and flow.
- Styling with some margins for the buttons as they were sticking together.
- Just realized that with multiple buttons this will get pretty ugly on smaller screens. Should we deal with this now or later?

### Checklist

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -159,9 +159,7 @@ export function clickedSignUp(campaignId, options = null, shouldRedirectToAction
 // Action: triggers the post signup affirmation modal.
 // This is for admin usage.
 export function clickedShowAffirmation() {
-  return dispatch => (
-    dispatch(openModal(POST_SIGNUP_MODAL))
-  );
+  return openModal(POST_SIGNUP_MODAL);
 }
 
 // Action: sends whether the user opted out of affiliate messaging.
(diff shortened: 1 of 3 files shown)


itsjoekent:
> Just realized that with multiple buttons this will get pretty ugly on smaller screens. Should we deal with this now or later?

We can maybe `overflow-x` it, I'm not that worried about it since campaign leads are not editing content on there phone

resources/assets/actions/signup.js
```diff
@@ -156,6 +156,14 @@ export function clickedSignUp(campaignId, options = null, shouldRedirectToAction
   };
 }
 
+// Action: triggers the post signup affirmation modal.
+// This is for admin usage.
+export function clickedShowAffirmation() {
+  return dispatch => (
+    dispatch(openModal(POST_SIGNUP_MODAL))
```
itsjoekent: This can just be, 

```
export function clickedShowAffirmation() {
   return openModal(POST_SIGNUP_MODAL);
}
```

right?

PR #501, opened 2017-11-13
Fixing campaign info not showing on the NS login page

itsjoekent:
### What does this PR do?
We were sending the campaign auth customization if you clicked "Log in" but not for users clicking signup buttons. This is because the login variables were only introduced in the php helpers & blade template.

This pr adds some logic so the client JS can send custom auth variables as well based on the current campaign to match the blade template logic.

https://www.pivotaltracker.com/story/show/152686373

### DIFF
--- a/resources/assets/actions/event.js
+++ b/resources/assets/actions/event.js
@@ -44,13 +44,21 @@ export function startQueue() {
 
 // Action: add an event to the queue.
 export function queueEvent(actionCreatorName, ...args) {
-  return {
-    type: actions.QUEUE_EVENT,
-    createdAt: Date.now(),
-    requiresAuth: true, // vLater - Allow more flexibility with configuring events
-    action: {
-      creatorName: actionCreatorName,
-      args,
-    },
+  return (dispatch, getState) => {
+    const northstarOptions = {};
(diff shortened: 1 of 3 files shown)


resources/assets/actions/event.js
```diff
@@ -44,13 +44,23 @@ export function startQueue() {
...
-    type: actions.QUEUE_EVENT,
-    createdAt: Date.now(),
-    requiresAuth: true, // vLater - Allow more flexibility with configuring events
-    action: {
-      creatorName: actionCreatorName,
-      args,
-    },
+  return (dispatch, getState) => {
+    const northstarOptions = {};
+    if (getState().campaign) {
```
weerd: Maybe set a `const campaign = getState().campaign` and then use the variable to grab the `title`, `coverImage.url` and `callToAction`?

itsjoekent: Good idea! 🌮

resources/assets/actions/event.js
```diff
@@ -44,13 +44,25 @@ export function startQueue() {
...
+      const { callToAction, coverImage, title } = getState().campaign;
```
weerd: Noice!

PR #502, opened 2017-11-13
removing collapsed footer

itsjoekent:
### What does this PR do?
Since we don't have the Forge JS stuffs running the footers can't be expanded on mobile. Rather than replicate the logic like we did for the Nav we just opt-ed to remove the collapsed classes on the elements themselves.

screen shot 2017-11-13 at 11 08 34 am
 
https://www.pivotaltracker.com/story/show/152720060


weerd:
@mendelB we'll eventually be switching to a new Forge so no need to include the old JS Forge library.

PR #503, opened 2017-11-13
Admin dash show pitch page

mendelB:
### What does this PR do?
- Adding button to admin dashboard to show the pitch page
- Button added only if the campaign has a landing page
- Converting the `Campaign` component to a `class`, so we can preserve and manipulate state

- Is it overkill to do all this for the admin landing page functionality? (I wasn't able to think of any better straightforward way)

### Checklist

### DIFF
--- a/resources/assets/components/Campaign/Campaign.js
+++ b/resources/assets/components/Campaign/Campaign.js
@@ -8,29 +8,33 @@ import { CampaignPageContainer, LandingPageContainer } from '../Page';
 import NotificationContainer from '../Notification';
 import AdminDashboardContainer from '../AdminDashboard';
 
-const Campaign = (props) => {
-  const { isAffiliated, useLandingPage, slug, clickedShowAffirmation } = props;
+class Campaign extends React.Component {
 
-  return (
-    <div>
-      <AdminDashboardContainer>
-        <a className="button -secondary margin-horizontal-md" href={`/next/cache/campaign_${slug}?redirect=${window.location.pathname}`}>
-          Clear Cache
-        </a>
-        <button className="button -secondary margin-horizontal-md" onClick={clickedShowAffirmation}>
-          Show Affirmation
-        </button>
-      </AdminDashboardContainer>
-      <NotificationContainer />
-      <ModalSwitch />
+  render() {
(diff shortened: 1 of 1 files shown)


resources/assets/components/Campaign/Campaign.js
```diff
@@ -8,29 +8,52 @@ import { CampaignPageContainer, LandingPageContainer } from '../Page';
...
-        <CampaignPageContainer {...props} />}
-    </div>
-  );
-};
+class Campaign extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      shouldShowLandingPage: false,
```
itsjoekent: Is there a reason we used the component state here instead of Redux?

mendelB: Not really... Is that the better route to take?

itsjoekent: Yeah we generally try to use Redux whenever possible. It doesn't have to be a merge blocker but we should make an `admin` part of the store where we keep track of this and any other forced states admins can apply.

mendelB: Gotcha. I can try and work this into using redux instead, it'll make adding an action page view button easier too.

DFurnes: This feels like a computed value to me, so ideally something we could just deal with in the render function (or if it's used in multiple places, using a selector function).

mendelB: @DFurnes How would this be dealt with in the render function? (Wouldn't a state-related thing need to be outside `render`?)

DFurnes: Oh, reading this more carefully now. Putting a new flag in the Redux store makes sense. 👍

I kinda wonder if it'd be useful to have a way for admins to flip their `isAffiliated` state at will (rather than adding a new specific override for showing/hiding the landing page), but that might add more complexity than it's worth.

mendelB: Yeah Joe mentioned something like that as well
https://dosomething.slack.com/archives/C3ASB4204/p1510601552000280

PR #506, opened 2017-11-13
Admin dash redux

mendelB:
### What does this PR do?
#503 only it's using the redux store to track admin override (which for now is just show landing page). 
Moved some calculation for `shouldShowLandingPage` to the container, as we don't really need the `isAffiliated` prop within the component itself and this makes it read a bit better I think.

Did I do this right?

#503 

### Checklist

### DIFF
--- a/resources/assets/actions/admin.js
+++ b/resources/assets/actions/admin.js
@@ -1,0 +1,11 @@
+import { SHOW_LANDING_PAGE } from '../actions';
+
+/**
+ * Action Creators: these functions create actions, which describe changes
+ * to the state tree (either as a result of application logic or user input).
+ */
+
+// Action: show the campaigns landing page
+export default function clickedShowLandingPage() {
+  return { type: SHOW_LANDING_PAGE };
+}
--- a/resources/assets/actions/admin.js
+++ b/resources/assets/actions/admin.js
@@ -6,6 +6,6 @@ import { SHOW_LANDING_PAGE } from '../actions';
  */
 
 // Action: show the campaigns landing page
-export default function clickedShowLandingPage() {
+export function clickedShowLandingPage() { // eslint-disable-line import/prefer-default-export
   return { type: SHOW_LANDING_PAGE };
 }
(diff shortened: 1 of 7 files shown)


resources/assets/actions/admin.js
```diff
@@ -0,0 +1,11 @@
+import { SHOW_LANDING_PAGE } from '../actions';
+
+/**
+ * Action Creators: these functions create actions, which describe changes
+ * to the state tree (either as a result of application logic or user input).
+ */
+
+// Action: show the campaigns landing page
+export default function clickedShowLandingPage() {
```
itsjoekent: I'd `export function clicked...` and add an eslint-ignore for the default export rule. otherwise the next dev has to update this & the index file to add another action

resources/assets/components/Campaign/CampaignContainer.js
```diff
@@ -2,15 +2,21 @@ import { connect } from 'react-redux';
...
-const mapStateToProps = state => ({
-  isAffiliated: state.signups.thisCampaign,
-  useLandingPage: state.campaign.landingPage !== null,
-  slug: state.campaign.slug,
-});
+const mapStateToProps = (state) => {
+  const shouldShowLandingPage = ! state.signups.thisCampaign || state.admin.shouldShowLandingPage;
+
+  return {
+    useLandingPage: state.campaign.landingPage !== null,
```
itsjoekent: Can we merge all of this logic into just 1 variable?

itsjoekent: eg:

```
const isSignedUp = state.signups.thisCampaign;
const forceLandingPage = state.admin.shouldShowLandingPage;
const hasLandingPage = state.campaign.landingPage !== null;

return {
   useLandingPage: forceLandingPage || (! isSignedUp && hasLandingPage),
   slug: state.campaign.slug,
}
```

mendelB: The reason I didn't is bc the `Show Landing Page` button is being shown only if `useLandingPage` is `true`, so that we don't show the button when there's no landing page available.

itsjoekent: Hm ok. Can we rename these then so it's a bit more clear? On the surface it's not clear what the difference is between `useLandingPage` and `shouldShowLandingPage`

mendelB: But I guess we can add that logic to `shouldShowLandingPage` and *still* make the `useLandingPage` prop available?

PR #507, opened 2017-11-13
fixing slugs

itsjoekent:
i will fill this out when i get a sec, probably

### DIFF
--- a/resources/assets/components/Navigation/TabbedNavigationContainer.js
+++ b/resources/assets/components/Navigation/TabbedNavigationContainer.js
@@ -11,6 +11,7 @@ import TabbedNavigation from './TabbedNavigation';
 import { campaignPaths } from '../../helpers/navigation';
 import { isCampaignClosed } from '../../helpers';
 import SignupButtonFactory from '../SignupButton';
+import { hasSlugFix } from '../../selectors/campaign';
 
 const mapStateToProps = state => ({
   hasActivityFeed: Boolean(state.campaign.activityFeed.length),
@@ -40,7 +41,12 @@ const TabbedNavigationContainer = (props) => {
   const additionalPages = pages
     .filter(page => ! page.fields.hideFromNavigation)
     .map((page) => {
-      const path = join('/us/campaigns', campaignSlug, campaignPaths.pages, page.fields.slug);
+      // TODO: Get rid of this slug fix conditional.
+      const path = hasSlugFix(page) ? (
+        join('/us/campaigns/', page.fields.slug)
+      ) : (
+        join('/us/campaigns', campaignSlug, campaignPaths.pages, page.fields.slug)
+      );
 
       return (
         <NavigationLink key={page.id} to={path}>{page.fields.title}</NavigationLink>
--- a/resources/assets/components/Navigation/TabbedNavigationContainer.js
+++ b/resources/assets/components/Navigation/TabbedNavigationContainer.js
@@ -40,15 +40,15 @@ const TabbedNavigationContainer = (props) => {
   const additionalPages = pages
     .filter(page => ! page.fields.hideFromNavigation)
     .map((page) => {
+      const baseCampaignsPath = '/us/campaigns';
       const pageHasCampaignSlug = page.fields.slug.indexOf(campaignSlug) >= 0;
       let pageSlug = page.fields.slug;
-      let path = '/us/campaigns';
 
       if (pageHasCampaignSlug) {
         pageSlug = pageSlug.replace(`${campaignSlug}/`, '');
       }
 
-      path += `/${campaignSlug}${campaignPaths.pages}${pageSlug}`;
+      const path = join(baseCampaignsPath, campaignSlug, campaignPaths.pages, pageSlug);
 
       return (
         <NavigationLink key={page.id} to={path}>{page.fields.title}</NavigationLink>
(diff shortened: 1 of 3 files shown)


resources/assets/components/Navigation/TabbedNavigationContainer.js
```diff
@@ -40,7 +41,12 @@ const TabbedNavigationContainer = (props) => {
   const additionalPages = pages
     .filter(page => ! page.fields.hideFromNavigation)
     .map((page) => {
-      const path = join('/us/campaigns', campaignSlug, campaignPaths.pages, page.fields.slug);
+      // TODO: Get rid of this slug fix conditional.
+      const path = hasSlugFix(page) ? (
+        join('/us/campaigns/', page.fields.slug)
+      ) : (
+        join('/us/campaigns', campaignSlug, campaignPaths.pages, page.fields.slug)
+      );
```
itsjoekent: @weerd so the process for updating this,

1. deploy this code to prod, no existing page slugs will break.
2. update every page in contentful to use the new slug format, include this `slugFix` flag. pages should continue working.
3. remove this `hasSlugFix` conditional from the codebase, assume all pages are using the new slug format

itsjoekent: the alternative you suggested, adding the flag to indicate backward compatibility would look like this,

1. update every page in contentful to have this `legacySlug` flag
2. deploy this code to prod 
3. update every page in contentful to use the new slug format and remove the flag
4. remove the flag from the code, assume all pages updated

its only 1 more step, but that additional step doubles the amount of work required

weerd:
Ok, so I think there's actually a simpler way to go about this an I think there's just a bit of confusion.

With the `scholarships` slug on `teens-for-jeans` campaign, what do you foresee the slug changing to in the contentful field and what will it look like as the actual path to the page?

itsjoekent:
Based on that gist I assumed the new slug would be `test-teens-for-jeans/pages/learn`

screen shot 2017-11-14 at 9 57 32 am

weerd:
See, there's the confusion... 

From the gist:
image

The campaign leads don't add `/page/` to the slug. Just `campaign_slug/subpage_slug`.

We programmatically add `/page/` when deciding that this will show up as a page (most common scenario) versus some other way (my only example at the moment is modals as an alternative, but there could be others).

So the code could be easily simplified and all it needs to do [here](https://github.com/DoSomething/phoenix-next/pull/507/files#diff-32e841fd05f866bd51df1076675f574dR44) is check if the string from the page slug field contains the full campaign slug (`teens-for-jeans`) as part of it (`scholarships` vs `teens-for-jeans/scholarships`). If it doesn't then prepend `teens-for-jeans/pages/` to the string, but if it _does_ then just insert `/pages/` in the string.

This is what we discussed as a team when we all met. I think it would simplify the logic and no need for any JSON additional content. And we could push to prod and slowly update the fields over time as needed but otherwise shouldn't break anything.

Does this make sense?

itsjoekent:
@weerd hows that?

itsjoekent:
er wait, one sec

itsjoekent:
ok, i think its working as you described. the slug in contentful is `test-teens-for-jeans/learn`

if i click the `learn` tab I get brought to `test-teens-for-jeans/pages/learn`

the other tabs (unmodified) all go to `test-teens-for-jeans/pages/*`

resources/assets/components/Navigation/TabbedNavigationContainer.js
```diff
@@ -40,7 +40,15 @@ const TabbedNavigationContainer = (props) => {
...
-      const path = join('/us/campaigns', campaignSlug, campaignPaths.pages, page.fields.slug);
+      const pageHasCampaignSlug = page.fields.slug.indexOf(campaignSlug) >= 0;
+      let pageSlug = page.fields.slug;
+      let path = '/us/campaigns';
+
+      if (pageHasCampaignSlug) {
+        pageSlug = pageSlug.replace(`${campaignSlug}/`, '');
+      }
+
+      path += `/${campaignSlug}${campaignPaths.pages}${pageSlug}`;
```
weerd: Looks much better!

One thing, can you switch to build the path using `join` like we do [here](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/Navigation/TabbedNavigationContainer.js#L43) @DFurnes has pointed out in the past that the `path.join()` method does a really good job of making sure it forms a proper url in case anything is missing from the string like `/` or whatever :)

resources/assets/components/Navigation/TabbedNavigationContainer.js
```diff
@@ -40,7 +40,15 @@ const TabbedNavigationContainer = (props) => {
...
-      const path = join('/us/campaigns', campaignSlug, campaignPaths.pages, page.fields.slug);
+      const baseCampaignsPath = '/us/campaigns';
+      const pageHasCampaignSlug = page.fields.slug.indexOf(campaignSlug) >= 0;
+      let pageSlug = page.fields.slug;
+
+      if (pageHasCampaignSlug) {
+        pageSlug = pageSlug.replace(`${campaignSlug}/`, '');
+      }
+
+      const path = join(baseCampaignsPath, campaignSlug, campaignPaths.pages, pageSlug);
```
weerd: Nitpicky, but you probs don't need the `baseCampaignsPath` var since you only use it once. Just toss the string in there!

PR #508, opened 2017-11-13
Admin dash action page button action

mendelB:
### What does this PR do?
- Adds show action page button to admin dashboard to display the action page.

action bronson

- Adding new prop to the admin store to account for 'show action page' override.
 - The reducer will need to toggle both states to make sure to turn off one override before turning on another.

Should we make the landing page button a `Link` as well so that the url is changed back to root in case an admin clicks the 'Show Action Page' button thus changing the url to `/action` and then clicks the `Show Landing Page` button - which for now is a plain button so the url remains `/action`?

https://www.pivotaltracker.com/story/show/152202725

### Checklist

### DIFF
--- a/resources/assets/components/Page/CampaignPage/CampaignPage.js
+++ b/resources/assets/components/Page/CampaignPage/CampaignPage.js
@@ -24,7 +24,7 @@ const CampaignPage = (props) => {
   const {
     affiliatePartners, affiliateSponsors, blurb, campaignLead, coverImage,
     dashboard, endDate, hasActivityFeed, isAffiliated, legacyCampaignId, match,
-    openModal, slug, subtitle, template, title, totalCampaignSignups,
+    openModal, shouldShowActionPage, slug, subtitle, template, title, totalCampaignSignups,
   } = props;
 
   // console.log(props.history);
@@ -72,7 +72,7 @@ const CampaignPage = (props) => {
             />
             <Route
               path={`${match.url}/action`}
-              render={() => (isClosed ?
+              render={() => (isClosed && ! shouldShowActionPage ?
                 <Redirect to={`${match.url}`} />
                 :
                 <ActionPageContainer />
@@ -144,6 +144,7 @@ CampaignPage.propTypes = {
   title: PropTypes.string.isRequired,
   totalCampaignSignups: PropTypes.number,
   openModal: PropTypes.func.isRequired,
+  shouldShowActionPage: PropTypes.bool,
 };
 
 CampaignPage.defaultProps = {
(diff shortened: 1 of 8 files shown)


resources/assets/components/Page/CampaignPage/CampaignPage.js
```diff
@@ -72,7 +72,7 @@ const CampaignPage = (props) => {
             />
             <Route
               path={`${match.url}/action`}
-              render={() => (isClosed ?
+              render={() => (isClosed && ! shouldShowActionPage ?
```
itsjoekent: I think this should probably be `(isClosed || ! shouldShowActionPage)` since this is meant to override the normal experience rules

resources/assets/components/Page/CampaignPage/CampaignPage.js
```diff
@@ -144,6 +144,7 @@ CampaignPage.propTypes = {
   title: PropTypes.string.isRequired,
   totalCampaignSignups: PropTypes.number,
   openModal: PropTypes.func.isRequired,
+  shouldShowActionPage: PropTypes.bool,
```
itsjoekent: don't worry about this for now, but in this case you could make `shouldShowActionPage` required and remove the need for a default prop added

resources/assets/components/Page/CampaignPage/CampaignPage.js
mendelB: But then if the campaign is indeed closed, wouldn't it redirect away from the action page regardless of the admin clicking 'Show Action Page' ?

itsjoekent: Is there another place we're redirecting from the action page?

mendelB: What I mean is it wouldn't render the action page to begin with

itsjoekent: Yes but I thought the point of the admin buttons was to override the normal experience rules?

mendelB: Right which is what it's doing in this case: ordinarily if the campaign is closed and the route requested is `.../action` it'll redirect to root without showing the action page. 

The `&& ! shouldShowAction` just adds a second clause  to not redirect if `shouldShowActionPage` is `true`

(unless I'm thoroughly confused. These reverse conditionals always throw me off)

itsjoekent: ohhhhhhhhhhhhhhh.


yep you're right

PR #509, opened 2017-11-14
set to use contentful preview api on thor/staging

mendelB:
### What does this PR do?
- Sets the Contentful API Client to use the Contentful preview API when on thor/staging environment

Questions:
 - ~Is using the `config` helper within `config` itself the right thing to do?~
 - ~Should the `$shouldUsePreview` be determined by some new environment key such as `USE_CONTENTFUL_PREVIEW=true/false` as opposed to the way it's done here?~ Using a .env key to set to use preview API or not.
 - Will Thor and the rest of the gang be able to defeat Thanos?

> We can use the preview api so we aren't publishing things that shouldnt be shipped yet!

jeff

https://www.pivotaltracker.com/story/show/152546533

### Checklist


config/contentful.php
```diff
@@ -3,6 +3,8 @@
  * @copyright 2016 Contentful GmbH
  * @license   MIT
  */
+$shouldUsePreview = config('app.env') === 'staging';
```
DFurnes: I think we should probably use `env('APP_ENV')` here since we're in a config file... I feel like otherwise we could end up in a weird loop when trying to cache config values. It could also make sense to have a separate `CONTENTFUL_USE_PREVIEW_API` (or w/e) toggle so we can explicitly set and unset this.

mendelB: What about the third question though?!

config/contentful.php
```diff
@@ -13,12 +15,12 @@
...
      * An API key for the above specified space
      */
-    'delivery.token' => env('CONTENTFUL_CONTENT_DELIVERY_API_KEY'),
+    'delivery.token' => env('CONTENTFUL_CONTENT_'.$tokenType.'_API_KEY'),
 
     /*
      * Controls whether Contentful's Delivery or Preview API is accessed
      */
-    'delivery.preview' => false,
+    'delivery.preview' => $shouldUsePreview,
```
weerd: Do we know what happens if we set `delivery.token` to the preview api key but `delivery.preview` to `false` what happens?

weerd: We should just better utilize env vars for this instead of adding logic:

```php
return [
  'delivery.token' => env('CONTENTFUL_CONTENT_API_KEY'),

  'delivery.token' => env('CONTENTFUL_USE_PREVIEW_API'),
]
```

```
// env file:

CONTENTFUL_CONTENT_API_KEY=asdasdasdadd123123asdad // key for preview or delivery depending on env
CONTENTFUL_USE_PREVIEW_API=true  // if using preview api key

// We can choose to keep the following vars for reference so we know which is being used in CONTENTFUL_CONTENT_API_KEY but otherwise these just don't get used:
CONTENTFUL_CONTENT_DELIVERY_API_KEY
CONTENTFUL_CONTENT_PREVIEW_API_KEY
```

config/contentful.php
DFurnes: Oh man haha, I actually didn't notice those questions were in the PR. ☺️

DFurnes: And the answer is "no" because Thanos is too powerful!!

config/contentful.php
DFurnes: I feel like both have an equal chance for user error, but I'm fine with this change. 👍

mendelB: >Do we know what happens if we set delivery.token to the preview api key but delivery.preview to false what happens?

It will error out due to a wrong API key (since the preview key doesn't work for the regular API and vice versa).

config/contentful.php
mendelB: tsk tsk where is your faith in our heroes!

config/contentful.php
weerd: Ok, cool. Good to know. Yeah, just create another env var for that and we set it there

.env.example
```diff
@@ -16,6 +16,8 @@ CACHE_DRIVER=redis
 SESSION_DRIVER=file
 QUEUE_DRIVER=sync
 
+CONTENTFUL_CONTENT_API_KEY
```
weerd: add an `=` sign at the end!

mendelB: No! I want things to break

PR #513, opened 2017-11-15
Adding initial posts routes and controllers for API routes.

weerd:
### What does this PR do?
This PR begins adding some of the Phoenix API endpoints that connect with Rogue to get data on campaign posts. This is the start of the work that will eventually allow us to split off the PN front-end from the server and allow the front-end app to get data via API endpoints.

Started also adding some basic info for endpoint documentation.

Refs https://www.pivotaltracker.com/story/show/151696124

### Checklist

### DIFF
--- a/app/Http/Controllers/Api/CampaignPostsController.php
+++ b/app/Http/Controllers/Api/CampaignPostsController.php
@@ -10,25 +10,30 @@ use App\Repositories\PostRepository;
 class CampaignPostsController extends Controller
 {
     /**
-     * [$rogue description]
-     * @var [type]
+     * The post repository.
+     *
+     * @var PostRepository
      */
     private $postRepository;
 
     /**
-     * [$rogue description]
-     * @var [type]
+     * Create a new CampaignPostsController instance.
+     *
+     * @var PostRepository $postRepository
      */
     public function __construct(PostRepository $postRepository)
     {
-        $this->middleware('auth:api');
+        $this->middleware('auth:api')->except(['index']);
 
         $this->postRepository = $postRepository;
     }
 
     /**
-     * [$rogue description]
-     * @var [type]
+     * Display a listing of the resource.
+     *
+     * @param  string  $id
+     * @param  Request $request
+     * @return array - JSON response
      */
     public function index($id, Request $request)
     {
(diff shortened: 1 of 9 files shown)


app/Http/Controllers/Api/CampaignPostsController.php
```diff
@@ -0,0 +1,55 @@
...
+    private $postRepository;
+
+    /**
+     * Create a new CampaignPostsController instance.
+     *
+     * @var PostRepository $postRepository
+     */
+    public function __construct(PostRepository $postRepository)
+    {
+        $this->middleware('auth:api')->except(['index']);
```
DFurnes: Oh, I really like this syntax. 👍

app/Http/Controllers/Api/CampaignPostsController.php
```diff
@@ -0,0 +1,55 @@
...
+
+        $this->postRepository = $postRepository;
+    }
+
+    /**
+     * Display a listing of the resource.
+     *
+     * @param  string  $id
+     * @param  Request $request
+     * @return array - JSON response
```
DFurnes: Let's either return an actual array here, or change this to `\Illuminate\Http\JsonResponse`.

app/Http/Controllers/Api/CampaignPostsController.php
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Store a newly created resource.
+     *
+     * @param  string $id
+     * @return object - JSON response
+     */
+    public function store($id)
+    {
+        // @TODO inactive for now.
+        return response()->json(false);
```
DFurnes: It could make sense to `abort(501);` for unfinished routes (for a "501 Not Implemented" response).

app/Http/Controllers/Api/CampaignPostsController.php
weerd: Yeah, seems like most of the newer syntax likes to method chain all the things ;)

app/Http/Controllers/Api/CampaignPostsController.php
weerd: ah, jes. I was going off of what we had in the `PhoenixLegacy` controller docblocks 😬

app/Http/Controllers/Api/CampaignPostsController.php
weerd: oh, nice. ok!

PR #515, opened 2017-11-15
adding twitter card meta tag for a large summary image

mendelB:
### What does this PR do?
Adds a meta tag for a large summary Twitter card so twitter shares can look Magnifique.
- also added the new Contentful API key name we're using to the `app.json`, for the Heroku review app. And removed the old ones no longer in use.

Boiteux:
image

Magnifique:
image

The other meta stuff should be read through our `og` (Original Gangster??) meta props.

https://www.pivotaltracker.com/story/show/152826443

### Checklist

### DIFF
--- a/app.json
+++ b/app.json
@@ -12,12 +12,6 @@
     "CONTENTFUL_CONTENT_API_KEY": {
       "required": true
     },
-    "CONTENTFUL_CONTENT_DELIVERY_API_KEY": {
-      "required": true
-    },
-    "CONTENTFUL_CONTENT_PREVIEW_API_KEY": {
-      "required": true
-    },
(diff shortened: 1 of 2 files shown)


weerd:
@mendelB @itsjoekent do we need this env item too? `CONTENTFUL_USE_PREVIEW_API`

mendelB:
It defaults to `null` so not using preview. So I guess it's technically not a *needed* item when we're not previewing unless you think it should always be present anyway?

weerd:
This is for the branch deploys right? We probs wanna pull from the Preview API for these right?

mendelB:
Sure I guess why not indeed? I'll throw in a new PR with those keys set

PR #517, opened 2017-11-16
Encode auth options before sending

itsjoekent:
### What does this PR do?
In a previous PR we added support for the client to pass campaign information along with the auth request (#501). This worked fine with Sincerely, Us, but we later discovered it doesn't work for Missing in History.

The reason it was breaking on Missing in History is that the callToAction uses the `&` symbol, which splits the json into another query parameter and prevents the server from parsing it. To resolve this, I just wrapped the potentially unsafe values with `encodeURIComponent`.

screen shot 2017-11-16 at 10 35 45 am

https://www.pivotaltracker.com/story/show/152686373


itsjoekent:
thanks!

PR #519, opened 2017-11-16
Updating app json

itsjoekent:
### What does this PR do?
- Added keys that were missing
- Alphabetizing everything
- Adding the Redis Addon

Getting Heroku staging up to date with everything

### DIFF
--- a/app.json
+++ b/app.json
@@ -2,17 +2,34 @@
   "name": "phoenix-next",
   "description": ":sparkles:",
   "env": {
-    "APP_ENV": "local",
     "APP_DEBUG": "false",
-    "NPM_CONFIG_PRODUCTION": "false",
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -2,19 +2,36 @@
   "name": "phoenix-next",
```
weerd: does the name need to change?

itsjoekent: 👍

itsjoekent:
also just for anyone casually following -- I mentioned to Diego IRL we could add environment specific stuff here if we want, but right now we're only relying on the app.json for the review apps so not a huge deal

itsjoekent:
screen shot 2017-11-16 at 1 56 26 pm

seems like it worked, from the review app on this PR ^

PR #521, opened 2017-11-17
Admin dash staff auth

mendelB:
### What does this PR do?
- Adds new selector to help determine user role.
- Using selector for `AdminDashboard` allowing 'staff' roles access as well as 'admin'.

Most staff could not see the admin dashboard!

### DIFF
--- a/resources/assets/components/AdminDashboard/AdminDashboardContainer.js
+++ b/resources/assets/components/AdminDashboard/AdminDashboardContainer.js
@@ -3,7 +3,7 @@ import AdminDashboard from './AdminDashboard';
 import { userHasRole } from '../../selectors/user';
 
 const mapStateToProps = state => ({
-  isAdmin: userHasRole(state, ['admin', 'staff']),
+  enabled: userHasRole(state, ['admin', 'staff']),
 });
 
 export default connect(mapStateToProps)(AdminDashboard);
--- a/resources/assets/components/AdminDashboard/AdminDashboardContainer.js
+++ b/resources/assets/components/AdminDashboard/AdminDashboardContainer.js
@@ -1,8 +1,9 @@
 import { connect } from 'react-redux';
 import AdminDashboard from './AdminDashboard';
+import { userHasRole } from '../../selectors/user';
 
 const mapStateToProps = state => ({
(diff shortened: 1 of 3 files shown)


resources/assets/components/AdminDashboard/AdminDashboardContainer.js
```diff
@@ -1,8 +1,9 @@
 import { connect } from 'react-redux';
 import AdminDashboard from './AdminDashboard';
+import { userHasRole } from '../../selectors/user';
 
 const mapStateToProps = state => ({
-  isAdmin: state.user.role === 'admin',
+  isAdmin: userHasRole(state, ['admin', 'staff']),
```
weerd: debating if calling this props `isAdmin` still makes sense, but not really a big deal ;)

DFurnes: Oh, agreed. Maybe rename to `enabled`?

resources/assets/selectors/user.js
```diff
@@ -7,3 +7,14 @@
...
+
+/**
+ * Find out if the user had one of provided roles.
+ *
+ * @param {object} state
+ * @param {array}  roles
+ * @returns {boolean}
+ */
+export function userHasRole(state, roles) {
+  return roles.includes(state.user.role);
```
DFurnes: This is so clean. 👏

resources/assets/components/AdminDashboard/AdminDashboardContainer.js
mendelB: Yeah that does make sense!

resources/assets/components/AdminDashboard/AdminDashboardContainer.js
```diff
@@ -1,8 +1,9 @@
 import { connect } from 'react-redux';
 import AdminDashboard from './AdminDashboard';
+import { userHasRole } from '../../selectors/user';
 
 const mapStateToProps = state => ({
-  isAdmin: state.user.role === 'admin',
+  enabled: userHasRole(state, ['admin', 'staff']),
```
weerd: 👍

PR #522, opened 2017-11-17
Social share modal link

mendelB:
### What does this PR do?
- adding option to the `makeShareLink` function to allow for either `modal` or `blocks` linking
- adding prop to `CampaignUpdate` component so that social links link out to modal and the heading link links to the block. 

I wasn't sure about how to handle this new 'modal/blocks' variable, so ended up going with a new `type` (not completely sold on that name either) option for the function. It does feel somewhat weird to be calling the `makeShareLink` twice in the container, but hopefully using the same `options` object twice dries it up somewhat. 
Also not completely sure about the `titleLink` prop name.

https://www.pivotaltracker.com/story/show/152351435
#481 

### Checklist

### DIFF
--- a/resources/assets/helpers/index.js
+++ b/resources/assets/helpers/index.js
@@ -228,7 +228,7 @@ export function makeHash(string) {
  */
 export function makeShareLink(
   resource,
-  options: { domain: string, slug?: string, key: string, type: "blocks" | "modals" },
+  options: { domain: string, slug?: string, key: string, type: "blocks" | "modal" },
 ) {
   switch (resource) {
     case 'campaigns':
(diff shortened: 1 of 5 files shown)


resources/assets/helpers/index.js
```diff
@@ -226,10 +226,13 @@ export function makeHash(string) {
  * @return {String}
  * @flow
  */
-export function makeShareLink(type, options: { domain: string, slug?: string, key: string }) {
+export function makeShareLink(
+  type,
+  options: { domain: string, slug?: string, key: string, type: string },
```
DFurnes: Let's enforce this as `blocks` or `modals` in the type:

```js
options: { domain: string, slug: string, key: string, type: "blocks" | "modals" },
```

DFurnes: Also maybe we can rename the first parameter to `resource`? (Or rename the second one to `subtype`?)

PR #523, opened 2017-11-17
Moving the hide step number out of json

itsjoekent:
### What does this PR do?
- [Adding Campaign Action Step Entity](https://github.com/DoSomething/phoenix-next/commit/54a516eecb64d3768319044f7dd633117c3b55eb). Just general housekeeping here by moving the code into a new entity. Of course, also added the `hideStepNumber` as a field on the action step, and it accepts both the new field and legacy json field.
- [Updating helper to be object and array friendly](https://github.com/DoSomething/phoenix-next/commit/d4aced2e08cb94ea658f13cc5bc7f8b17ad23ca9). This function is being used by our open graph logic already, and I figured with a tad repurposing it would also be useful here as well.
- [Updating action step wrapper for new field](https://github.com/DoSomething/phoenix-next/commit/085f131e92290a1744b50e0297f03a8f543dc030). JS broke since everything is under the `fields` object now. Also had to update where it referenced the `hideStepNumber`.


app/helpers.php
```diff
@@ -213,27 +212,36 @@ function get_legacy_campaign_data($id, $key = null)
...
  * @return mixed
  */
-function useOverrideIfSet($field, $campaign, $override)
+function useOverrideIfSet($field, $base, $override)
 {
-    if (! isset($campaign->{$field})) {
-        $base = null;
-    } else {
-        $base = $campaign->{$field};
+    $isBaseArray = is_array($base);
```
mendelB: Would casting to an object be useful here to avoid the rest of the conditionals for array/object?

itsjoekent: Yeah, I tried that first, the Contentful Dynamic Entry wasn't a fan (or vice-versa, going array). It seems to do a lot of hackery under the hood which makes it weird to work with, eg: here is a Slack message I sent Dave earlier...

screen shot 2017-11-17 at 3 19 45 pm

mendelB: Now that is some creepy stuff right there

DFurnes: I think Laravel's [`data_get`](https://laravel.com/docs/5.5/helpers#method-data-get) helper might let you avoid some of this extra logic.

itsjoekent:
- [Updated helper function to use Dave's suggestion of data_get](https://github.com/DoSomething/phoenix-next/pull/523/commits/adce5643844aa78e30d8dabce3be0dfff32d8549)

PR #524, opened 2017-11-20
Markdown External Links in New Tab

mendelB:
### What does this PR do?
- adds new [`markdown-it-attrs`](https://www.npmjs.com/package/markdown-it-attrs) lib to allow us to add attributes to the `<a>` tag generated by markdown links.
- parsing markdown content to find external links, and adding a `target="blank"` attribute to open in new tab

External links opening in the same tab is bad. However, we can't just inject HTML `<a>` tags with `target="blank"` wherever an external link is because we've decided to block HTML injection in the markdown conversion.

Adding this new plugin will allow for adding curly braces next to links and other markdown elements to give attributes to the generated HTML tag. 

I went the route of us manually adding these attributes to the links ourselves as opposed to having admins manually perform this in Contentful. My thinking was to avoid more info and labor on that end, and also avoid necessarily 'announcing' this new attribute-adding feature.
 

I couldn't find any regex around to serve our specific case so I had to conjure up something myself:
https://regex101.com/r/kaX9Kd/6

udpj8pi

https://www.pivotaltracker.com/story/show/148973233

### Checklist

### DIFF
--- a/resources/assets/components/Markdown/index.js
+++ b/resources/assets/components/Markdown/index.js
@@ -10,7 +10,7 @@ const contentfulImageFormat = url => (contentfulImageUrl(url, '1000'));
 const formatImageUrls = string => (string.replace(pattern, contentfulImageFormat));
 
 // https://regex101.com/r/kaX9Kd/6 (regex explanation)
-const externalUrlPattern = /\[.+?(?=])]\(http(s)?:\/\/(?!(www\.)?(next\.)?dosomething\.org).+(?=\))\)/g;
+const externalUrlPattern = /\[.+?(?=])]\(http(s)?:\/\/(?!(www\.)?(next\.)?dosomething\.org).+?(?=\))\)/g;
 const addAttr = link => (`${link}{target="_blank"}`);
 const formatExternalLinks = string => (string.replace(externalUrlPattern, addAttr));
 
--- a/resources/assets/components/Markdown/index.js
+++ b/resources/assets/components/Markdown/index.js
@@ -9,6 +9,10 @@ const pattern = /\/\/images\.contentful\.com.+\.(jpg|png)/g;
 const contentfulImageFormat = url => (contentfulImageUrl(url, '1000'));
 const formatImageUrls = string => (string.replace(pattern, contentfulImageFormat));
 
+// https://regex101.com/r/kaX9Kd/6 (regex explanation)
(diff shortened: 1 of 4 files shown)


resources/assets/components/Markdown/index.js
```diff
@@ -9,8 +9,20 @@ const pattern = /\/\/images\.contentful\.com.+\.(jpg|png)/g;
 const contentfulImageFormat = url => (contentfulImageUrl(url, '1000'));
 const formatImageUrls = string => (string.replace(pattern, contentfulImageFormat));
 
+// https://regex101.com/r/kaX9Kd/8 (regex explanation)
+const externalUrlPattern = /\[.+?(?=])]\(http(s)?:\/\/(?!(www\.)?(next\.)?dosomething\.org).+?(?=\))\)/g;
```
DFurnes: Instead of modifying the Markdown source, let's try using the [markdown-for-inline](https://github.com/markdown-it/markdown-it-for-inline) plugin to override this in the renderer:

```js
var iterator = require('markdown-it-for-inline');

const isExternal = url => new URL(url).hostname !== window.location.hostname;

markdownIt.use(iterator, 'url_new_win', 'link_open', function (tokens, idx) {
  const aIndex = tokens[idx].attrIndex('target');
  const url = tokens[idx].attrs['href'];

  if (aIndex < 0 && isExternalUrl(url)) {
    tokens[idx].attrPush(['target', '_blank']);
  } else {
    tokens[idx].attrs[aIndex][1] = '_blank';
  }
});
```

mendelB: whoa yeah, this'd be way cleaner!
What's the reason for: 
```
else {
  tokens[idx].attrs[aIndex][1] = '_blank';
 }
```

DFurnes: Was basing that off an example in the MarkdownIt docs... I think that conditional handles cases where a `target` attribute already exists on the link (like from another plugin). I don't think that's likely so it should be okay to replace it with this:

```js
markdownIt.use(iterator, 'url_new_win', 'link_open', function (tokens, index) {
  const url = tokens[index].attrs['href'];

  if (isExternalUrl(url)) {
    tokens[index].attrPush(['target', '_blank']);
  }
});
```

Let's also replace `idx` with `index` while we're at it!

mendelB: Got a new PR comin right up! (Had to change some of the indexing around to make it work - the attributes were being returned in arrays as opposed to objects)

PR #526, opened 2017-11-20
Reportback form reset

mendelB:
NOTE: Using #533 instead (same base-work, with some extra tracking logic).

### What does this PR do?
Only resetting the reportback form if the submission item from the store has a success object (meaning the submission was successful).

A quick way to avoid user frustration until we take more steps to refactor this component.

🐦 🐦 :one: 🗿 
https://www.pivotaltracker.com/story/show/152924905
https://www.pivotaltracker.com/story/show/152924977

### Checklist

### DIFF
--- a/resources/assets/components/ReportbackUploader/ReportbackUploader.js
+++ b/resources/assets/components/ReportbackUploader/ReportbackUploader.js
@@ -69,14 +69,13 @@ class ReportbackUploader extends React.Component {
     this.props.submitReportback(
       ReportbackUploader.setFormData(reportback),
       this.props.reportbackAffirmation,
-    );
-
-    // @TODO: only reset form AFTER successful RB submission.
-    // We'll make this a lot better once we switch to storing all the state
-    // in the Redux store @_@
-    this.form.reset();
-    this.setState({
-      media: this.defaultMediaState,
+    ).then(() => {
+      if (this.props.submissions.messaging.success) {
+        this.form.reset();
+        this.setState({
+          media: this.defaultMediaState,
+        });
+      }
     });
   }
 


resources/assets/components/ReportbackUploader/ReportbackUploader.js
```diff
@@ -69,14 +69,13 @@ class ReportbackUploader extends React.Component {
...
-    );
-
-    // @TODO: only reset form AFTER successful RB submission.
-    // We'll make this a lot better once we switch to storing all the state
-    // in the Redux store @_@
-    this.form.reset();
-    this.setState({
-      media: this.defaultMediaState,
+    ).then(() => {
+      if (this.props.submissions.messaging.success) {
```
itsjoekent: I noticed [here](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/ReportbackUploader/ReportbackUploader.js#L97) its possible the `submissions.messaging` var is undefined, but im assuming after any submission it would be defined & thus safe to access?

itsjoekent:
Actually, looks like you replicated this PR in the data tracking one, https://github.com/DoSomething/phoenix-next/pull/533/files#diff-201edebec1634a925cb7e8371eba4809R76

Mind closing this one & just merging the other?

resources/assets/components/ReportbackUploader/ReportbackUploader.js
mendelB: Yeah, that was my retionale. Is it worth using a lodash `get` to be safe?

itsjoekent: You could if you wanna be extra safe but I think you're fine

PR #535, opened 2017-11-27
Feeeed

itsjoekent:
### What does this PR do?
this pr ensures that any gaps between blocks is filled with a reportback photo. it might be a bit confusing so i'm gonna really break it down at the micro level, but first a macro level

before:
```
[  ] [  ]
[            ]
```

after:
```
[  ] [  ] [  ]
[            ]
```

i know i know, hold your applause.

1. renamed a bunch of functions. we were confusing two concepts (`points` and `blocks`, explained in the #context part of the PR) which made the code harder to update because the functions were referencing values that didn't represent the function name. [Feed selectors](https://github.com/DoSomething/phoenix-next/commit/fffb8c4623a6717404c7703c5d5d97549a0d0582), [Reportback selectors](https://github.com/DoSomething/phoenix-next/commit/6b28a65bb3a58046ef85c38533246a9a4dff7033)

2. prep work for the actual feature. first I decided to ditch the `.filter()` usage, because I was going to have to append blocks to the array and the `.filter()` function only accepts 1 value returned per callback. [new while loop](https://github.com/DoSomething/phoenix-next/commit/5bd7eb0b3e023007b0ba794c8aca84092bf72a56). after that, I moved the dummy reportback logic into a new inline function because as you can see from the comment stub, it's going to be re-used. [new rb generator](https://github.com/DoSomething/phoenix-next/commit/b68af1b97ae7f187a6cd4430534c77b16723f1f6)

3. the block fill logic. [there is a lot going on in the commit](https://github.com/DoSomething/phoenix-next/commit/a2de812e641e525e3720c453bf91a69991c4caeb), so let me break it up a bit. first of all, I ended up getting rid of the following logic,

```
if ((totalReportbackPoints + totalPoints) < targetPoints) {
   targetPoints = totalReportbackPoints + totalPoints;
```

it didn't actually seem to do anything and was just confusing things further. additionally by removing that logic I got to delete the reportback selector I created earlier and another variable.

second of all, the real meat on the bone

screen shot 2017-11-27 at 3 54 11 pm

basically what's going on here is we're checking how many points are left in the current row, and if the next block can fit on the row based on its point size. if it can't fit, that means we have a gap, and we should figure out how many reportbacks are needed to fill the gap. once we know that, we can do a quick loop to generate those blocks.

so for example, if we have a campaign update with a `two-thirds` display option, and the next block is another campaign update with `two-thirds`, we will have a remainder of 1 point in the first row. that 1 point will be filled with a reportback.

here is a more extreme example that has a 1 point block followed by a 3 point block (so there are 2 filler Reportbacks on the first row)

screen shot 2017-11-27 at 4 11 32 pm

OKAY, so, a lot of this PR revolves around renaming the points & blocks stuff. let me explain why i spent so much time fixing that naming.

the community feed is a grid, and the grid size is based on the constants defined in the feed selector. so as defined, it allows "3 blocks" per row, and 5 rows per page (assume a new page is made everytime you click the view more button). but there lies the problem, 3 blocks per row isn't an accurate description, because a single report back block might occupy 3 distinct grid spaces in the row. so thats where points system come in.

points define how much space the block occupies in the grid. and if you add up all of the points being used, you can figure out how much of the grid is empty (OR FULL?). this way, 3 distinct reportback blocks with a display option of `one-third`, and a single reportback block with a display option of `full` all equal the same number of points (3).

so now, if you look at some of my function renamings, it might make more sense why I did them.

eg: `totalBlocksInFeed` renamed to `totalBlockPointsInFeed`, because it's actually returning the total points of all of the blocks, not the total number of blocks.

https://www.pivotaltracker.com/story/show/152767131

### DIFF
--- a/resources/assets/actions/feed.js
+++ b/resources/assets/actions/feed.js
@@ -1,5 +1,9 @@
 import { FEED_INCREMENT_PAGE, fetchReportbacks } from '../actions';
-import { totalBlocksInFeed, totalReportbackBlocksInFeed, getBlockOffset } from '../selectors/feed';
+import {
+  totalBlockPointsInFeed,
+  totalReportbackBlockPointsInFeed,
+  getTotalVisibleBlockPoints,
+} from '../selectors/feed';
 
 /**
  * Action Creators: these functions create actions, which describe changes
@@ -18,8 +22,11 @@ export function conditionallyFetchReportbacks() {
 
     // The number of needed reportbacks are the number of reportback blocks in the
     // community feed + any additional offset that hasn't been filled by blocks.
-    const overflowReportbackBlocks = Math.max(0, getBlockOffset(state) - totalBlocksInFeed(state));
-    const neededReportbacks = totalReportbackBlocksInFeed(state) + overflowReportbackBlocks;
+    const overflowReportbackBlocks = Math.max(0, (
+      getTotalVisibleBlockPoints(state) - totalBlockPointsInFeed(state)
+    ));
+
+    const neededReportbacks = totalReportbackBlockPointsInFeed(state) + overflowReportbackBlocks;
 
     // Dispatch an HTTP request if we don't have enough reportbacks in the store.
     if (state.reportbacks.ids.length < neededReportbacks) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
       "integrity": "sha512-SBVzvCLvqA8/PBiyh2/leWOBtc0p5UU8iRzUIRuyiZwD4S3iMrf7gE5p6WutNdkhE0Jxq0dhQWoDjARvNArB9Q==",
       "dev": true,
       "requires": {
-        "eslint": "4.11.0",
+        "eslint": "4.12.0",
         "eslint-config-airbnb": "15.1.0",
         "eslint-plugin-import": "2.8.0",
         "eslint-plugin-jsx-a11y": "5.1.1",
@@ -77,7 +77,7 @@
         "file-loader": "1.1.5",
         "lodash": "4.17.4",
         "node-sass": "4.7.2",
-        "postcss-loader": "2.0.8",
+        "postcss-loader": "2.0.9",
         "sass-loader": "6.0.6",
         "url-loader": "0.6.2",
         "webpack": "3.8.1",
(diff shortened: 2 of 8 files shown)


weerd:
> i know i know, hold your applause.
🙄 haha

mendelB:
Thanks for the great write up! A lot to wrap the ole head around, but this looks awesome. One quick point: 

This stuff - 
image
was a quick fix to the rare scenario where we don't have enough reportbacks to fill the target amount of blocks. https://github.com/DoSomething/phoenix-next/pull/494
Might we still run into that problem?

itsjoekent:
Oh thats why I didn't seem to do anything, interesting. So in theory, if I make a new contentful campaign right now, on this branch I would see a bunch of empty RB cards... Yeah I can add the logic back in, but I'll need to make a quick modification since we're depleting RB's from the state now in two places.

mendelB:
Cool, yeah I'm sure there's a more elegant way to ensure we don't force non-existant reportbacks into the feed.

itsjoekent:
Ok so made a new contentful campaign that has a fake campaign id (so it cant retrieve RB's)

Previously it would display this:
screen shot 2017-11-28 at 11 22 36 am

Now it displays this:
screen shot 2017-11-28 at 11 22 44 am

And this is what a normal feed still looks like:
screen shot 2017-11-28 at 11 23 27 am

resources/assets/actions/feed.js
```diff
@@ -18,8 +22,11 @@ export function conditionallyFetchReportbacks() {
 
     // The number of needed reportbacks are the number of reportback blocks in the
     // community feed + any additional offset that hasn't been filled by blocks.
-    const overflowReportbackBlocks = Math.max(0, getBlockOffset(state) - totalBlocksInFeed(state));
-    const neededReportbacks = totalReportbackBlocksInFeed(state) + overflowReportbackBlocks;
+    const overflowReportbackBlocks = Math.max(0, (
```
weerd: could we rename this to `totolFillerReportbackBlocks`? not sure if `overflowReportbackBlocks` makes the most sense to me 🤔  Thoughts?

weerd: or `fillerReportbackBlocksNeeded`?

itsjoekent: hm

itsjoekent: just to update from slack, we're gonna do a few changes to this file in another PR

PR #536, opened 2017-11-27
Campaign update fixes for preview api

mendelB:
### What does this PR do?
Accounts for scenario where when using contentful's preview API - a `CampaignUpdate`s `author` field might not be set. 

- Set to `null` instead of trying to create an `Author` entity
- In the React component - change Author to defaultProp, and let the `Author` component's default props to take over.

Slack discussion https://dosomething.slack.com/archives/C3ASB4204/p1511818759000348

Fixes #

### Checklist

### DIFF
--- a/resources/assets/components/CampaignUpdate/CampaignUpdate.js
+++ b/resources/assets/components/CampaignUpdate/CampaignUpdate.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import { get } from 'lodash';
 
 import Card from '../Card';
 import Embed from '../Embed';
@@ -9,7 +10,7 @@ import Markdown from '../Markdown';
 import { ShareContainer } from '../Share';
 
 const CampaignUpdate = ({ id, author, content, link, shareLink, bordered, titleLink }) => {
-  const authorFields = (author && author.fields) || {};
+  const authorFields = get(author, 'fields', {});
 
   const isTweet = content.length < 144;
 
--- a/resources/assets/components/CampaignUpdate/CampaignUpdate.js
+++ b/resources/assets/components/CampaignUpdate/CampaignUpdate.js
@@ -9,7 +9,7 @@ import Markdown from '../Markdown';
 import { ShareContainer } from '../Share';
 
 const CampaignUpdate = ({ id, author, content, link, shareLink, bordered, titleLink }) => {
-  const { avatar, jobTitle, name } = author.fields;
+  const authorFields = (author && author.fields) || {};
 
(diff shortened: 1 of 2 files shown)


resources/assets/components/CampaignUpdate/CampaignUpdate.js
```diff
@@ -56,6 +56,9 @@ CampaignUpdate.propTypes = {
 CampaignUpdate.defaultProps = {
   link: null,
   bordered: true,
+  author: {
+    fields: {},
```
itsjoekent: Can we make this consistent with other places where content might be null? (just make it `author: null`, don't read from the object in JS unless certain it's safe)

resources/assets/components/CampaignUpdate/CampaignUpdateContainer.js
```diff
@@ -7,7 +7,9 @@ import { makeShareLink } from '../../helpers';
 
 const mapStateToProps = (state, props) => {
   const campaignUpdate = find(state.campaign.activityFeed, { id: props.id });
-  const { author, content, link } = campaignUpdate.fields;
+  const { content, link } = campaignUpdate.fields;
+  // If author field is unset (i.e. null), let's force through the defaultProps.
+  const author = campaignUpdate.fields.author || undefined;
```
itsjoekent: Hm, given above I think this could probably just be `const author = campaignUpdate.fields.author;`

resources/assets/components/CampaignUpdate/CampaignUpdate.js
itsjoekent: right so I would update that to make sure it can accept null. it's not a huge deal but it's just how we've been doing things in most places

mendelB: Ohh ok gotcha, that makes sense!

resources/assets/components/CampaignUpdate/CampaignUpdate.js
```diff
@@ -9,7 +9,7 @@ import Markdown from '../Markdown';
 import { ShareContainer } from '../Share';
 
 const CampaignUpdate = ({ id, author, content, link, shareLink, bordered, titleLink }) => {
-  const { avatar, jobTitle, name } = author.fields;
+  const authorFields = (author && author.fields) || {};
```
mendelB: Just realized I could probably use lodash `get` here. That'd be nicer

itsjoekent: yeah for sure, go for it if you want!

mendelB:
Added another accountability for when the `content` field isn't filled out in preview API. (or is `null` in general).

PR #538, opened 2017-11-28
Adding Campaign Lead to the Campaign entity

itsjoekent:
### What does this PR do?
- Adds the new campaign lead field (Staff entry type) to the Campaign object.
- Adds an email field to Staff
- Updates logic for fetching the campaign lead to support the new field and legacy json.

**Legacy JSON**

screen shot 2017-11-28 at 2 40 35 pm

**Campaign Lead Field**

screen shot 2017-11-28 at 2 32 19 pm

https://www.pivotaltracker.com/story/show/153177364

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -166,6 +166,7 @@ class Campaign extends Entity implements JsonSerializable
                 'url' => get_image_url($this->coverImage),
                 'landscapeUrl' => get_image_url($this->coverImage, 'landscape'),
             ],
+            'campaignLead' => $this->campaignLead ? new Staff($this->campaignLead->entry) : null,
             'affiliateSponsors' => $this->parseAffiliates($this->affiliateSponsors),
             'affiliatePartners' => $this->parseAffiliates($this->affiliatePartners),
             // @TODO: Why is it 'activity_feed' oy? ;/
--- a/resources/assets/components/Page/CampaignPage/CampaignPage.js
+++ b/resources/assets/components/Page/CampaignPage/CampaignPage.js
@@ -152,7 +152,7 @@ CampaignPage.defaultProps = {
   endDate: null,
   isAffiliated: false,
   totalCampaignSignups: 0,
-  campaignLead: undefined,
+  campaignLead: null,
 };
 
 export default CampaignPage;
--- a/resources/assets/components/Page/CampaignPage/CampaignPageContainer.js
+++ b/resources/assets/components/Page/CampaignPage/CampaignPageContainer.js
@@ -1,8 +1,9 @@
-import { get } from 'lodash';
 import { connect } from 'react-redux';
 
 import CampaignPage from './CampaignPage';
 import { convertExperiment, openModal } from '../../../actions';
+import { getCampaignLead } from '../../../selectors/campaign';
+
 
 /**
  * Provide state from the Redux store as props for this component.
@@ -16,7 +17,7 @@ const mapStateToProps = state => ({
   hasActivityFeed: Boolean(state.campaign.activityFeed.length),
   affiliateSponsors: state.campaign.affiliateSponsors,
   affiliatePartners: state.campaign.affiliatePartners,
-  campaignLead: get(state.campaign.additionalContent, 'campaignLead'),
+  campaignLead: getCampaignLead(state),
   legacyCampaignId: state.campaign.legacyCampaignId,
   shouldShowActionPage: state.admin.shouldShowActionPage,
   slug: state.campaign.slug,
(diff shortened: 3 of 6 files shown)


resources/assets/components/Page/CampaignPage/CampaignPage.js
```diff
@@ -152,7 +152,7 @@ CampaignPage.defaultProps = {
   endDate: null,
   isAffiliated: false,
   totalCampaignSignups: 0,
-  campaignLead: undefined,
+  campaignLead: null,
```
mendelB: I think that this was `undefined` to handle campaigns with no assigned campaign lead, to force falling back on the `defaultProps`.
I know you mentioned that that's not the best way though. 

Should we add a better fallback option here for that scenario? (as of now it would error out trying to get `email` from null in the [`CampaignFooter`](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/CampaignFooter/CampaignFooter.js#L14))

resources/assets/components/Page/CampaignPage/CampaignPageContainer.js
```diff
@@ -16,7 +17,7 @@ const mapStateToProps = state => ({
   hasActivityFeed: Boolean(state.campaign.activityFeed.length),
   affiliateSponsors: state.campaign.affiliateSponsors,
   affiliatePartners: state.campaign.affiliatePartners,
-  campaignLead: get(state.campaign.additionalContent, 'campaignLead'),
+  campaignLead: getCampaignLead(state),
```
weerd: Do we need a whole selector for getting the Campaign Lead?

app/Entities/Campaign.php
```diff
@@ -166,6 +166,7 @@ public function jsonSerialize()
                 'url' => get_image_url($this->coverImage),
                 'landscapeUrl' => get_image_url($this->coverImage, 'landscape'),
             ],
+            'campaignLead' => $this->campaignLead ? new Staff($this->campaignLead->entry) : null,
```
weerd: Ok, I confused myself... instead of a selector we could do a small check here to first look for `$this->campaignLead` and if it exists create a `new Staff()` otherwise check `$this->additionalContent->get('campaignLead')` and create a `new Staff()` (you can do something like in some of the other entities where we fake some info to allow creating a proper version of the entity) otherwise return `null`.

resources/assets/components/Page/CampaignPage/CampaignPage.js
itsjoekent: Oh hm yeah 👍

app/Entities/Campaign.php
itsjoekent: sure thing

app/Entities/Campaign.php
```diff
@@ -142,6 +142,37 @@ public function parseAffiliates($affiliates)
...
+    public function parseCampaignLead($campaignlead, $additionalContent)
+    {
+        if ($campaignlead) {
+            return new Staff($campaignlead->entry);
+        }
+
+        $email = 'help@dosomething.org';
+        $name = 'Us';
+
+        if ($additionalContent && $additionalContent['campaignLead']) {
```
weerd: I think you need to do an `isset()` check here. If `campaignLead` doesn't exist on the `$additionalContent` it's gonna toss an error for an `Undefined index`

```php
if ($additionalContent && isset($additionalContent['campaignLead'])) {
...
}
```

app/Entities/Campaign.php
```diff
@@ -142,6 +142,37 @@ public function parseAffiliates($affiliates)
...
+            $email = $additionalContent['campaignLead']['email'];
```
weerd: Also, in case in the additional content the `campaignLead` is specified, but the `email` or `name` isn't or there's a typo, you can use the null coalesce operator:

```php
$email = $additionalContent['campaignLead']['email'] ?? $email;
$name = $additionalContent['campaignLead']['name'] ?? $name;
```
🎉

itsjoekent: thats fun

weerd: Been wanting to use it on something reasonable for a while now, haha. Makes sense here!

app/Entities/Campaign.php
```diff
@@ -142,6 +142,32 @@ public function parseAffiliates($affiliates)
...
+     * @return array
+     */
+    public function parseCampaignLead($campaignlead, $additionalContent)
+    {
+        if ($campaignlead) {
+            return new Staff($campaignlead->entry);
+        }
+
+        $email = $additionalContent['campaignLead']['email'] ?? 'help@dosomething.org';
+        $name = $additionalContent['campaignLead']['name'] ?? 'Us';
```
weerd: Even better. Nice without the `isset()` in conditional 🎉

mendelB: That's so dope

weerd:
@itsjoekent Ok, I was wrong about trying to pop everything in the `Staff` entity. Let's just revert back to your prior commit and move forward from there. Only thing to add is to make sure the faked return array that's made to _look_ like a `Staff` entity contains all the same items as expected even if they're just `null` or dummy data.

PR #539, opened 2017-11-29
Removing share reducer

itsjoekent:
removes the (useless) share reducer to get rid of these warnings

screen shot 2017-11-29 at 12 52 32 pm


itsjoekent:
for the record, it now throws this in the console (but for the life of me I can't figure out why)

```
Unexpected key "share" found in previous state received by the reducer. Expected to find one of the known reducer keys instead: "admin", "blocks", "campaign", "competitions", "events", "experiments", "modal", "notifications", "reportbacks", "signups", "slideshow", "submissions", "uploads", "user", "quiz", "routing". Unexpected keys will be ignored.
```

DFurnes:
@itsjoekent If we're not using that `share` reducer anymore, it should be removed from the [server-provided state](https://github.com/DoSomething/phoenix-next/blob/c2b07737d9411a90f9deafe7af3f5be4a4c42c34/app/Http/Controllers/CampaignController.php#L72) too. Redux drops that warning in if there's a store w/o reducers to try to be helpful (in case you had a typo or something).

PR #540, opened 2017-11-29
Adding Reportback Settings and refactoring the uploader a bit

itsjoekent:
### What does this PR do?
In order to help clean up the json fields & also set up the new field for our new RB uploader design, this PR is mostly refactoring type work to introduce the new Reportback Uploader Settings. This contains all of our existing popular json fields & the new more information field. If the more information field is present, it will then also add a card next to the uploader with that markup.

Next PR will focus on the other styling stuff & the new reportback affirmation styling.

https://www.pivotaltracker.com/story/show/149011395


app/Entities/Campaign.php
```diff
@@ -187,6 +187,7 @@ public function jsonSerialize()
             'pages' => $this->pages,
             'landingPage' => $this->landingPage,
             'socialOverride' => $this->socialOverride ? new SocialOverride($this->socialOverride->entry) : null,
+            'reportbackUploaderSettings' => $this->reportbackUploaderSettings ? new ReportbackUploaderSettings($this->reportbackUploaderSettings->entry) : null,
```
weerd: We're going to be introducing a campaign settings content type to enable setting lots of configurations for things like this, enabling sixpack experiments toggle and noun/verb settings. So I think this is fine for now but we should shift it over in the future. Maybe add a `// @TODO` comment. I'm adding the config object today, so alternative is giving me a bit to set it up and then we pull stuff from there.

itsjoekent:
OK so this PR blew up a lil bit, mah bad.

new description:

### added support for the photo uploader step
this is a new action step that gives campaign leads more options when configuring the photo uploader. coupled with this were design changes to the photo uploader to give users more context about the process
screen shot 2017-12-05 at 2 02 59 pm

### action wrapper refactoring
the action steps wrapper was a bit wonky with everything going on in there. so I essentially moved all of the component rendering logic into tiny specific functions in the new action renderers file. feels better now!

### misc. fixes
- campaign action step was calling `->first()` on something that was null
-  In StepHeader the hideStepNumber prop should not be required (it's descendants don't require it), and I simplified the props being passed down to the template versions

app/Entities/CampaignActionStep.php
```diff
@@ -37,12 +37,12 @@ public function jsonSerialize()
             'type' => $this->getContentType(),
             'fields' => [
                 'title' => $this->title,
-                'displayOptions' => $this->displayOptions->first(),
+                'displayOptions' => $this->displayOptions ? $this->displayOptions->first() : 'one-third',
```
weerd: use `??`

itsjoekent: will that work if im calling the function on it?

weerd: I believe so if the function returns `null`, but try it and double check. It would only be a problem if the `displayOptions` field might not exist, but that should always be there... just might not have a value.

app/Entities/PhotoUploaderStep.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace App\Entities;
+
+use JsonSerializable;
+
+class PhotoUploaderStep extends Entity implements JsonSerializable
```
weerd: Call this `PhotoUploaderAction` since it may exist outside of the actions steps and also better to match the name of the content type on Contentful.

app/Entities/CampaignActionStep.php
```diff
@@ -37,12 +37,12 @@ public function jsonSerialize()
             'type' => $this->getContentType(),
             'fields' => [
                 'title' => $this->title,
-                'displayOptions' => $this->displayOptions->first(),
+                'displayOptions' => $this->displayOptions->first() ?? 'one-third',
```
weerd: actually, I just tried it myself and it doesn't work. _If_ it contained anything, it would return as a collection which is why `->first()` works... but if it's null, that method doesn't exist. You were right! Change it back! 😬

weerd:
@itsjoekent so the need to deal with [this](https://github.com/DoSomething/phoenix-next/pull/540/files#diff-dcf6db2fbb60f821677b41c220cf0f59R45) was happening because the the `PhotoUploaderComponent` was added in Contentful to the Action Steps before the code was ready to deal with it and it was breaking the campaign on local and staging.

weerd:
Is there a way to make some of these updates without blocking others abilities to work on the test TFJ campaign...?

Like maybe [what you did here](https://github.com/DoSomething/phoenix-next/pull/540/files#diff-df29a57b311e6ffd35546376d83f39eaR115) except instead of automatically returning the `new PhotoUploaderAction` entity, just break and not let anything happen... like as if the Photo Uploader Action wasn't added to the action steps. This could be PR #1 and then in PR #2 is where you actually handle processing it. This way it keeps stuff from breaking 😬

itsjoekent:
update after chatting IRL: the quick solution is to just test with this Photo Upload Action on a different campaign. going forward we'll try to make sure we don't parse the wrong entities in situations like this

app/Entities/PhotoUploaderStep.php
mendelB: Did we end up deciding to not use `PhotoUploaderAction` for this?

resources/assets/components/ActionPage/ActionStepRenderers.js
```diff
@@ -0,0 +1,165 @@
...
+    <div key="member_gallery" className="action-step">
+      <div className="margin-top-xlg margin-bottom-xlg margin-horizontal-md">
+        <h2 className="heading -emphasized legacy-step-header margin-top-md margin-bottom-md">
+          <span>Member Gallery</span>
+        </h2>
+        <PostGalleryContainer key="post_gallery" type="reportback" />
+      </div>
+    </div>
+  );
+}
```
mendelB: This clean up is awesome!

itsjoekent: Agh whoops, thanks for catchin' dat

app/Entities/CampaignActionStep.php
```diff
@@ -37,12 +37,12 @@ public function jsonSerialize()
...
             'fields' => [
                 'title' => $this->title,
-                'displayOptions' => $this->displayOptions->first(),
+                'displayOptions' => $this->displayOptions ? $this->displayOptions->first() : 'one-third',
                 'hideStepNumber' => $hideStepNumber,
                 'content' => $this->content,
                 'background' => get_image_url($this->background, 'landscape'),
                 'photos' => $this->parseActionStepPhotos($this->photos),
-                'customType' => $this->type ?: $this->customType->first(),
+                'customType' => $this->type ?: ($this->customType ? $this->customType->first() : null),
```
weerd: @itsjoekent I thought we chatted about not needing to do this anymore since it was only happening because the `PhotoUploaderAction` was being sent through as a `CampaignActionStep`...?

PR #542, opened 2017-12-01
Alternative landing page

mendelB:
### What does this PR do?

EDIT: All of the below was ported over to the existing landing page component. The `Page` entity has compatibility with old and new Landing Pages. And the `LandingPage` component will render the correct version based on the `isLegacyPitch` field from `additionalContent` in contentful.

Adding new alternative Landing Page component and container.

image

Based on info and designs from [here](https://docs.google.com/document/d/1nEFfJeHRkKgys7rBt3klKFvquJ6ATZaBwAsmx3emzJw/edit#)

Here's how it's shaped out to follow the design spec:

Using the new `Page` entity for this landing page, which allows us to now utilize the 'content' field in the Contentful Page to add the markdown for the main text:
image

There's a test landing page [here](https://app.contentful.com/spaces/81iqaqpfd8fy/entries/5p8Hu5Xa8gaqsqOYAuaCqg) for the missing-in-history test campaign.

We then add a Sidebar element which is using a Custom Block Entity (of the `static` type, though I don't know if that's necessary, as I ended up - ) using a `Card` component for this which meant no need for setting up the `json` field to work properly with the `Block` component
This allows for the yellow bar type CTA in the design:
image
Only issue is that the Sidebar field in Contentful allows for multiple elements so I wasn't sure how to treat this in the landing page where -as of now at least- we're only expecting one static block. I ended up just grabbing the first element from the sidebar array and making the assumption, that it's set properly. (@weerd confirmed this was cool for now).

For the grid layout, things seemed somewhat similar to what we have in [campaign sub pages](https://www.dosomething.org/us/campaigns/lose-your-v-card/pages/faq), so I used the nifty styling scheme for [CampaignSubPage](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/Page/CampaignSubPage/CampaignSubPage.js) with a `.campaign-subpage` parent, and then a `.primary` and `.secondary` div for the content and 'cta'. (With a small adjustment to ensure that the CTA isn't hidden on screens smaller than `$large`).

setting up for this test https://www.pivotaltracker.com/story/show/152925616
https://docs.google.com/document/d/1nEFfJeHRkKgys7rBt3klKFvquJ6ATZaBwAsmx3emzJw/edit#

### Checklist


resources/assets/components/Page/LandingPage/LandingPageAlt.js
```diff
@@ -0,0 +1,97 @@
...
+
+const LandingPage = (props) => {
+  const {
+    affiliateSponsors, blurb, coverImage, endDate, isAffiliated,
+    legacyCampaignId, pitchContent, sidebar, showPartnerMsgOptIn,
+    signupArrowContent, subtitle, tagline, template, title,
+  } = props;
+
+  // Grab the first Sidebar element, assuming here that we only want to be setting one sidebar
+  const sidebarCTA = sidebar[0].fields;
```
itsjoekent: why is sidebar an array if we only want the first element?

resources/assets/components/Page/LandingPage/LandingPageAlt.js
```diff
@@ -0,0 +1,97 @@
...
+
+  return (
+    <div>
+      <LedeBanner
```
itsjoekent: not related to your PR, but wowzers we need to refactor the lede banner... why is the parent component responsible for injecting all of this?!?

resources/assets/components/Page/LandingPage/LandingPageAlt.js
itsjoekent: ohh wait never mind, i read the bottom of your PR

itsjoekent:
@mendelB hmm, I see you imported `convertExperiment` but it doesn't seem to be used anywhere?

mendelB:
That was in the regular LandingPage, so I just copied it over, but I guess we don't need that there until we actually decide to run a test here right?

resources/assets/components/Page/LandingPage/LandingPageAlt.js
mendelB: YES

app/Entities/Page.php
```diff
@@ -37,8 +37,9 @@ public function jsonSerialize()
             'type' => $this->getContentType(),
             'fields' => [
                 'title' => $this->title,
-                'content' => $this->content,
+                'content' => $this->content ? $this->content : $this->additionalContent['pitchContent'],
                 'sidebar' => $this->parseSidebar($this->sidebar),
+                'isLegacyPitch' => array_get($this->additionalContent, 'isLegacyPitch', true),
```
mendelB: Instead of having to manually set this in contentful, we could also just compute this based on if the `content` field is `null` or if the `additionalContent['pitchContent']` is set (the former is better I imagine).

resources/assets/components/Page/LandingPage/LandingPageContainer.js
```diff
@@ -7,22 +7,31 @@ import { convertExperiment } from '../../../actions';
...
+    showPartnerMsgOptIn: get(state.campaign.additionalContent, 'displayAffilitateOptOut', false),
+    signupArrowContent: get(state.campaign.additionalContent, 'signupArrowContent', null),
+    subtitle: state.campaign.callToAction,
+    tagline: get(state.campaign.additionalContent, 'tagline'),
+    template: state.campaign.template,
+    title: state.campaign.title,
+    sidebar: landingPage.sidebar,
+    isLegacyPitch: landingPage.isLegacyPitch,
+  };
+};
```
mendelB: This change looks massive, but all we did was add `IslegacyPitch` and `sidebar` props, and change `pitchContent`, and turn this into a `{}` function

resources/assets/components/Page/LandingPage/landing-page.scss
```diff
@@ -2,36 +2,40 @@
...
-        margin-top: 48px;
-      }
+    div + div {
+      margin-top: 48px;
     }
+  }
+
+  .secondary {
+    display: unset;
+  }
```
mendelB: Another massive looking change, but it's just adding that `.secondary` class rule and fixing the indentation

resources/assets/components/Page/LandingPage/LandingPage.js
```diff
@@ -22,10 +24,28 @@ const formatToMarkup = data => (
...
+          <Markdown>{ pitchContent[0] }</Markdown>
+        </div>
+        <div className="secondary">
+          <Card title={sidebarCTA.title} className="rounded bordered" >
+            <Markdown className="padded" >{ sidebarCTA.content }</Markdown>
+          </Card>
+        </div>
+      </div>
+    );
+
```
mendelB: I have another PR coming that'll move the new pitch page stuff into it's own component so it's cleaner and we can run the experiment

app/Entities/Page.php
weerd: I'd argue that the latter is better. This way it'd also confirm that `pitchContent` is entered in the JSON field. With the former, someone could potentially put something in the `content` field on accident and it would think it's _not_ the legacy pitch even if it has `pitchContent` in the JSON field. Thoughts? 🤔

weerd: One caveat is that for Sixpack experiments we'd have to make sure that _both_ are filled out though right?

mendelB: Yeah true to the first point. And indeed in the next PR I removed this field altogether as they both need to be filled out anyway

mendelB: Instead, it adds a `LegacyPitchContent` field that grabs from the `additionalContent` or defaults to `null`

weerd: This ephemeral _next PR_ is it getting pushed up? haha ;)

mendelB: lol. NEVER!

resources/assets/components/Page/LandingPage/landing-page.scss
```diff
@@ -2,36 +2,40 @@
...
-      div + div {
-        margin-top: 48px;
-      }
+    div + div {
+      margin-top: 48px;
     }
+  }
+
+  .secondary {
+    display: unset;
```
weerd: Is `unset` supported on our target browsers?

weerd: Pretty cool rule, just want to make sure it works as intended :)

mendelB: Turns out it did not :( 
Changed to `display: block`

PR #543, opened 2017-12-04
Testing testing 1 2 3

mendelB:
### What does this PR do?
- Adds support for both versions of Pitch Page (Or in other words: the Landing Page primary content).
- Moving the new pitch page content into its own component.
- Adding sixpack A/B experiment to test between both versions of the pitch page content
- Adding experiment conversion for all experiments specified as `convertOnSignupIntent` in SignupButton when user clicks a Signup button.

We've decided to go with testing the legacy pitch page vs the new (mosaic?) pitch page.
The experiment will only run if the campaign settings 'allow A / B testing' is set to `true`. Any other campaign with a Landing Page will just default to the 'control' case which is the legacy pitch page.
Landing Pages will be expected to have both the old json format (pitch content as an array in the additional content json field.) as well as Markup content in the actual 'Content' field. The php `Page` entity will take care of setting those up, so that the component on the front end will just grab those fields in the container.

https://www.pivotaltracker.com/story/show/152925616
Test spec:
https://docs.google.com/document/d/1nEFfJeHRkKgys7rBt3klKFvquJ6ATZaBwAsmx3emzJw/edit#

### Checklist

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -188,6 +188,7 @@ class Campaign extends Entity implements JsonSerializable
             'landingPage' => $this->landingPage ? new Page($this->landingPage->entry) : null,
             'socialOverride' => $this->socialOverride ? new SocialOverride($this->socialOverride->entry) : null,
             'additionalContent' => $this->additionalContent,
+            'allowExperiments' => $this->campaignSettings ? $this->campaignSettings->allowExperiments : null,
         ];
     }
 }
--- a/resources/assets/components/Page/LandingPage/LandingPage.js
+++ b/resources/assets/components/Page/LandingPage/LandingPage.js
@@ -23,8 +23,8 @@ const formatToMarkup = data => (
 
 const LandingPage = (props) => {
   const {
-    affiliateSponsors, blurb, coverImage, endDate,
-    isAffiliated, legacyCampaignId, legacyPitchContent,
+    affiliateSponsors, blurb, convertExperiment, coverImage,
+    endDate, isAffiliated, legacyCampaignId, legacyPitchContent,
     pitchContent, sidebar, showPartnerMsgOptIn, signupArrowContent,
     subtitle, tagline, template, title,
   } = props;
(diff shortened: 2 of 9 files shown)


app/Entities/Campaign.php
```diff
@@ -188,6 +188,7 @@ public function jsonSerialize()
             'landingPage' => $this->landingPage ? new Page($this->landingPage->entry) : null,
             'socialOverride' => $this->socialOverride ? new SocialOverride($this->socialOverride->entry) : null,
             'additionalContent' => $this->additionalContent,
+            'allowExperiments' => $this->campaignSettings ? $this->campaignSettings->allowExperiments : null,
```
weerd: 🎉

resources/assets/components/Page/LandingPage/LandingPage.js
```diff
@@ -23,29 +23,14 @@ const formatToMarkup = data => (
 
 const LandingPage = (props) => {
   const {
-    affiliateSponsors, blurb, coverImage, endDate,
-    isAffiliated, isLegacyPitch, legacyCampaignId, pitchContent,
-    sidebar, showPartnerMsgOptIn, signupArrowContent, subtitle,
-    tagline, template, title,
+    affiliateSponsors, blurb, convertExperiment, coverImage,
+    endDate, isAffiliated, legacyCampaignId, legacyPitchContent,
+    pitchContent, sidebar, showPartnerMsgOptIn, signupArrowContent,
```
weerd: can we switch to calling `pitchContent` just `content`?

weerd: And `legacyPitchContent` to `legacyContent`.

weerd: @mendelB actually... don't worry about this now. We can update once the experiment is over and we've selected a winner.

resources/assets/components/SignupButton/SignupButton.js
```diff
@@ -35,6 +45,8 @@ const SignupButtonFactory = (WrappedComponent, source = null, sourceData = null)
     clickedSignUp: PropTypes.func.isRequired,
     template: PropTypes.string,
     trackEvent: PropTypes.func.isRequired,
+    experiments: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
+    convertExperiment: PropTypes.func.isRequired,
```
weerd: Should these items be required? Don't we have signup buttons that aren't involved in experiments?

mendelB: Exactly. So it's only going to use the experiments in state i.e: that have been 'activated' and thus added to the stores `component` object. Then it'll just kill all birds (Save the birds!) with a stone and convert any experiment that should be converted on signup intent.

PR #545, opened 2017-12-04
Configure Northstar resource server via Gateway for making API calls.

weerd:
### What does this PR do?
My other PR for API stuff was getting a little large so I decided to split out the work for using Gateway to connect to Northstar acting as a resource server.

Refs https://www.pivotaltracker.com/story/show/151696124

### DIFF
--- a/app/Console/Commands/SetupCommand.php
+++ b/app/Console/Commands/SetupCommand.php
@@ -59,6 +59,8 @@ class SetupCommand extends Command
 
         $this->runCommand('key:generate', 'Creating application key');
 
+        $this->runCommand('gateway:key', 'Fetching public key from Northstar');
+
         $this->runCommand('migrate', 'Running database migrations');
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,7 @@
             "php artisan optimize"
         ],
         "compile": [
-          "php artisan gateway:key"
+            "php artisan gateway:key"
         ]
     },
     "config": {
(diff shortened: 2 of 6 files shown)


app/Console/Commands/SetupCommand.php
```diff
@@ -59,6 +59,8 @@ public function handle()
 
         $this->runCommand('key:generate', 'Creating application key');
 
+        $this->runCommand('gateway:key', 'Fetching public key from Northstar');
```
DFurnes: Let's make sure we run `php artisan gateway:key` on each of our servers. We'll also want this to run on Heroku during our build process there, so let's add it to our `composer.json` [as a compile step](https://github.com/DoSomething/rogue/blob/fc1b29aca7d8a7c585cfc6fb1d8730a16e6fb2fb/composer.json#L58-L60).

```json
  "scripts": {
    "compile": [
      "@php artisan gateway:key"
    ]
  },
```

weerd: Ah jes! Good call. So this'll take care of our AWS env and anything on Heroku since it happens on each build 👌

itsjoekent: not a huge deal but it could also go here https://github.com/DoSomething/phoenix-next/blob/dev/bootstrap/setup.sh

weerd: @itsjoekent true, but for the time being we need it to also apply to the AWS instances along w/ the Heroku ones!

PR #546, opened 2017-12-04
changing pitchContent to optional prop and adding defaultProp

mendelB:
### What does this PR do?
Super small fix to the new Pitch Page situation:

We had the `pitchContent` prop marked as `required` in both the `LandingPage` and `LandingPageContentAlt` but on any legacy campaigns with Landing pages but no A/B testing, this value was `null` which logs an (non breaking) error.


weerd:
@mendelB after #585 is all set and merged, does this PR even apply anymore?

mendelB:
Indeed it does not! We shall bid this one sayonara 👋

PR #547, opened 2017-12-05
Mosaic arrows

mendelB:
### What does this PR do?
Adds the `CampaignSignupArrow` to the `MosaicTemplate`

I had done some work for this a while back so my memory is slightly hazy but, what made this complicated was the narrower width of the - large screen - half of Lede Banner, as well as the wider signup button on smaller screens. I ended up using the `-right` class for the arrow, and a some custom styling to finagle the arrow to look nice throughout the various breakpoints.

But on the whole, this is a pretty straightforward addition of the arrow to the mosaic template

Large:
image

Medium:
image

Small:
image

(That margin between arrow and Banner title on small and medium is spacing for the optional blurb)

https://www.pivotaltracker.com/story/show/153411597

### DIFF
--- a/resources/assets/components/CampaignSignupArrow/campaignSignupArrow.scss
+++ b/resources/assets/components/CampaignSignupArrow/campaignSignupArrow.scss
@@ -48,14 +48,14 @@
     }
 
     @include media($medium) {
-      background: url("../../../../node_modules/@dosomething/forge/assets/images/callout/arrow-left-black.png") 50% 50% / 35px no-repeat transparent !important;
+      background: url("~@dosomething/forge/assets/images/callout/arrow-left-black.png") 50% 50% / 35px no-repeat transparent !important;
       height: 19px;
       left: 0;
       margin-top: -9px;
       right: auto;
       top: 50%;
       width: 38px;
-      background-image: url("../../../../node_modules/@dosomething/forge/assets/images/callout/arrow-left-white.png") !important;
+      background-image: url("~@dosomething/forge/assets/images/callout/arrow-left-white.png") !important;
 
       transform: rotate(270deg);
     }
--- a/resources/assets/components/CampaignSignupArrow/campaignSignupArrow.scss
+++ b/resources/assets/components/CampaignSignupArrow/campaignSignupArrow.scss
@@ -48,14 +48,13 @@
     }
 
     @include media($medium) {
-      background: url("~@dosomething/forge/assets/images/callout/arrow-left-black.png") 50% 50% / 35px no-repeat transparent !important;
+      background: url("~@dosomething/forge/assets/images/callout/arrow-left-white.png") 50% 50% / 35px no-repeat transparent !important;
       height: 19px;
       left: 0;
       margin-top: -9px;
       right: auto;
       top: 50%;
       width: 38px;
-      background-image: url("~@dosomething/forge/assets/images/callout/arrow-left-white.png") !important;
 
       transform: rotate(270deg);
     }
--- a/resources/assets/components/CampaignSignupArrow/campaignSignupArrow.scss
+++ b/resources/assets/components/CampaignSignupArrow/campaignSignupArrow.scss
@@ -13,7 +13,7 @@
   }
 }
 
-.message-callout.-right {
+.message-callout.-mosaic-arrow {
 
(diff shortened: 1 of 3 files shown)


weerd:
@lkpttn are we cool with the down pointing arrow or should we be using that curved one that points down?

mendelB:
Ok Cool! The reason I went with this alternative was that in the mosaic template, the signup button is full width which clashes with the curved arrow format.

mendelB:
Luke wanted to go with the curved arrow on mobile, along with some slight altercations:
https://dosomething.slack.com/files/U025G4PTC/F8A7KCL1F/screen_shot_2017-12-06_at_2.45.04_pm.png

mendelB:
I added some adjustments based on Lukes feedback.

The default setting will be based on the `.-above` [forge](https://github.com/DoSomething/forge/blob/dev/scss/_components/_message-callout.scss) rules with some slight adjustments. So mobile will have the curved arrow. Anything larger will get the -copied over- rules from forge that apply to `.-right` scenario with a bunch of overrides. It's a bit messy since this while mosaic business is quite a departure from the layout for which these forge rules were applied in the first place.

resources/assets/components/CampaignSignupArrow/campaignSignupArrow.scss
```diff
@@ -12,3 +12,57 @@
...
+    }
+  }
+
+  .message-callout__copy::before {
+    @include media($larger) {
+      transform: none;
+    }
+
+    @include media($medium) {
+      background: url("../../../../node_modules/@dosomething/forge/assets/images/callout/arrow-left-black.png") 50% 50% / 35px no-repeat transparent !important;
```
mendelB: Is there a better way to grab an asset from forge?

DFurnes: You should be able to use `~@dosomething/forge/...` to automatically resolve the `node_modules` path.

mendelB: Ah just spotted https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/scss/base.scss#L4

resources/assets/components/CampaignSignupArrow/campaignSignupArrow.scss
```diff
@@ -12,3 +12,57 @@
...
+      height: 19px;
+      left: 0;
+      margin-top: -9px;
+      right: auto;
+      top: 50%;
+      width: 38px;
+      background-image: url("../../../../node_modules/@dosomething/forge/assets/images/callout/arrow-left-white.png") !important;
```
DFurnes: Can this be merged into the `background` shorthand above?

mendelB: How would that work? For context, I was sort of copying the format used in forge with the `background` and then an additional `background-image` white arrow for `-white`:

image

resources/assets/components/CampaignSignupArrow/campaignSignupArrow.scss
```diff
@@ -12,3 +12,57 @@
...
+
+    @include media($medium) {
+      background: url("~@dosomething/forge/assets/images/callout/arrow-left-black.png") 50% 50% / 35px no-repeat transparent !important;
+      height: 19px;
+      left: 0;
+      margin-top: -9px;
+      right: auto;
+      top: 50%;
+      width: 38px;
+      background-image: url("~@dosomething/forge/assets/images/callout/arrow-left-white.png") !important;
```
DFurnes: Can you replace the path on line 51 with this one (since it's overwriting the same property on the same selector)?

```diff
- background: url("~@dosomething/forge/assets/images/callout/arrow-left-black.png") 50% 50% / 35px no-repeat transparent !important;
+ background: url("~@dosomething/forge/assets/images/callout/arrow-left-white.png") 50% 50% / 35px no-repeat transparent !important;
```

mendelB: oh ok gotcha! fixing that now. (I'd tried balancing the two images better, and there was some wonkiness, but I think I was just doing it wrong)

PR #548, opened 2017-12-06
Typeform survey

mendelB:
### What does this PR do?
Adding framework for embedding a Typeform survey in a modal.

- [Added](https://github.com/DoSomething/phoenix-next/commit/35ef04956ffea8540b48619faf8f5140433206d5) the typeform embed js [script](https://admin.typeform.com/form/5888673/distribute/#/section/embedllecedor) in a partial. (And using the nifty [Laravel stacks](https://laravel.com/docs/5.5/blade#stacks) Diego showed me to push our script to the stack in the Campaign show view).

- [Adding](https://github.com/DoSomething/phoenix-next/commit/ab61bbd4b0b69444dbbd844d08d3fd919585cefc) a new `SurveyModal` component to our suite of modals, to create the pop up and embed the typeform modal when launched.

(The dataUrl is set as a constant in a new `nps_survey` constants file. Please ignore the other constant there it's meant to be part of the next PR).

I split this out as there's a ton of file changes and stuff involved here. But this is the beginnings of adding our NPS Survey from Typeform into a modal.

image

https://www.pivotaltracker.com/story/show/153366664

### Checklist


resources/views/campaigns/show.blade.php
```diff
@@ -14,4 +14,7 @@
         </div>
     </div>
 
+    @push('scripts')
```
itsjoekent: Is there a reason this is specific to the campaigns page?

mendelB: Yeah that's a good point. I was discussing this with Diego - because as of now the Modals and thus the NPS survey are kind of bound to the Campaign context, so -for now at least- it only pays to add this in for campaigns.

PR #549, opened 2017-12-06
Update to restore old code approach.

weerd:
### What does this PR do?
Reverts the code that transforms a couple fields in the `CampaignActionStep` entity. It was changed before due to the `PhotoUploaderAction` accidentally getting processed as a campaign action step, but this shouldn't ever be the case so reverting the code back since the change was unnecessary.

See conversation [here](https://github.com/DoSomething/phoenix-next/pull/540/files#r155355244).


itsjoekent:
sorry about that!

PR #550, opened 2017-12-07
Typeform survey launcher

mendelB:
### What does this PR do?
Adds core `Survey` component which functions as the launcher for the NPS (or any) survey in the Survey Modal.

EDIT:

- Adding a `SurveyModal` container along with user id, campaign id (legacy and contentful) to the `data-url` to be passed along to typeform.

The component sets a timer when user visits site (to be more specific, as of now this is only triggered on a campaign page and it's subpages, which is kind of our PN site right now), after the specified amount of elapsed seconds it triggers the `openModal` to open the survey modal which will embed the survey.

- Note that the only conditional set as of yet is that the user is authenticated. Much more to come, just trying to build this somewhat incrementally.

- This timer will refresh every time the user reloads a page, so anything outside of the SPA flow will restart this timer. (Perhaps we can handle this better once (or if) we introduce cookies or something to track user survey participation. Then we can perhaps track if they've immediately recently been on a PN page and we can just build on existing countdown).

- I shall remove the `<Survey />` from the `Campaign` component before merging this. Added for 

More to come!

https://www.pivotaltracker.com/story/show/153366664
https://www.pivotaltracker.com/story/show/153288461
#548 

### Checklist

### DIFF
--- a/resources/assets/components/Modal/configurations/SurveyModal.js
+++ b/resources/assets/components/Modal/configurations/SurveyModal.js
@@ -3,7 +3,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { SURVEY_DATA_URL } from '../../../constants/survey';
+const SURVEY_DATA_URL = 'https://dosomething.typeform.com/to/Bvcwvm';
 
 class SurveyModal extends React.Component {
   componentDidMount() {
--- a/resources/assets/components/Modal/configurations/SurveyModal.js
+++ b/resources/assets/components/Modal/configurations/SurveyModal.js
@@ -3,6 +3,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import { makeUrl } from '../../../helpers';
+
 const SURVEY_DATA_URL = 'https://dosomething.typeform.com/to/Bvcwvm';
 
 class SurveyModal extends React.Component {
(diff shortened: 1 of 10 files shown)


resources/assets/components/Modal/configurations/SurveyModal.js
```diff
@@ -11,10 +11,12 @@ class SurveyModal extends React.Component {
   }
 
   render() {
+    const { dataUrl, northstarId, campaignId, legacyCampaignId } = this.props;
+
     return (
       <div
         className="modal__slide typeform-widget"
-        data-url={this.props.dataUrl}
+        data-url={`${dataUrl}?northstar_id=${northstarId}&campaign_id=${campaignId}&legacy_campaign_id=${legacyCampaignId}`}
```
DFurnes: Let's create this with the `URL` constructor so those params get escaped if they ever have weird characters:

```js
const typeformUrl = new URL(dataUrl);
typeFormUrl.search = new URLSearchParams({'northstar_id': northstarId, 'campaign_id': campaignId, 'legacy_campaign_id': legacyCampaignId});

<div data-url={typeformUrl.href} ...
```

It could also be nice to make our own `url(path, query)` helper function to make this easier in the future?

resources/assets/components/Modal/configurations/SurveyModal.js
```diff
@@ -23,10 +25,16 @@ class SurveyModal extends React.Component {
 
 SurveyModal.propTypes = {
   dataUrl: PropTypes.string,
+  northstarId: PropTypes.string,
```
DFurnes: Should these be required?

resources/assets/components/Survey/Survey.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { SURVEY_MODAL } from '../Modal';
+import { SURVEY_COUNTDOWN } from '../../constants/survey';
```
DFurnes: If this isn't used elsewhere, I think it's fine to keep in this file.

mendelB:
Thanks @DFurnes! I'll work on getting those changes in.

PR #551, opened 2017-12-07
New affirmation

itsjoekent:
### What does this PR do?

**+** Added a new variable to the Redux store to keep track if we should be displaying the Reportback affirmation (and added associated actions/reducers).
**+** Added an optional "x" button to the Card component and the ability to pass in a callback when it's pressed. 
**+** Added more utility classes. Specifically one for vertical spacing, one for horizontal centering, and one for padding on mobile only.

**~** Changed the logic in the RB uploader on when we display messaging (Affirmation is displayed elsewhere now)
**~** Replaced `BlockWrapper` usage with `Card` in the RB uploader

**-** Removed a bunch of no longer needed SCSS from the RB uploader 

## Various states at different screen sizes
screen shot 2017-12-07 at 10 44 49 am
screen shot 2017-12-07 at 10 44 27 am
screen shot 2017-12-07 at 10 43 50 am

- The design of the media uploader changed to be full width (for now) after consulting with Luke. We can revisit this later to add the extra columns back, but I didn't want to play with all of the hardcoded values that are there and block the prod deploy even longer.

https://www.pivotaltracker.com/story/show/149011395


resources/assets/components/ReportbackUploader/ReportbackUploader.js
```diff
@@ -129,21 +135,26 @@ class ReportbackUploader extends React.Component {
                   { showQuantityField ? impactInput : null }
```
mendelB: Not directly related to this PR. But I bumped into an error on teens-for-jeans (but presumable this applies to any Campaign with no quantity.) where since the props `showQuantityField` was `false`, the `impact` field was not rendered, but in the [`handleSubmitForm`](https://github.com/DoSomething/phoenix-next/pull/551/files#diff-201edebec1634a925cb7e8371eba4809R65) `showQuantityField` seems to be forced to be `true` and so errors out when trying to grab `this.impact.value`.

resources/assets/components/Block/BlockWrapper.js
```diff
@@ -11,6 +11,8 @@ type BlockTitleLinkProps = {
   title: ?string,
 };
 
+// NOTE: BlockWrapper is being deprectaed in favor of Card.
```
weerd: 🎉

resources/assets/scss/base.scss
```diff
@@ -168,6 +168,16 @@ h1, h2, h3, p {
   padding-right: $half-spacing;
 }
 
+.padding-horizontal-md--mobile {
```
weerd: Lets not do media queries in these utility classes for now. Seems real weird it's suffixed w/ `mobile` but also has a `tablet` breakpoint 😱  I'll be switching us to TailwindCSS soon™ and this won't be a thing we'll need to worry about with how it implements responsiveness w/ utility classes. Do this in the component for now... add a specific class for this component and add the breakpoint styles for that specific class.

resources/assets/components/ReportbackUploader/ReportbackUploader.js
itsjoekent: @mendelB Thanks for catching that, I think I know what's going on

resources/assets/scss/base.scss
itsjoekent: okie dokes

itsjoekent:
thanks for the reviews team, sorry for the giant pr

PR #553, opened 2017-12-07
Updating quantity logic.

weerd:
### What does this PR do?
This PR updates how we handle quantity on the ReportbackUploader component. We now ask for the new total of items collected. If the user already had a quantity value from a prior submission we take the old value and subtract it from the new quantity to obtain the difference and send that to Rogue. It also updates the Reportback Uploader with the new Impact Display messaging to show a returning user how much they've collected from previous posts.

It also adds some validation logic for the quantity field (now that it became an optional field to show) and fixed that up. Also updates with some custom messaging.

_Note:_ Upcoming PR will handle the styling issue with the vertical bar breaking and not going full height!

I tried various scenarios out if the quantity is specified to show in the Reportback Uploader, including submitting a string of text as quantity and seeing how that gets added if a previous value existed. It ends up resulting in `NaN` which turns out to be fine because that gets submitted to PN and doesn't validate as an integer 🎉 Also tried submitting an empty field, or a number lower than the previous impact quantity and all works as expected.

Quantity Properly Validated when designated to show on campaign:
quantity validation

Quantity Display:
quantity display

Screen shot for reference of impact display on small screen, but for the most part the responsiveness is handled by the prior work in the RB Uploader restructuring and design work.
screen shot 2018-01-11 at 6 11 31 pm

Refs https://www.pivotaltracker.com/story/show/153297451

### Checklist

### DIFF
--- a/app/Http/Controllers/ReportbackController.php
+++ b/app/Http/Controllers/ReportbackController.php
@@ -47,7 +47,7 @@ class ReportbackController extends Controller
      */
     public function store(Request $request)
     {
-        $impactMessage = $request->input('previousImpact') > 0 ? 'The quantity needs to be greater than your previous quantity.' : 'The quantity field needs to be a number greater than 0.';
+        $impactMessage = $request->input('previousImpact') > 0 ? 'The quantity needs to be greater than your previous quantity of '.$request->input('previousImpact').'.' : 'The quantity field needs to be a number greater than 0.';
 
         $messages = [
             'media.required' => 'An uploaded photo is required.',
--- a/app/Http/Controllers/ReportbackController.php
+++ b/app/Http/Controllers/ReportbackController.php
@@ -50,7 +50,7 @@ class ReportbackController extends Controller
         $this->validate($request, [
             'media' => 'required|file|image',
             'caption' => 'required|min:4|max:60',
-            'impact' => 'required|numeric',
+            'impact' => 'required|integer|min:1',
             'whyParticipated' => 'required',
         ]);
 
--- a/app/Http/Controllers/ReportbackController.php
+++ b/app/Http/Controllers/ReportbackController.php
@@ -6,6 +6,7 @@ use Illuminate\Http\Request;
 use App\Services\PhoenixLegacy;
 use App\Services\UploadedMedia;
 use Illuminate\Support\Facades\Log;
+use App\Http\Requests\StorePostRequest;
 
 class ReportbackController extends Controller
 {
@@ -42,28 +43,11 @@ class ReportbackController extends Controller
     /**
      * Store a newly created resource in storage.
      *
-     * @param  \Illuminate\Http\Request  $request
+     * @param  \App\Http\Requests\StorePostRequest $request
      * @return \Illuminate\Http\Response
      */
-    public function store(Request $request)
+    public function store(StorePostRequest $request)
     {
-        $impactMessage = $request->input('previousImpact') > 0 ? 'The quantity needs to be greater than your previous quantity of '.$request->input('previousImpact').'.' : 'The quantity field needs to be a number greater than 0.';
-
-        $messages = [
-            'media.required' => 'An uploaded photo is required.',
-            'impact.required_if' => 'The quantity field is required.',
-            'impact.integer' => 'The quantity field needs to be a number.',
-            'impact.min' => $impactMessage,
-        ];
-
-        $this->validate($request, [
-            'media' => 'required|file|image',
-            'caption' => 'required|min:4|max:60',
-            'impact' => 'required_if:showImpact,1|integer|min:1',
-            'showImpact' => 'boolean',
-            'whyParticipated' => 'required',
-        ], $messages);
-
         $path = UploadedMedia::store($request->file('media'));
 
         $response = $this->phoenixLegacy->storeReportback(
--- a/app/Http/Controllers/ReportbackController.php
+++ b/app/Http/Controllers/ReportbackController.php
@@ -48,15 +48,19 @@ class ReportbackController extends Controller
     public function store(Request $request)
     {
         $messages = [
-            // override messages
+            'media.required' => 'An uploaded photo is required.',
+            'impact.required_if' => 'The quantity field is required.',
+            'impact.integer' => 'The quantity field needs to be a number.',
+            'impact.min' => 'The quantity field needs to be a number greater than 0.'
         ];
 
+        Log::info('Show Impact', [$request->input('showImpact')]);
         Log::info('Submission', $request->input());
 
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/ReportbackController.php
```diff
@@ -42,7 +42,7 @@ public function store(Request $request)
         $this->validate($request, [
             'media' => 'required|file|image',
             'caption' => 'required|min:4|max:60',
-            'impact' => 'required|numeric',
+            'impact' => 'required|integer|min:1',
```
weerd: changed this from `numeric` to `integer` since the former allows for floating points... and added `min:1` to keep users from entering negative values!

itsjoekent: @weerd wait what if i want to submit half a jean

itsjoekent:
@weerd is the new styling showing the total quantity in a different PR?

weerd:
There was no styling update to the quantity field!

weerd:
Oh wait, I see what you mean. I didn't see that in Abstract.... thought maybe it would be coming later and right now it's just a language change. @lkpttn mind clarifying?

app/Http/Controllers/ReportbackController.php
weerd: :trollface:

app/Http/Controllers/ReportbackController.php
```diff
@@ -47,12 +47,22 @@ public function index(Request $request)
      */
     public function store(Request $request)
     {
+        $impactMessage = $request->input('previousImpact') > 0 ? 'The quantity needs to be greater than your previous quantity.' : 'The quantity field needs to be a number greater than 0.';
```
weerd: Not ideal, and something we can improve on... maybe a [`FormRequest` class](https://laravel.com/docs/5.5/validation#form-request-validation)?

app/Http/Requests/StorePostRequest.php
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        return array_merge([
+            'caption' => 'required|min:4|max:60',
+            'impact' => 'required_if:showImpact,1|integer|min:1',
+            'media' => 'required|file|image',
+            'showImpact' => 'boolean',
+        ], $additionalRules);
+    }
+}
```
weerd: Turns out I had forgotten that currently, with the Refer-A-Friend feature, we need the Reportback submission form to account for both variations when there's a `PhotoUploadAction` and when there's a `ReferralAction` (that also has a photo upload). This whole situation is a little weird.

That being said, I found that setting up a Form Request helped DRY up the validation logic in both the `ReportbackController` and the `ReferralController`, and allowed extracting a bit of extra logic I was doing for some of the messaging customizations out. Ended up calling it the `StorePostRequest` request, because it was neither a solely `PhotoUploadRequest` or a `ReferralRequest`. I think this is ok _for now_...

Moving forward I would like product and tech to discuss the possibility of keeping actions distinct and separate from each other. So instead of having a mish-mosh of the Photo Upload AND Refer-A-Friend actions together in the RB uploader interface, the RB Uploader with Photo Upload would be one action. And if in a campaign we wanted both actions to upload a photo (but with no quantity field or why participated field, they'd be toggles in the `PhotoUploadAction` content-type) and refer-a-friend, the Referral Action would live alongside on the page but separate container and a separate form submission (like we've been talking about for multiple actions!). I feel this will make the code, logic, the building of pages by campaign leads and in general all our lives a lot easier. 😔

app/Http/Requests/StorePostRequest.php
```diff
@@ -0,0 +1,71 @@
...
+
+        if ($this->input('actionType') === 'photoUploaderAction') {
+            $additionalRules = [
+                'whyParticipated' => 'required',
+            ];
+        }
+
+        if ($this->input('actionType') === 'referralAction') {
+            $additionalRules = [
+                'friendEmail' => 'required|email',
```
weerd: @mendelB added validation for email in the referral ;)

app/Http/Requests/StorePostRequest.php
mendelB: This custom Request is dope! 
I def agree that this is a bit convoluted, and separating these two would make things way cleaner. In terms of experience though for the `team-dream` for example, two separate form submissions for photo and scholarship/RAF form might be inelegant no?

resources/assets/components/ReportbackUploader/ReportbackUploader.js
```diff
@@ -148,11 +157,30 @@ class ReportbackUploader extends React.Component {
         },
       }), {});
 
+    this.previousImpact = 35;
```
mendelB: Just a note to remember to take this out before merging.

resources/assets/components/ReportbackUploader/ReportbackUploader.js
```diff
@@ -82,10 +88,13 @@ class ReportbackUploader extends React.Component {
...
 
     const reportback = {
+      actionType: this.props.referralRB ? 'referralAction' : this.props.actionType,
       campaignId: this.props.legacyCampaignId,
       caption: this.caption.value,
-      impact: this.props.showQuantityField ? this.impact.value : 1,
+      impact: this.props.showQuantityField ? this.determineImpact(this.impact.value) : 1,
       media: this.state.media,
+      previousImpact: this.previousImpact,
+      showImpact: this.props.showQuantityField ? 1 : 0,
```
mendelB: What's the `showImpact` field for?

app/Http/Requests/StorePostRequest.php
weerd: Depends if we treat it as like a checklist of things to do (like we've talked about with multiple actions at times). That's why I think it's a broader conversation with product as well so we can cater the experience to feeling elegant while it still being separate actions.

resources/assets/components/ReportbackUploader/ReportbackUploader.js
weerd: oh crap! nice catch.

resources/assets/components/ReportbackUploader/ReportbackUploader.js
weerd: @mendelB this is specifically so we can have a conditional `required_if` rule that only validates the `impact` field **if** the `showImpact` boolean is set to true (or 1 in this scenario since after some testing I tried passing `true` but gets converted to a string `"true"` and that doesn't trigger the conditional rule correctly.

mendelB: Ohhh right yeah I see that now. Cool!

PR #555, opened 2017-12-11
Nps Survey env setup

mendelB:
### What does this PR do?
- Adds support for an env variable `SURVEY_ACTIVE` to toggle the survey's active status.
- Passing var along to client in `campaigns@show` for React.
- checking survey active status before showing survey.

I called the variable `SURVEY_ACTIVE`
and placed the config for it in `config/services`.
Please let me know if there's a better name / place for this variable.

Turning of NPS for deploy
https://dosomething.slack.com/archives/C2BPA7M8F/p1513004694000362?thread_ts=1513004418.000807&cid=C2BPA7M8F

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -297,6 +297,7 @@ function get_client_environment_vars()
         'SIXPACK_TIMEOUT' => config('services.sixpack.timeout'),
         'PUCK_URL' => config('services.analytics.puck_url'),
         'SURVEY_ACTIVE' => config('services.survey.active'),
+        'SURVEY_ENABLED' => config('services.survey.enabled'),
     ];
 }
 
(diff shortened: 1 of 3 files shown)


app/helpers.php
```diff
@@ -296,6 +296,7 @@ function get_client_environment_vars()
         'SIXPACK_COOKIE_PREFIX' => config('services.sixpack.prefix'),
         'SIXPACK_TIMEOUT' => config('services.sixpack.timeout'),
         'PUCK_URL' => config('services.analytics.puck_url'),
+        'SURVEY_ACTIVE' => config('services.survey.active'),
```
weerd: I think I'd like to stick w/ `enabled` terminology like in `SIXPACK_ENABLED` in ENV vars. Can we change to `SURVEY_ENABLED`?

weerd: Also alphabetize the list 💃

mendelB: Ahh right right! Will change

PR #557, opened 2017-12-11
Modal redesign 2

itsjoekent:
### What does this PR do?
This PR can't be shipped in pieces because if the first half get shipped without the second, we will lose functionality. So I'm throwing it up early, so it's not a surprise to anyone!

New modal designs require putting the "X" on Card components rather than a global X floating at the top right (Except for Surveys). The work so far is to pass the modal close function down to the necessary components so they can optionally append it to their Card components


itsjoekent:
Closing this PR but not deleting branch yet (so I can reference later)

I have a cleaner approach that will reduce the files involved and also allow this to be shipped in non-blocking pieces

PR #559, opened 2017-12-11
Adding the Modal broadcast system

itsjoekent:
### What does this PR do?
This PR introduces the Modal Broadcast System ™️ 

This system is going to be used so deeply nested components within a modal can have a clean interface to check if they are rendering within a modal or not, and apply whatever conditional logic is required. 

This approach uses the same architecture as [Puck](https://github.com/DoSomething/puck-client/blob/master/lib/Connector.js) in order to prevent components themselves from touching React context. 

In practice, it would look something like this.

```
import { ModalListener } from '../Modal';
import Card from './Card';

return ModalListener(Card);
```

Then within Card, you have a new prop called `withinModal`.

### DIFF
--- a/resources/assets/components/Modal/Modal.js
+++ b/resources/assets/components/Modal/Modal.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Portal from 'react-portal';
+import ModalBeacon from '../Modal';
 import './modal.scss';
 
 class Modal extends React.Component {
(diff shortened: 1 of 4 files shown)


DFurnes:
Whenever we get around to refactoring modals to render as normal routes (rather than Redux state), I think it'd make sense to use `<Route>` components for conditional logic like this.

itsjoekent:
i am closing this PR out along with the modal broadcast one. i have a (hopefully) less complicated solution

PR #560, opened 2017-12-11
Cards in modal

itsjoekent:
### What does this PR do?
Following up on #559 -

This PR adds the Modal Listener to the Card, which allows the card to determine if the close button to link to the close modal action. With this, I also created a Card Container. I will update all of the Card imports in a follow-up PR, so this current one has a few TODO's that need to be resolved after that happens.

To demo it working, I rigged up a console log function as the closeModal prop and set `withinModal` to true. (Check the console for the `hi` messages at the bottom)

screen shot 2017-12-11 at 4 20 57 pm

### DIFF
--- a/resources/assets/components/Card/Card.js
+++ b/resources/assets/components/Card/Card.js
@@ -11,13 +11,28 @@ const renderHeader = (title, link, onClose) => (
   </header>
 );
 
-const Card = ({ children, className = '', link = null, title = null, onClose }) => (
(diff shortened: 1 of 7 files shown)


DFurnes:
We're gonna need to retrofit some tests now that you're relying on your ModalBeacon context here.

itsjoekent:
See: https://github.com/DoSomething/phoenix-next/pull/559#issuecomment-355591541

PR #561, opened 2017-12-11
Adding posts to Redux store along with cleanups!

weerd:
### What does this PR do?
I'm working on a few things with the new API routes in Phoenix, and wanted to do a bit of cleanup for the store to start laying the foundation for using more middleware before/after actions are dispatched to the Redux store.

So this PR does a bit of `store.js` cleanup. It moves the `initialState` object out of the store to it's own file for easier editing, and also moves the custom middlewares being pushed to list of middleware into a separate file as well since this list will likely grow and be easier to manage in a separate file; both files are imported into the `/store/store.js` file.

It also adds a `post` property to the store and corresponding action/reducer files with no logic and an `apiMiddleware` that will also all play a role in an upcoming PR as I tie it all back together 🎗

Refs #562 

Fixes #

### DIFF
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -1,0 +1,9 @@
+/**
+ * Fetch posts for the specified campaign.
+ *
+ * @param  {String} $campaignId
+ * @return {[type]}
+ */
+export function fetchCampaignPosts($campaignId) {
+  console.log($campaignId);
+}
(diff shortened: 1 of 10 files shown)


resources/assets/actions/post.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Fetch posts for the specified campaign.
+ *
+ * @param  {String} $campaignId
+ * @return {[type]}
+ */
+export function fetchCampaignPosts($campaignId) {
+  console.log($campaignId);
```
DFurnes: 👾

weerd: hehe, isn't called yet ;)

PR #563, opened 2017-12-12
adding logging to the rb controller

itsjoekent:
### What does this PR do?
adds temporary logging to the reportback controller so we can diagnose an issue with RB's failing to upload


itsjoekent:
`[2017-12-12 15:43:40] local.INFO: RB Response: [420612]`

Issue #562, opened 2017-12-11
[Feature] Phoenix talks directly to Rogue for RB Posts activity


weerd:
_FEATURE OVERVIEW_
-----------------------

#### User Story
As a developer, I want to fully transition to using the new Phoenix API that talks directly to Rogue for all front-end functionality related to Reportback Posts.

#### Additional Information
Also want to take this opportunity to really harness the use of middleware with the Redux store and cleanup up a few items.

#### Why This Matters
We want to completely cut the cord with Phoenix-Ashes and connect directly with Rogue when it comes to Reportbacks data on the Phoenix platform.

#### Tasks
This feature includes the following tasks:
- [x] Cleanup the `store` for easily updating initial state and middleware.
- [ ] ...

PR #566, opened 2017-12-12
Cleaning up env vars and exposing a few more needed for API work.

weerd:
### What does this PR do?
This PR removes the env vars set up via Webpack which are a bit obscure in how they get implemented on build and harder to debug. Also, we totally stopped using them anywhere in the app in favor of adding env vars to the `window.ENV` object! So I moved/added more in there and got rid of the unused code and `dotenv` package from the app.

Refs https://github.com/DoSomething/phoenix-next/issues/562

### DIFF
--- a/resources/assets/constants/index.js
+++ b/resources/assets/constants/index.js
@@ -1,3 +1,5 @@
+/* global window */
+
 /**
  * Contains general constants for the application.
  */
--- a/resources/assets/constants/index.js
+++ b/resources/assets/constants/index.js
@@ -1,4 +1,5 @@
 /* global window */
+import { get } from 'lodash';
 
 /**
  * Contains general constants for the application.
@@ -6,4 +7,4 @@
 
 export const MEDIA_MEDIUM_SIZE_MIN = 759;
 
-export const PHOENIX_URL = window.ENV.PHOENIX_URL;
+export const PHOENIX_URL = get(window.ENV, 'PHOENIX_URL', null);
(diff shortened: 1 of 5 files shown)


resources/assets/constants/index.js
```diff
@@ -4,5 +4,4 @@
 
 export const MEDIA_MEDIUM_SIZE_MIN = 759;
 
-// @TODO will be updating to pull from ENV. Only temporary!
-export const PHOENIX_BASE_URL = '//phoenix.dev';
+export const PHOENIX_URL = window.ENV.PHOENIX_URL;
```
weerd: Told ya it was only temporary! 😏

resources/assets/constants/index.js
```diff
@@ -1,8 +1,10 @@
...
+
 /**
  * Contains general constants for the application.
  */
 
 export const MEDIA_MEDIUM_SIZE_MIN = 759;
 
-// @TODO will be updating to pull from ENV. Only temporary!
-export const PHOENIX_BASE_URL = '//phoenix.dev';
+export const PHOENIX_URL = get(window.ENV, 'PHOENIX_URL', null);
```
weerd: @itsjoekent @mendelB I had to use `get()` here since the `Navigation` component test we're failing. The `TabbedNavigation` uses the `MEDIA_MEDIUM_SIZE_MIN`, and I guess when it loads this file it also tries to set the value for `PHOENIX_URL` and since it's in the Jest test env, the `window` object does not exist!

PR #567, opened 2017-12-13
Blurb is the word

mendelB:
### What does this PR do?
- Changing the `blurb` property (used in Mosaic Lede Banner) to not be required.
~- Adjusting some of the padding so that it remains consistent with or without the blurb~ (maintaining blurb styling with or without the blurb for now.)

We don't want to require the blurb anymore. We want to be able to just have the info on the pitch page even for Mosaic campaigns.

https://www.pivotaltracker.com/story/show/153638522

### Checklist

### DIFF
--- a/resources/assets/components/LedeBanner/templates/MosaicTemplate.js
+++ b/resources/assets/components/LedeBanner/templates/MosaicTemplate.js
@@ -46,7 +46,7 @@ const MosaicTemplate = (props) => {
             <h2 className="lede-banner__headline-subtitle">{subtitle}</h2>
           </div>
 
-          <Markdown className="lede-banner__blurb">{blurb}</Markdown>
+          { blurb ? <Markdown className="lede-banner__blurb">{blurb}</Markdown> : null }
 
           { isAffiliated ? null : <SignupButton /> }
         </div>
--- a/resources/assets/components/LedeBanner/templates/MosaicTemplate.js
+++ b/resources/assets/components/LedeBanner/templates/MosaicTemplate.js
@@ -57,7 +57,7 @@ const MosaicTemplate = (props) => {
 
 MosaicTemplate.propTypes = {
   actionText: PropTypes.string.isRequired,
-  blurb: PropTypes.string.isRequired,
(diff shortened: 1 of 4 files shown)


lkpttn:
@mendelB I've added some size information to this card: https://www.pivotaltracker.com/n/projects/2019313/stories/153448673/comments/183657676

mendelB:
Thanks @lkpttn I'll tackle that in a separate PR. This is just removing the blurb field requirement so the Lede can work independently. Unless you're saying that we're entirely removing the blurb as opposed to just making it optional?

resources/assets/components/LedeBanner/templates/MosaicTemplate.js
```diff
@@ -46,7 +46,9 @@ const MosaicTemplate = (props) => {
             <h2 className="lede-banner__headline-subtitle">{subtitle}</h2>
           </div>
 
-          <Markdown className="lede-banner__blurb">{blurb}</Markdown>
+          <div className="lede-banner__blurb">
+            { blurb ? <Markdown>{blurb}</Markdown> : null }
+          </div>
```
weerd: If the blurb is `null`, then we still output an empty `lede-banner__blurb` div? That's not ideal, can you pop the div info along w/ the `Markdown` component and output together if blurb is set otherwise nothing is output.

weerd: Does the reasoning for this relate to: 

>(maintaining blurb styling with or without the blurb for now.)

mendelB: Right. I was keeping this for now rather than do something too extravagant to maintain the padding from the blurb with or without it. As removing that padding pushes the button too close to the header when the blurb is missing. Was thinking to work in something better along with the new mosaic lede banner changes upcoming.

weerd: Ah, ok. Yeah, link that upcoming PR to this one so we know it's related ;)

PR #569, opened 2017-12-13
Using only campaignLead email for contact link in footer

mendelB:
### What does this PR do?
Only using the `campaignLead` email for the footer contact link instead of name. 

image

image

Should we be removing the name property in general from the footer prop hierarchy?
 (`CampaignPage` and `Footer` both define a `campaignLead` object as a prop with *both* name and email. IS this now unnecessary?)

https://www.pivotaltracker.com/story/show/153558695

### DIFF
--- a/resources/assets/components/CampaignFooter/CampaignFooter.js
+++ b/resources/assets/components/CampaignFooter/CampaignFooter.js
@@ -3,21 +3,20 @@ import PropTypes from 'prop-types';
 
 import AffiliateCredits from '../AffiliateCredits';
 
-const CampaignFooterContact = ({ email, name }) => (
-  <div className="info-bar__secondary">
-    Questions? <a
-      href={`mailto:${email || CampaignFooterContact.defaultProps.email}`}
-    >Contact {name || CampaignFooterContact.defaultProps.name}</a>
-  </div>
-);
+const CampaignFooterContact = ({ email }) => {
+  const contactEmail = email || CampaignFooterContact.defaultProps.email;
+  return (
+    <div className="info-bar__secondary">
+      Questions? <a href={`mailto:${contactEmail}`}>Contact {contactEmail}</a>
+    </div>
+  );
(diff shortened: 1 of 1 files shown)


resources/assets/components/CampaignFooter/CampaignFooter.js
```diff
@@ -3,21 +3,20 @@ import PropTypes from 'prop-types';
...
 
-const CampaignFooterContact = ({ email, name }) => (
-  <div className="info-bar__secondary">
-    Questions? <a
-      href={`mailto:${email || CampaignFooterContact.defaultProps.email}`}
-    >Contact {name || CampaignFooterContact.defaultProps.name}</a>
-  </div>
-);
+const CampaignFooterContact = ({ email }) => {
+  const contactEmail = email || CampaignFooterContact.defaultProps.email;
```
weerd: Is there a reason you need to specify `||` the default prop? Isn't the point of the default prop to fill in that var if it's undefined? Is that because `email` might come in as `null`?

mendelB: Yeah it was just in case `email` was passed in as `null`. (Which -after checking on this-, seems to only be possible if the 'contact' is from an actual `Staff` entity and not `additionalContent`, and that contentful staff item did not have email set.)
https://github.com/DoSomething/phoenix-next/blob/dev/app/Entities/Campaign.php#L157

weerd: gotcha!

weerd:
>Should we be removing the name property in general from the footer prop hierarchy?
(CampaignPage and Footer both define a campaignLead object as a prop with both name and email. IS this now unnecessary?)

I would say as long as it's not required it's fine to keep in `CampaignPage`. The `campaignLead` should defs have the name on there, because I could see that being useful in the future.

PR #570, opened 2017-12-13
Adding helpful error messaging for failed media uploads

itsjoekent:
### What does this PR do?
screen shot 2017-12-13 at 3 11 59 pm

This PR tries to give users additional context as to why the photo might have failed to upload.

If we want to bump up how large a photo we accept, we'll need to set that in the server configs. This is just to give more clarity to the user.

### DIFF
--- a/resources/assets/components/ReportbackUploader/ReportbackUploader.js
+++ b/resources/assets/components/ReportbackUploader/ReportbackUploader.js
@@ -133,13 +133,18 @@ class ReportbackUploader extends React.Component {
       </div>
     );
 
+    const modifiedSubmissionsMessaging = { ...submissions.messaging };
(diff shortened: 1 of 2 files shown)


resources/assets/components/ReportbackUploader/ReportbackUploader.js
```diff
@@ -133,13 +133,18 @@ class ReportbackUploader extends React.Component {
       </div>
     );
 
+    const modifiedSubmissionsMessaging = { ...submissions.messaging };
+    if (modifiedSubmissionsMessaging.error && modifiedSubmissionsMessaging.error.fields.media) {
+      modifiedSubmissionsMessaging.error.fields.media.push('Try uploading a smaller file.');
+    }
+
```
weerd: Is there a reason we can just [modify the default Laravel validation message](https://laravel.com/docs/5.5/validation#custom-error-messages)?

PR #572, opened 2017-12-15
Activating the ApiMiddleware.

weerd:
### What does this PR do?
Hoooboy! This PR activates the use of a generic `ApiMiddleware` layer where hopefully as we move forward we transition to using this middleware with any API calls (eventually we'll account for both internal (PN) api requests AND external requests to other service, but went the simpler route for now). I'd also like for this to be the pattern moving forward when it makes sense to use a middleware so we can keep _actions_ more succinct.

All this does for now is make a call to PN API which calls Rogue and grabs RB Posts for a campaign. If not on production, then it outputs nothing, otherwise it'll output a table of the response data.

Middleware in Redux are great because they allow you to manipulate the action code and can run before _and_ after an action is passed to the store. You can even block an action from being passed if it doesn't meet a certain requirement.

> In real-life applications, middleware are often where most of the logic and complex code resides. They also hold most of the utility functionality, such as logging, error reporting, analytics, authorization, and more.They are commonly used to hold the most complicated and generic parts of an application. They have access to the current action, to the store, and to the `dispatch()` method, giving them more power than any other part of Redux. -- The Complete Redux Book

Refs #562

### DIFF
--- a/resources/assets/actions/api.js
+++ b/resources/assets/actions/api.js
@@ -1,0 +1,9 @@
+/**
+ * API request action creator
+ * @param  {Object} data
+ * @return {Object}
+ */
+export default function apiRequest(data) {
+  // @TODO: check if data is object and has type 'API'
+  return data;
+}
--- a/resources/assets/actions/api.js
+++ b/resources/assets/actions/api.js
@@ -5,7 +5,7 @@ import { API } from '../constants/action-types';
  * @param  {Object} data
  * @return {Object}
  */
-export default function apiRequest({ method, payload }) {
+export default function apiRequest(method, payload) {
   // @TODO: check if supplied arguments are ok!
   return ({
     type: API,
--- a/resources/assets/actions/api.js
+++ b/resources/assets/actions/api.js
@@ -1,9 +1,15 @@
+import { API } from '../constants/action-types';
+
 /**
  * API request action creator
  * @param  {Object} data
  * @return {Object}
  */
-export default function apiRequest(data) {
-  // @TODO: check if data is object and has type 'API'
-  return data;
+export default function apiRequest({ method, payload }) {
+  // @TODO: check if supplied arguments are ok!
+  return ({
+    type: API,
+    method,
+    payload,
+  });
 }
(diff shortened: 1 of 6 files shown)


resources/assets/actions/post.js
```diff
@@ -1,11 +1,30 @@
...
-  console.log($campaignId);
+export function fetchCampaignPosts() {
+  return (dispatch, getState) => {
+    const id = getState().campaign.id;
+    const legacyId = getState().campaign.legacyCampaignId;
+    // @TODO: for now prefer legacy ID over contentful ID
+    const campaignId = legacyId || id;
+
+    dispatch(apiRequest({
+      type: 'API',
```
DFurnes: Should this be using the constant from `constants/action-types`? Also, would it make sense for the action creator to set this on the action, and just have the `apiRequest` creator take a method, url, and payload?

weerd: oop! Yeah, vestige from when I was playing around with the code. Meant to import the constant!

As for the second question, I think I dig that. I only really created it before because the `fetchCampaignPosts` had to return a function to then get caught by the thunk middleware and allow passing in the `getState()` and also `dispatch()`... initially I didn't want to have to dispatch again, but couldn't help it. But with your suggestion it give the `apiRequest` more reason to exist and would be a bit more useful then :)

weerd: I'll update on Monday so we can merge and add to deploys 💃

resources/assets/actions/api.js
```diff
@@ -1,9 +1,15 @@
...
+
 /**
  * API request action creator
  * @param  {Object} data
  * @return {Object}
  */
-export default function apiRequest(data) {
-  // @TODO: check if data is object and has type 'API'
-  return data;
+export default function apiRequest({ method, payload }) {
```
DFurnes: How would you feel about making these into two distinct parameters `apiRequest(method, payload)`?

weerd: yeah, we can do that :)

PR #571, opened 2017-12-14
Fixing the RB uploader layout

itsjoekent:
### What does this PR do?
This PR adds the dual column layout back to the RB uploader. Not a 100% replica of Lukes mock yet but it's getting there.

screen shot 2017-12-14 at 10 53 33 am
screen shot 2017-12-14 at 10 49 37 am
screen shot 2017-12-14 at 10 50 45 am
screen shot 2017-12-14 at 10 50 11 am

### Any background context you want to provide?
- I mostly moved the existing form into a new variable & wrapped the various inputs with FlexCells.
- Had to update the media uploader centering so it worked in this new layout config

### What are the relevant tickets/cards?
https://www.pivotaltracker.com/story/show/149011395

PR #574, opened 2017-12-18
Attach the onSubmit handler to the button

itsjoekent:
### What does this PR do?
Fixes the upload button!

In [this](https://github.com/DoSomething/phoenix-next/pull/571/files#diff-201edebec1634a925cb7e8371eba4809R137) PR the form became a sibling of the submit button, so the form on submit no longer has anything to trigger it. This PR attaches the submit to the form.

### DIFF
--- a/resources/assets/components/ReportbackUploader/ReportbackUploader.js
+++ b/resources/assets/components/ReportbackUploader/ReportbackUploader.js
@@ -134,7 +134,7 @@ class ReportbackUploader extends React.Component {
     );
 
     const reportbackUploader = (
-      <form className="reportback-form" ref={form => (this.form = form)}>
+      <form className="reportback-form" onSubmit={this.handleOnSubmitForm} ref={form => (this.form = form)}>
         <Flex>
           <FlexCell width="full">
             { shouldDisplaySubmissionMessaging ? (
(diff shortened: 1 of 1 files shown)


resources/assets/components/ReportbackUploader/ReportbackUploader.js
```diff
@@ -134,7 +134,7 @@ class ReportbackUploader extends React.Component {
     );
 
     const reportbackUploader = (
-      <form className="reportback-form" onSubmit={this.handleOnSubmitForm} ref={form => (this.form = form)}>
```
mendelB: I don't know if this really matters, but do we care about removing the ability to submit the form with the 'Return'/'Enter' button?

itsjoekent: mmm, good point we could still leave the onSubmit handler here

weerd: Hmm, that's a good question...

itsjoekent:
@mendelB so the enter button works on this PR now, but you have to hit enter on anything besides the text area (Like you mentioned in Slack)

PR #575, opened 2017-12-19
Adding voter reg placeholder

itsjoekent:
### What does this PR do?
Adds a placeholder voter reg component with the URL format we'll need to use. For now it just prints the query on the page.

- The Entity is pretty blank (intentionally) because the actual frontend display isn't fully defined. But since we know the query we went ahead and did that.
- We do know, however, that voter reg is going to be used all over the place. so we went ahead and made that a new action type in contentful

https://www.pivotaltracker.com/story/show/153743799

### DIFF
--- a/resources/assets/components/ActionPage/ActionStepRenderers.js
+++ b/resources/assets/components/ActionPage/ActionStepRenderers.js
@@ -2,6 +2,7 @@ import React from 'react';
 import ActionStep from './ActionStep';
 import Revealer from '../Revealer';
 import SignupButtonFactory from '../SignupButton';
+import VoterRegistrationContainer from '../VoterRegistration';
 import { FlexCell } from '../Flex';
 import { PostGalleryContainer } from '../Gallery/PostGallery';
 import { ThirdPartyActionContainer } from '../Actions/ThirdPartyAction';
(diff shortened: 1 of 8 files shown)


resources/assets/components/ActionPage/ActionStepRenderers.js
```diff
@@ -163,3 +164,14 @@ export function renderLegacyGallery() {
...
 }
+
+/**
+ * Render the voter registration container.
+ *
+ * @return {Component}
+ */
+export function renderVoterRegistration() {
+  return (
+    <VoterRegistrationContainer key="voter-reg" />
```
mendelB: Is the `key` because it can be rendered from within a list of actions?

resources/assets/components/VoterRegistration/VoterRegistrationContainer.js
```diff
@@ -0,0 +1,15 @@
+import { connect } from 'react-redux';
+import VoterRegistration from './VoterRegistration';
+
+/**
+ * Provide state from the Redux store as props for this component.
+ */
+const mapStateToProps = state => ({
+  userId: state.user.id,
```
mendelB: I'm never sure if it's important to consistently use this [selector](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/selectors/user.js#L7) for fetching user ID? (I know as of now we're somewhat inconsistent about it)

resources/assets/components/ActionPage/ActionStepRenderers.js
itsjoekent: Yessir

resources/assets/components/VoterRegistration/VoterRegistrationContainer.js
itsjoekent: I personally don't find it particularly useful since it's such an easy property to reach, but I don't think you're wrong to suggest it either. /shrug

PR #576, opened 2017-12-19
Friend referral

mendelB:
### What does this PR do?

- Adds support for a 'referralRB' toggle in a campaigns additional content, which would add some specific 'info' fields to the reportback uploader form for the refer a friend feature.

You can test this on `test-teens-for-jeans` campaign.

Quoting explanation that was here prior:

"in the RB Uploader, in addition to the fixtures of the Media field and caption,
we have these -what we can call- "Info Fields", which consist of an optional "Quantity field" and a "Why Participated" field. (The quantity is more of a fixture because of how it always needs to be part of the RB package for now, so I mainly mean the 'why participated').

"It seems that all we're really doing with the Refer A Friend is adding some fields to replace the 'Why Participated', hence, re-doing everything in the RB felt extraneous.

"So I ended up adding support for a 'friendReferralRB' toggle field from contentful (additionalContent in the campaign), and based on this setting, the RB will render those appropriate fields in the form instead of the why_participated. (next step would be adding support for a different form submission prop if we need to submit this info to a new endpoint so we can save the data to our PN temp table).

### Old Screenshots
[LG](https://user-images.githubusercontent.com/12417657/34174156-051bf9ba-e4c6-11e7-89c8-389078544065.png), [MD](https://user-images.githubusercontent.com/12417657/34174178-1631c248-e4c6-11e7-8ec5-c518fe822472.png), [SM](https://user-images.githubusercontent.com/12417657/34174198-21c3ae32-e4c6-11e7-9e07-f2d9f7bf1046.png)

https://www.pivotaltracker.com/story/show/153297498

### Checklist
- [x] Added screenshot to PR description of related front-end updates on **small** screens.
- [x] Added screenshot to PR description of related front-end updates on **medium** screens.
- [x] Added screenshot to PR description of related front-end updates on **large** screens.

### DIFF
--- a/resources/assets/actions/reportback.js
+++ b/resources/assets/actions/reportback.js
@@ -163,11 +163,9 @@ export function toggleReportbackAffirmation(shouldShowAffirmation) {
 }
 
 // Async Action: submit a new reportback and place in submissions gallery.
-export function submitReportback(reportback, isReferralRB) {
+export function submitReportback(url, reportback) {
   return (dispatch) => {
     dispatch(storeReportback(reportback));
-
-    const url = `${window.location.origin}/next/${isReferralRB ? 'referrals' : 'reportbacks'}`;
 
     const token = document.querySelector('meta[name="csrf-token"]');
 
@@ -199,6 +197,16 @@ export function submitReportback(reportback, isReferralRB) {
       })
       .catch(error => console.log(error));
   };
+}
+
+export function submitReferralPost(post) {
+  const url = `${window.location.origin}/next/referrals`;
+  return submitReportback(url, post);
+}
+
+export function submitPhotoPost(post) {
+  const url = `${window.location.origin}/next/reportbacks`;
+  return submitReportback(url, post);
 }
 
 export function fetchUserReportbacks(userId, campaignId) {
--- a/resources/assets/actions/reportback.js
+++ b/resources/assets/actions/reportback.js
@@ -163,11 +163,11 @@ export function toggleReportbackAffirmation(shouldShowAffirmation) {
 }
 
 // Async Action: submit a new reportback and place in submissions gallery.
-export function submitReportback(reportback) {
+export function submitReportback(reportback, isReferralRB) {
   return (dispatch) => {
     dispatch(storeReportback(reportback));
 
-    const url = `${window.location.origin}/next/reportbacks`;
+    const url = `${window.location.origin}/next/${isReferralRB ? 'referrals' : 'reportbacks'}`;
 
     const token = document.querySelector('meta[name="csrf-token"]');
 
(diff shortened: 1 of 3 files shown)


itsjoekent:
I think it works to get something out the door but we def. need to re-think this entire thing at some point

mendelB:
For sure. I anticipate that once we restructure the RB uploader as a whole, it'll be way easier to fit this in in a cleaner more sustainable manner.

weerd:
I think getting all the pieces in place is ideal first, since I'll be going in at some point after the holidays and refactoring a lot of this and RB uploader stuff to talk directly with Rogue.

mendelB:
OK, @DoSomething/team-rocket I think this is ready for review. Part one of implementing Refer a Friend

mendelB:
UPDATE: Fixed the merge conflicts, and added working submission functionality for Refer a Friend RB's.

weerd:
@mendelB can you update the description and screenshots used with the latest? or add another comment w/ the updates and screenshots and reference it in the main description!

mendelB:
# Update:

- Resolved merge conflicts resulting from the revamp of the `ReportbackUploader`
- Added extra param to the `actions/reportback#submitReportback` method to indicate a referral reportback, which will send the submission to the proper endpoint.

Otherwise, it's all pretty much the same as it was before. Here are the updated screenshots based on the new look:

LG:
image

MD:
image

SM:
image

resources/assets/actions/reportback.js
```diff
@@ -163,11 +163,11 @@ export function toggleReportbackAffirmation(shouldShowAffirmation) {
 }
 
 // Async Action: submit a new reportback and place in submissions gallery.
-export function submitReportback(reportback) {
+export function submitReportback(reportback, isReferralRB) {
```
weerd: I'm debating if in the effort to individualize each "Action", like we're doing with the content types in Contentful, we make unique actions here as well. So a `submitReferralPost()`. And we could eventually rename this to `submitReportbackPost()` or `submitPhotoPost()`... Thoughts? 🤔

mendelB: Yeah, that'd definitely line up with how we're doing it on the content side! I guess for now that'd just mean keeping the endpoint stuff separate from the actual submission, since from [line 71](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/actions/reportback.js#L171) and on they're identical?

weerd: Yeah, it might mean separating that out. Something that we could also cleanup once submitting directly to Rogue

mendelB: @weerd I added an idea for how this might look with separate submission methods

mendelB: @weerd  @ashleybaldwin 
Can we merge this in as is, and move the work on separating the RB submission methods in a separate PR?

PR #577, opened 2017-12-19
adding migration for friend_referral table

mendelB:
### What does this PR do?
Opens the discussion regarding the data fields we're capturing for the 'Refer a Friend' Reportback.

Based the fields off of the pivotal card and the discussion in Slack
https://dosomething.slack.com/archives/C3ASB4204/p1513700247000116

https://www.pivotaltracker.com/story/show/153297514

### DIFF
--- a/database/migrations/2017_12_19_162611_create_friend_referral_table.php
+++ b/database/migrations/2017_12_19_162611_create_friend_referral_table.php
@@ -1,0 +1,36 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class CreateFriendReferralTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('friend_referrals', function (Blueprint $table) {
+            $table->increments('id');
+            $table->timestamps();
+            $table->string('friend_name');
+            $table->string('friend_email');
+            $table->text('friend_story');
+            $table->string('referrer_name');
+            $table->string('referrer_northstar_id');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::dropIfExists('friend_referrals');
+    }
+}


mendelB:
Maybe `referred_...` instead of `friend_...`?

database/migrations/2017_12_19_162611_create_friend_referrals_table.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class CreateFriendReferralsTable extends Migration
```
weerd: could we just call it the `referrals` table?

database/migrations/2017_12_19_162611_create_referrals_table.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::dropIfExists('friend_referrals');
```
weerd: need to rename this one too!

mendelB: yeesh, good catch! 😬

weerd:
Looks good! Do we also maybe want the `referrer_email` in case messaging wants to let them know they contacted their friend or something? 🤷‍♂️  @ngjo

DFurnes:
The referrer's name and email should be stored in Northstar, right?

mendelB:
Yeah, the stuff I have so far for the controller is fetching the Northstar User for obtaining the referrer name, so can def grab email too!

ngjo:
@weerd Sorry so I'm clear, the data we're gonna have (not necessarily the terms you'll use):

- Member who referred Northstar
- Member who referred email
- Member who referred name
- Member why participated (which in this case is the "why is ____ the best?"
- Referral Name
- Referral Email

Would it have a timestamp of the when the member submitted the RB - it's been useful for other exports (i.e. opt-in stuff). If that's easy to include, I would recommend that!

Also, it's probably not worth doing in this iteration, but just throwing it out there to get folks ~ marinating ~ on it, I could see a world where campaign leads would want to see if the referral is already a DS member or not.

mendelB:
@ngjo Cool! We'll have the timestamp for when the row was inserted into the database, which could serve as the timestamp for RB as well. Also, It wouldn't be difficult to do a search by email to see if the referred party is already a member. 

In terms of the why participated, so we're saving that in our database / CSV as the 'story', should we also be passing that along to Rogue (along with the usual data we've been passing along with RBs) as the 'Why Participated'? (This would make sense since as of now, I figured we'd just be passing in some placeholder for why participated, as it's a required field).

weerd:
@DFurnes for now since we didn't know where the ideal place for this was (could be argued it belongs in Rogue too since technically this is an action of referring a friend) we decided on a temp table for the time being in PN.

weerd:
> Member why participated (which in this case is the "why is ____ the best?"

@ngjo @mendelB after some discussions with @lkpttn we realized that the we can't use the `why_participated` as the field to hold this information because in current model if a user updates their RB "why participated" field it overwrites the data with the new info. But for the refer a friend, if a member refers more than one person we don't want to override each friend's story!

ngjo:
@mendelB that'd be great if we could see if the referral is a member or not! if it's at all a lift though, let's hold off on it and can reconsider post-launch of this.

@weerd that makes total sense - no need to do the why then! I was under the assumption that the `why participated` was going to be hacked to `why do you like your friend` but if it's staying the same, which makes sense to me,  I don't think there's any need to include in the export!

DFurnes:
> @DFurnes for now since we didn't know where the ideal place for this was (could be argued it belongs in Rogue too since technically this is an action of referring a friend) we decided on a temp table for the time being in PN.

I'm into that. Specifically asking why we're including a copy of referrer's name & email, since those are canonically stored in Northstar. We should just be storing `referrer_northstar_id`.

weerd:
> I'm into that. Specifically asking why we're including a copy of referrer's name & email, since those are canonically stored in Northstar. We should just be storing referrer_northstar_id.

I figured it would help with messaging since we're going to need to export a CSV of this data to pass along to messaging team, having that info available rather than collecting it all again (either on CSV generation, or by the messaging team) just seemed like it'd be useful...

DFurnes:
If we're handing this off to a partner in a CSV, we should be fetching those fields when generating the CSV (like we do with [CSV exports in Rogue](https://github.com/DoSomething/rogue/blob/master/app/Services/ExportService.php). If we're using this for messaging through Customer.io, all of a user's profile information is already accessible by Northstar ID.

ngjo:
From the discussions I've been in, at least for this campaign, this referral information is just for internal use for messaging in c.io!

weerd:
Ya, ok cool. We can remove this and just have the NS ID. It was more a suggestion from me to make the info available and accessible for other teams post export and less overhead when exporting. But I'm fine with not having it in during this temporary approach :) @mendelB sorry for the confusion. Mind removing the email and name for the referrer and just keep the NS ID?

mendelB:
@weerd For sure, not a problem! Do we want the check for if they're a member by email, or leave that out as well?

weerd:
@mendelB leave out for now. That can happen on CSV generation or something. But not necessary for this PR.

PR #579, opened 2017-12-20
Friend referral backend

mendelB:
### What does this PR do?
Adds route and controller action to handle reportback submissions of the 'Refer a Friend' variety.

I put the action in the reportback controller as this felt like a regular reportback at the end of the day. The `friendReferral` action takes care of validating and storing the fields we need locally in our temp table, and then passes the work on to the regular `store` action.

- We're also validating the regular RB fields in the `friendReferral` action since otherwise if for example *all* fields are blank, it'll only send back error messages for the friend referral specific ones. But for DRYness, would it be worth throwing the repeated validations in a class variable, and using that variable in both places?

- There's some static placeholder text for the `whyParticipated` field, as it's a required field in Rogue, but not needed for this type of RB.

EDIT: Added a `ReferralController` for storing referrals for more CRUDy controller setup. To prevent duplicating the regular RB storage logic, setup a new `storeReportback` method in the `PostRepository` and using that in the ReportbackController and ReferralController.

EDIT 2:
- Added a `UploadedMedia` service to DRY up storing and deleting the RB images.
- Moved the Reportback posting to the respective controllers.

#576  #577 
[#153297514](https://www.pivotaltracker.com/story/show/153297514)
[#153297498](https://www.pivotaltracker.com/story/show/153297498)

### DIFF
--- a/app/Http/Controllers/ReferralController.php
+++ b/app/Http/Controllers/ReferralController.php
@@ -70,7 +70,6 @@ class ReferralController extends Controller
 
         Log::info('RB Response:', $response);
 
-        // Delete the uploaded file.
         UploadedMedia::delete($path);
 
         return $response;
--- a/app/Http/Controllers/ReferralController.php
+++ b/app/Http/Controllers/ReferralController.php
@@ -5,17 +5,11 @@ namespace App\Http\Controllers;
 use App\Models\Referral;
 use Illuminate\Http\Request;
 use App\Services\PhoenixLegacy;
-use App\Repositories\PostRepository;
+use Illuminate\Support\Facades\Log;
+use App\Exceptions\InvalidFileUploadException;
 
 class ReferralController extends Controller
 {
-    /**
-     * The post repository.
-     *
-     * @var PostRepository
-     */
-    private $postRepository;
-
     /**
      * The legacy Phoenix API.
      *
@@ -26,9 +20,8 @@ class ReferralController extends Controller
     /**
      * ReferralController constructor.
      */
-    public function __construct(PostRepository $postRepository, PhoenixLegacy $phoenixLegacy)
+    public function __construct(PhoenixLegacy $phoenixLegacy)
     {
-        $this->postRepository = $postRepository;
         $this->phoenixLegacy = $phoenixLegacy;
 
         $this->middleware('auth')->only(['store']);
@@ -61,6 +54,34 @@ class ReferralController extends Controller
         // Static 'why participated' to pass validation.
         $request['whyParticipated'] = 'Refer-A-Friend';
 
-        return $this->postRepository->storeReportback($this->phoenixLegacy, $request);
+        $reportbackPhoto = $request->file('media');
+
+        if (! $reportbackPhoto->isValid()) {
+            throw new InvalidFileUploadException;
+        }
+
+        // Store the uploaded file.
+        $path = '/uploads/'.$reportbackPhoto->store('images', 'uploads');
+
+        $path = $this->uploadedMedia->store($request->file('media'));
+
+        $response = $this->phoenixLegacy->storeReportback(
+            auth()->id(),
+            $request->input('campaignId'),
+            [
+                'file_url' => config('app.env') !== 'local' ? config('app.url').'/next'.$path : 'https://placeimg.com/1000/768/animals',
+                'caption' => $request->input('caption'),
+                'quantity' => $request->input('impact'),
+                'why_participated' => $request->input('whyParticipated'),
+                'source' => 'phoenix-next',
+            ]
+        );
+
+        Log::info('RB Response:', $response);
+
+        // Delete the uploaded file.
+        app('files')->delete(public_path($path));
+
+        return $response;
     }
 }
--- a/app/Http/Controllers/ReferralController.php
+++ b/app/Http/Controllers/ReferralController.php
@@ -5,8 +5,8 @@ namespace App\Http\Controllers;
 use App\Models\Referral;
 use Illuminate\Http\Request;
 use App\Services\PhoenixLegacy;
+use App\Services\UploadedMedia;
 use Illuminate\Support\Facades\Log;
-use App\Exceptions\InvalidFileUploadException;
 
 class ReferralController extends Controller
 {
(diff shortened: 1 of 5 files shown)


weerd:
>I put the action in the reportback controller as this felt like a regular reportback at the end of the day. 

Totally get this logic, but something we learned about during a [Adam Wathan's conf talk](https://www.youtube.com/watch?v=MF0jFKvS4SI) at the last Laracon was to reduce complexity in controllers and make more controllers with the fixed set of expected CRUD actions instead. It was a pretty great talk!

So lets do a `FriendReferralController` controller class and use the typical `store()` action etc for this :)

Bonus is if we need to remove this down the line we just delete a controller file instead of having to clean up the `ReportbackController`!

mendelB:
Cool! I totally dig that, and that looks like a great talk.
Could we still send this off to use the `ReportbackController`s `store` action for the regular RB stuff, or should we be copying that logic over in this `FriendReferralController`?

mendelB:
@weerd Added the new `ReferralController` and `PostRepository` method

app/Http/Controllers/ReferralController.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /**
+     * ReferralController constructor.
+     */
+    public function __construct(PostRepository $postRepository, PhoenixLegacy $phoenixLegacy)
+    {
+        $this->phoenixLegacy = $phoenixLegacy;
+        $this->postRepository = $postRepository;
+
+        $this->middleware('auth', ['only' => ['store']]);
```
weerd: Lets switch to using the [method chaining options](https://github.com/illuminate/routing/blob/master/ControllerMiddlewareOptions.php#L31) which are a little more readable:

```php
$this->middleware('auth')->only(['store']);
```
You can actually pass a string or array :)

app/Repositories/PostRepository.php
```diff
@@ -45,4 +47,41 @@ public function getCampaignPosts($id, $query = [])
...
         return $this->rogue->get('v3/posts', $query);
     }
+
+    /**
+     * Send reportback to Rogue for storage.
+     *
+     * @param \App\Services\PhoenixLegacy $client
+     * @param \Illuminate\Http\Request    $request
+     */
+    public function storeReportback($client, $request)
```
weerd: I think I may have misinterpreted some stuff when discussing this part with you. Lets chat IRL when you have a moment because I'd like to not have to include all this logic in the repository, or at least simplify it. 🤔

app/Http/Controllers/ReferralController.php
```diff
@@ -0,0 +1,52 @@
...
+namespace App\Http\Controllers;
+
+use App\Models\Referral;
+use Illuminate\Http\Request;
+use App\Services\PhoenixLegacy;
+use App\Repositories\PostRepository;
+
+class ReferralController extends Controller
+{
+    /**
```
weerd: Add `phoenixLegacy` and `postRepository` as private props up here.

app/Repositories/PostRepository.php
weerd: OK @mendelB. Let's try calling this `storeReferralPost()` to specify what kind of post action we're specifically storing. However we want to aim to keep these methods in the repository short and sweet. Let's also not pass the `PhoenixLegacy` in as a param but include it via the constructor (ideally we'll remove this soon and just use the Rogue client).
Then for lines 59-66 lets see if we can create a `Services/UploadedMedia` class to DRY that logic up and then use it in the `ReportbackController` as well. There could be `store()` and `delete()` methods on the class, and the `store()` method returns the path to us for use in the API call in `$client->storeReportback()`.

Eventually, I'd like to remove the use of `PhoenixLegacy` to store RBs and switch to the `RogueClient` but it's ok for now. However, I could still see the `UploadedMedia` class still applying after the switcheroo ;)

Lemme know if you have any follow up questions!

mendelB: So the `ReportbackController` should *not* be making use of this `storeReferralPost` method (hence the specific naming), and should instead implement posting the storage on its own, (and as you mentioned utilizing the `UploadedMedia` service)?

mendelB: Meaning - both the `Referral` and `Reportback` `store` methods at the end of the day need to make use of the same logic to actually store the plain reportback (minus the special referral fields) to rogue. 
So should each controller still implement that on its own, or isn't that the intended usage of the `storeReportback` method in the repository?

weerd: Lets leave the `ReportbackController` alone for now. I want to move towards having a respective controller for each type of action (granted it's a little wonky here since both the regular action and RAF action have a photo upload component). We could try and wrap more actions into a single controller but that's gonna get out of control pretty quick.

So yeah, just implement the use of the `UploadedMedia` class in the `ReportbackController` to DRY up that aspect of the code, but otherwise, leave as is for now and don't import the `PostRepository` into it just yet.

mendelB: Sounds good 👍 
In that case though, do we even need to make use of the `PostRepository` to begin with or can we just implement the storage in the referral controller as well - same as RB controller?

weerd: Yeah, sounds good. Ignore using the `PostRepository` for now. We can resort to using that only for when dealing w/ `RogueClient` and keep it at that.

mendelB:
@weerd OK, this should all be as you specified. I hope I did properly!

app/Http/Controllers/ReferralController.php
```diff
@@ -0,0 +1,78 @@
...
+            'friend_name' => $request->input('friendName'),
+            'friend_email' => $request->input('friendEmail'),
+            'friend_story' => $request->input('friendStory'),
+            'referrer_northstar_id' => auth()->id(),
+        ]);
+
+        // Static 'why participated' to pass validation.
+        $request['whyParticipated'] = 'Refer-A-Friend';
+
+        $path = UploadedMedia::store($request->file('media'));
```
weerd: Noice. I dig it :) Glad it worked out alright!

app/Http/Controllers/ReportbackController.php
```diff
@@ -47,15 +54,9 @@ public function store(Request $request)
             'whyParticipated' => 'required',
         ]);
 
-        $reportbackPhoto = $request->file('media');
-
-        if (! $reportbackPhoto->isValid()) {
-            throw new InvalidFileUploadException;
-        }
-
+        return get_class($request->file('media'));
```
weerd: Was this left in from debugging?

mendelB: -yelp- Indeed it was! good ole doc block investigation

PR #581, opened 2017-12-22
Progress on RB Post Photo Uploader updates.

weerd:
### What does this PR do?
This PR does a bunch of updates to the Reportback Post Photo Uploader to address some new visual design updates and provide some bug fixes.

### Any background context you want to provide?
Sure. There's been a new design for the RB uploader and while some of the ideas have been implemented, this PR adds the rest and does a bit of cleanup.

### Responsive Sizes
Here's what the new design looks like across the different responsive sizes:

#### Small
screen shot 2018-01-02 at 8 43 47 am

#### Medium
screen shot 2018-01-02 at 8 44 26 am

#### Large
screen shot 2018-01-02 at 8 44 53 am

### Image Sizes In Uploader
And it also account for any sized images that are uploaded, not just square ones:
screen shot 2018-01-02 at 8 45 55 am

screen shot 2018-01-02 at 8 46 22 am

screen shot 2018-01-02 at 8 46 43 am

### Validations
And here it is with the field validations as setup by @itsjoekent but with some design updates:
screen shot 2018-01-04 at 9 37 14 am

_Note: I did find a bug with validation where now since we allow campaign leads to toggle showing the impact field and we default it to `1` if it's toggled off (since it's also required by Rogue at the moment) we never really validate the impact field since it will always have a value. I feel we can ignore this for now and reassess, especially once the switch to posting to Rogue happens and quantity on the post is a thing.

### What are the relevant tickets/cards?
Refs https://www.pivotaltracker.com/story/show/153757088

### Checklist
- [x] Added appropriate feature/unit tests.
- [x] Added screenshot to PR description of related front-end updates on **small** screens.
- [x] Added screenshot to PR description of related front-end updates on **medium** screens.
- [x] Added screenshot to PR description of related front-end updates on **large** screens.

PR #582, opened 2017-12-22
Voter reg action

mendelB:
### What does this PR do?
Turns the `VoterRegistration` component into a richer 'Action Step' action.

- adding markup based on the regular action step component. 
- using the typical action step properties needed. (As of now this is coming from `additionalContent`, these fields need to be added to the `VoterRegistrationAction` Entity)
- added the `additionalContent` field to `VoterRegistrationAction` Entity
- logic for building a URL with dynamic query params based on addtionalContent fields. (Should be turned into actual fields and models in contentful for a 'DynamicLink')

### AdditionalContent example
image

[#153853897](https://www.pivotaltracker.com/story/show/153853897)
[#153743799](https://www.pivotaltracker.com/story/show/153743799)

image

### DIFF
--- a/resources/assets/components/ActionPage/ActionStepRenderers.js
+++ b/resources/assets/components/ActionPage/ActionStepRenderers.js
@@ -170,8 +170,17 @@ export function renderLegacyGallery() {
  *
  * @return {Component}
  */
-export function renderVoterRegistration() {
+export function renderVoterRegistration(step, stepIndex) {
+  const { content, title, template, dynamicLink } = step.fields.additionalContent;
+
   return (
-    <VoterRegistrationContainer key="voter-reg" />
+    <VoterRegistrationContainer
+      key="voter-reg"
+      content={content}
+      title={title}
+      template={template}
+      stepIndex={stepIndex}
+      dynamicLink={dynamicLink}
+    />
   );
 }
--- a/resources/assets/components/ActionPage/ActionStepRenderers.js
+++ b/resources/assets/components/ActionPage/ActionStepRenderers.js
@@ -175,13 +175,17 @@ export function renderVoterRegistration(step, stepIndex) {
   const { template, dynamicLink } = additionalContent;
 
   return (
-    <VoterRegistrationContainer
-      key="voter-reg"
-      content={content}
-      title={title}
-      template={template}
-      stepIndex={stepIndex}
-      dynamicLink={dynamicLink}
-    />
+    <FlexCell>
+      <div className="action-step">
+        <VoterRegistrationContainer
+          key="voter-reg"
+          content={content}
+          title={title}
+          template={template}
+          stepIndex={stepIndex}
+          dynamicLink={dynamicLink}
+        />
+      </div>
+    </FlexCell>
   );
 }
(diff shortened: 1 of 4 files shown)


mendelB:
One important thing here:
Using the `makeUrl` helper which uses the JS `URL` constructor, URL encode the query etc and replaces the `:` with `%3A`.

Would this be a problem for turboVote, and should we skip using that for a turboVote type dynamic link?

resources/assets/components/ActionPage/ActionStepRenderers.js
```diff
@@ -170,8 +170,17 @@ export function renderLegacyGallery() {
  *
  * @return {Component}
  */
-export function renderVoterRegistration() {
+export function renderVoterRegistration(step, stepIndex) {
+  const { content, title, template, dynamicLink } = step.fields.additionalContent;
```
weerd: We know we want to use `title` and `content`... want to just go ahead and add those to the content model in Contentful and then pop into the entity, etc?

weerd:
> Would this be a problem for turboVote, and should we skip using that for a turboVote type dynamic link?

I think since the TurboVote link is super unique lets handle it differently to avoid issues.

resources/assets/components/VoterRegistration/VoterRegistration.js
```diff
@@ -1,21 +1,71 @@
...
+    link = makeUrl(baseUrl, query).href;
+  }
 
-  // Ignore the weird query format and the mismatching names. Wasn't our decision.
-  const query = `?r=user:${userId},campaign:${campaignRunId},source:web`;
 
   return (
-    <p className="margin-horizontal-md"><a href={join('/', query)}>TurboVote Link example</a></p>
+    <FlexCell width="full">
+      <div className={classnames('action-step')}>
```
weerd: So one thing I'm trying out in the [RB Uploader updates](https://github.com/DoSomething/phoenix-next/pull/581/files#diff-d7fddee1d3d6de8741222e368afcdea5R47) is to remove the scaffolding that is in the component for how action steps need to render. Mind removing the `<FlexCell width="full">` and `<div className={classnames('action-step')}>` out of the component and into the `ActionStepsRenderer`.

The main idea being, that if we want the `VoterRegistration` component rendered elsewhere on site, there's no reason it should be in a div with a class of `action-step` or necessarily in a `FlexCell` component with a `width` prop of `full`!

mendelB: Ah nice, I love that!

mendelB:
Only thing is that the header for this voter registration action won't look too good if this is rendered from anywhere else as of now since the styling for the step header is nested under `action-step` https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/ActionPage/actionPage.scss
Just something we'd need to bear in mind.

weerd:
> Only thing is that the header for this voter registration action won't look too good if this is rendered from anywhere else as of now since the styling for the step header is nested under action-step

Yep yep. I think it'll be something that as we work towards universalizing content to fall outside of just action steps we'll take some strides to make this better :)

PR #583, opened 2018-01-02
Action page root route

mendelB:
### What does this PR do?

Renders the `action` page from the root route under `CampaignPage` `/us/campaigns/my-cool-campaign`

- added dedicated `/community` route.
- [retained](https://github.com/DoSomething/phoenix-next/commit/8604bd16b5fc8c391b2d3a4593a7e1cc82f8f32a#diff-def96fba41962ab39482b9b2baa8aedbR35) fallback logic in case campaign is closed - and there is a Feed - we'll render the community tab (or `Feed`) instead.

NEW:
[e6618d6](https://github.com/DoSomething/phoenix-next/pull/583/commits/e6618d6f74aa78f29fa02d79360fcd9b0c4d04ef)
- mosaic template campaigns with empty fields will now also hide the community tab as per [#154056394](https://www.pivotaltracker.com/story/show/154056394) (I tacked this on to this PR since it's kind of contingent on not having the community tab as the root of the campaign page)

One caveat here is that the `action` tab won't be 'active' (or highlighted) unless the route explicitly matches `.../action`, meaning it won't actually be highlighted on the root page. (With community, we had only rendered the page from the root route, so we'd [defined](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/helpers/navigation.js#L6) the path as such. Here though, we wanted the root, as well as `/action` to render action, so we had to choose one of those for tab highlight activation, and went with `/action`).

https://www.pivotaltracker.com/story/show/153639716
https://www.pivotaltracker.com/story/show/154056394

### DIFF
--- a/resources/assets/components/Navigation/TabbedNavigationContainer.js
+++ b/resources/assets/components/Navigation/TabbedNavigationContainer.js
@@ -59,7 +59,7 @@ const TabbedNavigationContainer = (props) => {
     <Button className="-inline nav-button" onClick={() => clickedSignUp(legacyCampaignId)} text={actionText} />
   ), 'tabbed navigation', { text: actionText });
 
-  const shouldHideCommunity = (template === 'legacy') && ! hasActivityFeed;
+  const shouldHideCommunity = ! hasActivityFeed;
   const shouldHideAction = (isClosed || (shouldHideCommunity && additionalPages.length === 0));
 
   return (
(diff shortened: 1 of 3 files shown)


mendelB:
@joe I didn't go too crazy with it for now, but I added a couple of comments to hopefully lend at least some clarity to the weirder bits of routing logic

PR #585, opened 2018-01-03
Pitch landing page cleanup

mendelB:
### What does this PR do?

- Removing the Sixpack experiment for landing (pitch) pages 📦 📦 📦 📦 📦 📦  
- Removing the legacy landing (pitch) page, and associated fields and functions ✈️ 🛄 📟 

Not much going on here, just removing the experiment and the old landing page components, and crowning the `LandingPageContentAlt` as the `LandingPageContent` 👑 

EDIT:
Adding a `templates` folder and moving the `LandingPageContent` to a `PitchTemplate` component, as per Diego's idea below:

"So an interesting thought for this could be that the LandingPage component directory contains a templates directory and instead of rendering a LandingPageContent component it actually references a LandingPage/templates/PitchTemplate.js which is basically what is the current LandingPageContent but sets up an understanding of the hierarchy 😉"

# BEFORE MERGING 
- [ ] We must ensure all campaigns on Contentful that have a Landing Page, have the correct format to support the new Landing Page. (I believe @ashleybaldwin already took care of this?) (Is it pitch or landing page?? this has driven me crazy)

[#154013212](https://www.pivotaltracker.com/story/show/154013212)

### DIFF
--- a/app/Entities/Page.php
+++ b/app/Entities/Page.php
@@ -38,7 +38,6 @@ class Page extends Entity implements JsonSerializable
             'fields' => [
                 'title' => $this->title,
                 'content' => $this->content,
-                'legacyContent' => array_get($this->additionalContent, 'pitchContent', null),
                 'sidebar' => $this->parseSidebar($this->sidebar),
             ],
         ];
--- a/resources/assets/components/Page/LandingPage/LandingPage.js
+++ b/resources/assets/components/Page/LandingPage/LandingPage.js
@@ -5,7 +5,7 @@ import PropTypes from 'prop-types';
 
 import Enclosure from '../../Enclosure';
 import LedeBanner from '../../LedeBanner/LedeBanner';
-import LandingPageContent from './LandingPageContent';
+import PitchTemplate from './templates/PitchTemplate';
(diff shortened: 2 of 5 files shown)


weerd:
> Is it pitch or landing page?? this has driven me crazy

Pitch Landing Page 😏 

Pitch is a _type_ of **Landing Page**! We could eventually have _other_ types!

weerd:
So an interesting thought for this could be that the `LandingPage` component directory contains a `templates` directory and instead of rendering a `LandingPageContent` component it actually references a `LandingPage/templates/PitchTemplate.js` which is basically what is the current `LandingPageContent` but sets up an understanding of the hierarchy 😉

mendelB:
Yes! That would certainly help clarify the `Pitch` vs `Landing` Page situation.

weerd:
@mendelB try out the above suggestion for this PR and see how it goes :)

mendelB:
@weerd changed the layout as per your suggestion.

mendelB:
This one just needs a final review of the new `PitchTemplate` changes based on @weerd suggestion.

PR #586, opened 2018-01-04
Fixing logic for impact field.

weerd:
### What does this PR do?
While testing I realized I left some logic in to always show the impact field while I was working on the design. This update should properly adjust the logic so it only shows if specified in the Contentful `PhotoUploaderAction` toggle.

🏂


itsjoekent:
im surprised the linter didnt catch that 🤔

weerd:
Yeah... 🤔

PR #587, opened 2018-01-04
Remove slideshow close button

itsjoekent:
### What does this PR do?
in the updated modal designs Luke has the extra "close" button on the bottom no longer a thing, the only ways to close a modal are clicking anywhere on the screen outside the content, or the X button in the top right corner (which will soon only exist for modals that are not using a Card)


itsjoekent:
Correct, a prop with no value defaults to true

weerd:
The slide button is just a one-way thing?  User can't go back to prev slide right?

itsjoekent:
@weerd as of now they can't, but could always be something we add!

PR #589, opened 2018-01-05
Fixing a minor modal bug

mendelB:
### What does this PR do?

- adding a `lock` param to the `toggleChromeLock` helper method, to be able to specify locking/unlocking chrome.
- returning `NotFound` component in a `Card` from the `ContentModal` when no content exists (e.g. faulty contentful ID)

The bug is addressed in the [Pivotal Card](https://www.pivotaltracker.com/story/show/154087267). Essentially what's happening is that the - linked to - modal is attempting to load - it locks the screen and messes with the toggle setup for the affirmation modal which actually loads, but then locks the screen again when you dismiss it. So this adds a specification param so you can manually toggle adding or removing the chrome lock.

The other this addresses bad modal links such as an invalid contentful ID, where previously this would silently error out, and thus in this unaffiliated user scenario - post signup they wouldn't get to see the affirmation.

[#154087267](https://www.pivotaltracker.com/story/show/154087267)

### DIFF
--- a/resources/assets/actions/modal.js
+++ b/resources/assets/actions/modal.js
@@ -2,7 +2,7 @@ import { OPEN_MODAL, CLOSE_MODAL } from '../actions';
 import { toggleChromeLock } from '../helpers';
 
 export function openModal(modalType, contentfulId) {
-  toggleChromeLock();
+  toggleChromeLock(true);
   return { type: OPEN_MODAL, modalType, contentfulId };
 }
 
--- a/resources/assets/actions/modal.js
+++ b/resources/assets/actions/modal.js
@@ -1,12 +1,9 @@
 import { OPEN_MODAL, CLOSE_MODAL } from '../actions';
-import { toggleChromeLock } from '../helpers';
 
 export function openModal(modalType, contentfulId) {
-  toggleChromeLock(true);
   return { type: OPEN_MODAL, modalType, contentfulId };
 }
 
 export function closeModal() {
-  toggleChromeLock();
   return { type: CLOSE_MODAL };
 }
--- a/resources/assets/components/Modal/Modal.js
+++ b/resources/assets/components/Modal/Modal.js
@@ -1,8 +1,9 @@
+/* global document */
+
 import React from 'react';
 import PropTypes from 'prop-types';
 import Portal from 'react-portal';
 
-import { toggleChromeLock } from '../../helpers';
 import './modal.scss';
 
 class Modal extends React.Component {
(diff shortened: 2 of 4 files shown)


resources/assets/actions/modal.js
```diff
@@ -2,7 +2,7 @@ import { OPEN_MODAL, CLOSE_MODAL } from '../actions';
 import { toggleChromeLock } from '../helpers';
 
 export function openModal(modalType, contentfulId) {
-  toggleChromeLock();
+  toggleChromeLock(true);
```
DFurnes: Can we handle this with a [`onOpen` and `onClose`](https://github.com/tajo/react-portal/blob/master/READMEv3.MD) handlers [on the portal](https://github.com/DoSomething/phoenix-next/blob/a8cc5795e2677289dd683bc43126d3425f4038b9/resources/assets/components/Modal/Modal.js#L42) instead of imperatively toggling it here?

itsjoekent:
Just want to point out that there are two seperate issues being described here,

> The other this addresses bad modal links such as an invalid contentful ID, where previously this would silently error out, and thus in this unaffiliated user scenario - post signup they wouldn't get to see the affirmation.

The affirmation is contained within the `POST_SIGNUP_MODAL`, and links to blocks are in the `CONTENT_MODAL`. The code you changed would not fix an affirmation block missing from the page (if that was the intent)

mendelB:
@itsjoekent sorry, the explanation here a bit of a mish-mash and unclear. I actually was addressing the case of a missing content block (which could result from a bad modal ID in the URL) as opposed to a missing affirmation block.

itsjoekent:
okie dokes!

mendelB:
Since the content modal would error out, the affirmation modal - although it would be a valid and existing one, would fail to launch as well

resources/assets/helpers/index.js
```diff
@@ -373,10 +373,11 @@ export function showTwitterSharePrompt(href, quote) {
 /**
  * Toggle the lock on the app chrome.
  */
-export function toggleChromeLock() {
-  document.getElementById('chrome').classList.toggle('-lock');
-}
+export function toggleChromeLock(lock) {
+  const classList = document.getElementById('chrome').classList;
 
+  return lock ? classList.add('-lock') : classList.remove('-lock');
```
itsjoekent: can we change this to something more clear? eg: `isLocked`

mendelB: Or maybe `shouldLock`? (we're indicating intent as opposed to current state)

resources/assets/components/Modal/Modal.js
```diff
@@ -39,7 +41,12 @@ class Modal extends React.Component {
     const { shouldShowModal, children } = this.props;
 
     return (
-      <Portal closeOnEsc isOpened={shouldShowModal}>
+      <Portal
+        closeOnEsc
+        isOpened={shouldShowModal}
+        onOpen={() => toggleChromeLock(true)}
```
mendelB: Technically, we can not explicitly pass `true` and rely on the dom node being passed in as an argument (which seems to be the default and) which would be truthy as well, but I thought this would be more explicit and clear.

DFurnes: Can we inline the actual calls into this file? I don't think we should have a general helper function for this.

```jsx
const chrome = document.getElementById('chrome');

<Portal
  onOpen={() => chrome.classList.add('-lock')}
  onClose={() => chrome.classList.remove('-lock')}

```

mendelB: Sure thing!

PR #590, opened 2018-01-05
Action step coloring Proposition

mendelB:
### What does this PR do?

Here's my proposal for this Action Step Color testing business:

We have a general campaign setting for the 'theme color' (for now this is just in additionalContent), the PhotoHeader component can check the state for this color, and add it to its class list.
If a proper color is set, the CSS will override the default `background-color` with the specified one. This will also be the new grayscale for any images added to stepheaders

`test-teens-for-jeans` (purple)
image

`grab-mic` (purple)
image

If y'all think this is a good way to go, and we decide to indeed test this, we can then move this to a `PhotoHeaderAlt` apply a sixpack test between that and the usual one. (And add whichever colors we so desire to the suite)

[#154013236](https://www.pivotaltracker.com/story/show/154013236)

### DIFF
--- a/resources/assets/components/ActionPage/templates/MosaicStepHeaderTemplate.js
+++ b/resources/assets/components/ActionPage/templates/MosaicStepHeaderTemplate.js
@@ -3,14 +3,14 @@ import PropTypes from 'prop-types';
 import { FlexCell } from '../../Flex';
 
 import { convertNumberToWord } from '../../../helpers';
-import PhotoHeader from '../../PhotoHeader';
+import { PhotoHeaderContainer } from '../../PhotoHeader';
 
 const MosaicStepHeaderTemplate = ({ title, step, background, hideStepNumber }) => (
   <FlexCell width="full">
-    <PhotoHeader className="action-step__header" backgroundImage={background}>
+    <PhotoHeaderContainer className="action-step__header" backgroundImage={background}>
       { hideStepNumber ? null : <span>step { convertNumberToWord(step) }</span> }
       <h1>{ title }</h1>
-    </PhotoHeader>
+    </PhotoHeaderContainer>
   </FlexCell>
 );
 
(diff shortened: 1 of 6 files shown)


itsjoekent:
sorry can we take a step back, maybe I'm misremembering the sprint plan

- Blank gray boxes are littering the action page and that wasn't the original design intent
- Creative team added solid color png's to fix this but diego does not want them doing that
- Quick solution was to give them control over this on the action page
- The massive request that's likely unattainable without hard data proving its value is restyling the entire template with certain colors per campaign

If this is all true, I would suggest not using `additionalContent` here and making an actual value in the campaign settings model because this is a must-have feature we need regardless for our design. Can product weigh in here?

weerd:
>We have a general campaign setting for the 'theme color'...

I don't necessarily mind the idea of eventually setting a theme color, but I think this shouldn't be a setting for a overall theme color, but a setting on the action step so each action step can have it's own color specified. This was kind of the idea Campaigns was going for with the uploaded solid color PNGs per each step header. I think popping in additional content for now is fine, but do it in the action step.

itsjoekent:
Can we not add more JSON values though? This feels like a feature that is actively warranted & needed and the campaigns team does not enjoy using the json stuff

mendelB:
Ok, seems I misunderstood the intention here. So no tests. But an actual field within the Action Step to specify the coloring of the step header?

weerd:
@mendelB no actual field yet.

This should be a test, but a Sixpack test we can measure and see if changing the color has any traction.

weerd:
It actually doesn't need to have an `additionalContent` JSON setting if it's a Sixpack test. Lets just pass a predefined color for step header tests and call it a day. Wait till we see activity.

mendelB:
Ok, so one hardcoded color for the `StepHeaderAlt` (or whatever we end up calling it). What would the conversion be for this test since they're presumably already signed up if they're on the action page, perhaps user reporting back?

mendelB:
I'm gonna close this one out for now 😢

PR #592, opened 2018-01-08
Updating php unit tests

weerd:
### What does this PR do?
This PR does a bit of updating to our series of PHPUnit tests which were collecting dust. I've marked any tests that require requests to Contentful as `Incomplete` since we would rather figure out a better way to mock those than deal with requesting to Contentful for each test run.

I've also updated the Wercker file so it will run the tests when code is pushed up and it tries to build the app.

This cleanup was required because I'm adding a test or two to #591 and all the prior tests were failing 😬 

Testing! 🚥 

### Checklist

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -5,7 +5,7 @@ build:
       - leipert/composer-install@0.9.1
       - wercker/npm-install
       - script:
-          name: phpunit
+          name: run phpunit test suite
           code: |-
             php artisan key:generate
             php artisan gateway:key
--- a/wercker.yml
+++ b/wercker.yml
@@ -5,6 +5,12 @@ build:
       - leipert/composer-install@0.9.1
       - wercker/npm-install
       - script:
+          name: phpunit
+          code: |-
+            php artisan key:generate
+            php artisan gateway:key
+            vendor/bin/phpunit
+      - script:
           name: check code style
           code: npm run lint
       - script:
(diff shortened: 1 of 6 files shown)


wercker.yml
```diff
@@ -2,8 +2,17 @@ box: dosomething/ds-docker-php
 build:
     # The steps that will be executed on build
     steps:
+      - script:
+          name: configure environment
+          code: cp .env.example .env
       - leipert/composer-install@0.9.1
       - wercker/npm-install
+      - script:
```
weerd: I was debating if this script could actually come before the `npm-install` call since PHPUnit is really only dealing with the PHP code and if that fails what's the point of installing a bunch of node modules? 🤔

DFurnes: Unfortunately, Laravel's `elixir` helper throws an exception if the given asset (`app.js` or `app.css` can't be found), so I think we have to keep this near the end. 😩

weerd: Ah hokay!

weerd:
@DFurnes I think I might need to update to PHPUnit v6.0 since it seems to be ignoring the `markTestIncomplete()` on Wercker and running them anyways, which is why it's failing... 🤔

weerd:
Yeah, realized the Laravel upgrade is probs necessary as well :\ Ok, will comment out.

weerd:
@DFurnes Nevermind. Turns out I missed marking a test as incomplete!

DFurnes:
A ha! 🕵️

PR #593, opened 2018-01-08
Adding the sponsor to the lede banner

itsjoekent:
### What does this PR do?
- Adding the sponsor to the lede banner
- Shifting the arrow logic around a bit. In the next PR we can deal with the right angle arrow.

#### Large
screen shot 2018-01-08 at 1 54 26 pm

#### Medium
[see updated photo in comment](https://github.com/DoSomething/phoenix-next/pull/593#discussion_r160721634)

#### Small
screen shot 2018-01-08 at 1 53 58 pm

https://www.pivotaltracker.com/story/show/154037769

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
     "heroku-postbuild": "sh bootstrap/setup.sh"
   },
   "engines": {
-    "node": "6.x",
-    "npm": "4.x"
+    "node": "8.x",
+    "npm": "5.x"
   },
   "babel": {
     "presets": [
--- a/resources/assets/components/CampaignSignupArrow/CampaignSignupArrow.js
+++ b/resources/assets/components/CampaignSignupArrow/CampaignSignupArrow.js
@@ -4,7 +4,7 @@ import React from 'react';
 import './campaignSignupArrow.scss';
 
 const CampaignSignupArrow = ({ content, className }) => (
-  <div className={`message-callout -above -white ${className}`}>
+  <div className={`message-callout -below -white ${className}`}>
     <div className="message-callout__copy">
       <p>{content}</p>
     </div>
--- a/resources/assets/components/LedeBanner/lede-banner.scss
+++ b/resources/assets/components/LedeBanner/lede-banner.scss
@@ -16,13 +16,17 @@
     width: 100%;
     height: 70px;
 
-    @include media($medium) {
+    @include media($large) {
       width: 50%;
     }
   }
 
   .header__signup {
     padding-top: $base-spacing;
+
+    .button {
+      float: left;
+    }
   }
 }
 
@@ -97,5 +101,15 @@
 
   strong {
     color: $primary-color;
+  }
+}
+
+.lede-banner__arrow {
+  clear: both;
+  margin-top: $half-spacing;
+
+  @include media($larger) {
+    clear: none;
+    margin-top: 0;
   }
 }
--- a/resources/assets/components/LedeBanner/lede-banner.scss
+++ b/resources/assets/components/LedeBanner/lede-banner.scss
@@ -24,7 +24,7 @@
   .header__signup {
     padding-top: $base-spacing;
 
-    .button {
+    .button.-float {
       float: left;
     }
   }
--- a/resources/assets/components/LedeBanner/lede-banner.scss
+++ b/resources/assets/components/LedeBanner/lede-banner.scss
@@ -112,4 +112,14 @@
     clear: none;
     margin-top: 0;
   }
+
+  .message-callout.-white .message-callout__copy p {
+    font-size: 24px;
+  }
+}
+
+.lede-banner__sponsor {
+  .promotion .__copy, .promotion .__image {
+    max-width: 100%;
+  }
 }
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -13,8 +13,8 @@
     "heroku-postbuild": "sh bootstrap/setup.sh"
   },
   "engines": {
-    "node": "6.x",
-    "npm": "4.x"
+    "node": "8.x",
+    "npm": "5.x"
```
weerd: 🎉

resources/assets/components/CampaignSignupArrow/CampaignSignupArrow.js
```diff
@@ -4,7 +4,7 @@ import React from 'react';
 import './campaignSignupArrow.scss';
 
 const CampaignSignupArrow = ({ content, className }) => (
```
weerd: Not something I caught before but could we rename this component to `MessageCallout`? It's more than just an arrow 😬  Not necessary for this PR, but maybe do it in a quick follow-up PR?

resources/assets/components/LedeBanner/lede-banner.scss
```diff
@@ -16,13 +16,17 @@
     width: 100%;
     height: 70px;
 
-    @include media($medium) {
+    @include media($large) {
```
weerd: @lkpttn maybe this is more of a question for you... if we wait till Large size to switch to the 50/50 in the lede banner, the banner will take up a lot of vertical space. Could we just keep the 50/50 split on Medium size, but the join button, message callout and sponsor stay stacked vertically until Large? cc: @itsjoekent

resources/assets/components/CampaignSignupArrow/CampaignSignupArrow.js
itsjoekent: sure

package.json
mendelB: image
🎉

resources/assets/components/LedeBanner/lede-banner.scss
lkpttn: I just talked to Diego about this, he had a good suggestion about using 50% of the space in the banner at the medium size, but keeping the half/half layout. He can explain it more but I approved it!

weerd: @itsjoekent lets chat tomorrow about it!

package.json
mendelB: image

resources/assets/components/LedeBanner/lede-banner.scss
itsjoekent: screen shot 2018-01-10 at 11 00 44 am

here is the actual screenshot i should have uploaded instead for medium

package.json
weerd: escalated

resources/assets/components/LedeBanner/lede-banner.scss
```diff
@@ -16,13 +16,17 @@
...
       width: 50%;
     }
   }
 
   .header__signup {
     padding-top: $base-spacing;
+
+    .button.-float {
```
weerd: Probably don't need the modifier class when already targeting the `.button` from within the `.header__signup`. Should just be fine w/out `.-float`.

itsjoekent: the -float is intentional, if we apply the float when there isnt stuff under the button then the padding breaks and the button aligns with the bottom of the lede banner

resources/assets/components/LedeBanner/templates/MosaicTemplate.js
```diff
@@ -25,14 +28,32 @@ const MosaicTemplate = (props) => {
   };
 
   const signupArrowComponent = signupArrowContent ? (
-    <CampaignSignupArrow content={signupArrowContent} className="-mosaic-arrow" />
+    <div className="lede-banner__arrow">
```
weerd: Just curious, is this extra `div` and class necessary? Couldn't you just apply the styles to the `message-callout`?

resources/assets/components/LedeBanner/lede-banner.scss
weerd: Ah ok!

resources/assets/components/LedeBanner/templates/MosaicTemplate.js
itsjoekent: it doesnt seem to make any visual difference when i remove the div and apply the classname

resources/assets/components/LedeBanner/templates/MosaicTemplate.js
```diff
@@ -25,14 +28,30 @@ const MosaicTemplate = (props) => {
   };
 
   const signupArrowComponent = signupArrowContent ? (
-    <CampaignSignupArrow content={signupArrowContent} className="-mosaic-arrow" />
+    <CampaignSignupArrow content={signupArrowContent} className="lede-banner__arrow -mosaic-arrow" />
```
weerd: Also meant that without that extra `div`, you likely don't need the `lede-banner__arrow` class either and can just shift those styles over to the `message-callout` class, right?

PR #597, opened 2018-01-12
Moving sponsor out of the signup button component

itsjoekent:
### What does this PR do?
The sponsor logo should always appear, regardless of affiliation state.

screen shot 2018-01-12 at 10 12 20 am
screen shot 2018-01-12 at 10 12 14 am
screen shot 2018-01-12 at 10 12 05 am


weerd:
Post-merge 🌮

PR #601, opened 2018-01-16
changing NorthstarId dynamic url param to snake case

mendelB:
### What does this PR do?
Changes the `northstarId` special param within the dynamic URL logic we've built into the `ThirdPartAction` to be snake case (`northstar_id`).

[#154282288](https://www.pivotaltracker.com/story/show/154282288)

### DIFF
--- a/resources/assets/components/Actions/ThirdPartyAction/ThirdPartyAction.js
+++ b/resources/assets/components/Actions/ThirdPartyAction/ThirdPartyAction.js
@@ -15,7 +15,7 @@ const ThirdParyAction = (props) => {
   let contentLink = '';
 
   const urlParams = dynamicUrlParams.map((param) => {
-    const value = param === 'northstarId' ? userId : props[param];
+    const value = param === 'northstar_id' ? userId : props[param];
 
     return `${param}=${value}`;
   });


resources/assets/components/Actions/ThirdPartyAction/ThirdPartyAction.js
```diff
@@ -15,7 +15,7 @@ const ThirdParyAction = (props) => {
   let contentLink = '';
 
   const urlParams = dynamicUrlParams.map((param) => {
-    const value = param === 'northstarId' ? userId : props[param];
+    const value = param === 'northstar_id' ? userId : props[param];
```
itsjoekent: needs more comments. i want to see a novel.

mendelB:
Sorry, I should have clarified that. I updated the param in TypeForm as well to be snake case.

resources/assets/components/Actions/ThirdPartyAction/ThirdPartyAction.js
mendelB: I'm workin' on my novel, but it's of no `value` unfortunately.

PR #602, opened 2018-01-16
Initial progress on using TDD to drive out PN to Rogue API posting.

weerd:
### What does this PR do?
This PR begins using some TDD (test-driven development) to drive out the work for Phoenix posting reportback photos to Rogue via the `v3` endpoint on Rogue's end.

I marked a couple tests related to the uploading of media as incomplete with `Implement once we upgrade to Laravel 5.4.`. There's some nice stuff that came in 5.4 for helping to fake uploaded files, etc that makes testing said functionality easier. I'm debating on holding off on doing the testing for that until we upgrade, so we might be doing the upgrade sooner rather than later!

Refs https://www.pivotaltracker.com/story/show/154367578

### Checklist

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -1,10 +1,14 @@
 <?php
 
+use App\Exceptions\Handler;
 use App\Repositories\CampaignRepository;
+use DoSomething\Gateway\Testing\WithOAuthTokens;
+use Illuminate\Contracts\Debug\ExceptionHandler;
 
 abstract class TestCase extends Illuminate\Foundation\Testing\TestCase
 {
     use DatabaseSetup;
+    use WithOAuthTokens;
 
     /**
      * The base URL to use while testing the application.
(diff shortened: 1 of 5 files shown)


tests/TestCase.php
```diff
@@ -1,10 +1,14 @@
...
 
+use App\Exceptions\Handler;
 use App\Repositories\CampaignRepository;
+use DoSomething\Gateway\Testing\WithOAuthTokens;
+use Illuminate\Contracts\Debug\ExceptionHandler;
 
 abstract class TestCase extends Illuminate\Foundation\Testing\TestCase
 {
     use DatabaseSetup;
+    use WithOAuthTokens;
```
DFurnes: Totally subjective, but you can chain these together if you want!

```php
use DatabaseSetup, WithOAuthTokens;
```

tests/TestCase.php
```diff
@@ -27,6 +31,41 @@ public function createApplication()
         return $app;
     }
 
+    /**
+     * Helper method to assert validation errors based on error response structure.
+     *
+     * @param  string $field
+     */
+    protected function assertValidationError($field)
```
DFurnes: This is handy! 🙌

tests/TestCase.php
```diff
@@ -27,6 +31,41 @@ public function createApplication()
...
+    }
+
+    /**
+     * Disable exception handling in Laravel acceptance tests when need more
+     * verbose error information.
+     * https://gist.github.com/adamwathan/c9752f61102dc056d157
+     *
+     * @return StdClass
+     */
+    protected function disableExceptionHandling()
```
DFurnes: Clever! I assume it's something you'd pop in to dig up more detailed info on a test that's giving a 500? ⚠️

weerd: yar! Since Laravel takes over and Handles errors to deal with as HTTP responses. Something I learned from Adam 💃  In your test you drop `$this->disableExceptionHandling()` and basically creates new anonymous class and re-throws exception. Instead of having to re-throw manually in the `render()` in `App/Exceptions/Handler.php` and then forgetting to change it back!

PR #606, opened 2018-01-17
Adding the share component logic

itsjoekent:
### What does this PR do?
This PR adds
- the additionalContent field to the share action
- injecting the field within the action steps render process
- the logic that handles mapping the JSON into share links

Contentful JSON
screen shot 2018-01-17 at 3 15 02 pm

Component rendered
screen shot 2018-01-17 at 3 15 13 pm

Puck events
screen shot 2018-01-17 at 3 15 57 pm

https://www.pivotaltracker.com/story/show/154008960

### DIFF
--- a/app/Entities/ShareAction.php
+++ b/app/Entities/ShareAction.php
@@ -16,7 +16,9 @@ class ShareAction extends Entity implements JsonSerializable
         return [
             'id' => $this->entry->getId(),
             'type' => $this->entry->getContentType(),
-            'fields' => [],
+            'fields' => [
+                'additionalContent' => $this->additionalContent,
+            ],
         ];
     }
 }
--- a/resources/assets/components/ShareAction/ShareAction.js
+++ b/resources/assets/components/ShareAction/ShareAction.js
@@ -24,7 +24,7 @@ const ShareAction = (props) => {
   };
 
   return (
-    <div className="share-action margin-bottom-lg">
+    <div className="share-action margin-horizontal-md margin-bottom-lg">
       <ul>
         <li><a role="button" tabIndex="0" onClick={onFacebookClick}>share this</a></li>
         <li><a role="button" tabIndex="0" onClick={onFacebookClick}>hi luke</a></li>
--- a/resources/assets/components/ShareAction/ShareAction.js
+++ b/resources/assets/components/ShareAction/ShareAction.js
@@ -1,17 +1,13 @@
-/* global window */
-
-import PropTypes from 'prop-types';
 import React from 'react';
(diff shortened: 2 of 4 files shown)


resources/assets/components/ShareAction/ShareAction.js
```diff
@@ -1,17 +1,13 @@
...
+import PropTypes from 'prop-types';
 import { showFacebookSharePrompt } from '../../helpers';
 import './share-action.scss';
 
 const ShareAction = (props) => {
-  const { trackEvent } = props;
-
-  const link = window.location.href;
-  const trackingData = { link };
+  const { additionalContent, trackEvent } = props;
```
mendelB: Is the `trackEvent` actually being passed in from anywhere yet? I'm not seeing it in the container. I may just be confused cuz you do have an example puck log in the PR description.

resources/assets/components/ShareAction/ShareAction.js
```diff
@@ -23,19 +19,38 @@ const ShareAction = (props) => {
     });
   };
 
+  const hasLinks = additionalContent && additionalContent.links;
```
mendelB: Not using lodash `get` 😱

app/Entities/ShareAction.php
```diff
@@ -16,7 +16,9 @@ public function jsonSerialize()
         return [
             'id' => $this->entry->getId(),
             'type' => $this->entry->getContentType(),
-            'fields' => [],
+            'fields' => [
+                'additionalContent' => $this->additionalContent,
```
mendelB: We'll be changing this to a list of some sort of Link content type or something at some point I imagine?

itsjoekent: sure hope so! I'm just waiting on more concrete plans from product

resources/assets/components/ShareAction/ShareAction.js
itsjoekent: https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/ShareAction/ShareActionContainer.js

resources/assets/components/ShareAction/ShareAction.js
itsjoekent: I don't think its really necessary to import and use lodash for this

resources/assets/components/ShareAction/ShareAction.js
mendelB: Ahh ok, I actually took a look at the `PuckConnector` and I see it passing that prop through. Cool!

PR #608, opened 2018-01-18
Adding referrals table export command

mendelB:
### What does this PR do?
~Adding a `ReferralsExport` command to export the `referrals` table as a CSV.~

~There's also some logic in there to use a progress bar to visually display the status of the export.~

~- The CSV is being created and saved in the root of the project, would this be an issue in production? Might we want to have the command email the CSV someplace or something?~

~- Note that this will be exporting *all* rows every time the command is run, so this does need some way to query for only the non-exported rows.
https://dosomething.slack.com/archives/C3ASB4204/p1516287628001076~
https://dosomething.slack.com/archives/C3ASB4204/p1516289910000154

- Added an `exported` boolean column to the referrals table 
- Only exporting the referrals that have a `false` value for `exported` column, and updating appropriately
- Tested with 100,000 dummy rows, it works!

UPDATE:

- Adding an action in the `ReferralsController` to export the (necessary columns from the) referrals table as a CSV
- Adding route to hit this action (Though technically this probably should be a `patch/put` method since it's not idempotent, Diego advised to work with GET for this presumabely short-lived feature)
- Adding button in the campaigns admin dashboad to trigger the download. This button will only appear for campaigns that have the `referralsRB` specified as `true` in `additionalContent` and only for users with an `admin` role. Also added a confirmation alert, since this action is consequential for the database and should be used responsibly.

[#154360555](https://www.pivotaltracker.com/story/show/154360555)


mendelB:
OK, updated (added new stuff to description) and should be ready for review.

app/Console/Commands/ExportReferrals.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public function handle()
+    {
+        $referralsEloquentBuilder = Referral::where('exported', false);
+
+        $referrals = $referralsEloquentBuilder->get();
+
+        $bar = $this->output->createProgressBar($referrals->count());
+
+        $file = fopen('referrals.csv', 'w');
```
DFurnes: Where does this write to? Should we have an output path as one of the command arguments?

app/Console/Commands/ExportReferrals.php
```diff
@@ -0,0 +1,72 @@
...
+
+        $columns = ['id', 'created_at', 'friend_name', 'friend_email', 'friend_story', 'referrer_northstar_id'];
+
+        fputcsv($file, $columns);
```
DFurnes: Old school! 🤓 We also use the [`league/csv`](https://csv.thephpleague.com/9.0/) package is some other apps. It provides some nicer wrappers around these built-in functions (for example, so you don't have to call `fopen` and `fclose` by hand).

mendelB: Oh nice! Think it's worth using that package here instead?

weerd:
> There's also some logic in there to use a progress bar to visually display the status of the export.

Nice! I've always wanted to use the progress bar in command line for something 💃

app/Console/Commands/ExportReferrals.php
DFurnes: I think this is probably fine for now, but could save some time on future scripts!

app/Console/Commands/ExportReferrals.php
mendelB: It writes to the root of the project. I think that does make sense though. Will add!

weerd: Let's add a timestamp or some date to the name of the file so as multiple files are exported a user can keep track of when this happens.

weerd:
So something to consider for how this will be used. Someone has to SSH into the Prod environment and run this command... and then they also need to figure out how to download the file from the root directory on the Prod server to their local computer to then upload to Customer.io?

Which can be done, just confirming this is the intended workflow...

DFurnes:
> So something to consider for how this will be used. Someone has to SSH into the Prod environment and run this command... and then they also need to figure out how to download the file from the root directory on the Prod server to their local computer to then upload to Customer.io?
>
> Which can be done, just confirming this is the intended workflow...

That's a good point. It'd be nice to either toss this on S3 from the Artisan command (especially since  we won't have a reliable local filesystem after we move to Heroku) or maybe make an admin-only API route that lets people download this file from the admin menu.

mendelB:
Gonna revisit this in the AM to address @DFurnes points here.

mendelB:
Description has been updated. Chatted with @weerd @DFurnes over a Moscow Mule or two. They advised to proceed with a controller level CSV export logic, that can be called from the admin dashboard.

For review on Monday!

app/Http/Controllers/ReferralController.php
```diff
@@ -24,7 +25,7 @@ public function __construct(PhoenixLegacy $phoenixLegacy)
     {
         $this->phoenixLegacy = $phoenixLegacy;
 
-        $this->middleware('auth')->only(['store']);
+        $this->middleware('auth');
```
mendelB: Figuring that we probably want to authenticate the `csvExport` function as well.

app/Http/Controllers/ReferralController.php
```diff
@@ -74,4 +75,44 @@ public function store(Request $request)
...
+                    $referral->referrer_northstar_id,
+                ]);
+            }
+
+            fclose($file);
+
+            $referralsEloquentBuilder->update(['exported' => true]);
+        };
+
+        return response()->stream($callback, 200, $headers);
```
mendelB: I didn't end up using the php league [CSV package](https://csv.thephpleague.com) since I couldn't figure out how to stream the response as we're doing here. 
Meaning it would write all the rows to the CSV before actually returning the response, and this could take a lot of time if there are many rows to write, which could cause the response to time out.

resources/assets/components/AdminDashboard/CampaignDashboard/CampaignDashboard.js
```diff
@@ -1,17 +1,24 @@
...
   const { hasLandingPage, slug, clickedShowAffirmation, clickedShowLandingPage,
-    clickedShowActionPage, clickedRemoveSignUp, signupCreated, isSignedUp,
-    toggleReportbackAffirmation,
+    clickedShowActionPage, clickedRemoveSignUp, hasReferralRB, signupCreated,
+    isAdmin, isSignedUp, toggleReportbackAffirmation,
   } = props;
 
   const onSignUpClick = () => (! isSignedUp ? signupCreated() : clickedRemoveSignUp());
 
+  const onReferralExportClick = () => {
```
mendelB: The effect of clicking this button is kinda grave, since we'd be updating all exported rows as 'exported', ensuring they would not be exported again. So I felt like it might be important to diverge from the ole Hawaii missile alert approach and confirm this.

resources/assets/components/AdminDashboard/CampaignDashboard/CampaignDashboard.js
```diff
@@ -34,20 +41,27 @@ const CampaignDashboard = (props) => {
       <button className="button -secondary margin-md" onClick={() => toggleReportbackAffirmation(true)}>
         Show Reportback Affirmation
       </button>
+      { hasReferralRB && isAdmin ?
```
mendelB: Also felt like we might want to limit this ability to admins, otherwise any staff member might mistakenly click this button since we all love clicking buttons.

app/Http/Controllers/ReferralController.php
weerd: Not keen on having so much stuff going on in the Controller. Could we abstract this out a little. Not necessarily a whole dedicated class, but maybe create a `generate_streamed_csv()` helper function or something and pass in the necessary items?

resources/assets/components/AdminDashboard/CampaignDashboard/CampaignDashboard.js
```diff
@@ -1,17 +1,24 @@
...
-    clickedShowActionPage, clickedRemoveSignUp, signupCreated, isSignedUp,
-    toggleReportbackAffirmation,
+    clickedShowActionPage, clickedRemoveSignUp, hasReferralRB, signupCreated,
+    isAdmin, isSignedUp, toggleReportbackAffirmation,
   } = props;
 
   const onSignUpClick = () => (! isSignedUp ? signupCreated() : clickedRemoveSignUp());
 
+  const onReferralExportClick = () => {
+    const message = 'Please confirm your intent to export this data. This will permanently mark the exported records and cannot be undone.';
```
weerd: Maybe re-phrase "mark the exported records" to "mark the records as already exported"... does that seem any clearer? 🤔

app/Http/Controllers/ReferralController.php
mendelB: Yeah, that makes sense! I'll try and turn it into something more elegant.

mendelB:
@weerd Added a helper method for generating the actual CSV, was this along the lines of what you had in mind?

app/Http/Controllers/ReferralController.php
```diff
@@ -74,4 +75,28 @@ public function store(Request $request)
 
         return $response;
     }
+
+    public function csvExport()
+    {
+        $fileName = 'referrals_export_'.Carbon::now().'.csv';
```
DFurnes: Can we check the user's role here before continuing? We wouldn't want a normal user stumbling on this.

PR #611, opened 2018-01-18
Making the global modal exit button a conditional display

itsjoekent:
### What does this PR do?
Part of the modal design QA is moving the X button from the top right corner to the specific content in the modal. However, not all modals are treated equal, so this PR does some conditional magic as well.

screen shot 2018-01-18 at 3 44 24 pm
screen shot 2018-01-18 at 3 44 18 pm
screen shot 2018-01-18 at 3 44 12 pm

And here is a non-modal Card in the community feed to demonstrate the X not showing in this setting,
screen shot 2018-01-18 at 3 48 05 pm

I figured doing a pass-through of the `closeModal` function was the easiest way to accomplish this, but I'm definitely not convinced this is the most _proper_ solution. Refactoring the Card component to be container based, and updating all of those containers could be a PR in itself, so just did this, for now, to get the job done.

### DIFF
--- a/resources/assets/components/CampaignUpdate/CampaignUpdate.js
+++ b/resources/assets/components/CampaignUpdate/CampaignUpdate.js
@@ -9,13 +9,18 @@ import Byline from '../Byline';
 import Markdown from '../Markdown';
 import { ShareContainer } from '../Share';
 
-const CampaignUpdate = ({ id, author, content, link, shareLink, bordered, titleLink }) => {
+const CampaignUpdate = (props) => {
+  const {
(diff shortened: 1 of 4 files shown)


itsjoekent:
For reference, this was a much more "intensive" solution I tried a while back, which could be re-introduced if we suddenly have to start modifying a lot of different component states specifically within modals. Otherwise, I think the prop pass through is fine for now. 

https://github.com/DoSomething/phoenix-next/pull/559

lkpttn:
This looks good to me from a design perspective.

PR #617, opened 2018-01-22
Quiz refactor + Custom title

itsjoekent:
### What does this PR do?
1. Refactored the Quiz component a bit. Realized the fully-inline component was pretty but probably a bit confusing for someone else to jump into. Literally just copy-pasted the same code that was there but put it in named variables.

2. Added the `contentTitle` property to the Quiz Beta, which lets us change the "QUIZ" header to say anything the campaign lead wants. Title defaults to `Quiz` if nothing is given.

screen shot 2018-01-22 at 11 44 27 am
screen shot 2018-01-22 at 11 39 57 am

https://www.pivotaltracker.com/story/show/154293082

### DIFF
--- a/resources/assets/components/Quiz/Quiz.js
+++ b/resources/assets/components/Quiz/Quiz.js
@@ -49,7 +49,7 @@ const Quiz = (props) => {
   return (
     <div className="quiz">
       <div className="quiz__introduction">
-        <h1 className="quiz__title">{fields.contentTitle || Quiz.defaultProps.contentTitle}</h1>
+        <h1 className="quiz__title">{fields.contentTitle || Quiz.defaultProps.fields.contentTitle}</h1>
         <h2 className="quiz__subtitle">{fields.title}</h2>
         {introduction}
       </div>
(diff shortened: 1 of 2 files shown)


resources/assets/components/Quiz/Quiz.js
```diff
@@ -6,43 +6,60 @@ import { ShareContainer } from '../Share';
...
-  </div>
-);
+    </Conclusion>
+  ) : null;
+
+  return (
+    <div className="quiz">
+      <div className="quiz__introduction">
+        <h1 className="quiz__title">{fields.contentTitle || Quiz.defaultProps.fields.contentTitle}</h1>
+        <h2 className="quiz__subtitle">{fields.title}</h2>
```
weerd: This is a little confusing... maybe it should use `title` and `subtitle` in Contentful, instead of `contentTitle` for the title and `title` for the subtitle @_@

itsjoekent: Yeah I hear ya, I think the original naming scheme was designed to mirror the [Lede Banner](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/LedeBanner/templates/MosaicTemplate.js#L63) in terms of how the markup corresponded visually on the page, but definitely doesn't mean it should stay that way

weerd: Ok, well lets change it up when there's an opportunity. Styles are easier to change around than Contenful field names.

PR #622, opened 2018-01-22
Renaming content title to subtitle, reversing name usage

itsjoekent:
### What does this PR do?
- Renamed contentTitle to subTitle
- Reversed usage of subtitle and title

screen shot 2018-01-22 at 4 04 15 pm
screen shot 2018-01-22 at 4 04 06 pm

- contentTitle was confusing and we wanted to get this changed [sooner rather than later](https://github.com/DoSomething/phoenix-next/pull/617#discussion_r163062381)
- the subtitle in the markup referencing the title field was also confusing

### DIFF
--- a/app/Entities/Quiz.php
+++ b/app/Entities/Quiz.php
@@ -59,7 +59,7 @@ class Quiz extends Entity implements JsonSerializable
             'type' => $this->getContentType(),
             'fields' => [
                 'title' => $this->title,
-                'contentTitle' => $this->contentTitle,
+                'subTitle' => $this->subTitle,
                 'slug' => $this->slug,
                 'introduction' => $this->introduction,
                 'conclusion' => $this->conclusion,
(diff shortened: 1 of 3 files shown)


app/Entities/Quiz.php
```diff
@@ -59,7 +59,7 @@ public function jsonSerialize()
             'type' => $this->getContentType(),
             'fields' => [
                 'title' => $this->title,
-                'contentTitle' => $this->contentTitle,
+                'subTitle' => $this->subTitle,
```
weerd: cool! but `subtitle` is one word, not two so no camel-case 😬

weerd: other than that, looks good!

itsjoekent:
@weerd camel case resolved!

itsjoekent:
@weerd can you hit the approve button instead of the comment one? github wont let me merge it

PR #625, opened 2018-01-23
Updating campaign routing logic so the quiz is not gated

itsjoekent:
### What does this PR do?
For Grab The Mic we want to use quizzes as a conversion point that can bypass the Landing Page. This required adding logic that checks if the current page is a quiz page. I achieved this by using the React Router props and subtracting the matched path from the entire URL, and seeing if the remainder string starts with `/quiz/`. In addition to this logical addition, I also broke the statement which determined whether to display a landing page into several parts. Hopefully, it is easy to follow this given all of the different conditionals floating around.

https://www.pivotaltracker.com/story/show/154565578

### DIFF
--- a/resources/assets/components/Campaign/CampaignContainer.js
+++ b/resources/assets/components/Campaign/CampaignContainer.js
@@ -2,13 +2,21 @@ import { connect } from 'react-redux';
 
 import Campaign from './Campaign';
 
-const mapStateToProps = (state) => {
+const mapStateToProps = (state, props) => {
   const hasLandingPage = state.campaign.landingPage !== null;
   const isSignedUp = state.signups.thisCampaign;
 
-  const shouldShowLandingPage = hasLandingPage &&
-    (! isSignedUp || state.admin.shouldShowLandingPage) &&
-    ! state.admin.shouldShowActionPage;
+  const { location, match } = props;
+  const isQuiz = location.pathname.replace(match.url, '').startsWith('/quiz/');
+
+  const blockLandingPage = state.admin.shouldShowActionPage || isQuiz;
+  let shouldShowLandingPage = false;
+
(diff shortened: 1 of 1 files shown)


resources/assets/components/Campaign/CampaignContainer.js
```diff
@@ -2,13 +2,21 @@ import { connect } from 'react-redux';
...
   const hasLandingPage = state.campaign.landingPage !== null;
   const isSignedUp = state.signups.thisCampaign;
 
-  const shouldShowLandingPage = hasLandingPage &&
-    (! isSignedUp || state.admin.shouldShowLandingPage) &&
-    ! state.admin.shouldShowActionPage;
+  const { location, match } = props;
+  const isQuiz = location.pathname.replace(match.url, '').startsWith('/quiz/');
+
+  const blockLandingPage = state.admin.shouldShowActionPage || isQuiz;
```
weerd: Maybe rename `blockLandingPage` to `ignoreLandingPage`? the term `block` has very specific meaning in our system 😬 

Also, if `state.admin.shouldShowActionPage` is `true` that would mean `blockLandingPage` is `true` which seems to be counter what the var name is implying? Am I just confusing things?

itsjoekent: `ignore` works fine for me, but the logic you described is intentional. if `**admin**.shouldShowActionPage` is true then that means bypass the landing page.

weerd: oh duh. yeah, ok. Sorry been staring at too many consecutive PRs @_@

itsjoekent:
Yeah, I think on the prior PR I wrote the routing logic was getting a bit crazy, I still think it's pretty crazy. I have some ideas on making this way simpler but it would require product changes.

PR #627, opened 2018-01-23
adding cursor pointer rule for a tags in list

mendelB:
Super quick: I noticed that the links in the list of share links in [Grab the Mic](https://next-thor.dosomething.org/us/campaigns/grab-mic) (in 'STEP ONE') on thor does not turn to a pointer when you hover (perhaps because they're nested inside `<li>`s?)

Just added the explicit rule in the stylesheet.

### DIFF
--- a/resources/assets/components/ShareAction/share-action.scss
+++ b/resources/assets/components/ShareAction/share-action.scss
@@ -8,6 +8,5 @@
     a {
       cursor: pointer;
     }
-
   }
 }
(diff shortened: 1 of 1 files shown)


weerd:
@mendelB it's actually because they are `<a>` anchor links without the `href=""` attribute!

We'll want to consider how to better treat items on the page that we want to behave as links but apparently are not.... 🤔

PR #630, opened 2018-01-24
Differentiating between run and campaign in voter reg. link

itsjoekent:
### What does this PR do?
- Updated VoterRegAction container to pull in the campaign id
- Fixed component throwing error if content is null
- Fixed duplicate key bug if multiple voter reg actions are on the page 

screen shot 2018-01-24 at 11 24 22 am

screen shot 2018-01-24 at 12 32 20 pm

https://www.pivotaltracker.com/story/show/154590055

### DIFF
--- a/resources/assets/components/VoterRegistration/VoterRegistration.js
+++ b/resources/assets/components/VoterRegistration/VoterRegistration.js
@@ -29,6 +29,9 @@ const VoterRegistration = (props) => {
     link = makeUrl(baseUrl, query).href;
   }
 
+  const formattedContent = (
+    content || VoterRegistration.defaultProps.content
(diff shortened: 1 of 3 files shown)


resources/assets/components/VoterRegistration/VoterRegistration.js
```diff
@@ -39,7 +42,7 @@ const VoterRegistration = (props) => {
         stepIndex={stepIndex}
       />
       <FlexCell width="two-thirds">
-        <Markdown>{ content.replace(/:::[a-zA-Z]*:::/gi, link) }</Markdown>
+        <Markdown>{ formattedContent }</Markdown>
```
mendelB: 👌

PR #633, opened 2018-01-24
Collapse API Middleware Sample Response

weerd:
### What does this PR do?
This PR just wraps the API Middleware Response that I've been using for debugging, etc in a group collapse so that it's not filling up other's consoles in the browser!


weerd:
@itsjoekent 😘

PR #635, opened 2018-01-26
Adding Survey component tests

mendelB:
### What does this PR do?
Adds some test cases for the [`Survey`](https://github.com/DoSomething/phoenix-next/blob/surveyComponentTests/resources/assets/components/Survey/Survey.js) component.

We'll soon be launching an NPS Survey, using the `Survey` component. I figured it'd be prudent to add some tests for what can potentially be a somewhat intrusive feature since there's a somewhat complicated flow of hiding / showing the survey to a user in that component so hopefully this makes us feel safer!

I'm still a jest novice so please forgive if these are somewhat shoddily written 😬. (I tried!)

### Checklist
- [x] Added appropriate feature/unit tests. 🎊

### DIFF
--- a/resources/assets/components/Survey/Survey.test.js
+++ b/resources/assets/components/Survey/Survey.test.js
@@ -6,24 +6,10 @@ import { getTime } from 'date-fns';
 
 import Survey from './Survey';
 import { set } from '../../helpers/storage';
+import LocalStorageMock from '../../__mocks__/localStorageMock';
 
 // Faking timers to be able to interact with and mock our timed survey component
 jest.useFakeTimers();
-
-// Mocking local storage for tests
-class LocalStorageMock {
-  constructor() {
-    this.store = {};
-  }
-
-  getItem(key) {
-    return this.store[key] || null;
-  }
-
-  setItem(key, value) {
-    this.store[key] = value.toString();
-  }
-}
 
 // Setting random userId for all test
 const userId = '1234';
--- a/resources/assets/components/Survey/Survey.test.js
+++ b/resources/assets/components/Survey/Survey.test.js
@@ -50,7 +50,6 @@ const toggleUserCompletion = (value) => {
 }
 
 test('it launches the survey when `ENV`s `SURVEY_ENABLED` is toggled on', () => {
-
   const openModalMock = mountModal();
 
   jest.runAllTimers();
@@ -59,7 +58,6 @@ test('it launches the survey when `ENV`s `SURVEY_ENABLED` is toggled on', () => 
 });
 
 test('it does not launch the survey when `ENV`s `SURVEY_ENABLED` is toggled off', () => {
-
   global.ENV.SURVEY_ENABLED = false;
 
   const openModalMock = mountModal()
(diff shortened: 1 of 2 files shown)


resources/assets/components/Survey/Survey.test.js
```diff
@@ -0,0 +1,144 @@
...
+import { getTime } from 'date-fns';
+
+import Survey from './Survey';
+import { set } from '../../helpers/storage';
+
+// Faking timers to be able to interact with and mock our timed survey component
+jest.useFakeTimers();
+
+// Mocking local storage for tests
+class LocalStorageMock {
```
weerd: Noice! Would be useful to have this be a mock class to import in other tests, but fine for now since only used here :)

DFurnes: Yeah, it could be handy to throw this into the top-level [`__mocks__` directory](https://github.com/DoSomething/phoenix-next/tree/dev/resources/assets/__mocks__) so we remember it's available in the future.

resources/assets/components/Survey/Survey.test.js
```diff
@@ -0,0 +1,144 @@
...
+
+  jest.runAllTimers();
+
+  expect(openModalMock).toHaveBeenCalledTimes(0);
+});
+
+test('it waits 60 seconds to launch survey', () => {
+  const openModalMock = mountModal();
+
+  jest.runTimersToTime(59000);
```
DFurnes: This is super handy. ⏱

resources/assets/components/Survey/Survey.test.js
mendelB: Yeah makes sense! Especially since we'd be using it in some tests I've got in the boiler for #646. I'll try and set this up using the `__mocks__`

resources/assets/components/Survey/Survey.test.js
```diff
@@ -0,0 +1,144 @@
...
+
+  expect(openModalMock).toHaveBeenCalledTimes(0);
+
+  Object.defineProperty(window.location, 'search', {
+    writable: true,
+    value: '',
+  });
+});
+
+test('it does not launch the survey is the user is marked as dismissed less then 30 days ago', () => {
```
weerd: typo! "it does not launch the survey _if_ the user..."

resources/assets/components/Survey/Survey.test.js
```diff
@@ -0,0 +1,144 @@
...
+  set(`${userId}_dismissed_survey`, 'timestamp', Date.now());
+
+  const openModalMock = mountModal();
+
+  jest.runAllTimers();
+
+  expect(openModalMock).toHaveBeenCalledTimes(0);
+});
+
+test('it launches the survey is the user is marked as dismissed more then 30 days ago', () => {
```
weerd: similar typo! "it launches the survey _if_ the user..."

mendelB:
I'm not sure that I used the mocks directory correctly. But I'll continue looking into this. It seems like we can do some cool stuff https://facebook.github.io/jest/docs/en/manual-mocks.html

mendelB:
@DFurnes @weerd Just want to confirm that the `localStorageMock` addition is OK before merging.

PR #636, opened 2018-01-26
Adding Byline to the Affirmation

itsjoekent:
### What does this PR do?
- Adds a new reference field to the Affirmation for linking a Staff entry
- Some styling tweaks after consulting Luke

screen shot 2018-01-26 at 11 29 33 am
screen shot 2018-01-26 at 11 29 22 am
screen shot 2018-01-26 at 11 29 14 am

In order to keep backward compatibility, the Affirmation now has two author fields. The old author field remained the same (Plain text field), but the new field (Staff ref) will supersede the legacy if it exists in the entry.

Last PR for this? https://www.pivotaltracker.com/story/show/152767171


weerd:
> In order to keep backward compatibility, the Affirmation now has two author fields. The old author field remained the same (Plain text field), but the new field (Staff ref) will supersede the legacy if it exists in the entry.

Hoping the Contentful Migration API will help with cleaning this all up 😬

app/Entities/Affirmation.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /**
+     * Convert the object into something JSON serializable.
+     *
+     * @return array
+     */
+    public function jsonSerialize()
+    {
+        $authorValue = $this->newAuthor ?: $this->author;
+        $authorField = null; // TODO: Migrate this field.
```
weerd: TODO 💯

weerd: Can we just call this `$author`? This was actually confusing me thinking it was some other field in Contentful and not what get set based on what `$authorValue` turns out to be.

itsjoekent:
@weerd done!

itsjoekent:
@ashleybaldwin note: we should update all of the affirmations currently in use before the next deplyo

PR #641, opened 2018-01-29
Feature flags backend proposal

mendelB:
## _PULL REQUEST OVERVIEW_
A proposal for some backend setup for feature flag cookies.

### What does this PR do?
- Adds a helper method that can parse through a list (of e.g. cookies) and return a list of feature flags.
- Uses this helper method in the `CampaignsController#show` action to pass feature flags along with the state to the front end.

This is related to the NPS Survey feature, where we want to display the survey feature to only a percentage of traffic. The idea here is that we'll have a cookie being set in Fastly config similarly to how it was done with [this instance](https://github.com/DoSomething/devops/issues/293) of splitting ashes and PN traffic. 
 
So for n percent of traffic, the request will hit us with a set cookie, such as `feature_flag_some_shiny_feature=true`. And we'll pass that along with the state in the response in a list of `featureFlags` as `some_shiny_feature => true`.

Then in React we can work with a feature wrapper component that can use this set state to optionally display the feature/s (I've got something in the works for that if we like this approach).

### Checklist

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -55,10 +55,14 @@ class CampaignController extends Controller
      * @return \Illuminate\View\View
      */
     public function show($slug)
+    public function show(Request $request, $slug)
     {
         $campaign = $this->campaignRepository->findBySlug($slug);
 
         $env = get_client_environment_vars();
+
+        $cookies = Request::cookie();
+        $featureFlags = get_feature_flags($cookies);
 
         // The slug argument will not contain `/blocks` or other path extensions, hence `::url()`.
         $socialFields = get_social_fields($campaign, Request::url());
@@ -73,6 +77,7 @@ class CampaignController extends Controller
                 'id' => auth()->id(),
                 'role' => auth()->user() ? auth()->user()->role : null,
             ],
+            'featureFlags' => $featureFlags,
         ]);
     }
 }
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -398,3 +398,21 @@ function generate_streamed_csv($columns, $records)
 
     return fclose($file);
 }
+
+/**
+ * Get a list of feature flags from provided array.
+ *
+ * @param  array $list
+ * @return array
+ */
+function get_feature_flags($list) {
+    $featureFlags = [];
+
+    foreach ($list as $key => $value) {
+        if (preg_match('/feature_flag_/', $key)) {
+            $featureFlags[str_replace('feature_flag_', '', $key)] = $value;
+        }
+    }
+
+    return $featureFlags;
+}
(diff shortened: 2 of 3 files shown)


app/Http/Controllers/CampaignController.php
```diff
@@ -55,11 +55,15 @@ public function index()
      * @return \Illuminate\View\View
      */
     public function show($slug)
+    public function show(Request $request, $slug)
     {
         $campaign = $this->campaignRepository->findBySlug($slug);
 
         $env = get_client_environment_vars();
 
+        $cookies = Request::cookie();
```
weerd: I'm curious is there's something we could do at a higher level in the RouteProvider or in maybe a middleware to handle this so it doesn't just apply to campaigns... 🤔

mendelB:
Counter proposal:

@itsjoekent raised a valid point, that it would be way simpler for us to take care of the traffic distribution ourselves in React.

Contrary to the scenario I linked to in the description (where traffic was being assigned Ashes or Pheonix Next (Phoenix (next))), here, we have control over all our traffic.

So we can perhaps set up a `TrafficDistribution` component to randomly assign 5 percent of users to the feature, and set the feature flag value in `localstorage` as `true`/`false`.

weerd:
>So we can perhaps set up a TrafficDistribution component to randomly assign 5 percent of users to the feature, and set the feature flag value in localstorage as true/false.

Yep. I dig the idea of us being able to control it better than having to head into Fastly to adjust %. What're the thoughts on how the traffic is distributed? Is it just a random number generator assigning true when a value is generated for 5% of the time? A little fuzzy on the execution here.

mendelB:
Yeah something like `Math.random() <= 0.05` and then setting say `feature_flag_nps` as `true`/`false` in localstorage so we can persist the assignment per user?

weerd:
Cool ok. Further question... what's the benefit of having this done as a component and localStorage rather than doing it as a helper function or something on the backend and setting a global window var from there? I'm fine either way, just curious which is more ideal... 🤔

weerd:
I could see it being better as a component that could wrap other components individually and defining % for that wrapped component specifically. I think in localStorage we'd want to store that the "item" showed for a user so they ideally don't see it again (like the NPS survey).... 🤔

mendelB:
Yeah, that makes sense. so like 
```
<TrafficDistribution percentage={0.5}> 
  <Survey/>
</TrafficDistribution>
```
with TrafficDist taking care of the logic and localstorage setting?

DFurnes:
Yeah, I think that makes great sense.

mendelB:
Sweeet! OK, I'll go ahead and close this PR, and get to work on implementing this in React.

Thanks for the helpful input everyone!!

PR #643, opened 2018-01-30
Adding northstar user fields to referrer objects before passing to csv builder

mendelB:
### What does this PR do?
Adding some Northstar user fields to the `Referral` records before generating a CSV export.

We need some identifier fields for the referring user in our exported CSV of Referral records. 
I know that this is adding some clutter to the controller action which is something @weerd [wanted to avoid](https://github.com/DoSomething/phoenix-next/pull/608#discussion_r163068305). But I wasn't sure how to keep that [csv generating helper method](https://github.com/DoSomething/phoenix-next/blob/dev/app/helpers.php#L383) abstract, while also fetching these new fields that we needed for the CSV. 
Is there perhaps a better way? Maybe just make that helper method specific to our use case, and move all the logic there?

https://dosomething.slack.com/archives/C2BPA7M8F/p1517260740000423

### DIFF
--- a/app/Http/Controllers/ReferralController.php
+++ b/app/Http/Controllers/ReferralController.php
@@ -86,7 +86,16 @@ class ReferralController extends Controller
 
         $referrals = $referralsEloquentBuilder->get();
 
-        $columns = ['id', 'created_at', 'friend_name', 'friend_email', 'friend_story', 'referrer_northstar_id'];
+        // Adding some referrer user details to the referral records, to be able to export them with the CSV.
+        foreach ($referrals as $referral) {
+            $user = gateway('northstar')->getUser('id', $referral->referrer_northstar_id);
+
+            $referral->referrer_first_name = $user->first_name;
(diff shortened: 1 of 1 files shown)


weerd:
One question... @DFurnes is this ok, hitting NS on each record? Isn't there a way to send a series of IDs to NS and get back records? Although not sure how much that would complicate things 🤔

mendelB:
This would have us only making one API call to Northstar. (If you think that this too convoluted though, I can revert this commit)

weerd:
Ok, cool. Let's have @DFurnes weigh in, cause one thing I remember from doing a call where we clump all NS ids together is that w/ Mongo IDs and their length, eventually you could reach the max limit on a URL string. I remember having to do batches in Gladiator when tackling this. But let's have Furnes weigh in on how significant multiple individual calls to NS would be and if it's not a big deal, lets revert and go with the first option.

DFurnes:
Yeah I think keeping this as individual calls is probably simpler for now. :+1:

PR #646, opened 2018-01-30
[#154484254] adding traffic distribution component

mendelB:
### What does this PR do?
- Adds a `TrafficDistribution` component to handle exposing a feature to n percent of traffic.

We started this process in #641 but ended up deciding to approach distributing a feature across traffic from within React.

What this component does is act as a wrapper for any feature (any component etc.) and takes a couple of props:
- `percentage` (int): specifying the percentage of all traffic that should be exposed to this feature.
- `feature` (string): the name of the feature. For labeling and persistence purposes.
- `children` (node): the wrapped component.

It then checks localstorage to see if the user has already been designated to view the feature or not and acts appropriately. Otherwise, it'll do some simple math to randomly assign the user based on specified percentage and store their designation in localstorage.

(I'll add some documentation if this is approved. I've got some tests on the burner, but it would pay to wait for the approval of #635 cause there's some overlap in the testing helpers/setup we need to use.

#641 

### Checklist

### DIFF
--- a/resources/assets/components/TrafficDistribution/TrafficDistribution.js
+++ b/resources/assets/components/TrafficDistribution/TrafficDistribution.js
@@ -1,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { get, set } from '../../helpers/storage';
+
+const TrafficDistribution = ({ percentage, feature, children }) => {
+  let shouldSeeFeature;
+
+  const storedValue = get(feature, 'object');
+
+  if (! storedValue) {
+    shouldSeeFeature = (Math.random() * 100) <= percentage;
+
+    set(feature, 'object', { showFeature: shouldSeeFeature });
+  } else {
+    shouldSeeFeature = storedValue.showFeature;
+  }
+
+  return (
+    shouldSeeFeature ? <span>{ children }</span> : null
+  );
+};
+
+TrafficDistribution.propTypes = {
+  percentage: PropTypes.number.isRequired,
+  feature: PropTypes.string.isRequired,
+  children: PropTypes.node.isRequired,
+};
+
+export default TrafficDistribution;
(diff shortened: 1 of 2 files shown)


resources/assets/components/TrafficDistribution/TrafficDistribution.js
```diff
@@ -0,0 +1,30 @@
...
+  if (! storedValue) {
+    shouldSeeFeature = (Math.random() * 100) <= percentage;
+
+    set(feature, 'object', { showFeature: shouldSeeFeature });
+  } else {
+    shouldSeeFeature = storedValue.showFeature;
+  }
+
+  return (
+    shouldSeeFeature ? <span>{ children }</span> : null
```
weerd: Is there a specific reason for using a `<span>` to wrap the children? Might be more ideal to use a `<div>` since there's less surprises with a `div` as a block level element than a `span` which is inline.

mendelB: Oops, yeah no special reason. I'll switch to `div`!

mendelB: Switched!

resources/assets/components/TrafficDistribution/TrafficDistribution.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { get, set } from '../../helpers/storage';
+
+const TrafficDistribution = ({ percentage, feature, children }) => {
+  let shouldSeeFeature;
+
+  const storedValue = get(feature, 'object');
```
DFurnes: Do we want to prefix this localstorage key, like `TrafficDistribution_${feature}`?

PR #647, opened 2018-01-31
Fixing NavLink highlights & tab order

itsjoekent:
### What does this PR do?
Fixes the tab ordering (Action page is the new default page). Additionally, I fixed the highlighting so the root page `/` will still highlight either "Action" or "Community" depending on what the default page is.

screen shot 2018-01-31 at 11 31 10 am
screen shot 2018-01-31 at 11 31 06 am
screen shot 2018-01-31 at 11 31 00 am

https://www.pivotaltracker.com/story/show/154379917

### DIFF
--- a/resources/assets/components/Navigation/TabbedNavigationContainer.js
+++ b/resources/assets/components/Navigation/TabbedNavigationContainer.js
@@ -82,7 +82,6 @@ const TabbedNavigationContainer = (props) => {
   );
 
   const actionPagePath = join('/us/campaigns', campaignSlug, campaignPaths.action);
-
   const ActionNavigationLink = () => (
     shouldHideAction ? null : (
       <NavigationLink
--- a/resources/assets/components/Navigation/TabbedNavigationContainer.js
+++ b/resources/assets/components/Navigation/TabbedNavigationContainer.js
@@ -1,3 +1,5 @@
+/* global window */
+
 import React from 'react';
 import { join } from 'path';
 import { get } from 'lodash';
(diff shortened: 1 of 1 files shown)


resources/assets/components/Navigation/TabbedNavigationContainer.js
```diff
@@ -62,11 +64,48 @@ const TabbedNavigationContainer = (props) => {
...
+      >Action</NavigationLink>
+    )
+  );
+
+  const communityPagePath = join('/us/campaigns', campaignSlug, campaignPaths.community);
+  const CommunityNavigationLink = () => (
+    shouldHideCommunity ? null : (
+      <NavigationLink
+        to={communityPagePath}
+        isActive={defaultPageIsActiveFunction(shouldHideAction && ! shouldHideCommunity)}
```
mendelB: Wouldn't the `defaultPageIsActiveFunction ` always return to `null` in this case, since `isDefaultPage` will be `false`? If the `ActivityFeed` is being rendered, we [always](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/Page/CampaignPage/CampaignPage.js#L40) redirect to `/community`, which should cause the tab to be active.

Also, I could be misunderstanding something. Our routing logic makes me somewhat dizzy 😵

itsjoekent: I think you misread that statement you linked, the first part is checking `isClosed = true` which makes that irrelevant here. Knowing that should (hopefully?) clear up the confusion around it always being null/false, as its actually possible the community tab renders on the `/` without `/community` in the URL

itsjoekent: Oh wait, hmm, yeah our routing rules are driving me insane. I think it's fine this way because it's covering all cases which is probably better since we have no idea what the scope is

mendelB: Yeah fair enough! 👍

weerd: Yeah, we need to simplify our routing rules as we move forward with stuff. It's getting way too convoluted.

PR #668, opened 2018-02-05
[#154734388] adding some styling edits

mendelB:
## _PULL REQUEST OVERVIEW_
Some super quick edits to the styling of the Section Header

image

https://dosomething.slack.com/archives/C2BPA7M8F/p1517863001000435?thread_ts=1517858302.000753&cid=C2BPA7M8F
[#154734388](https://www.pivotaltracker.com/story/show/154734388)

### DIFF
--- a/resources/assets/components/ActionPage/actionPage.scss
+++ b/resources/assets/components/ActionPage/actionPage.scss
@@ -2,7 +2,9 @@
 
 .action-step {
   width: 100%;
-  margin-bottom: $base-spacing;
+  // @TODO Once we start nesting actions within the action step (as more of a 'section')
+  // we should re-eveluate the margin here
+  margin-bottom: 36px;
 
   > .flex {
     margin: 0 $base-spacing / 2;
--- a/resources/assets/components/SectionHeader/section-header.scss
+++ b/resources/assets/components/SectionHeader/section-header.scss
@@ -13,11 +13,12 @@
   }
 
   .-emphasized:after {
-    margin-top: -10px;
-    margin-left: 5px;
+    margin-top: -12px;
+    margin-left: 4px;
   }
 
   h1.section-header__title {
+    text-transform: uppercase;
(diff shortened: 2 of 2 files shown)


weerd:
Thoughts on also using `margin-bottom: 48px`, so `$section-spacing on `.action-steps`? They feel a little more compacted now @lkpttn

screen shot 2018-02-05 at 6 06 59 pm

lkpttn:
@weerd Agreed, we should add more space in between the actions. The new headers leave them looking a little samey, so I'd like to use space to distinguish them a bit.

mendelB:
Should we be waiting on adding more space beneath the `action-step` till we have the individual actions (Eg. social share, voter reg) embedded within the action step as opposed to existing independently on the action page?

(Meaning, adding this space as of now would sometimes create a gap between the step and the action, as opposed to between the steps themselves:)
image


instead of:

image

lkpttn:
^ I would say this isn't good but I can see where the complications are coming from. Would it be easier to think of it as space before any given step header?

mendelB:
That definitely would be easier. We can throw in `24px` of `margin-top` for a combined `48`?

weerd:
Nope! Don't mix margin top and bottom for this. For one thing, margins collapse, so if you make a top margin 24px and bottom margin 24px, the result is 24px.

weerd:
We've defined everything with margin-bottom and switching to use both top and bottom is going to cause complication elsewhere. I think we should maybe set it to an inbetween value for now (36px) and do another dedicated pass on these in the future.

mendelB:
Yeesh collapsing margins don't sound good. Ok I like that compromise. I think once we formalize this Actions Section > actions business it'll be way more straightforward with just a desired `margin-bottom`

weerd:
Yeah, defs. Lets wait to re-address this once we move to all actions being isolated components and "action steps" _just_ being blocks that _contain_ actions!

PR #670, opened 2018-02-06
Direct "share" route for SMS users.

DFurnes:
### What does this PR do?
This PR adds a "direct share page", which will allow Gambit to capture real "reportbacks" for SMS share actions (instead of that murky business where we ask for screenshots!)

🦈 Adds an un-gated MVP "share page" for SMS users, e.g. `/us/campaigns/gtm-test-page/share?href=https%3A%2F%2Fwww.dosomething.org&id=59820f96a0bfad79b63516fd `.

🎓 Fixes issue (at least with Vim's EditorConfig plugin) where `*.php` settings weren't respected for `*.blade.php` files.

👤 Adds our [test app](https://developers.facebook.com/apps/159247331463057/dashboard/)'s "App ID" to the example `.env` file so we don't manually have to complete this step. This app is configured to work with the `phoenix.test`local domain so share flows are testable.

🐞 Uses the [debug build](https://goo.gl/TBYxCB) of the Facebook SDK so that we get some additional logging and error-checking when working on share features locally.

This ended up being more complicated than it seemed at first:

#### Why not build this into the existing client-side app?
I had a lot of trouble getting our current React routing rules to allow a non-authed campaign page without forcing the "landing page" UI. I also didn't find a clean way to override Puck to get the user ID from the query string (rather than the Redux store) without potentially mucking other things up.

#### Why doesn't this automatically redirect to the Facebook share prompt?
The call to `FB.ui('share', {...});` is blocked by the browser's built-in popup blocker _unless_ it is initiated by a "click" event handler. I briefly experimented with redirecting to the plain [Share URL](https://developers.facebook.com/docs/sharing/reference/share-dialog#redirect) but this doesn't allow us to see the "success" status after a user posts (both success and failure send users to our `redirect_uri` with same `#_=_` hash).

#### How could we make this nicer?
I was thinking that displaying Mendel's beautiful "share action" (#662) on this page would be a nice way to make this experience feel less awkward sans-redirect. __Update:__ See below!

[#154844995](https://www.pivotaltracker.com/story/show/154844995)

### Checklist


mendelB:
> I had a lot of trouble getting our current React routing rules to allow a non-authed campaign page without forcing the "landing page" UI.

I don't know if this is helpful at this point, but Joe (peace be with 👼) worked on un-gating the Quiz, and to circumvent this issue he added some logic to the `CampaignContainer` to have the `ShouldShowLandingPage` set to `false` (which would allow the `CampaignPage` and routing to be rendered), all based on the URL.
https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/Campaign/CampaignContainer.js#L12

Granted this is added complexity to an already complex set of conditions, but it is sitting there if you want to use it 🤷

DFurnes:
Oh my! There it is, indeed. 🕵️ I'd like to consider taking some time in an upcoming sprint to start to untangle some of our routing issues, so my vote would be to avoid adding further conditions for now.

mendelB:
Totally, fair enough!

DFurnes:
Updated this to use the share action introduced in #662 and it's lookin' good!

screen recording 2018-02-07 at 04 19 pm

mendelB:
Doooooooooope! Is that an iPhone simulator?

DFurnes:
Yep yep, just [Simulator](https://developer.apple.com/library/content/documentation/IDEs/Conceptual/iOS_Simulator_Guide/Introduction/Introduction.html) from XCode! 📱

resources/assets/helpers/index.js
```diff
@@ -378,9 +378,23 @@ export function showTwitterSharePrompt(href, quote) {
...
   const urlObject = new URL(path);
-  urlObject.search = new URLSearchParams(query);
+  urlObject.search = new URLSearchParams(queryParameters);
 
   return urlObject;
 }
+
+/**
+ * Share a link by generating a Facebook share prompt.
+ * Get a callback if the share is successful or not.
```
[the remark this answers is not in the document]
DFurnes: Oh gee. Hahaha. 🤦‍♂️

DFurnes:
> Also, a breakaway app from the main campaign app??

Yeah, mostly to be extra sure I didn't break other things. I _would_ like to start experimenting with how we can better break up our app bundle in the future, though - it's getting pretty hefty at 750k!

.editorconfig
```diff
@@ -18,6 +18,10 @@ insert_final_newline = true
 indent_style = space
 indent_size = 4
 
+[*.blade.php]
```
weerd: 👏

weerd:
> Yeah, mostly to be extra sure I didn't break other things. I would like to start experimenting with how we can better break up our app bundle in the future, though - it's getting pretty hefty at 750k!

Planning on starting a discussion in the team GH discussion to think about this in relation to some of the upcoming Page stuff we wanna do too :)

resources/assets/helpers/index.js
```diff
@@ -376,11 +376,24 @@ export function showTwitterSharePrompt(href, quote) {
...
   return urlObject;
 }
+
+/**
+ * Get the query-string value at the given key.
+ *
+ * @param  {String}   key
+ * @return {String}
+ */
+export function query(key) {
```
weerd: Would we want to pass an optional param for the `path` so if not present we only automatically default to use the `window.location`? Just a thought; not vital for merge :)

resources/views/partials/social.blade.php
```diff
@@ -28,7 +28,7 @@
      var js, fjs = d.getElementsByTagName(s)[0];
      if (d.getElementById(id)) {return;}
      js = d.createElement(s); js.id = id;
-     js.src = "//connect.facebook.net/en_US/sdk.js";
+     js.src = "//connect.facebook.net/en_US/{{ config('app.debug') ? 'sdk/debug' : 'sdk' }}.js";
```
weerd: noice!

resources/assets/helpers/index.js
DFurnes: Nice idea!

PR #677, opened 2018-02-09
Share action tests

mendelB:
Adding some tests for the good ole [ShareAction](https://github.com/DoSomething/phoenix-next/blob/dev/resources/assets/components/ShareAction/ShareAction.js) component!

Also added a (somewhat meager) Facebook global object mock, to help with testing that we're triggering an FB share modal, and tracking the (successful or canceled) share.

### DIFF
--- a/resources/assets/components/ShareAction/ShareAction.test.js
+++ b/resources/assets/components/ShareAction/ShareAction.test.js
@@ -2,15 +2,14 @@ import React from 'react';
 import { shallow } from 'enzyme';
 
 import ShareAction from './ShareAction';
-import setFBshare from '../../__mocks__/facebookShareMock.js';
+import setFBshare from '../../__mocks__/facebookShareMock';
 
 describe('ShareAction component', () => {
-
   const url = 'https://dosomething.org';
   const trackingData = { url };
 
   let trackEventMock = jest.fn();
-  let openModalMock  = jest.fn();
+  let openModalMock = jest.fn();
 
   const getShallow = () => shallow(
     <ShareAction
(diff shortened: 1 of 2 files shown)


resources/assets/components/ShareAction/ShareAction.test.js
```diff
@@ -0,0 +1,98 @@
...
+import setFBshare from '../../__mocks__/facebookShareMock';
+
+describe('ShareAction component', () => {
+  const url = 'https://dosomething.org';
+  const trackingData = { url };
+
+  let trackEventMock = jest.fn();
+  let openModalMock = jest.fn();
+
+  const getShallow = () => shallow(
```
weerd: Just curious but would `beforeEach()` here make sense for creating this component before each test?

weerd: oh, nevermind, you use it later in the `describe()`

mendelB: Also what I found tricky about this is that we need a reference to the `shallow` object for testing, and I wasn't sure how to maintain a reference to it through `beforeEach`.

mendelB: Hmm... maybe by declaring it above it scope, and using a `let`? 🤔

mendelB: That worked! And I think it's slightly more elegant. I'll update this and you can confirm

weerd: Noice! I also wonder if [`beforeAll()`](https://facebook.github.io/jest/docs/en/setup-teardown.html#one-time-setup) would help with the scope issue 🤔

mendelB: Oh wait yes!

mendelB: OK @weerd sorry I got a little antsy here. But I *think* what we have based on your feedback is an improvement.

PR #679, opened 2018-02-09
Reportback uploader affirmation

mendelB:
### What does this PR do?
This PR:

- Adds a `PostReportbackModal` which displays the reportback affirmation message to the user after a successful RB submission.
- Removes the old ReportbackAffirmation card and associated logic for displaying the card under the RB uploader
- Triggers the PostReportbackModal from the Admin Campaigns Dashboard's 'Show Reportback Affirmation' button

reportback-affirmation-modal

I am super excited about this! This will ensure our users actually see the affirmation and hopefully feel more like the submission was concretely successful.

[#154850816](https://www.pivotaltracker.com/story/show/154850816)

### DIFF
--- a/resources/assets/actions/reportback.js
+++ b/resources/assets/actions/reportback.js
@@ -16,7 +16,6 @@ import {
   REQUESTED_USER_SUBMISSIONS,
   REQUESTED_USER_SUBMISSIONS_FAILED,
   RECEIVED_USER_SUBMISSIONS,
-  TOGGLE_REPORTBACK_AFFIRMATION,
   queueEvent,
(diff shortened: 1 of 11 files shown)


lkpttn:
@mendelB Can we trigger this modal to appear without jumping back up to the top of the page? Ideally we can maintain the position and page load. 

Also can we add in our loading animation button state? Or would that be a different card.

mendelB:
@lkpttn Totally, I didn't love that it did that either. I think both of those things might need to be separate cards since the way the general modal system is set up now it pops to the top of the page and the loading business is a different concern.

PR #680, opened 2018-02-09
Add optional "post-affiliation button".

DFurnes:
### What does this PR do?
This PR adds an optional "post-affiliation button" to the lede banner. If the `affiliatedActionLink` additionalContent field is set, it'll be placed in the same location as the "Join Us" button pre-affiliation. It can be used to link to anywhere else on the page by anchor link (or theoretically anywhere else).

screen recording 2018-02-09 at 04 03 pm

In particular, these changes were made:

🆔 Adds `id` attributes to generic action steps & the voter reg component. These are then linkable, like so `/us/campaigns/test-teens-for-jeans/action#step-419uoCEXUQs4GEEEmEICa2`

👁 Adds a "hash observer" to the history instance, which will look for hashes in URLs and then see if there's a matching element on the page. This fixes our longstanding issue with not being able to send people to anchor links.

💱 Extracted helpers from the `ScrollConcierge` so we can do the same smooth scrolling here.

🦋 Finally, adds the "affiliated action" button. It's just a link!

 - ~Right now, the hash observer just directly jumps you to that part of the page (like a browser normally would). It could be nice to use our Concierge-style smooth scrolling here.~ Done in 65a8736!
 - __Question:__ I made the step anchors based on the Contentful IDs... would it maybe make more sense to just have these as `#step-1`, `#step-2`, etc? There's downsides (what if steps aren't sequentially numbered, how would we link to non-step actions), but that's perhaps friendlier looking.

[#154380351](https://www.pivotaltracker.com/story/show/154380351)

### Checklist


DFurnes:
@weerd @lkpttn This is ready for review!

weerd:
> I made the step anchors based on the Contentful IDs... would it maybe make more sense to just have these as #step-1, #step-2, etc? There's downsides (what if steps aren't sequentially numbered, how would we link to non-step actions), but that's perhaps friendlier looking.

Yeah, I think we should use the Contentful IDs like you did, especially if we're aiming to do away with using logic to output step number and instead it's just a `preTitle` text field.

resources/assets/components/LedeBanner/lede-banner.scss
```diff
@@ -14,11 +14,19 @@
...
     @include media($large) {
       width: 50%;
     }
+
+    &.-action {
+      background: #fff;
+      color: #000;
+
+      // HACK: Fix awkward blue border on click.
+      &:focus { border: 0 };
```
weerd: This has been plaguing us for a while 😬

resources/assets/components/LedeBanner/templates/MosaicTemplate.js
```diff
@@ -54,6 +57,14 @@ const MosaicTemplate = (props) => {
     </div>
   ), 'lede banner', { text: actionText });
 
+  const actionButton = affiliatedActionLink ? (
+    <div className="header__signup">
+      <Link className={classnames('button', '-action')} to={affiliatedActionLink}>
+        {affiliatedActionText || 'Take Action'}
```
weerd: 👌

resources/assets/helpers/createHistoryHashObserver.js
```diff
@@ -0,0 +1,84 @@
+/* global document window MutationObserver */
```
weerd: @_@

resources/assets/components/LedeBanner/lede-banner.scss
DFurnes: Yeah. Something to re-think with Forge 7!

PR #682, opened 2018-02-12
Fix two issues with null values.

DFurnes:
### What does this PR do?
This PR fixes two kabooms:

1. If `response` in the Redux API middleware is `null`, it'd explode since it couldn't get a `.data` property from it. This just adds a check that `response` exists before we try to get things from it.

2. Grabs the `affiliatedActionText` properties from `additionalContent` more carefully, since it's provided as `null` when empty (rather than an `{}` empty object like I'd assumed!)

🙅‍♂️ 

References [this convo](https://dosomething.slack.com/archives/C2BPA7M8F/p1518459824000076).

### Checklist


weerd:
> If response in the Redux API middleware is null, it'd explode since it couldn't get a .data property from it. This just adds a check that response exists before we try to get things from it.

😱

PR #684, opened 2018-02-12
Switching Herok to use Nginx and configuring it.

### LINKED ISSUE #681: Photo Reportback Uploading on `phoenix-preview` bug
### Current Behavior
What is the current behavior?

**1.**
> On Thor, an error is being thrown when this photo is being uploaded. Prod, its fine. I think its because of Heroku.

image1

**2:**
Reporting back with the photo uploader seems to randomly return a `500` error with the following json response:
```json
{"error":{"code":500,"message":"Looks like something went wrong. We've noted the problem and will try to get it fixed!"}}
```

I grabbed some logs from Heroku:
screen shot 2018-02-09 at 12 11 54 pm
[ds-phoenix-staging-logs-1518208119644.txt](https://github.com/DoSomething/phoenix-next/files/1712351/ds-phoenix-staging-logs-1518208119644.txt)

The above would happen with images of a smaller size, otherwise, I'd generally bump into the `500` error.

### Expected Behavior
Describe the behavior you are expecting.
Both cases **do not** occur on prod. This seems to be a Heroku or thor/staging specific problem.

#### Additional Information
_Any other info or references to cards/tickets from other services (Trello, etc)._
Pivotal ID: [#155037862](https://www.pivotaltracker.com/story/show/155037862)

### Steps to Reproduce
Please provide detailed steps for reproducing the issue.
Go through the regular reportback flow with either the image attached above or some smaller file. 

[Here's a fun example](https://user-images.githubusercontent.com/12417657/36052148-7703178a-0dba-11e8-865a-cd3539a5906a.jpg)

weerd:
### What does this PR do?
This PR updates our Heroku instance to build our `web` process using Nginx instead of Apache. It also adds similar configurations to what Rogue is using to help dictate max file upload sizes.

Fixes #681

### DIFF
--- a/.user.ini
+++ b/.user.ini
@@ -1,0 +1,3 @@
+upload_max_filesize = 10M
+post_max_size = 10M
+memory_limit = 256M
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,3 @@
 web: composer compile && vendor/bin/heroku-php-nginx -C nginx.conf public/
+
+release: php artisan migrate --force
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: vendor/bin/heroku-php-apache2 /public/
+web: composer compile && vendor/bin/heroku-php-nginx -C nginx.conf public/
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
-web: composer compile && vendor/bin/heroku-php-nginx -C nginx.conf public/
+web: composer warmup && vendor/bin/heroku-php-nginx -C nginx.conf public/
 
 release: php artisan migrate --force
--- a/app.json
+++ b/app.json
@@ -9,6 +9,7 @@
     },
     "APP_LOG": "errorlog",
     "APP_URL": "https://ds-phoenix-staging.herokuapp.com/",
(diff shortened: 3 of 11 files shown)


DFurnes:
Looks good to me! ⛵️ 

One gotcha: [route caching](https://laravel.com/docs/5.5/controllers#route-caching) doesn't work with closure-based routes (maybe triggering off the artisan one?) so we might have to disable that to get the Heroku build working.

weerd:
Are we using closure-based routes anywhere?

DFurnes:
Ahhh, I think [route macros](https://github.com/DoSomething/phoenix-next/blob/a41b1e826ac1dfd458a06a5fa9b16e25fc93f8b4/routes/web.php#L11) count as closures. 😖 FWIW, this is built-in once we upgrade to 5.5!

weerd:
Ok, I'll remove the command for now and re-add after upgrading! Nice catch!

DFurnes:
Hmm, things still [do not look super happy](https://ds-phoenix-staging-pr-684.herokuapp.com) on the review app:

screen shot 2018-02-12 at 4 58 07 pm

weerd:
oh boo.

PR #688, opened 2018-02-16
Updating docs and adding a stylesheet.

weerd:
### What does this PR do?
This PR does a few updates to the docs content but mostly it removes the horizontal rules from titles on every page, and adds a stylesheet to add some nicer `h1` heading styles :)

@DFurnes made a good point in an earlier PR as to why we were using `***` to create horizontal rules after page titles, since it was semantically unnecessary and mostly used for styling purposes.


DFurnes:
screen shot 2018-02-16 at 12 13 03 pm


(╯°□°)╯︵ ┻━┻

weerd:
😬

PR #697, opened 2018-02-22
Re-adding a key to list items output in the action steps.

weerd:
### What does this PR do?
This PR re-adds a key to outputted items in the action steps wrapper to eliminate the warnings popping up.

It also updates a couple component tests that were failing.

### Checklist


weerd:
Thanks @mendelB 💃

PR #698, opened 2018-02-23
Adding Twitter to the ShareAction

### LINKED ISSUE #695: Add Twitter Share to the Share Action
### User Story
As a Campaign Lead, I want the new Share Action to work with Twitter. As well as the ability to toggle a Share Action to be a Twitter or Facebook share.

#### Additional Information
Pivotal ID: [#154526005](https://www.pivotaltracker.com/story/show/154526005)

### Why This Matters
This feature will work towards enabling us to depart from the screenshot Reportback format for social share actions.

### Tasks
This feature includes the following tasks:
- [ ] Add Twitter sharing functionality to the Share Action
- [ ] Add Twitter share tracking to the Share Action so we can capture the RB automatically
- [ ] Enable toggling between Facebook and Twitter for the Share Action on Contentful

mendelB:
### What does this PR do?
This PR adds the ability to toggle the `ShareAction` to use Facebook or Twitter for the share button.

- added new `socialPlatform` field to the ShareAction Contentful Entry.
- based on this field the ShareAction will render a Facebook or Twitter share button
- added twitter share functionality and puck tracking

Also some additions to the `showTwitterSharePrompt` function, enabling it to receive an optional `callback` param, which it will trigger once the twitter share intent window is closed.

Being able to show a post share affirmation to the user for the twitter share proved to be somewhat difficult, and unlike Facebook, we do not have a callback for post success or canceled shares.

I went with setting up a `setInterval` to check for when the new twitter intent window is `closed`, and once that's true the callback is triggered. This is admittedly not the most optimal pattern but the best I was able to come up with under these circumstances!

Pivotal ID [#154526005](https://www.pivotaltracker.com/story/show/154526005)
closes #695 

### Checklist
Visually the share is identical to how it is now (save for Twitter replacing Facebook in the button text, so I'll not attach the screenshots)

### DIFF
--- a/resources/assets/components/ShareAction/ShareAction.js
+++ b/resources/assets/components/ShareAction/ShareAction.js
@@ -5,16 +5,16 @@ import Card from '../Card';
 import Embed from '../Embed';
 import Markdown from '../Markdown';
 import { POST_SHARE_MODAL } from '../Modal';
-import { showFacebookSharePrompt } from '../../helpers';
+import { showFacebookSharePrompt, showTwitterSharePrompt } from '../../helpers';
 
 import './share-action.scss';
 
 const ShareAction = (props) => {
-  const { title, content, link, openModal, trackEvent } = props;
+  const { title, content, link, socialPlatform, openModal, trackEvent } = props;
 
   const onFacebookClick = (url) => {
     const trackingData = { url };
-    trackEvent('clicked share action', trackingData);
+    trackEvent('clicked facebook share action', trackingData);
 
     showFacebookSharePrompt({ href: url }, (response) => {
       if (response) {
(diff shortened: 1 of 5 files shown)


weerd:
> I went with setting up a setInterval to check for when the new twitter intent window is closed, and once that's true the callback is triggered. This is admittedly not the most optimal pattern but the best I was able to come up with under these circumstances!

Just a thought and maybe a bad idea, but could we hook into a `mouseDown` event (or some other reasonable event) instead of using a `setInterval`? Is it actually the user closing the intent window or does it close on its own? I'm also not completely sure I recall how these work otherwise 🤔

resources/assets/components/ShareAction/ShareAction.js
```diff
@@ -54,6 +77,7 @@ ShareAction.propTypes = {
   link: PropTypes.string.isRequired,
   openModal: PropTypes.func.isRequired,
   trackEvent: PropTypes.func.isRequired,
+  socialPlatform: PropTypes.oneOf(['twitter', 'facebook']).isRequired,
```
weerd: 👌

mendelB:
It closes on its own when you actually tweet. But otherwise, you need to manually x out the page. Also annoyingly, there's no way to hook any events into the twitter window itself since it's not owned by our domain so you get the cross-origin error. So this felt like the only way to tell that they'd either tweeted or canceled.

mendelB:
Also, just realized the field is returned as an array 😳and we weren't grabbing it correctly. Fixed, and set it to default to 'facebook' if that's a good idea?

weerd:
Cool! Let's give it a whirl then!

PR #699, opened 2018-02-23
Un-gate quiz results

mendelB:
### What does this PR do?
This PR un-gates the Quiz, and allows user to see their quiz results without prompting the log-in flow.

Users who are already logged in will continue to be auto signed up for the Campaign but will also see their quiz results seamlessly.

Also changed the formatting of the `details` param when [dispatching a signup](https://github.com/DoSomething/phoenix-next/blob/ungate-quiz-results/resources/assets/actions/quiz.js#L49) from the quiz for logged in users. Previously we were passing through an object `{ source: 'quiz' }` which was causing an error and `500` from Rogue. Changed this over to a string. (`'source:quiz'`)

**Question**: There's still some logic in the [`quizConvert`](https://github.com/DoSomething/phoenix-next/blob/ungate-quiz-results/resources/assets/actions/quiz.js#L34) action that deals with an unauthenticated user, and triggers the Northstar flow. We're only calling this method now if they're authenticated so it won't bother us, but I can remove it if we think that it's not going to be needed anymore? -- UPDATE: REMOVED.

Fixes [#154881463](https://www.pivotaltracker.com/story/show/154881463)

### DIFF
--- a/resources/assets/actions/quiz.js
+++ b/resources/assets/actions/quiz.js
@@ -66,7 +66,9 @@ export function completeQuiz(quizId) {
     }
 
     document.querySelector('.main').scrollIntoView(true);
-    return dispatch(quizConvert(quizId));
+
+    return getState().user.id ?
+      dispatch(quizConvert(quizId)) : dispatch(viewQuizResult(quizId));
   });
 }
 
--- a/resources/assets/actions/quiz.js
+++ b/resources/assets/actions/quiz.js
@@ -3,7 +3,6 @@
 import { find } from 'lodash';
 import {
   clickedSignUp,
-  queueEvent,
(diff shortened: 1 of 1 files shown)


weerd:
> Question: There's still some logic in the quizConvert action that deals with an unauthenticated user, and triggers the Northstar flow. We're only calling this method now if they're authenticated so it won't bother us, but I can remove it if we think that it's not going to be needed anymore?

@mendelB Yeah, I think that should be fine given how we're planning on treating the Quiz moving forward.

mendelB:
@weerd OK I got rid of that snippet that would stash the quiz in localstorage and send the user for auth, as well as the part that would address grabbing that stashed quiz data if available.

PR #693, opened 2018-02-20
Display "action step" blocks (like voter reg.) in affirmation modal.

DFurnes:
### What does this PR do?
This PR allows editors to place "action step" blocks into the affirmation modal. This allows us to push other actions than sharing the campaign, like registering to vote.

screen shot 2018-02-20 at 3 39 14 pm

And here's the same page [on tablet](https://user-images.githubusercontent.com/583202/36450824-36ef18f0-165d-11e8-9d1d-573d914a9ccb.png) & [on mobile](https://user-images.githubusercontent.com/583202/36450795-11a279c0-165d-11e8-8e65-43fbe152aa0c.png)!

### Any background context you want to provide?
This involved more refactoring than I originally estimated. In particular:

🌵 Extract a single generalized `ActionStepBlock` component out of the `ActionStepWrapper` (which is used to render the entire action page). 2172965

📳 Add `Affirmation` as a possible "action step" block, and then use the `ActionStepBlock` in the affirmation modal! This should keep existing behavior for current campaigns, but allows us to start to introduce all the fun new stuff here over time. 3ccc211

🏭 Pull "page-specific" elements (like built-in section headers & column layout wrappers) into the `ActionStepWrapper` component, so `ActionStepBlock`is _just_ the step itself. b76858e

💱 Creates a `ModalControls` component that overlays the "close card" button, so that we don't have to muddy the rest of our components by passing this down through the hierarchy. e7a0390

🆔 Finally, this makes _all_ steps on the action page linkable! (Not just section headers.) 967bb8e

### What are the relevant tickets/cards?
[#154732073](https://www.pivotaltracker.com/story/show/154732073)

### Checklist
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.
- [x] Added screenshot to PR description of related front-end updates on **small** screens.
- [x] Added screenshot to PR description of related front-end updates on **medium** screens.
- [x] Added screenshot to PR description of related front-end updates on **large** screens.

PR #700, opened 2018-02-23
WIP: Tidy up pages and blocks

DFurnes:
### What does this PR do?
This PR moves some things around to be more discoverable: it puts the main "action block" component from #693 into the `./Actions` directory with its other friends, and moves `./ActionPage` into the `./Pages` directory with its brethren. 

I wanted to move all the "actions" into the `./Block` directory so we can have them all work the same way (give Contentful JSON, get rendered block), but there's still some special parameters for "action blocks" that we need to be rid of (`stepIndex` & `isSignedUp`).

This was tackled as part of [#154881362](https://www.pivotaltracker.com/story/show/154881362).

### Checklist


weerd:
> with its brethren.

😂

weerd:
> I wanted to move all the "actions" into the ./Block directory so we can have them all work the same way (give Contentful JSON, get rendered block), but there's still some special parameters for "action blocks" that we need to be rid of (stepIndex & isSignedUp).

@DFurnes I don't necessarily mind having the `Actions` be in their own `./Actions` directory, since they are very specific components and won't necessarily work like all others. It helps maintain an understanding of these actions as being very specific things, contain logic that "does some action" and thus separated. 

My general rule for the `Page` directory and the different components in there is that the name of the component is suffixed with the name of the directory.... so `CampaignPage`, `LandingPage`, `ActionPage` all make sense to go in the containing `Page` directory. Currently most of the new action components and how we've been moving forward with them is with an `Action` suffix, `LinkAction`, `VoterRegistrationAction`, `ThirdPartyAction`, etc.

For the `Blocks` directory I could see things like the `ContentBlock`, and `GalleryBlock` going in there, and any future custom blocks we make for new features/new scenarios, but they're all suffixed with `Block`.

DFurnes:
Welp, this is ready for review Monday morning! Everything seems good from informal testing on my local, but may be best to hold off merging until after we do our prod deploy nonetheless. 😺

PR #706, opened 2018-02-26
Adding contentful migrations for the CampaignUpdate Contentful model

mendelB:
### What does this PR do?
This PR adds two migrations for the CampaignUpdate Contentful model:
1. Adding an Affiliate Logo field. 
   - Optional field
   - Accepts an asset link
2. Making the Author field optional

This is to support adding Affiliate Link campaign updates that pop [#154880493](https://www.pivotaltracker.com/story/show/154880493).

The affiliateLogo field will work similarly to the way it works in the [LinkAction](https://app.contentful.com/spaces/81iqaqpfd8fy/content_types/linkAction/fields).

The author field is now optional since that left-hand side of the Blocks footer will now contain either the author *or* the sponsor promo if the affiliate logo was added:
image

It's kind of a toggle situation between the two fields, and I wasn't sure if there's a better/clearer way to handle this?

Also, I promise I won't run this migration till the work is done to support these changes!

hashtag contentfulMigrationsFTW


weerd:
Number them! `2018_02_26_002_....` and `2018_02_26_003_....` since I already added `001` 😄

mendelB:
Ohhh right I knew I was forgetting something!

contentful/migrations/2018_02_26_002_add_affiliate_logo_field_to_campaign_updates.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function (migration) {
+  const page = migration.editContentType('campaignUpdate');
```
weerd: Also just realized the `const` should probs not be named `page` ;)

weerd: Tossing up a quick PR :)

PR #708, opened 2018-02-27
Move Lede Banner into individual pages, and allow toggling it!

DFurnes:
### What does this PR do?
This pull request moves the `LedeBanner`, `Dashboard`, and other "page chrome" into each individual page (rather than in the top-level campaign route). This allows us to easily toggle it case-by-case... for example, so we can hide it on a quiz page:

screen shot 2018-02-27 at 3 40 20 pm

I created `LedeBannerContainer` and `DashboardContainer` so that these did not need props to be passed in from each parent component (and I think moves us closer to idiomatic Redux usage). I left a `dashboard` prop injected into each of the parent pages, so that the `Dashboard` component could be toggled based on whether or not it's set on the campaign. Not sure if that's ideal or not... 🤔

Moving the rest of the page layout into each specific Page component fits into larger refactoring goals that @weerd & I had talked about, but also makes [the real change](https://github.com/DoSomething/phoenix-next/pull/708/commits/3446b2bc00dfd9c4edbc540da1e97e6b41b4087a) pretty darn simple! Nice!

[#154881362](https://www.pivotaltracker.com/story/show/154881362)

### Checklist

### DIFF
--- a/app/Entities/Quiz.php
+++ b/app/Entities/Quiz.php
@@ -66,6 +66,7 @@ class Quiz extends Entity implements JsonSerializable
                 'comparison' => $this->comparison,
                 'callToAction' => $this->callToAction,
                 'questions' => $this->parseQuestionsFromJson(),
+                'additionalContent' => $this->additionalContent,
             ],
         ];
     }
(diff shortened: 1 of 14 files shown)


DFurnes:
Yeah, I think that would be a good idea. 👍 We can definitely take a few more passes on cleaning up these `Page` components over the next couple of sprints.

PR #709, opened 2018-02-27
Progress on showing an action for a quiz result.

### LINKED ISSUE #704: Update quiz to show different results based on age.
### User Story
As a quiz user, I want content presented to me based on my age so that I get a registry link only if I can register.

#### Additional Information
Pivotal ID: [#154881437](https://www.pivotaltracker.com/story/show/154881437)

### Why This Matters
We should be filtering people based on age to get them to the content we want them to do. Without this, we estimate we would be losing out on (x) Facebook shares at the end.

### Definition Of Done
Given a user gets to the end of the quiz, when they see their results then they are prompted to either register or share depending on age.

### Tasks
This feature includes the following tasks:
- [ ] Quiz contentful migration add results field #705

weerd:
### What does this PR do?
This PR Adds dynamic quiz result actions to the results shown from the Quiz - for now - specifically tailored for the `give-a-spit-about-cancer` quiz we're running.

- adding a `results` field to the Quiz Contentful Entry which is an array of references to the potential result actions
- adding a `resultActions` field to the Quiz Contentful Entry which is a reference to the `type` and `id` of the potential result actions
- adding internal logic to render the correct result action based on the users' response to the first question in the quiz
- some styling fixes 

We introduced a slight hack to the quiz so that based on the results of the first answer (Age), we'd be able to return a specific result action from the quiz.

It works like so:

We have a field of `results` containing an array of references to the potential result actions.
We specify the `type`s and `id`s of said results in a `resultActions` field. (This is set within the `questions` JSON field in contentful).
Once the user clicks 'Get Results' we check to see what they answered to the first questions and based on the response find the resultAction corresponding to the proper action 'type' and set that as the `resultActionId`.
Within the Quiz, we find the action with that ID, and render it as the quiz result!

quiz

Pivotal ID [#154881437](https://www.pivotaltracker.com/story/show/154881437)
closes #704 
#705 

### Checklist

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -106,7 +106,7 @@ class Campaign extends Entity implements JsonSerializable
     /**
      * Parse and extract data for an action step.
      *
-     * @param  Entry $entry
+     * @param  Entry $step
      * @return array
      */
     public function parseActionStep($step)
--- a/app/Entities/Quiz.php
+++ b/app/Entities/Quiz.php
@@ -33,6 +33,8 @@ class Quiz extends Entity implements JsonSerializable
      */
     private function parseQuestionsFromJson()
     {
+        // @TODO: Not sure why this was not named additionalContent like every other
+        // instance where we use an additionalContent JSON field @_@
         $questions = $this->questions->get('items');
         if (! $questions) {
             return [];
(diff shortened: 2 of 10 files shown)


mendelB:
@weerd I added in the rendering of the various result actions. It is functional!

mendelB:
@DFurnes Ok, I think this is ready for review.

PR #718, opened 2018-03-01
Query params fun

mendelB:
## _PULL REQUEST OVERVIEW_

Frustratingly, the [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) which we use to deal with safely [constructing](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/helpers/index.js#L408) and [parsing](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/helpers/index.js#L422) URL params acts funky with older versions of Safari and presumably some other browsers. We tried to appease them by stringifying things to no avail.

Upon @DFurnes [recommendation](https://dosomething.slack.com/archives/C2BPA7M8F/p1519941483000320), we're installing the `queryString` npm package and using that to do reliably preform our URL param manipulations. 

browser-compatibility-meme

PIVOTAL ID [#155533170](https://www.pivotaltracker.com/story/show/155533170)
#710 😢

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2663,6 +2663,11 @@
       "integrity": "sha1-9lNNFRSCabIDUue+4m9QH5oZEpA=",
       "dev": true
     },
+    "decode-uri-component": {
+      "version": "0.2.0",
+      "resolved": "https://registry.npmjs.org/decode-uri-component/-/decode-uri-component-0.2.0.tgz",
+      "integrity": "sha1-6zkTMzRYd1y4TNGh+uBiEGu4dUU="
+    },
     "deep-diff": {
       "version": "0.3.4",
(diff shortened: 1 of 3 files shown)


resources/assets/components/Modal/configurations/SurveyModal.js
```diff
@@ -29,7 +29,7 @@ class SurveyModal extends React.Component {
     };
 
     const typeformUrl = makeUrl(SURVEY_DATA_URL, typeformQuery);
-
+    debugger
```
weerd: 😱

mendelB: LOL

PR #720, opened 2018-03-02
Puck tracking voter reg

mendelB:
### What does this PR do?
This PR adds `puck` tracking to the VoterRegistrationAction link.

- adding `contentfulId` prop to VoterRegistration so we can track it
- adding `modalType` prop (fetching it from the state through the container) so we can track when the voter reg is in a modal, and what type of modal (e.g. signup affirmation vs content modal (content modals are not currently supported, but will be what I'm working on next)). If it's not a modal we'll track it as `null`

The modal tracking logic is based on our current modal setup which will hopefully be changing soon. So just bear in mind the temporary nature of tracking the `modalType` through `state`. (Soon we'll hopefully be able to plainly see if it's a modal (and perhaps the type of modal if relevant) through the windows URL location).

Updated the tests to use `shallow` instead of `render` to be able to test interacting with the button for tracking tests. (As an aside, I'm always unsure -- are tracking tests a concern of the component? Is it indeed relevant to be testing tracking data?)

Pivotal ID [#155479093](https://www.pivotaltracker.com/story/show/155479093)
[Some -Slack convo- context for the data setup](https://dosomething.slack.com/archives/C2BPA7M8F/p1520009934000069)

### Checklist

### DIFF
--- a/resources/assets/components/Actions/VoterRegistrationAction/VoterRegistrationAction.js
+++ b/resources/assets/components/Actions/VoterRegistrationAction/VoterRegistrationAction.js
@@ -26,12 +26,19 @@ const VoterRegistrationAction = (props) => {
     source: 'web',
   };
 
+  const parsedLink = link && dynamicString(link, tokens);
+
+  const handleClick = () => {
+    const trackingData = { url: parsedLink, modal: modalType }
+    trackEvent('clicked voter registration action', trackingData)
+  }
+
   return (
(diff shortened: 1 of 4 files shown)


weerd:
> As an aside, I'm always unsure -- are tracking tests a concern of the component? Is it indeed relevant to be testing tracking data?

🤔 I eventually want to rethink how we do tracking so it's less involved, uses a simple middleware to send data to Puck, etc and then ideally it's not a concern of the component when testing.

PR #725, opened 2018-03-02
Configure standardized block page for SMS shares.

DFurnes:
### What does this PR do?
This PR implements a standardized "block" page (based on the existing feature to let us link directly to blocks in the activity feed). This will allow campaign leads & the messaging team to spin up social shares and edit them without being scared off by writing JSON.

blockpageshare

I'd recommend reviewing this pull request commit-by-commit:

🚜 Properly transform any blocks passed in the `pages` field, like we do elsewhere. 970db81

🔷 Simplify the `/blocks/:id` layout to match the SMS Share prototype. cd7fe13

👓 Look for blocks in any block reference field for the `/blocks/:id` page, not _just_ the community feed. This will allow us to link action page blocks or completely unique "link-only" blocks. 1b6af74

🔓 Allow anonymous users to visit `/blocks/:id` without being show the landing page. In the future, we should refactor this so that auth checks happen within each individual page. a618329

🔑 Extract `isAuthenticated` selector & use it to determine login status. b4d4e54

🍰 Finally, swap that `isAuthenticated` selector to use a new "isAuthenticated" boolean from the server so we can safely swap out `user.id` using the `?user_id=` query string without implying that the person is _really_ logged in. ec4aa28, 9ee9d48

This is part of an ongoing effort to make these elements more flexible, and is made possible due to the work in #700, #708, and #711. Rather than making a new "SMS Share" content type, we can just use the existing "Share Block" here!

[#155551584](https://www.pivotaltracker.com/story/show/155551584)

### Checklist

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -70,7 +70,7 @@ class CampaignController extends Controller
         ])->with('state', [
             'campaign' => $campaign,
             'user' => [
-                'id' => auth()->id(),
+                'id' => auth()->id() ?: request()->query('user_id'),
                 'isAuthenticated' => auth()->check(),
                 'role' => auth()->user() ? auth()->user()->role : null,
             ],
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -71,6 +71,7 @@ class CampaignController extends Controller
             'campaign' => $campaign,
             'user' => [
                 'id' => auth()->id(),
+                'isAuthenticated' => auth()->check(),
                 'role' => auth()->user() ? auth()->user()->role : null,
             ],
         ]);
--- a/resources/assets/selectors/user.js
+++ b/resources/assets/selectors/user.js
@@ -5,7 +5,7 @@
  * @returns {String}
  */
 export function isAuthenticated(state) {
-  return Boolean(state.user.id);
+  return state.user.isAuthenticated;
 }
 
 /**
--- a/resources/assets/selectors/user.js
+++ b/resources/assets/selectors/user.js
@@ -4,6 +4,16 @@
  * @param state
  * @returns {String}
  */
+export function isAuthenticated(state) {
+  return Boolean(state.user.id);
+}
+
+/**
+ * Get the user id from the state.
+ *
+ * @param state
+ * @returns {String}
+ */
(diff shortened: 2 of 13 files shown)


DFurnes:
I'm going to stop adding stuff to this PR now, since it grew a bit heavy with the `isAuthenticated` selector refactoring commit. I'll add back in the "post-share" action & ability to customize it in Contentful in a follow-up PR! 😺

resources/assets/selectors/user.js
```diff
@@ -1,3 +1,13 @@
+/**
+ * Get the user id from the state.
+ *
+ * @param state
+ * @returns {String}
+ */
+export function isAuthenticated(state) {
```
weerd: I wonder if we should only be passing the `state.user` to the function, so that it only ever expects an object which contains a user `id` and not necessarily an object that could be of potentially different structure with the `user` object at a different level or as the sole object. Also maybe have it default to an empty object?

```javascript
export function isAuthenticated(userState = {}) {
...
}
```

app/Http/Controllers/CampaignController.php
```diff
@@ -71,6 +71,7 @@ public function show($slug)
             'campaign' => $campaign,
             'user' => [
                 'id' => auth()->id(),
+                'isAuthenticated' => auth()->check(),
```
weerd: 👏

resources/assets/selectors/user.js
weerd: or calling the param just `user`, but figured `userState` made it more obvious.

DFurnes: Mmm, not sure! I think the general pattern here is to pass the whole state, so that if we were to move this to a separate part of the tree later on we would just have one place to change it (and to avoid potential bugs from passing the wrong portion of the state).

DFurnes: Ideally, I'd love to be able to say "this function must get an object of type `UserState`" but that's probably something for down the road with more static typing investments...

weerd: 🤔 hokay!

DFurnes: We've never actually closed the loop on this (by actually using the library), but I've used [reselect's documentation](https://github.com/reactjs/reselect#example) as an example for how these should be structured.

PR #726, opened 2018-03-05
Abstracting a Modal launcher component

mendelB:
### What does this PR do?
This PR adds a dedicated `ModalLauncher` component which can take care of launching a specified modal after a specified amount of time.

Also, using this component from within the `Survey` component, allowing for some nice trimming down of the `Survey` component itself.

This is the beginnings of the work to be able to create a VoterRegistration Modal pop-up similar to the Survey pop up modal. Now that we have this general ModalLauncher, we can use it for VoterRegistration as well.

### Checklist
If approved documentation shall follow.

### DIFF
--- a/resources/assets/components/ModalLauncher/ModalLauncher.js
+++ b/resources/assets/components/ModalLauncher/ModalLauncher.js
@@ -1,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class ModalLauncher extends React.Component {
+  constructor() {
+    super();
+
+    this.state = {
+      count: 0,
+    };
+
+    this.incrementOrLaunch = this.incrementOrLaunch.bind(this);
+  }
+
+  componentDidMount() {
+    this.timer = setInterval(this.incrementOrLaunch, 1000);
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.timer);
+  }
+
+  incrementOrLaunch() {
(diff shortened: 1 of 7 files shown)


resources/assets/components/ModalLauncher/ModalLauncher.js
```diff
@@ -0,0 +1,43 @@
...
+
+    this.state = {
+      count: 0,
+    };
+
+    this.incrementOrLaunch = this.incrementOrLaunch.bind(this);
+  }
+
+  componentDidMount() {
+    this.timer = setInterval(this.incrementOrLaunch, 1000);
```
DFurnes: Instead of incrementing a counter value, could we set the timeout to the "launch" time?

```js
this.timer = setInterval(this.launchModal, 1000 * this.props.countdown);
```

mendelB: Totally! 
I'm trying to recall if there was a reason we didn't do this in the first place, as that feels like it should have been the obvious approach to this, right? 🤔

mendelB:
Updated to use `setTimeout`, thus drastically simplifying the whole deal!

resources/assets/components/ModalLauncher/ModalLauncher.js
mendelB: Whatever the case, updating this as you suggest seems to work perfectly and the tests are passing so...

PR #728, opened 2018-03-05
Adds optional post-share button to the standard Share Action.

DFurnes:
### What does this PR do?
This pull request adds the ability to have a post-share button on the standard "Share Action" block, just like we did with the SMS Share prototype. Unlike in the prototype though, we can use this on _any_ Share Action now (on mobile, desktop, you name it!)

screen shot 2018-03-05 at 4 41 12 pm

It also fixes what turns out was a long-standing issue where _all_ share affirmations would show the content from the first share on the given campaign, instead of showing the corresponding content for that particular share block.

I've kept these in `additionalContent` for this PR because I'd like some feedback on how best to model this kind of thing in Contentful. Two separate "Affirmation Button" and "Affirmation Link" fields?

Another idea I was toying with was a "Block" relationship that'd let us just display an optional block alongside this affirmation message... adds a bit more admin overhead, but could be handy!

[#155551584](https://www.pivotaltracker.com/story/show/155551584)

### Checklist

### DIFF
--- a/resources/assets/components/ShareAction/ShareAction.js
+++ b/resources/assets/components/ShareAction/ShareAction.js
@@ -10,7 +10,7 @@ import { showFacebookSharePrompt, showTwitterSharePrompt } from '../../helpers';
 import './share-action.scss';
 
 const ShareAction = (props) => {
-  const { title, content, link, socialPlatform, openModal, trackEvent } = props;
+  const { id, title, content, link, socialPlatform, openModal, trackEvent } = props;
 
   const onFacebookClick = () => {
     const trackingData = { url: link };
@@ -19,7 +19,8 @@ const ShareAction = (props) => {
     showFacebookSharePrompt({ href: link }, (response) => {
       if (response) {
         trackEvent('share action completed', trackingData);
-        openModal(POST_SHARE_MODAL);
+        // @TODO: Render a <Modal>...</Modal> component here!
+        openModal(POST_SHARE_MODAL, id);
       } else {
         trackEvent('share action cancelled', trackingData);
       }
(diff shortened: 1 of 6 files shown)


resources/assets/components/ShareAction/ShareAction.js
```diff
@@ -68,10 +69,17 @@ const ShareAction = (props) => {
...
+  confirmationAction: null,
+  confirmationActionLink: null,
   content: null,
 };
 
 ShareAction.propTypes = {
+  id: PropTypes.string.isRequired,
+  // affirmation: PropTypes.string,
+  // confirmationAction: PropTypes.string,
+  // confirmationActionLink: PropTypes.string,
```
mendelB: Are these pending us rendering an actual `<Modal>` post share?

DFurnes: Oof, yeah these were from experimenting with that but I should've removed 'em!

weerd:
> Another idea I was toying with was a "Block" relationship that'd let us just display an optional block alongside this affirmation message... adds a bit more admin overhead, but could be handy!

Not sure if I'm understanding correctly...

So part of the issue is that there's an `affirmation` field in the `ShareAction`, and you're debating on names for housing the possibility of a `confirmationAction` and `confirmationActionLink` fields along with the message, correct?

And you mention the possibility of instead using a `blocks` reference field to display a block along with the message?

What if we got rid of the `affirmation` field in the `ShareAction`, added the `blocks` ref field instead, and then spruced up the `affirmation` content-type and what you reference is an `affirmation` block? This way they could be used in multiple spots on the same campaign, etc and ideally all the affirmation related info is contained in that instance of the `affirmation` content-type.

DFurnes:
OH MY GOD, DUH! I forgot we have a _literal_ Affirmation Block! Of course!! 👍 👍 

Alright, I'm very in favor of using a block reference field here then. I'll see about hooking that up next, and we can write a ticket to migrate this text `affirmation` field into a real block in an upcoming sprint.

PR #730, opened 2018-03-06
Modal launcher v2

mendelB:
### What does this PR do?
This PR updates the `ModalLauncher` to run the logic as to whether to show or hide the timed modal. This is mostly abstracted from the old Survey component, which, thanks to this is no longer needed at all! We can just render a general ModalLuancher and provide it some specific info as props.

This will be the necessary work to be able to render a VoterRegistrationAction timed modal, and that exists in an upcoming PR! As well as any other timed modal we might need in the future 🔮 

NOTE: I changed the format for the localstorage key when dismissing or 'hiding' the timed modal to be more general. The issue is that we've already been setting this a certain way for the NPS Typeform Survey, so we need to still support that format for a while I imagine?

#726 

### Checklist

### Pre deploy checklist
- [ ] Change the `env` variable on prod and staging from `SURVEY_ENABLED` to `NPS_SURVEY_ENABLED`
- [ ] Change the callback hook query on Typeform from `?finished_nps=1` to `?hide_nps_survey=1`


mendelB:
UPDATE: 
- Fixed an issue
- Updated tests
- Rebased from master
- Updated env variables

resources/assets/components/ModalLauncher/ModalLauncher.js
```diff
@@ -1,25 +1,79 @@
...
+      shouldNotSee = get(`${userId}_finished_survey`, 'boolean');
+    }
+
+
+    // @see: SurveyModal.js
+    const dismissalTime = get(`${userId}_dismissed_${type}`, 'timestamp');
+    // Check if the survey was dismissed over 30 days ago.
+    const isDismissed = isTimestampValid(dismissalTime, (30 * 1440 * 60 * 1000));
+
+    return env[`${type.toUpperCase()}_ENABLED`] && ! shouldNotSee && ! isDismissed;
```
DFurnes: Let's use our [client-side `env` helper](https://github.com/DoSomething/phoenix-next/blob/e6c7b396241795bb581a648f751d932640bb9cde/resources/assets/helpers/index.js#L350-L358) for this!

PR #733, opened 2018-03-06
A slight scroll observer adjustment.

mendelB:
### What does this PR do?
This PR fixes an issue where a link to a step on a page would not receive the scroll functionality it so craves.

(A link coming from markdown seems to be coming through as a "POP" action causing us to [skip away from using the scroll functionality](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/helpers/createHistoryHashObserver.js#L52))
[This convo in the slack-o-sphere](https://dosomething.slack.com/archives/C3ASB4204/p1520004652000764)

**Not So Great**:
scrollll

**Fantastico**
scroll2

This doesn't solve the issue on initial page loads with a hash in the [URL](https://www.dosomething.org/us/campaigns/grab-mic/action#step-7bdcOGPUAgKociOgEE4S4g) but it's hopefully a start.

### DIFF
--- a/resources/assets/helpers/createHistoryHashObserver.js
+++ b/resources/assets/helpers/createHistoryHashObserver.js
@@ -49,7 +49,7 @@ export default (history, timeout = 1000) => {
       reset();
     }
 
-    if (action !== 'PUSH') {
+    if (action !== 'PUSH' && location.pathname !== window.location.pathname) {
       return;
     }
 
--- a/resources/assets/helpers/createHistoryHashObserver.js
+++ b/resources/assets/helpers/createHistoryHashObserver.js
@@ -49,10 +49,6 @@ export default (history, timeout = 1000) => {
       reset();
     }
 
-    if (action !== 'PUSH' && location.pathname !== window.location.pathname) {
-      return;
-    }
-
     if (typeof location.hash !== 'string') {
       return;
     }
--- a/resources/assets/helpers/createHistoryHashObserver.js
+++ b/resources/assets/helpers/createHistoryHashObserver.js
@@ -44,7 +44,7 @@ export default (history, timeout = 1000) => {
     return false;
   };
 
-  history.listen((location, action) => {
+  history.listen((location) => {
     if (timeoutId) {
       reset();
     }


DFurnes:
I think part of this problem was unrelated... [that action step](https://app.contentful.com/spaces/81iqaqpfd8fy/entries/5Rw4OWMZ20AQe00ICW8o6Q) was linking to an absolute production URL `https://www.dosomething.org/us/campaigns...`, and so it'd [open in a new tab](https://github.com/DoSomething/phoenix-next/blob/ecf58359378d8ff605a2602d7fa05ea8490ad026/resources/assets/helpers/index.js#L113-L115) on production.

resources/assets/helpers/createHistoryHashObserver.js
```diff
@@ -49,7 +49,7 @@ export default (history, timeout = 1000) => {
       reset();
     }
 
-    if (action !== 'PUSH') {
+    if (action !== 'PUSH' && location.pathname !== window.location.pathname) {
```
DFurnes: Messing around with this a bit on my local & I think we might be able to remove this conditional altogether.

mendelB:
I ended up testing with a relative URL as well and it resulted in the same issue, it seems to use a "POP" action though it's routing to the id on the page.

resources/assets/helpers/createHistoryHashObserver.js
mendelB: Oh, that would be way simpler! I guess cuz it doesn't hit this function unless it's navigating to an id on the same page?

DFurnes: Yeah, it shouldn't as far as I know...

DFurnes:
To make Markdown links scroll nicely, I think we'd want a top-level event listener on the document looking for any click events on `a` tags, seeing if they're on the same domain, and if not handing it off to this same `scrollToElement` helper.

mendelB:
Meaning doing so *instead* of removing that `"PUSH"` conditional?

DFurnes:
From what I can tell, removing the `"PUSH"` conditional doesn't cause these links to scroll smoothly on it's own (which is weird, I would've assumed it would...)

mendelB:
Yeah indeed weird... it's a Stranger Thing.

PR #734, opened 2018-03-07
Adding voter registration action modal with static content

mendelB:
### What does this PR do?
This PR adds a `VoterRegistrationModal` component (along with the proper container, exports, and ModalSwitch `case`).

The modal is hardcoded with static content since [we've decided](https://dosomething.slack.com/archives/C3ASB4204/p1520279950000760) on showing this same modal across all campaigns that are feature-flagged for it to appear.

The actual rendering of this modal under some `TrafficDistribution` and feature flagging will be upcoming pending approval of the ModalLauncher changes #730 

image

### DIFF
--- a/resources/assets/components/Modal/Modal.js
+++ b/resources/assets/components/Modal/Modal.js
@@ -4,7 +4,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Portal from 'react-portal';
 import {
-  POST_SIGNUP_MODAL, POST_SHARE_MODAL, BLOCK_MODAL,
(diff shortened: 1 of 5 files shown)


lkpttn:
Is this a screenshot from the site? We have double Xs to close the modal here.

mendelB:
It's from my local dev, I'm not sure why some modals still have that top-level X. Looking into it

mendelB:
It is no longer! We just had to specify hiding it. I think this was to support modals with no card X ability which may no longer be the case.

lkpttn:
Yeah, all modals should be presented in the carded fashion, so we can look at removing that extra code.

mendelB:
Rebased from master to resolve merge conflicts but no new changes so gonna 🚢!

PR #736, opened 2018-03-07
Disabling experiment participation on all campaigns via a script.

weerd:
### What does this PR do?
This PR adds a generic migration to set all the `allowExperiments` fields on each `campaignSettings` content-type to `false` to disable participation in an experiment. I wanted to try this method out instead of changing each campaign's settings manually, one by one.

I placed this in a dedicated `transforms` directory, since it's not really a migration to run once, but something we could run a few times to clear out experiment participation across campaigns :)

Related to upcoming PR for a Sixpack experiment on the `ledeBanner`.

Hopefully this works!

Refs PID [#155479845](https://www.pivotaltracker.com/story/show/155479845)


weerd:
Yeah, wasn't super keen on that, but all _required_ items 🤷‍♂️

PR #737, opened 2018-03-07
Bug Fix && Content block updates

mendelB:
### What does this PR do?
This PR
- [900d122](https://github.com/DoSomething/phoenix-next/commit/900d1222d87d41c60a218ab4af0f293acd12a7ce) adds a `campaignUpdate` case to our Contentful Entry block parser, since that type is allowed as an `affirmationBlock` field in `ShareActions`
- [f89b4b](https://github.com/DoSomething/phoenix-next/commit/f89b4b22f4ab41280f2921a16f44846388f19688) updates our `ContentModal` to allow for general `block` content, doing the proper entry lookup and using the `ContentfulEntry` component to properly render the `json` fields.
- [d50b170](https://github.com/DoSomething/phoenix-next/commit/d50b1705e58b103c347e93837659f4a2004cce76) updates the `CampaignUpdate` component to not override it's props in the container, by manually performing the entry lookup.

I bumped into this issue trying to test out the post share action affirmation modal, and attaching a `campaignUpdate` as the `affirmationBlock`. This KO'd the app since we didn't parse the block as a CampaignUpdate. Upon fixing this, I bumped into some other issues with campaignUpdates props being overridden in the `CampaignUpdate` component which led to some necessary updates to the `ContentModal` so that we no longer need to do those overrides, and to bring it up to speed with our new 'general block' `ContentfulEntry` abilities.


resources/assets/components/CampaignUpdate/CampaignUpdateContainer.js
```diff
@@ -1,24 +1,17 @@
...
     domain: window.location.origin,
     slug: state.campaign.slug,
     key: props.id,
   };
 
   return {
-    affiliateLogo,
-    author,
-    content,
-    link,
```
weerd: Nice cleanup!

resources/assets/components/Modal/containers/ContentModalContainer.js
```diff
@@ -4,34 +4,13 @@ import ContentModal from '../configurations/ContentModal';
...
-  }
+  const id = state.modal.contentfulId;
 
-  const database = [
-    ...state.campaign.pages,
-    ...state.campaign.activityFeed,
-  ];
+  const json = find(state.campaign.pages, { id })
+    || find(state.campaign.actionSteps, { id })
+    || find(state.campaign.activityFeed, { id });
```
mendelB: We seem to be repeating this entry lookup a bunch. I imagine it might be nice to consolidate them all in a helper function?

DFurnes: I'm into that. 👍

resources/assets/components/Modal/configurations/ContentModal.js
```diff
@@ -1,48 +1,25 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Card from '../../Card';
-import Markdown from '../../Markdown';
-import NotFound from '../../NotFound';
-import { CampaignUpdateContainer } from '../../CampaignUpdate';
+import ContentfulEntry from '../../ContentfulEntry';
+import ModalControls from '../ModalControls';
 
 const ContentModal = (props) => {
```
DFurnes: Do we want to rename this now that it doesn't _just_ hold a CampaignUpdate? `BlockModal`?

resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -42,10 +42,12 @@ const ContentfulEntry = ({ json = DEFAULT_BLOCK, stepIndex, isSignedUp }: Conten
       return (
         <CampaignUpdateContainer
           id={json.id}
+          affiliateLogo={json.fields.affiliateLogo}
           author={json.fields.author}
           content={json.fields.content}
           displayOptions={json.fields.displayOptions}
           link={json.fields.link}
+          bordered={false}
```
weerd: The no bordered will just be for when it appears in a modal right? But in the page feed it's still bordered?

resources/assets/components/Modal/configurations/ContentModal.js
mendelB: I dig.

resources/assets/components/ContentfulEntry/ContentfulEntry.js
mendelB: Hmm, I believe this is the only place that we're rendering a CampaignUpdate from now so it'd always not be bordered

mendelB: I'll try and fix that real quick

weerd: Ok, yeah, cause they should be bordered in the feed... 😬

DFurnes: Could we unify the styles so they look okay in both places?

Maybe a like a `rgba(0, 0, 0, 0.15)` border instead of `#dcdcdc`?

DFurnes: ALSO, LOOK AT THOSE LITTLE GITHUB COLOR PREVIEWS. WOW. 🎨

mendelB: whoah. Yeah unifying would be dope, since it'd be challenging to account for whether it's a modal vs feed cuz unless we want to tamper with the json or start looking at the window location?

weerd: yeah, was thinking maybe that could be an option too so they don't need to have subtle differences in style.

also, the GitHub color previews are so cool! haha!

mendelB: It seems to still look the same, unfortunately:
image

mendelB: Perhaps it's a cop-out or just ingenious you decide. But changing the border to a darker color from within the `modal.scss` file -- only for modal `article`s does the trick!

mendelB:
UPDATES:
- renamed the `ContentModal` to `BlockModal` and adjusted the calls to said modal type all over the place
- added a specific `border-color` rule for `<article>`s in modals to address the border issue. (and removed the prop in `ContentfulEntry` so it'll work as it did before in the feed etc.)
- added a helper method to find contentful entries by id from state. (replacing the three places we were calling this method with the helper call)
- using the `BlockModal` from within the `PostSignupModal` for a quick win 🎐

mendelB:
image

app/Entities/Entity.php
```diff
@@ -68,6 +68,8 @@ public function parseBlock($block)
                 return new LinkAction($block->entry);
             case 'affirmation':
                 return new Affirmation($block->entry);
+            case 'campaignUpdate':
+                return new CampaignUpdate($block->entry);
```
weerd: This was killing me on Teens For Jeans test campaign, haha!

resources/assets/components/Modal/configurations/BlockModal.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ContentfulEntry from '../../ContentfulEntry';
+import ModalControls from '../ModalControls';
+
+const BlockModal = (props) => {
+  const { json, closeModal } = props;
+
+  json.fields.bordered = false;
```
DFurnes: Can we add a `// HACK:` comment here explaining this?

resources/assets/helpers/index.js
```diff
@@ -438,6 +438,20 @@ export function query(key, url = window.location) {
...
 
+/**
+ * Find an entry from within the campaign by given ID param.
+ * (Returns false if not found).
+ *
+ * @param  {Object} state
+ * @param  {String} id
+ * @return {Object|Boolean}
+ */
+export function findContentfulEntry(state, id) {
```
weerd: Noice!

resources/assets/components/Modal/modal.scss
```diff
@@ -31,6 +31,10 @@
     > .block-wrapper {
       margin: 0;
     }
+
+    article {
+      border-color: $black;
```
DFurnes: What's this for? Can we add a comment, or a more descriptive class to hook onto?

resources/assets/helpers/index.js
```diff
@@ -438,6 +438,20 @@ export function query(key, url = window.location) {
   return search[key];
 }
 
+/**
+ * Find an entry from within the campaign by given ID param.
+ * (Returns false if not found).
```
DFurnes: Is this accurate? It should return `undefined` or `null`, ideally.

resources/assets/components/Modal/configurations/BlockModal.js
weerd: oh yeah, forgot about this.  @mendelB I think lets go with @DFurnes suggestion of using rgba value and not forcing removing the border. I'd rather get into a pattern of finding design solution that appease the blocks showing up in Modals and on the Page without changes like this if possible :)

resources/assets/helpers/index.js
mendelB: I think `find` itself returns `undefined` but since we have the `||` operators it renders as a boolean, or truthy value if found.

Should we change this to be found value or `undefined` if not found?

resources/assets/components/Modal/configurations/BlockModal.js
mendelB: Ugh that was left in there from when I was tinkering around 🤒

mendelB: https://github.com/DoSomething/phoenix-next/pull/737#discussion_r172987979

resources/assets/components/Modal/modal.scss
mendelB: It's the solution to this: https://github.com/DoSomething/phoenix-next/pull/737#discussion_r172987979

I'll add in a clarifying comment!

resources/assets/helpers/index.js
DFurnes: Can you double-check that with a `console.log` or testing on the console? If all of the pieces of this boolean expression return `undefined`, it _shooooould_ return the last value as far as I know.

resources/assets/components/Modal/modal.scss
DFurnes: Oh I see that now! Let's definitely add a `// HACK:` describing that & we can revisit sometime later.

resources/assets/components/ContentfulEntry/ContentfulEntry.js
weerd: Oh gotcha. Yeah, it's because it's a transparent value but the background it's transparent on is the card (which is white) and not the modal.

resources/assets/helpers/index.js
mendelB: You're absolutely right, it does return `undefined`!
Here's what tripped me up:
image

resources/assets/components/ContentfulEntry/ContentfulEntry.js
DFurnes: Try `background-clip: padding-box` on the card!

mendelB: By Jove! That works!

mendelB:
UPDATES:
- return value spec in docblock for `findContentfulEntry`
- use `background-clip: border-box` and dave's magic rgba value for `.bordered` instead of a modal hack
- remove a thing

resources/assets/helpers/index.js
DFurnes: I think the console is tricking you into using this [window.find](https://developer.mozilla.org/en-US/docs/Web/API/Window/find) method instead!

mendelB: Aha! That makes sense, and is something to bear in mind for the future!

weerd:
> use background-clip: border-box

Oh yeah! Nice 💃

PR #739, opened 2018-03-08
Updates and fixes

weerd:
### What does this PR do?
This PR updates some entities to correctly grab the `type` as a string for an entity instead of grabbing a complete object (see images below). It also updates a few straggling `SignupButtonFactory`s that remained, so it's now officially wiped!

Entity type as an object (incorrect)
screen shot 2018-03-07 at 5 24 56 pm

Entity type as a string (correct)
screen shot 2018-03-07 at 5 24 24 pm

- 5260b1a Updates entities to grab `type` as a string.
- 79e70f8 Removes the last straggling `SignupButtonFactory`s components in favor of `SignupButton` (missing `Revealer`, but @DFurnes is on that!)

### DIFF
--- a/resources/assets/components/ContentfulEntry/renderers.js
+++ b/resources/assets/components/ContentfulEntry/renderers.js
@@ -46,7 +46,7 @@ export function renderPhotoUploader(step, isSignedUp) {
     <div key="reportback_uploader" className="margin-bottom-lg">
       <PuckWaypoint name="photo_uploader_action-top" />
       <div className="margin-horizontal-md">
-        <ReportbackUploaderContainer actionType={step.type.sys.id} {...step.fields} />
+        <ReportbackUploaderContainer actionType={step.type} {...step.fields} />
       </div>
       <PuckWaypoint name="photo_uploader_action-bottom" />
     </div>
--- a/resources/assets/components/Feed/Feed.js
+++ b/resources/assets/components/Feed/Feed.js
@@ -38,8 +38,8 @@ const renderFeedItem = (block, index) => (
  * @returns {XML}
  */
 const Feed = (props) => {
-  const { actionText, blocks, callToAction, campaignId, dashboard, signedUp, hasPendingSignup,
-    isAuthenticated, canLoadMorePages, clickedViewMore, clickedSignUp } = props;
+  const { blocks, callToAction, dashboard, signedUp, hasPendingSignup,
+    isAuthenticated, canLoadMorePages, clickedViewMore } = props;
 
   const shouldShowRevealer = (isAuthenticated && ! signedUp) || canLoadMorePages;
   const revealer = (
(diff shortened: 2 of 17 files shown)


resources/assets/components/Feed/Feed.js
```diff
@@ -38,17 +38,16 @@ const renderFeedItem = (block, index) => (
  * @returns {XML}
  */
 const Feed = (props) => {
-  const { actionText, blocks, callToAction, campaignId, dashboard, signedUp, hasPendingSignup,
-    isAuthenticated, canLoadMorePages, clickedViewMore, clickedSignUp } = props;
+  const { blocks, callToAction, dashboard, signedUp, hasPendingSignup,
+    isAuthenticated, canLoadMorePages, clickedViewMore } = props;
 
-  const viewMoreOrSignup = signedUp ? clickedViewMore : () => clickedSignUp(campaignId);
+  const shouldShowRevealer = (isAuthenticated && ! signedUp) || canLoadMorePages;
```
mendelB: Now that we're only showing 'view more' in the revealer and not the `actionText` (and not running a signup conversion on click in such a case), we probably don't need to show the revealer button in the case of `(isAuthenticated && ! signedUp)`, right?

resources/assets/components/ContentfulEntry/renderers.js
```diff
@@ -46,7 +46,7 @@ export function renderPhotoUploader(step, isSignedUp) {
     <div key="reportback_uploader" className="margin-bottom-lg">
       <PuckWaypoint name="photo_uploader_action-top" />
       <div className="margin-horizontal-md">
-        <ReportbackUploaderContainer actionType={step.type.sys.id} {...step.fields} />
+        <ReportbackUploaderContainer actionType={step.type} {...step.fields} />
```
mendelB: I believe we could also be rid of the `get(json, 'type.sys.id')` in the [`parseContentfulType` helper](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/helpers/index.js#L453).

resources/assets/components/Quiz/Quiz.js
```diff
@@ -59,22 +58,11 @@ const Quiz = (props) => {
 
   const showResultingAction = () => {
     const action = find(fields.results, { id: selectedResult });
-
-    const actionProps = {
-      ...action.fields,
-      content: `${fields.conclusion}\n${action.fields.content}`,
```
mendelB: I think we need this left in there so that we combine the content from the action with the dynamic result from the quiz.

mendelB:
Ooh I'm late

DFurnes:
Haha we were _reckless_! 😈

resources/assets/components/Feed/Feed.js
DFurnes: The "signup button" is now built-in to the revealer, if the user isn't already signed up!

mendelB:
Livin' on the edge 🥂

resources/assets/components/ContentfulEntry/renderers.js
DFurnes: I think you're right!

resources/assets/components/Feed/Feed.js
mendelB: Oh oh oh I see that now. Dope!

PR #740, opened 2018-03-08
Lede banner design sixpack experiment

weerd:
### What does this PR do?
This PR adds a Sixpack A/B test for the `LedeBanner` to showcase the new "cover" design as an alternative test in a lede banner experiment. This experiment will affect _ALL_ campaigns, so the prior script in #736 removed experiment participation in all campaigns and I only re-enabled it for _Grab The Mic_.

Alternative A (default - mosaic)
screen shot 2018-03-08 at 2 24 10 pm

Alternative B (cover)
screen shot 2018-03-08 at 2 24 21 pm

Fixes [Pivotal ID #155479845](https://www.pivotaltracker.com/story/show/155479845)

### DIFF
--- a/resources/assets/components/Experiment/Experiment.js
+++ b/resources/assets/components/Experiment/Experiment.js
@@ -1,8 +1,9 @@
 /* global window */
 
 import React from 'react';
-import PropTypes from 'prop-types';
 import { get } from 'lodash';
+import PropTypes from 'prop-types';
+
 import { assertTestPasses } from '../../helpers/experiments';
 
 class Experiment extends React.Component {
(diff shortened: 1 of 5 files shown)


weerd:
> Does the experiment default to the control alternative if the user is signed up? And if not -- do we want to perhaps add a preTest conditional to ensure they're not signed up so that anyone already signed up for grab-mic doesn't get a whole new lede banner, or are we cool with that happening?

So I've only applied this to the landing page. It's not something that will continue past into the action page for a campaign which was my assumption, but I will clarify with @lkpttn and @ashleybaldwin

weerd:
Haha, once the experiment is done we can do what you did in #741

PR #741, opened 2018-03-08
Landing page sweeeeep

mendelB:
## _PULL REQUEST OVERVIEW_

I was doing some work on the Landing Page, and could not resist the temptation to sweep the heck out of it and utilize the `LedeBannerContainer` instead of all those repetitive props. It's literally the same exact thing, just added in two new props to the LedeBanner container that I guess before would only show up in the LedeBanner on the landing page but were only relevant to the Signup button, and thus only admins would've seen the disparity.


mendelB:
omg I'm an idiot. #740. But we should do this eventually!

PR #742, opened 2018-03-09
Sticky CTA on landing page

mendelB:
### What does this PR do?
This PR 
- adds a sticky CTA on the Landing Page
- adds a new utility class in `base.scss` called `hide-on-mobile`. Guess what it does?

Pivotal ID [#155401683](https://www.pivotaltracker.com/story/show/155401683)

sticky-cta

### DIFF
--- a/resources/assets/components/Page/LandingPage/LandingPage.js
+++ b/resources/assets/components/Page/LandingPage/LandingPage.js
@@ -65,11 +65,16 @@ const LandingPage = (props) => {
         <Enclosure className="default-container margin-lg pitch-landing-page">
           <PitchTemplate pitchContent={pitchContent} sidebarCTA={sidebarCTA} />
         </Enclosure>
+
+        <CallToActionContainer
+          className="-sticky"
+          content={tagline}
+        />
       </div>
 
       <PuckWaypoint name="landing_page_cta-top" />
       <CallToActionContainer
-        className="legacy border-top border-radius-none bg-off-white padding-lg"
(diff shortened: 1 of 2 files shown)


DFurnes:
> adds a new utility class in `base.scss` called `hide-on-mobile`. Guess what it does?

think think think

resources/assets/components/Page/LandingPage/LandingPage.js
```diff
@@ -65,11 +65,16 @@ const LandingPage = (props) => {
         <Enclosure className="default-container margin-lg pitch-landing-page">
           <PitchTemplate pitchContent={pitchContent} sidebarCTA={sidebarCTA} />
         </Enclosure>
+
+        <CallToActionContainer
+          className="-sticky"
```
DFurnes: I wonder if we should just make a boolean `sticky` prop that sets this class... e.g:

```jsx
<CallToActionContainer content={tagline} sticky />
```

That way other components don't need to know the right classname to enable this.

mendelB: That seems like a solid idea!

mendelB:
UPDATE:
Added the `sticky` prop to CallToAction component, to determine `-sticky` class name

DFurnes:
Can you swap out [the other places we manually set this class](https://github.com/DoSomething/phoenix-next/search?utf8=✓&q=%22-sticky%22&type=) as well? After that, 👍 👍 to merge!

mendelB:
Ach my bad. I don't know why I assumed those were the only places and didn't take a second to do a search 🤦‍♂️

PR #744, opened 2018-03-09
Moving developer onbaording to docs

weerd:
### What does this PR do?
This PR move the Developer Onboarding document from the wiki over into our docs directory for display in GitBooks.

Also fixes link for `editLink` plugin, which did not get updated when we switched to having a `master` branch!

### DIFF
--- a/docs/development/developer_onboarding.md
+++ b/docs/development/developer_onboarding.md
@@ -39,7 +39,7 @@ There are a couple of other applications and sites that you will want access to 
 
 ## Tips and Tidbits 🍩
 
-### Staging Environment
+### Preview Environment
 Once a _Pull Request_ is merged, a deploy to our staging environment is automatically triggered.
 
 The staging environment can be accessed at [https://phoenix-preview.dosomething.org](https://phoenix-preview.dosomething.org).
--- a/docs/development/developer_onboarding.md
+++ b/docs/development/developer_onboarding.md
@@ -7,87 +7,21 @@ However, before running `php artisan phoenix:setup`, you are going to need a few
 
 
 ### Northstar
-For the most part, you can use the staging **Northstar** URL for your development environment: `https://northstar-thor.dosomething.org`
+For the most part, you can use the staging Northstar URL for your development environment: `https://northstar-thor.dosomething.org`
 
 To gain access to an ID and SECRET, you will need to have the proper permissions set on your DoSomething account. Let a developer on the team know if you are unable to access [Aurora (staging)](https://aurora-thor.dosomething.org/clients).
 
 
 ### Contentful
-Be sure to setup your [Contentful account](https://www.contentful.com/sign-up/) and request access to the project **Phoenix** project space, so that you can set the proper _Contentful Space ID_ and _Contentful API Key_ in your setup.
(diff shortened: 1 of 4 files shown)


docs/development/developer_onboarding.md
```diff
@@ -0,0 +1,114 @@
...
+### Contentful
+Be sure to setup your [Contentful account](https://www.contentful.com/sign-up/) and request access to the project **Phoenix** project space, so that you can set the proper _Contentful Space ID_ and _Contentful API Key_ in your setup.
+
+Once you gain access to Contentful and the **Phoenix** space, you can find the values you need in the **Space settings** tab and click on **API Keys** in Contentful. Then select the **Phoenix Next - local development** token.
+
+
+### .ENV (noun: /Dot EE EN VEE/)
+Here's an example of how a typical `.env` file might look for a developer (with `random-string` replacing any auth data):
+
+```
```
DFurnes: Can we just link to `.env.example` here so there's less chance of this getting out of sync?

docs/development/developer_onboarding.md
```diff
@@ -0,0 +1,114 @@
...
+### Access and Permissions
+There are a couple of other applications and sites that you will want access to for development:
+
+- The [Dosomething Team](https://dashboard.heroku.com/teams/dosomething/overview) on [Heroku](https://www.heroku.com/) (To access Preview Apps, and more).
+- The [Dosomething Organization](https://app.wercker.com/dosomething) on [Wercker](https://app.wercker.com) (To access and re-trigger auto builds for _Pull Requests_).
+
+
+## Tips and Tidbits 🍩
+
+### Staging Environment
```
DFurnes: Can we call this "QA Environment"?

weerd: Is that the official? I thought it was `Staging`?

DFurnes: I thought [we'd settled on QA](https://git.io/vAhcT) but I'm open to "staging" if people feel that's a better choice. Let's just make sure to document whatever we've settled on in that issue if it's changed.

weerd: > QA: app named northstar-qa, and URL profile-qa.dosomething.org

Oh yeah... but we also went with `phoenix-preview.ds.org`.... 🤔

weerd: I'm so confused now, lol 😅

DFurnes: My only preference is that we name environments consistently between applications, so that should be `phoenix-qa.dosomething.org`.

weerd: ok, so we could also leave `phoenix-preview.ds.org` for campaign leads and internal but for devs it's `phoenix-qa.ds.org`? I really don't want to have to switch it _again_ on campaign leads since we switched them from `next-thor.ds.org` @_@

DFurnes: To be fair, I think we could just set up a domain redirect there. I'd also really like to double-down on that "Preview" button (once we can figure out this [URL template kerfuffle](https://www.contentfulcommunity.com/t/malformed-content-preview-urls/719)) and that'll drastically cut down on leads actually typing URLs by hand.

weerd:
@DFurnes I renamed the title from `Staging` to `Preview` to match the url in the text that follows. I can update this once we actually have a `phoenix-qa.ds.org` since that doesn't currently exist.

PR #745, opened 2018-03-09
Activate voter registration timed modal!

mendelB:
### What does this PR do?
This PR activates the VoterRegistrationAction timed modal!

- 🎏 added a `featureFlags` prop to the `Campaign` component and set up the Voter Registration Timed Modal in a `ModalLauncher` if the `showVoterRegistrationModal` feature flag is set to `true`. (set in Additional Content in Contentful).
- 🖲 set up tracking in local storage for when a user clicks the VoterRegistrationAction button (sets a hide voter reg modal value), or when the user dismisses the Voter Registration Timed Modal (sets a dismissed voter reg modal value)
- ✅ set the `.env` variable for the Voter Registration Modal feature `VOTER_REG_MODAL_ENABLED=true`, and setting things up in `config/services` (added some slight changes there to group all timed modal variables together) and in the View Composer.
- 🆔 set a static `contentfulId` prop in the VoterRegistrationModal's render of the `VoterRegistrationModalAction` for puck tracking data.
- 🗒 added a test case for localstorage tracking to the `VoterRegistrationAction`
- 🍪added Dave style emoji's to PR description to maintain sanity and morale.

#730 
#734 

### Before Merge
- [x] add `VOTER_REG_MODAL_ENABLED=true` to staging variables config

### Before Deploy
- [ ] add `VOTER_REG_MODAL_ENABLED=true` to production variables config


weerd:
> 🍪added Dave style emoji's to PR description to maintain sanity and morale. 

😏

weerd:
@mendelB mind quickly fixing the merge conflict ^^

mendelB:
All conflicts resolved! Send me to the middle east why don't you.

weerd:
image

app/Providers/AppServiceProvider.php
```diff
@@ -45,7 +45,8 @@ public function boot(Router $router)
                 'SIXPACK_COOKIE_PREFIX' => config('services.sixpack.prefix'),
                 'SIXPACK_ENABLED' => config('services.sixpack.enabled'),
                 'SIXPACK_TIMEOUT' => config('services.sixpack.timeout'),
-                'NPS_SURVEY_ENABLED' => config('services.nps_survey.enabled'),
+                'NPS_SURVEY_ENABLED' => config('services.timed_modals.nps_survey.enabled'),
+                'VOTER_REG_MODAL_ENABLED' => config('services.timed_modals.voter_reg_modal.enabled'),
```
weerd: May eventually want to rethink a bit how we do these. Would rather not have to keep adding env vars for each new modal we include. Fine for now, but I feel there's a better alternative™ 🤔

mendelB: Yeah, that's a really fair point! Was thinking about this in terms of a way to shut off a timed modal across *all* campaigns as opposed to individual Campaign Settings in Contentful, but it's just kinda awkward to be needing a new ENV var for each one. Is there some alternative way, maybe like the contentful migration route like the [experiments disable one](https://github.com/DoSomething/phoenix-next/blob/master/contentful/transforms/disable_experiments_on_all_campaign_settings.js)?

resources/assets/components/Modal/containers/VoterRegistrationModalContainer.js
```diff
@@ -1,6 +1,14 @@
 import { connect } from 'react-redux';
 import VoterRegistrationModal from '../configurations/VoterRegistrationModal';
 import { closeModal } from '../../../actions/modal';
+import { getUserId } from '../../../selectors/user';
+
+/**
+ * Provide state from the Redux store as props for this component.
+ */
+const mapStateToProps = state => ({
+  northstarId: getUserId(state),
```
weerd: Don't we normally just call this `userId` in other component containers?

resources/assets/components/actions/VoterRegistrationAction/VoterRegistrationAction.test.js
```diff
@@ -24,7 +29,14 @@ test('VoterRegistrationAction is rendered as a card component with a button', ()
   expect(wrapper.find('.button').length).toEqual(1);
 });
 
-test('VoterRegistrationAction calls the event tracker prop function when the button is clicked', () => {
+describe('clicking the VoterRegistrationAction button', () => {
   wrapper.find('.button').simulate('click');
-  expect(trackEventMock).toHaveBeenCalled();
+
+  it('calls the event tracker prop function', () => {
```
weerd: Random, but thoughts on using `test()` instead of `it()`.... I feel `it()` drives you to write the sentence for the test in a very specific fashion that doesn't always work great with the test. Thoughts? 🤔

resources/assets/components/Modal/containers/VoterRegistrationModalContainer.js
mendelB: Is it not better to use the selector or should we always just grab it directly from state as we have been doing generally?

app/Providers/AppServiceProvider.php
weerd: Yeah potentially that, or even having a `timedModals` multiple ref field on a campaign (or page?) and actually having a `timedModal` content type which takes a block and some other info, like the amount of time to wait, etc. 🤔

resources/assets/components/actions/VoterRegistrationAction/VoterRegistrationAction.test.js
mendelB: Hmm I see what you mean. Sometimes it feels so elegant to use 'it' and it has a nice flow in the test output, but I can see it being too forced occasionally. But I guess I'm not really sure what better language to use here? Is the flow from the `describe` description to the `it` description not working?

weerd: The `describe` I think still works, but gives more options for how you end up finishing the sentence to describe the test. For example, one I'm working on now is `describe('TextSubmissionAction component')...` -> `test('is rendered as a card component with a form and submit button'...)`. More of a semantic thing, but leaves it more open and also feels more obvious _what_ the test actually is... since it's wrapped in `test()` :)

mendelB: lol I hear ya, makes sense. 
(Though I think you may be offending someone  here)
07-pennywise-it-1990 w710 h473

app/Providers/AppServiceProvider.php
mendelB: How would that affect something like Survey that we want site-wide?

resources/assets/components/actions/VoterRegistrationAction/VoterRegistrationAction.test.js
weerd: I will now be forever haunted in my dreams 😱

app/Providers/AppServiceProvider.php
mendelB: Or have we considered some Contentful Content Type that is like a site wide or cross campaign affect setting?

resources/assets/components/actions/VoterRegistrationAction/VoterRegistrationAction.test.js
mendelB: I'll put in a good word for you. He won't be on your case. (We do Svedka shots together on weekends sometimes)

resources/assets/components/Modal/containers/VoterRegistrationModalContainer.js
weerd: oh I just mean rename it from `northstarId` to `userId`.

mendelB: ohhh lol dunno why I did it like that. Yeah I'll update!

app/Providers/AppServiceProvider.php
weerd: > Or have we considered some Contentful Content Type that is like a site wide or cross campaign affect setting?

Yeah, maybe something like this. I think once we get to something like a site-wide configuration setting, then we can chat how to refactor this down the line. But that's a very good point. Forgot about the site-wide modals.

PR #754, opened 2018-03-13
Hide member gallery if the user is not signed up

mendelB:
### What does this PR do?
This PR hides the Member Gallery if the user is not signed up for the campaign.
[Slack](https://dosomething.slack.com/archives/C3ASB4204/p1520948094000357)

### DIFF
--- a/resources/assets/components/Page/ActionPage/ActionSteps.js
+++ b/resources/assets/components/Page/ActionPage/ActionSteps.js
@@ -98,7 +98,7 @@ const ActionSteps = (props) => {
     ));
   }
 
-  if (template === 'legacy' || ! hasActivityFeed) {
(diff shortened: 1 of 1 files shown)


resources/assets/components/Page/ActionPage/ActionSteps.js
```diff
@@ -98,7 +98,7 @@ const ActionSteps = (props) => {
     ));
   }
 
-  if (template === 'legacy' || ! hasActivityFeed) {
+  if (isSignedUp && (template === 'legacy' || ! hasActivityFeed)) {
     stepComponents.push(renderLegacyGallery());
```
DFurnes: Oh, this just reminded me: let's make a card to make this into a "normal" block that could be placed anywhere that we accept `ContentfulEntry` components.

PR #758, opened 2018-03-14
Contentful Migration - LinkAction `buttonText` field

mendelB:
## _PULL REQUEST OVERVIEW_

image

### What does this PR do?
This PR adds a contentful migration that creates an optional `buttonText` field for the Link Action Contentful model.

This will be an optional field that - if filled out, will cause a button link to appear in the link action instead of the embed link.

[Pivotal #155866139](https://www.pivotaltracker.com/story/show/155866139)

### DIFF
--- a/contentful/migrations/2018_03_14_001_add_button_text_field_to_link_action.js
+++ b/contentful/migrations/2018_03_14_001_add_button_text_field_to_link_action.js
@@ -1,0 +1,10 @@
+module.exports = function (migration) {
+  const linkAction = migration.editContentType('linkAction');
+
+  linkAction.createField('buttonText')
+    .name('Button Text')
+    .type('Symbol')
+    .required(false);
+
+  linkAction.moveField('buttonText').afterField('link');
+}


mendelB:
Yeah, I feel like it's cool since as you said it's scoped to the LinkAction which is safe from the actionText override.

DFurnes:
Yeah, that seems like good reasoning. I'm happy with `buttonText`. 👍

contentful/migrations/2018_03_14_001_add_button_text_field_to_link_action.js
```diff
@@ -0,0 +1,10 @@
+module.exports = function (migration) {
+  const linkAction = migration.editContentType('linkAction');
+
+  linkAction.createField('buttonText')
+    .name('Button Text')
+    .type('Symbol')
+    .required(false);
```
DFurnes: Let's add some help text after running this to explain that if this field is left blank, we just won't show a button.

PR #759, opened 2018-03-14
Remove share prototype.

DFurnes:
### What does this PR do?
We're handling this with standard blocks now! 🔥

This was a temporary prototype to figure out whether a better SMS-to-web flow would boost reportbacks. It turns out it did!

N/A

### Checklist


mendelB:
And @DFurnes just confirming: we can just direct to a `/blocks/share-action-id` and we have all that functionality baked in right?

weerd:
@mendelB awesome gif 😉

DFurnes:
> And @DFurnes just confirming: we can just direct to a `/blocks/share-action-id?user_id=...` and we have all that functionality baked in right?

Correct! The new flow would be to create a "Share Action" w/ the desired URL and link to that. We can use the `?user_id=` query parameter there (and anywhere else!) to "partially" authenticate a user so that they'd send their Northstar ID along to Puck without having to fully auth.

The extra nice thing is that we can start to think about doing this exact same thing for any other block - so directly linking to the "Voter Registration block" or a "Campaign Update" or anything else too!

PR #760, opened 2018-03-14
Link action button text

mendelB:
### What does this PR do?
This PR adds the `buttonText` field to the `LinkAction` component.

- adding `buttonText` field to LinkAction Contentful Entry
- adding prop to `LinkAction` replacing the Embed link with a button link if it's filled out

**SMALL**
https://user-images.githubusercontent.com/12417657/37425957-900d1622-279b-11e8-95f9-7846fe2e2607.png

**MEDIUM**
https://user-images.githubusercontent.com/12417657/37425930-80f3945e-279b-11e8-8b2f-b0c2c289c49c.png

**LARGE**
https://user-images.githubusercontent.com/12417657/37425907-6f1df8b4-279b-11e8-97d8-da0a158e4814.png

[Pivotal #155866139](https://www.pivotaltracker.com/story/show/155866139)
#758 

### Checklist
- [X] Documentation added for new features/changed endpoints.
- [X] Added screenshot to PR description of related front-end updates on **small** screens.
- [X] Added screenshot to PR description of related front-end updates on **medium** screens.
- [X] Added screenshot to PR description of related front-end updates on **large** screens.


docs/content-publishing/actions/link-action.md
```diff
@@ -15,4 +19,5 @@ The Link Action consists of three fields:
 - **title (required)**: The title that will show up in the yellow bar atop the Link Action.
 - **content (optional)**: content in Markdown format that will appear within the card atop the link.
 - **link (required)**: a valid URL which will be embedded within the card.
+- **buttonText (required)**: Text for button to link which will replace the embedded link.
```
DFurnes: This should be **(optional)** I think!

resources/assets/components/actions/LinkAction/LinkAction.js
```diff
@@ -29,9 +34,11 @@ const LinkAction = (props) => {
           <Markdown className="padded">{content}</Markdown>
           : null }
 
-        <div role="button" tabIndex="0" onClick={onLinkClick} className="link-wrapper">
-          <Embed className="padded" url={link} />
-        </div>
+        { ! buttonText ? (
```
DFurnes: Subjective, but does this read better the other way around? I'm afraid the `!` gets a little lost if skimming.

```jsx
{ buttonText ? (
  null
) : (
  <div role="button" tabIndex="0" onClick={onLinkClick} className="link-wrapper">
    <Embed className="padded" url={link} />
  </div>
) }
```

resources/assets/components/actions/LinkAction/link-action.scss
```diff
@@ -0,0 +1,9 @@
+// @import '../../../scss/next-toolbox.scss';
+
+.link-action {
+  .button {
```
DFurnes: Can we target this with a `.link-button` selector instead so we don't add specificity? (Or can we combine this with other "card buttons" like this as a standard component that we could import here?)

docs/content-publishing/actions/link-action.md
mendelB: gah yes!

resources/assets/components/actions/LinkAction/LinkAction.js
mendelB: I'm dramatic so I love exclamation points on stuff. but yes I'll switch lol

resources/assets/components/actions/LinkAction/link-action.scss
mendelB: Sorry, what if we did the former actually? I realized the reason I used the specificity was bc otherwise the general `.button` rules would override the rules here. But what if we do call it `link-button`, and nest it under `.card` in `/card.scss` instead, is that an appropriate specificity application?

DFurnes: Ah, was hoping we'd be able to avoid that. If it's already blah, then I'm fine going along with a `.button .card-button` selector to give it the needed specificity.

weerd: `.button .button-attached`? all buttons in cards are technically a "card button". This is just a special variation that attaches to the bottom and is full width.

mendelB: gotcha. It does indeed seem to be the only way to avoid the default `.button` rules 😞

mendelB:
EDIT: added a `.button.card-button` rule to `card.scss` and replaced some points we use those rules with the class

PR #765, opened 2018-03-15
Embed patch

mendelB:
### What does this PR do?
This PR
- adds a custom dispatch configuration to the `Embed` object we create, so that we timeout requests way earlier.
- adding a `catch` to the `Embed` construction so that we can cache bad requests and timeouts (we catch the thrown exception, and cache the error message so that we don't repeatedly try this request again in a short time span)
- adding a case to the `Embed` component on the front end, to simply output a link in case the request fails.

This is a super quick patch so we can fix this and not lost RBs for team-dream tonight!

[#156024745](https://www.pivotaltracker.com/story/show/156024745)
[Slack - this message above and below for context](https://dosomething.slack.com/archives/C2BPA7M8F/p1521146734000553)

### DIFF
--- a/app/Http/Controllers/EmbedController.php
+++ b/app/Http/Controllers/EmbedController.php
@@ -3,6 +3,7 @@
 namespace App\Http\Controllers;
 
 use Embed\Embed;
+use Embed\Http\CurlDispatcher;
 use Illuminate\Http\Request;
 use Illuminate\Http\JsonResponse;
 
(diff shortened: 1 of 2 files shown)


weerd:
> so that we can cache bad requests and timeouts

"cache" or "catch"?

weerd:
ah, the code explained! it was "cache" 💃

mendelB:
>"cache" or "catch"?

Jeepers that's a confusing sentence!

"Catching" the Exception, so that we can "cache" the request.

Issue #769, opened 2018-03-16
[Feature] Implement Text Submission Action reportback post to Rogue


weerd:
### User Story
As team rocket, we want to connect the text submission reportback post to the V3 endpoint in rogue.

#### Additional Information
_Any other info or references to cards/tickets from other services (Trello, etc)._
Pivotal ID: [#155866202](https://www.pivotaltracker.com/story/show/155866202)

### Why This Matters
This feature will allow our members to submit a text only reportback post to Rogue, expanding the number of available actions our members can perform on the platform.

### Tasks
This feature includes the following tasks:
- [x] Adding text submission action React component #756 
- [x] Adding migration to create text submission action content type #766 
- [x] Add entity in backend to transform text submission action instance #770 
- [x] Connecting text submission action entity to component #771
- [x] Text submission action request to Phoenix API #782 
- [x] Handling multiple action storing #788 
- [x] Connect text submission action to API middleware and submit post to Rogue #796 
- [x] Update text submission action to react to form submission, showing validation #803 & #806 
- [x] Show affirmation for all successful submissions #807 
- [x] Updates the exception messages, and fixes action bucket id #811

PR #771, opened 2018-03-16
Connecting text submission action entity to component

weerd:
### What does this PR do?
This PR connects the data from the `TextSubmissionAction` entity, over into the `ContentfulEntry` component and associated renderer.

It also alphabetizes the list of cases when deciding which renderer to use 😉 

I also disabled the button on the `TextSubmissionAction` since it _could_ accidentally get added to a campaign page after this PR is merged and want it to be obvious that it can't be used just yet!

image

Refs [Pivotal Tracker ID #155866202](https://www.pivotaltracker.com/story/show/155866202)
Refs #769

### DIFF
--- a/resources/assets/components/ContentfulEntry/renderers.js
+++ b/resources/assets/components/ContentfulEntry/renderers.js
@@ -198,7 +198,7 @@ export function renderTextSubmissionAction(data) {
     <div key={`text-submission-action-${contentfulId}`} className="margin-horizontal-md">
       <PuckWaypoint name="text_submission_action-top" waypointData={{ contentfulId }} />
       <TextSubmissionActionContainer {...data.fields} />
-      <PuckWaypoint name="text_submission_action-top" waypointData={{ contentfulId }} />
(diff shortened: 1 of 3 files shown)


resources/assets/components/ContentfulEntry/renderers.js
```diff
@@ -183,6 +185,24 @@ export function renderLinkAction(step) {
...
+ * @return {Component}
+ */
+export function renderTextSubmissionAction(data) {
+  const contentfulId = data.id;
+
+  return (
+    <div key={`text-submission-action-${contentfulId}`} className="margin-horizontal-md">
+      <PuckWaypoint name="text_submission_action-top" waypointData={{ contentfulId }} />
+      <TextSubmissionActionContainer {...data.fields} />
+      <PuckWaypoint name="text_submission_action-top" waypointData={{ contentfulId }} />
```
mendelB: `text_submission_action-top` -> `text_submission_action-bottom` 🔝🎩

weerd: Good catch! Copy-pasta mistake 😬

PR #775, opened 2018-03-21
Updating the Cover template copy

weerd:
### What does this PR do?
This PR makes a quick copy update to the Cover template experiment copy.


weerd:
Not sure... it's the copy I was given, but good 👁 

@ashleybaldwin grammar check?

Issue #780, opened 2018-03-22
Create New Refactored Quiz


mendelB:
### User Story
As a developer, I want a new Quiz component, content type, and Contentful Model in a formalized more sustainable format, using the updated design mocks.

#### Additional Information
Pivotal ID: [#155783121](https://www.pivotaltracker.com/story/show/155783121)

### Why This Matters
This feature will allow us to create more quizzes in a reliable manner, that convert well.

### Tasks
This feature includes the following tasks:
- [x] Create a fresh new Quiz component #773 #774 
- [x] Create new Quiz Contentful Model/Content-Type and Entry  #778 #779

PR #773, opened 2018-03-19
adding new alternative quiz component

mendelB:
### What does this PR do?
This PR adds a new alternative Quiz component suite featuring two primary changes for now from the legacy Quiz:

- quiz choice selections stored in local state, enabling a way less cluttered layout with no reducers or actions necessary.
- visual changes based on the design specs (attached to the pivotal card)
- also, instead of displaying an error when the quiz is submitted incomplete, we just disable the submit button until all questions are answered 💡 

Beyond that, it's more or less following a similar format with the main `QuizAlt` component that renders `QuizQuestion`s which each render `QuizChoice`s, along with a top-level `QuizConclusion` which is used as the CTA/Quiz submission button. 

The main thing missing at this point is the award system based on the selections, as well as the corresponding Action. 

📸 
[**Small**](https://user-images.githubusercontent.com/12417657/37618065-44483afc-2b8b-11e8-9748-116bb37c9380.png)

[**Medium**](https://user-images.githubusercontent.com/12417657/37616432-4430b7c4-2b86-11e8-81cf-958e6b1c17c5.png)

[**Large** incomplete quiz](https://user-images.githubusercontent.com/12417657/37616361-fb045aa6-2b85-11e8-8d2a-0ec0a013117b.png)
[**Large** complete quiz](https://user-images.githubusercontent.com/12417657/37616378-18588942-2b86-11e8-9c6f-a0bb9c889c1b.png)

### What are the relevant tickets/cards?
[Pivotal ref #155783121](https://www.pivotaltracker.com/story/show/155783121)

### Checklist
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.
- [x] Added screenshot to PR description of related front-end updates on **small** screens.
- [x] Added screenshot to PR description of related front-end updates on **medium** screens.
- [x] Added screenshot to PR description of related front-end updates on **large** screens.

PR #784, opened 2018-03-23
Major quiz updates

mendelB:
### What does this PR do?
This PR majorly updates the new Quiz component to be at feature parity with the legacy quiz (QuizBeta/`LegacyQuiz`).

- moving some logic over to the `QuizContainer` to have it pass the proper props to the Quiz from the `additionalContent` field and the others, and also doing some work to try and find the quiz based on the routers `slug` param (`.../quiz/:quizSlug`), and returning a `NotFound` component if a quiz with the said slug is non-existent - [SEE UPDATE](https://github.com/DoSomething/phoenix-next/pull/784#issuecomment-376288552)
- adding a `quiz-helper` file with a `calculateResult` function that can parse through the user selections at the end of the quiz and return the winning result and result block to be displayed to the user
- added some tests (not exhaustive but a healthy start. Still need some testing for the container that's doing a bit of heavy lifting of its own now.)

This more or less brings the quiz up to par with how it was in a hopefully much cleaner and more sustainable format. Main thing remaining now is to 
- formalize the Quiz Content Type fields in contentful
- wrap the heavy JSON fields (`questions`, `choices`, etc.) in a nice Contentful SDK GUI to make it easier to edit. 
And importantly:
- the `resultBlocks` field -which along with most other fields right now is in `AdditionalContent`- must be equipped with all the desired fields of the block (as opposed to just attaching a reference link) which is certainly not ideal. 
So I think the immediate next step should be a migration script to add a new field for `resultBlocks` (or a better name?)

**Also**
As of now, this is not being rendered at all from anywhere in the app. We can technically flip the switch and start rendering this in favor of the `legacyQuiz` whenever we so desire. (We'd just need to reference the new `Quiz` entry we have which is a copy/based on the old `bone-marrow` quiz in `give-spit...`).

#780 

### Checklist
- [somewhat] Added appropriate feature/unit tests.

(Visually it's the same as it was in #773)

# Important Announcement
We ended up changing the format quite a bit in #789 working off of these changes, so closing this one.


weerd:
> So I think the immediate next step should be a migration script to add a new field for resultBlocks (or a better name?) 
👍

resources/assets/components/Quiz/Quiz.js
```diff
@@ -1,61 +1,15 @@
...
 import { Flex, FlexCell } from '../Flex';
 import QuizQuestion from './QuizQuestion';
 import QuizConclusion from './QuizConclusion';
-import DashboardContainer from '../Dashboard/DashboardContainer';
-import LedeBannerContainer from '../LedeBanner/LedeBannerContainer';
-import TabbedNavigationContainer from '../Navigation/TabbedNavigationContainer';
+import ContentfulEntry from '../ContentfulEntry';
 
-import './quiz.scss';
+import calculateResult from './quiz-helpers';
```
weerd: Thoughts on just naming this `helpers.js` since it's already namespaced within the `Quiz` component?

resources/assets/components/Quiz/Quiz.js
```diff
@@ -65,15 +19,26 @@ class Quiz extends React.Component {
...
     this.selectChoice = this.selectChoice.bind(this);
     this.completedQuiz = this.completedQuiz.bind(this);
-    this.completeQuiz = this.completeQuiz.bind(this);
   }
 
-  selectChoice(questionId, choiceId) {
-    const choices = Object.assign({}, this.state.choices);
-    choices[questionId] = choiceId;
-    this.setState({ choices });
+  getResult() {
```
weerd: You could switch these methods to use class properties and thus eliminate all the `this.getResult = this.getResult.bind(this);` type statements. Not necessary for merge and maybe happens in a separate PR as an update.

resources/assets/components/Quiz/Quiz.js
mendelB: Yeah I think that's a better idea

resources/assets/components/Quiz/Quiz.js
mendelB: Definitely would love to do that in a follow up PR!

resources/assets/components/Quiz/QuizContainer.js
```diff
@@ -1,14 +1,89 @@
...
+            <NotFound />
+          ) : (
+            <Quiz {...props} />
+          )
+        }
+      </Enclosure>
+
+      { props.showLedeBanner ? <CallToActionContainer sticky hideIfSignedUp /> : null }
+    </div>
+  </div>
```
weerd: Interesting. I feel this is a little odd, but kinda get it. It feels like it's fuzzying up the lines of what the Quiz is. The pattern in our app has been that when we call the container for a component, it's pretty much calling the main component, but the container helps decorate it w/ additional props and actions. This feels like it's doing much more of this.

May we can chat this through a bit... 🤔

mendelB: **Postscript**: A quick chat through clarified the need to be rid of the enclosing `LedeBanner` et al logic, and keeping the container clean and minimal (just passing through the props etc.)

mendelB:
UPDATE: Removed all the `LedeBanner` misc, so we just have the plain Quiz content which shall fit the theme of abstracting all block content into pages etc.

Added a `QuizWrapper` to deal with handling the 'not found' scenario. 
Exporting both `QuizWrapper` and `Quiz` so the test file can use the plain `Quiz`.

resources/assets/components/Quiz/Quiz.js
DFurnes: FWIW, you should also only need to do the `.bind(this)` trick in the constructor when you're calling a method as an event handler. Since this is just a normal method call from the `render` method, it'll work either way.

resources/assets/components/Quiz/Quiz.js
```diff
@@ -65,15 +37,26 @@ class Quiz extends React.Component {
...
+    const result = find(this.props.results, { id: resultId });
+
+    const resultBlockId = get(this.state, 'results.resultBlockId');
+    const resultBlock = find(this.props.resultBlocks, { id: resultBlockId });
+
+    if (! (resultBlock && resultBlock.fields)) {
+      return null;
+    }
+
+    resultBlock.fields.content = `${result.content}\n\n${resultBlock.fields.content}`;
```
DFurnes: It might be nice to pull the `resultId` and `result` variables down closer to this line, and add a comment explaining what's going on. At first glance, I wasn't sure why we were modifying the block here.

It also feels like a potential issue that we have to use `get` when reading from the application's own state. Can we instead set an initial state value of `null` for this property when we set `this.state` in the constructor?

```js 
// Prepend the "quiz result" text to the specified block.
const resultId = this.state.results.resultId;
const result = find(this.props.results, { id: resultId });
resultBlock.fields.content = `${result.content}\n\n${resultBlock.fields.content}`;
```

resources/assets/components/Quiz/Quiz.js
mendelB: Mmm true! So I guess the only one which we'd need to `bind` would be `this.selectChoice`.
(`this.completeQuiz` is also a handler but we're using the `() =>` there which keeps us in scope (and seems to be a fine practice unless you're passing the handler to a lower component which is the case with `selectChoice`))

resources/assets/components/Quiz/Quiz.js
```diff
@@ -87,13 +70,21 @@ class Quiz extends React.Component {
...
       });
-      this.setState({ showResults: true });
+
+      const results = calculateResult(this.state.choices, this.props.questions);
+      this.setState({ showResults: true, results });
     }
   }
 
+  selectChoice(questionId, choiceId) {
+    const choices = Object.assign({}, this.state.choices);
```
DFurnes: This might be a nice place to use [object spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax):

```js
this.setState({
    choices: {
        ...this.state.choices,
        [questionId]: choiceId,
    }
});
```

resources/assets/components/Quiz/Quiz.js
```diff
@@ -108,7 +99,7 @@ class Quiz extends React.Component {
       />
     ));
 
-    const quizConclusion = showResults ? <h1>{conclusionText}</h1> : (
+    const quizConclusion = showResults ? this.getResult() : (
```
DFurnes: Would it make sense to rename this method to `renderResult`? Or `renderConclusion`?

resources/assets/components/Quiz/Quiz.js
mendelB: It's also occurring to me now that I didn't think through the safety logic well enough.
```
if (! (resultBlock && resultBlock.fields)) {
  return null;
}
```
Should we even sensibly be making this check here, or relying on the fact that a proper resultBlock has been assigned? And if we should, should the `result.content` be returned instead of `null`? And should we also be ensuring that `result` and `result.content` are existent?

(e.g.)
```
if (! resultBlock) {
  return result && result.content ? result.content : null;
}
```

I guess I'm just not clear on the lengths we have to take here while remaining somewhat sensible.

Also if we do have the above safety net, is it ok if that goes between 
```
const resultId = this.state.results.resultId;
const result = find(this.props.results, { id: resultId });
```
 and
`resultBlock.fields.content = `${result.content}\n\n${resultBlock.fields.content}`?

DFurnes: I'm always a bit concerned when we have multiple return types from a function, for example a React element or string here. Should we be wrapping this text in a "default" containing element when there isn't a result block?

resources/assets/components/Quiz/Quiz.js
```diff
@@ -1,79 +1,48 @@
...
-];
-
-const defaultSubmitButtonText = 'Get Results';
-
-const defaultCallToAction = 'Click **"Get Results"** to find out your likelihood for a match';
-
-const defaultConclusionText = 'Genetic compatibility is key for finding a match, which means the more diverse the registry is, the more lives we’ll save.';
-
-const defaultShowLedeBanner = false;
+const QuizWrapper = props => (
```
weerd: @mendelB mind adding a `@TODO` here regarding the `QuizWrapper` and that ideally it's just a temp thing until we figure out how to better use `Pages` to nestle a quiz in it?

mendelB:
LOL the amount of times `@todo` must have been tagged across Github 😬

mendelB:
We're going to end up changing the format quite a bit in #789 working off of these changes, so closing this one.

PR #785, opened 2018-03-26
Contentful Migrations update

mendelB:
### What does this PR do?
This PR updates two of our Contentful Migration scripts to work with the `validations` field quirk.

[Slack convo](https://dosomething.slack.com/archives/C3ASB4204/p1521821160000253)

Evidentally, we cannot set the `validations` field directly on a field that is an array, instead, one must use the `items` fields, and reset all the fields along with `validations`
https://www.contentfulcommunity.com/t/confusing-validations-error-from-the-migrations-cli/776/4

(The changes from these migrations have already been set manually on Contentful. This is just so we have a working backfill of these migrations).

### DIFF
--- a/contentful/migrations/2018_03_02_001_allow_blocks_in_pages.js
+++ b/contentful/migrations/2018_03_02_001_allow_blocks_in_pages.js
@@ -1,11 +1,16 @@
 module.exports = function (migration) {
   const block = migration.editContentType('campaign');
 
+  // https://www.contentfulcommunity.com/t/confusing-validations-error-from-the-migrations-cli/776/4
   block.editField('pages')
-    .validations([{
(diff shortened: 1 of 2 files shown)


contentful/migrations/2018_03_02_001_allow_blocks_in_pages.js
```diff
@@ -1,11 +1,16 @@
 module.exports = function (migration) {
   const block = migration.editContentType('campaign');
 
+  // https://www.contentfulcommunity.com/t/confusing-validations-error-from-the-migrations-cli/776/4
```
weerd: use the `@see https://....` Just helps define a reference in a comment :)

PR #786, opened 2018-03-26
LinkAction `link` field migration

mendelB:
### What does this PR do?
This PR adds a Contentful Migration script to update the `LinkAction` content type's `link` field validation to allow for either a valid URL, or a `tel:` link to a valid US telephone number.

I copied over the in-house contentful regex for this, only adding the bounds to allow for either the URL *or* the telephone link, as well as requiring the `tel:` at the beginning if it is indeed a telephone link.
Monkeying around with the Regex seems to show this will work splendidly.
https://regex101.com/r/8isj1Y/1

[Pivotal #156276783](https://www.pivotaltracker.com/story/show/156276783)

### Checklist

image


weerd:
@lkpttn I wonder if we could repurpose your visual design for a link (with the blue segment on the side and anchor link icon) for this, and make it have a blue segment on the side with a telephone icon?

Just a thought and not necessary for merge!

mendelB:
Yeah that would be dope! And to add to that - 
It might be helpful to also re-imagine the embed for how it might look without an image, title, description -- in cases where the embed fails for example and all we have is a URL. Which would work nicely for telephone numbers as well!

PR #788, opened 2018-03-27
Handling multiple action storing

weerd:
### What does this PR do?
This PR adds some updates to help with processing the text submission action when it is submitted. It gets validated by the `PostRequest` and then if all looks good continues on its way within the `store()` method on the `CampaignPostsController`, which passes it to the `storePost()` in the `PostRepository`. From here it gets prepped as a multipart form data and sent to Rogue and returns a response.

An update was made to the `Handler.php` as well, because we were only handling Laravel `ValidationException`s and not validation exceptions from Gateway (aliases as `GatewayValidationException`). So errors coming from Rogue were not being handled properly; `422` validation errors with its proper messaging was being turned into a `500` error with a generic server error message. But no longer!

Refs [Pivotal ID #155866202](https://www.pivotaltracker.com/story/show/155866202)

### Checklist

### DIFF
--- a/app/Http/Controllers/Api/CampaignPostsController.php
+++ b/app/Http/Controllers/Api/CampaignPostsController.php
@@ -58,41 +58,10 @@ class CampaignPostsController extends Controller
      */
     public function store($id, PostRequest $request)
     {
-        // @TODO: Reactivate this once in upcoming pull request:
+        $request->merge(['campaign_id' => $id]);
 
-        // $this->validate($request, [
-        //     'media' => 'required',  //@TODO: add file|image
-        //     'caption' => 'required|min:4|max:60',
-        //     'impact' => 'required|integer|min:1',
-        //     'whyParticipated' => 'required',
-        // ]);
+        $data = $this->postRepository->storePost($request->all());
 
-        // $request->merge(['campaign_id' => $id]);
-
-        // // Append the Campaign Run ID to Request if Legacy Campaign
-        // if (is_legacy_id($id)) {
-        //     $legacyCampaign = $this->campaignRepository->findByLegacyCampaignId($id);
-        //     $legacyCampaign = $legacyCampaign->jsonSerialize();
-
-        //     $request->merge(['campaign_run_id' => $legacyCampaign['legacyCampaignRunId']]);
-        // }
-
-        // // @TODO: Rename the following items on the front-end so they more easily
-        // // match with the params Rogue is expecting and thus don't need to manipulate
-        // // these params here.
-        // $request->merge(['file' => $request->file('media')]);
-        // $request->offsetUnset('media');
-
-        // $request->merge(['quantity' => intval($request->input('impact'))]);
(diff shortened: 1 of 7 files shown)


app/Http/Requests/PostRequest.php
```diff
@@ -54,7 +55,7 @@ public function rules()
 
             case 'text':
                 return [
-                    'text' => 'required',
+                    'text' => 'required|max:256',
```
mendelB: Not `666` in the end? 😢

resources/assets/actions/post.js
```diff
@@ -28,10 +28,11 @@ export function fetchCampaignPosts() {
 /**
  * Store posts for the specified campaign.
  *
+ * @param {String} id
  * @param  {FormData} data
  * @return {function}
  */
-export function storeCampaignPost(data) {
+export function storeCampaignPost(id, data) {
```
mendelB: Just curious about the switch from getting this `id` from the request `date` itself -- Is this more direct in terms of intention? (meaning having the id explicitly declared as a param, indicating the endpoint and campaign in a more pronounced way? cuz I notice now that we need to merge this id into the data within the controller.)

app/Http/Controllers/Api/CampaignPostsController.php
```diff
@@ -58,41 +58,10 @@ public function index($id, Request $request)
...
-        // $request->offsetUnset('impact');
-
-        // $request->merge(['why_participated' => $request->input('whyParticipated')]);
-        // $request->offsetUnset('whyParticipated');
-
-        // $data = $this->postRepository->storePost($request->all());
-
-        // return response()->json($data, 201);
-
-        return null;
```
mendelB: Were all these deleted stuff relevant to standard photo reportbacks?

app/Http/Requests/PostRequest.php
weerd: LOL 👺

resources/assets/actions/post.js
weerd: Yeah. I decided to remove it so that it doesn't have to be added via the `FormData` object when building it, especially since the variable name is camelcase and Rogue wants snake-case (we add it afterwards in the controller as the snake-cased var name). It also forces the need to pass it in and helps match the method signature of `CampaignPostsController`'s `store()` method. The "campaign" part of the method name suggests it's a specific campaign and thus explicitly passing it helps further the context. This isn't just a random post to Rogue, but a post for a specific campaign.

mendelB: 👌 Makes sense!

app/Http/Controllers/Api/CampaignPostsController.php
weerd: Yes, but only for submitting photo reportbacks via Postman, since current photo reportbacks go through the `ReportController`. So removing all this doesn't affect anything for now and will accommodate photo posts as I update that logic in the future! 🔮

PR #789, opened 2018-03-27
Alt quiz layout

mendelB:
## _PULL REQUEST OVERVIEW_

#784 
refs #780 

# What the jupiter is this?
This'd remove the `QuizWrapper` component from the `Quiz` so that the `Quiz` can handle the 'not found' scenario itself.
Also, we'll now be passing along the json fields themselves to the quiz from the container.

Would love to hear your thoughts on this! I'm quite unsure about a number of areas here tbh.

# Context
These changes are based on a discussion with @weerd regarding not breaking out of the pattern of having the container just pass through the props from state, and not having a QuizWrapper caught in the middle of things.

# Question
Would it be nicer for the container to handle passing along the props from the fields instead of having the nested JSON in the `Quiz` itself which leads to some clunkiness? I feel like it'd be more elegant for it to be that way.

The catch is that then we'd need some other way to inform Quiz in case the JSON was not found (perhaps with the old `notFound` prop?) and also, we'd need to make *all* props optional for such a case, since if no json is found we can't pass those props through.

# Awesome Hack
You can see the primary changes to #784 [here](https://github.com/DoSomething/phoenix-next/compare/major-quiz-updates...alt-quiz-layout?expand=1)

# EDIT
Based on [discussion](https://dosomething.slack.com/archives/C2BPA7M8F/p1522246692000536) and feedback, we're going to be rendering the Quiz utilizing the `ContentfulEntry` and `BlockPage` system. So for this PR, we're removing the containers work of passing over the fields to Quiz, and getting rid of the enclosing presentation markup in Quiz as well. Also no more 'not found' scenario rendering needed, since `ContentfulEntry` will niftily handle that on its own.


resources/assets/components/Quiz/Quiz.js
```diff
@@ -1,83 +1,38 @@
...
-    this.completeQuiz = this.completeQuiz.bind(this);
-  }
-
-  selectChoice(questionId, choiceId) {
-    const choices = Object.assign({}, this.state.choices);
-    choices[questionId] = choiceId;
-    this.setState({ choices });
   }
 
   completedQuiz() {
```
weerd: Small nitpick, maybe rename to `completeQuiz` since this is taking the action to complete the quiz and get the results?

mendelB: That's [this one](https://github.com/DoSomething/phoenix-next/pull/789/files#diff-07dbc25425fced330db58db5403dac95R40) no?

resources/assets/components/Quiz/QuizContainer.js
```diff
@@ -6,9 +7,16 @@ import Quiz from './Quiz';
...
-});
+const mapStateToProps = (state, ownProps) => {
+  const { slug } = ownProps.match.params;
+
+  const quiz = find(state.campaign.quizzes, { fields: { slug } });
+
+  const fields = quiz ? quiz.fields : null;
+
+  return {
+    fields,
```
weerd: maybe `return fields ? { ...fields } : null` so we spread the props into the component and then in the PropTypes we don't need to use the `shape` type and just use all the prop names?

weerd:
I do like this one a little better. Feels more self-contained and then once we have a proper approach with how we use `Page`es etc, should be a little easier to rework this into a Page :)

resources/assets/components/Quiz/QuizContainer.js
mendelB: But then we'd be fine with making all the props optional? (To account for the `notFound` scenario)

weerd: oh yeah, forgot about that 🤔

mendelB: tho tbf since most things are nested under `additionalContent` for now, we'd still be able to shape `additionalContent` stuff with an `isRequired`. Not so for `title` though.

DFurnes:
Yeah, I like the direction this is heading!

resources/assets/components/Quiz/QuizContainer.js
DFurnes: I think I feel safer with named props, since it's clearer what dependencies the component has at a glance, and it isn't tied to the format that Contentful gives us. Ideally we could handle pulling relevant elements out of the `fields` object inside `ContentfulEntry` in the future.

weerd: Wait. Ok, so what if we do `return fields ? { ...fields } : {}`...

Then when checking, use `this.props.id`, it will result in `undefined` and the `NotFound` component renders, right? Would this work ok? Then you only have to set default props for the non-required props if any (although most seem like they're required).

mendelB:
Ok! Passing through `...fields` and using `id` to determine found status. This did mean we needed to make the `additionalContent` and `title` (and eventually all extracted fields) optional props.

mendelB:
Alrighty! This should be ready to 🚢 based on the decisions we made in regards to using `ContentfulEntry` and `BlockPage`.

I'll open another PR with the required changes necessary to have that part working. For now, this Quiz as is won't see the light of day and will just chill in its directory where it belongs. (Why and how did that just turn into a spiteful rant against the Quiz?? I love the Quiz!)

resources/assets/components/Quiz/Quiz.js
```diff
@@ -1,183 +1,153 @@
...
-    this.completedQuiz = this.completedQuiz.bind(this);
-    this.completeQuiz = this.completeQuiz.bind(this);
   }
 
-  selectChoice(questionId, choiceId) {
-    const choices = Object.assign({}, this.state.choices);
-    choices[questionId] = choiceId;
-    this.setState({ choices });
-  }
+  evaluateQuiz() {
```
weerd: 👌

resources/assets/components/Quiz/Quiz.js
```diff
@@ -1,183 +1,153 @@
...
-  completedQuiz() {
-    return every(this.props.questions, question => (
+    return every(questions, question => (
       !! this.state.choices[question.id]
     ));
   }
 
   completeQuiz() {
-    if (this.completedQuiz()) {
+    if (this.evaluateQuiz()) {
```
weerd: Nice, this reads a lot better with the new method verbiage.

resources/assets/components/Quiz/QuizContainer.js
```diff
@@ -3,12 +3,4 @@ import { PuckConnector } from '@dosomething/puck-client';
...
-/**
- * Provide state from the Redux store as props for this component.
- */
-const mapStateToProps = state => ({
-  dashboard: state.campaign.dashboard,
-});
-
-// Export the container component.
-export default connect(mapStateToProps)(PuckConnector(Quiz));
+export default connect()(PuckConnector(Quiz));
```
weerd: haha! the `QuizContainer` totally shrunk!

PR #793, opened 2018-03-28
Quiz rendering

mendelB:
### What does this PR do?
This PR adds rendering support for the brand new Quiz ⚡️

- updating the `FindContentfulEntry` helper method to work with a general `identifier` param, and find either by slug or id
- updating the `BlockPageContainer` to pass both `id` and `slug` from `match` params
- updating the `/quiz` route to render a `BlockPageContainer`
- adding cases to the `ContentfulEntry` to handle rendering a beta/legacy or new/regular Quiz
- updating LegacyQuiz to work with this new format

## Edit:
~Whoops, there are some issues with the `LegacyQuiz` straight up rendering as a `ContentfulEntry` in a BlockPage. Changed things back so that we're rendering the LegacyQuiz from the `/quiz` route.~
Nope. Made some changes to the LegacyQuiz so that it can indeed render nicely through `BlockPagesContainer`, and `ContentfulEntry`. So now we've got backward compatibility and can use either quiz on Contentful.

Refs #780 

### DIFF
--- a/resources/assets/components/Page/BlockPage/BlockPageContainer.js
+++ b/resources/assets/components/Page/BlockPage/BlockPageContainer.js
@@ -6,9 +6,9 @@ import { findContentfulEntry } from '../../../helpers';
  * Provide state from the Redux store as props for this component.
  */
 const mapStateToProps = (state, ownProps) => {
-  const { id } = ownProps.match.params;
+  const { id, slug } = ownProps.match.params;
 
-  const json = findContentfulEntry(state, id);
+  const json = findContentfulEntry(state, (id || slug));
 
   return { json };
 };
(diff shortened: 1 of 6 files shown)


weerd:
> ~Whoops, there are some issues with the LegacyQuiz straight up rendering as a ContentfulEntry in a BlockPage. Changed things back so that we're rendering the LegacyQuiz from the /quiz route. ~
Nope. Made some changes to the LegacyQuiz so that it can indeed render nicely through BlockPagesContainer, and ContentfulEntry. So now we've got backward compatibility and can use either quiz on Contentful.

enjoyed the stream of consciousness description 😉

resources/assets/components/Page/BlockPage/BlockPageContainer.js
```diff
@@ -6,9 +6,9 @@ import { findContentfulEntry } from '../../../helpers';
  * Provide state from the Redux store as props for this component.
  */
 const mapStateToProps = (state, ownProps) => {
-  const { id } = ownProps.match.params;
+  const { id, slug } = ownProps.match.params;
 
-  const json = findContentfulEntry(state, id);
+  const json = findContentfulEntry(state, (id || slug));
```
weerd: Should we switch around the order to prefer the `slug` over the `id`?

mendelB: I mean optimally those two would never collide, right? Or are you saying we want to start prioritizing over ID's slugs in general?

weerd: Oh true. Just was my initial gut reaction b/c with an `||` statement, if the first check passes it ignores the second, but in this scenario we'd only be passing a single item, so I think you're right!

PR #794, opened 2018-03-29
Migration script for resultBlocks field

mendelB:
### What does this PR do?
This PR adds a Contentful Migration script to create a new `resultBlocks` field for the `Quiz` Content Type.

This is probably the most crucial field that we need to formalize since as things are now, one would need to manually paste in all the necessary fields for the result blocks into Additional Content 😨 

Refs #780 

image


mendelB:
One thing I wasn't entirely sure about with this one was should the field be required?

We do have a [fallback in place](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/Quiz/Quiz.js#L70) for if no `resultBlock` is found for the Quiz result, but do we want to encourage actual Blocks as quiz results, or either way is OK?

weerd:
I don't think it should be required, but maybe it's a question for @ashleybaldwin?

contentful/migrations/2018_03_29_001_add_result_blocks_field_to_quiz.js
```diff
@@ -0,0 +1,14 @@
+module.exports = function (migration) {
+  const quiz = migration.editContentType('quiz');
+
+  quiz.createField('resultBlocks')
+    .name('Result Blocks')
+    .type('Link')
+    .linkType('Entry')
```
weerd: Is this only a single reference and not a many reference? The old quiz beta results field was a many reference, so you could include both alternatives depending on outcome of quiz... either a link to external site if over 18 or a share action if under 18. Does this work differently now if it's a single reference?

mendelB: omg no this should be multi reference 🤦‍♂️ 
I made this last night after too much champaign it seems. I'll update!

mendelB: OK, updated! (Yeesh, good catch!)

mendelB:
Confirmed with @ashleybaldwin, optional field it is!

PR #798, opened 2018-03-30
Updating the SignupButton to allow disabling it via JSON.

weerd:
### What does this PR do?
This PR provides a quick hack to disable the signup button on a campaign if a `disableSignup` property is added to the JSON `additionalContent` field.

This was required to help disable signups on the Rinse, Recycle Repeat College Competition campaign to limit the pool, while not closing the campaign and allowing Reportback submissions.

### DIFF
--- a/resources/assets/components/SignupButton/SignupButton.js
+++ b/resources/assets/components/SignupButton/SignupButton.js
@@ -31,17 +31,17 @@ const SignupButton = (props) => {
     });
   };
 
+  // Have signups been disabled for this campaign?
+  if (disableSignup) {
+    return null;
+  }
+
   const baseCopy = text || campaignActionText;
 
   // If no source-specific override, don't opt in to the A/B test.
   const sourceOverride = get(sourceActionText, trafficSource);
   if (! sourceOverride) {
-    return (
-      disableSignup ?
-        null
-        :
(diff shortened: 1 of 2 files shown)


resources/assets/components/SignupButton/SignupButton.js
```diff
@@ -35,54 +36,64 @@ const SignupButton = (props) => {
   // If no source-specific override, don't opt in to the A/B test.
   const sourceOverride = get(sourceActionText, trafficSource);
   if (! sourceOverride) {
-    return <Button className={className} onClick={() => onSignup(baseCopy)} text={baseCopy} />;
```
DFurnes: Could we instead short-circuit and return `null` above these other two cases?

```js
// Have signups been disabled for this campaign?
if (disableSignup) {
    return null;
}

// If no source-specific override, don't opt in to the A/B test.
const sourceOverride = get(sourceActionText, trafficSource);
if (! sourceOverride) {
    return <Button className={className} onClick={() => onSignup(baseCopy)} text={baseCopy} />;
}

// ...
```

weerd: oh ya. that feels cleaner. haven't had my ☕️yet... 😪

weerd:
Much cleaner @DFurnes 🛁

PR #802, opened 2018-04-02
Add loading indicator for GraphQL gallery pagination.

DFurnes:
### What does this PR do?
This pull request refactors the GraphQL gallery to use the `Query` component (rather than the `graphql` HOC), which now seems to be the suggested way of doing things in the [Apollo documentation](https://www.apollographql.com/docs/react/features/pagination.html). I think it looks a lil' more readable so seems like a win!

I've also added support for parsing different types of loads from the `networkState` prop - so we show a spinner on initial load, but then just disable & style the "Load More" button when fetching subsequent pages (while the rest of the results stay intact). ⏳ 

It may be helpful to compare [before](https://github.com/DoSomething/phoenix-next/blob/0c57c975d84d022825ad4538a8d5cd6ca5533275/resources/assets/components/CampaignGalleryBlock/CampaignGalleryBlockContainer.js) and [after](https://github.com/DoSomething/phoenix-next/blob/5802c427a8fafb94c758b5b700cdd6bd9d52a330/resources/assets/components/CampaignGalleryBlock/CampaignGalleryBlockContainer.js) when reviewing this!

[#155944555](https://www.pivotaltracker.com/story/show/155944555)

### Checklist

### DIFF
--- a/resources/assets/components/CampaignGalleryBlock/CampaignGalleryBlockContainer.js
+++ b/resources/assets/components/CampaignGalleryBlock/CampaignGalleryBlockContainer.js
@@ -6,6 +6,7 @@ import gql from 'graphql-tag';
 
 import CampaignGalleryBlock from './CampaignGalleryBlock';
 import ErrorBlock from '../ErrorBlock/ErrorBlock';
+import { NetworkStatus } from '../../constants';
 
 /**
  * Provide state from the Redux store as props for this component. (In
@@ -47,7 +48,7 @@ const CampaignGalleryQuery = ({ campaignId }) => (
   >
     {({ data, error, networkStatus, variables, fetchMore }) => {
       // On initial load, just display a loading spinner.
-      if (networkStatus === 1) {
+      if (networkStatus === NetworkStatus.LOADING) {
         return <div className="spinner -centered" />;
       }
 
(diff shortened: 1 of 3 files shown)


DFurnes:
> @weerd [had spoken](https://github.com/DoSomething/phoenix-next/pull/784#discussion_r177189457) about the pattern of how we use Containers (merely passing state props, and rendering the main component), and I'm wondering how this might apply here, what with this being a new pattern we're introducing with GraphQL? 
> 
> I guess as we implement this more across the app we'll be doing more work to formalize how we run the query, and HOC, or would this just be an OK exception to the rule in general?

Yeah. I think as we continue to invest in GraphQL it'll make sense to eventually extract out our own `BaseQuery` class that's responsible for the "broad strokes" of how to fetch and paginate results (and say, replacing a component with a loading spinner on that initial load).

I deliberated over splitting this out into `CampaignGalleryBlockContainer` (Redux), `CampaignGalleryBlockQuery` (GraphQL), and `CampaignGalleryBlock` components as well, but wasn't sure if that was worth the overhead of swapping between three different places? Might try that in a follow-up PR though and see how it feels...

mendelB:
Ah nice, `BaseQuery` idea sounds interesting! In general, I'm like never completely sure of the correct approach between separating code across files or keeping it in one place for easier reference... As with everything I'm sure it depends...

Anyways, thanks for clarifying!

resources/assets/components/CampaignGalleryBlock/CampaignGalleryBlockContainer.js
```diff
@@ -34,45 +38,58 @@ const query = gql`
...
  */
-const queryOptions = props => ({
-  variables: {
-    campaignId: props.campaignId,
-    count: 9,
-    page: 1,
-  },
-});
+const CampaignGalleryQuery = ({ campaignId }) => (
+  <Query
```
weerd: Cooooool!

weerd:
@mendelB thanks for asking about the container pattern stuff :)

post-merge 👍! Looking awesome @DFurnes

PR #803, opened 2018-04-02
TsText Submission Action reacting to responses

weerd:
### What does this PR do?
This PR enables the `TextSubmissionAction` component to response appropriately to responses received in the redux store regarding the results of a form submission. If the response has errors, it shows the error message list and highlights the labels/inputs of items with errors, but if the response is successful, it shows a success message.

text submission action validation

One thing I wasn't 💯% on, was how I clear the form after successful submission. It works great, just not sure if ideal and if using the `componentDidUpate()` lifecycle method is cool with that. I initial tried to do a simple `this.form.reset()` type thing, but this can't be done if using React and taking the ["controlled components"](https://reactjs.org/docs/forms.html#controlled-components) approach (which is the recommended approach with form fields). So instead you have to reset the state. I check agains the previous state and current state to see if the response status was successful. We only really want to reset the form each time a successful submission is made, otherwise keep the state and this approach seems to accomplish this! 🤷‍♂️ 

Additionally, I'm contemplating a future refactor where instead of having a `helpers/forms.js` with a few helpful functions, instead we formalize a `FormResponse` class and add those as methods to the class. It would allow for much more versatility and usefulness, ability to decorate the response further and also test against receiving an instance of said class... something to contemplate 🤔 

Next steps include removing the Laravel `ValidationException` message and overriding it with something with language that is a little nicer. Also need to setup an `Affirmation` to launch if the submission was successful instead of just the validation message in the form.

Refs [Pivotal ID #155866202](https://www.pivotaltracker.com/story/show/155866202)

### Checklist

### DIFF
--- a/resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
+++ b/resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
@@ -1,11 +1,13 @@
 /* global FormData */
 
 import React from 'react';
+import { has } from 'lodash';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
 import Card from '../../Card';
 import FormValidation from '../../utilities/Form/FormValidation';
+import { getFieldErrors } from '../../../helpers/forms';
 
 import './text-submission-action.scss';
 
@@ -13,6 +15,15 @@ class TextSubmissionAction extends React.Component {
   state = {
     textValue: '',
   };
+
+  componentDidUpdate = (prevProps) => {
+    const prevResponse = prevProps.submissions.items[this.props.id] || null;
+    const response = this.props.submissions.items[this.props.id] || null;
+
+    if (! has(prevResponse, 'status.success') && has(response, 'status.success')) {
+      this.resetForm();
+    }
+  }
 
   handleSubmit = (event) => {
     event.preventDefault();
@@ -32,7 +43,6 @@ class TextSubmissionAction extends React.Component {
       }));
     }
 
-    // Send this off to the backend API to validate and send off to Rogue.
     this.props.storeCampaignPost(this.props.campaignId, formData);
   }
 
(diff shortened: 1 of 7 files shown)


resources/assets/helpers/forms.js
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param  {Object} response
+ * @return {Object}
+ */
+export function getFieldErrors(response) {
+  if (! has(response, 'status.error.fields')) {
+    return null;
+  }
+
+  return response.status.error.fields;
```
mendelB: Might this be slightly more concise with a lodash `get`?
(`get(response, 'status.error.fields')`)

resources/assets/components/utilities/Form/FormValidation.js
```diff
@@ -0,0 +1,42 @@
...
+
+FormValidation.propTypes = {
+  response: PropTypes.shape({
+    data: PropTypes.object,
+    status: PropTypes.object.isRequired,
+    type: PropTypes.string.isRequired,
+  }).isRequired,
+};
+
+export default FormValidation;
```
mendelB: This is neat! 👌

resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
```diff
@@ -13,6 +16,15 @@ class TextSubmissionAction extends React.Component {
     textValue: '',
   };
 
+  componentDidUpdate = (prevProps) => {
+    const prevResponse = prevProps.submissions.items[this.props.id] || null;
+    const response = this.props.submissions.items[this.props.id] || null;
+
+    if (! has(prevResponse, 'status.success') && has(response, 'status.success')) {
```
mendelB: Sorry, my brain is fighting me here. Why are we checking that the `prevResponse` does not have a successful status?

resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
```diff
@@ -13,6 +16,15 @@ class TextSubmissionAction extends React.Component {
     textValue: '',
   };
 
+  componentDidUpdate = (prevProps) => {
```
mendelB: As far as I understand it, I think using a lifecycle method for the form reset approach makes perfect sense here! However, should we be using the [`getDerivedStateFromProps`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) instead?
https://reactjs.org/docs/react-component.html#unsafe_componentwillupdate
```
Note that you cannot call this.setState() here; nor should you do anything else (e.g. dispatch a Redux action) that would trigger an update to a React component before UNSAFE_componentWillUpdate() returns.
```

resources/assets/helpers/forms.js
weerd: oh, true!

resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
weerd: Yeah, this was a little funky. Basically, prior to submitting the form successfully, both should be returning `false`. Then when the forms submits successfully, it returns `false` & `true` which is the condition I specifically want. If the user continues to submit, I don't want it to keep resetting the form when they go to type another message AND the redux store hasn't changed so `success` from prior submission is still indicated. This accounts for the "correct" scenario and allows multiple submissions :)

resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
mendelB: 🤦‍♂️ 
yeesh. never mind me then.

resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
mendelB: Ok, I *think* I'm getting you here. And lastly, where's the `this.props.id` being set? I'm not seeing that in the container.

mendelB: Ach but of course, in `/renderers`!

weerd: Not in the container. It's the actual ContentfulID of the component set in the renderer when using the `TextSubmissionAction` component: https://github.com/DoSomething/phoenix-next/blob/c4fe14eb6cfcbc741c12e7473cfdda9bea2e8fe2/resources/assets/components/ContentfulEntry/renderers.js#L194-L204

weerd: woah! GitHub now adds a mini code file embed?!?!

mendelB: Dope!!!

PR #804, opened 2018-04-02
Display reactions in GraphQL gallery.

DFurnes:
### What does this PR do?
This PR makes a few additions to the GraphQL-powered gallery:

🥙 I flattened the `url` and `text` properties into the post [in the GraphQL gateway](https://github.com/DoSomething/graphql/commit/65f708fbe0e3bbf4fc6ea227504a3f00f708b08c). This makes more sense to me when querying (since for us `media` isn't a real "thing" with a unique ID, and for a text-only post, the `text` isn't really even media). ddf3256

❤️ Reactions are now displayed on the gallery! 81e8529

🕵️ I split out the container component into `CampaignGalleryBlockContainer` (Redux) and `CampaignGalleryBlockQuery` (GraphQL), and extracted more generalized "paginated query" logic into a `PaginatedQuery` class that gets imported there (based on @mendelB's [feedback](https://github.com/DoSomething/phoenix-next/pull/802#pullrequestreview-108691424)!). I think this feels a little cleaner and nicer to work with, time will tell! 3e7fe6e

🔗 Fixes an issue where the `Authorization` header was not being properly set. 948def39

screen shot 2018-04-02 at 4 34 45 pm

I'd recommend reviewing things commit-by-commit!

[#155944555](https://www.pivotaltracker.com/story/show/155944555)

### Checklist

### DIFF
--- a/resources/assets/components/CampaignGalleryBlock/CampaignGalleryBlock.js
+++ b/resources/assets/components/CampaignGalleryBlock/CampaignGalleryBlock.js
@@ -18,8 +18,8 @@ const CampaignGalleryBlock = (props) => {
             <ReportbackItem
               id={String(post.id)}
               type={post.type}
-              caption={post.media.text}
-              url={post.media.url}
+              caption={post.text}
+              url={post.url}
               firstName={get(post, 'user.firstName') || 'A Doer'}
             />
           </Card>
--- a/resources/assets/components/CampaignGalleryBlock/CampaignGalleryBlock.js
+++ b/resources/assets/components/CampaignGalleryBlock/CampaignGalleryBlock.js
@@ -18,6 +18,9 @@ const CampaignGalleryBlock = (props) => {
             <ReportbackItem
               id={String(post.id)}
               type={post.type}
+              reaction={{ total: post.reactions, reacted: post.reacted }}
+              toggleReactionOn={() => console.log('Coming soon!')}
+              toggleReactionOff={() => console.log('Coming soon!')}
               caption={post.text}
               url={post.url}
               firstName={get(post, 'user.firstName') || 'A Doer'}
--- a/resources/assets/graphql.js
+++ b/resources/assets/graphql.js
@@ -10,13 +10,17 @@ import { onError } from 'apollo-link-error';
 import { env } from './helpers';
 
 // Create an authentication link with the user's access token.
-const accessToken = window.AUTH.jwt;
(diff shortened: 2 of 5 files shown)


resources/assets/components/CampaignGalleryBlock/CampaignGalleryBlock.js
```diff
@@ -18,8 +18,11 @@ const CampaignGalleryBlock = (props) => {
             <ReportbackItem
               id={String(post.id)}
               type={post.type}
-              caption={post.media.text}
-              url={post.media.url}
+              reaction={{ total: post.reactions, reacted: post.reacted }}
+              toggleReactionOn={() => console.log('Coming soon!')}
+              toggleReactionOff={() => console.log('Coming soon!')}
```
weerd: 🔜

resources/assets/graphql.js
```diff
@@ -10,13 +10,17 @@ import { onError } from 'apollo-link-error';
 import { env } from './helpers';
 
 // Create an authentication link with the user's access token.
-const accessToken = window.AUTH.jwt;
-const authenticationLink = setContext((request, { headers }) => {
+const authenticationLink = setContext((request, context) => {
+  const accessToken = window.AUTH.jwt;
```
weerd: Just FYI on campaign pages I expose this as: https://github.com/DoSomething/phoenix-next/blob/d019f0e0d8cc42ac792fd1ba36c78b2d4da682e2/app/Http/Controllers/CampaignController.php#L64

and then retrieve it in as part of the redux store when sending API requests as `getState().user.token;`. Would it make more sense to unify these and keep as window prop like you have here so it's the same approach?

DFurnes: Oh, hmm! I do kinda like `window.AUTH` here since it's not Redux-specific, but agreed it's weird to have it stored in multiple places.

weerd: Yeah, that seems fine with me. I can update on my end. Where do you set it? Is it set on every "page", so if I try nabbing it from the `window` it'll be fine while on campaign page?

DFurnes: Yes! It's set on every view with a [wildcard view composer](https://github.com/DoSomething/phoenix-next/blob/d019f0e0d8cc42ac792fd1ba36c78b2d4da682e2/app/Providers/AppServiceProvider.php#L54-L59).

PR #806, opened 2018-04-02
Fix for issues with older Safari support of FormData methods.

weerd:
### What does this PR do?
This PR provides a fix for the lack of support in some browsers (*cough* IE, *cough* Safari) for the methods on the `FormData` class. Instead of passing the `FormData` as the data directly to the `storeCampaignPost()` method and trying to use the unsupported `get()` method to grab some of the data to provide better context to the backend, we instead will now pass an object with the data as properties of the object. Makes things a lot easier.

Blessing in disguise? 🤷‍♂️ 

Fixes #

### DIFF
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -33,30 +33,32 @@ export function fetchCampaignPosts() {
 /**
  * Store posts for the specified campaign.
  *
- * @param {String} id
- * @param  {FormData} data
+ * @param {String} campaignId
+ * @param  {Object} data
  * @return {function}
  */
-export function storeCampaignPost(id, data) {
-  if (! (data instanceof FormData)) {
-    throw Error(`The supplied data must be an instance of FormData, instead it is an instance of ${data.constructor.name}.`);
+export function storeCampaignPost(campaignId, data) {
+  if (! (data.body instanceof FormData)) {
+    throw Error(`The supplied data.body must be an instance of FormData, instead it is an instance of ${data.body.constructor.name}.`);
   }
+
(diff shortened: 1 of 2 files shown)


resources/assets/actions/post.js
```diff
@@ -33,30 +33,32 @@ export function fetchCampaignPosts() {
...
- * @param {String} id
- * @param  {FormData} data
+ * @param {String} campaignId
+ * @param  {Object} data
  * @return {function}
  */
-export function storeCampaignPost(id, data) {
-  if (! (data instanceof FormData)) {
-    throw Error(`The supplied data must be an instance of FormData, instead it is an instance of ${data.constructor.name}.`);
+export function storeCampaignPost(campaignId, data) {
```
mendelB: Would it at all be a good idea to use flow here to enforce the property types of `data`?
I just remembered seeing this here:
https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/helpers/index.js#L263-L283

weerd: Hmmm, yeah, I think the consensus has been that Flow has been a pain in the 🍑; spent more time struggling with the Flow syntax and getting it to work than just building a feature. We don't really use Flow much either, and might look to alternatives. We've mostly relied on `PropTypes` in components and eschewed using Flow. I think for now the error throwing is ok. Dave's been wanting to look at TypeScript to see if that works a bit better. I'd prefer a just JS (not superset of JS) solution, but not sure what else is out there that is better.

DFurnes: Man, I do really want some static type checking though!

PR #813, opened 2018-04-06
Basic modals refactor

weerd:
### What does this PR do?
This PR refactors how we execute showing a modal after an event (like an action submission). This uses [React Portals](https://reactjs.org/docs/portals.html) to simplify how to inject components into a container that exists outside the main app container. The aim was to help simplify how we do modals and make it easier to just pop any component inside a modal without needing to add a secondary custom variation on a component that is specific for modals (which is the current process).

It also updates/improves on the styling for the modal so it can be more receptive to whatever components we throw inside of the modal.

This currently only supports injecting _one_ component within a modal.

new modal experience

Prior to this update, the TextSubmissionAction component was forced to use the affirmation content message from the PhotoUploaderAction. However, this update will allow for using a custom affirmation content message specifically for the TSA (once the affirmationContent` field is added to the TSA content-type on Contentful; upcoming PR).

### Checklist

### DIFF
--- a/app/Entities/TextSubmissionAction.php
+++ b/app/Entities/TextSubmissionAction.php
@@ -21,6 +21,7 @@ class TextSubmissionAction extends Entity implements JsonSerializable
                 'textFieldLabel' => $this->textFieldLabel,
                 'textFieldPlaceholder' => $this->textFieldPlaceholder,
                 'buttonText' => $this->buttonText,
+                'affirmationContent' => $this->affirmationContent ?: null,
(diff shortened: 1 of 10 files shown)


weerd:
image

When there's lots of text, the modal container scrolls to allow seeing the full extent of the content.

weerd:
Also works well on mobile:
image

Even if the modal has a ton of content, it scrolls nicely:
image

mendelB:
This is so awesome BTW!

PR #817, opened 2018-04-09
Adding Quiz Fields Contentful migration

mendelB:
### What does this PR do?
This PR adds a Contentful CLI Migration Script to add a `results` and `questions` field to the `Quiz` content type, in anticipation of us utilizing the JSON FOrm Editor extension to allow for these complex JSON fields in Contentful! 🎈 

#780 

image

### DIFF
--- a/contentful/migrations/2018_04_09_001_add_fields_to_quiz.js
+++ b/contentful/migrations/2018_04_09_001_add_fields_to_quiz.js
@@ -1,0 +1,16 @@
+module.exports = function (migration) {
+  const quiz = migration.editContentType('quiz');
+
+  quiz.createField('results')
+    .name('Results')
+    .type('Object')
+    .required(true);
+
+  quiz.createField('questions')
+    .name('Questions')
+    .type('Object')
+    .required(true);
+
+  quiz.moveField('results').beforeField('resultBlocks');
+  quiz.moveField('questions').beforeField('resultBlocks');
+}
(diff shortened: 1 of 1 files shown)


contentful/migrations/2018_04_09_001_add_fields_to_quiz.js
```diff
@@ -0,0 +1,16 @@
...
+    .type('Object')
+    .required(true);
+
+  quiz.createField('questions')
+    .name('Questions')
+    .type('Object')
+    .required(true);
+
+  quiz.moveField('results').beforeField('resultBlocks');
+  quiz.moveField('questions').beforeField('resultBlocks');
```
weerd: actually... do we want the `questions` to be above the `results` which are above the `resultBlocks`?

mendelB: I did it this way for two reasons:
1. When crafting a quiz, ([according to the way BuzzFeed explains it](https://www.buzzfeed.com/catesish/how-to-make-a-buzzfeed-quiz?utm_term=.usomAPyOR#.pb5Zpxlgz)) you should have your results defined first, and then you can define the questions that lead to those results.
2. When defining choices for the questions, you can reference those already defined result IDs

Does this make sense?

mendelB: (This might be an argument for moving the `resultBlocks` field on top of `questions` as well...

weerd: Oh nice. Ok, that makes sense to me! I guess then ya, move the `questions` under `resultBlocks` :)

mendelB: Done!

PR #819, opened 2018-04-09
Analytics middleware

weerd:
### What does this PR do?
This PR  provides an alternative way to track events in Phoenix, using an Analytics Middleware. The middleware is currently configured to work with Puck, but we could extend it to work with other analytics services in the future, like Google Analytics (or just easily send certain events to one service and other events to another). Instead of requiring components to be wrapped with the `PuckConnector` higher-order component, this approach simply provides a middleware that is triggered by a simple action method. No need to wrap a container component, or even require a container component for a component that you want to track events on. Just pull in the action method, and trigger the event 👌 

There is a generic `trackAnalyticsEvent()` method, or more specific `trackPuckEvent()` and `trackGoogleAnalyticsEvent()` (GA one is not implemented, just there for example).

~~My biggest question is that I suspect it is not a great idea to instantiate a new `Puck` engine each time an event is triggered. So I'm looking into ways to approach this as a Singleton, (but that would ideally happen in the Puck Client library) or maybe stick the Puck client on the `window` for the time being and check there if there's an instance already? I dunno, but could use some advice!~~ 

Check the [update](https://github.com/DoSomething/phoenix-next/pull/819#issuecomment-380248456) below!

This otherwise works great and triggers events! Here's an example:

image

Fixes #

### Checklist


mendelB:
> My biggest question is that I suspect it is not a great idea to instantiate a new Puck engine each time an event is triggered. So I'm looking into ways to approach this as a Singleton, (but that would ideally happen in the Puck Client library) or maybe stick the Puck client on the window for the time being and check there if there's an instance already? I dunno, but could use some advice!

I think the window idea is nice. So like whether we end up using a `puck` Singelton instance or not, we could set a global `PuckClient` object in say [`init.js`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/init.js) and use that within the middleware to emit the puck event? So we wouldn't have to check for the instance but just use the globally defined PuckClient object.

mendelB:
Kinda like how as of now, the way it traditionally works is using the `PuckProvider` as a 'global' context (which carries it's own `Engine` instance), and the `Provider` would pull the `trackEvent` using the Context magic.

So we'd just be taking care of defining and using that global Engine on our own

resources/assets/middleware/analytics.js
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * @param  {String} name
+ * @param  {Object} data
+ * @param  {Object} state
+ * @return {void}
+ */
+const analyzeWithPuck = (name, data, state) => {
+  const Puck = new PuckClient({
+    source: 'phoenix-next',
+    // Not sure why Puck Client needs this as a function:
```
mendelB: Hmm could this have to do with tracking across a session? So if they're not logged in when the app is instantiated we wouldn't be [just passing in](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/App.js#L22) a static `null` id value, but rather we use this function to check for the ID each event?

resources/assets/actions/analytics.js
```diff
@@ -0,0 +1,43 @@
...
+ * Track an analytics event with a specified service.
+ *
+ * @param  {String} name
+ * @param  {Object} data
+ * @param  {String} service
+ * @return {Object}
+ */
+export function trackAnalyticsEvent(name, data, service) {
+  return {
+    type: TRACK_ANALYTICS_EVENT,
```
DFurnes: This doesn't feel like a separate `TRACK_ANALYTICS_EVENT` action since it doesn't modify the store. I'd love to look into something [like this package](https://github.com/markdalgleish/redux-analytics) (even if implemented ourselves) where we annotate the action with analytics info.

resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
```diff
@@ -37,6 +37,16 @@ class TextSubmissionAction extends React.Component {
   handleSubmit = (event) => {
     event.preventDefault();
 
+    // @TODO: clean this up and set data into variable to be passed for multiple
+    // trackPuckEvent calls.
+    this.props.trackPuckEvent('clicked-text-submission-action-button', {
```
DFurnes: Similarly, this feels like it could just be a plain ol' helper function (rather than an action we have to inject).

resources/assets/middleware/analytics.js
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Send event to analyze with Puck.
+ *
+ * @param  {String} name
+ * @param  {Object} data
+ * @param  {Object} state
+ * @return {void}
+ */
+const analyzeWithPuck = (name, data, state) => {
+  const Puck = new PuckClient({
```
DFurnes: The easiest way to make this a singleton is to pull it out of the function (so it persists in the module scope).

```js
let puckClient = null;

const analyzeWithPuck = (name, data, state) => {
  if (! puckClient) {
    puckClient = new PuckClient({
        // ...
    });

```

resources/assets/middleware/analytics.js
DFurnes: Yeah, the intent here was that we could change the value in the store and it'd continue to be reflected correctly in Puck (without a page reload). We don't do this in practice, though.

weerd: Ah, yeah, that makes sense.

resources/assets/middleware/analytics.js
weerd: Aha!

resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
weerd: Ah, yeah. That's a good point. I think I do like it better if it is a helper.

resources/assets/actions/analytics.js
weerd: That's fair. This was my initial attempt but figured it needed to be further refined. I may want to chat because I _think_ I get what you mean, but not 100% sure.

weerd:
Thanks for all the awesome input @DFurnes and @mendelB! After a further chat with @DFurnes, he provided some good insight on a variation to the approach which is a bit simpler and truly separates running Analytics independent of going through the Redux process. As [@DFurnes pointed out](https://github.com/DoSomething/phoenix-next/pull/819#discussion_r180451488), it also didn't make sense to execute an `action` when the Redux store wasn't being updated when tracking an event.

So instead, now any Redux actions that _want_ to have the action tracked, can simply add a `meta.analytics` property to their payload, and the `analyticsMiddleware` will recognize this and track the event. The `analytics` payload contains a `name`, `service`, and `payload` property to help define the analytics tracking.

However, if you want to track a event that is not an action that can be dispatched, feel free to do so with the `trackAnalyticsEvent()`, `trackGoogleAnalyticsEvent()`, or `trackGoogleAnalyticsEvent()`! Those do not involve the middleware, so no Redux-y setup is required. Just pull in the function and call it with the appropriate arguments!

resources/assets/helpers/analytics.js
```diff
@@ -0,0 +1,103 @@
...
+
+  Puck.trackEvent(name, data);
+}
+
+
+/**
+ * Watch the given parameters for changes in their state
+ * and record it to Google Analytics.
+ */
+export function googleAnalyticsTrack(history) {
```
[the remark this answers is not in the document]
weerd: Yeah, I wasn't sure. That sounds better. I can rename 👍

mendelB:
This is super exciting!

PR #820, opened 2018-04-10
Lil update for the timed out embeds

mendelB:
### What does this PR do?
This PR updates the output for a timed out Embed request from our `Embed` component so that it's at least slightly nicer.

*BEFORE*
image

*AFTER*
image

I am certainly no design wizard, so any recommendations are welcome! This is just a quick pass to at least get something legible out the door 🤒 

[Pivotal #156024745](https://www.pivotaltracker.com/story/show/156024745)


resources/assets/components/Embed/index.js
```diff
@@ -39,7 +41,14 @@ class Embed extends React.Component {
         </a>
       );
     } else if (this.state.requestFailed) {
-      embed = <a href={this.props.url}>{this.props.url}</a>;
+      embed = (
+        <a href={this.props.url}>
+          <Figure className="padded margin-bottom-none bordered rounded" image={linkIcon} alt="link icon" alignment="left-collapse" size="small">
+            <h3>{ this.props.url }</h3>
```
DFurnes: Should we truncate this if it's over N characters? Wondering how this would look if we had a longer URL.

mendelB: That is a good idea! Done!

resources/assets/components/Embed/index.js
```diff
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import classnames from 'classnames';
 import { Phoenix } from '@dosomething/gateway';
+import { truncate } from 'lodash';
```
weerd: oh nice! of course Lodash has a function for this. LOL.

PR #823, opened 2018-04-11
Contentful Migration script for ContentBlock 'imageAlignment' field

mendelB:
### What does this PR do?

This PR adds a new Contentful Migration script for the `ContentBlock` Content-Type which adds a new `imageAlignment` field

Not married to this naming! So if you have a better suggestion prey tell 😀 

I don't know of any way to mess with the Appearance settings from a migration script, but I will duly add a radio button appearance, as well as a description.

[Slack](https://dosomething.slack.com/archives/C3ASB4204/p1523392554000097)

image

### DIFF
--- a/contentful/migrations/2018_04_11_001_add_image_alignment_field_to_content_block.js
+++ b/contentful/migrations/2018_04_11_001_add_image_alignment_field_to_content_block.js
@@ -1,0 +1,12 @@
+module.exports = function(migration) {
+  const contentBlock = migration.editContentType('contentBlock');
+
+  contentBlock
+    .createField('imageAlignment')
+    .name('Image Alignment')
+    .type('Symbol')
+    .required(false)
+    .validations([{ in: ['Right', 'Left'] }]);
+
+  contentBlock.moveField('imageAlignment').beforeField('additionalContent');
+};


contentful/migrations/2018_04_11_001_add_image_alignment_field_to_content_block.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function(migration) {
+  const contentBlock = migration.editContentType('contentBlock');
+
+  contentBlock
+    .createField('imageAlignment')
+    .name('Image Alignment')
+    .type('Symbol')
```
weerd: Is it still considered a `Symbol`? Isn't that for a text field?

contentful/migrations/2018_04_11_001_add_image_alignment_field_to_content_block.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function(migration) {
+  const contentBlock = migration.editContentType('contentBlock');
+
+  contentBlock
+    .createField('imageAlignment')
```
weerd: I think I dig the name :)

contentful/migrations/2018_04_11_001_add_image_alignment_field_to_content_block.js
mendelB: I think it's still considered a text field, we're just using validation to make it appear as two radio button options. (I tested a bit manually in a sandbox)

weerd: oh, nice. ok! Just making sure 💃

PR #825, opened 2018-04-11
Append 'legacy' to all 'content block' references

mendelB:
### What does this PR do?

This PR shall append 'Legacy' to all Content Block references.

We are working on formally introducing a new `ContentBlock` component that will replace the current working version.

[Slack](https://dosomething.slack.com/archives/C3ASB4204/p1523390187000400)


resources/assets/components/Block/LegacyContentBlock.js
```diff
@@ -55,12 +57,15 @@ const ContentBlock = props => {
         ) : null}
         {photoComponent}
       </Flex>
-      <PuckWaypoint name="content-block__bottom" waypointData={{ title }} />
+      <PuckWaypoint
+        name="legacy-content-block__bottom"
```
mendelB: @ashleybaldwin Just want to confirm that it's OK to rename this event?

DFurnes: Are we good with this "breaking" any queries/graphs relying on this stat?

resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -90,8 +90,8 @@ class ContentfulEntry extends React.Component<Props, State> {
       case 'competition':
         return renderCompetitionStep(json);
 
-      case 'contentBlock':
-        return renderContentBlock(json, stepIndex);
+      case 'legacyContentBlock':
```
DFurnes: Will this break existing content blocks? Do we need a migration?

mendelB: It shouldn't since we don't actually have any `contentBlock` types anywhere. We were manually overwriting the `campaignActionStep` to be `contentBlock`, so I changed it there as well
https://github.com/DoSomething/phoenix-next/pull/825/files#diff-927b208f806cb78d14c0f0587928cb8bR526

resources/assets/components/Block/LegacyContentBlock.js
mendelB: Yeah, I was thinking this particular data point might not really matter since it's a waypoint. But I'll undo this for now, and we can always change if need be later on.

resources/assets/components/ContentfulEntry/ContentfulEntry.js
DFurnes: Great! 👍

resources/assets/components/Block/LegacyContentBlock.js
```diff
@@ -7,16 +7,16 @@ import SectionHeader from '../SectionHeader';
 import Markdown from '../Markdown';
 import { Flex, FlexCell } from '../Flex';
 
-import './contentBlock.scss';
+import './legacyContentBlock.scss';
```
weerd: not sure why we didn't have this as a hyphenated name like all other scss files 🤔 Either way, not a big deal if we don't update it since we'll eventually deprecate this block. Up to you :)

mendelB: I can change this real quick and do a mergeroo!

mendelB: Ok changed it!

PR #829, opened 2018-04-12
Content block component (skeletal version)

mendelB:
### What does this PR do?

This PR adds a very basic static version of the ContentBlock component along with a renderer and `ContentfulEntry` case.

image
(It's admittedly not looking too hot yet)

### Checklist

### DIFF
--- a/resources/assets/components/ContentfulEntry/renderers.js
+++ b/resources/assets/components/ContentfulEntry/renderers.js
@@ -245,7 +245,7 @@ export function renderContentBlock(data) {
       key={`content-block-${contentfulId}`}
       className="margin-horizontal-md margin-bottom-lg"
     >
-      <ContentBlock {...data.fields} />
+      <ContentBlock id={contentfulId} {...data.fields} />
     </div>
   );
 }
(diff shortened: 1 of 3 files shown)


resources/assets/components/ContentfulEntry/renderers.js
```diff
@@ -229,3 +230,22 @@ export function renderTextSubmissionAction(data) {
...
+ */
+export function renderContentBlock(data) {
+  const contentfulId = data.id;
+
+  return (
+    <div
+      key={`content-block-${contentfulId}`}
+      className="margin-horizontal-md margin-bottom-lg"
+    >
+      <ContentBlock {...data.fields} />
```
weerd: Similar to the `TextSubmissionAction` can we pass it the `id={data.id}` or maybe we use the var so `id={contentfulId}`. Useful to have it on the block!

resources/assets/components/blocks/ContentBlock/ContentBlock.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+
+const ContentBlock = () => <div className="content-block">Content Block!</div>;
```
weerd: Minimalist!

resources/assets/components/ContentfulEntry/renderers.js
mendelB: Done!

PR #832, opened 2018-04-12
Simplify how we handle `modal/:id` routes.

DFurnes:
### What does this PR do?

This pull request updates the `/modal/:id` routes to use the new Modal component, and also preserve real URLs! You can now copy-paste the direct link to a modal or use the back/forward buttons in your browser to open and close it! It all works the way you'd expect! 💥 

screencast

#### Why are we using `StaticRouter`?
We need to "freeze" the background context so it doesn't change when we open a modal. The [example](https://reacttraining.com/react-router/web/example/modal-gallery) from the docs uses a `Switch` for this, but that only works one level deep - when I tried this, our [nested `Redirect`](https://github.com/DoSomething/phoenix-next/blob/c7ea1392ce314063741b0d7738967c7253c36ea4/resources/assets/components/Page/CampaignPage/CampaignPage.js#L102) would always trigger and close the modal! Oy!

[#156351561](https://www.pivotaltracker.com/story/show/156351561)

### Checklist

### DIFF
--- a/resources/assets/components/utilities/ModalLayer/ModalLayer.js
+++ b/resources/assets/components/utilities/ModalLayer/ModalLayer.js
@@ -4,6 +4,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Route } from 'react-router-dom';
 import { StaticRouter } from 'react-router';
+import ReactRouterPropTypes from 'react-router-prop-types';
 
 import ContentfulEntryContainer from '../../ContentfulEntry/ContentfulEntryContainer';
 import Modal from '../../utilities/Modal/Modal';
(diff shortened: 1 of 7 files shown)


resources/assets/components/utilities/ModalLayer/ModalLayer.js
```diff
@@ -0,0 +1,77 @@
+/* global window */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Route } from 'react-router-dom';
+import { StaticRouter } from 'react-router';
+import ReactRouterPropTypes from 'react-router-prop-types';
+
+import ContentfulEntryContainer from '../../ContentfulEntry/ContentfulEntryContainer';
+import Modal from '../../utilities/Modal/Modal';
```
weerd: move above! short -> long import lengths!

PR #834, opened 2018-04-13
Updating Submission Gallery query to filter by post type.

weerd:
### What does this PR do?
This PR updates the `SubmissionGalleryBlock` to pass a prop for the `type` and updates the query to pass the type along to filter posts by type.

Wait to merge this until the [graphql PR](https://github.com/DoSomething/graphql/pull/6) is merged!


weerd:
@DFurnes I figured we'd make the `type` required since we're using the Submission Gallery under each specific action type. Thoughts? 🤔

DFurnes:
> @DFurnes I figured we'd make the `type` required since we're using the Submission Gallery under each specific action type. Thoughts? 🤔 

That makes sense to me!

PR #835, opened 2018-04-13
Add the Content Block -- the real deal

mendelB:
### What does this PR do?

This PR Adds the fully constructed `ContentBlock` component with styling and some testing!

👉 adjusting the `imageAlignment` field in the `ContentBlock` entity to be lowercased

👉 cleaning the `renderer` function a little bit and removing the styling classes from there (they were interfering with the proper internal styling of the ContentBlock)

👉 adjusting the `SectionHeader` component so that the `title` prop is optional (ContentBlock has an optional title and uses `SectionHeader`)

- I dunno how great these tests are, I tried to focus on the specific tricky variables we had and then just general light content testing.

- Everything's pretty straightforward aside from the `imageAlignment`/`image` balancing act. I *tried* to keep it as clean and rational as I could. Please let me know if there's anything improvable there.

- I had to add some weird styling so that the content wasn't hidden on mobile. (the `overflow: hidden` messed with things as well as a missing `margin-bottom`)

- On mobile, the Image appears *above* the content, contrary to the [current campaign action step](https://user-images.githubusercontent.com/12417657/38754793-2bc201fc-3f31-11e8-844f-f0042adbcd1b.png). Is this a problem for us?

### Checklist

###PICS

**Large**
image

**Medium**
image

**Small**
image

### DIFF
--- a/resources/assets/components/Figure/figure.scss
+++ b/resources/assets/components/Figure/figure.scss
@@ -48,7 +48,7 @@
   }
 }
 
-.content-block {
+.figure.-one-third {
   &.figure.-right-collapse > .figure__media {
     margin-left: $base-spacing;
   }
--- a/resources/assets/components/Figure/figure.scss
+++ b/resources/assets/components/Figure/figure.scss
@@ -20,18 +20,47 @@
   margin-bottom: 0;
 }
 
-.figure.-left-collapse {
-  text-align: left;
+.figure {
+  &.-right-collapse,
+  &.-left-collapse {
+    text-align: left;
+
+    @include media($tablet) {
+      > .figure__media {
+        margin-right: gutter();
+        margin-bottom: 0;
+      }
+
+      > .figure__body {
+        overflow: hidden;
+      }
+    }
+  }
 
   @include media($tablet) {
-    > .figure__media {
-      float: left;
-      margin-right: gutter();
-      margin-bottom: 0;
+    &.-right-collapse > .figure__media {
+      float: right;
     }
 
-    > .figure__body {
-      overflow: hidden;
+    &.-left-collapse > .figure__media {
+      float: left;
+    }
+  }
+}
+
+.content-block {
+  &.figure.-right-collapse > .figure__media {
+    margin-left: $base-spacing;
+  }
+
+  &.figure.-left-collapse > .figure__media {
+    margin-right: $base-spacing;
+  }
+  &.figure > .figure__media {
+    width: 33%;
+
+    @include media($small) {
+      width: 100%;
     }
   }
 }
--- a/resources/assets/components/Figure/figure.scss
+++ b/resources/assets/components/Figure/figure.scss
@@ -56,6 +56,7 @@
   &.figure.-left-collapse > .figure__media {
     margin-right: $base-spacing;
   }
+
   &.figure > .figure__media {
     width: 33%;
 
--- a/resources/assets/components/blocks/ContentBlock/ContentBlock.js
+++ b/resources/assets/components/blocks/ContentBlock/ContentBlock.js
@@ -10,11 +10,7 @@ import './content-block.scss';
 const ContentBlock = props => {
   const { content, image, imageAlignment, superTitle, title } = props;
 
-  const imageAlignmentClass =
-    // ensure proper responsiveness setting for the figure component (only required for 'left')
-    (imageAlignment === 'left' && 'left-collapse') ||
-    imageAlignment ||
-    ContentBlock.defaultProps.imageAlignment;
+  const defaultImageAlignment = ContentBlock.defaultProps.imageAlignment;
 
   const contentNode = content ? <Markdown>{content}</Markdown> : null;
 
@@ -31,8 +27,9 @@ const ContentBlock = props => {
           <Figure
             image={image}
             alt="content-block"
-            alignment={imageAlignmentClass}
+            alignment={`${imageAlignment || defaultImageAlignment}-collapse`}
             size="large"
+            className="content-block"
           >
             {contentNode}
           </Figure>
--- a/resources/assets/components/blocks/ContentBlock/ContentBlock.js
+++ b/resources/assets/components/blocks/ContentBlock/ContentBlock.js
@@ -20,9 +20,11 @@ const ContentBlock = props => {
 
   return (
     <div className="content-block">
-      <div className="margin-horizontal-md">
-        <SectionHeader preTitle={superTitle} title={title} hideStepNumber />
-      </div>
+      {title ? (
+        <div className="margin-horizontal-md">
+          <SectionHeader preTitle={superTitle} title={title} hideStepNumber />
+        </div>
+      ) : null}
 
       <div className="margin-horizontal-md">
         {image ? (
--- a/resources/assets/components/blocks/ContentBlock/ContentBlock.js
+++ b/resources/assets/components/blocks/ContentBlock/ContentBlock.js
@@ -2,8 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import Markdown from '../../Markdown';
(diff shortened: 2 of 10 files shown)


sbsmith86:
HI THIS IS MY FAVORITE PR OF ALL PRs

YOU ARE A JEWEL FOR THIS. 

jewel gif-downsized

mendelB:
Thanks Shae! I was hoping you'd notice 😂

resources/assets/components/blocks/ContentBlock/ContentBlock.js
```diff
@@ -1,6 +1,60 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
-const ContentBlock = () => <div className="content-block">Content Block!</div>;
+import Markdown from '../../Markdown';
+import SectionHeader from '../../SectionHeader';
+import { Figure } from '../../Figure';
```
weerd: nit-pick! Shortest to longest!

mendelB: ah dang, I was trying to be careful about that and must have forgotten!

resources/assets/components/blocks/ContentBlock/ContentBlock.js
```diff
@@ -1,6 +1,60 @@
...
+    imageAlignment ||
+    ContentBlock.defaultProps.imageAlignment;
+
+  const contentNode = content ? <Markdown>{content}</Markdown> : null;
+
+  return (
+    <div className="content-block">
+      <div className="margin-horizontal-md">
+        <SectionHeader preTitle={superTitle} title={title} hideStepNumber />
+      </div>
```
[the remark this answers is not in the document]
mendelB: In such a scenario, we still get the yellow bar from the `SectionHeader`
image

Do we not need that?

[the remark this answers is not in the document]
weerd: Nope. we don't want that. We want the `ContentBlock` to be stackable and "look" like continuous blocks of content. Only when there's a `title` do we output the "section-like" title to give the impression of a new section.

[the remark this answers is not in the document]
mendelB: ah ok, I'll add a ternary then. Should it be contingent on there being a `title` or `superTitle`? Or if there's no title we skip it even if there's a `superTitle`?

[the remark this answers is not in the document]
weerd: I think just the title. That's the main item. `superTitle` and eventually `subTitle` should only output if a `title` is provided. Otherwise the editor is just doing it wrong 😬

resources/assets/components/blocks/ContentBlock/ContentBlock.js
```diff
@@ -1,6 +1,60 @@
...
+import { Figure } from '../../Figure';
+import SectionHeader from '../../SectionHeader';
 
-export default ContentBlock;
+import './content-block.scss';
+
+const ContentBlock = props => {
+  const { content, image, imageAlignment, superTitle, title } = props;
+
+  const imageAlignmentClass =
```
DFurnes: This might read better using two separate conditionals:

```js
let imageAlignmentClass = imageAlignment || ContentBlock.defaultProps.imageAlignment;
if (imageAlignmentClass == 'left') {
  imageAlignmentClass = 'left-collapse';
}


```

resources/assets/components/blocks/ContentBlock/ContentBlock.js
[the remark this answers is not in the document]
mendelB: lol fair enough, will do!

resources/assets/components/blocks/ContentBlock/content-block.scss
```diff
@@ -0,0 +1,45 @@
...
+    color: $black;
+    font-family: $primary-font-family;
+  }
+
+  .figure > .figure__media {
+    width: 100%;
+  }
+
+  @include media($medium) {
+    .figure {
```
DFurnes: These overrides make me a little uncomfortable, since they tightly couple this component to the `Figure`'s markup and class names. Could we add these as additional `size` options in Figure?

resources/assets/components/blocks/ContentBlock/content-block.scss
```diff
@@ -0,0 +1,45 @@
+@import '../../../scss/next-toolbox.scss';
+
+.content-block {
+  width: 100%;
+  margin-bottom: 36px;
```
DFurnes: Is there a variable or helper class we can use here?

resources/assets/components/blocks/ContentBlock/ContentBlock.js
mendelB: lol that was how I initially had it and then thought I was being smart 🤕

resources/assets/components/blocks/ContentBlock/content-block.scss
mendelB: Not that I was able to find.
In general, I wasn't sure about the spacing here. This is copied over from the [`LegacyContentBlock`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/Block/legacy-content-block.scss#L5-#L7)
 
There was discussion [here](https://github.com/DoSomething/phoenix-next/pull/668) regarding this. At that point we were under the impression that we'd be moving over to `SectionBlocks` where we could account for actions being embedded within the section so we could space each 'section' apart from each other. But now that the 'thing' under a `ContentBlock` might be an associated action - where we wouldn't want much spacing in between them -- I guess we're back to the same question of how to space this?

One thought I had was to take care of the `margin-bottom` in the `actionSteps` where we could space accordingly for `contentBlocks` and corresponding actions. similarly to how we handle [column widths](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/Page/ActionPage/ActionSteps.js#L87) if that makes sense.

resources/assets/components/Figure/figure.scss
```diff
@@ -20,18 +20,48 @@
...
+
+    &.-left-collapse > .figure__media {
       float: left;
-      margin-right: gutter();
-      margin-bottom: 0;
     }
+  }
+}
+
+.content-block {
```
DFurnes: I don't think we should have to override `.content-block` from `Figure`. Should this be a new "size" modifier?

resources/assets/components/blocks/ContentBlock/content-block.scss
DFurnes: > One thought I had was to take care of the `margin-bottom` in the `actionSteps` where we could space accordingly for `contentBlocks` and corresponding actions. similarly to how we handle [column widths](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/Page/ActionPage/ActionSteps.js#L87) if that makes sense.

Yeah, I feel like this makes the most sense in `ActionSteps` (or whatever that becomes in our new "everything a page" world). Since that's all in flux, this seems okay to leave for now. 👍

resources/assets/components/Figure/figure.scss
mendelB: Oh yeah, that does feel way better! I'm just not sure on the naming/categorization of this -- would `x-large` be appropriate?

mendelB: ...or `one-third` kinda like the flex stuff?

weerd: Yeah, I think that's fine for now. After the whole slugs/pages changes, I want to spend some time implementing Tailwind CSS and that'll change a bunch of stuff across the board for styles anyhoo.

mendelB:
Updated to use `one-third` "size" prop for styling. I'll merge once this builds successfully! (no real changes here)

PR #838, opened 2018-04-16
Remove /pages/ from campaign subpage slugs and redirect existing ones

weerd:
### What does this PR do?
This PR provides the first step in the effort to revise how our system works with slugs for campaign subpages. It removes the `/pages/` from the URLs to campaign subpages and also updates the backend routes to handle redirecting requests to campaigns that contain `/pages/`.

Also, instead of redirecting campaign subpages with a slug that cannot be matched, back to the main campaign root page, we're going to just let the campaign subpage show a NotFound component.

Refs [Pivotal ID #156351793](https://www.pivotaltracker.com/story/show/156351793)

### Checklist

### DIFF
--- a/resources/assets/components/Page/CampaignPage/CampaignPage.js
+++ b/resources/assets/components/Page/CampaignPage/CampaignPage.js
@@ -77,10 +77,6 @@ const CampaignPage = props => {
             }}
           />
           <Route
-            path={`${match.url}/pages/:slug`}
-            component={CampaignSubPageContainer}
-          />
-          <Route
             path={`${match.url}/blocks/:id`}
             component={BlockPageContainer}
           />
@@ -88,8 +84,15 @@ const CampaignPage = props => {
             path={`${match.url}/quiz/:slug`}
             component={BlockPageContainer}
           />
-          {/* If no route matches, just redirect back to the main page: */}
-          <Redirect from={`${match.url}/:anything`} to={`${match.url}`} />
+          {/* @deprecate: remove this Route specification with `/pages/:slug` */}
+          <Route
+            path={`${match.url}/pages/:slug`}
+            component={CampaignSubPageContainer}
+          />
+          <Route
+            path={`${match.url}/:slug`}
+            component={CampaignSubPageContainer}
+          />
         </Switch>
       </div>
       <CampaignFooter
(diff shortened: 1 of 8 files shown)


resources/assets/components/Page/CampaignPage/CampaignPage.js
```diff
@@ -88,8 +84,15 @@ const CampaignPage = props => {
             path={`${match.url}/quiz/:slug`}
             component={BlockPageContainer}
           />
-          {/* If no route matches, just redirect back to the main page: */}
-          <Redirect from={`${match.url}/:anything`} to={`${match.url}`} />
+          {/* @deprecate: remove this Route specification with `/pages/:slug` */}
+          <Route
```
DFurnes: Let's redirect this to the new path rather than rendering both routes:

```js
<Redirect from={`${match.url}/pages/:slug`} to={`${match.url}/:slug`}>
```

routes/web.php
```diff
@@ -20,6 +20,11 @@
 $router->get('us/campaigns', 'CampaignController@index');
 $router->redirect('campaigns', 'us/campaigns');
 
+// Redirect routes for campaign specific URLs containing "/pages/".
+$router->get('us/campaigns/{slug}/pages/{clientRoute?}', function ($slug, $clientRoute) {
```
DFurnes: I think it makes sense to do this in the app itself, rather than introducing more server-side logic.

resources/assets/components/Page/CampaignPage/CampaignPage.js
weerd: Ah yeah, ok. I'll update.

weerd: _Update!_ Tried doing the redirect on the client side instead and there seems to be something odd going on w/ the `to` attribute in the `Redirect` component, which doesn't accept the `:slug` token properly. Chatted with @DFurnes and we'll leave it with server redirecting for now.

mendelB: BTW just thought this was interesting -- a workaround for this issue:
https://stackoverflow.com/a/46064986/5657278

PR #840, opened 2018-04-16
Fix Puck event tracking for ModalLauncher.

DFurnes:
### What does this PR do?

This pull request re-adds the `open modal` Puck event to the two modals launched by the `ModalLauncher`, since we need that event to track conversion on those two flows.

screen shot 2018-04-16 at 4 51 16 pm

This event [was triggered](https://github.com/DoSomething/phoenix-next/pull/836/files#diff-dc9bb216b849169350e03ebe7c39bcc2L38) in the old modal implementation, but didn't make it's way over to the new one.

To reduce some noise, I was thinking we'd only track modals that don't appear as part of the normal flow. (Since post-signup and post-reportback modals always appear alongside their associated events, and the `modal/:id` pages should be tracked as normal page views).

N/A

### Checklist

### DIFF
--- a/resources/assets/components/ModalLauncher/ModalLauncher.js
+++ b/resources/assets/components/ModalLauncher/ModalLauncher.js
@@ -58,8 +58,19 @@ class ModalLauncher extends React.Component {
   };
 
   render() {
+    // Set the proper Puck "event name" so we can
+    // continue tracking opens of timed modals.
+    let type = this.props.type;
+    if (type === 'voter_reg_modal') {
+      type = 'VOTER_REGISTRATION_MODAL';
+    } else if (type === '') {
+      type = 'SURVEY_MODAL';
+    }
+
     return this.state.showModal ? (
-      <Modal onClose={this.handleClose}>{this.props.render()}</Modal>
+      <Modal trackingId={type} onClose={this.handleClose}>
+        {this.props.render()}
+      </Modal>
     ) : null;
   }
 }
--- a/resources/assets/components/ModalLauncher/ModalLauncher.js
+++ b/resources/assets/components/ModalLauncher/ModalLauncher.js
@@ -1,8 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { get } from 'lodash';
 
 import Modal from '../utilities/Modal/Modal';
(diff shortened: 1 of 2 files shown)


DFurnes:
For context, here's the [original discussion](https://github.com/DoSomething/phoenix-next/commit/3c9e6a503d36947d6648f22f4d640bb81a24cbcc#commitcomment-28605925) where @mendelB caught this!

resources/assets/components/ModalLauncher/ModalLauncher.js
```diff
@@ -58,8 +58,19 @@ class ModalLauncher extends React.Component {
   };
 
   render() {
+    // Set the proper Puck "event name" so we can
+    // continue tracking opens of timed modals.
+    let type = this.props.type;
+    if (type === 'voter_reg_modal') {
+      type = 'VOTER_REGISTRATION_MODAL';
+    } else if (type === '') {
```
mendelB: Wouldn't the `type` be `'nps_survey'` in this case?

DFurnes: OMG I AM REAL DUMB. Yes! 😇

weerd:
Yeah, I think I prefer this approach instead of only handling tracking in the `ModalLauncher`. This feels a bit more versatile for future use cases. I dig it! It doesn't add to much to the `Modal` that muddies it up.

resources/assets/components/ModalLauncher/ModalLauncher.js
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { get } from 'lodash';
 
 import Modal from '../utilities/Modal/Modal';
-import { get, set } from '../../helpers/storage';
+import { get as getStorage, set as setStorage } from '../../helpers/storage';
```
weerd: random, but we should probs name these storage methods like this in general to not confuse w/ the Lodash `get` even in files where Lodash `get` isn't being used!

DFurnes: Yeah, I agree. It's tripped me up a few times. 🏪

PR #844, opened 2018-04-17
Contentful slug transform script

mendelB:
### What does this PR do?

This PR adds a script that uses the [contentful-management-api (js sdk)](https://contentful.github.io/contentful-management.js/contentful-management/5.0.0/) to transform all `Page` - `slug` fields over to the [new format](https://github.com/orgs/DoSomething/teams/team-rocket/discussions/17)

To run this script (in command line from `phoenix-next` project root directory):
```
$ node contentful/management-api-scripts/[transformation-filename] --space-id [contentful space id] --access-token [contentful-access-token]
```

- We tried doing this with our usual transformation process using the Contentful Migration CLI, however, we ran into a roadblock with needing a `Campaign`s own `slug`field  in order to transform it's individual linked `pages` `slug` fields (`campaign-slug/page-slug`). However, the CLI only allows us access to the `Page`s individually when targeting `Page` directly (without any means of knowing the source campaigns slug), or only the `linkEntry` reference to the pages when targeting the `campaign` itself for the transformation. [Slack thread](https://dosomething.slack.com/archives/C3ASB4204/p1523914344000281)

- There would be *way* fewer API calls being made in the script, however, frustratingly, the [`include`](https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/links) and [`resolveLinks`](https://www.contentful.com/developers/docs/javascript/tutorials/using-js-cda-sdk/#retrieving-entries-with-search-parameters) options don't seem to be working at all with this sdk.

- I tested this all using my personal Contentful space, modeled with Campaigns and pages etc. So this *should* work with our space as well.
comic

https://www.pivotaltracker.com/story/show/156771576 -- I think

### DIFF
--- a/contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
+++ b/contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
@@ -11,7 +11,7 @@ if (!spaceId || !accessToken) {
   return;
 }
 
-async function transformPageSlugs() {
+async function getEnvironment() {
   const client = contentful.createClient({
     // This is the access token for this space. Normally you get both ID and the token in the Contentful web app
     accessToken: accessToken,
@@ -22,7 +22,11 @@ async function transformPageSlugs() {
 
   // This API call will request the environment with the specified - as of now hardcoded - id
   // (Contentful requires this scoping, as the `space.getEntries` is being deprecated)
-  const environment = await space.getEnvironment('master');
+  return space.getEnvironment('master');
+}
+
+async function transformPageSlugs() {
+  const environment = await getEnvironment();
 
   // Now that we have an environment, we can get entries from that environment
   const campaigns = await environment.getEntries({
--- a/contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
+++ b/contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
@@ -62,7 +62,12 @@ async function transformPageSlug(campaign, page, environment) {
   const campaignSlug = get(campaign.fields.slug, locale);
   const pageSlug = get(pageEntry.fields.slug, locale);
 
-  if (!campaignSlug || !pageSlug || pageSlug.indexOf(campaignSlug) === 0) {
+  if (
+    !campaignSlug ||
+    !pageSlug ||
+    pageSlug.indexOf(campaignSlug) === 0 ||
+    pageSlug.indexOf(campaignSlug) === 1
+  ) {
     return;
   }
 
--- a/contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
+++ b/contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
@@ -3,12 +3,19 @@ const contentful = require('contentful-management');
 const spaceId = process.argv[3];
 const accessToken = process.argv[5];
 
+const locale = 'en-US';
+
 if (!spaceId || !accessToken) {
   console.log(
     'Please provide the space-id and access token arguments in the following format:',
   );
   console.log('--space-id [space-id] --access-token [access-token]');
   return;
+}
+
+// Helper Function
+function sleep(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
 }
 
 async function getEnvironment() {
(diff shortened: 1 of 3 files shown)


weerd:
@mendelB mind giving a quick example of how the command is run somewhere in the PR description :)

contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+async function transformPageSlug(campaign, page, environment) {
+  const pageEntry = await environment.getEntry(page.sys.id);
+
+  if (!pageEntry.fields.slug || !campaign.fields.slug) {
+    return;
+  }
+
+  pageEntry.fields.slug[locale] = `${campaign.fields.slug[locale]}/${
```
weerd: Use `join()` from `path` (@see the `TabbedNavigationContainer`) to generate the new slug since it's made to deal w/ urls well :)

weerd: Also, before reconstructing the slug to include the campaign slug, I think it'd be best to first do a check to see if the campaign slug is already part of the page slug value. Similar to how we do a check here: https://github.com/DoSomething/phoenix-next/blob/74db19f823f36bcbf0ec7d0e80717a440e7b7bdc/resources/assets/components/Navigation/TabbedNavigationContainer.js#L49

I definitely know that a page on TestTFJ already has the campaign slug as part of the page slug value!

mendelB:
@weerd Updated!

contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
```diff
@@ -0,0 +1,80 @@
...
+  });
+}
+
+async function transformPageSlug(campaign, page, environment) {
+  const pageEntry = await environment.getEntry(page.sys.id);
+
+  const campaignSlug = get(campaign.fields.slug, locale);
+  const pageSlug = get(pageEntry.fields.slug, locale);
+
+  if (!campaignSlug || !pageSlug || pageSlug.indexOf(campaignSlug) === 0) {
```
weerd: Instead of `===` do we want it to be just `>=` to be safe? So the campaign slug exists _somewhere_ in the string; very likely it'll be the first one. Not a big deal, just a small concern :)

mendelB: Yeah, I can totally change to that! (My thinking was that we *only* cared if it was at the very beginning)

weerd: I just would prefer to be safe incase the value for the slug contained the campaign slug but started with a `/` before it. I _think_ that might mean the campaign slug is index `1` and not `0` but I haven't actually tried that. I could very well be wrong! haha 😅

weerd: I doubt we have that and likely only have 1 maybe 2 page slugs with the campaign slug though. Which is why it's probs fine :)

contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
```diff
@@ -0,0 +1,80 @@
...
+
+  pageEntry.fields.slug[locale] = join(campaignSlug, pageSlug);
+
+  pageEntry.update().then(page => page.publish());
+
+  console.log(`Updated Page! [ID: ${pageEntry.sys.id}]`);
+  console.log(`--> ${pageEntry.fields.slug[locale]}\n`);
+
+  // API breather room
+  sleep(1000);
```
DFurnes: Do we need to `await` this?

contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
mendelB: 👍 I added that condition to the mix.

contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
```diff
@@ -0,0 +1,80 @@
+const { join } = require('path');
+const { get } = require('lodash');
+const contentful = require('contentful-management');
+
+const spaceId = process.argv[3];
+const accessToken = process.argv[5];
```
DFurnes: It might be handy to use [minimist](https://github.com/substack/minimist) for this so we don't accidentally mix up the argument order!

contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
```diff
@@ -62,7 +62,12 @@ async function transformPageSlug(campaign, page, environment) {
   const campaignSlug = get(campaign.fields.slug, locale);
   const pageSlug = get(pageEntry.fields.slug, locale);
 
-  if (!campaignSlug || !pageSlug || pageSlug.indexOf(campaignSlug) === 0) {
+  if (
+    !campaignSlug ||
+    !pageSlug ||
+    pageSlug.indexOf(campaignSlug) === 0 ||
+    pageSlug.indexOf(campaignSlug) === 1
```
weerd: oh @mendelB just use `>= 0` and it should account for it. Just means it found it, you don't need to account for every location 😉

contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
weerd: that's a cool little package! 🕶

mendelB: Oh swag! I'll add that in

contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
mendelB: Ah, yes! 😬

mendelB:
Updates based on feedback:
- Added `minimist`
- cleaned conditional
- added `await`
- removed one of the `sleep`s.

PR #847, opened 2018-04-18
WIP: Shrink vendor bundle.

DFurnes:
### What does this PR do?
This pull request attempts to shrink our hefty vendor bundle a bit. I've also added a `npm run analyze` command that uses the lovely [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer) to [visualize what's taking up space](https://user-images.githubusercontent.com/583202/38949318-043061d2-4310-11e8-8004-594c3766357e.png)!

| Step | Bundle Size | Bundle Size (gzipped) |
|------|-------------|-----------------------|
| _(initial)_ | 843kb | 257kb |
| 🌾 Swapped `lodash` references for `lodash-es`, which is bundled as tree-shakable ES modules. (7431547) | 850kb | 256kb |
| 🎡 Export ES modules from Gateway so we don't pull in full Lodash when it just needs a few tiny helpers. (DoSomething/gateway#19, …) | 774kb | 233kb |

I used the wonderful `

Fixes #

### Checklist


weerd:
🔥

DFurnes:
Rebased to fix a `package-lock.json` conflict.

DFurnes:
Alright, this is ready for review! I'm pretty happy with where things are at: 843kb → 642kb (-201kb) for the vendor bundle (down to 197kb gzipped)! Way smaller! 🙆‍♂️

There's also another fun "bonus" savings I realized while digging into all this... we're not currently gzipping assets on production. 😱 We definitely fixed this in the past, but not for the last time it seems! After this is merged, I'm going to see about re-adding this to our Fastly config so we can get that sweet sweet performance boost.

weerd:
I think my favorite thing is the awesome table in the description! 🎉 nice @DFurnes 👏

resources/assets/helpers/index.js
```diff
@@ -368,7 +368,7 @@ export function isCampaignClosed(endDate) {
     return false;
   }
 
-  return isPast(parse(endDate));
+  return isBefore(new Date(endDate), new Date());
```
weerd: I looked into this with some of the stuff I'm working on to refactor the page nav, and `parse` is actually used in both `isBefore()` and `isPast()` so I think you can just pass the date as is and don't necessarily need to pass the first param date through `new Date()`:

```
function isBefore (dirtyDate, dirtyDateToCompare) {
  var date = parse(dirtyDate)
  var dateToCompare = parse(dirtyDateToCompare)
  return date.getTime() < dateToCompare.getTime()
}
```

weerd:
> I'm also interested in making a CI task to graph this over time... hack week?? 😺

That'd be real cool!

weerd:
> There's also another fun "bonus" savings I realized while digging into all this... we're not currently gzipping assets on production. 😱 We definitely fixed this in the past, but not for the last time it seems! After this is merged, I'm going to see about re-adding this to our Fastly config so we can get that sweet sweet performance boost.

I never really understood where the gzipping happens? Is that something we set on the server? Or because we use Fastly we set it via our CDN?

DFurnes:
> I never really understood where the gzipping happens? Is that something we set on the server? Or because we use Fastly we set it via our CDN?

There's multiple options - we could gzip at build time (and then just set the appropriate header when serving it), gzip on demand on the server (adds CPU load), or let Fastly handle it. Since we put Fastly in front of most of our traffic, it seems like a handy (and easy) place to handle it!

resources/assets/helpers/index.js
DFurnes: This is true!

PR #851, opened 2018-04-20
Contentful Management API scripts refactor

mendelB:
### What does this PR do?

For the new content management script I was working on, I realized there's a bunch of setups that kinda needs to get repeated. So I figured it'd be nice to set up a general 'Client' that takes care of the basic configuration and fetching the environment client we use for entry updating and then yields to whatever custom script is running.

I hope this setup makes sense and is somewhat clean!

This will pave the way for upcoming PR with another page manipulation script.

[Pivotal #156771576](https://www.pivotaltracker.com/story/show/156771576)

### DIFF
--- a/contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
+++ b/contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
@@ -1,81 +1,49 @@
 const { join } = require('path');
-const { get } = require('lodash');
-const parseArgs = require('minimist');
-const contentful = require('contentful-management');
+const { contentManagementClient } = require('./contentManagementClient');
 
-const locale = 'en-US';
+const LOCALE = contentManagementClient.LOCALE;
+const { sleep, getField } = contentManagementClient.helpers;
 
-const args = parseArgs(process.argv, {
-  alias: { 'access-token': 'accessToken', spaceId: 'space-id' },
-});
-
-const { spaceId, accessToken } = args;
-
-if (!spaceId || !accessToken) {
-  console.log(
-    'Please provide the space-id and access token arguments in the following format:',
-  );
-  console.log('--space-id [space-id] --access-token [access-token]');
-  return;
-}
-
-// Helper Function
-function sleep(ms) {
-  return new Promise(resolve => setTimeout(resolve, ms));
-}
-
-async function getEnvironment() {
-  const client = contentful.createClient({
-    // This is the access token for this space. Normally you get both ID and the token in the Contentful web app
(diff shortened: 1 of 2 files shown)


contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
```diff
@@ -1,81 +1,49 @@
...
   console.log(`Updated Page! [ID: ${pageEntry.sys.id}]`);
-  console.log(`--> ${pageEntry.fields.slug[locale]}\n`);
+  console.log(`--> ${pageEntry.fields.slug[LOCALE]}\n`);
 
   // API breather room
   await sleep(1000);
 }
 
-transformPageSlugs();
+contentManagementClient.init(function(environment) {
```
weerd: Where does `environment` get passed in?

mendelB: That comes from [here](https://github.com/DoSomething/phoenix-next/pull/851/files#diff-0b148c3293c7c052c64d53008b7fc9beR22)

I thought about it like the migration scripts which pass the `migration` object to the exported functions

contentful/management-api-scripts/contentManagementClient.js
```diff
@@ -0,0 +1,59 @@
+const { get } = require('lodash');
+const parseArgs = require('minimist');
+const contentful = require('contentful-management');
+
+const LOCALE = 'en-US';
+
+async function initContentManagementClient(callback) {
+  const args = parseArgs(process.argv, {
+    alias: { 'access-token': 'accessToken', spaceId: 'space-id' },
```
weerd: I might not have caught this in prior PR, but here for `access-token` the key is hyphenated, but in `spaceId` the key is not? I think `access-token` and `accessToken` need to be switched right?

contentful/management-api-scripts/2018_04_17_001_transform_page_slugs.js
weerd: ah cool. ok!

contentful/management-api-scripts/contentManagementClient.js
mendelB: Oh LOL totally missed this too!
When I initially set it up, I was switching around the order, and it seems that it doesn't care which way you do it which is cool. But will fix for consistency!

weerd: > it seems that it doesn't care which way you do it which is cool.

😅 haha

PR #852, opened 2018-04-20
Activity community page transformation script

mendelB:
### What does this PR do?

This PR adds a new Content Management API script that will create an ~Action and~ Community Page for each campaign and port over the ~`actionSteps` and~ `activityFeed` blocks over to ~Action and~ Community ~respectively~.

To run this script (in command line from `phoenix-next` project root directory):
```bash
$ node contentful/management-api-scripts/[transformation-filename] --space-id [contentful space id] --access-token [contentful-access-token]
```

To run this script for just a single campaign, you can add the optional `--campaign-id` argument with the campaign's Contentful ID:
```bash
$ node contentful/management-api-scripts/[transformation-filename] --space-id [contentful space id] --access-token [contentful-access-token] --campaign-id [contentful-campaign-id]
```

Edge cases I was able to think of:
- We'll skip any campaign with no `slug` or `internalTitle`
- We'll only create the pages *if* the campaign has the respective fields set. (So if it doesn't have any `activityFeed` blocks, we won't create a Community page for it) -- do we want to create *both* pages no matter what?

I based the `attempt` helper function - `async await` coupling off of this [post](https://blog.grossman.io/how-to-write-async-await-without-try-catch-blocks-in-javascript/)

[Pivotal #156771690](https://www.pivotaltracker.com/story/show/156771690)


mendelB:
**Update**:

- Added the ability to pass a `campaign-id` argument for running the script on a single campaign
- Refactored a bunch to include safety catches for each API call to ensure we're logging properly, and not breaking down from a single failed call. It adds a bunch of noise to the script but I think it's worth it. 
- Added an `attempt` helper method, so we can catch and log errors while maintaining the use of `async` in what I *think* is a fairly elegant way relative to having to use `.then`s

mendelB:
**Update:**

- Added `/helpers` file, and added a few more helper functions for what seem to be repeating bits of logic.
- Moved out the logic for action Pages and actionSteps. This script is now soley creating Community pages from `activityFeed` fields.

mendelB:
I think this is now ready for review!

mendelB:
**Update**:

Added a log file, to help preserve our logs, and prevent losing data from a 'terminal space overflow' (holy crap that sounds awesome)

contentful/management-api-scripts/helpers.js
```diff
@@ -0,0 +1,88 @@
...
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+function getField(entry, field, defaultVal = null) {
+  return get(entry.fields[field], LOCALE, defaultVal);
+}
+
+// Catch and log any callback errors
+function attempt(callback) {
+  return callback().catch(error => log(error.message));
```
DFurnes: This is really clever! ⚾️

contentful/management-api-scripts/2018_04_20_001_campaign_activity_feed_to_community_page.js
```diff
@@ -0,0 +1,107 @@
...
+    log('--------------------------------------------\n');
+    return;
+  }
+
+  log(`Processing Campaign! [ID: ${campaign.sys.id}]\n`);
+
+  // Create a new community 'Page' with the activityFeed blocks from the source campaign
+  const communityPage = await attempt(() =>
+    environment.createEntry('page', {
+      fields: {
```
DFurnes: It could be slick to have a helper that maps over a simple object & formats fields like this for you, e.g:

```js
environment.createEntry('page', withFields({
   internalTitle: `${campaignInternalTitle} Community Page`,
   title: 'Community Page',
   slug: join(campaignSlug, 'community'),
   blocks: campaignActivityFeed
});
```

contentful/management-api-scripts/2018_04_20_001_campaign_activity_feed_to_community_page.js
```diff
@@ -0,0 +1,107 @@
...
+        internalTitle: {
+          [LOCALE]: `${campaignInternalTitle} Community Page`,
+        },
+        title: {
+          [LOCALE]: 'Community Page',
```
DFurnes: Should we title this "Community" since that's what shows on the tab navigation?

mendelB:
OK **FINAL Update**: 🤞
- Added Winston package for logging
- Added a `withFields` helper for entry creation
- Renamed `Community Page` to `Community`

PR #853, opened 2018-04-20
Conditionally load polyfills if browser requires them.

DFurnes:
### What does this PR do?
This pull request conditionally loads polyfills using [Polyfill.io](http://polyfill.io), a service created by the Financial Times team that delivers just the polyfills a user's browser needs based on it's user-string. Additionally, we short-circuit and avoid making a request at all for browsers with the newest web platform features (where we can be pretty darn sure they'll have everything else).

This saves most users an additional 71kb – bringing our vendor bundle down to 568kb, 173kb gzipped!

N/A

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -283,11 +283,6 @@
         "webpack-merge": "4.1.2"
       }
     },
-    "@publica/url-polyfill": {
-      "version": "0.5.8",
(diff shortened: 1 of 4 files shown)


DFurnes:
One thing I've noticed is that we no longer include [regenerator-runtime](https://www.npmjs.com/package/regenerator-runtime) when this is enabled, so we'll have to hold on using `async` and `await` in client-side code temporarily. Looking into whether we can have separate build targets to solve this, or if not going to manually re-include it in the build!

PR #859, opened 2018-04-26
Updating the campaign page routing to simplify it.

weerd:
### What does this PR do?
This PR updates the Campaign page routing to help simplify things a bit, and remove some of the convoluted logic to help support some upcoming fixes to the page navigation.

Separated this out into a different PR so that the page navigation updates PR doesn't get too unwieldy!

Refs [Pivotal ID #156351793](https://www.pivotaltracker.com/story/show/156351793)

### DIFF
--- a/resources/assets/components/pages/CampaignPage/CampaignPage.js
+++ b/resources/assets/components/pages/CampaignPage/CampaignPage.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { get } from 'lodash';
+import { join } from 'path';
 import PropTypes from 'prop-types';
 import ReactRouterPropTypes from 'react-router-prop-types';
 import { Switch, Route, Redirect } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/CampaignPage/CampaignPage.js
```diff
@@ -1,13 +1,12 @@
 import React from 'react';
-import { get } from 'lodash';
+import { join } from 'path';
```
weerd: I ended up bringing in `join` because I noticed some funky behavior with using template literals to construct the urls for React Router when you tried to hit a url with an existing forward slash; it would double up the slashes (eg: `/us/campaign-name//action`.

resources/assets/components/pages/CampaignPage/CampaignPage.js
```diff
@@ -49,43 +32,54 @@ const CampaignPage = props => {
...
           </Modal>
         ) : null}
+
         <Switch>
           <Route
             path={`${match.url}`}
             exact
-            component={ActionPageOrActivityFeed}
+            render={() => {
+              if (isCampaignClosed && hasCommunityPage) {
```
DFurnes: It might read a little nicer if we just vary the path here:

```js
const path = isCampaignClosed && hasCommunityPage ? 'community' : 'action';

return <Redirect to={join(match.url, path)} />;
```

weerd: Oh yeah, I dig that!

resources/assets/components/pages/CampaignPage/CampaignPage.js
```diff
@@ -49,43 +32,53 @@ const CampaignPage = props => {
...
+
+              return <Redirect to={join(match.url, path)} />;
+            }}
           />
+
           <Route
-            path={`${match.url}/action`}
-            component={ActionPageOrActivityFeed}
+            path={join(match.url, 'action')}
+            component={ActionPageContainer}
```
mendelB: Are we cool with people being able to hit the action page even after a campaign is closed?

PR #869, opened 2018-04-30
Adding new PhotoSubmissionAction entity and other setup.

weerd:
### What does this PR do?
This PR simply adds an entity file for the new `PhotoSubmissionAction` along with an case entry in the `ContentfulEntry` component.

💅 

### DIFF
--- a/app/Entities/PhotoSubmissionAction.php
+++ b/app/Entities/PhotoSubmissionAction.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace App\Entities;
+
+use JsonSerializable;
+
+class PhotoSubmissionAction extends Entity implements JsonSerializable
+{
+    public function jsonSerialize()
+    {
+        return [
+            'id' => $this->entry->getId(),
+            'type' => $this->getContentType(),
+            'fields' => [
+                'title' => $this->title,
+                'captionFieldLabel' => $this->captionFieldLabel,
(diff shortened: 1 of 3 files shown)


app/Entities/PhotoSubmissionAction.php
```diff
@@ -0,0 +1,29 @@
...
+                'title' => $this->title,
+                'captionFieldLabel' => $this->captionFieldLabel,
+                'captionFieldPlaceholder' => $this->captionFieldPlaceholder,
+                'showQuantityField' => $this->showQuantityField,
+                'quantityFieldLabel' => $this->quantityFieldLabel,
+                'quantityFieldPlaceholder' => $this->quantityFieldPlaceholder,
+                'whyParticipatedFieldLabel' => $this->whyParticipatedFieldLabel,
+                'whyParticipatedFieldPlaceholder' => $this->whyParticipatedFieldPlaceholder,
+                'buttonText' => $this->buttonText,
+                'affirmationContent' => $this->affirmationContent ?: null,
```
mendelB: Why do we need the manual `null` set here?

weerd: Yeah, I couldn't remember but it's something I did in the `TextSubmissionAction`. I'm gonna double check while working on the photo submission and revise this in both locations if not needed :)

PR #874, opened 2018-05-01
Updating logic to disable access to Action page if the campaign is closed

weerd:
### What does this PR do?
This PR updates the logic in the `CampaignPage` component to disable access to the Action page if the campaign is closed (and it has a Community page to redirect to). It does not disable access to the Action page if the user is an Admin. cc @ashleybaldwin 😸 

Tested this a bit and seems to work as expected, even if the conditional reads a little odd. We'll likely move this logic to allow pages to decide their "visibility" instead of having a slew of conditionals in the `CampaignPage`, but that's an upcoming change!

### Checklist

### DIFF
--- a/resources/assets/components/pages/CampaignPage/CampaignPageContainer.js
+++ b/resources/assets/components/pages/CampaignPage/CampaignPageContainer.js
@@ -3,6 +3,7 @@ import { connect } from 'react-redux';
 
 import CampaignPage from './CampaignPage';
 import { isCampaignClosed } from '../../../helpers';
+import { userHasRole } from '../../../selectors/user';
 import { convertExperiment, clickedHideAffirmation } from '../../../actions';
 
 /**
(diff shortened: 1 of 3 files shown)


resources/assets/components/pages/CampaignPage/CampaignPageContainer.js
```diff
@@ -14,6 +15,7 @@ const mapStateToProps = state => ({
   campaignLead: get(state, 'campaign.campaignLead.fields', null),
   hasActivityFeed: Boolean(state.campaign.activityFeed.length),
   hasCommunityPage: Boolean(state.campaign.activityFeed.length),
+  isAdmin: userHasRole(state, 'admin'),
```
mendelB: Do we want to allow the `'staff'` role as well?

weerd: We can add that if asked for it. Let's keep it as just Admin's so for staff the experience of the campaign is closer to the user experience. We can easily change if needed. Good ques tho 👍

PR #875, opened 2018-05-01
Quick fix to simplify logic for Action page.

weerd:
### What does this PR do?
This PR fixes logic bug when user is not an admin. Decided to split out the logic so it is a bit simpler and works correctly.

Non-admin users were not able to get to the Action page, when the campaign wasn't closed and it did have a Community page.

### DIFF
--- a/resources/assets/components/pages/CampaignPage/CampaignPage.js
+++ b/resources/assets/components/pages/CampaignPage/CampaignPage.js
@@ -49,11 +49,15 @@ const CampaignPage = props => {
           <Route
             path={join(match.url, 'action')}
             render={() => {
-              if (isAdmin || (!isCampaignClosed && !hasCommunityPage)) {
+              if (isAdmin) {
                 return <ActionPageContainer />;
               }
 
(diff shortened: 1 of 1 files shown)


resources/assets/components/pages/CampaignPage/CampaignPage.js
```diff
@@ -49,11 +49,15 @@ const CampaignPage = props => {
           <Route
             path={join(match.url, 'action')}
             render={() => {
-              if (isAdmin || (!isCampaignClosed && !hasCommunityPage)) {
+              if (isAdmin) {
                 return <ActionPageContainer />;
               }
 
-              return <Redirect to={join(match.url, 'community')} />;
+              if (isCampaignClosed && hasCommunityPage) {
```
DFurnes: Could these be one `isCampaignClosed && hasCommunityPage && !isAdmin` conditional?

weerd: I can try that.

PR #878, opened 2018-05-02
Photo submission action component setup

weerd:
### What does this PR do?
This PR adds the new `PhotoSubmissionAction` component and some of the setup work to help get it connected. It only outputs the form with basic rendering but not much else for now.

image

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -917,7 +917,7 @@
     },
     "babel-eslint": {
       "version": "8.2.2",
-      "resolved": "http://registry.npmjs.org/babel-eslint/-/babel-eslint-8.2.2.tgz",
+      "resolved": "https://registry.npmjs.org/babel-eslint/-/babel-eslint-8.2.2.tgz",
       "integrity": "sha512-Qt2lz2egBxNYWqN9JIO2z4NOOf8i4b5JS6CFoYrOZZTDssueiV1jH/jsefyg+86SeNY3rB361/mi3kE1WK2WYQ==",
       "dev": true,
       "requires": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -917,7 +917,7 @@
     },
     "babel-eslint": {
       "version": "8.2.2",
-      "resolved": "https://registry.npmjs.org/babel-eslint/-/babel-eslint-8.2.2.tgz",
+      "resolved": "http://registry.npmjs.org/babel-eslint/-/babel-eslint-8.2.2.tgz",
       "integrity": "sha512-Qt2lz2egBxNYWqN9JIO2z4NOOf8i4b5JS6CFoYrOZZTDssueiV1jH/jsefyg+86SeNY3rB361/mi3kE1WK2WYQ==",
       "dev": true,
       "requires": {
@@ -4676,7 +4676,7 @@
     },
     "event-stream": {
       "version": "3.3.4",
-      "resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
+      "resolved": "http://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
       "integrity": "sha1-SrTJoPWlTbkzi0w02Gv86PSzVXE=",
       "dev": true,
       "requires": {
--- a/resources/assets/components/ContentfulEntry/renderers.js
+++ b/resources/assets/components/ContentfulEntry/renderers.js
@@ -203,6 +203,14 @@ export function renderTextSubmissionAction(data) {
   );
 }
 
+export function renderPhotoSubmissionAction(data) {
+  const contentfulId = data.id;
+
+  console.log(contentfulId);
+
+  return null;
+}
+
 /**
  * Render the affirmation step.
  *
--- a/resources/assets/components/ContentfulEntry/renderers.js
+++ b/resources/assets/components/ContentfulEntry/renderers.js
@@ -10,6 +10,7 @@ import LinkActionContainer from '../actions/LinkAction/LinkActionContainer';
 import ShareActionContainer from '../actions/ShareAction/ShareActionContainer';
 import ThirdPartyActionContainer from '../actions/ThirdPartyAction/ThirdPartyActionContainer';
 import TextSubmissionActionContainer from '../actions/TextSubmissionAction/TextSubmissionActionContainer';
+import PhotoSubmissionActionContainer from '../actions/PhotoSubmissionAction/PhotoSubmissionActionContainer';
 import SubmissionGalleryBlockContainer from '../blocks/SubmissionGalleryBlock/SubmissionGalleryBlockContainer';
 import VoterRegistrationActionContainer from '../actions/VoterRegistrationAction/VoterRegistrationActionContainer';
 
@@ -208,7 +209,14 @@ export function renderPhotoSubmissionAction(data) {
 
   console.log(contentfulId);
 
-  return null;
+  return (
+    <div
+      key={`photo-submission-action-${contentfulId}`}
+      className="margin-horizontal-md margin-bottom-lg"
+    >
+      <PhotoSubmissionActionContainer />
+    </div>
+  );
 }
 
 /**
--- a/resources/assets/components/ContentfulEntry/renderers.js
+++ b/resources/assets/components/ContentfulEntry/renderers.js
@@ -206,15 +206,11 @@ export function renderTextSubmissionAction(data) {
 }
 
 export function renderPhotoSubmissionAction(data) {
-  const contentfulId = data.id;
   const fields = setNullFieldsToUndefined(data.fields);
 
   return (
-    <div
-      key={`photo-submission-action-${contentfulId}`}
-      className="margin-horizontal-md margin-bottom-lg"
-    >
-      <PhotoSubmissionActionContainer id={contentfulId} {...fields} />
+    <div className="margin-horizontal-md margin-bottom-lg">
+      <PhotoSubmissionActionContainer id={data.id} {...fields} />
     </div>
   );
 }
--- a/resources/assets/components/ContentfulEntry/renderers.js
+++ b/resources/assets/components/ContentfulEntry/renderers.js
@@ -4,6 +4,7 @@ import { PuckWaypoint } from '@dosomething/puck-client';
 import Affirmation from '../Affirmation';
 import { LegacyContentBlock } from '../Block';
 import ContentBlock from '../blocks/ContentBlock/ContentBlock';
+import { setNullFieldsToUndefined } from '../../helpers/forms';
 import { ReportbackUploaderContainer } from '../ReportbackUploader';
 import { SubmissionGalleryContainer } from '../Gallery/SubmissionGallery';
 import LinkActionContainer from '../actions/LinkAction/LinkActionContainer';
@@ -194,7 +195,7 @@ export function renderTextSubmissionAction(data) {
         name="text_submission_action-top"
         waypointData={{ contentfulId }}
       />
-      <TextSubmissionActionContainer id={data.id} {...data.fields} />
+      <TextSubmissionActionContainer id={contentfulId} {...data.fields} />
       <SubmissionGalleryBlockContainer type="text" />
       <PuckWaypoint
         name="text_submission_action-bottom"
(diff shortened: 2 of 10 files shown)


package-lock.json
```diff
@@ -917,7 +917,7 @@
     },
     "babel-eslint": {
       "version": "8.2.2",
-      "resolved": "https://registry.npmjs.org/babel-eslint/-/babel-eslint-8.2.2.tgz",
+      "resolved": "http://registry.npmjs.org/babel-eslint/-/babel-eslint-8.2.2.tgz",
```
DFurnes: Why does this keep happening?! 🤔

resources/assets/components/ContentfulEntry/renderers.js
```diff
@@ -203,6 +205,20 @@ export function renderTextSubmissionAction(data) {
   );
 }
 
+export function renderPhotoSubmissionAction(data) {
+  const contentfulId = data.id;
+  const fields = setNullFieldsToUndefined(data.fields);
+
+  return (
+    <div
+      key={`photo-submission-action-${contentfulId}`}
```
DFurnes: This component only returns a single item, so you shouldn't need a `key` here.

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -0,0 +1,179 @@
...
+
+class PhotoSubmissionAction extends React.Component {
+  state = {
+    captionValue: '',
+    file: this.defaultMediaState,
+    quantityValue: '',
+    showModal: false,
+  };
+
+  componentDidUpdate = prevProps => {
```
DFurnes: We only need to use class properties for event handlers. Lifecycle methods don't have any binding issues.

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -0,0 +1,179 @@
...
+  resetForm = () => {
+    this.setState({
+      captionValue: '',
+      quantityValue: '',
+    });
+  };
+
+  render() {
+    return (
+      <React.Fragment>
```
DFurnes: This doesn't need to be wrapped in a `<React.Fragment>` since there's only one child here.

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -0,0 +1,179 @@
...
+    file: this.defaultMediaState,
+    quantityValue: '',
+    showModal: false,
+  };
+
+  componentDidUpdate = prevProps => {
+    console.log(prevProps);
+  };
+
+  defaultMediaState = {
```
DFurnes: Should we just inline this into `this.state`? It feels confusing to keep it in a separate property just to initialize `this.state`, since any changes made here (or in the state itself) later on would not be reflected here.

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -0,0 +1,179 @@
...
+                    <label
+                      className={classnames('field-label')}
+                      htmlFor="caption"
+                    >
+                      {this.props.captionFieldLabel}
+                    </label>
+                    <input
+                      className={classnames('text-field')}
+                      type="text"
+                      id="caption"
```
DFurnes: Does this need an `id`? (Thinking about what we've talked about with having more than one form on a page.)

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -0,0 +1,179 @@
...
+      captionValue: '',
+      quantityValue: '',
+    });
+  };
+
+  render() {
+    return (
+      <React.Fragment>
+        <Card
+          id={this.props.id}
```
DFurnes: What's the intended use case for this?

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionActionContainer.js
```diff
@@ -0,0 +1,22 @@
...
+  clearPostSubmissionItem,
+  storeCampaignPost,
+} from '../../../actions/post';
+
+const mapStateToProps = state => ({
+  campaignId: state.campaignId,
+  legacyCampaignId: state.campaign.legacyCampaignId,
+  legacyCampaignRunId: state.campaign.legacyCampaignRunId,
+  submissions: state.postSubmissions,
+  type: 'photo',
```
DFurnes: Can this just be assigned in `ContentfulEntry` as a normal prop?

resources/assets/helpers/forms.js
```diff
@@ -44,3 +44,19 @@ export function getFieldErrorMessages(response) {
...
 }
+
+/**
+ * Set null properites to undefined to help with React defaultProps.
+ *
+ * @param  {Object} fields
+ * @return {Object}
+ */
+export function setNullFieldsToUndefined(fields) {
+  const updatedFields = {};
```
DFurnes: I think this could be simplified with Lodash's [`omitBy`](https://lodash.com/docs/4.17.10#omitBy) and [`isNull`](https://lodash.com/docs/4.17.10#isNull):

```js
export function setNullFieldsToUndefined(fields) {
  return omitBy(fields, isNull);
}

```

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
weerd: It's not just for initializing state, but also for when the form is reset after submitting. Ran into this in the original uploader.

package-lock.json
weerd: 🤷‍♂️

resources/assets/components/ContentfulEntry/renderers.js
weerd: Most of the other rendered components in that file contain a `key` prop, so I was following the pattern. I can remove from the rest if desired.

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
weerd: Gotcha. I did this in the `TextSubmissionAction` so was following what I saw, since using that component as a helpful guide for this one.

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionActionContainer.js
weerd: Yeah probably. Again, was just following the pattern in `TextSubmissionAction`. Technically this can only ever be `photo` and in the `TextSubmissionAction` it can only ever be `text`, so maybe it's something that is hardcoded in the component, considering that if a different value is passed if it's a prop, that would cause to probably not work correctly.

resources/assets/helpers/forms.js
weerd: Oh yeah, I did that. Forgot about `omitBy`. Cool! 💯

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
weerd: It's required. Without it you don't have the accessibility to click on the label to focus on the associated input. The `name` attribute on the input doesn't cut it.

I want to address the multiple "same" forms on page in a future PR/ticket. It'll just mean appending the Contentful ID to these series of attributes.

resources/assets/components/ContentfulEntry/renderers.js
weerd: @DFurnes removed from the PSA renderer method, and will remove from others in a different PR so the number of changed files stays low 😬

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionActionContainer.js
weerd: Made `type` a class property :)

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
DFurnes: Oh, didn't know that's how that worked! Bummer that it can't just read from something scoped to the `form` itself. 😖

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
DFurnes: Would the IDs generated by the page in [ActionSteps](https://github.com/DoSomething/phoenix-next/blob/946e0dea35adf60c0cd903dc2cbf005678bb86da/resources/assets/components/pages/ActionPage/ActionSteps.js#L102), for example, serve the same purpose?

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
DFurnes: Ok, gotcha. Without a `@TODO` it's hard to know what's on your mind!

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -0,0 +1,183 @@
...
+    file: this.defaultMediaState,
+    quantityValue: '',
+    showModal: false,
+  };
+
+  componentDidUpdate(prevProps) {
+    console.log(prevProps);
+  }
+
+  type = 'photo';
```
DFurnes: Where is this being used (or going to be used)? If it's a "static" value, I'm not sure it belong as a class property.

DFurnes: Just looked at how `this.props.type` is used in TextSubmissionAction... I think I would just inline it or make a constant closer to where it's being used.

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
weerd: Yar. Sorry about that! Code edits were growing and was just trying to keep it small as I was duplicating the approaches from the Text Submission Action 😬

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
weerd: Potentially. That's going to change a bit given that those are going to all be variations on a content block in the future, "Action Steps" as they are as "distinct containers" aren't going to be a thing. Is there a particular concern here though? Maybe I don't really see the issue with adding an `id` to the component which otherwise seems innocuous to me 🤔 But I might be missing something...

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
weerd: Ah, ok. Yeah I'll go with the `const` approach in that `handleSumbmit` when I get to fleshing it out then. I'll revise the `TextSubmissionAction` to do the same. Prefer that over inlining since it will be used in multiple places in the `handleSubmit` func. 👍

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
DFurnes: It feels to me like something a generic page could handle as well, rather than having to update all of our components to accept an `id` prop. If you'd like more hands-on control, this works too though!

weerd: Aha! I was missing something. I hadn't realized that the `Card` component wasn't outputting an `id` attribute/accepting an `id` attribute. 🤦‍♂️ I fell victim again to stuff that was in the `TextSubmissionAction` as I was mimicking it's structure/behavior. Looks like I need a follow-up PR to the TSA to clean that one up a bit 😢

PR #879, opened 2018-05-02
Fix field name, add hide from nav

mendelB:
### What does this PR do?

This PR fixes field naming for `activity_feed`.
Adds `hideFromNavigation` to new community pages


weerd:
> Adds hideFromNavigation to new community pages

Where does it do this?

PR #881, opened 2018-05-02
Add migration to create subtitle field in Pages

mendelB:
### What does this PR do?

This PR adds a Contentful Migration script to create a `subtitle` field in the `Page` content type.

[#157014308](https://www.pivotaltracker.com/story/show/157014308)

### DIFF
--- a/contentful/migrations/2018_05_02_001_add_subtitle_to_page_content_type.js
+++ b/contentful/migrations/2018_05_02_001_add_subtitle_to_page_content_type.js
@@ -1,0 +1,12 @@
+module.exports = function(migration) {
+  const page = migration.editContentType('page');
+
+  page
+    .createField('subtitle')
+    .name('Subtitle')
+    .type('Symbol')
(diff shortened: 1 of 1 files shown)


contentful/migrations/2018_05_02_001_add_subtitle_to_page_content_type.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function(migration) {
+  const page = migration.editContentType('page');
+
+  page
+    .createField('subtitle')
```
weerd: `subTitle` to stay with how we case this field in other content-types!

mendelB: Yikes! I double checked the field in ContentBlock, but checked the title and not the id 😩

mendelB: Fixed!

PR #884, opened 2018-05-04
New profile page setup

katiecrane:
### What does this PR do?

Initial setup for Face! Looks like this so far:

image

Questions:
Does this page need to be gated given that we will definitely want to be able to test as non-admin users?

This is the container in which to build out the user profile. Next time we can actually add in information. 

[Card](https://www.pivotaltracker.com/story/show/156511955)

### Checklist

### DIFF
--- a/resources/assets/components/App.js
+++ b/resources/assets/components/App.js
@@ -11,6 +11,7 @@ import graphqlClient from '../graphql';
 import { getUserId } from '../selectors/user';
 import { initializeStore } from '../store/store';
 import CampaignContainer from './Campaign/CampaignContainer';
+import ProfilePage from './pages/ProfilePage/ProfilePage';
 
 const App = ({ store, history }) => {
   initializeStore(store);
@@ -26,7 +27,7 @@ const App = ({ store, history }) => {
         <ApolloProvider client={graphqlClient}>
           <ConnectedRouter history={history}>
             <Switch>
-              <Route path="/profile" render={() => <h1>PROFILE</h1>} />
+              <Route path="/profile" component={ProfilePage} />
               <Route path="/us/campaigns/:slug" component={CampaignContainer} />
             </Switch>
           </ConnectedRouter>
--- a/resources/assets/components/App.js
+++ b/resources/assets/components/App.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Provider } from 'react-redux';
-import { Route } from 'react-router-dom';
+import { Route, Switch } from 'react-router-dom';
 import { ApolloProvider } from 'react-apollo';
 import { ConnectedRouter } from 'react-router-redux';
 import { PuckProvider } from '@dosomething/puck-client';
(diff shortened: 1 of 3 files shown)


DFurnes:
> Does this page need to be gated given that we will definitely want to be able to test as non-admin users?

I'm leaning towards "no", because the traffic we get to `next.dosomething.org` should just be staffers and we won't be setting up the Fastly rule to expose this on `www.dosomething.org` until it's ready.

katiecrane:
@DFurnes Ahhhhh okay I didn't fully understand that yesterday, but given that I am also in favor of no gating.

weerd:
Yeah, I think the "gating" can come later as long as no sensitive info is being exposed or whatever, but like @DFurnes said, it'd just be staffers and their respective profile.

resources/assets/components/App.js
```diff
@@ -11,6 +11,7 @@ import graphqlClient from '../graphql';
 import { getUserId } from '../selectors/user';
 import { initializeStore } from '../store/store';
 import CampaignContainer from './Campaign/CampaignContainer';
+import ProfilePage from './pages/ProfilePage/ProfilePage';
```
weerd: short to long arrangement of imports :)

resources/assets/components/App.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Provider } from 'react-redux';
-import { Route } from 'react-router-dom';
+import { Route, Switch } from 'react-router-dom';
```
weerd: same same, short to long arrangement of imports, ala Laravel-style.

resources/assets/components/App.js
```diff
@@ -25,7 +26,10 @@ const App = ({ store, history }) => {
       >
         <ApolloProvider client={graphqlClient}>
           <ConnectedRouter history={history}>
-            <Route path="/us/campaigns/:slug" component={CampaignContainer} />
+            <Switch>
+              <Route path="/profile" component={ProfilePage} />
```
weerd: Adding the `/us/`? Maybe this is a for later thing..?

routes/web.php
```diff
@@ -16,6 +16,11 @@
 $router->get('next/logout', 'AuthController@getLogout')->name('logout');
 $router->redirect('auth/login', 'next/login'); // Fix for hard-coded redirect in Gateway! <goo.gl/2VPxDC>
 
+// Profile
+$router->get('/profile', function () {
```
weerd: Related to question above regarding adding the `/us/`? Does that matter for profiles?

resources/assets/components/App.js
katiecrane: @weerd ahhh I thought it would yell at me if I messed that up! good eye, thanks! definitely would have bugged me when I eventually noticed!

resources/assets/components/App.js
katiecrane: Great question, I really know nothing about this and what the reason to add it is. If we want it eventually I don't see why we wouldn't add now?

weerd: It's for #global, and if we eventually go back to providing our site for different locales that prefix will vary to help translate things, so like `/mx/profiles`. I think best to add it in so adding it in later, doesn't lead to any surprises!

resources/assets/components/App.js
DFurnes: 🦅 👀

routes/web.php
DFurnes: Yeah, I don't know what "translating" a user's profile means & I'm not a huge fan of the country-code prefix URLs, but I think it's a decision we live with now since Ashes enforces it on all the links it generates. 😬

weerd: Yep. Exactly. Not a fan of it either, but we've already committed to it 😬 

I was debating "what" in a profile would get translated, but there could definitely be some things like headings etc. 🤷‍♂️

PR #892, opened 2018-05-09
Community campaign subpage

mendelB:
### What does this PR do?

This PR adds support for routing to a  'Community' Page (`/community`) on a campaign (which has a Community Page but not an `activityFeed` field).

- Adding parsing for a `customBlock` content type to the contentful Entity `parseBlocks` method (to enable handling custom blocks which exist in many an `activityFeed` and will be ported to the community pages)
- Accounting for a Community page in the `page`s field when setting the `hasCommunityPage` boolean for `CampaignPage` routing
- Adding community page specific rendering in `campaignSubPage` to render the community page with blocks. (Will be implemented for pages in general down the line)
- Rendering the `campaignSubPage` from the `/community` route on campaigns which have a community page and no activity feed
- Adding the `hideFromNavigation` field to the `Page` Entity (temporary as we want to get rid of that logic in the future)

The setup is somewhat intensely specific to the weird world we have now with balancing community pages and `activityFeed`s. Once we sunset `activityFeed` entirely and move over to pages for everything, this will be able to be cleaned up majorly!

The `isCommunity` prop is to circumvent the general methodoligy the `campaignSubPage` uses to find the appropriate page. We also don't have access to the `match` prop (and anyways don't have a `:slug`) when on the `/community` route.

### DIFF
--- a/app/Entities/Page.php
+++ b/app/Entities/Page.php
@@ -31,7 +31,8 @@ class Page extends Entity implements JsonSerializable
      * @param  array $blocks
      * @return array
      */
-    public function parseBlocks($blocks) {
+    public function parseBlocks($blocks)
+    {
         $parsedBlocks = parent::parseBlocks($blocks);
 
         if (ends_with($this->slug, 'community')) {
--- a/app/Entities/Page.php
+++ b/app/Entities/Page.php
@@ -26,6 +26,22 @@ class Page extends Entity implements JsonSerializable
     }
 
     /**
+     * Parse blocks, and reverse parsed blocks for community pages.
+     *
+     * @param  array $blocks
+     * @return array
+     */
+    public function parseBlocks($blocks) {
+        $parsedBlocks = parent::parseBlocks($blocks);
+
+        if (ends_with($this->slug, 'community')) {
+            return $parsedBlocks->reverse()->values();
+        }
+
+        return $parsedBlocks;
+    }
+
+    /**
      * Convert the object into something JSON serializable.
      *
      * @return array
--- a/app/Entities/Page.php
+++ b/app/Entities/Page.php
@@ -42,6 +42,9 @@ class Page extends Entity implements JsonSerializable
                 'content' => $this->content,
                 'sidebar' => $this->parseSidebar($this->sidebar),
                 'blocks' => $this->parseBlocks($this->blocks),
+                // @TODO: we want to eventually remove the need for hideFromNavigation field
+                // in favor of always linking to pages referenced in the `pages` field.
+                'hideFromNavigation' => $this->hideFromNavigation,
             ],
         ];
     }
--- a/app/Entities/Page.php
+++ b/app/Entities/Page.php
@@ -35,7 +35,7 @@ class Page extends Entity implements JsonSerializable
     {
         $parsedBlocks = parent::parseBlocks($blocks);
 
-        if (ends_with($this->slug, 'community')) {
+        if (ends_with(rtrim($this->slug, '/'), 'community')) {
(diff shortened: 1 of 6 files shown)


app/Entities/Page.php
```diff
@@ -25,6 +25,23 @@ public function parseSidebar($sidebarItems)
...
+     * Parse blocks, and reverse parsed blocks for community pages.
+     *
+     * @param  array $blocks
+     * @return array
+     */
+    public function parseBlocks($blocks)
+    {
+        $parsedBlocks = parent::parseBlocks($blocks);
+
+        if (ends_with($this->slug, 'community')) {
```
weerd: Do we need to worry if someone accidentally enters a value for the slug as `some-campaign-name/community/`; ending slash?

mendelB: `ends_with` wouldn't handle that. Is this a scenario we need to account for? (I guess we could use regex instead?)

resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
```diff
@@ -15,14 +16,22 @@ import CampaignPageNavigationContainer from '../../CampaignPageNavigation/Campai
...
+    campaignEndDate,
+    isCommunity,
+    match,
+    noun,
+    pages,
+    tagline,
+    verb,
+  } = props;
+
+  const pageSlug = isCommunity ? 'community' : match.params.slug;
```
weerd: Slightly confused... why wouldn't `community` be part of the `match.params.slug`? 

For `faq` and `scholarship` pages it seems to work fine and find the page in the list. If we're here to render a community page, it's already established in the `CampaignPage` that a page for community exists in the pages list.

https://github.com/DoSomething/phoenix-next/pull/892/files#diff-f22cfc1593dfc65f4beca9bf1de156b6R72

weerd: Oh wait. Nevermind. Now I get it. It's not using `Route` to determine the component thus `match` is not available! Now I understand better what you meant in the description 👍

mendelB: I think the issue is that this is a statically declared route `.../community` as opposed to the others where we're routing it through `.../:slug`. 

In those cases the trailing slug is set as a param under `match.params.slug`. In the case of community -- until we remove the `/community` route entirely when we do away with `activity_feed`s, we need to deal with this scenario manually.  

Also, we'd need to find the community page entry anyways, so figure we'd just use the existing subPage `find` logic.

mendelB: Ah damn, you spoke! ...github 🙄

mendelB:
@weerd We do have several other places the check for `community` slug  through JS `endsWith` (or just compare slugs in general) without the `/` protection. Do we want to add that safeguard in those places as well?
([here](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js#L24), [here](https://github.com/DoSomething/phoenix-next/pull/892/files#diff-485aa9148b8caa1bc41166a564dfdde6R19) and [here](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/helpers/index.js#L558) is what I could find offhand)

weerd:
@mendelB I think so, yes. But separate PR and different approach since those other instances are all in JS. Maybe a similar Lodash method or just make a small helper function?

mendelB:
👍 Will do! I'll merge this one in then.

PR #893, opened 2018-05-09
Update ActivityFeed to Community Page transformation script

mendelB:
### What does this PR do?

This PR includes some updates to the Contentful Management API script for updating Campaigns `activity_feed`s to Community Pages:

- ignores any custom blocks of `type` `reportbacks`
- empties the `activity_feed` field of the transformed campaign
- reverse the block ordering for any campaign specified as `reverseActivityFeedOrder: false` in `additionalContent`. (so that we can blanket reverse all community pages, and not have to migrate this logic to pages)

With the support on the front end from #892, we'll be able to run this script and transition campaigns to the new path one by one.

### DIFF
--- a/contentful/management-api-scripts/2018_04_20_001_campaign_activity_feed_to_community_page.js
+++ b/contentful/management-api-scripts/2018_04_20_001_campaign_activity_feed_to_community_page.js
@@ -1,4 +1,5 @@
 const { join } = require('path');
+const { get } = require('lodash');
 const {
   attempt,
   constants,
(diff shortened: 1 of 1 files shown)


weerd:
> reverse the block ordering for any campaign specified as reverseActivityFeedOrder: false in additionalContent. (so that we can blanket reverse all community pages, and not have to migrate this logic to pages)

I might be confusing myself, but we want to reverse the activity feed for campaigns with `reverseActivityFeedOrder: true` right? Because that means they're using the old model where the campaign updates were arranged in the list manually to have the latest at the top... or am I confused?

mendelB:
lol yeah this took me a minute...

So we always auto reverse the activity feeds in PN *unless* `reverseActivityFeedOrder` is set to `false` which'd mean the feed is the usual contentful top to bottom order and editors are manually dragging new blocks to the top as opposed to just leaving them on the bottom and letting us reverse the feed.

So this would reverse those old feeds so that *all* community pages are bottom to top, and we'll always just reverse the feed on our end.

mendelB:
Regular `activity_feed`: 

3
2
1

we reverse to 

1
2
3

Old activity feeds:

1
2
3

`reverseActivityFeedOrder: false`

so we don't reverse.

this would flip those old ones around so again:

3
2
1

and we reverse

PR #894, opened 2018-05-09
Add a badge to unpublished posts.

DFurnes:
### What does this PR do?

This PR just adds a lil' badge to unpublished & rejected posts, so it's clear to staff members (who can see everything) or users (whose posts may be awaiting review) that it won't be visible in the public gallery.

screen shot 2018-05-09 at 1 17 02 pm

This was prompted by some confusion earlier today around how Phoenix Next galleries worked.

N/A

### Checklist

### DIFF
--- a/resources/assets/components/utilities/PostCard/PostBadge.js
+++ b/resources/assets/components/utilities/PostCard/PostBadge.js
@@ -1,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const PostBadge = ({ status, tags }) =>
+  status !== 'ACCEPTED' || tags.includes('hide-in-gallery') ? (
+    <div
+      className="post-badge"
+      title="This post is awaiting review or hidden in the gallery."
+    />
+  ) : null;
+
+PostBadge.propTypes = {
+  status: PropTypes.bool.isRequired,
+  tags: PropTypes.arrayOf(PropTypes.string),
+};
+
+PostBadge.defaultProps = {
+  tags: [],
+};
+
+export default PostBadge;
--- a/resources/assets/components/utilities/PostCard/PostBadge.js
+++ b/resources/assets/components/utilities/PostCard/PostBadge.js
@@ -1,13 +1,27 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const PostBadge = ({ status, tags }) =>
-  status !== 'ACCEPTED' || tags.includes('hide-in-gallery') ? (
-    <div
-      className="post-badge"
-      title="This post is awaiting review or hidden in the gallery."
-    />
-  ) : null;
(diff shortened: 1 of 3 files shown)


DFurnes:
This applies to all posts displayed in the GraphQL (Rogue v3) gallery, by the way. I just chose two text posts since they were the easiest thing to snag without personal data in them. 👁

resources/assets/components/utilities/PostCard/post.scss
```diff
@@ -60,3 +61,19 @@
     }
   }
 }
+
+.post-badge {
+  display: block;
+  position: absolute;
+  top: -15px;
+  left: -15px;
```
weerd: Not required for merge but a potential update, instead of using a square and offsetting to hide and cut it off w/ overflow to simulate a triangle, could use a CSS triangle made from borders.

resources/assets/components/utilities/PostCard/PostBadge.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const PostBadge = ({ status, tags }) => {
+  if (status === 'REJECTED') {
+    return (
+      <div
+        className="post-badge -rejected"
+        title="This post has been marked as rejected."
```
mendelB: Mozilla has an [interesting bit](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#Accessibility_concerns) about the accessibility concerns with using `title`. Is this something we need to worry about or is this just for internal labeling?

resources/assets/components/utilities/PostCard/post.scss
DFurnes: Oh yeah, nice! Done in f52661c.

resources/assets/components/utilities/PostCard/post.scss
```diff
@@ -65,15 +65,12 @@
...
-  top: -15px;
-  left: -15px;
-  width: 30px;
-  height: 30px;
-  background: $yellow;
-  color: #fff;
-  transform: rotate(-45deg);
+  top: 0;
+  left: 0;
+  border-top: $base-spacing solid $yellow;
```
weerd: We might want to specify `height: 0` and `width: 0` since the defaults for those are `auto` and may cause issues in some browsers!

resources/assets/components/utilities/PostCard/PostBadge.js
DFurnes: That's a good thought. It's primarily intended to reduce confusion for staffers, but would also display to users re: their own posts. Lemme play around a bit.

DFurnes:
Alright, I think this is clearer & also more accessible. An all-around win!

screen shot 2018-05-09 at 2 00 08 pm

PR #895, opened 2018-05-09
Quick fixes for console errors with Embed and LazyImage components.

weerd:
### What does this PR do?
This PR fixes some bugs with the `Embed` and `LazyImage` components that were causing a few warnings to show up in the browser console.

image 1

🌵 🤠

### DIFF
--- a/resources/assets/components/utilities/Embed/Embed.js
+++ b/resources/assets/components/utilities/Embed/Embed.js
@@ -46,7 +46,7 @@ class Embed extends React.Component {
         rel="noopener noreferrer"
       >
         <div className="embed">
-          <LazyImage className="embed__image" src={image} background />
+          <LazyImage className="embed__image" src={image} hasBackgroundImage />
           <div className="embed__content padded">
             <div className="margin-vertical-md margin-right-md">
               <h3 className="line-break">
@@ -56,16 +56,24 @@ class Embed extends React.Component {
                   <PlaceholderText size="medium" />
                 )}
               </h3>
-              <p className="color-gray">
+              <div>
                 {data ? (
-                  truncate(data.description, { length: 240 })
+                  <p className="color-gray">
+                    {truncate(data.description, { length: 240 })}
+                  </p>
                 ) : (
                   <PlaceholderText size="large" />
                 )}
-              </p>
-              <p className="footnote font-bold caps-lock">
-                {data ? data.provider.name : <PlaceholderText size="small" />}
-              </p>
+              </div>
+              <div className="footnote">
+                {data ? (
+                  <p className="color-gray font-bold caps-lock">
+                    {data.provider.name}
+                  </p>
+                ) : (
+                  <PlaceholderText size="small" />
+                )}
+              </div>
             </div>
           </div>
           {badged ? (
--- a/resources/assets/components/utilities/Embed/Embed.js
+++ b/resources/assets/components/utilities/Embed/Embed.js
@@ -46,7 +46,7 @@ class Embed extends React.Component {
         rel="noopener noreferrer"
       >
         <div className="embed">
-          <LazyImage className="embed__image" src={image} hasBackgroundImage />
+          <LazyImage className="embed__image" src={image} asBackground />
           <div className="embed__content padded">
             <div className="margin-vertical-md margin-right-md">
               <h3 className="line-break">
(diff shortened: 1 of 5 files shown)


resources/assets/components/utilities/LazyImage.js
```diff
@@ -67,16 +67,16 @@ class LazyImage extends React.Component {
 }
 
 LazyImage.propTypes = {
-  alt: requiredIf(PropTypes.string, props => !props.background),
-  background: PropTypes.bool,
+  alt: requiredIf(PropTypes.string, props => !props.hasBackgroundImage),
   className: PropTypes.string,
+  hasBackgroundImage: PropTypes.bool,
```
weerd: Renamed this props since it was confusing, and now follows more inline with how we name boolean properties.

resources/assets/components/utilities/LazyImage.js
```diff
@@ -54,7 +54,7 @@ class LazyImage extends React.Component {
 
     return (
       <img
-        {...this.props}
+        className={this.props.className}
```
weerd: This was causing an issue where it was adding the `hasBackgroundImage` (previously `background`) as an attribute to the `` which React was getting angry about. There aren't any other props we need, other than `className` that would actually apply, so the spread operator isn't really needed here.

resources/assets/components/utilities/LazyImage.js
DFurnes: I don't feel this is clearer. This prop instructs `LazyImage` to display the given `src` as a background image, not a description of something it "has".

weerd: Oh fair. Suggestions on a rename. Just `background` is definitely not clear and almost looks like we're adding an old-school background attribute to an element. `asBackgroundImage`?

resources/assets/components/utilities/Embed/Embed.js
```diff
@@ -56,16 +56,24 @@ class Embed extends React.Component {
...
-                  truncate(data.description, { length: 240 })
+                  <p className="color-gray">
+                    {truncate(data.description, { length: 240 })}
+                  </p>
                 ) : (
                   <PlaceholderText size="large" />
                 )}
-              </p>
-              <p className="footnote font-bold caps-lock">
-                {data ? data.provider.name : <PlaceholderText size="small" />}
```
DFurnes: Want to make the element in `PlaceholderText` a `span` instead so it can be nested? The JSX markup looks a bit cleaner, and let's it inherit font size from the text it's replacing.

resources/assets/components/utilities/Embed/embed.scss
```diff
@@ -7,6 +7,10 @@
   @include media($mobile) {
     flex-direction: column;
   }
+
+  .footnote {
+    margin-top: 36px;
```
DFurnes: What's this for? Should we use `$base-spacing`?

resources/assets/components/utilities/LazyImage.js
DFurnes: I don't mind `background`, but `asBackground` or `backgrounded` could be fine alternatives.

resources/assets/components/utilities/Embed/Embed.js
weerd: Yeah, that works for me! Wasn't sure if that was an ok change and if needed to have a certain display quality for the placeholder text stuff to render accordingly.

resources/assets/components/utilities/Embed/embed.scss
weerd: The change in structure (now not having `p + p`) changed the styles so the footnote was up next to the description above. Did `36px` cause that's what it was before and `$base-spacing` is only `24px`. But if switching to using `span` for the placeholder, this becomes a mute point and can be removed :)

resources/assets/components/utilities/Embed/Embed.js
DFurnes: Should be safe to change since the element is set to display as `inline-block`.

weerd: Ha, I didn't investigate that deeply. Just saw the warning about no `div` in `p` 😏

resources/assets/components/utilities/Embed/embed.scss
DFurnes: I feel like I _have_ to call you out on [mute point](http://www.dictionary.com/e/moot-point-vs-mute-point/). Even the grammar police aren't above the law! 👮

resources/assets/components/utilities/Embed/Embed.js
DFurnes: Yeah sorry, I clearly should have tested this a bit more! 😬

resources/assets/components/utilities/Embed/embed.scss
weerd: hahaha, well done my young Padawan 👍

PR #899, opened 2018-05-10
Performing loader functionality on mount as well as on update

mendelB:
Update to set the loader on the LazyImage component both after mounting *and* updating props. (Quick bug fix to address RBs never rendering their images.)


DFurnes:
Just a quick linty-lint fix!

PR #902, opened 2018-05-10
Standalone 'Page' backend setup

mendelB:
### What does this PR do?

This PR adds the initial setup for our standalone Pages (static pages, article pages) on the backend.

- Adds a `PageRepository` with a findBySlug method for fetching the proper page entry from Contentful
- Adds a `PageController` with a `show` method still stubbed to the `501` but fetching the page using the `PageRepository` (so if the entry is not found we'd return a `404` #progress).
- Routing `us/{slug}` to the `PageController@show`

This is the first half of implementation for standalone Page support.

Q: Should we add caching to the PageRepository@findBySlug à la the [CampaignRepository](https://github.com/DoSomething/phoenix-next/blob/page-backend/app/Repositories/CampaignRepository.php#L128) or is that overkill for this?

[#157006215](https://www.pivotaltracker.com/story/show/157006215)

### DIFF
--- a/app/Repositories/PageRepository.php
+++ b/app/Repositories/PageRepository.php
@@ -43,8 +43,8 @@ class PageRepository
 
         if (! $pages->count()) {
             throw new ModelNotFoundException;
-        } else {
-            return new Page($pages[0]);
         }
+
+        return new Page($pages[0]);
     }
 }
--- a/app/Repositories/PageRepository.php
+++ b/app/Repositories/PageRepository.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace App\Repositories;
+
+use App\Entities\Page;
+use Contentful\Delivery\Query;
+use Contentful\Delivery\Client as Contentful;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+
+class PageRepository
+{
+    /**
+     * Contentful client instance.
+     */
+    private $contentful;
+
+    /**
+     * PageRepository constructor.
+     *
+     * @param Contentful $contentful
+     */
+    public function __construct(Contentful $contentful)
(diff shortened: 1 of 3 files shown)


app/Repositories/PageRepository.php
```diff
@@ -0,0 +1,50 @@
...
+     */
+    public function findBySlug($slug)
+    {
+        $query = (new Query)
+            ->setContentType('page')
+            ->where('fields.slug', $slug)
+            ->setInclude(1)
+            ->setLimit(1);
+
+        $pages = $this->contentful->getEntries($query);
```
DFurnes: Not necessary for this PR, but I imagine we'll eventually want to cache this request.

app/Repositories/PageRepository.php
```diff
@@ -0,0 +1,50 @@
...
+
+        if (! $pages->count()) {
```
DFurnes: This might read a little nicer as a guard:

```php
if (! $pages->count()) {
}

return new Page($pages[0]);

```

weerd:
> Q: Should we add caching to the PageRepository@findBySlug à la the CampaignRepository or is that overkill for this?

No caching for now. I want to re-assess how we do caching and refactor it a bit so it's more universal than how it's currently setup; being specific to campaigns.

app/Repositories/PageRepository.php
DFurnes: Oh whoops, I see now this was a question in the PR. 😅

weerd: Defs in the future, but I'd like to refactor how we do caching to make a solution that works well for both campaigns and static pages.

PR #905, opened 2018-05-10
Community migration updates

mendelB:
2 quick last minute additions for the `activity_feed` --> community page migration:

- Adding the newly generated Community Page to the *top* of the Campaign's Pages field so that the 'community' tab is at the beginning of the navigation (or at least directly following the 'action' tab).
- Disabling title link and sharing for Campaign Updates, since moving the blocks from the campaigns activity feed hides them from our Contentful Entry finding method (which means the links don't work anymore).

### DIFF
--- a/contentful/management-api-scripts/2018_04_20_001_campaign_activity_feed_to_community_page.js
+++ b/contentful/management-api-scripts/2018_04_20_001_campaign_activity_feed_to_community_page.js
@@ -90,7 +90,7 @@ async function addCommunityPageFromActivityFeed(environment, campaign) {
       logger.info(`  - Created Community Page! [ID: ${id}]\n`);
 
       // Add a Link to the new Page to the campaigns Pages field
-      campaign.fields.pages[LOCALE].push(linkReference(id));
+      campaign.fields.pages[LOCALE].unshift(linkReference(id));
       // Remove activity_feed blocks from campaign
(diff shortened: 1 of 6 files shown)


resources/assets/components/CampaignUpdate/CampaignUpdate.js
```diff
@@ -63,17 +66,24 @@ const CampaignUpdate = props => {
...
+            variant="icon"
+            parentSource="campaignUpdate"
+            className="clear-none -right-icon"
+          /> */}
       </footer>
     </Card>
   );
 };
 
+// *1:
```
DFurnes: This is really clean! ☯️

contentful/management-api-scripts/2018_04_20_001_campaign_activity_feed_to_community_page.js
```diff
@@ -90,7 +90,7 @@ async function addCommunityPageFromActivityFeed(environment, campaign) {
       logger.info(`  - Created Community Page! [ID: ${id}]\n`);
 
       // Add a Link to the new Page to the campaigns Pages field
-      campaign.fields.pages[LOCALE].push(linkReference(id));
+      campaign.fields.pages[LOCALE].unshift(linkReference(id));
```
weerd: 👌

PR #908, opened 2018-05-11
Photo submission action tweaks and bug fixes

weerd:
### What does this PR do?
This PR fixes a bug with the validation errors after switching to using a `fields` property to map errors against. This led to any field items without errors to still exist in the `errors` object, but with value of `undefined`. Visually this caused _all_ fields to have the `has-errors` class applied when checking against the `errors` object and seeing if it `has()` a particular property!

It also adds some useful logging to the `store()` method on the `CampaignPostsController`, and makes sure to remove user details that are not necessary for logging.

Refs #562


resources/assets/helpers/index.js
```diff
@@ -600,11 +600,22 @@ export function report(error) {
 }
 
 /**
- * Set null properites to undefined to help with React defaultProps.
+ * Set null properites to undefined.
```
mendelB: (Not really relevant to this PR but just out of curiosity:) Does this actually set the properties to `undefined` or just removes the fields from the returned object?

weerd: Ah! Good point!

PR #914, opened 2018-05-14
Ensure page is fully loaded before starting app.

DFurnes:
### What does this PR do?

This pull request fixes up a bug I introduced in #910. Since we now kick off network requests for scripts in the header, there's a chance that they'll load before the rest of the DOM has been parsed and run into issues (for example, not being able to find `window.ENV`). To avoid that, I've delayed our "initial render" until the DOM is ready and delayed an `env()` call that would happen before initial render.

We didn't have to do this before because the scripts were located at the very end of the body, after our `window.STATE`, `window.AUTH` and `window.ENV` scripts.

I'll write one up!

### Checklist

### DIFF
--- a/resources/assets/components/App.js
+++ b/resources/assets/components/App.js
@@ -25,7 +25,7 @@ const App = ({ store, history }) => {
         history={history}
         puckUrl={env('PUCK_URL')}
       >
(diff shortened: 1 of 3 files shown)


weerd:
Post-merge 👍

PR #916, opened 2018-05-14
Link Action - CTA template

mendelB:
### What does this PR do?

This PR adds a CTA style template for the Link Action component.

This is for Article / Standalone Pages.
I Copied and altered some of the styling and structure over from the `CallToAction` component.

[#157015532](https://www.pivotaltracker.com/story/show/157015532)

### Checklist

**LARGE**
image

**MEDIUM**
image

**SMALL**
image


lkpttn:
Looks good!

resources/assets/components/actions/LinkAction/templates/CTATemplate.js
```diff
@@ -0,0 +1,46 @@
...
+
+import './cta-template.scss';
+
+const onLinkClick = link => {
+  window.open(link, isExternal(link) ? '_blank' : '_self');
+  trackPuckEvent('clicked link action', { link });
+};
+
+const CallToAction = props => (
+  <Card className="cta-template rounded padded text-centered bg-black dark caps-lock">
```
DFurnes: I wonder if there's a shared presentational component that can be extracted here to share this layout & styling between both blocks that use it? If so, not sure what to call it hehe… 😆

resources/assets/components/actions/LinkAction/templates/CTATemplate.js
```diff
@@ -0,0 +1,46 @@
...
+    <div className="cta-template__title margin-top-lg">{props.title}</div>
+
+    <div className="cta-template__content margin-bottom-md">
```
DFurnes: I wonder if each of these `div` elements should have `.padded`, instead of (or in addition to?) the outside container, so we don't have to worry about which side to apply the margin to on each one.

mendelB: Here's what that would look like: (The title would just be farther up on the card)
**padded**
image
**margin**
image

resources/assets/components/actions/LinkAction/templates/CTATemplate.js
mendelB: Yelp. I should have named this `CTATemplate`! Also, the content and styling are somewhat different here then the `CallToAction` which has multiple themes and all sorts of CTA specific fields, so I wasn't sure how to combine the two properly.

resources/assets/components/actions/LinkAction/LinkAction.js
```diff
@@ -1,12 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import CTATemplate from './templates/CTATemplate';
```
weerd: Let's keep our studly-case standard `CtaTemplate`. While I know it might look weird with `Cta` not being all in caps, it'll cause less confusion if we follow the standard naming rule!

mendelB: Ooh this actually annoyed me like hell, and I didn't realize we were meant to do it that way! Sweet!

resources/assets/components/actions/LinkAction/templates/CTATemplate.js
```diff
@@ -0,0 +1,46 @@
...
+import './cta-template.scss';
+
+const onLinkClick = link => {
+  window.open(link, isExternal(link) ? '_blank' : '_self');
+  trackPuckEvent('clicked link action', { link });
+};
+
+const CTATemplate = props => (
+  <Card className="cta-template rounded padded text-centered bg-black dark caps-lock">
+    <div className="cta-template__title margin-top-lg">{props.title}</div>
```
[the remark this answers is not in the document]
mendelB: Is an `h3` appropriate here?

resources/assets/components/actions/LinkAction/templates/CTATemplate.js
DFurnes: Hmm, I also wonder if we used `h3` + `p` + `p` tags (instead of all `div`s), if we'd get the right spacing for free. Worth a quick try, but otherwise I'm good leaving this as-is.

resources/assets/components/actions/LinkAction/templates/CTATemplate.js
```diff
@@ -0,0 +1,44 @@
...
+
+CtaTemplate.propTypes = {
+  title: PropTypes.string.isRequired,
+  content: PropTypes.string.isRequired,
+  link: PropTypes.string.isRequired,
+  buttonText: PropTypes.string,
+};
+
+CtaTemplate.defaultProps = {
+  buttonText: "Let's do this!",
```
mendelB: Hmm.. should the `buttonText` be required here? I initially made it optional since it's optional on the Link Action as it stands now, but come to think of it, content is also optional and it's required here. Also not sure about the `Let's do this!` default text...

resources/assets/components/actions/LinkAction/templates/CtaTemplate.js
```diff
@@ -0,0 +1,44 @@
...
+const onLinkClick = link => {
+  window.open(link, isExternal(link) ? '_blank' : '_self');
+  trackPuckEvent('clicked link action', { link });
+};
+
+const CtaTemplate = props => (
+  <Card className="cta-template rounded padded text-centered bg-black dark caps-lock">
+    <h3 className="cta-template__title margin-top-lg">{props.title}</h3>
+
+    <p className="cta-template__content color-white">{props.content}</p>
```
weerd: Oh, one reason I didn't suggest to change this to a `p` is because the content is Markdown right? Shouldn't we use the markdown component?

mendelB: Hmm, the field in contentful is a markdown editor since it's markdown in the default link action, but I didn't think this one was meant to be markdown?

weerd: It's probably not, but it brings about a complication if it's set as a Markdown field, you could very well end up with hashtags and other markdown characters outputting in a paragraph tag. 🤔 We might just want to wrap this and output in Markdown, and if it looks weird, its up to the editor to adjust it.

mendelB: Ok, that sounds like a fair compromise!

mendelB: Added! Though this did necessitate narrowing the `cta-template__content` scss rules to the `p` inside it to effect the rendered Markdown element. Is that OK?

PR #918, opened 2018-05-15
Add migration for new campaign fields

mendelB:
### What does this PR do?

This PR adds a Contentful CLI Migration script which adds new fields to the Campaign Content Type

- Staff Pick
- Cause (checklist of specified causes)
- Scholarship Amount

A cool thing is we can [change the editor interface for a field directly from the migration](https://github.com/contentful/contentful-migration#changeeditorinterface-fieldid-widgetid-settings--void)! (We do it here for cause. I stopped short of adding helper text - which we can also do). You just need to update your contentful-cli to be able to run this.

I set Scholarship Amount as an integer, please LMK if it should be a Symbol instead. (I ask bc it was specified in the card as `number?`)

[#157006958](https://www.pivotaltracker.com/story/show/157006958)

image

### DIFF
--- a/contentful/migrations/2018_05_15_001_add_fields_to_campaign.js
+++ b/contentful/migrations/2018_05_15_001_add_fields_to_campaign.js
@@ -1,0 +1,49 @@
+module.exports = function(migration) {
+  const campaign = migration.editContentType('campaign');
+
+  campaign
+    .createField('staffPick')
+    .name('Staff Pick')
+    .type('Boolean')
+    .required(false)
+    .localized(true);
+
+  campaign
+    .createField('cause')
+    .name('Cause')
+    .type('Array')
+    .items({
+      type: 'Symbol',
+      validations: [
+        {
+          in: [
+            'Animals',
+            'Bullying',
+            'Disasters',
+            'Discrimination',
+            'Education',
+            'Environment',
+            'Homelessness',
+            'Mental Health',
+            'Physical Health',
+            'Poverty',
+            'Relationships',
+            'Sex',
+            'Violence',
+          ],
+        },
+      ],
+    })
+    .required(false)
+    .localized(true);
+
+  campaign.changeEditorInterface('cause', 'checkbox');
+
+  campaign
+    .createField('scholarshipAmount')
+    .name('Scholarship Amount')
+    .type('Integer')
+    .required(false);
+
+  campaign.moveField('additionalContent').afterField('scholarshipAmount');
+};


contentful/migrations/2018_05_15_001_add_fields_to_campaign.js
```diff
@@ -0,0 +1,49 @@
...
+
+  campaign
+    .createField('staffPick')
+    .name('Staff Pick')
+    .type('Boolean')
+    .required(false)
+    .localized(true);
+
+  campaign
+    .createField('cause')
```
mendelB: Should this field be called `causes`? (That's what ashes calls it. It also seems to have a single 'primary' and multiple 'secondary' causes, is that something we care about here, or are we fine with a simple array?)

weerd: I think maybe go with `causes` since it can be one or many. I might run the `secondary` causes stuff with @ashleybaldwin and @lkpttn to get thoughts on that as well.

contentful/migrations/2018_05_15_001_add_fields_to_campaign.js
```diff
@@ -0,0 +1,49 @@
...
+            'Discrimination',
+            'Education',
+            'Environment',
+            'Homelessness',
+            'Mental Health',
+            'Physical Health',
+            'Poverty',
+            'Relationships',
+            'Sex',
+            'Violence',
```
weerd: Might be best to lowercase all of these.

contentful/migrations/2018_05_15_001_add_fields_to_campaign.js
```diff
@@ -0,0 +1,49 @@
...
+    })
+    .required(false)
+    .localized(true);
+
+  campaign.changeEditorInterface('cause', 'checkbox');
+
+  campaign
+    .createField('scholarshipAmount')
+    .name('Scholarship Amount')
+    .type('Integer')
```
weerd: I could maybe see arguments for a string vs an integer, but need to think about this for a hot sec 🤔

contentful/migrations/2018_05_15_001_add_fields_to_campaign.js
mendelB: Good call! Another (not at all that helpful but still fun) contentful form widget might be Upcasing checklist or array symbol values in the editor for presentation!

contentful/migrations/2018_05_15_001_add_fields_to_campaign.js
```diff
@@ -0,0 +1,49 @@
+module.exports = function(migration) {
+  const campaign = migration.editContentType('campaign');
+
+  campaign
+    .createField('staffPick')
+    .name('Staff Pick')
+    .type('Boolean')
+    .required(false)
+    .localized(true);
```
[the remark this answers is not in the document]
weerd: Not sure. I suspect so, but in all my migrations I've been specifying it for every field regardless of whether `true` or `false`... 🤷‍♂️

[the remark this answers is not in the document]
mendelB: Fair enough, that's what I did in the end!

contentful/migrations/2018_05_15_001_add_fields_to_campaign.js
```diff
@@ -0,0 +1,50 @@
...
+
+  campaign
+    .createField('staffPick')
+    .name('Staff Pick')
+    .type('Boolean')
+    .required(false)
+    .localized(false);
+
+  campaign
+    .createField('causes')
```
mendelB: Based on what [@DFurnes said](https://dosomething.slack.com/archives/C2BPA7M8F/p1526404453000061) it seems as though this should be a single `cause` field as opposed to a list of `causes`? Since Ashes is returning a single primary cause, and Rogue seems to be using that one cause to [group campaigns](https://github.com/DoSomething/rogue/blob/master/app/Services/CampaignService.php#L162).

weerd: Yeah, that makes sense then. Switch it to singular.

contentful/migrations/2018_05_15_001_add_fields_to_campaign.js
weerd: @DFurnes thoughts on this. I think integer should be fine.

DFurnes: Yeah, I like using an integer here – could avoid confusion over whether or not to include `$` or commas in the number, and I don't think we'll likely want non-numeric scholarship amounts.

PR #919, opened 2018-05-15
Make quantity field optional

weerd:
### What does this PR do?
This PR updates the `PhotoSubmissionAction` to allow campaign leads to decide whether to show or hide the `quantity` field on the form. It also makes sure to validate the `quantity` field _if_ it is set to show up on the form using the `showQuantityField` option in the Contentful PSA content type.

**With Quantity Field**
kapture 2018-05-15 at 12 15 10

**Without Quantity Field**
kapture 2018-05-15 at 12 16 58

I initially approached this by adding a bunch of logic to the `PostRequest` to deal with only validating the `quantity` if the `show_quantity` was set to `true`. This logic needed to be in place because I was still submitting a value for the `quantity` via the front-end `FormData` submission, and thus even though not required it would still validate against the `int` and `min:1` rules for the `quantity` since it was still present in the form data. 

Upon that realization, it made it more obvious that this could be significantly simplified by just not including a `quantity` property in the `FormData` if the `showQuantityField` was `false`. This turned out to be the better, and simpler approach via adding a little bit of logic in the `PhotoSubmissionAction` to not include the `quantity` property.

Refs #562 

### Checklist

### DIFF
--- a/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
+++ b/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
@@ -54,11 +54,18 @@ class PhotoSubmissionAction extends React.Component {
     }
   }
 
-  fields = {
-    file: 'media',
-    text: 'caption',
-    quantity: 'quantity',
-    why_participated: 'whyParticipated',
+  fields = () => {
+    const items = {
+      file: 'media',
+      text: 'caption',
+      why_participated: 'whyParticipated',
+    };
+
+    if (this.props.showQuantityField) {
+      items.quantity = 'quantity';
+    }
+
+    return items;
   };
 
   handleChange = event => {
@@ -88,8 +95,9 @@ class PhotoSubmissionAction extends React.Component {
         type,
         id: this.props.id,
         // Associate state values to fields.
-        ...mapValues(this.fields, value => this.state[`${value}Value`]),
+        ...mapValues(this.fields(), value => this.state[`${value}Value`]),
         file: this.state.mediaValue.file || '',
+        show_quantity: this.props.showQuantityField ? 1 : 0,
       },
       this.props,
     );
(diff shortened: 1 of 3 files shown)


resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -88,8 +95,9 @@ class PhotoSubmissionAction extends React.Component {
         type,
         id: this.props.id,
         // Associate state values to fields.
-        ...mapValues(this.fields, value => this.state[`${value}Value`]),
+        ...mapValues(this.fields(), value => this.state[`${value}Value`]),
         file: this.state.mediaValue.file || '',
+        show_quantity: this.props.showQuantityField ? 1 : 0,
```
weerd: One thing to note here... initially this wasn't a ternary and just the `this.props.showQuantityField` since it's a boolean and worked just fine with the validation on the `PostRequest`.

However, I preferred to play it safe and in the validation for the `show_quantity` field, apart from being `required`, I wanted to ensure it's a `boolean`. 
https://github.com/DoSomething/phoenix-next/pull/919/files#diff-57052eff73cbefde96a02bdd41513dccR61

When doing that the PSA validation broke because the `boolean` validation will not register the string `"true"` (how it gets submitted via the form data request) as a [boolean according to the Laravel validation docs](https://laravel.com/docs/5.6/validation#rule-boolean). Thus the ternary to ensure it's a boolean value that can be validated appropriately on the backend.

mendelB: Iiinteresting, Is the `show_quantity` field a rogue requirement, or is that just for us to display in the form or not?

BTW a cool thing [I bumped into](https://stackoverflow.com/a/7820695/5657278) that you could use here instead of a ternary is to use the `+` operator to auto convert the value:
```js
show_quantity: + this.props.showQuantityField,
```

weerd: I dig that, but worried it'd cause more confusion and require some thinking to realize what's going on. I think if the stuff going on was more straightforward I'd be down with changing it, but there's already a bunch of conditional logic happening in that file, and I think I'd rather keep it more straightforward. Still cool suggestion 💃 

Alternatively, I _could_ use that and just add a quick comment next to it 🤔 Thoughts?

mendelB: I feel ya. Clarity > Brevity

PR #920, opened 2018-05-15
Add additional tokens to Link Action and Share Action.

DFurnes:
### What does this PR do?

This PR adds a few extra dynamic tokens to the Link Action and Share Action:

 - `{northstarId}` - this is an alias for `{userId}`, which we should use instead.
 - `{campaignRunId}` - the legacy campaign run ID for the current campaign.
 - `{source}` - like the voter registration action, this is hard-coded to `web`.

🗼 

[#157515339](https://www.pivotaltracker.com/story/show/157515339)

### Checklist


DFurnes:
One more fix for an issue I just thought of – if we're showing a preview of a tokenized URL, we'd never cache that data (since each URL would be unique to the user & campaign). I've updated the Embed component to strip any tokens from the given URL when generating the preview to avoid this.

weerd:
Nice! good call @DFurnes 👍

PR #921, opened 2018-05-16
Master template cleanup

weerd:
### What does this PR do?
This PR simply cleans up the script loading logic in the **master.blade.php** file by extracting it into a partial, the template is more straightforward.

🌵


mendelB:
BTW there's no way moving this to a partial would affect the load order right?

weerd:
Don't think so since Laravel needs to process the template and then caches the views it generates and calls them.

PR #922, opened 2018-05-16
Add new fields to Campaign Entity

mendelB:
### What does this PR do?

This PR adds a couple of new fields to the `Campaign` Entity based on some additional fields we've added to the Contentful Content Type

[#157010624](https://www.pivotaltracker.com/story/show/157010624)

### Checklist


app/Entities/Campaign.php
```diff
@@ -205,6 +205,9 @@ public function jsonSerialize()
             'additionalContent' => $this->additionalContent,
             'allowExperiments' => $this->campaignSettings ? $this->campaignSettings->allowExperiments : null,
             'actionText' => array_get($this->campaignSettings, 'actionText') ?: 'Join Us',
+            'staffPick' => $this->staffPick,
+            'cause' => $this->cause,
```
weerd: Just curious, does this return an array with a single item or a single string?

mendelB: Just a single string here.

PR #923, opened 2018-05-16
Voter registration nav button

weerd:
### What does this PR do?
This PR adds a "Register to Vote" link in the site wide top navigation so authenticated members can click on a link that takes them to dosomething.vote! This only shows up on large screen, and otherwise hidden to anything below that.

image

Refs [Pivotal ID #156275921](https://www.pivotaltracker.com/story/show/156275921)

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4906,7 +4906,7 @@
     },
     "event-stream": {
       "version": "3.3.4",
-      "resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
+      "resolved": "http://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
       "integrity": "sha1-SrTJoPWlTbkzi0w02Gv86PSzVXE=",
       "dev": true,
       "requires": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -4906,7 +4906,7 @@
     },
     "event-stream": {
       "version": "3.3.4",
-      "resolved": "http://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
+      "resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -4906,7 +4906,7 @@
     },
     "event-stream": {
       "version": "3.3.4",
-      "resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
+      "resolved": "http://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
```
DFurnes: (╯°□°)╯︵  dʇʇɥ

weerd: damn it!

resources/views/partials/navigation.blade.php
```diff
@@ -16,18 +16,25 @@
...
-                    </ul>
-                @else
-                    <a href="{{ route('login', isset($campaign) ? get_login_query($campaign) : null) }}">Log In</a>
-                @endif
-            </li>
-        </ul>
+
+        <div class="wrapper">
+            @if (Auth::user())
+                <a href="https://www.dosomething.vote/?r=campaignID:8017,campaignRunID:8022,source:web,source_details:nav" class="button" target="_blank">Register to Vote</a>
```
mendelB: btw is there a reason we don't throw the `userId` into this link?

weerd: Ashley mentioned it's just a hardcoded link for now and not a referral.

DFurnes: Hmm, I think we do want `userId` in here... otherwise Chompy has to do extra work to try to match up a user's form entries with the correct Northstar ID. @ashleybaldwin Can you help clarify what we want here?

weerd: I can defs add it and should be a quick update!

PR #924, opened 2018-05-17
implement basic index action

mendelB:
### What does this PR do?

This PR adds basic implementation for the `CampaignsController#index` for the `/api/v2/campaigns` endpoint.
(also includes a tiny unrelated doc link fix)

All this does as of now is return the value of the `CampaignRepository`'s `getAll` method which queries for all Contentful campaigns.

I believe though that we do need to return Ashes campaigns as well from this endpoint to fully backfill support for the ashes campaigns index. [More context in this thread](https://dosomething.slack.com/archives/C3ASB4204/p1526496169000595)

Also, this doesn't include params for filtering of any kind as of yet. Hopefully will be able to utilize [some existing code](https://dosomething.slack.com/archives/C3ASB4204/p1526481348000338) from Northstar to implement that next.
[#157410894](https://www.pivotaltracker.com/story/show/157410894)

### Checklist

### DIFF
--- a/app/Http/Controllers/Api/CampaignsController.php
+++ b/app/Http/Controllers/Api/CampaignsController.php
@@ -31,7 +31,9 @@ class CampaignsController extends Controller
      */
     public function index()
     {
-        abort(501, 'Currently not implemented.');
+        $data = $this->campaignRepository->getAll();
+
+        return response()->json(['data' => $data]);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Api/CampaignsController.php
```diff
@@ -31,7 +31,9 @@ public function __construct(CampaignRepository $campaignRepository)
      */
     public function index()
     {
-        abort(501, 'Currently not implemented.');
+        $data = $this->campaignRepository->getAll();
+
+        return response()->json(['data' => $data]);
```
weerd: I think this should be ok for now, but as we move along with this, I'd like for us to re-assess where the caching for contentful queries happens, where the flattening happens, etc so we have obvious answers for where this stuff happens between controller & repository classes :)

mendelB: 👌

PR #926, opened 2018-05-17
Hides reactions UI in ungrated gallery when not authenticated.

weerd:
### What does this PR do?
This PR updates the `PostCard` to only show reactions on a post if the user is authenticated and can actually react on the item.

Refs [Pivotal ID #155866183](https://www.pivotaltracker.com/story/show/155866183)


weerd:
FYI, forgot to mention that I got rid of the `ensureAuth()` helper function because it wasn't used anywhere in favor of the simpler `isAuthenticated()` :)

PR #927, opened 2018-05-17
Bunch o' style fixes

weerd:
### What does this PR do?
This PR fixes up a bunch of small issues with styles:
- Fixes the nav so the dropdown does not get cut-off by the body content of the page, causing only a sliver of the "Log out" link to be able to be clicked.
- Fixes the nav so when the dropdown shows, it doesn't push the rest of the content page down.
- Universally changes the font on buttons to "Source Sans Pro" instead of "Proxima Nova".
- Slight fixes to the "Register to Vote" button in the nav.

**Dropdown content push bug**
bug

**Dropdown cut-off bug, with colors for easier visual**
other bug

Refs [Pivotal ID #156275921](https://www.pivotaltracker.com/story/show/156275921)

### DIFF
--- a/resources/assets/scss/navigation.scss
+++ b/resources/assets/scss/navigation.scss
@@ -19,7 +19,7 @@
       color: $white;
       display: block;
       float: left;
-      font-weight: $weight-normal;
+      margin: 5px 0;
       text-shadow: none;
       text-transform: none;
     }
@@ -36,6 +36,8 @@
 }
 
 .navigation__menu {
+  overflow: auto;
+
   > .wrapper {
     @include media($medium) {
       float: right;
@@ -48,3 +50,7 @@
     float: left;
   }
 }
+
+// .navigation__secondary {
+//   background: red;
+// }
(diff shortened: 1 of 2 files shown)


resources/assets/scss/navigation.scss
```diff
@@ -34,9 +36,17 @@
...
   > .wrapper {
     @include media($medium) {
       float: right;
     }
   }
 }
+
+.navigation__primary {
+  @include media($medium) {
+    float: left;
```
DFurnes: What's this one for? Might be worth leaving a comment for future reference.

weerd: I don't think it necessarily requires a comment 🤔 I noticed that it's not floated but the other items on the right are. This causes the primary nav container to spread the whole width and could cause bugs in some browsers. I just find it more of a best practice to properly represent the containers and how they layout. I can add a comment if you feel it's necessary, but was more something that I felt should be done and just wasnt'. 🤷‍♂️

PR #929, opened 2018-05-18
Initial progress on showing quantity.

weerd:
# 🚧Work In Progress 🚧

### What does this PR do?
This PR enables showing a member's prior quantity if they already submitted a photo and quantity, so that they can see their progress, particularly for a collection campaign, and update their total easily based off of their prior amount.

...

Fixes #

### Checklist


resources/assets/helpers/forms.js
```diff
@@ -1,6 +1,26 @@
...
+
+/**
+ * Calculate the difference between a total value and a submitted value.
+ * Returns submitted value if calculation cannot be completed.
+ *
+ * @param  {Number|String} total
+ * @param  {Number|String} submittedValue
+ * @return {Number}
+ */
+export function calculateDifference(total, submittedValue) {
```
weerd: This might be better placed in the more general `helpers/index.js` but since it returns the `submittedValue` if the calculation can't be done (which seemed to apply more for form submission stuff) I decided to place into the `forms.js` file. 🤔

resources/assets/helpers/forms.js
```diff
@@ -1,6 +1,26 @@
...
+  if (!isInteger(Number(total))) {
+    return submittedValue;
+  }
+
+  if (!isInteger(Number(submittedValue))) {
+    return submittedValue;
+  }
```
mendelB: Can these be combined into one conditional, or is this a conscious separation?

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -137,12 +148,18 @@ class PhotoSubmissionAction extends React.Component {
   };
 
   resetForm = () => {
+    const signup = get(
+      this.props.submissions.items[this.props.id],
+      'data.signup.data',
+      null,
```
DFurnes: Do we need to explicitly set `null` as a default value here, or is `undefined` okay?

resources/assets/helpers/forms.js
```diff
@@ -1,6 +1,26 @@
...
+ * @param  {Number|String} submittedValue
+ * @return {Number}
+ */
+export function calculateDifference(total, submittedValue) {
+  if (!isInteger(Number(total))) {
+    return submittedValue;
+  }
+
+  if (!isInteger(Number(submittedValue))) {
+    return submittedValue;
```
DFurnes: Should this return a default like `0` if the submittedValue is invalid? Or throw?

app/Http/Requests/PostRequest.php
```diff
@@ -29,6 +29,7 @@ public function messages()
                 return [
                     'file.required' => 'An uploaded photo is required.',
                     'quantity.required_if' => 'The quantity field is required.',
+                    'quantity.min' => $this->setMinQuantityMessage(),
```
DFurnes: Neat!

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
weerd: I think `undefined` would probably be fine, but since I defaulted to setting it as `null` in the initial state in the `constructor()` I figured I'd be consistent 🤷‍♂️

resources/assets/helpers/forms.js
weerd: If it's invalid (meaning it's just a string of letters or field was left entirely empty) I think we want the validation on the Laravel end to cover that so the message pops up. So the reason it returns the `submittedValue` is if someone enters something like `asdasd`,  it would keep that, return it to let the validator do it's job :)

DFurnes: Hmm, alright. Let's at least update the `@return` to indicate this might return a string.

resources/assets/helpers/forms.js
weerd: More or less a conscious separation after trying a few things. But could possibly combine w/ `&&` if y'all feel strongly about it. Just need to consider if that affects logic at all which I don't think it does 🤔

resources/assets/helpers/forms.js
weerd: oh! yeah, good call.

resources/assets/helpers/forms.js
DFurnes: I like combining the two if statements, but maybe separating each part of the condition into a named boolean like `const invalidTotal = !isInteger(Number(total))`.

mendelB: no biggie either way I don't think

resources/assets/helpers/forms.js
```diff
@@ -1,6 +1,25 @@
...
+ */
+export function calculateDifference(total, submittedValue) {
+  const validTotal = total && isInteger(Number(total));
+  const validValue = submittedValue && isInteger(Number(submittedValue));
+
+  if (!validTotal || !validValue) {
+    return submittedValue;
+  }
+
+  return Number(submittedValue) - Number(total);
```
weerd: Switched this around a bit based on feedback. While reworking this I learned that casting an empty form submission to an integer using `Number()`, it converts it to `0` 🤔... 

Thus the need to use `&&`:
```js
const validValue = submittedValue && isInteger(Number(submittedValue));
```

PR #931, opened 2018-05-18
Add contentful migration for ReferralSubmissionAction content type

mendelB:
### What does this PR do?

This PR adds a Contentful Migration Script to create a ReferralSubmissionAction Content Type with some basic fields.

I only added the fields that felt like a given deal for this action, following a similar format to the text and photo submission actions. 

[#157363643](https://www.pivotaltracker.com/story/show/157363643)

image


mendelB:
Just updating here that we aren't adding any label customization as of yet.

PR #933, opened 2018-05-18
Some API and component cleanup and updates.

weerd:
### What does this PR do?
This PR does some quick cleanup and updates to the API when fetching Campaign Posts, and also updates the `VoterRegistrationActionContainer` to use the user ID selector.

Fixed this stuff up while working on an approach for the PSA quantity display updates, but decided I need to go in a different direction. The updates still seemed valid though, thus the PR!

🌵

### DIFF
--- a/app/Http/Controllers/Api/CampaignPostsController.php
+++ b/app/Http/Controllers/Api/CampaignPostsController.php
@@ -48,7 +48,11 @@ class CampaignPostsController extends Controller
      */
     public function index($id, Request $request)
     {
-        return $this->postRepository->getCampaignPosts($id, $request->all());
+        $query = $request->all();
+
+        $query['filter']['campaign_id'] = $id;
+
+        return $this->postRepository->getPosts($id, $query);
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Api/CampaignPostsController.php
```diff
@@ -48,7 +48,11 @@ public function __construct(CampaignRepository $campaignRepository, PostReposito
      */
     public function index($id, Request $request)
     {
-        return $this->postRepository->getCampaignPosts($id, $request->all());
+        $query = $request->all();
+
+        $query['filter']['campaign_id'] = $id;
+
+        return $this->postRepository->getPosts($id, $query);
```
mendelB: Was this meant to be just `getPosts($query)`? since it seems like the `getPosts` method only accepts a `$query` argument.

weerd: oh! good catch!

PR #936, opened 2018-05-21
Psa get user campaign signups

weerd:
### What does this PR do?
This PR adds some initial work to help with showcasing the prior quantity on the `PhotoSubmissionAction` is a member has already submitted a reportback. 

To accomplish this, we need to talk to Rogue and get the member's prior signup. So this PR adds a new API endpoint to grab the singup information and adds the necessary controllers and repository. On the client side of things, instead of using the API middleware and the Redux store, it made more sense to deal with storing the signup in the local state for the component. However, this required updating the API  methods so calls can be made _outside_ of using the middleware approach. So now there's a `helpers/api.js` file with methods for making requests, and the api middleware uses these methods as well. This currently only applies for the `getRequest()`, but eventually can more `postRequest()` into the helper file as well.

One caveat is that if there are multiple `*SubmissionAction`s on a campaign action page, it means each will be doing their own request to get the signup and storing the data locally. While not ideal, this current setup helps for when we eventually switch this call to a GraphQL call, and that does some nice caching of requests and avoids the additional calls etc; slated for a future update.

Refs [Pivotal ID #157633315](https://www.pivotaltracker.com/story/show/157633315)

### Checklist


resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -43,11 +46,25 @@ class PhotoSubmissionAction extends React.Component {
...
+  componentDidMount() {
+    const request = getUserCampaignSignups(
+      this.props.userId,
+      this.props.legacyCampaignId,
+      this.props.legacyCampaignRunId,
+    );
+
+    // @TODO: handle if errors.
+    request.then(response => {
+      this.setSignup(response.data['0']);
```
weerd: Not really doing much with the data yet; just setting it into state here and also when a successful submission is made and the returned data contains the new signup info (https://github.com/DoSomething/phoenix-next/pull/936/files#diff-4cb2e283ba27748d30bb6d53ba5a8e15R143) Will be utilized in future for showcasing the quantity, etc, but didn't want this PR to get to crazy.

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -111,12 +128,19 @@ class PhotoSubmissionAction extends React.Component {
     });
   };
 
+  setSignup = data => {
```
DFurnes: I wonder if it might make sense to treat methods like this as event handlers and name `handleSignupResponse` (a la `handleClick`)? I think that'd also help make it clearer why this is safe to use in `componentDidMount`.

resources/assets/helpers/api.js
```diff
@@ -0,0 +1,45 @@
...
+
+import { PHOENIX_URL } from '../constants';
+
+/**
+ * Send a GET request.
+ *
+ * @param  {Object} payload
+ * @return {Object}
+ */
+export function getRequest(payload) {
```
DFurnes: I think my slight preference would be to use named `url, query` arguments here instead of an object (since it's clearer what the parameters are at a glance). (And then the Redux middleware can handle translating action into those params.)

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
DFurnes: Why is this `response.data['0']` instead of `response.data`?

weerd: Data comes in from Rogue as an array since it's the index signups endpoint, just filtered to get the single signup. Debated having the response from the custom `getUserCampaignSignups()` return a single item.

DFurnes: A ha, that makes sense now!! Next question - why `'0'` instead of an integer`0`?

weerd: Oh, that's just cause I was being dumb 🤦‍♂️

resources/assets/helpers/api.js
weerd: Yeah, I think that makes sense and would defs be clearer. `payload` is a bit more Redux-y anyhoo

Issue #925, opened 2018-05-17
Migrate from Campaign 'Action Steps' to 'Action Pages'


mendelB:
### Current Behavior

What is the current behavior?
Campaigns have an `actionSteps` reference field (Campaign Contentful Content Types and corresponding Entities -that is) within which we store the links to Campaign Action Steps and our other assorted actions and blocks for the action page.

We also add a hardcoded `/action` route to all active campaigns, which renders the action steps in `ActionPage` and `ActionSteps` components.

### Expected Behavior
In the spirit of our migration over to Pages, where we source every subpage on a campaign from a Contentful `Page` entry (and as we've done with the Activity Feed -> Community Pages), we'd like to migrate over to a world of Action Pages.

This would include:

- Moving all Campaigns `actionSteps` blocks over to a new `Page` in the campaigns `pages` field.
  - While doing so, we'd like to transform all legacy `CampaignActionSteps` to `ContentBlocks` and if necessary, port the CampaignActionSteps `photos` field over to an `ImagesBlock`
- Manipulating the navigation logic to only include a `/action` tab if the Campaign does *not* have a new 'Action Page' (ensuring support for both Action Pages and ActionSteps until this migration is complete.)
- Updating our Routing logic to properly render either the `ActionPageContainer` for `actionSteps` or the correct Action page.

Some questions regarding the rendering of an Action Page instead of `actionSteps`:
The current [`ActionPage`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/pages/ActionPage/ActionPage.js) component we use to render `actionSteps`, as well as the `ActionSteps` component, contain several logical pieces of rendering functionality specific to actions steps that we would lose if blindly rendering a series of `blocks` in a page. These include:

- [filtering and truncating](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/pages/ActionPage/ActionPage.js#L22) steps if the user isn't signed up
- rendering a [revealer](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/pages/ActionPage/ActionSteps.js#L21) button to the user if they're not signed up
- [adding section headers](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/pages/ActionPage/ActionSteps.js#L72) to voter registration actions
[...]

PR #938, opened 2018-05-22
Action page navigation setup

mendelB:
### What does this PR do?

This PR adds some logic to enable rendering the `ActionPage` for migrated Campaigns, using the action Page instead of Campaign `actionSteps`.

- don't manually add an `action` tab if there's an action page
- render the action page blocks as the action steps if it exists and there are no `actionSteps`

This means we've got backfilled support for `actionSteps` as well as support for the new action Pages, so we can migrate piecemeal and ensure everything is working before pulling the switch on `actionSteps`

I went with simply replacing the `steps` prop but still maintaining the use of the regular flow for action steps (using the `ActionPage` component etc.) since there are a bunch of moving pieces involved in rendering the action page which simply rendering the blocks in a `campaignSubPage` would lose. (described more thoroughly in #925) 
Q: was this the right move? Or do we want to focus on moving this over completely to a regular page setup now - or perhaps after we migrate?

### DIFF
--- a/resources/assets/components/CampaignPageNavigation/CampaignPageNavigation.js
+++ b/resources/assets/components/CampaignPageNavigation/CampaignPageNavigation.js
@@ -38,8 +38,10 @@ const CampaignPageNavigation = ({
     });
   }
 
+  const hasActionPage = pages.find(page => page.fields.slug.endsWith('action'));
+
   // Conditional whether to include Action page.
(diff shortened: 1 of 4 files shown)


mendelB:
Updated to add a lil safety check to the `pages` finder and fetching the `actionSteps`

PR #941, opened 2018-05-24
Content block update

mendelB:
### What does this PR do?

This PR adds a quick refactor and a touch of clean up to the `ContentBlock` component:

- ~Always return a `Figure` component even~ Style content ~with `Flex`~ when there is no `image` prop to ensure proper formatting of the content
- Use the `withoutNulls` helper in the renderer so that we can do away with explicitly grabbing the `defaultProp`for `imageAlignment` and let React handle that work ☺️ 

I was testing out the action navigation changes with a new action 'page' and noticed that the ContentBlock was spreading across the entire page when there was no image:
image

But by ~always returning `Figure`~ styling properly:
image

### DIFF
--- a/resources/assets/components/blocks/ContentBlock/__snapshots__/ContentBlock.test.js.snap
+++ b/resources/assets/components/blocks/ContentBlock/__snapshots__/ContentBlock.test.js.snap
@@ -65,11 +65,19 @@ exports[`ContentBlock component it works beautifully with a mere content prop 1`
   <div
     className="margin-horizontal-md"
   >
-    <Markdown
-      className={null}
+    <Figure
+      alignment="right-collapse"
+      alt="content-block"
+      image={null}
+      imageClassName={null}
+      size="one-third"
     >
-      hi there
-    </Markdown>
+      <Markdown
+        className={null}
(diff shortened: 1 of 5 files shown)


weerd:
Would it not work with a simple class and width declaration? It doesn't really feel like a solution that needs flex either, since it's just one area of content and nothing showing on the other side to require special flex-box layout. I really think a simple class with a width designation should be fine... 🤔

mendelB:
Ugh I overthought this... lemme fix this

mendelB:
Updated to plain styling with classname based on `image` prop being provided.

resources/assets/components/blocks/ContentBlock/__snapshots__/ContentBlock.test.js.snap
```diff
@@ -65,11 +65,21 @@ exports[`ContentBlock component it works beautifully with a mere content prop 1`
   <div
     className="margin-horizontal-md"
   >
-    <Markdown
+    <Flex
```
weerd: Does the test still require `Flex`? Or is this the outer flex element?

PR #943, opened 2018-05-24
Add additional fields to the LegacyCampaign Entity

mendelB:
### What does this PR do?

This PR adds some additional fields to the `LegacyCampaign` Entity.

I had been under the mistaken impression that the new fields from #922 only needed to be added to the `Campaign` Entity. This will add -the available ones- to the `LegacyCampaign` as well.

[#157010624](https://www.pivotaltracker.com/story/show/157010624)

### DIFF
--- a/app/Entities/LegacyCampaign.php
+++ b/app/Entities/LegacyCampaign.php
@@ -45,6 +45,12 @@ class LegacyCampaign implements JsonSerializable
                 'url' => $this->legacyCampaign['cover_image']['default']['uri'],
                 'landscapeUrl' => $this->legacyCampaign['cover_image']['default']['sizes']['landscape']['uri'],
             ],
+            'staffPick' => $this->legacyCampaign['staff_pick'],
+            'cause' => $this->legacyCampaign['causes']['primary']['name'],
+            'additionalContent' => [
+                'noun' => $this->legacyCampaign['reportback_info']['noun'],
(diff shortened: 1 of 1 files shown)


app/Entities/LegacyCampaign.php
```diff
@@ -45,6 +45,12 @@ public function jsonSerialize()
                 'url' => $this->legacyCampaign['cover_image']['default']['uri'],
                 'landscapeUrl' => $this->legacyCampaign['cover_image']['default']['sizes']['landscape']['uri'],
             ],
+            'staffPick' => $this->legacyCampaign['staff_pick'],
+            'cause' => $this->legacyCampaign['causes']['primary']['name'],
+            'additionalContent' => [
+                'noun' => $this->legacyCampaign['reportback_info']['noun'],
+                'verb' => $this->legacyCampaign['reportback_info']['verb'],
```
mendelB: Should `noun` and `verb` actually contain the `singular` and `plural` fields, to match the format we have in Contentful? (singular would always be `null` since AFAIK ashes only returns one noun and verb and it's always plural)
```php
'additionalContent' => [
                'noun' => [
                    'singular' => null,
                    'plural' => $this->legacyCampaign['reportback_info']['noun'],
                ],
                'verb' => [
                    'singular' => null,
                    'plural' => $this->legacyCampaign['reportback_info']['verb'],
                ],
            ],
```

Issue #942, opened 2018-05-24
Campaigns Index endpoint


mendelB:
### User Story
As a developer, I want to implement an index endpoint for `/campaigns` on Phoenix.

#### Additional Information
After some discussion, we've fine-tuned our need at the moment to adding support for batched ID requests to this endpoint. E.G.
`/campaigns?ids=1,2,3`

We'd limit this to a small number of IDs (in the 10-15 range), and merge a return value of the Contentful campaigns for all provided Contentful IDs and (transformed-to-align-schema) Ashes campaigns based on the provided ashes IDs.

Eventually, we'd like to implement a caching system using Fastly and a Contentful webhook so that services can be comfortable sending us multiple 'show' requests instead of the somewhat inelegant batched ID requests.

_Any other info or references to cards/tickets from other services (Trello, etc)._
Pivotal ID: [#157410894](https://www.pivotaltracker.com/story/show/157410894)
Pivotal ID: [#155672260](https://www.pivotaltracker.com/n/projects/2019429/stories/155672260)

### Why This Matters
This is another step towards a full-fledged Phoenix Campaigns API.
We'd also like to act as a proxy for the Ashes `/campaigns` index endpoint, to help do away with multiple services querying Ashes, and instead, in addition to us serving Contentful campaigns from this endpoint, we'd also be supporting Ashes campaigns.

### Tasks
This feature includes the following tasks:
- [ ] Add batched ID `/campaigns` endpoint support for Contentful and Ashes IDs

mendelB:
https://dosomething.slack.com/archives/C3ASB4204/p1527193987000147
^ we discuss the value to query against for Contentful campaigns (`legacyCampaignId` vs. `sys.id`) and determine that the approach should be to query likely legacy IDs against Contentful, then filter the IDs that were not found, and pass those over to an Ashes query.

weerd:
Side note, I think we want something like: `/campaigns?filter[id]=1,2,3`.

Rogue's docs provide some good [examples](https://github.com/DoSomething/rogue/blob/master/docs/endpoints/posts.md)!

DFurnes:
Yep! We've settled on a pattern [suggested by JSON API](http://jsonapi.org/recommendations/#filtering), where we "namespace" filters under a `filter[]` query parameter. This way we wouldn't run into naming collisions one day if, for example, we wanted to filter something by pages (e.g. `?filter[page]=something&page=2`).

mendelB:
[...]

PR #946, opened 2018-05-25
Fetch index of Legacy Campaigns By ID

mendelB:
### What does this PR do?

This PR adds basic setup to enable fetching an index of legacy campaigns from Ashes.

- Adds a `getCampaigns` method to the `PhoenixLegacy` Service [4140f43](https://github.com/DoSomething/phoenix-next/commit/4140f43ff5232404dd66ffcf002546908c0415c4)
- Adds a `findByLegacyCampaignIds` method to the `CampaignRepository` [299a6c5](https://github.com/DoSomething/phoenix-next/commit/299a6c56f7ef614e24b5cd0bbe5788cc15fcbf06)
- Removes `data` index in the `LegacyCampaign` Entity constructor so that an index of campaign `data` can be used to construct `LegacyCampaigns`s as well

This is the first step towards the #942 feature.

Qs:
- Should the `findByLegacyCampaignIds` or `getCampaigns` methods be doing any validation to ensure that there aren't more than n `$ids` provided? (Since we discussed limiting this call to 10-15 IDs)
  - or also, in the case of `getCampaigns` should we only be allowing calls with an `ids` query param?
- How's the naming? I simply used the plural of the two existing 'show' methods we had in use.

#942 
[#157410894](https://www.pivotaltracker.com/story/show/157410894)

### DIFF
--- a/app/Entities/LegacyCampaign.php
+++ b/app/Entities/LegacyCampaign.php
@@ -20,7 +20,7 @@ class LegacyCampaign implements JsonSerializable
      */
     public function __construct($legacyCampaign)
     {
-        $this->legacyCampaign = $legacyCampaign['data'];
+        $this->legacyCampaign = $legacyCampaign;
     }
 
     /**
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -120,7 +120,8 @@ class CampaignRepository
      * @param  array $ids
      * @return \Illuminate\Support\Collection
      */
-    public function findByLegacyCampaignIds($ids) {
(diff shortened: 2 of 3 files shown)


app/Entities/LegacyCampaign.php
```diff
@@ -20,7 +20,7 @@ class LegacyCampaign implements JsonSerializable
      */
     public function __construct($legacyCampaign)
     {
-        $this->legacyCampaign = $legacyCampaign['data'];
+        $this->legacyCampaign = $legacyCampaign;
```
weerd: Nice. This feels cleaner 👌

PR #930, opened 2018-05-18
Referral Submission Action Component

mendelB:
### What does this PR do?

This PR adds the `ReferralSubmissionAction` component.

### Any background context you want to provide?
I copied *heavily* from the fantastic setup of the `TextSubmission` and `PhotoSubmission` action components since a lot of this follows an identical pattern. 
I imagine once we flesh this out more there may be opportunity to refactor some common form methods into helpers to DRY things up

More work to be done on the form field end since we want to support more than just an `email` field. [slack convo](https://dosomething.slack.com/archives/C3ASB4204/p1526653625000714)

### What are the relevant tickets/cards?
[#157363643](https://www.pivotaltracker.com/story/show/157363643)

### Checklist

* [ ] Documentation added for new features/changed endpoints.
* [x] Added appropriate feature/unit tests.
* [ ] Added screenshot to PR description of related front-end updates on **small** screens.
* [ ] Added screenshot to PR description of related front-end updates on **medium** screens.
* [ ] Added screenshot to PR description of related front-end updates on **large** screens.

(that lil icon in the input is just from my web extension)
image

PR #948, opened 2018-05-25
Tsa updates

weerd:
### What does this PR do?

This PR updates the Text Submission Action to bring it up to date with the newer code logic/approaches from the more recent updates to the Photo Submission Action and the Referral Submission Action.

🌵

### DIFF
--- a/resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
+++ b/resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
@@ -1,11 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import { has, get, mapValues } from 'lodash';
+import { get, has, invert, mapValues } from 'lodash';
 
 import Card from '../../utilities/Card/Card';
 import Modal from '../../utilities/Modal/Modal';
 import Button from '../../utilities/Button/Button';
+import { withoutUndefined } from '../../../helpers';
 import Markdown from '../../utilities/Markdown/Markdown';
 import FormValidation from '../../utilities/Form/FormValidation';
 import { getFieldErrors, setFormData } from '../../../helpers/forms';
@@ -79,7 +80,14 @@ class TextSubmissionAction extends React.Component {
 
     const formResponse = has(submissionItem, 'status') ? submissionItem : null;
 
-    const errors = getFieldErrors(formResponse);
+    const formErrors = getFieldErrors(formResponse);
+
+    // Associate errors to component field names.
+    const errors = withoutUndefined(
+      formErrors
+        ? mapValues(invert(this.fields), value => formErrors[value])
+        : null,
+    );
 
     return (
       <React.Fragment>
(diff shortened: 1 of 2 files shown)


resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
```diff
@@ -1,18 +1,30 @@
...
+import { getFieldErrors, setFormData } from '../../../helpers/forms';
 
 class TextSubmissionAction extends React.Component {
+  static getDerivedStateFromProps(nextProps) {
+    const response = nextProps.submissions.items[nextProps.id] || null;
+
+    if (has(response, 'status.success')) {
+      return {
+        showModal: true,
+        textValue: '',
```
weerd: I liked @mendelB approach in the RSA where the form is reset here, instead of how it's done in the PSA setting a `shouldResetForm` property to `true` and then using life cycle methods to see when that's changed. It had to be done in this way in the PSA (at least for now), but for less complex components it I preferred this simpler approach.

mendelB: Oh, but I do wish I could take credit for that idea 😇 
https://github.com/DoSomething/phoenix-next/pull/930#discussion_r189624573

resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
```diff
@@ -170,8 +164,8 @@ TextSubmissionAction.propTypes = {
   className: PropTypes.string,
   id: PropTypes.string.isRequired,
   initPostSubmissionItem: PropTypes.func.isRequired,
-  legacyCampaignId: PropTypes.string,
-  legacyCampaignRunId: PropTypes.string,
+  legacyCampaignId: PropTypes.string, // eslint-disable-line react/no-unused-prop-types
+  legacyCampaignRunId: PropTypes.string, // eslint-disable-line react/no-unused-prop-types
```
mendelB: Why do we still need these props?

weerd: They get used when the props get passed to `setFormData()` and added to the `details` JSON. It had them before and figured I'd leave them in, but eslint squawks at me.

mendelB: Oh right! eslint keepin' us on our toes 🐔

DFurnes: Hmm, this feels like a dependency we'd easily miss later on (since there's no way to know that it's actually being used without clicking in to that helper function).

How would you feel about updating the `setFormData` helper to take a `details` argument so it looks like this:

```js
// TextSubmissionAction.js
setFormData({
 // ...
}, {
  campaign_id: props.campaignId,
  legacy_campaign_id: props.legacyCampaignId,
  legacy_campaign_run_id: props.legacyCampaignRunId,
});

// helpers.js:
export function setFormData(values, details) {
  // ...
  if (details) {
    formData.append('details', JSON.stringify(details));
  }

  return formData;
}
```

weerd: Yeah, I was debating it a bit. I think that's a good suggestion though! I dig it 💃 

Since this is already merged, I'll do a separate PR with that update in a bit!

PR #954, opened 2018-06-04
Add Bertly shortened link to SocialDriveAction component

mendelB:
### What does this PR do?

This PR adds an API call to the `SocialDriveAction` component to fetch a shortened Bertly link via our Phoenix Bertly proxy endpoint.

What happens is we make the API call once the component mounts, and have all the buttons disabled until the state has it's `shortenedLink` property set (which happens as a callback once the API response is returned)

**UPDATE** I snuck in some changes to the Facebook share link handler to use the more robust share action flow. (I know it's completely repetitive of the `ShareAction`. Hoping we can clean this up as we refactor. Ashley really wanted this in. #blameCanada)

I added a new `postRequest` method to the `/helpers/api` so that we can avoid setting the authentication headers and `PHOENIX_URL` and whatnot from the `socialDriveAction` component.

I'm not sure that my `token` param approach was the right one for this method, but it seemed like an easy way to pass along the needed user token from state.

I also did not add an error `catch` to the API call, since if it errors out it'll just stay in the 'disabled' state forever. (wasn't sure of a better way for now, but I'm sure we can tweak this as we continue fleshing this out)

The git diffs for the actual `render()` markup is all kaboodled and thus gonna be annoying to review. You can just look out for some `disabled` attributes on `button`s, a [utility class name](157741116) and the [link bar value](https://github.com/DoSomething/phoenix-next/pull/954/files#diff-6fa015ab9fa63a40da148508a9827503R68) being conditioned on the `shortenedLink` state property being 'truthy'.
 
Refs [Pivotal ID #157741116](https://www.pivotaltracker.com/story/show/157741116)

jun-04-2018 14-31-31

### DIFF
--- a/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
+++ b/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
@@ -8,7 +8,6 @@ import linkIcon from './linkIcon.svg';
 import Card from '../../utilities/Card/Card';
 import Embed from '../../utilities/Embed/Embed';
 import { postRequest } from '../../../helpers/api';
-import { setFormData } from '../../../helpers/forms';
 import {
   dynamicString,
   handleFacebookShareClick,
@@ -28,10 +27,8 @@ class SocialDriveAction extends React.Component {
 
     const href = dynamicString(this.props.link, { userId });
 
-    const formData = setFormData({ url: withoutTokens(href) });
-
-    postRequest('api/v2/links', formData, token).then(response =>
-      this.setState({ shortenedLink: response.url }),
+    postRequest('api/v2/links', { url: withoutTokens(href) }, token).then(
+      response => this.setState({ shortenedLink: response.url }),
     );
   }
 
--- a/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
+++ b/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
@@ -2,112 +2,142 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
+import classnames from 'classnames';
 
 import linkIcon from './linkIcon.svg';
+import { postRequest } from '../../../helpers/api';
+import { setFormData } from '../../../helpers/forms';
 import Card from '../../utilities/Card/Card';
 import Embed from '../../utilities/Embed/Embed';
 import {
   dynamicString,
   handleFacebookShareClick,
   handleTwitterShareClick,
+  withoutTokens,
 } from '../../../helpers';
 
 import './social-drive.scss';
 
-const SocialDriveAction = props => {
-  const { link, showPageViews, userId } = props;
+class SocialDriveAction extends React.Component {
+  state = {
+    shortenedLink: null,
+  };
 
-  const href = dynamicString(link, { userId });
+  componentDidMount() {
+    const { userId, token } = this.props;
 
-  const handleCopyLinkClick = () => {
+    const href = dynamicString(this.props.link, { userId });
+
+    const formData = setFormData({ url: withoutTokens(href) });
+
+    postRequest('api/v2/links', formData, token).then(response =>
+      this.setState({ shortenedLink: response.url }),
+    );
+  }
+
+  handleCopyLinkClick = () => {
     const linkInput = document.querySelector('#social-drive-link');
     linkInput.select();
     document.execCommand('copy');
   };
 
-  return (
-    <Card
-      title="Your Online Drive"
-      className="social-drive-action rounded bordered"
-    >
-      <div className="padded">
-        <Embed url={link} />
-      </div>
+  render() {
+    const { link, showPageViews } = this.props;
+    const shortenedLink = this.state.shortenedLink;
 
-      <div className="padded link-area">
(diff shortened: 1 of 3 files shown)


resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -2,112 +2,142 @@
...
+    const formData = setFormData({ url: withoutTokens(href) });
 
-  const handleCopyLinkClick = () => {
+    postRequest('api/v2/links', formData, token).then(response =>
+      this.setState({ shortenedLink: response.url }),
+    );
+  }
+
+  handleCopyLinkClick = () => {
     const linkInput = document.querySelector('#social-drive-link');
```
DFurnes: Oops, I think I missed this in the previous PR - let's use a [ref](https://reactjs.org/docs/refs-and-the-dom.html) for this.

resources/assets/helpers/api.js
```diff
@@ -21,6 +21,25 @@ export function getRequest(url, query) {
...
+ * @param  {String} url
+ * @param  {Object} query
+ * @param  {String} token
+ * @return {Object}
+ */
+export function postRequest(url, query, token) {
+  const client = new RestApiClient(PHOENIX_URL, {
+    headers: {
+      Authorization: `Bearer ${token}`,
+      'Content-Type': 'multipart/form-data',
```
DFurnes: We can send this as a JSON payload on our end, I think.

mendelB:
**UPDATE** I snuck in some changes to the Facebook share link handler to use the more robust share action flow. (I know it's completely repetitive of the `ShareAction`. Hoping we can clean this up as we refactor. Ashley really wanted this in. #blameCanada)

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -2,112 +2,160 @@
...
+  }
+
+  componentDidMount() {
+    loadFacebookSDK();
+
+    const { userId, token } = this.props;
+
+    const href = dynamicString(this.props.link, { userId });
+
+    postRequest('api/v2/links', { url: withoutTokens(href) }, token).then(
```
DFurnes: We want the short-link to be for the tokenized URL, so we shouldn't be calling `withoutTokens` here right?

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -2,112 +2,160 @@
...
+    const { link, showPageViews } = this.props;
+    const shortenedLink = this.state.shortenedLink;
+
+    return (
+      <Card
+        title="Your Online Drive"
+        className="social-drive-action rounded bordered"
+      >
+        <div className="padded">
+          <Embed url={link} />
```
DFurnes: It might make sense to use `withoutTokens` here, though.

resources/assets/helpers/api.js
```diff
@@ -21,6 +21,25 @@ export function getRequest(url, query) {
...
+ *
+ * @param  {String} url
+ * @param  {Object} query
+ * @param  {String} token
+ * @return {Object}
+ */
+export function postRequest(url, query, token) {
+  const client = new RestApiClient(PHOENIX_URL, {
+    headers: {
+      Authorization: `Bearer ${token}`,
```
DFurnes: Wondering if it makes more sense to just read from `window.AUTH.token` here rather than having to grab and pass it from the Redux store... not something we need to decide now, though!

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
mendelB: This is already after the `dynamicString` ran so there *shouldn't* be any more tokens in here right? I added it because I had gotten an error (I *think* from Bertly) when I mistakenly passed through the unparsed link which still had `{userId}` in it. My thinking was to prevent such an error if the link had some unknown token in it. Does that make sense, or should I remove this call?

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
mendelB: [`Embed`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/utilities/Embed/Embed.js#L26) makes the call on it's own like a boss

resources/assets/helpers/api.js
mendelB: Yeah, I can see why that'd be cleaner!

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
DFurnes: Oh you're totally right, we have tokenized at this point. I wonder if we should always remove unknown tokens when we process something with `dynamicString`... either way, I'm good to merge as-is for now.

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
DFurnes: Ah, of course! Ez.

PR #955, opened 2018-06-04
Fix for validation issue when not using quantity field.

weerd:
### What does this PR do?

This PR fixes a bug when the quantity field is not displayed on the Photo Submission Action. While the `quantity` field would _not be validated as "required"_ if the `show_quantity` field value was `false`, the form request would still validate that the quantity value needed to both be an integer and a minimum value of 1. Thus, the PSA form, without a quantity field would never be able to be successfully submitted.

To solve this issue, I set up a separate `getPhotoRules()` method which checks against the `show_quantity` boolean value to see if it should add any validation rules for the `quantity` and all seems to work great 🍸 

I encountered this bug when switching some older Contentful Campaigns to use the new PSA instead of the old Photo Uploader.

🌵

### DIFF
--- a/app/Http/Requests/PostRequest.php
+++ b/app/Http/Requests/PostRequest.php
@@ -28,8 +28,7 @@ class PostRequest extends FormRequest
             case 'photo':
                 return [
                     'file.required' => 'An uploaded photo is required.',
-                    'quantity.required_if' => 'The quantity field is required.',
-                    'quantity.min' => $this->setMinQuantityMessage(),
+                    'quantity.min' => $this->getMinQuantityMessage(),
                     'text.max' => 'The caption field may not be greater than :max characters.',
                     'text.min' => 'The caption field may not be less than :min characters.',
                     'text.required' => 'The caption field for your photo is required.',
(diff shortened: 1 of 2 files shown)


app/Http/Requests/PostRequest.php
```diff
@@ -75,9 +68,32 @@ public function rules()
...
+    private function getPhotoRules()
+    {
+        $rules = [
+            'file' => 'required|file|image',
+            'show_quantity' => 'required|boolean',
+            'text' => 'required|min:4|max:60',
+            'why_participated' => 'required',
+        ];
+
+        if (filter_var($this->input('show_quantity'), FILTER_VALIDATE_BOOLEAN)) {
```
weerd: It technically worked fine with the values of `"0"` or `"1"` which come from the form data, but I preferred converting to a boolean to avoid any string funny-business and also for piece of mind 😌

weerd:
Hold off on this one... I realized late last night that the issue actually stems from the `quantity` being passed through via `FormData`, and thus being present in the form request. Since it's present with a value of `null` it is still getting validated. The ideal solution removes it from the `FormData` object that is sent through to the backend.

weerd:
Closing this in favor of the solution in #959

PR #958, opened 2018-06-05
Display clicks in the Social Drive Action.

DFurnes:
### What does this PR do?

This pull request adds the "total clicks" interface to the Social Drive Action, using Bertly's `/clicks` endpoint. This allows users to see how many people have clicked on their personalized share link!

At the moment, this is done using a chained HTTP request in our LinkController proxy, but could probably be added as a query-string option to that `POST /` route in Bertly.

screen shot 2018-06-05 at 12 15 21 pm

For now, any failures will gracefully display the unshortened link and zero clicks. We can consider other fallback options in a follow-up PR as well.

Refs [Pivotal ID #157741138](https://www.pivotaltracker.com/story/show/157741138).

### Checklist

### DIFF
--- a/app/Http/Controllers/Api/LinkController.php
+++ b/app/Http/Controllers/Api/LinkController.php
@@ -44,10 +44,10 @@ class LinkController extends Controller
 
         $url = $request->input('url');
         $host = parse_url($url, PHP_URL_HOST);
-        if (in_array($host, $whitelist)) {
-            $url = $this->bertly->shorten($url);
+        if (! in_array($host, $whitelist)) {
+            return ['url' => $url, 'count' => '0'];
         }
 
-        return response()->json(['url' => $url]);
+        return $this->bertly->shorten($url);
     }
 }
--- a/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
+++ b/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
@@ -37,8 +37,8 @@ class SocialDriveAction extends React.Component {
 
     const href = dynamicString(this.props.link, { userId });
 
-    postRequest('api/v2/links', { url: withoutTokens(href) }, token).then(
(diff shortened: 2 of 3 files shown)


resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -143,7 +143,7 @@ class SocialDriveAction extends React.Component {
               <span className="page-views__text caps-lock">
                 total page views
               </span>
-              <h1 className="page-views__amount">5</h1>
+              <h1 className="page-views__amount">{this.state.count || 0}</h1>
```
mendelB: Do you think we should default this to `N/A` or some other indicator that we're not actually applying the clicked amount?

app/Http/Controllers/Api/LinkController.php
```diff
@@ -44,10 +44,10 @@ public function store(Request $request)
 
         $url = $request->input('url');
         $host = parse_url($url, PHP_URL_HOST);
-        if (in_array($host, $whitelist)) {
-            $url = $this->bertly->shorten($url);
+        if (! in_array($host, $whitelist)) {
+            return ['url' => $url, 'count' => '0'];
```
mendelB: Eventually, do you think we'd want to return an error of some sort or `404` in this scenario?

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
DFurnes: Yeah, I wasn't sure what the best way to handle this is. I could see a "?" or "N/A" being better.

app/Http/Controllers/Api/LinkController.php
DFurnes: Yeah, that might be better. Wasn't sure how strict we wanted to be.

PR #959, opened 2018-06-05
Psa validation bug fix part deux

weerd:
### What does this PR do?

This PR fixes a bug when the quantity field is not displayed on the Photo Submission Action. 

While the `quantity` field would _not be validated as "required"_ if the `show_quantity` field value was `false`, the form request would still validate that the quantity value needed to both be an integer and a minimum value of 1. This was because after the updates happened to add logic to show a member's prior quantity, the changes resulted in a value for quantity _always_ being passed even if indicated that the field should not show. The value being passed was `null` and thus the validation rules, upon seeing an existing property for `quantity`, would appropriately validate them.

To solve this issue, I updated the `PhotoSubmissionAction` to include some logic and _only_ include values for `quanity` and `previousQuantity` if `showQuantity` is `true`, and all seems to work great 🍸 

I encountered this bug when switching some older Contentful Campaigns to use the new PSA instead of the old Photo Uploader.

🌵

### DIFF
--- a/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
+++ b/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
@@ -136,8 +136,6 @@ class PhotoSubmissionAction extends React.Component {
    * @return {void}
    */
   handleFileUpload = media => {
-    console.log(media);
-
     this.setState({
       mediaValue: media,
     });
--- a/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
+++ b/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
@@ -164,9 +164,9 @@ class PhotoSubmissionAction extends React.Component {
     );
 
     if (this.props.showQuantityField) {
-      values['previousQuantity'] = get(this.state.signup, 'quantity', 0);
+      values.previousQuantity = get(this.state.signup, 'quantity', 0);
 
-      values['quantity'] = calculateDifference(
+      values.quantity = calculateDifference(
         get(this.state.signup, 'quantity', null),
         this.state.quantityValue,
       );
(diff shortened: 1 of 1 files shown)


weerd:
I added docblocks to the methods in the React Component since it was getting a little on the long side, and the docblocks help break it up visually and clarify what each method is doing :) Thoughts?

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -22,6 +22,12 @@ import {
 import './photo-submission-action.scss';
 
 class PhotoSubmissionAction extends React.Component {
+  /**
+   * Lifecycle method invoked before every render.
```
DFurnes: Is this before every render or just when the component receives new props?

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -91,18 +117,38 @@ class PhotoSubmissionAction extends React.Component {
...
   };
 
+  /**
+   * Handle file upload event.
+   *
+   * @param  {Object} media
+   * @return {void}
+   */
   handleFileUpload = media => {
+    console.log(media);
```
DFurnes: 👾

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
weerd: Before every render, according to the [React docs](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops).

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
weerd: Dang it!

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
DFurnes: A ha! Good to know. 💡

PR #960, opened 2018-06-05
Adjustments to the `parseBlock` method

mendelB:
### What does this PR do?

This PR adds some cases to the `parseBlock` method to account for `quiz`es and `campaignActionStep`s.

Also swapped the default case to merely return the `$block->entry` to avoid parsing some unknown block as a `campaignActionStep`

Threw in the Quiz case as a first step towards [Pivotal ID #157750196](https://www.pivotaltracker.com/story/show/157750196)

We were bumping into errors anytime we had an unrecognized block in a Contentful reference where it's be rendered as a `campaignActionStep` and [break down the system](https://i.ytimg.com/vi/RSAbLRnGiWg/maxresdefault.jpg)

[Slack](https://dosomething.slack.com/archives/C3ASB4204/p1528222293000020)
[also Slack](https://dosomething.slack.com/archives/C3ASB4204/p1528156878000035)
[Slack as well](https://dosomething.slack.com/archives/C3ASB4204/p1528131998000440)

 [Pivotal ID #157750196](https://www.pivotaltracker.com/story/show/157750196)

### DIFF
--- a/app/Entities/Entity.php
+++ b/app/Entities/Entity.php
@@ -63,6 +63,7 @@ class Entity implements ArrayAccess, JsonSerializable
             case 'callToAction':
                 return new CallToAction($block->entry);
             case 'campaignActionStep':
+            case 'componentActionStep':
                 return new CampaignActionStep($block->entry);
             case 'campaignUpdate':
                 return new CampaignUpdate($block->entry);
--- a/app/Entities/Entity.php
+++ b/app/Entities/Entity.php
@@ -86,6 +86,8 @@ class Entity implements ArrayAccess, JsonSerializable
                 return new PhotoSubmissionAction($block->entry);
             case 'photoUploaderAction':
                 return new PhotoUploaderAction($block->entry);
+            case 'quiz':
+                return new Quiz($block->entry);
             case 'referralSubmissionAction':
                 return new ReferralSubmissionAction($block->entry);
             case 'shareAction':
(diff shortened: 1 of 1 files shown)


app/Entities/Entity.php
```diff
@@ -62,6 +62,8 @@ public function parseBlock($block)
                 return new Affirmation($block->entry);
             case 'callToAction':
                 return new CallToAction($block->entry);
+            case 'campaignActionStep':
```
weerd: I can't remember well enough, but any concerns that we defaulted to `new CampaignActionStep($block->entry)` because there were scenarios from earlier efforts where `$block->getContentType()` would return something that we just wanted to default to an action step? Not sure.... maybe just confirm that some very early on campaigns (Sincerely Us) work as expected with this change... 🤔

mendelB: Ooh I had a slight suspicion that there was a reason for this default. I'll do some 🕵️ and see what I can find

mendelB: OK, I *think* this was to deal with `componentActionStep`s so that they'd render as a `campaignActionStep` as well
(https://app.contentful.com/spaces/81iqaqpfd8fy/entries/jzM9WgF7jM6cM6YkGKiym?previousEntries=3RDCSjtWc0QW6yG6Kqqsuq)
Can we just add a case for `componentActionStep`?

weerd: Yeah, I think that sounds good. I've been working on removing all instances of `componentActionSteps` so we can eventually delete the content type anyhoo. Will be easier if there's a specific case for it ;)

mendelB: Oh, awesome. OK, added in the new case.

PR #962, opened 2018-06-06
Implement Quizzes as `resultBlocks`

mendelB:
### What does this PR do?

This PR adds the ability to set a Quiz as a potential `resultBlock` for a Quiz (for Logic jumping etc.).

- Added a bit of work to the `/helpers#calculateResult`, so that it finds the winning `resultBlock` and `result` and returns those rather than just the IDs. (also changed Quiz state to store them instead of IDs as well) [3457548](https://github.com/DoSomething/phoenix-next/commit/3457548915077f402f9499589908300a92052b30)
- Added a `quizResultBlockHandler` function which gets called on quiz completion, to see if the winning resultBlock is a Quiz, and if so, to redirect to that Quiz's `slug`. [78beff9](https://github.com/DoSomething/phoenix-next/commit/78beff91503ea7d304e24b8b8d5ec7f30dc88777)
  - tightened the `replace` logic a bit to ensure we only replace the quiz slug at the end of the URL [41639d9](https://github.com/DoSomething/phoenix-next/commit/41639d996771c61a213d986cae90783ff3cb12c9)
- update those tests! Including adding a new case for the quiz result block redirect, adding support for the location and history, and updating the helper's test for its new functionality [409ecbe](https://github.com/DoSomething/phoenix-next/commit/409ecbe7a326cc18594343c1cb5e0f7873c1166e)

I initially tried just adding another logic case to the `renderResult` function, but that was problematic since it gets called within the `render` function and thus React would cry if we'd redirect during a render, so I added the lil `quizResultBlockHandler` function to be called right before updating the `state` with the quiz results.

Refs [Pivotal ID #157750196](https://www.pivotaltracker.com/story/show/157750196)

### Checklist

jun-06-2018 14-33-46

### DIFF
--- a/resources/assets/components/Quiz/Quiz.js
+++ b/resources/assets/components/Quiz/Quiz.js
@@ -71,8 +71,8 @@ class Quiz extends React.Component {
 
       // Retain the current pathname while replacing the active quiz's slug with the resultBlocks slug
       const newPath = location.pathname.replace(
-        `quiz/${slug}`,
-        `quiz/${resultBlockSlug}`,
+        new RegExp(`/quiz/${slug}$`),
+        `/quiz/${resultBlockSlug}`,
       );
 
       history.push(newPath);
--- a/resources/assets/components/Quiz/Quiz.js
+++ b/resources/assets/components/Quiz/Quiz.js
@@ -1,6 +1,6 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import { every } from 'lodash';
+import PropTypes from 'prop-types';
 import ReactRouterPropTypes from 'react-router-prop-types';
 
 import calculateResult from './helpers';
--- a/resources/assets/components/Quiz/Quiz.js
+++ b/resources/assets/components/Quiz/Quiz.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { every } from 'lodash';
+import ReactRouterPropTypes from 'react-router-prop-types';
 
 import calculateResult from './helpers';
 import { Flex, FlexCell } from '../Flex';
@@ -8,6 +9,7 @@ import QuizQuestion from './QuizQuestion';
 import Share from '../utilities/Share/Share';
 import QuizConclusion from './QuizConclusion';
 import ContentfulEntry from '../ContentfulEntry';
+import ScrollConcierge from '../ScrollConcierge';
 
 import './quiz.scss';
 
@@ -54,9 +56,28 @@ class Quiz extends React.Component {
         this.props.resultBlocks,
       );
 
+      this.quizResultBlockHandler(results.resultBlock);
+
       this.setState({ showResults: true, results });
     }
   };
+
+  // If the winning resultBlock is a Quiz, navigates to the new resultBlock's slug
+  quizResultBlockHandler(resultBlock) {
+    if (resultBlock && resultBlock.type === 'quiz') {
+      const { location, history, slug } = this.props;
+
+      const resultBlockSlug = resultBlock.fields.slug;
+
+      // Retain the current pathname while replacing the active quiz's slug with the resultBlocks slug
+      const newPath = location.pathname.replace(
+        `quiz/${slug}`,
+        `quiz/${resultBlockSlug}`,
+      );
+
+      history.push(newPath);
+    }
+  }
 
   selectChoice = (questionId, choiceId) => {
     this.setState({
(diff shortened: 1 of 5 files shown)


resources/assets/components/Quiz/Quiz.js
```diff
@@ -38,11 +40,36 @@ class Quiz extends React.Component {
...
+      );
+
+      this.quizResultBlockHandler(results.resultBlock);
+
       this.setState({ showResults: true, results });
     }
   }
 
+  // If the winning resultBlock is a Quiz, navigates to the new resultBlock's slug
+  quizResultBlockHandler(resultBlock) {
```
mendelB: Is this function name the worst that ever was?

resources/assets/components/Quiz/Quiz.js
```diff
@@ -38,11 +40,36 @@ class Quiz extends React.Component {
...
 
-      const results = calculateResult(this.state.choices, this.props.questions);
+      const results = calculateResult(
+        this.state.choices,
+        this.props.questions,
+        this.props.results,
+        this.props.resultBlocks,
+      );
+
+      this.quizResultBlockHandler(results.resultBlock);
```
mendelB: I felt like it made sense to call this before `setState`, since we wouldn't care for the new state anyway if this is a Quiz `resultBlock`

weerd:
Oh man, I'm fearing a solid rebase on my end after this 😬

resources/assets/components/Quiz/Quiz.js
```diff
@@ -38,11 +40,36 @@ class Quiz extends React.Component {
...
+    if (resultBlock && resultBlock.type === 'quiz') {
+      const { location, history, slug } = this.props;
+
+      const resultBlockSlug = resultBlock.fields.slug;
+
+      // Retain the current pathname while replacing the active quiz's slug with the resultBlocks slug
+      const newPath = location.pathname.replace(
+        new RegExp(`/quiz/${slug}$`),
+        `/quiz/${resultBlockSlug}`,
+      );
```
mendelB: I tried to play this as safely as possible, hence the `/quiz` prefix to really make sure that we're replacing the correct bit in the URL.

The issue with this might be that the Quiz is technically accessible from `/block/:quiz-slug` in which case this `replace` wouldn't work, and a logic jump would be defunct. In other words, this makes certain assumptions about the routing rules.

mendelB:
>Oh man, I'm fearing a solid rebase on my end after this 😬
Ah shoot, you're right 😖

resources/assets/components/Quiz/Quiz.js
```diff
@@ -91,6 +111,8 @@ class Quiz extends React.Component {
 
     return (
       <Flex className="quiz">
+        <ScrollConcierge />
```
mendelB: I added this because the Quiz in a logic jump would be rendered but the page wouldn't scroll up on the redirect which was incredibly awkward (since generally having finished all questions, you'd be on the bottom of the page). However, @ashleybaldwin had mentioned that we'd be adding the option for a `LedeBanner` etc for the upcoming Quiz campaign, and this would cause the quiz to *always* scroll past the `LedeBanner` which mightn't be a good thing?

mendelB:
*UPDATE* Rebased off of master to bring in the updates from #963. Please note that some of my comments are now collapsed but still do apply.

weerd:
>I initially tried just adding another logic case to the renderResult function, but that was problematic since it gets called within the render function and thus React would cry if we'd redirect during a render, so I added the lil quizResultBlockHandler function to be called right before updating the state with the quiz results.

Yeah, I experienced something as well. Initially trying to auto-submit in the `render()` before it returns the component, but React didn't seem cool with that and also just felt a bit gross and like mixing intentions.

resources/assets/components/Quiz/Quiz.js
mendelB: Just noting here: we discussed IRL. It'll only be called from the `completeQuiz` function which will be triggered from the `componentDidUpdate` for an `autoSubmit` quiz, or the conclusion button for non `autoSubmit` quizzes

resources/assets/components/Quiz/Quiz.js
weerd: Lets keep this for now and see how it goes. I can create a ticket to reassess in the near future.

resources/assets/components/Quiz/Quiz.js
```diff
@@ -47,12 +49,36 @@ class Quiz extends React.Component {
...
+      );
+
+      this.quizResultBlockHandler(results.resultBlock);
 
       this.setState({ showResults: true, results });
     }
   };
 
+  // If the winning resultBlock is a Quiz, navigates to the new resultBlock's slug
+  quizResultBlockHandler(resultBlock) {
```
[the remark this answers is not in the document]
mendelB: Oops, yes totally!

PR #607, opened 2018-01-17
Upgrading to Laravel 5.4

weerd:
### What does this PR do?
I'd like to try a piecemeal process of performing all the Laravel upgrade steps to move from 5.3 to 5.4 (and then eventually from 5.4 to 5.5). 

I've added an upgrade checklist to indicate what steps in the upgrade process have been completed based on the [upgrade guide](https://laravel.com/docs/5.4/upgrade) in the Laravel documentation. With this approach others can take over when they have some free time and complete a step in the process and eventually we'll do the full upgrade.

This is a little experiment to see if we can upgrade Laravel as a team!
go team

There's lots of useful stuff in Laravel 5.4 and also in 5.5 that would be great to have access too. Specifically it'd be great to be able to use the file upload faker for testing!

When reviewing the code changes, it might be best to go [commit-by-commit](https://github.com/DoSomething/phoenix-next/pull/607/commits) which easily follows the checklist items below!

### What are the relevant tickets/cards?
Refs https://www.pivotaltracker.com/story/show/154419780

### Checklist:
- [x] Updating Dependencies
- [x] Authorization (not applicable)
- [x] Blade (not applicable)
- [x] Bootstrappers (not applicable)
- [x] Broadcasting
- [x] Collections (not applicable)
- [x] Container (not applicable)
- [x] Console (not applicable)
- [x] Database (not applicable)
- [x] Eloquent (not applicable)
- [x] Events (not applicable)
- [x] Exceptions (not applicable)
- [x] Mail (not applicable)
- [x] Queue (not applicable)
- [x] Redis (not applicable, not using clusters)
- [x] Routing
- [x] Sessions (not applicable)
- [x] Testing
- [x] Translation
- [x] URL Generation
- [x] Validation
- [x] Miscellaneous

PR #964, opened 2018-06-07
Upgrading to Laravel 5.5!

weerd:
### What does this PR do?
Similar to #607 I'd like to try a piecemeal process of performing all the Laravel upgrade steps to move from 5.4 to 5.5. 

<p align="center">engage upgrade</p>

There's an upgrade checklist to indicate what steps in the upgrade process have been completed based on the [upgrade guide](https://laravel.com/docs/5.5/upgrade) in the Laravel documentation. With this approach others can take over when they have some free time and complete a step in the process and eventually we'll do the full upgrade.

When reviewing the code changes, it might be best to go [commit-by-commit](https://github.com/DoSomething/phoenix-next/964/commits) which easily follows the checklist items below!

Refs https://www.pivotaltracker.com/story/show/158189642

### Checklist:
* [x] Updating Dependencies
* [x] Artisan
* [x] Authorization
* [x] Cache
* [x] Eloquent
* [x] Exception Format
* [x] Filesystem
* [x] Mail
* [x] Queues
* [x] Requests
* [x] Testing
* [x] Translation
* [x] Validation
* [x] Views
* [x] Miscellaneous


DFurnes:
The default exception format was changed in this release, but we're currently overriding in our [Handler](https://github.com/DoSomething/phoenix-next/blob/0cb44212c8cf31ad4a7ce794d439f4eede4f5718/app/Exceptions/Handler.php) to set a custom format (with nested `success` and `error` objects). I'd love to settle on the simpler Laravel style, but seems out of scope of this PR since our overrides continue to work.

DFurnes:
Also worth shouting out some interesting changes in the "Request" section of the upgrade guide, but none that affect us in this application.

DFurnes:
Rebased to clear up the `composer.lock` conflict.

DFurnes:
@weerd @mendelB I think this is ready for merge!

config/session.php
```diff
@@ -122,7 +122,10 @@
     |
     */
 
-    'cookie' => 'laravel_session',
+    'cookie' => env(
```
DFurnes: We may want to set `SESSION_COOKIE` to `laravel_session` before we deploy, so we don't lose existing sessions. On the other hand, I could see it making sense to rip off the bandaid & have these cookie names be app-specific so they don't conflict (since a Northstar cookie would otherwise both be `laravel_session`).

DFurnes:
Fixed up a fatal error on the campaigns index caused by the new response format from the campaign repository's `getAll` (also seen [on Preview](http://phoenix-preview.dosomething.org/us/campaigns)). If we don't decide to merge this, I can cherry-pick out.

config/session.php
mendelB: The latter sounds like a good idea IMO.

DFurnes:
Fixed one more tiny issue and fixed in a497c66. Laravel now includes a built-in `redirect` method on the router that overrides our macro, but doesn't support redirects w/ arguments. For now, I've ripped out our macro and used a closure route to handle remaining redirects.

routes/web.php
```diff
@@ -26,17 +26,18 @@
 $router->redirect('campaigns', 'us/campaigns');
 
 // Non campaign pages
-$router->redirect('/{slug}', 'us/{slug}');
```
mendelB: So we're fine with these no longer being redirected?

DFurnes: They're handled by the `campaigns/{slug}/{clientRoute?}` redirect below!

DFurnes: Oh wait, I see. They totally should be. Whoops!

mendelB: DAVE. Our new content strategy!

mendelB: Ben is wagging his finger disapprovingly in the corner rn

DFurnes: Hah! Fixed in 9081638.

weerd:
> I'd love to settle on the simpler Laravel style, but seems out of scope of this PR since our overrides continue to work.

Yeah, I can create a ticket to switch over to the Laravel style, and we'll want to do a bit of testing/let other teams know using the `/campaign` endpoint!

PR #965, opened 2018-06-07
Add gating and signup conversion functionality to the Quiz

mendelB:
### What does this PR do?

This PR adds functionality to auto signup users on quiz submission (for non auto-submit quizzes), as well as support to ensure unauthenticated users can see their quiz result after going through the Northstar authentication flow.

### background context:
There are some moving parts to this, so I'll attempt to summarize what in tarnation is going on here:

#### ⚜️Our Task 
<image src="https://user-images.githubusercontent.com/12417657/41127185-07cdadf6-6a78-11e8-88a5-5ac40d348825.jpg" width="40%" />
We want to auto sign-up users for the current campaign when they submit a completed quiz. 
The simple idea is that In the `completeQuiz` function - if the Quiz is not set to `autoSubmit`- (we don't want to auto sign up users for a quiz if there is no explicit submit button with some context) - than we will sign up the user for the current campaign before showing them their quiz results. 

#### ⚜️It's complicated 😬 
<image src="https://user-images.githubusercontent.com/12417657/41128383-0080016c-6a7c-11e8-844d-2269531ce01a.jpg" width="40%"/>

This becomes slightly complicated for unauthenticated users -- who in order to be signed up for the campaign need to first go through the login flow on Northstar, which means they're leaving our domain with only the [current URL](https://github.com/DoSomething/phoenix-next/blob/quiz-gating-and-signup-conversion/app/Http/Controllers/AuthController.php#L47) to guide them back. Once they're redirected back to the quiz it just starts over and they need to re-take the quiz to see their results. LAME.

#### ⚜️Diego's bright idea 💡 
<image src="https://user-images.githubusercontent.com/12417657/41128440-2cecfa8e-6a7c-11e8-8009-03728c090621.gif" width="40%" />
@weerd helped (well, more like I helped, he did) craft a solution whereby - before we trigger the signup - we equip the current URL with some helpful query parameters based on the results they should be seeing. Then we can simply check the query params whenever we render a Quiz and alter the state accordingly! Setting the appropriate `result` and `resultBlock` and then scrubbing the URL of these extra params.

I also had to finagle the `signupCreated` action and reducer to ensure that we wouldn't be popping up a post signup affirmation modal and ruining the zen. (This is something that may require some more refactoring/discussion as there seems to be an `OPENED_POST_SIGNUP_MODAL` action type hanging about which'd trigger the affirmation modal, as well as the `shouldShowAffirmation` being set from the `signupCreated` reducer case).

#### ⚜️Matt 🖖 
<image src="https://user-images.githubusercontent.com/12417657/41129235-1a0b701e-6a7f-11e8-8ad4-7e10aaeebd81.png" />
We also snuck in a fix for an issue @mshmsh5000 flagged regarding us not formatting the cover image URL passed from campaigns to Northstar for login flow, causing Yuge images to be rendered on Login pages [21088ee](https://github.com/DoSomething/phoenix-next/commit/21088ee14448ed7efe870c7c9ec406c7812d256f)

[Slack convo](https://dosomething.slack.com/archives/C2BPA7M8F/p1528408079000550)

Refs [Pivotal ID #157751720](https://www.pivotaltracker.com/story/show/157751720)

jun-07-2018 18-26-13

### DIFF
--- a/resources/assets/actions/event.js
+++ b/resources/assets/actions/event.js
@@ -1,7 +1,7 @@
 import * as actions from '../actions';
-import { isTimestampValid } from '../helpers';
-import { getArray, EVENT_STORAGE_KEY } from '../helpers/storage';
 import { isAuthenticated } from '../selectors/user';
+import { getArray, EVENT_STORAGE_KEY } from '../helpers/storage';
+import { isTimestampValid, contentfulImageUrl } from '../helpers';
 
 // Action: remove completed event from storage.
 export function completedEvent(index) {
@@ -52,7 +52,9 @@ export function queueEvent(actionCreatorName, ...args) {
       const { callToAction, coverImage, title } = getState().campaign;
 
       northstarOptions.title = encodeURIComponent(title);
-      northstarOptions.coverImage = coverImage.url;
+      northstarOptions.coverImage = encodeURIComponent(
+        contentfulImageUrl(coverImage.url, '800', '600', 'fill'),
+      );
       northstarOptions.callToAction = encodeURIComponent(callToAction);
     }
 
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -171,12 +171,15 @@ export function clickedSignUp(
           // If Drupal denied our signup request, check if we already had a signup.
           dispatch(checkForSignup(campaignId));
         } else {
-          // Create signup and track any data before redirects.
-          dispatch(signupCreated(campaignId));
-
-          // Take user to the action page if campaign is open.
           const endDate = get(state.campaign.endDate, 'date', null);
           const isClosed = isCampaignClosed(endDate);
+
+          // Create signup and track any data before redirects.
+          dispatch(
+            signupCreated(campaignId, shouldRedirectToActionTab && !isClosed),
+          );
+
+          // Take user to the action page if campaign is open.
           if (shouldRedirectToActionTab && !isClosed) {
             dispatch({ type: OPENED_POST_SIGNUP_MODAL });
             dispatch(push(campaignActionUrl));
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -141,7 +141,14 @@ export function clickedSignUp(
 
     // If the user is not logged in, handle this action later.
     if (!isAuthenticated(state)) {
-      return dispatch(queueEvent('clickedSignUp', campaignId, details));
+      return dispatch(
+        queueEvent(
+          'clickedSignUp',
+          campaignId,
+          details,
+          shouldRedirectToActionTab,
+        ),
+      );
     }
 
     // If we already have a signup, just go to the action page.
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -26,7 +26,7 @@ import {
  */
 
 // Action: a new signup was created for a campaign.
-export function signupCreated(campaignId) {
+export function signupCreated(campaignId, shouldShowAffirmation = true) {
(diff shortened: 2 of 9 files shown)


resources/assets/actions/signup.js
```diff
@@ -141,7 +142,14 @@ export function clickedSignUp(
 
     // If the user is not logged in, handle this action later.
     if (!isAuthenticated(state)) {
-      return dispatch(queueEvent('clickedSignUp', campaignId, details));
+      return dispatch(
+        queueEvent(
+          'clickedSignUp',
+          campaignId,
+          details,
+          shouldRedirectToActionTab,
```
mendelB: Not sure why this param wasn't being passed through, but it was causing a redirect to the action tab after login (as well as an affirmation modal pop up)

resources/assets/actions/event.js
```diff
@@ -52,7 +52,9 @@ export function queueEvent(actionCreatorName, ...args) {
       const { callToAction, coverImage, title } = getState().campaign;
 
       northstarOptions.title = encodeURIComponent(title);
-      northstarOptions.coverImage = coverImage.url;
+      northstarOptions.coverImage = encodeURIComponent(
+        contentfulImageUrl(coverImage.url, '800', '600', 'fill'),
```
weerd: Not necessary for this PR, but we could see if there's a better crop option for the layout on the NS login page since it's more of a portrait crop than a landscape. Something we can ask @lkpttn about!

resources/assets/components/Quiz/Quiz.js
```diff
@@ -1,29 +1,47 @@
...
 
 import './quiz.scss';
 
 class Quiz extends React.Component {
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
+
+    // grab state overrides from the query parameters
```
weerd: Nit-pick; capitalize the `g`. All comments seem to start with capitals.

resources/assets/components/Quiz/Quiz.js
```diff
@@ -1,29 +1,47 @@
...
+    const resultId = query('resultId');
+    const resultBlockId = query('resultBlockId');
+    // Also ensuring user authentication before applying the showResults query param override
+    const showResults = query('showResults') && props.isAuthenticated;
```
weerd: nice 👌

resources/assets/components/Quiz/Quiz.js
```diff
@@ -1,29 +1,47 @@
...
 
 class Quiz extends React.Component {
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
+
+    // grab state overrides from the query parameters
+    const resultId = query('resultId');
+    const resultBlockId = query('resultBlockId');
```
weerd: Just to confirm, if `query()` doesn't find `resultId` and `resultBlockId`, what value is assigned to them?

resources/assets/actions/event.js
mendelB: Definitely! I just copied over the options from LedeBanner in a hurry.

resources/assets/components/Quiz/Quiz.js
mendelB: It'll be `undefined`. I didn't think that was an issue, but can throw in an `|| null` if it matters?

weerd: Nah, I think it's ok. Just wanted to make sure. Maybe instead update the `query()` helper func to indicate that it returns `String|Undefined`?

mendelB: Yeah that's a good idea, I'll do that!

PR #966, opened 2018-06-08
Quick fix for the Quiz constructor

mendelB:
### What does this PR do?

This PR will adjust our Quiz URL scrubbing method to use `replaceState` instead of `pushState` so that if we are indeed scrubbing params from the URL, the user won't be stuck in a history loop. (where you'd click back, hit the old URL with the params, but then we'd scrub them again pushing you forward......)


weerd:
Ha! I think I experienced this when testing and just wasn't sure what was going on, lol.

PR #967, opened 2018-06-11
Filter out action page if campaign is closed

mendelB:
### What does this PR do?

This PR ensures that any Action Page would be filtered out of the Campaign Page Navigation when a campaign is closed

Last bit of work to ensure we're all set up for the grand Action Page Migration™£¢¡∞§¶£ºª 

Refs [Pivotal ID #156771690](https://www.pivotaltracker.com/story/show/156771690)

### DIFF
--- a/resources/assets/components/CampaignPageNavigation/CampaignPageNavigation.js
+++ b/resources/assets/components/CampaignPageNavigation/CampaignPageNavigation.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import { isActionPage } from '../../helpers';
 import { prepareCampaignPageSlug } from '../../helpers/campaign';
 import PageNavigation from '../utilities/PageNavigation/PageNavigation';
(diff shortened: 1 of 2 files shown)


resources/assets/components/CampaignPageNavigation/CampaignPageNavigation.js
```diff
@@ -17,11 +17,16 @@ const CampaignPageNavigation = ({
     return null;
   }
 
+  const isActionPage = page =>
+    page.type === 'page' && page.fields.slug.endsWith('action');
```
mendelB: Should I throw this method into `/helpers`?

weerd: Yeah, might end up being useful elsewhere!

mendelB: Ok cool! I'll add this in than

PR #970, opened 2018-06-12
Add SocialOverride functionality for Pages

mendelB:
### What does this PR do?

This PR enables `socialOverride` functionality for General Pages (guides/articles/static-pages)

- adds `socialOverride` field to the `Page` Entity's `jsonSerialize`
- adds a general `get_social_fields` helper method for Contentful entries / changes the campaign-specific method name to `get_campaign_social_fields`
- sets the `socialFields` variable when rendering a page from the `PageController` using the `get_social_fields` helper method

The `get_social_fields` helper is pretty specific to campaigns, so switched up the naming to make room for this more general `get_social_fields` method.

I ran into some issues since contrary to the [`CampaignRepository#findBySlug`](https://github.com/DoSomething/phoenix-next/blob/page-social-override/app/Repositories/CampaignRepository.php#L168), in the [`PageRepository#findBySlug`](https://github.com/DoSomething/phoenix-next/blob/page-social-override/app/Repositories/PageRepository.php#L48) we're not returning the JSON `encode/decode`d object. But being that we probably might end up adding caching in the near future, it felt like a good idea to enable this `get_social_fields` method to comply with either the decoded object or a `DynamicEntry` (Which is the current case with Pages)

Refs [Pivotal ID #158163700](https://www.pivotaltracker.com/story/show/158163700)

### DIFF
--- a/app/Repositories/PageRepository.php
+++ b/app/Repositories/PageRepository.php
@@ -28,7 +28,7 @@ class PageRepository
      * Find a page by its slug.
      *
      * @param  string $slug
-     * @return object
+     * @return \Contentful\Delivery\DynamicEntry
      * @throws \Illuminate\Database\Eloquent\ModelNotFoundException
      */
     public function findBySlug($slug)
--- a/app/Repositories/PageRepository.php
+++ b/app/Repositories/PageRepository.php
@@ -28,7 +28,7 @@ class PageRepository
      * Find a page by its slug.
      *
      * @param  string $slug
-     * @return stdClass
+     * @return object
      * @throws \Illuminate\Database\Eloquent\ModelNotFoundException
      */
     public function findBySlug($slug)
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -279,7 +279,7 @@ function get_social_fields($entry)
     if (! $socialOverride) {
         return;
     }
- 
+    
     return [
         'title' => $socialOverride->title,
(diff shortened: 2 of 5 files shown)


app/Repositories/PageRepository.php
```diff
@@ -28,7 +28,7 @@ public function __construct(Contentful $contentful)
      * Find a page by its slug.
      *
      * @param  string $slug
-     * @return stdClass
+     * @return object
```
mendelB: Should this be more specific? Something like `Contentful\Delivery\DynamicEntry` or `\Page` or something? (Same for the docblock in `get_social_fields` I guess)

weerd: Setting to `\Contentful\Delivery\DynamicEntry` seems cool to me.

app/helpers.php
```diff
@@ -262,11 +262,41 @@ function useOverrideIfSet($field, $base, $override)
 /**
  * Determine the fields to display in the social share.
  *
+ * @param  stdClass|object $entry
+ * @return array|null
+ */
+function get_social_fields($entry)
+{
+    // If this is an Entry which has been cast as JSON and thus completely JSON serialized.
```
weerd: Yeah, I think this makes sense, but maybe just add a `@TODO` to reassess once we take a look at caching again.

PR #974, opened 2018-06-13
Hide question number quiz

mendelB:
### What does this PR do?

This PR adds a `hideQuestionNumber` toggle to the quiz to enable hiding the `preTitle` on `QuizQuestions`s with the `Question #`.

This is especially useful for logic jumped / nested quizzes where the question numbers would awkwardly begin again from 1

Refs [Pivotal ID #158307276](https://www.pivotaltracker.com/story/show/158307276)

image

### DIFF
--- a/contentful/migrations/2018_06_12_002_add_hide_question_number_to_quiz.js
+++ b/contentful/migrations/2018_06_12_002_add_hide_question_number_to_quiz.js
@@ -1,0 +1,12 @@
+module.exports = function(migration) {
+  const quiz = migration.editContentType('quiz');
+
+  quiz
+    .createField('hideQuestionNumber')
+    .name('Hide Question Number')
+    .type('Boolean')
+    .required(false)
+    .localized(false);
+
+  quiz.moveField('hideQuestionNumber').afterField('autoSubmit');
+};
--- a/app/Entities/Quiz.php
+++ b/app/Entities/Quiz.php
@@ -70,6 +70,7 @@ class Quiz extends Entity implements JsonSerializable
                 'results' => $this->parseResults($this->results),
                 'questions' => $this->parseQuestions($this->questions),
                 'resultBlocks' => $this->parseBlocks($this->resultBlocks),
+                'hideQuestionNumber' => $this->hideQuestionNumber,
                 'additionalContent' => $this->additionalContent,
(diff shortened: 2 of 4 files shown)


contentful/migrations/2018_06_12_002_add_hide_question_number_to_quiz.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function(migration) {
+  const quiz = migration.editContentType('quiz');
+
+  quiz
+    .createField('hideQuestionNumber')
```
mendelB: How's the naming? Inspired by the `hideStepNumber` field. 
Should it perhaps be plural hideQuestionNumber*s* for the quiz and singular for the `QuizQuestion`? I considered that that might just be confusing.

contentful/migrations/2018_06_12_002_add_hide_question_number_to_quiz.js
```diff
@@ -0,0 +1,12 @@
...
+    .name('Hide Question Number')
+    .type('Boolean')
+    .required(false)
+    .localized(false);
+
+  quiz.moveField('hideQuestionNumber').afterField('autoSubmit');
```
mendelB: Wasn't positive about the field positioning here, but felt like it might be nice to have those two toggles together, and not hidden all the way under the long questions / results fields.

contentful/migrations/2018_06_12_002_add_hide_question_number_to_quiz.js
weerd: Make it plural for the Quiz.

contentful/migrations/2018_06_12_002_add_hide_question_number_to_quiz.js
weerd: ya, sounds good to me :)

contentful/migrations/2018_06_12_002_add_hide_question_number_to_quiz.js
weerd: Oh actually, nevermind. I see how more how it's used and would be confusing I think having to switch back and forth from plural to singular. 🤕

PR #975, opened 2018-06-13
Add rainbow logo for Pride!

DFurnes:
### What does this PR do?

This pull request updates the logo on Phoenix for Pride 2018! 🏳️‍🌈

screen shot 2018-06-13 at 11 32 22 am

I was trying to think about how we could make this easier to swap across the board (since we have three different front-ends a user may hit in the critical path), but couldn't come up with anything that I was satisfied with. I think this will be less of an issue as we continue to tear down Ashes.

Refs [Pivotal ID #157407505](https://www.pivotaltracker.com/story/show/157407505)

### Checklist


weerd:
🏳️‍🌈 🎉

mendelB:
> I was trying to think about how we could make this easier to swap across the board (since we have three different front-ends a user may hit in the critical path), but couldn't come up with anything that I was satisfied with. I think this will be less of an issue as we continue to tear down Ashes.

@DFurnes not sure if this is worth pursuing or a good idea, but would it at all make sense to add a modifier class to Forge such as `-pride-logo` or something, which, if applied to the `.navigation-logo` would add the pride one instead? 

If this is at all a good idea, or even separately, how ridiculous would it be to add a touch of logic into the `navigation.blade.php` files, which conditionally add the pride modifier class _if_ the current month is June? 
e.g.
```blade
<a class="navigation__logo {{ Carbon\Carbon::now()->month === 6 ? '-pride-logo' : '' }}" href="http://www.dosomething.org"><span>DoSomething.org</span></a>
```

Or is that just putting an undue amount of effort onto the blade files for something only relevant to one month a year which we can pretty easilly do manually?

DFurnes:
My one kinda-sorta reservation with having a permanent `-pride-logo` class was that we'd end up inlining that [rainbow-logo.svg](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/scss/rainbow-logo.svg) in our CSS file all year, adding an extra 5kb that's generally unused.

Ideally, we'd build this part of the UI in React, and could then do dead-code elimination based on whether or not that class is applied (and then swapping the logo could just be a matter of re-deploying).

Now that being said… 5kb is not really breaking the camel's back when you consider our overall bundle size, and it may be worth the small performance overhead to save us the trouble next year! I'm down for whatever approach you feel like taking here!

PR #976, opened 2018-06-13
Add title hiding toggle to Quiz

mendelB:
### What does this PR do?

~This PR adds the ability to set the Quiz title to be hidden~
- ~Contentful CLI migration for a new `hideTitle` field in Quiz.~
- ~`hideTitle` field in the Quiz Entity `jsonSerialize`~
- ~Quiz component will hide the `title` based on the `hideTitle` prop~

UPDATE: We actually just changed the `title` field to optional and problem solved no sweat!

Refs [Pivotal ID #158261009](https://www.pivotaltracker.com/story/show/158261009)

image


app/Entities/Quiz.php
```diff
@@ -66,6 +66,7 @@ public function jsonSerialize()
             'fields' => [
                 'title' => $this->title,
                 'slug' => $this->slug,
+                'hideTitle' => $this->hideTitle,
```
weerd: Instead of a new field to hide the title, why don't we just NOT make it required, and make the default prop `null`?

mendelB: Haha, omg that is a *great* point!

mendelB: Updated! I reckon we just can make that `title` field change directly in Contentful right?

weerd:
@mendelB wrap the `<h2 className="quiz__title">{this.props.title}</h2>` in a ternary, or it's going to output an empty `h2`!

PR #977, opened 2018-06-13
Using campaign sub page to render the quiz with lede banner.

weerd:
### What does this PR do?

This PR switches to using the `CampaignSubPage` for rendering out the Quiz, so that it gets access to the `LedeBanner`. It also does a little cleanup and a sort-of hacky update to pass along a prop to tell the Lede Banner not to show the signup button.

image

I'm not a big fan of this solution,  BUT it brings us part-way to moving to where we want to be and eventually using the same component to render all "pages" for a campaign. Some refactor work will address the `@TODO`'s I've included _after_ the migration is run to move `Actions` into their own page for the final step of everything a page. Then the real work begins ™

### Checklist

### DIFF
--- a/resources/assets/components/Quiz/Quiz.js
+++ b/resources/assets/components/Quiz/Quiz.js
@@ -193,6 +193,12 @@ class Quiz extends React.Component {
   render() {
     return (
       <Flex className="quiz">
+        {/*
+          @TODO: removed the ScrollConcierge because with the lede banner, the scroll would
+          always go past the banner to the content of the quiz. Initially, added so after a
+          potential logic jump with nested quiz, the quiz would scroll back up to see the
+          first question in the next, connected quiz. Need to find a better solution.
+        */}
         <FlexCell width="two-thirds">
           <h1 className="quiz__heading">Quiz</h1>
           <h2 className="quiz__title">{this.props.title}</h2>
--- a/resources/assets/components/Quiz/Quiz.js
+++ b/resources/assets/components/Quiz/Quiz.js
@@ -10,7 +10,6 @@ import QuizQuestion from './QuizQuestion';
 import Share from '../utilities/Share/Share';
 import QuizConclusion from './QuizConclusion';
 import ContentfulEntry from '../ContentfulEntry';
-import ScrollConcierge from '../ScrollConcierge';
 import { calculateResult, resultParams, appendResultParams } from './helpers';
 
 import './quiz.scss';
@@ -194,8 +193,6 @@ class Quiz extends React.Component {
   render() {
     return (
       <Flex className="quiz">
-        <ScrollConcierge />
-
         <FlexCell width="two-thirds">
           <h1 className="quiz__heading">Quiz</h1>
           <h2 className="quiz__title">{this.props.title}</h2>
@@ -211,7 +208,7 @@ Quiz.propTypes = {
   autoSubmit: PropTypes.bool.isRequired,
   additionalContent: PropTypes.shape({
     callToAction: PropTypes.string.isRequired,
-    introduction: PropTypes.string.isRequired,
+    introduction: PropTypes.string,
     submitButtonText: PropTypes.string,
   }).isRequired,
   clickedSignUp: PropTypes.func.isRequired,
@@ -239,6 +236,9 @@ Quiz.propTypes = {
 };
 
 Quiz.defaultProps = {
+  additionalContent: {
+    introduction: null,
+  },
   resultBlocks: null,
   submitButtonText: 'Get Results',
 };
--- a/resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
+++ b/resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import Enclosure from '../../Enclosure';
+import ContentfulEntry from '../../ContentfulEntry';
 import { CallToActionContainer } from '../../CallToAction';
 import CampaignSubPageContent from './CampaignSubPageContent';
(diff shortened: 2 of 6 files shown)


resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
```diff
@@ -15,30 +16,44 @@ import './campaign-subpage.scss';
...
-    <div className="main clearfix">
-      {props.dashboard ? <DashboardContainer /> : null}
-      <CampaignPageNavigationContainer />
-      <Enclosure className="default-container margin-top-lg margin-bottom-lg">
-        <CampaignSubPageContent {...props} />
-      </Enclosure>
-      <CallToActionContainer sticky hideIfSignedUp />
+const CampaignSubPage = props => {
+  // @TODO: after Action page migration, determine a better solution (Page has own gate toggle? etc).
+  const displaySignup = !props.json ? true : false; // eslint-disable-line no-unneeded-ternary
```
mendelB: Why do we need this to be a ternary, if `!props.json` will be a boolan value?

resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
```diff
@@ -15,30 +16,44 @@ import './campaign-subpage.scss';
...
+  );
+};
 
 CampaignSubPage.propTypes = {
   dashboard: PropTypes.shape({
     id: PropTypes.string,
     type: PropTypes.string,
     fields: PropTypes.object,
   }),
+  json: PropTypes.object, // eslint-disable-line react/forbid-prop-types
```
mendelB: Would it be a good idea to call this maybe `fields` or `jsonFields`? I was a bit confused by what this was until I saw the container

resources/assets/components/pages/CampaignSubPage/CampaignSubPageContent.js
```diff
@@ -94,7 +94,7 @@ CampaignSubPageContent.propTypes = {
       fields: PropTypes.shape({
         title: PropTypes.string,
         slug: PropTypes.string,
-        content: PropTypes.string.isRequired,
+        content: PropTypes.string,
```
mendelB: Do you think we should add some sort of safeguard or default for this `content` prop? Since if it's `null`, `Markdown` will still try and render it and break down.

resources/assets/components/Quiz/Quiz.js
```diff
@@ -194,8 +193,6 @@ class Quiz extends React.Component {
   render() {
     return (
       <Flex className="quiz">
-        <ScrollConcierge />
```
mendelB: Should we maybe add a `@TODO` here regarding the scroll issue we discussed?

resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
DFurnes: Yeah, 👍 to the confusion. Was just jumping around trying to figure out where this got defined when I landed on this comment! 😆

resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
DFurnes: Yeah, I'd love a comment with more context here since I'm a little confused by it too.

DFurnes: Also, if we need to convert to Boolean, we can do so with `Boolean(props.json)` to silence the linter warning.

resources/assets/components/pages/CampaignSubPage/CampaignSubPageContent.js
DFurnes: Yeah, I'm a little surprised the linter isn't yelling about this one. Shouldn't it need a `defaultProp` if not required?

DFurnes: Oh, maybe because it's inside a `PropTypes.shape`. I guess that's trickier.

weerd: Yeah, this was weird and yelling at me for it being required and being null.

resources/assets/components/Quiz/Quiz.js
weerd: Sure!

resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
DFurnes: Yep, figured that it was object/null but still confused exactly why there are the two cases. Could you add a comment explaining what scenarios this is an object and what ones it's null?

resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
weerd: Haha, I left it called the same thing as what was used `BlockPage.js`, but I can change to be clearer. Didn't like the name anyhoo.

weerd: Eventually, the idea being if we wrap everything into a single component to output Campaign Pages, then it'd just be content searched for within the `pages` prop type, instead of passing it along as a separate prop with JSON data.

resources/assets/components/pages/CampaignSubPage/CampaignSubPageContent.js
weerd: @mendelB added that [safe guard](https://github.com/DoSomething/phoenix-next/pull/977/commits/f971c4c6fab80626ef887c62f834d4b5ace56782#diff-080d8fb3cc0b3d1adf6af43cf5a3da9eR63) just in case.

PR #978, opened 2018-06-13
Updating to reverse boolean using not operator.

weerd:
### What does this PR do?

This PR fixes an error in the logic for the `displaySignup` prop.

### DIFF
--- a/resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
+++ b/resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
@@ -18,7 +18,7 @@ import './campaign-subpage.scss';
  */
 const CampaignSubPage = props => (
   <div>
-    <LedeBannerContainer displaySignup={Boolean(props.entryContent)} />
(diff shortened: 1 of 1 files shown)


resources/assets/components/pages/CampaignSubPage/CampaignSubPage.js
```diff
@@ -18,7 +18,7 @@ import './campaign-subpage.scss';
  */
 const CampaignSubPage = props => (
   <div>
-    <LedeBannerContainer displaySignup={Boolean(props.entryContent)} />
+    <LedeBannerContainer displaySignup={Boolean(!props.entryContent)} />
```
mendelB: Doesn't the `!` operator return a Boolean without the `Boolean()` function, or do we just want to be explicit about it?

weerd: I'd rather be explicit about it. While this is all _temporary_, I think this section has been confusing enough.

PR #979, opened 2018-06-14
Auto scroll positioning for nested quizzes and quiz results

mendelB:
### What does this PR do?

This PR adds ScrollConcierge capability to the Quiz to ensure proper screen positioning for nested quizzes and quiz results

*Imma try and document the train of thought here since this is a bit of a weird scenario:*

The idea is to ensure that users are scrolled to proper screen positioning when moving from one quiz to another (since presumably, they've scrolled down a bit to complete the first quiz and now need to be nudged back to the top (and this needs to happen *only* for a nested quiz, since we don't want to skip over the `LedeBanner` for the initial quiz)).

The catch is that when rendering the quiz results - where, once again, the screen may need to be nudged to the top - the `ScrollConcierge` may already have been set to show if we're on a nested quiz and if so it wouldn't re-render for the results - thus sadly keeping the user scrolled a bit past the result block.

jun-14-2018 11-14-15

😢 

So to prevent this, we set a timer in the constructor to almost immediately reset the `renderScrollConcierge` to `false`, so that once we get to render the results we'll simply be able to toggle it to `true` again and properly re-render the ScrollConcierge.

jun-14-2018 11-16-35

🍺 😄 
 
[Slack](https://dosomething.slack.com/archives/C3ASB4204/p1528983096000473)

### DIFF
--- a/resources/assets/components/Quiz/Quiz.js
+++ b/resources/assets/components/Quiz/Quiz.js
@@ -1,7 +1,7 @@
 /* global window */
 import React from 'react';
 import PropTypes from 'prop-types';
-import { find, every } from 'lodash';
+import { find, every, get } from 'lodash';
 import ReactRouterPropTypes from 'react-router-prop-types';
 
 import { query } from '../../helpers';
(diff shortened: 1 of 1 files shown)


resources/assets/components/Quiz/Quiz.js
```diff
@@ -41,7 +42,12 @@ class Quiz extends React.Component {
         resultBlock,
       },
       showResults,
+      // Show the scroll concierge for nested quizzes
+      renderScrollConcierge: get(props.additionalContent, 'isNestedQuiz'),
     };
+
+    // Quickly hide scroll concierge so that we can re-render it when showing the quiz results
+    setTimeout(() => this.setState({ renderScrollConcierge: false }), 500);
```
weerd: Would like to eventually revisit this and see if there's a better solution. Still feels a little hacky, but ok for now.

PR #980, opened 2018-06-14
Fixing bug when missing match prop and breaking community page.

weerd:
### What does this PR do?

This PR fixes a bug which would break Community pages on Campaigns due to a missing property when setting things up in the `CampaignSubPageContainer`.

Also does not make the end date on a campaign required in the `CampaignSubPageContent` component and defaults the value to `null`. This was showing up as a warning on Grab The Mic which does not have an end date!

### DIFF
--- a/resources/assets/components/pages/CampaignSubPage/CampaignSubPageContainer.js
+++ b/resources/assets/components/pages/CampaignSubPage/CampaignSubPageContainer.js
@@ -1,4 +1,4 @@
-import { get } from 'lodash';
+import { get, has } from 'lodash';
 import { connect } from 'react-redux';
 
 import CampaignSubPage from './CampaignSubPage';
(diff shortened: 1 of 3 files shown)


resources/assets/components/pages/CampaignSubPage/CampaignSubPageContainer.js
```diff
@@ -8,10 +8,14 @@ import { findContentfulEntry } from '../../../helpers';
  * Provide state from the Redux store as props for this component.
  */
 const mapStateToProps = (state, ownProps) => {
-  const { id, slug } = ownProps.match.params;
+  let entryContent = null;
 
-  // @TODO: temporary retrieval of single camapaign page (quiz) based on matched id or slug.
-  const entryContent = findContentfulEntry(state, id || slug);
+  if (has(ownProps, 'match.props', null)) {
```
mendelB: Is this meant to be 'match.params'?

PR #983, opened 2018-06-18
Add Contentful Querying by Legacy Campaign IDs

mendelB:
### What does this PR do?

This PR updates our lovely `CampaignRepository#findByLegacyCampaignIds ` function to query Contentful first in its quest to return campaigns by a list of Legacy IDs.

We'll first ask Contentful for all campaigns it can find using the `legacyCampaignId` field, and only pass along the remaining non-found IDs to Ashes.

This is another step towards a `/index` endpoint for Campaigns, supporting the batched `filter[id]=1,2,3` query.

### Question:
We probably don't want to be returning the fully constructed Contentful `Campaign` entity from this endpoint, correct? Do we perhaps want to create a mini or truncated Campaign entity which we could use for this scenario as well as in the `getAll` method?

#942

### DIFF
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -93,9 +93,9 @@ class CampaignRepository
     {
         $query['ids'] = implode(',', $ids);
 
-        $campaigns = $this->phoenixLegacy->getCampaigns($query);
+        $results = $this->phoenixLegacy->getCampaigns($query);
 
-        $campaigns = collect($campaigns['data'])->map(function ($campaign) {
+        $campaigns = collect($results['data'])->map(function ($campaign) {
             return new LegacyCampaign($campaign);
         });
     }
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -133,19 +133,30 @@ class CampaignRepository
 
     /**
      * Find a list of campaigns by their legacy IDs.
+     * Attempts to fetch campaigns from Contentful and falls back to Phoenix Ashes.
      *
      * @param  array $ids
      * @return \Illuminate\Support\Collection
      */
     public function findByLegacyCampaignIds($ids)
     {
-        $query['ids'] = implode(',', $ids);
+        $query = (new Query)
+            ->setContentType('campaign')
+            ->where('fields.legacyCampaignId', $ids, 'in')
+            ->setInclude(1);
 
-        $legacyCampaigns = $this->phoenixLegacy->getCampaigns($query);
+        $campaignsArray = iterator_to_array($this->contentful->getEntries($query));
 
-        return collect($legacyCampaigns['data'])->map(function ($legacyCampaign) {
-            return new LegacyCampaign($legacyCampaign);
+        $contentfulCampaigns = collect($campaignsArray)->map(function ($campaign) {
+            return new Campaign($campaign);
         });
+
+        // List of IDs returned from Contentful.
+        $foundIds = $contentfulCampaigns->pluck('legacyCampaignId')->all();
+        // Query from Phoenix Ashes using remaining IDs.
+        $legacyCampaigns = $this->getLegacyCampaigns(array_diff($ids, $foundIds));
+
+        return $contentfulCampaigns->merge($legacyCampaigns);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


app/Repositories/CampaignRepository.php
```diff
@@ -116,19 +133,30 @@ public function findByLegacyCampaignId($id)
...
     public function findByLegacyCampaignIds($ids)
     {
-        $query['ids'] = implode(',', $ids);
+        $query = (new Query)
+            ->setContentType('campaign')
+            ->where('fields.legacyCampaignId', $ids, 'in')
+            ->setInclude(1);
 
-        $legacyCampaigns = $this->phoenixLegacy->getCampaigns($query);
+        $campaignsArray = iterator_to_array($this->contentful->getEntries($query));
```
weerd: What's the need for `iterator_to_array` here?

mendelB: the return value of `getEntries` is a `ResourceArray` which made it really awkward to be able to iterate or grab the entries from it. I got the idea from [`getAll()`](https://github.com/DoSomething/phoenix-next/blob/master/app/Repositories/CampaignRepository.php#L47)

DFurnes: Hmm, can you try directly passing the iterator to `collect()`? It looks like the constructor is able [to parse](https://github.com/laravel/framework/blob/8a74cf931b9df281707cef973ee9fba5b18793e8/src/Illuminate/Support/Collection.php#L1603-L1626) a bunch of different types of input, including iterators.

mendelB: I did try that but ended up with this error:
`array_keys() expects parameter 1 to be array, object given`

(And just confirming that you mean passing the returned result directly to `collect`?
```php
$campaignsArray = $this->contentful->getEntries($query);

$contentfulCampaigns = collect($campaignsArray)->map(function ($campaign) {
  return new Campaign($campaign);
});
```
)

mendelB: I think that error might be from later on as the `collect($campaignsArray)` itself *does* work. It's just that the results are trapped in a nested `items` field:
```
Collection {#797 ▼
  #items: {#799 ▼
    +"sys": {#798 ▶}
    +"total": 2
    +"limit": 100
    +"skip": 0
    +"items": array:2 [▼
      0 => DynamicEntry {#534 ▶}
      1 => DynamicEntry {#795 ▶}
    ]
  }
}
```

DFurnes: In that dump, `#items` is the internal field on the collection instance that stores the values. You should be able to then call `->toArray()` on the collection to export them as a plain array.

DFurnes: Oh oh, wait I see - the _internal_ items key there. Sorry! I'm surprised that gets properly handled when using `iterator_to_array` the other way, though... strange.

mendelB: Replacing `iterator_to_array ` with `collect($campaignsArray->getIterator())` seems to work as well if you think that's more elegant?

DFurnes: Ahh, that's so interesting. I guess otherwise it picks up on [this case](https://github.com/laravel/framework/blob/8a74cf931b9df281707cef973ee9fba5b18793e8/src/Illuminate/Support/Collection.php#L1619-L1620) and passes the [JSON](https://github.com/contentful/contentful-core.php/blob/3edd8f0cfb8d1719c4faa1d226e7e83833afd723/src/Resource/ResourceArray.php#L95-L109) to the collection. Bleh! Maybe just `collect($campaignsArray->getItems())`?

DFurnes: That actually makes me think we should rename `$campaignsArray` since it's not an array (yet), or inline the `->getItems()` into the query itself like this:

```php
$campaignsArray = $this->contentful->getEntries($query)->getItems();
$contentfulCampaigns = collect($campaignsArray)->map(function ($campaign) {
  return new Campaign($campaign);
});
```

mendelB: Ah if only 😩 
`Call to undefined method Contentful\ResourceArray::getItems()`

mendelB: `dd(get_class_methods($campaignsArray));`

```
array:11 [▼
  0 => "__construct"
  1 => "getTotal"
  2 => "getLimit"
  3 => "getSkip"
  4 => "jsonSerialize"
  5 => "count"
  6 => "getIterator"
  7 => "offsetExists"
  8 => "offsetGet"
  9 => "offsetSet"
  10 => "offsetUnset"
]
```

DFurnes: RIP, `getIterator` it is! Looks like we're a few releases behind on the Contentful SDK.

PR #985, opened 2018-06-19
Add method to find Campaigns from Contentful by IDs

mendelB:
### What does this PR do?

This PR adds a new method to the `CampaignRepository` to batch query campaigns from Contentful by a list of IDs

**Q**:
I wasn't sure whether to name this `findByContentfulIds` or if we wanted this to be implied as the de-facto primary ID by calling it `findByIds`?

#942

### DIFF
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -160,6 +160,26 @@ class CampaignRepository
     }
 
     /**
+     * Find a list of campaigns by their IDs.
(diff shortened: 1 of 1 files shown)


DFurnes:
> I wasn't sure whether to name this `findByContentfulIds` or if we wanted this to be implied as the de-facto primary ID by calling it `findByIds`?

I think calling this `findByIds` seems reasonable to me.

PR #988, opened 2018-06-19
Remove old submission gallery related code

weerd:
### What does this PR do?

This PR removes code related to the old `ComponentActionStep` with a type of `submission-gallery` and the associated `SubmissionGallery` component.

Also adds in the submission gallery using the new GraphQL related component along with the new PSA, which I hadn't realized was actually missing! 😬 

Refs [Pivotal ID #158465366](https://www.pivotaltracker.com/story/show/158465366)

### Checklist


resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -55,7 +54,7 @@ class ContentfulEntry extends React.Component<Props, State> {
     }
 
     // Otherwise, find the corresponding component & render it!
-    const { json = DEFAULT_BLOCK, stepIndex = 1, isSignedUp } = this.props;
```
DFurnes: Does this mean we can kill the `isSignedUp` prop on this component too? 😺

weerd: Yeah, probs. I was a bit weary to remove in case it _might_ impact something, but I don't think anything else is using it.

PR #989, opened 2018-06-19
Updated Campaigns Index Action

mendelB:
### What does this PR do?

This PR updates the `Campaigns#index` action to process a `filter[id]=1,2,3` query.

This is a bit of a hefty action. I imagine that once we bring this action fully up to speed (full index support, pagination etc.) we can get to clean this up some more 🙏

#942

### DIFF
--- a/app/Http/Controllers/Api/CampaignsController.php
+++ b/app/Http/Controllers/Api/CampaignsController.php
@@ -51,9 +51,9 @@ class CampaignsController extends Controller
         // All remaining IDs are presumed to be Contentful IDs.
         $contentfulIds = array_diff($idsArray, $legacyIds);
 
-        $legacyCampaigns = count($legacyIds) ? $this->campaignRepository->findByLegacyCampaignIds($legacyIds) : collect();
+        $legacyCampaigns = $this->campaignRepository->findByLegacyCampaignIds($legacyIds);
 
-        $contentfulCampaigns = count($contentfulIds) ? $this->campaignRepository->findByIds($contentfulIds) : collect();
+        $contentfulCampaigns = $this->campaignRepository->findByIds($contentfulIds);
 
         $campaigns = $contentfulCampaigns->merge($legacyCampaigns);
 
--- a/app/Http/Controllers/Api/CampaignsController.php
+++ b/app/Http/Controllers/Api/CampaignsController.php
@@ -40,10 +40,11 @@ class CampaignsController extends Controller
 
         $idsArray = explode(',', $ids);
 
-        // We limit filter[id] requests to a maximum of 10 IDs.
-        if (count($idsArray) > 10) {
-            return response()->json('Exceeded limit of 10 IDs', 422);
-        }
+        $request->query('filter')['id'] = $idsArray;
+
+        $this->validate($request, [
+            'filter.id' => 'max:10'
+        ]);
 
         // Extract the legacy IDs.
         $legacyIds = array_filter($idsArray, 'is_legacy_id');
--- a/app/Http/Controllers/Api/CampaignsController.php
+++ b/app/Http/Controllers/Api/CampaignsController.php
@@ -2,6 +2,7 @@
 
 namespace App\Http\Controllers\Api;
 
+use Illuminate\Http\Request;
 use App\Http\Controllers\Controller;
 use App\Repositories\CampaignRepository;
 
@@ -29,11 +30,38 @@ class CampaignsController extends Controller
      *
      * @return \Illuminate\Http\JsonResponse
      */
-    public function index()
+    public function index(Request $request)
     {
-        $data = $this->campaignRepository->getAll();
+        $this->validate($request, [
+            // We're only supporting requests with an ID filter query.
+            'filter.id' => 'required',
+        ]);
 
-        return response()->json(['data' => $data]);
+        $ids = $request->query('filter')['id'];
+
+        $idsArray = explode(',', $ids);
+
+        // We limit filter[id] requests to a maximum of 10 IDs.
+        if (count($idsArray) > 10) {
+            // Not sure that this is the right thing to do here:
+            return response()->json('Exceeded limit of 10 IDs', 422);
+        }
+
+        // Extract the legacy IDs.
+        $legacyIds = collect($idsArray)->filter(function($id) {
+            return is_legacy_id($id);
+        })->all();
+
+        // All remaining IDs are presumed to be Contentful IDs.
+        $contentfulIds = array_diff($idsArray, $legacyIds);
+
+        $contentfulCampaigns = $this->campaignRepository->findByLegacyCampaignIds($legacyIds);
+
+        $legacyCampaigns = $this->campaignRepository->findByIds($contentfulIds);
+
+        $campaigns = $contentfulCampaigns->merge($legacyCampaigns);
+
+        return response()->json(['data' => $campaigns]);
     }
 
     /**
--- a/app/Http/Controllers/Api/CampaignsController.php
+++ b/app/Http/Controllers/Api/CampaignsController.php
@@ -34,7 +34,7 @@ class CampaignsController extends Controller
     {
         $ids = array_get($request->query('filter'), 'id');
 
-        if (!$ids) {
+        if (! $ids) {
             return response()->json(['data' => []]);
         }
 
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Api/CampaignsController.php
```diff
@@ -29,11 +30,38 @@ public function __construct(CampaignRepository $campaignRepository)
...
      * @return \Illuminate\Http\JsonResponse
      */
-    public function index()
+    public function index(Request $request)
     {
-        $data = $this->campaignRepository->getAll();
+        $this->validate($request, [
+            // We're only supporting requests with an ID filter query.
+            'filter.id' => 'required',
+        ]);
```
mendelB: This works fine for AJAX requests. For regular HTTP requests, this redirects to `us/campaigns` upon failure

app/Http/Controllers/Api/CampaignsController.php
```diff
@@ -29,11 +30,38 @@ public function __construct(CampaignRepository $campaignRepository)
...
+        $legacyIds = collect($idsArray)->filter(function($id) {
+            return is_legacy_id($id);
+        })->all();
+
+        // All remaining IDs are presumed to be Contentful IDs.
+        $contentfulIds = array_diff($idsArray, $legacyIds);
+
+        $contentfulCampaigns = $this->campaignRepository->findByLegacyCampaignIds($legacyIds);
+
+        $legacyCampaigns = $this->campaignRepository->findByIds($contentfulIds);
```
mendelB: @weerd My question about empty ID arrays was referring to this and the above method calls. Suppose all IDs in the request were legacy IDs, we'd then be calling this with an empty array, and then `findByIds` would be querying Contentful for no reason.

So I was wondering if it'd be our job here to ensure `count($contentfulIds/legacyIds)` before calling these methods, or should those methods themselves make that determination before querying?

app/Http/Controllers/Api/CampaignsController.php
```diff
@@ -29,11 +30,38 @@ public function __construct(CampaignRepository $campaignRepository)
...
+        })->all();
+
+        // All remaining IDs are presumed to be Contentful IDs.
+        $contentfulIds = array_diff($idsArray, $legacyIds);
+
+        $legacyCampaigns = $this->campaignRepository->findByLegacyCampaignIds($legacyIds);
+
+        $contentfulCampaigns = $this->campaignRepository->findByIds($contentfulIds);
+
+        $campaigns = $contentfulCampaigns->merge($legacyCampaigns);
```
mendelB: @weerd My question about empty ID arrays was referring to this and the above method calls. Suppose all IDs in the request were legacy IDs, we'd then be calling this with an empty array, and then `findByIds` would be querying Contentful for no reason.

So I was wondering if it'd be our job here to ensure `count($contentfulIds/legacyIds)` before calling these methods, or should those methods themselves make that determination before querying?

app/Http/Controllers/Api/CampaignsController.php
weerd: It seems odd to see the validator not used in a method related to a POST request, but I think it technically should be fine 🤔

mendelB: Is it like an anti-pattern, or just out of the ordinary?
I guess it's a weird case here that we're like only allowing requests with a certain filter through?

weerd: oh actually, we could just remove this...

If `filter.id` is not provided, than we should not have any IDs and thus `count` is 0 and we just return an empty array.

mendelB: Ah true!! Why bother validating

mendelB: But that does beg this question as well
https://github.com/DoSomething/phoenix-next/pull/989#discussion_r196583168

mendelB: ...Or is it no biggie if we make those blank requests, than no sweat!

weerd: looking into that now. Ideally with nothing there we don't make those requests.

app/Http/Controllers/Api/CampaignsController.php
weerd: Maybe something like:

```php
$legacyCampaigns = count($legacyIds) ? $this->campaignRepository->findByLegacyCampaignIds($legacyIds) : [];

$contentfulCampaigns = count($contentfulIds) ? $this->campaignRepository->findByIds($contentfulIds) : [];

 $campaigns = $contentfulCampaigns->merge($legacyCampaigns);
```
If not ids for either, it's just an empty array, and when merged should be all good.

weerd: We can probs after refactor and clean this up a bit once it's all working :)

mendelB: Makes sense! I'll get this in.

app/Http/Controllers/Api/CampaignsController.php
```diff
@@ -29,11 +30,36 @@ public function __construct(CampaignRepository $campaignRepository)
...
-        return response()->json(['data' => $data]);
+        if (! $ids) {
+            return response()->json(['data' => []]);
+        }
+
+        $idsArray = explode(',', $ids);
+
+        // We limit filter[id] requests to a maximum of 10 IDs.
+        if (count($idsArray) > 10) {
+            return response()->json('Exceeded limit of 10 IDs', 422);
```
mendelB: Should this be returning some sort of object, or raising some sort of error, or are we fine with returning a simple JSON string?

app/Http/Controllers/Api/CampaignsController.php
```diff
@@ -29,11 +30,36 @@ public function __construct(CampaignRepository $campaignRepository)
...
+        }
+
+        // Extract the legacy IDs.
+        $legacyIds = collect($idsArray)->filter(function ($id) {
```
DFurnes: This could be simplified to `array_filter($idsArray, 'is_legacy_id')` if you want to save some space.

app/Http/Controllers/Api/CampaignsController.php
```diff
@@ -29,11 +30,36 @@ public function __construct(CampaignRepository $campaignRepository)
...
+            return is_legacy_id($id);
+        })->all();
+
+        // All remaining IDs are presumed to be Contentful IDs.
+        $contentfulIds = array_diff($idsArray, $legacyIds);
+
+        $legacyCampaigns = count($legacyIds) ? $this->campaignRepository->findByLegacyCampaignIds($legacyIds) : collect();
```
DFurnes: I think it might be a good idea to pull this conditional into `findByLegacyCampaignIds`, so we can always call the function here and the repository knows it doesn't need to fire off an API request if the array is empty.

app/Http/Controllers/Api/CampaignsController.php
weerd: What does this look like as a response? Does Laravel's handler take over and provide an API response with related error structure?

DFurnes: Let's use a validator for this, e.g:

```php
validator($request->query, ['ids' => 'required|min:1'])->validate();
```

weerd: It's my bad. Initially @mendelB had a validator in here and I recommended removing it. I think lets keep it with a max at `10`, and just return empty array if no ids in filter provided.

mendelB:
Updated based on feedback most notably:

- Using a new `TruncatedCampaign` Entity for constructing the returned campaign entries. (I took the liberty of using this in the `getAll` function as well) I based the fields to keep off of what we had in `LegacyCampaign`
- Checking and returning for empty ID arrays in the `CampaignRepository` instead of the controller
- Converting the `filter[id]` to an array so that we could use the native validation. (I *think* this is what y'all were recommending to do?)

app/Http/Controllers/Api/CampaignsController.php
```diff
@@ -29,11 +30,36 @@ public function __construct(CampaignRepository $campaignRepository)
...
      * @return \Illuminate\Http\JsonResponse
      */
-    public function index()
+    public function index(Request $request)
     {
-        $data = $this->campaignRepository->getAll();
+        $ids = array_get($request->query('filter'), 'id');
 
-        return response()->json(['data' => $data]);
+        if (! $ids) {
```
DFurnes: Now that `findByLegacyCampaignIds` and `findByIds` can handle empty arrays without making a request, can we remove this separate empty response?

DFurnes:
> Converting the `filter[id]` to an array so that we could use the native validation. (I *think* this is what y'all were recommending to do?)

Yeah, I'm glad it worked! It could be fun to see if this would make sense as a `parseQueryString()` helper that iterates over any values and parses them as CSVs, but this is totally cool to ship IMO.

app/Http/Controllers/Api/CampaignsController.php
mendelB: The only trouble here was that `explode` returns an array with an empty string element even when `$ids` was `null`, so I just added a ternary to address that scenario.

mendelB:
@DFurnes I guess I'll ship as is for now if that's cool, since I'm not sure I completely understand how that helper would work. Would love to tackle adding that in eventually!

DFurnes:
Yeah, ship away! ⛵️

PR #990, opened 2018-06-21
Fixes to the Campaign Index action

mendelB:
### What does this PR do?

This PR fixes a couple of silly mistakes I made in the zero hour of #989 

- Fixed an improperly defined ternary to ensure we weren't exploding an empty string
- Properly reset the `filter[id]` query parameter

🤦‍♂️

### DIFF
--- a/app/Http/Controllers/Api/CampaignsController.php
+++ b/app/Http/Controllers/Api/CampaignsController.php
@@ -34,7 +34,7 @@ class CampaignsController extends Controller
     {
         $ids = array_get($request->query('filter'), 'id');
 
-        $idsArray = count($ids) ? explode(',', $ids) : [];
+        $idsArray = $ids ? explode(',', $ids) : [];
 
         // Resetting the filter[id] param value to an array so that we can properly validate.
         $request->query('filter')['id'] = $idsArray;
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Api/CampaignsController.php
```diff
@@ -34,10 +34,10 @@ public function index(Request $request)
     {
         $ids = array_get($request->query('filter'), 'id');
 
-        $idsArray = count($ids) ? explode(',', $ids) : [];
+        $idsArray = $ids ? explode(',', $ids) : [];
 
         // Resetting the filter[id] param value to an array so that we can properly validate.
-        $request->query('filter')['id'] = $idsArray;
+        $request->request->add(['filter' => ['id' => $idsArray]]);
```
DFurnes: `$request->request`! Awkward!

mendelB: ikr?? I was hoping for something more elegant but this was the only working method I could find out there.

app/Http/Controllers/Api/CampaignsController.php
```diff
@@ -34,10 +34,10 @@ public function index(Request $request)
     {
         $ids = array_get($request->query('filter'), 'id');
```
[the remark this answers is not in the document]
mendelB: Exactly. I felt like it wasn't worth a default arg either since it's any way easier to just check for `$ids` truthiness.

PR #993, opened 2018-06-21
Attach query params to the nested quiz path

mendelB:
### What does this PR do?

This PR ensures we retain query parameters when redirecting a user to a nested Quiz

This is specifically important for SMS quizzes, where we'd like to keep the quiz un-gated while retaining User identity through the `user_id=` query parameter.

Refs [Pivotal ID #157751760](https://www.pivotaltracker.com/story/show/157751760)

### DIFF
--- a/resources/assets/components/Quiz/Quiz.js
+++ b/resources/assets/components/Quiz/Quiz.js
@@ -120,10 +120,11 @@ class Quiz extends React.Component {
       const resultBlockSlug = resultBlock.fields.slug;
 
       // Retain the current pathname while replacing the active quiz's slug with the resultBlocks slug
-      const newPath = location.pathname.replace(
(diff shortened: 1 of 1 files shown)


mendelB:
Hmm that's a good point. I'm seeing it actually persist now! When I was testing with Ashley it wasn't showing up, but I suspect now that we may have been looking at the wrong Keen.io event stream. Do you think it's a good idea anyway for us to attach the `search` or is there no use?

DFurnes:
Unless we're reading that query string from the URL for analytics (which we don't really want to be doing), it won't do anything to pass along. I'd vote against so we don't confuse ourselves.

mendelB:
Fair indeed! I'll close this poor guy out.

PR #994, opened 2018-06-27
Syncing env files

weerd:
### What does this PR do?

This PR adds a new composer package that helps to make sure our `.env` files are properly synced within our team. It follows the awesome blog post [Dot Env Files: The Silent Killer](https://tighten.co/blog/dot-env-files-the-silent-killer) by @calebporzio that shows how to use this package along with some composer scripts and git hooks to make sure we don't forget to update `.env` and `.env.example` files when environment variables are added or removed.

**NOTE: Make sure to run the following command in your repository directory!**
```
$ git config core.hooksPath githooks
```

### Checklist


githooks/post-checkout.sh
```diff
@@ -0,0 +1,13 @@
...
+# This hook is called with the following parameters:
+#
+# $3 - A flag indicating whether the checkout was a branch or file checkout
+
+isFileCheckout=0
+isBranchCheckout=1
+
+if [ $3 -eq $isBranchCheckout ]
+then
+    composer install
```
mendelB: Would this hook be running even if the branch checkout was done outside of the Homestead environment? 
Also, do you see it becoming possibly annoying to have to `composer install` for every branch checkout (personally -- sometimes I'll be doing rapid checkouts between branches...), or do you think it's a worthy sacrifice anyway?

weerd: So many questions!

weerd: I think the `chmod 777` is set on the file so you shouldn't _have_ to do it as well, but something we can see how it works out.

weerd: > Would this hook be running even if the branch checkout was done outside of the Homestead environment?

Not sure, and something I was curious about as well and figured we'd see how it shook out as we used it.

weerd: > Also, do you see it becoming possibly annoying to have to composer install for every branch checkout (personally -- sometimes I'll be doing rapid checkouts between branches...), or do you think it's a worthy sacrifice anyway?

Might be annoying, but it's much more annoying to run into trippy bugs because we're missing an environment variable. They're are the dumbest kind of headache!

mendelB: I feel ya. And remembering the Contentful API Key Mishap feel inclined to agree!

weerd:
Ran into an issue on Heroku (but could be just a general build step error not specific to Heroku), where the `.env` file was missing by the time the `post-install-cmd` runs for the `php artisan env:check` command. It errors out, since the file is missing to be able to run a check on it. Attempting to see if adding the `.env` creation command (which runs in `post-root-package-install` script as well) prior to the `env:check` command helps out 🤔

weerd:
This also doesn't necessarily resolve making sure our env variables are added on our Heroku Prod and Staging environments since the whole process on Heroku is a bit different 🤔 But this does help ensure that environments are in sync among the team during development.

weerd:
I'm going to close this PR for now. It's been sitting open for a while. I'll come back to it eventually and try to implement, but right now too much time has passed and it's just collecting dust.

PR #995, opened 2018-06-28
Remove activityFeed related fields and props

mendelB:
### What does this PR do?

This PR removes the deprecated `activityFeed` field and props from our application

Since we've moved over to 'Community Page's, the Campaign `activityFeed` is deprecated and no longer in use so:

- removed the `Campaign` Entity `activityFeed` field, and `parseActivityField` method
- removed the defunct `fillReportbackPosts ` method
- trashed the `activityFeed` props and logic from `CampaignPage`

#982

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -20,90 +20,6 @@ use Contentful\Delivery\Asset;
 class Campaign extends Entity implements JsonSerializable
 {
     /**
-     * Fill with specified number of Reportback Post items.
-     *
-     * @param  \Contentful\Delivery\DynamicEntry $entry
-     * @return array
-     *
-     * @todo Temporary reportback post item filler based on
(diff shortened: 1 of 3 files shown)


app/Entities/Campaign.php
```diff
@@ -190,11 +106,6 @@ public function jsonSerialize()
             'campaignLead' => $this->parseCampaignLead($this->campaignLead, $this->additionalContent),
             'affiliateSponsors' => $this->parseAffiliates($this->affiliateSponsors),
             'affiliatePartners' => $this->parseAffiliates($this->affiliatePartners),
-            // @TODO: Why is it 'activity_feed' oy? ;/
```
weerd: The fabled comment from @DFurnes has finally had an opportunity to be addressed!

mendelB: This was my fave. RIP 😢

PR #1000, opened 2018-06-29
Action Steps and  Activity Feed cleanup

mendelB:
### What does this PR do?

This PR removes an assortment of fields, logic, and React props related to the now deprecated `actionSteps` and `activityFeed` Campaign fields.

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -311,7 +311,7 @@ function get_campaign_social_fields($campaign, $uri)
 
     if (str_contains($uri, [$blockPath, $modalPath])) {
         $blockId = last(explode('/', $uri));
-
+        // @TODO with the activityFeed field now deprecated, we need to re-visit this.
         $block = array_first($campaign->activityFeed, function ($value) use ($blockId) {
             return $value->id === $blockId;
         });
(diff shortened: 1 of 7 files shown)


app/helpers.php
```diff
@@ -311,7 +311,7 @@ function get_campaign_social_fields($campaign, $uri)
 
     if (str_contains($uri, [$blockPath, $modalPath])) {
         $blockId = last(explode('/', $uri));
-
+        // @TODO with the activityFeed field now deprecated, we need to re-visit this.
```
DFurnes: Yeah, it'd be nice to have a general `get_social_metadata()` helper that we pass a Contentful Entry to based on the route. It feels awkward to me that we have routing logic inside a random helper method.

DFurnes: It'd be nice to revisit our data fetching logic next quarter, since the whole "fetch all the things and then iterate through them" approach definitely doesn't scale. 😆

mendelB: Ah so you're saying, let's just fetch only the specific entry needed, to begin with, based on the route, and then simply extract its social metadata?

DFurnes: Yep!

PR #1001, opened 2018-06-29
Adding in script for Google Tag Manager.

weerd:
### What does this PR do?

This PR adds the Google Tag Manager script for ad tracking. It can be toggled by adding the Google Tag Manager ID as an environment variable; if ID is not present as env variable then the script is not included. We only really want to add the ID to an env variable on the production environment.

I ended up removing the `<noscript>` fallbacks in both the Facebook Pixel script and the Google Tag Manager Script because it's not really useful in a client side JS rendered site. It is something we can add back in if/when we switch to a server rendering react components.

Refs [Pivotal ID #158462761](https://www.pivotaltracker.com/story/show/158462761)

### Checklist


resources/views/partials/facebook_pixel_script.blade.php
```diff
@@ -1,14 +1,14 @@
...
-  {if(f.fbq)return;n=f.fbq=function(){n.callMethod?
-  n.callMethod.apply(n,arguments):n.queue.push(arguments)};
-  if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0';
-  n.queue=[];t=b.createElement(e);t.async=!0;
-  t.src=v;s=b.getElementsByTagName(e)[0];
-  s.parentNode.insertBefore(t,s)}(window, document,'script',
-  'https://connect.facebook.net/en_US/fbevents.js');
-  fbq('init', '{{config('services.analytics.facebook_pixel_id')}}');
-  fbq('track', 'PageView');
+    !function(f,b,e,v,n,t,s)
```
DFurnes: ⏩

resources/views/partials/facebook_pixel_script.blade.php
```diff
@@ -1,14 +1,14 @@
...
+    if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0';
+    n.queue=[];t=b.createElement(e);t.async=!0;
+    t.src=v;s=b.getElementsByTagName(e)[0];
+    s.parentNode.insertBefore(t,s)}(window, document,'script',
+    'https://connect.facebook.net/en_US/fbevents.js');
+    fbq('init', '{{config('services.analytics.facebook_pixel_id')}}');
+    fbq('track', 'PageView');
 </script>
-<noscript><img height="1" width="1" style="display:none" src="https://www.facebook.com/tr?id={{config('services.analytics.facebook_pixel_id')}}&ev=PageView&noscript=1" /></noscript>
+<?php // When/if we implement server-side rendering of React, we should add the noscript code back in! ?>
```
DFurnes: Wanna make this a Blade comment? `{{-- When/if we implement server-side... --}}`

weerd: haha, doh!

PR #1005, opened 2018-07-02
CampaignPage component portation

mendelB:
### What does this PR do?

This PR moves over everything from the `CampaignPage` component into its own `CampaignRoute` component and renames all `CampaignSubPage` paraphernalia to `CampaignPage`

No real changes here, this is only structural. Next step will be some refactoring to hopefully clean these up a bit.

The idea is to pull all Campaign routing related logic to an aptly named `CampaignRoute` component and to be rid of the idea of a Campaign Sub Page, now that all campaign pages are indeed Pages!


resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
```diff
@@ -0,0 +1,120 @@
...
+import { Switch, Route, Redirect } from 'react-router-dom';
+
+import Modal from '../../utilities/Modal/Modal';
+import CampaignFooter from '../../CampaignFooter';
+import BlockPageContainer from '../../pages/BlockPage/BlockPageContainer';
+import ActionPageContainer from '../../pages/ActionPage/ActionPageContainer';
+import CampaignPageContainer from '../../pages/CampaignPage/CampaignPageContainer';
+import PostSignupModalContainer from '../../pages/PostSignupModal/PostSignupModalContainer';
+
+const CampaignPage = props => {
```
DFurnes: Should we rename this to `const CampaignRoute`?

PR #1007, opened 2018-07-02
Campaign Page Content Changes

mendelB:
### What does this PR do?

This PR updates the `CampaignPageContent` component rendering logic to enable rendering Campaign Pages using the 'everything is a page' approach.

Also, removed the `/community` route, which means we do lose the redirect when there's no community page existant.

Finally, removed the deprecated `/pages/:slug` route.

Previously, we were only allowing `/community` pages to be rendered using the `blocks` field, with everything else defaulting to the old 'Campaign Sub Page' approach using the `content` field.

This would allow any Page rendered to use the `blocks` field, ~and will fallback to using the `content` field if it's filled out.~

This also eliminates the formatting of the old Campaign Sub Pages, to skew towards rendering things as simple blocks, so that what used to be:

image

Will now be:
 
image

A cta can be added in the `sidebar` field which could look something like:

image

~Eliminating the CTA sidebar, and rendering the content in a simple `Markdown` component. (While keeping the bottom CTA for all pages. (*should we remove that as well?*))~

### Some questions that came to mind are:
- Should we just remove the bottom CTA for good?
- Do we actually want the **sidebar** CTA, or more generally -- do we want to be rendering the page sidebar block/s for all campaign pages, allowing a primary / secondary visual layout as it used to be. (@weerd I recall you mentioning a while ago that this was part of the Page vision)
- Do we want to get a management API script going to update all Campaign Pages and pull their Content into a ContentBlock in the `blocks` field? (We can then remove the pesky logic here, and better clarify the matter for editors. (This does also raise the question of how to apply this for landing pages.))
- I dunno why I'm vomiting all of these thoughts here. Maybe a Discussion or something would be better? I'm sure these are things you've considered already, it just hit me more succinctly when working on this PR.


weerd:
> Do we actually want the sidebar CTA, or more generally -- do we want to be rendering the page sidebar block/s for all campaign pages, allowing a primary / secondary visual layout as it used to be. (@weerd I recall you mentioning a while ago that this was part of the Page vision)

image

We do have a field for a simple siderbar on the `Page` content type, so could we use that?

mendelB:
@weerd Would it make sense to add the sidebar functionality to Campaign Pages akin to the Pitch Page? 
(I'm imagining applying the `primary`, `secondary` styling rules *if* sidebar content is available, and if so, rendering the sidebar blocks in `secondary`, if that sounds correct?)

weerd:
Yeah, I'm thinking of it as more of just a general page thing, given that the sidebar is a field available on all pages!

mendelB:
**Post Script:**

@weerd Ran through the vision for Pages.
Roughly speaking, here's how they'll work:

- **Content** will be rendered as a `primary` div spread across two thirds of the page.
- **Sidebar** blocks will be rendered in a `secondary` portion across the remaining third (There can be multilpe sidebar blocks)
- **Blocks** will be rendered underneath the above section. 




*Edge Cases*:
- If there's no **Content** provided, we'll ignore the **Sidebar** blocks.
- If there's no **Sidebar** but there is **Content**, we'll still keep the **Content** limited to two thirds across the page. (Tentative. Pending a @lkpttn consultation)

We'd like to eventually utilize this flow across all pages - *General Page*, *Campaign Page*, *Pitch Page*, (perhaps by creating a `PageContent` type that implements this strategy.) 
For now we'll implement this across all pages, and consolidate when applicable.

See https://github.com/orgs/DoSomething/teams/team-rocket/discussions/8 for a broader discussion of the above.

weerd:
Yep yep @mendelB.

Plus, I think we could utilize this structure for all rows; so a div that is the row, and then a `primary` and `secondary` items in each row. I think that could work as a general pattern :)

mendelB:
Updated to new format as explained above!

DFurnes:
This is so nice and simple! 🙌

ashleybaldwin:
I love this!! This is going to make everything so easy on the admin side.

PR #1012, opened 2018-07-03
Removing total signups output in Dashboard and related code.

weerd:
### What does this PR do?

This PR removes the total signups output from the Dashboard (which wasn't really used anyhoo), and gets rid of a lot of the related code for getting the total signups.

To get the total signups for a campaign to potentially showcase in the Dashboard, we make a call to Ashes for the information. Turns out we make the call _even if we're not showing the total signups_ in the Dashboard, which is not idea. Plus the call has started failing for whatever reason, responding with `500` server error codes. Since we can't currently get the total signups for a campaign from Rogue directly, we decided it best to just remove this and any associated code given that it's not really being used.

The error I was seeing:
screen shot 2018-07-03 at 2 43 48 pm

🌵 

### Checklist

### DIFF
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -33,18 +33,6 @@ class SignupController extends Controller
     }
 
     /**
-     * Get the total signups for this campaign.
-     *
-     * @param  int $campaignId
-     * @return \Illuminate\Http\Response
-     */
-    public function total($campaignId)
-    {
-        // Count = 1 is a hack to make the API actually return something. Don't ask.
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/SignupController.php
```diff
@@ -32,18 +32,6 @@ public function index(Request $request)
...
 
-    /**
-     * Get the total signups for this campaign.
-     *
-     * @param  int $campaignId
-     * @return \Illuminate\Http\Response
-     */
-    public function total($campaignId)
-    {
-        // Count = 1 is a hack to make the API actually return something. Don't ask.
```
mendelB: > Don't ask.
k 😆

PR #1015, opened 2018-07-05
GitBook v2 related updates!

weerd:
### What does this PR do?

This PR performs a few updates to get the repository updated to work with GitBook v2!

...

Refs [Pivotal ID #]()


docs/content-publishing/community/campaign-update/affiliate-update.md
```diff
@@ -0,0 +1,24 @@
+# Affiliate Update
```
mendelB: Is this page a copy of the [`campaign-update/README.md](https://github.com/DoSomething/phoenix-next/pull/1015/files#diff-5b57a288e1c6d570dd6d3d56a8a5063f) above?

weerd: Yar. A quirk with the new GitBooks. Looks like when it migrated, if a summary item referenced a hashtag in a file, then it would extract that into it's own separate file. Had to clean that all up, and evidently missed one.

PR #1016, opened 2018-07-09
Puck updates - Adding isAuthenticated field.

mendelB:
### What does this PR do?

This PR updates to the new puck version `1.5.0` and adds the `isAuthenticated` prop to the `PuckProvider`

This gives us the new `user.isAuthenticated` field!

Refs [Pivotal ID #158838971](https://www.pivotaltracker.com/story/show/158838971)

### DIFF
--- a/resources/assets/components/App.js
+++ b/resources/assets/components/App.js
@@ -8,7 +8,7 @@ import { PuckProvider } from '@dosomething/puck-client';
 
 import { env } from '../helpers';
 import graphqlClient from '../graphql';
-import { getUserId } from '../selectors/user';
+import { getUserId, isAuthenticated } from '../selectors/user';
 import { initializeStore } from '../store/store';
 import ProfilePage from './pages/ProfilePage/ProfilePage';
 import CampaignContainer from './Campaign/CampaignContainer';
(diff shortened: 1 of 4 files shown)


resources/assets/components/App.js
```diff
@@ -17,11 +17,14 @@ import GeneralPageContainer from './pages/GeneralPage/GeneralPageContainer';
...
 
+  const state = store.getState();
+
   return (
     <Provider store={store}>
       <PuckProvider
         source="phoenix-next"
-        getUser={() => getUserId(store.getState())}
+        getUser={() => getUserId(state)}
+        isAuthenticated={() => isAuthenticated(state)}
```
DFurnes: Let's add this to any events [triggered via the helper](https://github.com/DoSomething/phoenix-next/blob/15ba557ee670fffdbf8e4d6b31ad06ae6dea713a/resources/assets/helpers/analytics.js#L30-L35) as well.

mendelB: Ah right! Completely forgot that 🤕

mendelB: Updated!

PR #1024, opened 2018-07-11
Add reset submission call to fix looping bug

mendelB:
### What does this PR do?

This PR is a proposed fix to an infinite loop we were experiencing in our RB Submission Forms.

[This thread here](https://dosomething.slack.com/archives/C2BPA7M8F/p1531331017000136)
> So to recap:
It seems that with the updated version of React, the `getDerivedStateFromProps` lifecycle method is invoked for *every* re-render.
This seems not to have been the case in earlier versions, which was why, although we still maintained a successful submission in our store, we still weren’t getting caught in this infinite loop.
Now that it is happening though, what we could do is simply call the `resetPostSubmissionItem` method once we register a successful post, ensuring that the next time `getDerivedStateFromProps` is invoked, we will no longer reset the state to `shouldResetForm: true` causing the cycle to begin all over again. 
Like so https://git.io/fNI4H

### DIFF
--- a/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
+++ b/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
@@ -32,6 +32,8 @@ class PhotoSubmissionAction extends React.Component {
     const response = nextProps.submissions.items[nextProps.id] || null;
 
     if (has(response, 'status.success')) {
+      nextProps.resetPostSubmissionItem(nextProps.id);
+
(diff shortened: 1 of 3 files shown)


weerd:
@mendelB just want to confirm this works with both a PSAction and subsequent TSAction and seeing appropriate (and different) messaging on the respective affirmation modals.

mendelB:
I tested on `test-teens-for-jeans` which has both, and can confirm - working, and different affirmation messages for each 👌

PR #1025, opened 2018-07-12
Campaign Page block rendering / goodbye action page

mendelB:
### What does this PR do?

This PR updates the block rendering in `CampaignPage` and removes the `/ActionPage` components.

- always render the `subPage` blocks, whether there's content or not (seemed to be an oversight from #1007 )
- add `actionSteps` page formatting logic, to apply a default `primary` class to all blocks so that they're two-thirds width on `$large` pages
- remove the `/action` route in favor of the default `:campaign-slug/:slug` route
  - this does away with some logic we had to redirect to `/community` for closed campaigns. I believe the ultimate intention was for leads to unpublish the `action` page from closed campaigns?
- 🔥🗑 the old ActionPage components which causes us to lose the following:
  - truncation of action steps and revealer CTA button for un-signed up users users
  - legacy gallery for signed up users on legacy campaigns

To further consolidate all campaign pages into our `CampaignPage` component. We removed the `ActionPage` associated components and imported some of the ActionPage rendering logic into the main `CampaignPageContent` component, applying the width styling logic to all blocks.

### Checklist

screen shot 2018-07-12 at 2 32 50 pm

screen shot 2018-07-12 at 2 36 10 pm


mendelB:
One final page field we're not yet dealing with in `CampaignPageContent` is the `subtitle` field. Is that something we want to perhaps display as a header on top of the page or something, or just ignore?

resources/assets/components/pages/CampaignPage/CampaignPageContent.js
```diff
@@ -26,28 +27,54 @@ const CampaignPageContent = props => {
...
+
+    // Only setting full column width for Content Blocks with an image
+    if (type === 'contentBlock' && json.fields.image) {
+      fullWidth = true;
+    }
+
+    return (
+      <div
+        className={classnames('margin-bottom-lg clear-both', {
+          primary: !fullWidth,
```
DFurnes: Where's this `primary` class coming from? Wondering if we should rename it to avoid conflicts.

mendelB: It's from the [`campaign-page.scss`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/pages/CampaignPage/campaign-page.scss#L8) (left over from the old campaignSubPage format.
A global search didn't yield it being defined anywhere else (at least within Phoenix), though it does look like it's being utilized by the [`PitchTemplate`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/pages/LandingPage/templates/PitchTemplate.js#L9)

mendelB:
**NOTE**:
We encountered a bit of a quirk with the Page `sidebar` blocks, where they for some odd reason are *not* being serialized with their entity's `jsonSerialize` method. 
Practically speaking this means that any CTA's added to the sidebar (to maintain consistency with the old hardcoded sidebar CTA's for subpages) will be limited to a `light` background. 

If @lkpttn @ashleybaldwin find this blocking, we could always quickly update to account for an array `visualStyle` field.

mendelB:
> Quick ques... if the Action page is unpublished, what happens with a link that goes to it is clicked (for example some shared link or whatever). Does it go to community?

Since that redirect routing logic is gone, you'd just hit the Not Found page as with any other non-existant page slug

image

weerd:
Ok, cool. I think that's fine for now. We can reassess and decide if we end up redirecting to whatever is the first item in the array of `pages` that is "published" as the default behavior.

mendelB:
Ah yeah that sounds like a reasonable default!

lkpttn:
To be honest I think we'll be removing those CTA blocks so I don't mind the visual treatment. In fact if it's easier to just remove them now I'd rather do that.

mendelB:
OK, that's awesome since the CTA's won't be automatically added anymore, and won't show up unless an editor adds them manually.

mendelB:
refs #925

PR #1028, opened 2018-07-13
initial face work

deguzmankevin:
### What does this PR do?

This PR adds much of the first part of the new profile page (aka FACE) to phoenix-next. 

This is the first actual React/Redux/GraphQL code I've ever written - please let me know if I'm doing anything seriously wrong or not using them how they're meant to be used! 

So far it's pretty [similar](https://ibb.co/e9koP8) to the wireframe and is mostly functional

Refs [#157662685](https://www.pivotaltracker.com/story/show/157662685)


routes/web.php
```diff
@@ -6,7 +6,7 @@
  * @var \Illuminate\Routing\Router $router
  * @see \App\Providers\RouteServiceProvider
  */
-
+// $router->redirect('/us/profile', 'https://identity-qa.dosomething.org');
```
deguzmankevin: routing in general has been troubling, still a WIP (shoutout to Mendel for all the help though!)

resources/assets/components/pages/ProfilePage/AccountNavigation.js
```diff
@@ -0,0 +1,79 @@
...
+  handleCampaignClick() {
+    this.setState({ accountActive: false });
+  }
+  handleAccountClick() {
+    this.setState({ accountActive: true });
+  }
+
+  render() {
+    let component;
+    if (this.state.accountActive) {
```
deguzmankevin: I know there's a better way to render a component based on the route, will try to figure it some other time

routes/web.php
```diff
@@ -17,7 +17,12 @@
 $router->redirect('auth/login', 'next/login'); // Fix for hard-coded redirect in Gateway! <goo.gl/2VPxDC>
 
 // Profile
-$router->view('/us/profile', 'app');
+// $router->redirect('/us/profile', '/us/profile/account');
```
DFurnes: You should be able to use the `view` route helper for all these & remove the `ProfileController`:

```php
$router->view('/us/profile/{clientRoute?}', 'app')
    ->where('clientRoute', '.*');
$router->view('/northstar/{id}', 'app');
```

resources/assets/components/pages/ProfilePage/ProfileContainer.js
```diff
@@ -0,0 +1,9 @@
+import { connect } from 'react-redux';
+
+import ProfilePage from './ProfilePage';
+
+const mapStateToProps = state => ({
+  userId: String(state.user.id),
```
DFurnes: Let's use the [selector pattern](https://hackernoon.com/selector-pattern-painless-redux-store-destructuring-bfc26b72b9ae) to grab this from the store:

```js
import { getUserId } from '../../../selectors/user';

  userId: getUser(state),
});
```

resources/assets/components/pages/ProfilePage/AccountQuery.js
```diff
@@ -0,0 +1,38 @@
...
+      email
+    }
+  }
+`;
+
+const AccountQuery = ({ userId }) => (
+  <Query query={ACCOUNT_QUERY} queryName="user" variables={{ userId }}>
+    {({ loading, error, data }) => {
+      if (loading) {
+        return 'Loading...';
```
DFurnes: We can return `<div className="spinner -centered" />` to make this look a bit cleaner.

resources/assets/components/pages/ProfilePage/AccountQuery.js
```diff
@@ -0,0 +1,38 @@
...
+      }
+      if (error) {
+        return `Error! ${error.message}`;
```
DFurnes: Let's use an [`ErrorBlock`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/ErrorBlock/ErrorBlock.js) here:

resources/assets/components/pages/ProfilePage/AccountNavigation.js
```diff
@@ -0,0 +1,79 @@
...
+        </div>
+      </div>
+    );
+  }
+}
+
+export default AccountNavigation;
+
+AccountNavigation.propTypes = {
+  user: {
```
DFurnes: I think this probably needs to be wrapped in a `PropTypes.shape` to be a [valid proptype](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes):

```js
AccountNavigation.propTypes = {
  user: PropTypes.shape({
    firstName: PropTypes.string,
  }).isRequired,
};
```

resources/assets/components/pages/ProfilePage/Account.js
```diff
@@ -0,0 +1,67 @@
...
+  user: {
+    firstName: PropTypes.string,
+    lastName: PropTypes.string,
+    email: PropTypes.string,
+    mobile: PropTypes.string,
+  },
+};
+
+Account.defaultProps = {
+  user: PropTypes.object,
```
DFurnes: The `defaultProps` property is used to set a default value if a prop is omitted, so for example we might say `{ firstName: 'A Doer', ... }` here. We can probably set this proptype as required above since it's coming from our GraphQL query though, so we don't need a fallback in this case.

resources/assets/components/pages/ProfilePage/Account.js
```diff
@@ -0,0 +1,67 @@
...
+        </button>
+      </a>
+    </div>
+  </div>
+);
+
+export default Account;
+
+Account.propTypes = {
+  user: {
```
DFurnes: Same question as below with prop-type format!

resources/assets/components/pages/ProfilePage/Account.js
```diff
@@ -0,0 +1,67 @@
...
+    </div>
+    <div className="margin-top-lg">
+      <h5>Phone Number</h5>
+      <div className="margin-top-md">
+        <p>{props.user.mobile}</p>
+      </div>
+    </div>
+    <div className="margin-top-lg">
+      <a
+        href="https://identity-qa.dosomething.org" // profile.dosomething.org for prod, should there be a redirect once profile is edited?
```
DFurnes: We [expose some environment variables](https://github.com/DoSomething/phoenix-next/blob/e29f0b5ca8bfc9d2ade33e7631dcc81818797696/app/Providers/AppServiceProvider.php#L26-L40) in a `window.ENV` object, so this can be accessed here by importing the [`env` client-side helper](https://github.com/DoSomething/phoenix-next/blob/e29f0b5ca8bfc9d2ade33e7631dcc81818797696/resources/assets/helpers/index.js#L394-L402):

```
<a href={env('NORTHSTAR_URL')} ...>
```

resources/assets/components/pages/ProfilePage/Account.js
```diff
@@ -0,0 +1,67 @@
...
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        <button className="button" type="button">
```
DFurnes: Totally nitpicky, but you can attach this `"button"` class directly to the wrapping `a` tag! In general, an `a` is used for linking between pages and `button` is used for non-navigational actions.

resources/assets/components/pages/ProfilePage/Account.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Account = props => (
+  <div className="bg-gray padding-bottom-lg wrapper">
+    <h2 className="caps-lock league-gothic-sm">Your Account </h2>
+    <div className="margin-top-lg">
+      <h3>Account Info</h3>
+    </div>
+    <div className="margin-top-lg">
```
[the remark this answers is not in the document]
DFurnes: I think that's a great idea! 👍 Maybe a little `FormItem` component?

resources/assets/components/pages/ProfilePage/AccountNavigation.js
DFurnes: Take a look at [React Router's documentation](https://reacttraining.com/react-router/web/example/basic) for ideas here! You should be able to use `<Route>` components to render each of these pages depending on the URL. 📃

app/Http/Controllers/ProfileController.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    /**
+     * Display the specified resource.
+     *
+     * @param  int  $id
+     * @return \Illuminate\Http\Response
+     */
+    public function show()
+    {
+        $user = auth()->user();
```
weerd: No real need to save a `$user` variable here since only using it in one place. You can pop the `auth()->user()` into the `if (...)` conditional :)

routes/web.php
deguzmankevin: thanks for the suggestion! I had it like that at some point, but added back the `ProfileController` to handle logic for going to the profile page when someone isn't signed in. Is there a simple way to handle that through react?

mendelB: Just updating here that @deguzmankevin and I chatted about using a React Route `render` method a la @DFurnes [suggestion above](https://github.com/DoSomething/phoenix-next/pull/1028/files#r202428091) to handle the auth/non-auth logic. And spoke about using the `/next/login` route instead of redirecting to Northstar directly to handle logging the user in and persisting the session.

PR #1030, opened 2018-07-16
Purging Feed components and misc

mendelB:
### What does this PR do?

This PR 🔪with 🔥the `Feed` components, actions, reducers, selectors, intialStates, and soul from the application.

(Also snuck in a fix to some annoying test warnings [b7f366d](https://github.com/DoSomething/phoenix-next/pull/1030/commits/b7f366d89bf07d1eb7830841ccda8612eebcb707))

We used this to render our legacy `ActivityFeed`, which is no longer. If at a later date we decide to utilize these codes, we shall come back here for reference.


weerd:
post-merge 👏

PR #1034, opened 2018-07-16
Sixpack Experiment component: dynamic experiment participation

weerd:
### What does this PR do?
This PR adds a `SixpackExperiment` component to potentially handle content-based A/B tests so that content editors can run their own experiments. The intention is that this new component will eventually be combined/remove the need for the old `Experiments` component, so that setting up Sixpack experiments is easier not just for content editors on Contentful, but also making it easier to setup some of the more involved tests that require engineering setup!

Example dynamic experiment setup in Sixpack:
image

For the time being the code in this PR only handles creating experiments dynamically in Sixpack, when a `SixpackExperiment` content type is attached to a reference field in Contentful, and participating a user into a selected test alternative.

The actual _conversion_ of a test will be in an upcoming PR, just aiming to keep this PR more streamlined.

I added a `SixpackExperimentContainer` but technically, it's currently unnecessary since I'm not using the campaign ID or slug it's providing to the `SixpackExperiment` component. I left it in because it might come into play. Also wanted to gauge and see what you all thought of possibly _prepending_ the campaign slug to the experiment name, so as more and more experiments get added, they can be easier to find in Sixpack via the prepended slug? 🤔 Something like `grab_mic_some_experiment_title`, `grab_mic_other_experiment_title`, `teens_for_jeans_landing_page_experiment`, etc...

Refs [Pivotal ID #158394982](https://www.pivotaltracker.com/story/show/158394982)

### Checklist


resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -130,6 +131,9 @@ class ContentfulEntry extends React.Component<Props, State> {
       case 'shareAction':
         return renderShareAction(json);
 
+      case 'sixpackExperiment':
+        return renderSixpackExperiment(json);
```
DFurnes: Would it make sense to inline this here since the function is so small?

```js
case 'sixpackExperiment':
    return <SixpackExperimentContainer id={data.id} {...withoutNulls(json)} />;
```

weerd:
PS> I'm refraining from creating the Contentful migration with the `sixpackExperiment` content type definition until _after_ I complete the conversion logic, since we should know all the necessary fields after that 😄

resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -130,6 +131,14 @@ class ContentfulEntry extends React.Component<Props, State> {
       case 'shareAction':
         return renderShareAction(json);
 
+      case 'sixpackExperiment':
+        return (
+          <SixpackExperimentContainer
+            id={json.id}
+            {...withoutNulls(json.fields)}
```
mendelB: Is there a reason we use `withoutNulls` here, or is this just a good pattern for whenever we pass our JSON fields to Components?

resources/assets/components/utilities/SixpackExperiment/SixpackExperiment.js
```diff
@@ -0,0 +1,51 @@
...
+    const alternativeOptions = alternatives.map(item =>
+      snakeCase(item.fields.title),
+    );
+
+    const selectedAlternative = participateBeta(
+      snakeCase(title),
+      alternativeOptions,
+    );
+
+    selectedAlternative.then(response => {
```
mendelB: Would it be a good idea to have a `catch` to default to the control alternative?

resources/assets/components/ContentfulEntry/ContentfulEntry.js
weerd: @DFurnes made a recommendation that since this was just simply returning the component, it might be nicer to not require a call to a corresponding "renderer" function with basically the same thing.

but in terms of doing it inside the `{...}` instead of assigning a variable to the result of passing fields to that func, I was down with giving it a whirl. We could switch things up to do that in other places if we like it :)

resources/assets/components/utilities/SixpackExperiment/SixpackExperiment.js
weerd: Oh yeah, I was contemplating adding a catch and it returns the first item (default) in the alternatives array. In Contentful I was planning on making the `alternatives` ref field require at least 1 item. I can add that real quick!

weerd:
> Side question: Is it part of our plan to add conditions to these experiments akin to what we had specified in JSON in the old format?

I'm seeing what makes sense to bring in. For now, the main requirement is just participation in experiment and conversion. Working on the conversion part now and that's where I'm trying to figure out how "conditions" would work and _if_ I can make them work 💃

resources/assets/components/ContentfulEntry/ContentfulEntry.js
mendelB: Ah gotcha! I was mainly curious about the usage of `withoutNulls` as opposed to just passing the `json.fields` as-is to the component since there aren't any default props?

PR #1035, opened 2018-07-17
Facebook Share Action to Rogue

mendelB:
### What does this PR do?

This PR adds a call to post successful Facebook share actions to Rogue for authenticated users on a campaign Share Action.

I got the `type` and `details.platform` fields as well as more context from this doc https://github.com/DoSomething/chompy/wiki/How-Historic-FB-Share-Records-are-Imported

Refs [Pivotal ID #158772936](https://www.pivotaltracker.com/story/show/158772936)

### DIFF
--- a/resources/assets/components/actions/ShareAction/ShareAction.js
+++ b/resources/assets/components/actions/ShareAction/ShareAction.js
@@ -61,8 +61,11 @@ class ShareAction extends React.Component {
 
     showFacebookShareDialog(url)
       .then(() => {
-        // @TODO: Once Rogue is ready to accept this post we'll activate.
-        this.storeSharePost();
+        // Send share post to Rogue for authenticated users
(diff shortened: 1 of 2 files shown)


resources/assets/components/actions/ShareAction/ShareAction.js
```diff
@@ -25,11 +27,45 @@ class ShareAction extends React.Component {
...
   }
 
+  storeSharePost = () => {
+    const type = 'share-social';
+
+    const action = get(this.props.additionalContent, 'action', 'default');
+
+    const { id, campaignId, campaignRunId, legacyCampaignId } = this.props;
+
+    const formData = setFormData(
```
DFurnes: Sorta unrelated to this pull request, but it might make sense to rename this `createFormData`. At first read I was kinda thinking it was mutating something somewhere else (like on a form in the DOM or something).

PR #1038, opened 2018-07-19
Add IDs to the rendered blocks on campaign pages

mendelB:
### What does this PR do?

This PR adds an `id` props to the blocks rendered by Campaign Pages, to bring back support for anchor links.

We previously would add the Contentful entry ID as a prop to blocks being rendered by the retired [`ActionPage`](https://github.com/DoSomething/phoenix-next/pull/1025/files#diff-3f7676fcbfa0a4ecba1bfd24a77047ffL72). 
This allowed us to add 'anchor' links, to a specific block on a page, most notably, we would do this through an `additionalContent.affiliatedActionLink` field on campaigns e.g. this 'register to vote' button linking to the voter registration action on the page (which no longer works 😭)
image

### DIFF
--- a/resources/assets/components/pages/CampaignPage/CampaignPageContent.js
+++ b/resources/assets/components/pages/CampaignPage/CampaignPageContent.js
@@ -46,6 +46,7 @@ const CampaignPageContent = props => {
           primary: !fullWidth,
         })}
         key={json.id}
+        id={`block-${json.id}`}
       >
         <ContentfulEntry json={json} />
       </div>


resources/assets/components/pages/CampaignPage/CampaignPageContent.js
```diff
@@ -46,6 +46,7 @@ const CampaignPageContent = props => {
           primary: !fullWidth,
         })}
         key={json.id}
+        id={`block-${json.id}`}
```
mendelB: It used to be `step-${json.id}` so I figured `block` was a nice substitute?

DFurnes: Makes sense to me!

PR #1040, opened 2018-07-20
Error logging for failed POST requests to Phoenix backend

mendelB:
### What does this PR do?

This PR adds error logging for failed POST requests made to the Phoenix backend from our API middleware using the Gateway `RestApiClient`.

We want to be able to analyze and recognize when one of our calls is failing, specifically regarding the Share action post to Rogue #1035 where we don't have any failure handling exposed to the user.

Refs [Pivotal ID #158772936](https://www.pivotaltracker.com/story/show/158772936/comments/192093651)

### DIFF
--- a/resources/assets/middleware/api.js
+++ b/resources/assets/middleware/api.js
@@ -2,9 +2,11 @@
 
 import { RestApiClient } from '@dosomething/gateway';
 
+import { report } from '../helpers';
 import { PHOENIX_URL } from '../constants';
 import { getRequest } from '../helpers/api';
 import { API } from '../constants/action-types';
+import { trackPuckEvent } from '../helpers/analytics';
 
 /**
  * Send a GET request and dispatch actions.
@@ -68,7 +70,14 @@ const postRequest = (payload, dispatch) => {
       });
     })
     .catch(error => {
+      report(error);
+      trackPuckEvent('phoenix_failed_post_request', {
+        url: payload.url,
+        body: payload.body,
+        error,
+      });
       console.log('🚫 failed response; caught the error!');
+
       const response = error.response;
 
       dispatch({


resources/assets/middleware/api.js
```diff
@@ -68,7 +70,14 @@ const postRequest = (payload, dispatch) => {
       });
     })
     .catch(error => {
+      report(error);
+      trackPuckEvent('phoenix_failed_post_request', {
+        url: payload.url,
+        body: payload.body,
```
mendelB: I wasn't completely sure what to add here as data in general. I figured the body might be important, so we can determine which RB post request failed (text/photo/social) Issue is this would be a FormData object which I don't think is very helpful.

1. Would we want a helper to convert formData to JSON and log that?
2. Maybe just log the `type` field, so we know which RB it was?

resources/assets/middleware/api.js
```diff
@@ -68,7 +70,14 @@ const postRequest = (payload, dispatch) => {
       });
     })
     .catch(error => {
+      report(error);
```
mendelB: I'm not quite sure how this looks on the New Relic end, as I wasn't able to test this properly, but it seemed reasonable to me to just log the error as is, similarly to how we do it in [`ContentfulEntry`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/ContentfulEntry/ContentfulEntry.js#L43)?

resources/assets/middleware/api.js
```diff
@@ -68,7 +70,14 @@ const postRequest = (payload, dispatch) => {
...
+      trackPuckEvent('phoenix_failed_post_request', {
```
mendelB: And of course -- how's the naming for this? 😊

resources/assets/middleware/api.js
weerd: Yes to the `type` field.

As for the `formData`, yeah, I think it would be useful to turn it into JSON. At a minumum you should be able to loop through it and make an object to output 🤔 

The `formData` object has some looping methods like `FormData.entries()` or `FormData.values()` but the caveat is that they're not widely supported!

resources/assets/middleware/api.js
DFurnes: These should show up as normal exceptions in New Relic!

resources/assets/middleware/api.js
mendelB: Ok, do you think we're cool with returning `type` only, or the entire JSON serialized response?

Also, it looks like `get` and `entries` are not supported in Internet Explorer or Safari mobile. I guess we can add a check to ensure support before calling these methods in our helpers, and for those platforms be resigned to not having the form data in the event logs?

image

weerd: > I guess we can add a check to ensure support before calling these methods in our helpers, and for those platforms be resigned to not having the form data in the event logs?

Yeah, I think this would be what I suggest. Go with `entries()` which seems to be the iterator that has the most support, and check against it :)

resources/assets/middleware/api.js
weerd: Looks good. Only suggestion _might be_ whether we want to specify the post request failed getting sent through to Phoenix vs making it through Phoenix and failed on the Rogue side. We want the former right?

mendelB: I guess we really should have both in our logs? And this would just be for analytics to allow us to consolidate specific failures more easily which might mean this could just be a catch-all for both?

DFurnes: Yeah, I think we'd be mostly relying on New Relic & logs for tracking down real issues. The Puck event would just let us see big picture trends on different types of API requests.

weerd: Cool cool. Sounds good to me then 💃

mendelB:
Updated to add and use a form helper to extract the FormData entries into a regular Object

mendelB:
omg that's hilarious `prettier` put the `eslint-disable` comment *on top* of the code

resources/assets/helpers/forms.js
```diff
@@ -99,8 +99,8 @@ export function getFormData(formData) {
 
   const formDataObject = {};
 
+  // eslint-disable-next-line no-restricted-syntax
```
DFurnes: Oh woah, didn't know `eslint-disable-next-line` was a thing. I'd just resorted to global per-file disables when this had happened to me in other places. Neat! 🦁

PR #1041, opened 2018-07-20
Fix routing

deguzmankevin:
### What does this PR do?

This PR makes the browser routing on the profile page more of the react/react-router way and deletes some unnecessary components

It's simpler than before and it's easy to add another section to the nav bar if we ever decide to!


routes/web.php
```diff
@@ -18,7 +18,9 @@
 
 // Profile
 $router->redirect('/northstar/{id}', '/us/profile/info');
-$router->redirect('/us/profile/campaigns', '/us/profile/info'); //placeholder for now while figuring out routing
+$router->get('/us/profile/campaigns', function () {
```
deguzmankevin: I tried a few things so this wouldn't be redundant, but couldn't think of a good solution. Is there any way to have any route of the form '/us/profile/{slug}' redirect if a user isn't signed in, without needing a controller? The problem with using '/us/profile/{slug}' is that the slug could be anything besides info/campaigns and it still goes to the profile page (ie /us/profile/hello would work). Maybe there's a clean way to verify that the slug is info/campaigns?

mendelB: @deguzmankevin 
You could try using a `/{slug}` route, and appending a `where` clause to the route like so
```
$router->get('/us/profile/{slug}', function () {
    return auth()->user() ? view('app') : redirect('/next/login');
})->where([
	'slug' => 'campaigns|info'
]);
```

(I'm not positive that this will work but worth a whirl)

mendelB: Probs cleaner as `->where('slug','campaigns|info');`

resources/assets/components/pages/ProfilePage/ProfileRoute.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Switch, Route } from 'react-router-dom';
+
+import Account from './Account';
+import TestCampaign from './TestCampaign';
+
+const ProfileRoute = props => (
+  <div>
+    <Switch>
+      <Route path="/us/profile/info" render={() => <Account {...props} />} />
```
mendelB: Do you need to use `render` to pass along the `props` or does that happen automatically with `component={...}`?

resources/assets/components/pages/ProfilePage/AccountNavigation.js
```diff
@@ -3,71 +3,46 @@ import { NavLink } from 'react-router-dom';
...
-        </div>
       </div>
-    );
-  }
-}
+    </div>
+    <div className="container bg-gray margin-top-lg">
+      <div className="wrapper bg-gray">
+        <Enclosure className="">
+          <ProfileRoute {...props} />
```
mendelB: Way cleaner! 👏

resources/assets/components/pages/ProfilePage/ProfileRoute.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Switch, Route } from 'react-router-dom';
+
+import Account from './Account';
+import TestCampaign from './TestCampaign';
+
+const ProfileRoute = props => (
+  <div>
```
mendelB: Small fry, but do we need this surrounding `<div>`?

resources/assets/components/pages/ProfilePage/ProfileRoute.js
deguzmankevin: yeah, `component={...}` wasn't passing along the props so I used render. I also tried doing `component={AccountContainer}` and passing along props through that container to `Account`, but it didn't work either!

resources/assets/components/pages/ProfilePage/ProfileRoute.js
deguzmankevin: good catch you're totally right! don't need it at all :wave:

deguzmankevin:
@mendelB yeah I think that's a good idea! also if you look in the #team-bleed channel, there was some discussion about the difficulty/workload of the campaigns portion. It looks like I'll be starting with displaying all of a user's posts instead of campaigns!

PR #1042, opened 2018-07-20
Experiments refactor cleanup

weerd:
### What does this PR do?

This PR does a bit of refactor/cleanup for Experiments related code I noticed could use a bit of sprucing 🌲 up. I started a bit of this in my Sixpack Component PR, but to keep it from getting to huge, I figured I'd separate this out 😄 

It's probably better to review this PR [commit by commit](https://github.com/DoSomething/phoenix-next/pull/1042/commits), but here's the general gist of what it does:

- Renames the action constants to follow a better pattern, and moves them to the more appropriate location in the `/constants/action-types.js` file.
- It consolidates the `experiments.js` middleware and the `experimentsApi.js` middleware into just one `experiments.js` middleware to contain all the logic. When I first tackled and setup the middlewares for experiments, it seemed to make sense at the time to have a separate middleware deal with Sixpack API stuff and another for Redux store stuff. But they kind of started to mix together and every time I came back to work on something with experiments I'd be confused as to which did what and why. So to simplify things, they've been combined!
- It also consolidates the `helpers/experiments.js` file and the `helpers/sixpack.js` files into just a `helpers/experiments.js` file. Again, this was a bit confusing and didn't make the most sense to keep them separate and combining helps to simplify things.

Everything should technically work the same as before. I foresee some additional cleanup and simplification coming in the aforementioned Sixpack Component PR, so mostly just review the cleanup and not a re-thinking of the logic just yet! 

Refs [Pivotal ID #158394982](https://www.pivotaltracker.com/story/show/158394982)


resources/assets/constants/action-types.js
```diff
@@ -8,3 +8,9 @@ export const POST_SUBMISSION_INIT_ITEM = 'POST_SUBMISSION_INIT_ITEM';
 export const POST_SUBMISSION_PENDING = 'POST_SUBMISSION_PENDING';
 export const POST_SUBMISSION_RESET_ITEM = 'POST_SUBMISSION_RESET_ITEM';
 export const POST_SUBMISSION_SUCCESSFUL = 'POST_SUBMISSION_SUCCESSFUL';
+
+export const EXPERIMENT_ADD_TEST_TO_STORE = 'EXPERIMENT_ADD_TEST_TO_STORE';
```
DFurnes: I wonder if these could be more concise, like `EXPERIMENT INIT`, `EXPERIMENT_CONVERT`, `EXPERIMENT_PARTICIPATE`, and `EXPERIMENT_UPDATE`?

resources/assets/store/middlewares.js
```diff
@@ -1,13 +1,7 @@
...
 import experimentsMiddleware from '../middleware/experiments';
-import experimentsApiMiddleware from '../middleware/experimentsApi';
 
-const middlwares = [
-  analyticsMiddleware,
-  apiMiddleware,
-  experimentsApiMiddleware,
-  experimentsMiddleware,
-];
+const middlwares = [analyticsMiddleware, apiMiddleware, experimentsMiddleware];
```
DFurnes: Sneaky typo! `middlwares` → `middlewares `

resources/assets/constants/action-types.js
weerd: Yeah, I was contemplating shorter constant names vs more obvious explanation as to what it's doing. So would `EXPERIMENT_INIT` be what `EXPERIMENT_INIT_PARTICIPATION` is, and `EXPERIMENT_PARTICIPATE` be `EXPERIMENT_PARTICIPATE`?

I dig that a bit, but debating if it follows the lingo of how the experiments actually run. Participation technically happens at initialization.... 🤔

resources/assets/store/middlewares.js
weerd: ha! I think that's been like that since the beginning! It `export default middlwares;` as well. I'll fix for upcoming PR. Nice catch!

resources/assets/constants/action-types.js
DFurnes: Huh, maybe I'm confused about how Sixpack works! Why are `EXPERIMENT_INIT_PARTICIPATION` and `EXPERIMENT_PARTICIPATE` separate things?

weerd: haha, you suggested `EXPERIMENT_PARTICIPATE`. I'm just trying to understand which of the names you suggested match up with what from the existing constants I renamed :)

`EXPERIMENT_INIT_PARTICIPATION` = initializes experiment participation by the user; sixpack associates unique ID for user with being participated into an experiment.
`EXPERIMENT_ADD_TEST_TO_STORE` = we add specific test that the user has been participated into to the Redux store for future reference.
`EXPERIMENT_UPDATE_TEST_IN_STORE` = indicate in the store which A/B test the user got assigned.
`EXPERIMENT_CONVERT_ON_TEST` = convert user on the assigned test.

I am trying to see if I can simplify this down in upcoming PRs, and likely just minimize to 3 constants, but this PR was only meant to refactor and rename without changing around the number of constants, etc.

DFurnes: Sorry if I wasn't clear - the names were meant to match up in the same order currently listed:

`EXPERIMENT_ADD_TEST_TO_STORE` → `EXPERIMENT_INIT`
`EXPERIMENT_CONVERT_ON_TEST` → `EXPERIMENT_CONVERT`
`EXPERIMENT_INIT_PARTICIPATION` → `EXPERIMENT_PARTICIPATE`
`EXPERIMENT_UPDATE_TEST_IN_STORE` → `EXPERIMENT_UPDATE`

I didn't realize `EXPERIMENT_ADD_TEST_TO_STORE` only runs once the user is participated, so that's where my confusion came in. Definitely on board with trying to simplify these further in a future PR!

PR #1044, opened 2018-07-23
Social drive Puck Events

mendelB:
### What does this PR do?

This PR edits the Puck event tracking for the `SocialDriveAction` component
- adds a new puck event `phoenix_clicked_copy_to_clipboard` on copy link to clipboard clicks
- updates the puck event tracking `data` fields to use the original `link` prop instead of the shortened link
- adds a tracking `data.url` field for twitter share clicks

Refs [Pivotal ID #159035714](https://www.pivotaltracker.com/story/show/159035714)

### DIFF
--- a/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
+++ b/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
@@ -45,6 +45,9 @@ class SocialDriveAction extends React.Component {
   handleCopyLinkClick = () => {
     this.linkInput.current.select();
     document.execCommand('copy');
+    trackPuckEvent('phoenix_clicked_copy_to_clipboard', {
+      url: this.props.link,
+    });
   };
 
   handleFacebookShareClick = url => {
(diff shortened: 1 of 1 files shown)


resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -45,17 +45,22 @@ class SocialDriveAction extends React.Component {
   handleCopyLinkClick = () => {
     this.linkInput.current.select();
     document.execCommand('copy');
+    trackPuckEvent('phoenix_clicked_copy_to_clipboard', {
```
deguzmankevin: probably not important, but just wondering why this puck event is in snake case but the others aren't??

mendelB: Good question! we [recently established some conventions](https://github.com/orgs/DoSomething/teams/tech-leads/discussions/3) to use going forward for Puck events, which among other things suggests using 🐍casing. 

Most other events are still relics of the wild wild west puck event system™

deguzmankevin: ah I see, thanks for the explanation!

PR #1051, opened 2018-07-31
Empty State, User Posts, and Refactoring

deguzmankevin:
### What does this PR do?

This PR adds the user posts section to the profile page, along with an empty state to the profile page for when a user has no uploads!

I've also refactored a lot of the code and renamed components, so the flow of nested React Components makes a little more sense!

I should've made a separate PR for the user posts gallery my bad 😬!! I've refrained from further work on it anyways until we know whether or not we want to ship it or wait and work towards the user campaigns gallery end goal.

#### web empty state
screen shot 2018-07-30 at 1 46 08 pm

#### web posts gallery
screen shot 2018-07-30 at 1 50 10 pm

#### mobile empty state
screen shot 2018-08-01 at 3 08 37 pm

#### mobile account info
screen shot 2018-08-01 at 2 58 29 pm

#### mobile posts gallery
screen shot 2018-08-01 at 3 02 20 pm

#### tablet empty state
screen shot 2018-08-01 at 3 08 47 pm

#### tablet account info
screen shot 2018-08-01 at 3 11 11 pm

#### tablet posts gallery
screen shot 2018-08-01 at 3 01 48 pm

#### with routing as /account/profile
screen shot 2018-08-07 at 10 02 22 am

Refs [#159276274](https://www.pivotaltracker.com/story/show/159276274)


resources/assets/components/pages/ProfilePage/EmptyStateBlock.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const EmptyStateBlock = () => (
+  <div className="margin-top-md margin-bottom-lg">
+    <p>
+      You {"don't"} have any uploads...yet! Find a cause you care about <br />and
```
weerd: What's `{"don't"}`?

weerd: If this was to resolve the single quote giving your trouble, you should just be able to escape it using `You don\'t have...`

weerd: Also, don't use the `<br />`. If you want that to look nicer, just separate the 2 sentences into 2 separate `<p>...</p>` :)

weerd: Last note on why `<br />`s are not ideal is that they still apply on small and medium screens, so the decision to add a `<br />` to make the text read nicely on large screens, will very likely end up looking weird and breaking awkwardly on the other sizes!

resources/assets/components/pages/ProfilePage/ProfileRoute.js
```diff
@@ -1,14 +1,22 @@
...
 import { Switch, Route } from 'react-router-dom';
+import PropTypes from 'prop-types';
 
 import Account from './Account';
-import TestCampaign from './TestCampaign';
+import UserPostsQuery from './UserPostsQuery';
 
 const ProfileRoute = props => (
   <Switch>
     <Route path="/us/profile/info" render={() => <Account {...props} />} />
```
weerd: Just curious why is the path `/us/profile/info` and not `/us/profile/account`?

weerd:
@deguzmankevin could you add screenshots for medium and small sizes too so we have an idea of how it looks on those sizes?

resources/assets/components/pages/ProfilePage/EmptyStateBlock.js
deguzmankevin: Yeah for some reason the escape character wasn't working! If I don't escape it at all, it looks fine but then I get the `no-unescaped-entities` error from React in terminal.

Thanks for all the info on `<br />`, I had no idea it could cause problems like that!

resources/assets/components/pages/ProfilePage/ProfileRoute.js
deguzmankevin: @weerd `/profile/account` just sounded weird to me because I always thought they were synonyms! I agree with @mendelB, it's pretty confusing - are they different things? Should we change the prefix?

resources/assets/components/pages/ProfilePage/ProfilePage.js
```diff
@@ -1,12 +1,35 @@
...
+        <h1 className="league-gothic-lg caps-lock">
+          Welcome, {props.user.firstName}!
+        </h1>
+        <AccountNavigation {...props} />
+      </div>
+    </div>
+
+    <div className="container bg-gray margin-top-lg">
+      <div className="wrapper bg-gray">
+        <Enclosure className="">
```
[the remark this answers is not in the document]
deguzmankevin: ah no! good catch - removing this!

resources/assets/components/pages/ProfilePage/EmptyStateBlock.js
mendelB: I feel like our convention for this has been to use [HTML Entities](https://www.w3schools.com/html/html_entities.asp) for special chars? This one would be `&#39;`

resources/assets/components/pages/ProfilePage/ProfilePage.js
```diff
@@ -1,12 +1,35 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
 import Enclosure from '../../Enclosure';
-import AccountQueryContainer from './AccountQueryContainer';
+import AccountNavigation from './AccountNavigation';
+import ProfileRoute from './ProfileRoute';
 
 const ProfilePage = props => (
   <Enclosure className="container margin-bottom-lg">
```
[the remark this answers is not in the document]
deguzmankevin: thanks Mendel!

deguzmankevin:
@weerd am I supposed to use the chrome inspect tool to see how it looks on small and medium screens? I'm not too familiar with it

resources/assets/components/pages/ProfilePage/EmptyStateBlock.js
deguzmankevin: ahh ok that makes sense!

resources/assets/components/pages/ProfilePage/ProfileRoute.js
mendelB: Also, @deguzmankevin, @weerd, do you think it makes sense to rename the `Account` and `AccountNavigation` components to `Profile`  and `ProfileNavigation` just for consistency? (I found it slightly confusing that we had some things as `Account` and some as `Profile`). Or was there a specific rationale in naming them `Account`?

mendelB:
@deguzmankevin If it's helpful, I'm happy to quickly go over how I go about screenshotting for Medium and Small screens using the inspector.

resources/assets/components/pages/ProfilePage/ProfileRoute.js
weerd: Yeah, I like your suggestion @mendelB. I was getting a bit confused as well. @deguzmankevin could we go with Mendel's suggestion and just make everything `Profile`, etc? I think that would be more straightforward. We can help with any naming issues that otherwise pop up from this :)

resources/assets/components/pages/ProfilePage/EmptyStateBlock.js
weerd: Oh nice, yeah, when in doubt (or something's not working) use the html entity. Good call @mendelB 👍

weerd:
> @weerd am I supposed to use the chrome inspect tool to see how it looks on small and medium screens? I'm not too familiar with it

There's a variety of ways. The simplest is just reduce your browser window until you hit the medium breakpoint, screenshot, and then again for the small breakpoint. But @mendelB might know of some secret, cooler ways to do it too 😉

mendelB:
drunk

resources/assets/components/pages/ProfilePage/ProfileRoute.js
deguzmankevin: thanks for the suggestions @mendelB @weerd! The rationale for using `Account` was to be consistent with the account tab on the navigation bar, ie clicking the `Account` tab renders the `Account` component. Would keeping it as `Account` be confusing? 

To be even more consistent, the route really should be `us/profile/account`, so I'll go ahead and change that back. Renaming `AccountNavigation` to `ProfileNavigation` definitely makes sense to me too!

deguzmankevin:
I've added the screenshots and made some changes for mobile/tablet! However, there's a gray shadow on the right side of the navigation bar on mobile view - it has to do with the `.page-navigation::before` class, in particular, this line:
` background: linear-gradient(90deg,hsla(0,0%,96%,0) 0,#f6f6f6 75%); ` 

any ideas on what this is for??

deguzmankevin:
interesting, spotify uses `/account/profile`...it's all so confusing 😆 
screen shot 2018-08-01 at 4 28 19 pm

weerd:
Maybe we follow suit w/ what Spotify has? 

`/us/account/profile` and `/us/account/posts`

I feel `account -> profile` makes more sense to me than `profile -> account` :)

deguzmankevin:
@weerd that sounds good to me! so the navigation tab should say `Profile` instead of `Account`, right? @lkpttn any thoughts on all this profile vs account stuff?

weerd:
@deguzmankevin yeah, I think so. @lkpttn @ashleybaldwin @ngjo does this make sense? Not sure who signs off on this. 

**tldr;** Instead of calling the whole thing Profile, it's Account. Account contains both the user's "profile" and a record of their "posts" which can be found at the following URLs `/us/account/profile` and `/us/account/posts` respectively.

Thoughts?

ngjo:
hi! I think that that makes sense to have it under "Account" but would especially want @lkpttn to jump in here w/ final sign-off as he's done a bunch of competitive research on this stuff!

One flag I'd have is that we probably don't want to call it /posts regardless since that feels more like an internal naming convention. I'd suggest /uploads or something similar -- thoughts @lkpttn ?

lkpttn:
Yeah I'm in favor of `/account/*`. 

I'd still like the navigation to say Campaigns instead of Posts, even if we're just showing posts right now.

lkpttn:
@weerd So if we want public profiles in the future, would having `posts` and 'profile' make things tricky. I could see a public "profile" including a user's posts on it and we might end up crossing the streams a little bit.

weerd:
@lkpttn not really. A public profile would just be something like `/accounts/:id/profile`. And the page would just showcase info that's public, but if we _know_ it's the user viewing their own account, they'd see more detailed info or whatever.

resources/assets/components/utilities/PageNavigation/page-navigation.scss
```diff
@@ -12,6 +12,12 @@
     z-index: $tabbed-navigation-zindex;
   }
 
+  &.-no-fade {
+    &::before {
+      display: none; // go away!
```
weerd: @deguzmankevin you can probs delete the silly comment I added for it to "go away!" 😂

mendelB: I quite enjoyed that tbh

weerd:
Making a note here for future reference! Chatted with @deguzmankevin to figure out a visual bug that was showing up with the page navigation. The page navigation uses a `page-navigation` class to get most of the visual styling and interaction styling from the page navigation we use on campaign tabbed nav for campaign pages.

However, when viewing the account page on small screens it would show a weird gray faded area to the right. On the campaign tabbed nav on small screen we use this faded element to signal if there are tabbed nab items to the right that are getting cut off and the menu can be scrolled horizontally. It works great on a campaign page where the tabbed nav sits on a grey background. However, on the account page, the nav sits on a white background, so the faded element on small screens on the account page looks misplaced.

To resolve this for now, I recommend adding in a modifier class on the nav in the account page to remove this element on those pages. (Alternatively we could just change the background gradient color on the element to white).

After our big "Ashes to Ashes" push, we'll be doing work to cleanup and improve a lot of UI on the site and we can address this with a better solution then.

mendelB:
Huge! Congrats Kevin, this is some stellar work! 💪 
image

PR #1053, opened 2018-08-02
Move Landing Page Logic to Campaign Page / Closed Campaign State Logic

mendelB:
### What does this PR do?

This PR

- moves the logic determining whether to show the Landing Page over to the `CampaignPage` component
- ensures Landing Page won't show up for closed campaigns
- abstracts the logic into a helper function called ...`shouldShowLandingPage`! 💡 

I was adding the logic to ensure landing pages did not show for closed campaigns and bumped into a helpful `@TODO` comment begging us to move the logic from the `CampaignContainer` into the respective pages themselves, so took the opportunity to do so!

It did feel like a good idea to throw the whole mess of logic into a helper method to clean the container and so that we can easily pull it into other components if needed in the future.
helper

Refs [Pivotal ID #159476747](https://www.pivotaltracker.com/story/show/159476747)

### DIFF
--- a/resources/assets/components/Campaign/CampaignContainer.js
+++ b/resources/assets/components/Campaign/CampaignContainer.js
@@ -2,32 +2,9 @@ import { connect } from 'react-redux';
 import { get } from 'lodash';
 
 import Campaign from './Campaign';
-import { isSignedUp } from '../../selectors/signup';
 
-const mapStateToProps = (state, props) => {
-  const hasLandingPage = state.campaign.landingPage !== null;
-
-  // @TODO: Move these into the pages themselves.
-  const { location, match } = props;
-  const isQuiz = location.pathname.replace(match.url, '').startsWith('/quiz/');
-  const isBlock = location.pathname
-    .replace(match.url, '')
-    .startsWith('/blocks/');
-
-  const ignoreLandingPage =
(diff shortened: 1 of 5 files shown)


resources/assets/components/Campaign/CampaignContainer.js
```diff
@@ -2,32 +2,9 @@ import { connect } from 'react-redux';
...
-    shouldShowLandingPage = !isSignedUp(state);
-  }
-
-  return {
-    shouldShowLandingPage,
-    featureFlags: get(state.campaign.additionalContent, 'featureFlags'),
-  };
-};
+const mapStateToProps = state => ({
+  featureFlags: get(state.campaign.additionalContent, 'featureFlags'),
```
weerd: So much cleaner!

resources/assets/helpers/index.js
```diff
@@ -380,6 +381,30 @@ export function isCampaignClosed(endDate) {
...
 }
 
+/**
+ * Check if the Landing Page should be shown for campaign.
+ *
+ * @param  {Object}       state
+ * @param  {Boolean|null} shouldIgnore
+ * @return {Boolean}
+ */
+export function shouldShowLandingPage(state, shouldIgnore) {
```
mendelB: On second thought, I'm going to try to rename the `shouldIgnore` a bit better.

PR #1054, opened 2018-08-03
Provide fallbacks and better logic for a couple Legacy Campaign fields.

weerd:
### What does this PR do?
This PR updates the LegacyCampaign Entity to handle data from Ashes regarding cover images a little better. We found there may be some campaigns that don't have a cover image, or don't have a cover image with a "landscape" crop assigned and thus breaks the parsing of the data.

It also updates handling in the LegacyCampaign Entity for the Ashes run ID. Thanks to global, not all campaigns will have a value under `en` for the current run. Since there are a couple different languages Ashes supported, instead of adding logic to address each one, I figured it would be easier to just take the first item in the `current` array and grab the ID from it, instead of dealing with each potential index of `en`, `es-mx`, `pt-br`, etc... Thanks to @mendelB for flagging this one! 🏅 

🌵 

### Checklist
- [ ] Documentation included in `/docs` for added/updated API endpoints.
- [ ] Added relevant Runscope API monitoring tests.

### DIFF
--- a/app/Entities/LegacyCampaign.php
+++ b/app/Entities/LegacyCampaign.php
@@ -47,6 +47,18 @@ class LegacyCampaign implements JsonSerializable
     }
 
     /**
+     * Parse the campaign run ID.
+     *
+     * @param  array $campaignRuns
+     * @return string
+     */
+    public function parseCampaignRunId($campaignRuns) {
+        $current = array_shift($campaignRuns['current']);
+
+        return $current['id'];
+    }
+
+    /**
      * Convert the object into something JSON serializable.
      *
      * @return array
@@ -56,7 +68,7 @@ class LegacyCampaign implements JsonSerializable
         return [
             'id' => null,
             'legacyCampaignId' => $this->legacyCampaign['id'],
-            'legacyCampaignRunId' => $this->legacyCampaign['campaign_runs']['current']['en']['id'],
+            'legacyCampaignRunId' => $this->parseCampaignRunId($this->legacyCampaign['campaign_runs']),
             'type' => 'campaign',
             'title' => $this->legacyCampaign['title'],
             'slug' => null,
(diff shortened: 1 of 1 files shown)


app/Entities/LegacyCampaign.php
```diff
@@ -23,6 +23,42 @@ public function __construct($legacyCampaign)
...
+     * @return array
+     */
+    public function parseCoverImage($coverImage)
+    {
+        $url = null;
+        $landscapeUrl = null;
+
+        if (isset($coverImage['default'])) {
+            $url = $coverImage['default']['uri'];
+            $landscapeUrl = isset($coverImage['default']['sizes']['landscape']) ? $coverImage['default']['sizes']['landscape']['uri'] : null;
```
mendelB: Could using [`array_get`](https://laravel.com/docs/5.6/helpers#method-array-get) here be of use, so you don't need the ternary?

app/Entities/LegacyCampaign.php
```diff
@@ -23,6 +23,42 @@ public function __construct($legacyCampaign)
...
+
+    /**
+     * Parse the campaign run ID.
+     *
+     * @param  array $campaignRuns
+     * @return string
+     */
+    public function parseCampaignRunId($campaignRuns)
+    {
+        $current = array_shift($campaignRuns['current']);
```
mendelB: ahh nice!

mendelB:
>  I figured it would be easier to just take the first item in the current array and grab the ID from it, instead of dealing with each potential index of en, es-mx, pt-br, etc...

And `en` would show up first in case both keys existed, right? Or would the ordering not be predictable? (And do we just not have any campaigns with both languages, to begin with?)

app/Entities/LegacyCampaign.php
weerd: oh ya. Good call 💃

weerd:
> And en would show up first in case both keys existed, right? Or would the ordering not be predictable? (And do we just not have any campaigns with both languages, to begin with?)

I'll double check this to make sure!

PR #1057, opened 2018-08-07
Switching to Sixpack class and enabling conversion.

weerd:
### What does this PR do?

This PR creates a new `Sixpack` class to help handle all experiments on a page. I worked with @DFurnes to initially stub out the idea. He had suggested trying to move away from storing the experiments in the Redux store and instead having a dedicated singleton class to handle experiments on a page and from this initial pass it seems to work rather well.

Being a singleton, the`Sixpack` class can only be instantiated once and will house all the experiments on a page (that means it keeps track of experiments on a page load, OR switching to another page (like a tabbed nav page, but still need to test this out a bit).

Experiments can be setup via Contentful and will activate an alternative when the page is loaded. In Contentful, an editor can select what "actions" or events to convert the experiment on; if none are specified, then it will convert on any of the possible actions.

This PR only adds the changes to include the `Sixpack` class, some constants, and also adds some new console logging for when a Sixpack test is participated or converted:

sixpack console log

The capacity to execute the conversions will happen in middleware (upcoming PR) which will listen for actions related to signups, posts, etc.

### DIFF
--- a/resources/assets/constants/index.js
+++ b/resources/assets/constants/index.js
@@ -20,3 +20,7 @@ export const NetworkStatus = {
   READY: 7,
   ERROR: 8,
 };
+
+export const SIXPACK_EXPERIMENT_SIGNUP_ACTION = 'signup';
+export const SIXPACK_EXPERIMENT_REPORTBACK_POST_ACTION = 'reportbackPost';
+export const SIXPACK_EXPERIMENT_CLICKED_BUTTON_ACTION = 'clickedButton';
--- a/resources/assets/services/Sixpack.js
+++ b/resources/assets/services/Sixpack.js
@@ -1,0 +1,140 @@
+/* global window */
+
+import client from 'sixpack-client';
+
+import {
+  SIXPACK_EXPERIMENT_SIGNUP_ACTION,
+  SIXPACK_EXPERIMENT_REPORTBACK_POST_ACTION,
+  SIXPACK_EXPERIMENT_CLICKED_BUTTON_ACTION,
+} from '../constants';
+import { sixpackLog } from '../helpers/loggers';
+
+class Sixpack {
+  constructor() {
+    this.reset();
+
+    const env = window.ENV || {};
+
+    if (!env.SIXPACK_BASE_URL) {
+      throw new Error('Missing Sixpack configuration settings.');
+    }
+
+    this.client = new client.Session({
+      base_url: env.SIXPACK_BASE_URL,
+      cookie_name: env.SIXPACK_COOKIE_PREFIX || 'sixpack',
+    });
+  }
+
+  conversions = [
+    SIXPACK_EXPERIMENT_SIGNUP_ACTION,
+    SIXPACK_EXPERIMENT_REPORTBACK_POST_ACTION,
+    SIXPACK_EXPERIMENT_CLICKED_BUTTON_ACTION,
+  ];
+
+  experiments = {};
+
+  /**
+   * Add experiment and supplied values to the experiments list.
+   *
+   * @param {String} experimentName
+   * @param {Object} experimentSettings
+   */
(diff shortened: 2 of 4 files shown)


resources/assets/services/Sixpack.js
```diff
@@ -0,0 +1,140 @@
...
+import {
+  SIXPACK_EXPERIMENT_SIGNUP_ACTION,
+  SIXPACK_EXPERIMENT_REPORTBACK_POST_ACTION,
+  SIXPACK_EXPERIMENT_CLICKED_BUTTON_ACTION,
+} from '../constants';
+import { sixpackLog } from '../helpers/loggers';
+
+class Sixpack {
+  constructor() {
+    this.reset();
```
mendelB: Why do we need to reset the `this.experiments` in the constructor, wouldn't it always be empty at this point?

resources/assets/constants/index.js
```diff
@@ -20,3 +20,7 @@ export const NetworkStatus = {
   READY: 7,
   ERROR: 8,
 };
+
+export const SIXPACK_EXPERIMENT_SIGNUP_ACTION = 'signup';
+export const SIXPACK_EXPERIMENT_REPORTBACK_POST_ACTION = 'reportbackPost';
+export const SIXPACK_EXPERIMENT_CLICKED_BUTTON_ACTION = 'clickedButton';
```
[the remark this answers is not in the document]
DFurnes: That makes sense to me. I do sometimes wonder about the developer ergonomics of exporting and importing constants (as opposed to throwing a runtime exception on invalid input or ✨ type checking ✨). But I definitely see the argument in favor of 'em!

resources/assets/services/Sixpack.js
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * Participate in specified experiment.
+   *
+   * @param  {String} experimentName
+   * @param  {Array}  testAlternatives
+   * @param  {Object} options
+   * @return {Promise}
+   */
+  participate(experimentName, testAlternatives = [], options = {}) {
```
mendelB: Are we removing support for the `trafficFraction` param? (I know you had it as a @TODO on the `participateBeta` you added)

resources/assets/services/Sixpack.js
weerd: Yeah, but just in case™ 😉

weerd: Didn't feel strongly about removing it, but if you do feel free to let me know!

resources/assets/services/Sixpack.js
weerd: `trafficFraction` can be passed as part of the `options` now. along with `kpi` and any other options 💃

weerd:
> So this Singleton is going to replace the /helpers/experiments, and the redux / experiments middleware?

Yep yep! I'll be cleaning the redux store out and any old code related to the old approach once this new approach is all set!

resources/assets/services/Sixpack.js
mendelB: Nah I was just curious if there was some fluke you'd bumped into. I always love me some

resources/assets/services/Sixpack.js
mendelB: Oh, nice! Where does that get passed over to `sixpack` though?

resources/assets/constants/index.js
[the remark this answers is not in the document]
weerd: Yeah, if we feel it doesn't end up helping out, we can remove. Just something I wanted to try out for now :)

resources/assets/services/Sixpack.js
weerd: That's in the next part coming up. It's going to be a field on the `sixpackExperiment` content type in Contentful 💃

PR #1058, opened 2018-08-08
Visual Bug Fix for TSA Field Labels

mendelB:
### What does this PR do?

This PR fixes a visual bug that cropped up in the `TextSubmissionAction` component where field labels were getting cut off.

- adds a scss file for the `TextSubmissionAction` with a rule to fix the `.field-label`

Refs [Pivotal ID #159628338](https://www.pivotaltracker.com/story/show/159628338)

image
image


weerd:
I also wonder if it would be worth targeting a class across action types like the PSA, TSA and RSA and fixing it across them? 🤔

mendelB:
Would we do that by adding a new additional class, or implementing this rule across all `.field-label`s?

weerd:
I would say a new class like `submission-action` or something that can be applied to all. But let's leave like this for now. It might be a moot point in the near-ish future when we switch to Tailwind CSS...

PR #1059, opened 2018-08-08
Migration to add the sixpack experiment content type.

weerd:
### What does this PR do?

This PR adds a migration to add the new Sixpack Experiments content type into the system:

image

### DIFF
--- a/contentful/migrations/2018_08_08_001_add_sixpack_experiment_content_type.js
+++ b/contentful/migrations/2018_08_08_001_add_sixpack_experiment_content_type.js
@@ -1,0 +1,47 @@
+module.exports = function(migration) {
+  const sixpackExperiment = migration
+    .createContentType('sixpackExperiment')
+    .name('Sixpack Experiment')
+    .description('Used to help set up A/B testing experiments.')
+    .displayField('internalTitle');
+
+  sixpackExperiment
+    .createField('internalTitle')
+    .name('Internal Title')
+    .type('Symbol')
+    .required(true)
+    .localized(false);
+
+  sixpackExperiment
+    .createField('title')
+    .name('Title')
+    .type('Symbol')
+    .required(true)
+    .localized(true);
+
+  sixpackExperiment
+    .createField('alternatives')
+    .name('Test Alternatives')
+    .type('Array')
+    .items({
+      type: 'Link',
+      linkType: 'Entry',
+    })
+    .validations([{ size: { min: 2 } }])
+    .required(true)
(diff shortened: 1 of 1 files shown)


contentful/migrations/2018_08_08_001_add_sixpack_experiment_content_type.js
```diff
@@ -0,0 +1,47 @@
...
+      linkType: 'Entry',
+    })
+    .validations([{ size: { min: 2 } }])
+    .required(true)
+    .localized(false);
+
+  sixpackExperiment
+    .createField('trafficFraction')
+    .name('Traffic Fraction')
+    .type('Integer')
```
mendelB: Do we want validations for this field? Like will it be 1-100 or something?

contentful/migrations/2018_08_08_001_add_sixpack_experiment_content_type.js
```diff
@@ -0,0 +1,47 @@
...
+    .required(false)
+    .localized(false);
+
+  sixpackExperiment
+    .createField('kpi')
```
mendelB: I'm not sure I completely understand the `kpi` thing but isn't it the ability to convert an experiment against multiple KPI's? How would this work for us?

contentful/migrations/2018_08_08_001_add_sixpack_experiment_content_type.js
weerd: Oh yeah, potentially. I was going to end up taking whatever integer gets passed and in the entity file, dividing by 100. If the value is anything not above 0, then setting it to `null` and assuming default traffic fraction of 100. But I can see validation here being helpful :)

contentful/migrations/2018_08_08_001_add_sixpack_experiment_content_type.js
weerd: I still don't fully get how `kpi` works in Sixpack, but it's a param that can be passed and I figured we can start using it and figure out how it should work. I tried it locally and seems like it could be useful:

image

Basically it adds a dropdown where you can select by the KPI... whatever that means ;)

weerd:
Updated migration:
image

contentful/migrations/2018_08_08_001_add_sixpack_experiment_content_type.js
mendelB: I freaking love business jargon 🕴

PR #1065, opened 2018-08-13
Sixpack component middleware updates

weerd:
### What does this PR do?

This PR wraps up the final puzzle piece with the new Sixpack Experiments component and approach.

Actions that are considered "convert-able" should pass a `meta` property called `sixpackExperments.conversion` with a value designating the type of conversion as seen here.

The middleware picks up on this property, and if found it will convert all current experiments that allow conversion on that property's value (for example on `reportbackPost`).

This approach lets us use the `SixpackExperiment` component for both Contentful-based tests by an editor, and in code tests by a developer (for more complex tests or tests not feasible via Contentful).

```js
// for experiments that come from Contentful:
<SixpackExperiment {...json} />

// for experiments defined in code:
<SixpackExperiment
  title="developer test in code"
  alternatives={[
    { ...jsonFromContentful },
    AlternatePage,
    () => <div>simple third option!</div>,
  ]}
/>
```
In an experiment defined in code, you could pass a variety of types in the alternatives array that would end up rendering out the proper component.

### Checklist


resources/assets/components/utilities/SixpackExperiment/SixpackExperiment.js
```diff
@@ -66,10 +66,17 @@ class SixpackExperiment extends React.Component {
...
   render() {
-    return this.state.selectedAlternative ? (
-      <ContentfulEntry json={this.state.selectedAlternative} />
+    const selectedAlternative = this.state.selectedAlternative;
+
+    if (!selectedAlternative) {
+      return <Placeholder />;
+    }
+
+    return typeof selectedAlternative === 'object' &&
```
weerd: I wasn't sure if there was maybe a better way to test against types here. I checked and React components register as `object` as well, thus the need to check inside the object for an `id` which would _very likely_ mean it's a contentful based json object.

DFurnes: Perhaps [`_.isPlainObject`](https://lodash.com/docs/4.17.10#isPlainObject)?

DFurnes: Or the inverse, [`React.isValidElement`](https://reactjs.org/docs/react-api.html#isvalidelement)!

weerd: dang. I always forget to consider Lodash magic ✨ Good cal @DFurnes!!

I think maybe `_.isPlainObject`... would `React.isValidElement` work if passed an anonymous function that renders out html? `() => <div>simple third option!</div>`

DFurnes: Yeah, I feel good about checking for a plain object & rendering w/ `ContentfulEntry` if so.

weerd:
> If one of the alternatives is a node or function, wouldn't we have trouble trying to grab the fields.title from it in our componentDidMount?

@mendelB yeah, tis a good point. I'm working on a _in code_ experiment next with the Lede Banners and can tackle approaching that in a real world example. For the time being this currently works with Contentful tests, but will get it to work w/ dev tests as I try and actual example! Defs a good note!

weerd:
Turns our `React.isValidElement` was the better approach. Switched it in an upcoming PR :)

PR #1066, opened 2018-08-13
Lede banner sixpack experiment

weerd:
### What does this PR do?

This PR uses the new [`SixpackExperiment`](https://github.com/DoSomething/phoenix-next/tree/master/resources/assets/components/utilities/SixpackExperiment) component to easily add an in-code Sixpack experiment. We no longer need to create duplicates/alternates of the component we want to test and named slightly different. Instead, with the new setup, we can just pull from our library of components.

To help make this feasible, I updated some of the behavior in the `LedeBannerContainer` component. When calling this component directly, it automagically includes props for the subsequent `LedeBanner` component that gets rendered. However, if you passed properties to override what the `LedeBannerContainer` sets, it wouldn't work, since the props the component sets override anything you pass in as props.

Example:
```js
<LedeBannerContainer title="Some new title for the banner!" />
// The title still ends up being what is set in the container and not this prop!
```
But by altering the props set in the container as such:
```js
title: get(props, 'title', state.campaign.title),
```
It will prefer the prop passed, and if none was passed, will default to the value the container originally set.

You can see in the [experiment for this PR](https://github.com/DoSomething/phoenix-next/pull/1066/files#diff-3d3cc00faaa209b063033174e7dddcacR36), the `coverImage` property is overridden on the `LedeBannerContainer`.

This could be useful behavior to implement in other component containers when we want to provide override-able properties.

[Small screen](https://user-images.githubusercontent.com/105849/44103913-7341c662-9fbb-11e8-8528-8dfee8b235b6.png)

[Medium screen](https://user-images.githubusercontent.com/105849/44103942-86deb612-9fbb-11e8-9669-62a93df9cdbb.png)

[Large Screen](https://user-images.githubusercontent.com/105849/44103960-947af5b0-9fbb-11e8-83e5-450725295752.png)

Refs [Pivotal ID #158806897](https://www.pivotaltracker.com/story/show/158806897)

### Checklist


resources/assets/components/pages/LandingPage/LandingPage.js
```diff
@@ -5,71 +5,49 @@ import PropTypes from 'prop-types';
...
-          subtitle={subtitle}
-          blurb={blurb}
-          coverImage={coverImage}
-          legacyCampaignId={legacyCampaignId}
-          endDate={endDate}
-          template={template}
-          affiliateSponsors={affiliateSponsors}
-          signupArrowContent={signupArrowContent}
-          showPartnerMsgOptIn={showPartnerMsgOptIn}
+      {campaignId === '3pwxnRZxociqMaQCMcGOyc' ? (
```
weerd: Not the most ideal, but we're specifically targeting this single campaign for this experiment, so I figured it was alright. Will need to consider how to best handle something like this in the future, or if wanting to target a handful of specific campaigns... 🤔

mendelB: We never really formalized this -- but could the `additionalContent.featureFlags` be useful for cases like this?

à la what we do in the Campaign component:

https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/Campaign/Campaign.js#L38
https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/Campaign/CampaignContainer.js#L29

weerd: Yeah, potentially. Just debating if this is an entirely developer setup experiment, it might make sense to keep it all in code, rather than have an item in the CMS that lingers if we forget to remove it... it's a possibility though 🤔 Good thinking :)

weerd: Also if the `SixpackExperiment` is set to appear if there is something set for `additionalContent.featureFlags`, it could mean some tests would appear in same place unexpectedly (since `<SixpackExperiment />` is added manually with some specified props, but the feature flag could accidentally be added to other campaigns in Contentful. I think having the single place to control these non-dynamic experiments out-weighs the possible mixups or bugs.

mendelB: Def see your point regarding one centralized location for static information (#ryhmes).
I guess if we had a more formalized system for the feature flagging it might make more sense?

resources/assets/components/LedeBanner/LedeBannerContainer.js
```diff
@@ -19,13 +19,13 @@ const mapStateToProps = state => ({
     null,
   ),
   blurb: state.campaign.blurb,
```
mendelB: Can the `blurb` not be overridden from props?

resources/assets/components/LedeBanner/templates/JumboTemplate.js
```diff
@@ -0,0 +1,79 @@
...
+      <div className="wrapper margin-horizontal-auto">
+        <h1 className="jumbo-lede-banner__headline-title">{title}</h1>
+
+        <h2 className="jumbo-lede-banner__headline-subtitle">{subtitle}</h2>
+
+        {blurb ? (
+          <Markdown className="jumbo-lede-banner__blurb">{blurb}</Markdown>
+        ) : null}
+
+        {isAffiliated ? null : (
```
mendelB: Do we not want the `affiliatedActionLink` button for this template?

resources/assets/components/pages/LandingPage/LandingPageContainer.js
```diff
@@ -11,19 +11,21 @@ const mapStateToProps = state => {
   const landingPage = state.campaign.landingPage.fields;
 
   return {
```
mendelB: Should we also discard all the unused props from this container?

mendelB:
Is the `cover` template kaput or is the `jumbo` template just a slightly different alternative?

resources/assets/components/LedeBanner/templates/JumboTemplate.js
weerd: I ignored it for now. Is it being used? Also, I think we should start keeping a list of `additionalContent` feature flags that we want to eventually formalize...

resources/assets/components/LedeBanner/LedeBannerContainer.js
weerd: Oh yeah, it should be, but in this experiment it's not going to be. I'll update that!

weerd:
> Is the cover template kaput or is the jumbo template just a slightly different alternative?

Likely kaput, but since `jumbo` was different enough and not to confuse I wanted to have a different one, and with templates it makes it easy. Will likely be deleting `cover` in due time.

resources/assets/components/LedeBanner/templates/JumboTemplate.js
weerd: Cool. I'm going to leave off and if it turns out we want it back in, it'll be an opportunity to formalize it!

weerd: Defs good note. Meant to mention that in the PR discussion and totes forgot! 😬

resources/assets/components/pages/LandingPage/LandingPageContainer.js
weerd: Oh jes! 👍

weerd: All set 💃

PR #1069, opened 2018-08-16
Sixpack experiments purge old code

weerd:
### What does this PR do?

This PR removes a bunch 🍇 of the old Sixpack experiments stuff and additional cruft that built up over time both on the JS and PHP sides.

It also exposes the Sixpack class via `window.Sixpack`, and remove the `window.sixpack` that the Sixpack NPM package makes available, but we don't use. It also allows us to use `window.Sixpack` on Prod now, which will be useful!

🛁 lots o' cleaning!

🌵


weerd:
😂

PR #1074, opened 2018-08-24
Adding new Person content type.

weerd:
### What does this PR do?

This PR adds a new `Person` content type migration that will replace the existing `Staff` content type.

🤺 

### DIFF
--- a/contentful/content-types/person.js
+++ b/contentful/content-types/person.js
@@ -1,0 +1,57 @@
+module.exports = function(migration) {
+  const person = migration
+    .createContentType('person')
+    .name('Person')
+    .description('A person employed or associated with DoSomething.org.')
+    .displayField('name');
+
+  person
+    .createField('name')
+    .name('Name')
+    .type('Symbol')
+    .required(true)
+    .localized(false);
+
+  person
+    .createField('type')
+    .name('Type')
+    .type('Symbol')
+    .validations([
+      {
+        in: ['staff', 'intern', 'board member', 'advisory board member'],
+      },
+    ])
+    .required(true)
+    .localized(true);
+
+  person.changeEditorInterface('type', 'radio');
+
+  person
+    .createField('jobTitle')
+    .name('Job Title')
+    .type('Symbol')
+    .required(false)
+    .localized(true);
+
+  person
+    .createField('email')
+    .name('Email')
+    .type('Symbol')
+    .required(false)
+    .localized(false);
+
+  person
+    .createField('avatar')
+    .name('Avatar')
+    .type('Link')
+    .linkType('Asset')
+    .required(false)
+    .localized(false);
+
+  person
+    .createField('description')
+    .name('Description')
+    .type('Text')
+    .required(false)
+    .localized(true);
+};


contentful/content-types/person.js
```diff
@@ -0,0 +1,57 @@
...
+
+  person
+    .createField('name')
+    .name('Name')
+    .type('Symbol')
+    .required(true)
+    .localized(false);
+
+  person
+    .createField('type')
```
weerd: `type` work as the name for this field?

contentful/content-types/person.js
```diff
@@ -0,0 +1,57 @@
...
+
+  person
+    .createField('email')
+    .name('Email')
+    .type('Symbol')
+    .required(false)
+    .localized(false);
+
+  person
+    .createField('avatar')
```
weerd: Figured I'd stay with `avatar` but lemme know if maybe we prefer `photo`. Also, probably need to add `alternateAvatar` or `alternatePhoto` field for the baby pics 🤔

DFurnes: I _very slightly_ prefer `photo` I think, just since it's less jargony. But no strong feels!

contentful/content-types/person.js
DFurnes: Makes sense to me.

contentful/content-types/person.js
```diff
@@ -0,0 +1,57 @@
...
+  person
+    .createField('type')
+    .name('Type')
+    .type('Symbol')
+    .validations([
+      {
+        in: ['staff', 'intern', 'board member', 'advisory board member'],
+      },
+    ])
+    .required(true)
```
DFurnes: Does this mean someone has to be one of those four options? What about prior staffers/interns?

contentful/content-types/person.js
weerd: I think I might change it to `photo` just cause `alternateAvatar` just sounds a bit weird to me.

contentful/content-types/person.js
```diff
@@ -0,0 +1,79 @@
...
+
+  person
+    .createField('description')
+    .name('Description')
+    .type('Text')
+    .required(false)
+    .localized(true);
+
+  person
+    .createField('advisoryCommittee')
```
weerd: Needed to add this field since the [advisory board is split into different categories](https://www.dosomething.org/us/about/advisory-board).

But now that I think of it, maybe we call this something different like `team` and then allow options for the staff as well. So among the options for committees if the person type is `advisory board member`, if the person is a `staff` then we could have options for `product`, `tech`, `campaigns`, etc?

Thoughts? Also thoughts on name... maybe `category` instead of `team` to be extra generic?

contentful/content-types/person.js
mendelB: [image](https://www.youtube.com/watch?v=kXraSkgssFk)

contentful/content-types/person.js
weerd: Suggestions? Not make it required or add those as options? `alumni`, `former staff`, `former intern`?

contentful/content-types/person.js
weerd: LOL 😂

contentful/content-types/person.js
mendelB: If the new Advisory board web pages are just going to be `Page`s, and we're linking these `Person`'s in `blocks`, do we need to categorize them on the Entries themselves, or can we just put the appropriate people in their Pages?

weerd: Yeah, that's true and makes things simpler. Okie doke. Will remove!

contentful/content-types/person.js
mendelB: I think `former staff`, `former intern` makes sense. Or we could add an `active` or `current` -or something- boolean field?

weerd: Oh yeah. I think I like the suggestion for something like a `status` and can be `active` or `inactive`.

While `former staff` and `former intern` could work, then we would also have to consider `former *` for board and advisory members. At least with a `status` we can keep type to be continuous and just indicate if still around or not.

weerd: Errr, duh. Can't be `status` if boolean. Will make it `active`. Unless there's a better name for that. I didn't feel `current` would be as obvious as `active` @mendelB

mendelB: Haha true fact, that'd be awkward to read!
Hmm I *think* active seems right. `employed` and such is also staff specific.

weerd: Yeah. I guess `active` makes the most sense! Good suggestion for a boolean tho! 👍

PR #1082, opened 2018-08-28
Sixpack landing page bug fix

weerd:
### What does this PR do?

This PR finalizes the changes to Landing Pages to get them working properly with Sixpack Experiments. Moving forward, all campaign landing pages need to use the `landingPage` content type!

**Example landing page with Alt A**
image

**Example landing page with Alt B**
image

Most of the changes happened in #1084, and this PR finalizes to add the Sixpack support.

Refs [Pivotal ID #159907997](https://www.pivotaltracker.com/story/show/159907997)

### Checklist

### DIFF
--- a/resources/assets/components/pages/LandingPage/LandingPageContainer.js
+++ b/resources/assets/components/pages/LandingPage/LandingPageContainer.js
@@ -7,7 +7,7 @@ import LandingPage from './LandingPage';
  * Provide state from the Redux store as props for this component.
  */
 const mapStateToProps = (state, ownProps) => {
-  console.log(ownProps);
+  const landingPage = get(ownProps, 'landingPage.fields', ownProps);
 
   return {
     campaignId: state.campaign.id,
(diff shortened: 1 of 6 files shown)


weerd:
Also tested to make sure conversion works on signup and it does! 💃

resources/assets/components/pages/LandingPage/LandingPageContainer.js
```diff
@@ -6,8 +6,11 @@ import LandingPage from './LandingPage';
 /**
  * Provide state from the Redux store as props for this component.
  */
-const mapStateToProps = state => {
-  const landingPage = state.campaign.landingPage.fields;
+const mapStateToProps = (state, ownProps) => {
+  // @TODO: while we have landing pages as either a page content type
+  // or a landingPage content type, the ownProps is structured a bit
+  // differently. Revise once all landing pages use landingPage type.
+  const landingPage = get(ownProps, 'landingPage.fields', ownProps);
```
mendelB: Just to confirm I understand this correctly:
If the Landing Page is the old `Page` type - the fields come through directly from `CampaignPage` and we thus need to grab the `ownProps.landingPage.fields`.
But if the Landing Page is a `LandingPage` content type, then the fields come through from `ContentfulEntry` which passes the `json.fields` - thus we can defualt to `ownProps` directly?

weerd: Yep!

PR #1083, opened 2018-08-28
Adding Get Out the Vote link in the footer

ngjo:
### What does this PR do?

Adds the Get out the vote link to the footer. 

This was a request that came out of today's management meeting.

N/A

### DIFF
--- a/resources/views/partials/footer.blade.php
+++ b/resources/views/partials/footer.blade.php
@@ -49,6 +49,7 @@
     <div class="footer__column -links">
         <h4>Get Involved</h4>
         <ul>
+          <li><a href="https://vote.dosomething.org/workwithus">Get Out the Vote!</a></li>
           <li><a href="{{ phoenixLink('us/about/easy-scholarships') }}">Scholarships</a></li>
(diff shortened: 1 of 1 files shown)


mendelB:
@ngjo Do we need to add this to Ashes (cough) as well?

resources/views/partials/footer.blade.php
```diff
@@ -49,6 +49,7 @@
     <div class="footer__column -links">
         <h4>Get Involved</h4>
         <ul>
+          <li><a href="https://vote.dosomething.org/workwithus">Get Out the Vote!</a></li>
```
mendelB: Does this open in a new tab? And if not, would we want it to, or doesn't matter?

ngjo: I believe the SEO best practice is that if it opens on the same platform, it should be in the same tab. So, I guess the question is if we count Instapage as our platform...in this case I think I'd keep it in this tab since this instapage is DS branded. But I don't feel particularly strongly either way!

ngjo:
> Do we need to add this to Ashes (cough) as well?

Already added! How 'bout that irony that an Ashes change was easier because I was afraid of the codes 😝

PR #1084, opened 2018-08-28
Landing page entity and logic

weerd:
### What does this PR do?

This PR updates how we render out Landing Pages for a campaign. It adds a new `LandingPage` entity to take advantage of the new `landingPage` content type on Contentful. Moving forward we want to use this content type instead of the `page` content type so we can know better when we're dealing with the uniqueness of a landing page.

If a landing page is of type `page` it will render out using the old approach, otherwise if it is of type `landingPage` it will render using the new approach.

It now uses the `ContentfulEntry` component to render out the landing page (which sets things up for an upcoming Sixpack bug fix for landing pages). This PR also fixes some naming to rename the `pitchContent` variable to just be `content`.

Refs [Pivotal ID #159907997](https://www.pivotaltracker.com/story/show/159907997)

### Checklist

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -86,7 +86,7 @@ class Campaign extends Entity implements JsonSerializable
      */
     public function parseLandingPage($landingPage)
     {
-        if (!$landingPage) {
+        if (! $landingPage) {
             return null;
         }
 
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -79,6 +79,30 @@ class Campaign extends Entity implements JsonSerializable
     }
 
     /**
+     * Parse the landing page for the campaign.
+     *
+     * @param  DynamicEntry $landingPage
+     * @return array
+     */
+    public function parseLandingPage($landingPage)
+    {
+        if (!$landingPage) {
+            return null;
+        }
+
+        if ($landingPage->entry->getContentType()->getId() === 'sixpackExperiment') {
+            return new SixpackExperiment($landingPage->entry);
+        }
+
+        // @TODO: remove once all landing pages use the LandingPage content type.
+        if ($landingPage->entry->getContentType()->getId() === 'page') {
+            return new Page($landingPage->entry);
+        }
+
+        return new LandingPage($landingPage->entry);
+    }
+
+    /**
      * Convert the object into something JSON serializable.
      *
      * @return array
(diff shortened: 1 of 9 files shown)


resources/assets/components/pages/CampaignPage/CampaignPage.js
```diff
@@ -17,9 +17,18 @@ import './campaign-page.scss';
...
  * @returns {XML}
  */
-const CampaignPage = props =>
-  props.shouldShowLandingPage ? (
-    <LandingPageContainer {...props} />
+const CampaignPage = props => {
+  // @TODO: temporary variable to select component to use based on type.
+  // Will be removed once all landing pages use the LandingPage content type.
+  const landingPageComponent =
+    props.landingPage.type === 'page' ? (
```
mendelB: Should we have some fail-safe in case `props.landingPage` is `null`?

app/Entities/Campaign.php
```diff
@@ -78,6 +78,30 @@ public function parseCampaignLead($campaignlead, $additionalContent)
         ];
     }
 
+    /**
+     * Parse the landing page for the campaign.
+     *
+     * @param  DynamicEntry $landingPage
+     * @return array
+     */
+    public function parseLandingPage($landingPage)
```
mendelB: Do we not want to use `parseBlock` here because the `LandingPage` is only ever a property of a `Campaign`?

app/Entities/Campaign.php
```diff
@@ -78,6 +78,30 @@ public function parseCampaignLead($campaignlead, $additionalContent)
...
+    {
+        if (! $landingPage) {
+            return null;
+        }
+
+        if ($landingPage->entry->getContentType()->getId() === 'sixpackExperiment') {
```
mendelB: I think if you call `$landingPage->getContentType()` (without accessing `entry`) you get the `id` for free cuz of https://github.com/DoSomething/phoenix-next/blob/master/app/Entities/Entity.php#L156.

resources/assets/components/pages/CampaignPage/CampaignPage.js
```diff
@@ -17,9 +17,18 @@ import './campaign-page.scss';
...
+      <LandingPageContainer {...props} />
```
mendelB: I forget, was there a reason we even bothered passing through `...props` to the `LandingPageContainer`, since it grabs all the `props` from `state` anyway?

resources/assets/components/pages/CampaignPage/CampaignPage.js
weerd: oh wait. You mean since it's out of the return ternary. I see what you mean now!

app/Entities/Campaign.php
weerd: Ya. Plus need to do some extra logic in here to deal w/ Sixpack stuff for upcoming PR.

resources/assets/components/pages/CampaignPage/CampaignPage.js
weerd: 🤷‍♂️ Just kept it as we were doing it before. Aiming to get rid of this in near future anyhoo

weerd:
@mendelB switched to have it [use a function instead](https://github.com/DoSomething/phoenix-next/pull/1084/commits/0ecfa067efd0f4ce51dc9248dcd7a78f3098b16d). This way, it doesn't get called if `props.shouldShowLandingPage` is false 💃

weerd:
> I think if you call $landingPage->getContentType() (without accessing entry) you get the id for free

Great call @mendelB! I switched this up as well, but forgot to push. Just pushed it and will merge :)

PR #1085, opened 2018-08-28
Contentful SDK update Part 1

mendelB:
### What does this PR do?

This PR updates our Contentful packages ([contentful/contentful](https://github.com/contentful/contentful.php), and [contentful/laravel](https://github.com/contentful/contentful-laravel)) to `2.0`

This is v1 of our upgrade process. We were wayyyyy behind, and the `3.0` upgrade has a lot more breaking changes, so figured it'd be a good idea to go step by step.

This one was pretty straight-forward, and we only had to deal with [these breaking changes](https://github.com/contentful/contentful.php/blob/master/CHANGELOG.md#changed-5)

Refs [Pivotal ID #](https://www.pivotaltracker.com/story/show/158440484)

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -1,9 +1,9 @@
 <?php
 
 use App\Entities\Campaign;
-use Contentful\ImageOptions;
 use Contentful\Delivery\Asset;
 use App\Services\PhoenixLegacy;
+use Contentful\File\ImageOptions;
 use Illuminate\Support\HtmlString;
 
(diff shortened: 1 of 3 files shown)


weerd:
> This is v1 of our upgrade process. We were wayyyyy behind, and the 3.0 upgrade has a lot more breaking changes, so figured it'd be a good idea to go step by step.

defs! good call 👍

PR #1086, opened 2018-08-29
Adding Person entity parsing.

weerd:
### What does this PR do?

This PR adds an entity for the `Person` class which will parse the `person` content-type from Contentful.

🌵 

### Checklist

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -59,14 +59,10 @@ class Campaign extends Entity implements JsonSerializable
      */
     public function parseCampaignLead($campaignlead, $additionalContent)
     {
-        if ($campaignlead->getContentType() === 'person') {
-            return new Person($campaignlead->entry);
-        }
-
-        // @TODO (2018-08-29): Should eventually be removed and replaced by the
-        // Person content type and class.
-        if ($campaignlead->getContentType() === 'staff') {
-            return new Staff($campaignlead->entry);
+        if ($campaignlead) {
+            // @TODO (2018-08-29): Staff should eventually be removed and replaced by the
+            // Person content type and class for all entries.
+            return $campaignlead->getContentType() === 'staff' ? new Staff($campaignlead->entry) : new Person($campaignlead->entry);
         }
 
         // @TODO (2018-08-29): we should do away with this additional content item.
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -59,10 +59,17 @@ class Campaign extends Entity implements JsonSerializable
      */
     public function parseCampaignLead($campaignlead, $additionalContent)
     {
-        if ($campaignlead) {
+        if ($campaignlead->getContentType() === 'person') {
+            return new Person($campaignlead->entry);
+        }
+
+        // @TODO (2018-08-29): Should eventually be removed and replaced by the
+        // Person content type and class.
+        if ($campaignlead->getContentType() === 'staff') {
             return new Staff($campaignlead->entry);
         }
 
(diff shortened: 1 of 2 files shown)


app/Entities/Person.php
```diff
@@ -0,0 +1,31 @@
...
+            'id' => $this->entry->getId(),
+            'type' => $this->getContentType(),
+            'fields' => [
+                'name' => $this->name,
+                'type' => $this->type,
+                'active' => $this->active,
+                'jobTitle' => $this->jobTitle,
+                'email' => $this->email,
+                'photo' => get_image_url($this->photo, 'square'),
+                'alternatePhoto' => get_image_url($this->alternatePhoto, 'square'),
```
weerd: I think we've debated this before but would it be useful to make the `alternatePhoto` and `photo` be an array with properties like `square` and `thumb` which has URLs to the different sizes instead of just defaulting to the `square` image and needing to re-request from Contentful in the component when we want a thumbnail size? 🤔 Seems like it could be useful...

app/Entities/Campaign.php
```diff
@@ -59,10 +59,17 @@ public function parseAffiliates($affiliates)
      */
     public function parseCampaignLead($campaignlead, $additionalContent)
     {
-        if ($campaignlead) {
+        if ($campaignlead->getContentType() === 'person') {
+            return new Person($campaignlead->entry);
+        }
+
+        // @TODO (2018-08-29): Should eventually be removed and replaced by the
```
mendelB: Dated `@TODO`s -- neat!

app/Entities/Campaign.php
```diff
@@ -59,10 +59,17 @@ public function parseAffiliates($affiliates)
      */
     public function parseCampaignLead($campaignlead, $additionalContent)
     {
-        if ($campaignlead) {
+        if ($campaignlead->getContentType() === 'person') {
```
mendelB: Do we have Campaigns where the `campaignLead` field is `null` -- meaning I guess that it used to default to `additionalContent`, which'd error out when we call `getContentType`?

app/Entities/Person.php
mendelB: That sounds like it'd be useful! What about the `contentfulImageUrl` helper method we have on the front-end? Could we alternatively pass through the default image URL, and then let us do the processing on the front-end, or on second thought -- is it more useful to have the more robust processing we get on the back-end?

weerd: That's partly my question. Might be useful to have `square` and `thumb` made available via the backend (we already were doing it technically for the square version) since we mostly use those for person photo images. And have the `contentfulImageUrl` process on the front-end for less usual usage?...

app/Entities/Campaign.php
weerd: Trying it out. Might make us more accountable ;)

app/Entities/Campaign.php
weerd: Dang! I keep doing that and you keep catching it. haha 🤦‍♂️

app/Entities/Person.php
DFurnes: Do we need to re-request from the backend for different crops? I thought that was something we could handle client-side with query parameters (e.g. `?w=200&h=200` for thumb?)

weerd: True, but we're technically doing that with the `square`. And for the most part we ended up using the "thumb" more often than the full square image... Maybe default URL like @mendelB said and then the front-end decides all crops?

DFurnes: I'd definitely cast my vote for handling client-side with the helper if we can, since it keeps the "decision" for formatting that image closer to where it's used.

PR #1089, opened 2018-08-31
Updates to address using default URL in both Person and Staff.

weerd:
### What does this PR do?

This PR updates the `Staff` entity to return the default author photo URL and not a pre-formatted one. Instead we want the front-end React components to decide what sizing of a photo they want using the Contentful Image API.

It also makes it so that Campaign Updates and Affirmations support either using a `staff` or `person` content type. The `staff` content type will be deprecated but to make the transition easier, we are supporting both until everyone is shift over to be a `person`.

This PR also updates references to use `photo` instead of `avatar` which is the new name for the person's photo in the `person` content type.

### Checklist

### DIFF
--- a/resources/assets/components/CampaignUpdate/CampaignUpdate.js
+++ b/resources/assets/components/CampaignUpdate/CampaignUpdate.js
@@ -9,6 +9,7 @@ import Embed from '../utilities/Embed/Embed';
 // import Share from '../utilities/Share/Share';
 import Byline from '../utilities/Byline/Byline';
 import SponsorPromotion from '../SponsorPromotion';
+import { contentfulImageUrl } from '../../helpers';
 import Markdown from '../utilities/Markdown/Markdown';
 
 const CampaignUpdate = props => {
@@ -26,6 +27,7 @@ const CampaignUpdate = props => {
   } = props;
 
   const authorFields = get(author, 'fields', {});
+  const authorPhoto = authorFields.photo || undefined;
 
   const isTweet = content && content.length < 144;
 
@@ -61,7 +63,11 @@ const CampaignUpdate = props => {
         ) : (
           <Byline
             author={authorFields.name}
-            avatar={authorFields.avatar || undefined}
+            photo={
+              authorPhoto
+                ? contentfulImageUrl(authorPhoto, 175, 175, 'fill')
+                : undefined
+            }
             jobTitle={authorFields.jobTitle || undefined}
             className="float-left"
           />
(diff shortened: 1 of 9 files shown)


resources/assets/components/utilities/Byline/Byline.js
```diff
@@ -34,15 +34,15 @@ Byline.propTypes = {
   author: PropTypes.string,
   className: PropTypes.string,
   jobTitle: PropTypes.string,
-  avatar: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+  photo: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
DFurnes: Hmm, do you remember what the `PropTypes.object` is used for here?

resources/assets/components/CampaignUpdate/CampaignUpdate.test.js
```diff
@@ -6,7 +6,7 @@ import CampaignUpdate from './CampaignUpdate';
 
 const author = {
   fields: {
-    avatar: 'http://example.com/avatar-aang.jpg',
+    photo: 'http://example.com/avatar-aang.jpg',
```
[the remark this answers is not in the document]
weerd: 😂

resources/assets/components/utilities/Byline/Byline.js
weerd: Not 100% sure... should I remove now that we specifically want a string? This actually brings up a good point I forgot to ask about.

I initially had the call to `contentfulImageUrl()` within the `Byline` component, but then decided against it since the default image is just an svg DS Logo and not something to process, which made me think that seemed right, since the `Byline` should only care about getting a URL to an image. So I popped it in the `CampaignUpdate` component, but wanted to make sure that felt like the right approach to y'all... thoughts? 🤔

resources/assets/components/CampaignUpdate/CampaignUpdate.js
```diff
@@ -61,7 +63,11 @@ const CampaignUpdate = props => {
         ) : (
           <Byline
             author={authorFields.name}
-            avatar={authorFields.avatar || undefined}
+            photo={
+              authorPhoto
+                ? contentfulImageUrl(authorPhoto, 175, 175, 'fill')
```
mendelB: Something of a random side note -- but might it be nice to eventually have some formal formatting options we could pass to the `contentfulImageUrl` akin to the options we have on the PHP helper, as we move more and more of the image URL formatting to the front end?

weerd: @mendelB ya, I dig that. Helps to keep consistency in the sizes referenced. Mind making a ticket in PT for it in the Backlog?

PR #1091, opened 2018-09-07
Add management API scripts to derive Persons from Staff entries

mendelB:
### What does this PR do?

This PR adds a contentful-migration-api script that simply creates new `Person` entries for each of our existing legacy `Staff` entries. **This was successfully run on Master**

It also adds a script to replace any link to the legacy Staff, with a link to its equivalent Person entry. **This was not run yet. This is actually consequential, so pending approval** (Ran it on a `test` environment and it executed successfully.)

Also:
- updates the client to accept an `environment-id` argument to specify environment (and defaults to `dev` for safety)
- Adds `member` option to list of validations for `Person.type` (we have a few `Staff` entries that seem to be members (wasn't sure whether to characterize as MC or VIP or Partner, so just went with `member` for now))

Figured we can split the Staff -> Person migration into a few parts:
1. simply script to create Person copies of existing Staff **DONE**
2. update the Person entries to add the additional new fields. (`type`, `active`) **DONE using Contentful UI** (PR updates migration file accordingly. (If we don't like the `member` choice we can totally change to wtvr.))
3. find all entries linking to staff entries and swap those for the person entry of the same name. **Pending Approval**

### DIFF
--- a/contentful/management-api-scripts/2018_09_07_002_replace_staff_linked_entries_with_person.js
+++ b/contentful/management-api-scripts/2018_09_07_002_replace_staff_linked_entries_with_person.js
@@ -20,21 +20,11 @@ async function replaceStaffLinkedEntriesWithPerson(environment, staff) {
 
   logger.info(`Processing Staff! [ID: ${staff.sys.id}] [Name: ${staffName}]\n`);
 
-  let people;
-
-  if (staffName === 'Margot') {
-    // Special case to account for two Staff entries for Margot.
-    people = await environment.getEntries({
-      content_type: 'person',
-      'fields.name[match]': 'Margot Harris',
-    });
-  } else {
-    // Find the Person entries of the same name.
-    people = await environment.getEntries({
-      content_type: 'person',
-      'fields.name[match]': staffName,
-    });
-  }
+  // Find the Person entries of the same name.
+  const people = await environment.getEntries({
+    content_type: 'person',
+    'fields.name[match]': staffName,
+  });
 
   // Hard return if there is no Person equivalent.
   if (!people.items.length) {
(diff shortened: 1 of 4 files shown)


contentful/management-api-scripts/2018_09_07_002_replace_staff_linked_entries_with_person.js
```diff
@@ -0,0 +1,139 @@
...
+const logger = createLogger('replace_staff_linked_entries_with_person');
+
+async function replaceStaffLinkedEntriesWithPerson(environment, staff) {
+  const staffName = getField(staff, 'name');
+
+  logger.info(`Processing Staff! [ID: ${staff.sys.id}] [Name: ${staffName}]\n`);
+
+  let people;
+
+  if (staffName === 'Margot') {
```
weerd: To avoid this could we just manually switch the duplicate Margot entries ref to use the correct one?

weerd: I went ahead and switched the 3 campaigns that referenced the duplicate Margot to reference the correct Margot staff entry. We can probs remove this conditional now :)

contentful/management-api-scripts/2018_09_07_002_replace_staff_linked_entries_with_person.js
```diff
@@ -0,0 +1,139 @@
...
+    });
+  }
+
+  // Hard return if there is no Person equivalent.
+  if (!people.items.length) {
+    return;
+  }
+
+  // Our Staff names are unique, so there shouldn't be more then one match.
+  const person = people.items[0];
```
weerd: Just FYI, I also just made the `name` field unique in the `Person` content type. Seems like that will help to avoid duplicate entries!

contentful/management-api-scripts/2018_09_07_002_replace_staff_linked_entries_with_person.js
mendelB: Awesome, thanks! I'll remove this conditional.

mendelB:
@weerd How do we feel about merging & running the second script? I tested again on `test` environment after removing the conditional, and it works 👌

PR #1096, opened 2018-09-13
Updating entities to reflect Staff entity removal.

weerd:
### What does this PR do?

This PR does some quick cleanup of the codebase to remove references to the now deprecated `Staff` entity and content type.

🗑 


app/Entities/Campaign.php
```diff
@@ -50,9 +50,9 @@ public function parseAffiliates($affiliates)
     public function parseCampaignLead($campaignlead, $additionalContent)
     {
         if ($campaignlead) {
-            // @TODO (2018-08-29): Staff should eventually be removed and replaced by the
-            // Person content type and class for all entries.
-            return $campaignlead->getContentType() === 'staff' ? new Staff($campaignlead->entry) : new Person($campaignlead->entry);
+            // @TODO (2018-09-13): We should make the CampaignLead field required and thus
+            // no longer need a conditional check here.
```
mendelB: Looks like there are about 10-15 campaigns without a campaign lead, so shouldn't be too big of a deal to update those and make this field required.

weerd: Yeah defs!

PR #1098, opened 2018-09-17
Un-bold Markdown Paragraph Links

mendelB:
### What does this PR do?

This PR ensures `<a>` tags in paragraphs have a 'normal' `font-weight`

I assumed the `markdown.scss` file was the correct place for this rule. I left in the old rule declaring plain `<a>` as bolded, since this card specified 'links in paragraph elements'.

Refs [Pivotal ID #159745386](https://www.pivotaltracker.com/story/show/159745386)

**BEFORE**
image

**AFTER**
image

### DIFF
--- a/resources/assets/components/utilities/Markdown/markdown.scss
+++ b/resources/assets/components/utilities/Markdown/markdown.scss
@@ -69,6 +69,10 @@
     font-weight: 700;
   }
 
+  p a {
+    font-weight: $weight-normal;
+  }
+
   h2,
   h3 {
     font-family: $primary-font-family;


weerd:
>  I left in the old rule declaring plain <a> as bolded, since this card specified 'links in paragraph elements'.

I don't think we want that.

I think it's safe to just do this, and remove the `a` from the declaration. 

```css
strong {
  font-weight: 700;
}
```

All `<a>`s in Markdown are going to show up in some `<p>` tag.

mendelB:
Ah ok I can get rid of that then. We'll still need the new explicit rule though to override the [forge <a> bolding](https://github.com/DoSomething/forge/blob/dev/scss/_base/_elements.scss#L58)

weerd:
> Ah ok I can get rid of that then. We'll still need the new explicit rule though to override the forge bolding

Cool, yeah, but no need to over-qualify it. It's already sitting inside the `.markdown` scss class... so `.markdown a` should be enough to override the Forge one :)

PR #1099, opened 2018-09-17
Update Markdown headers styling

mendelB:
### What does this PR do?

This PR updates the styling of our Markdown header tags.

I also removed some ruling from the Content Block styling that wasn't being used (since we use a `SectionHeader` component to take care of the title) but was affecting the Markdown headers.

**MOCKS**
md classes

**BEFORE**
image

**AFTER**
image

Refs [Pivotal ID #159745211](https://www.pivotaltracker.com/story/show/159745211)

### DIFF
--- a/resources/assets/components/utilities/Markdown/markdown.scss
+++ b/resources/assets/components/utilities/Markdown/markdown.scss
@@ -1,9 +1,6 @@
 @import '../../../scss/next-toolbox.scss';
 
 .markdown {
-  h1,
-  h2,
-  h3,
   p {
     color: inherit;
   }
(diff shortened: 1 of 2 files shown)


resources/assets/components/utilities/Markdown/markdown.scss
```diff
@@ -73,17 +70,32 @@
     font-weight: $weight-normal;
   }
 
+  h1 {
+    font-family: $secondary-font-family;
+    font-size: 41px;
+    font-weight: $weight-normal;
+    text-transform: uppercase;
+    color: #000000;
```
weerd: `$black`?

mendelB: Luke had the `h1` specifically as `#000000` as opposed to the `$blacks - #222` (the rest of the `h`s), but I honestly can't tell the difference. Should I change?

weerd: Oh interesting 🤔 @lkpttn mind weighing in? If we end up using `#000` our `$black` variable is a bit of a misnomer and for future Forge should be `$off-black`...?

PR #1103, opened 2018-09-20
Social Share Tray in Content Pages (and updates)

mendelB:
### What does this PR do?

This PR adds the `SocialShareTray` component to `GeneralPage`s!

- 🔘 [5af656a](https://github.com/DoSomething/phoenix-next/commit/5af656a84d855f91279f87cea9d4ee8e25100ed8) Adds `platforms` prop so that we can specify platforms for the tray (defaults to all of 'em)
- 🎋[1c09553](https://github.com/DoSomething/phoenix-next/commit/1c0955363410154a0b34ed7545f1b90ad7381b39) Adds `title` prop so we can customize the title (defaults to 'Share on social media').

The logic I used for rendering the buttons based on `platform` prop seems somewhat non-DRY but it was the best I could think of without going too crazy about this.

Refs [Pivotal ID #159745577](https://www.pivotaltracker.com/story/show/159745577)

image

image

### DIFF
--- a/resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
+++ b/resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
@@ -83,7 +83,7 @@ class SocialShareTray extends React.Component {
   };
 
   render() {
-    const { shareLink } = this.props;
+    const { shareLink, platforms } = this.props;
     const trackLink = this.props.trackLink || this.props.shareLink;
 
     return (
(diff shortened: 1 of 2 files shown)


resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
```diff
@@ -83,49 +83,59 @@ class SocialShareTray extends React.Component {
...
+
+          {platforms.includes('email') ? (
+            <ShareButton
+              className="email"
+              disabled={!shareLink}
+              icon={emailIcon}
+              text="Email"
+              onClick={() => this.handleEmailShareClick(shareLink, trackLink)}
+            />
+          ) : null}
```
weerd: Seems fine for now. Could be refactored into a loop that decides things based on the platform type, etc.

PR #1104, opened 2018-09-21
Caching and page updates

weerd:
### What does this PR do?

This PR fixes our caching system for Campaigns and updates it to add Pages into the mix as well. While working to add Authors to pages, there was a bit of a discrepancy with how Page and a Campaign were being returned which muddled some expected behavior when parsing blocks in the `Entity`. This update now unifies how we handle both Campaigns and Pages, ensuring both return a decoded JSON string as expected. It also simplifies and unifies the caching approach for both when finding a page or campaign by slug.

It also adds a new trait called `QueriesContentful`. While updating the `findBySlug()` method in both respective repository classes, it became obvious that both methods had a lot of similar code, so moving it into a trait made the most sense.

🌵 

### Checklist


app/Repositories/QueriesContentful.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    protected function getEntryFromSlugAsJson($type, $slug)
+    {
+        $query = (new Query)
+                ->setContentType($type)
+                ->where('fields.slug', $slug)
+                ->setInclude(3)
+                ->setLimit(1);
+
+        $entry = $this->contentful->getEntries($query);
```
weerd: Question: Is it odd that this trait has a call here to `$this->contentful`, but it's dependent on the contentful client being added to the class using the trait? Is this a bad pattern? Should I opt for passing in the client instead? 🤔

app/Repositories/PageRepository.php
```diff
@@ -33,18 +33,18 @@ public function __construct(Contentful $contentful)
...
+            });
+        }
 
-        if (! $pages->count()) {
+        if ($page === 'not_found') {
             throw new ModelNotFoundException;
         }
 
-        return new Page($pages[0]);
+        return json_decode($page);
```
weerd: I contemplated further DRYing the `PageRepository->findBySlug()` and `CampaignRepository->findBySlug()` since there's more overlap in code here... maybe moving all this into another method on the trait, but ended up deciding it might be best to keep this separate so behavior could be more easily modified between the two repository types... 🤔

app/Repositories/QueriesContentful.php
mendelB: Hmm, would it be helpful to add a `catch` or a conditional to ensure that the `contentful` client is set before invoking it? Although I def do see the argument in favor of manually passing the client to the method, though - that might get messy if we abstract other query methods to the Trait...

app/Repositories/QueriesContentful.php
```diff
@@ -0,0 +1,40 @@
...
+     */
+    protected function getEntryFromSlugAsJson($type, $slug)
+    {
+        $query = (new Query)
+                ->setContentType($type)
+                ->where('fields.slug', $slug)
+                ->setInclude(3)
+                ->setLimit(1);
+
+        $entry = app('contentful.delivery')->getEntries($query);
```
weerd: Found a middle-ground to not having to add another param to pass the contentful client, or importing the class and seems to work well.

mendelB: Ah way cool!

app/Http/Controllers/PageController.php
```diff
@@ -36,7 +36,7 @@ public function show($slug)
 
         return view('app', [
             'socialFields' => get_social_fields($page),
-            'pageTitle' => $page->title,
+            'pageTitle' => $page->fields->title,
```
mendelB: We might be able to refactor some of the logic in `get_social_fields` now that `$page` is also a `json_decode`d value. https://github.com/DoSomething/phoenix-next/blame/master/app/helpers.php#L230

(although I guess it can't really hurt to have it work both ways)

app/helpers.php
```diff
@@ -63,9 +63,9 @@ function scriptify($json = [], $store = 'STATE')
  * @param  \Closure  $callback
  * @return mixed
  */
-function remember($key, $minutes, Closure $callback, $skipCache = false)
+function remember($key, $minutes, Closure $callback)
```
mendelB: Was there a specific reason for removing the `$skipCache` variable here? Or do you just prefer the more explicit conditional pattern for caching/not caching?

mendelB:
Now that we're going to be caching Content Pages as well, should we add a ticket to get an [admin dashboard](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/AdminDashboard/AdminDashboard.js) for content pages as well so that admins can reset the cache as we do with Campaigns?

app/Http/Controllers/PageController.php
weerd: yeah, I refrained from it for now since it wasn't vital, but I figure we'll completely revisit/refactor this when we add the new Metadata content type!

app/helpers.php
weerd: Yeah, I felt it was simpler if we removed it. I felt it was confusing to even bother calling `remember()` if we don't want it to. Thus moving the logic to decide to skip cache back into the controller and simplifying the `remember()` helper function :)

PR #1106, opened 2018-09-24
Adds authors to static pages.

weerd:
### What does this PR do?

This PR adds Authors to static pages (ideally 2 authors max; will be adding validation to Contentful to limit to 2) and also adds some static page title sizing and responsive fixes.

- [Small Screen](https://user-images.githubusercontent.com/105849/45962398-c602e400-bfee-11e8-92e2-0035b8aa6596.png)
- [Medium Screen](https://user-images.githubusercontent.com/105849/45962369-b388aa80-bfee-11e8-934a-ecd1a8d9bd62.png)
- [Large Screen](https://user-images.githubusercontent.com/105849/45962348-a8ce1580-bfee-11e8-9581-4f743a4e98b0.png)

Refs [Pivotal ID #159745284](https://www.pivotaltracker.com/story/show/159745284)

### Checklist

### DIFF
--- a/resources/assets/components/pages/GeneralPage/general-page.scss
+++ b/resources/assets/components/pages/GeneralPage/general-page.scss
@@ -41,6 +41,25 @@
     }
   }
 
+  .general-page__authors {
+    display: inline-block;
+    margin: $base-spacing auto 0;
+    text-align: left;
+
+    @include media($large) {
+      margin-top: 36px;
+    }
+
+    .byline {
+      padding: $half-spacing 0;
+
+      @include media($medium) {
+        float: left;
+        padding: 0 $base-spacing;
+      }
+    }
+  }
+
   .default-container {
     width: 95%;
     margin-left: 2.5%;
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,10 +17,10 @@
       "dev": true,
       "requires": {
         "@babel/types": "7.0.0-beta.44",
-        "jsesc": "^2.5.1",
-        "lodash": "^4.2.0",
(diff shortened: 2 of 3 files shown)


resources/assets/components/pages/GeneralPage/general-page.scss
```diff
@@ -41,6 +41,25 @@
...
+  .general-page__authors {
+    display: inline-block;
+    margin: $base-spacing auto 0;
+    text-align: left;
+
+    @include media($large) {
+      margin-top: 36px;
+    }
+
+    .byline {
```
DFurnes: Nested selectors like this make me a little nervous (as opposed to explicitly applying a class to the thing we're styling, and ideally scoping it safely within the module it modifies), but not a blocker!

weerd: it's scoped within the `.general-page__authors` class which is a div containing only the single or multiple byline items. I think this should be cool for now until we change everything to use TailWind!

DFurnes: Yep, I saw it was scoped! My concern is that referencing `.byline` here in `general-page.scss` is not very discoverable, so if we make markup or style changes in `byline.scss` we're less likely to check that the overrides you've added here continue to function.

weerd:
Oh! Gotcha, yeah, good call.

resources/assets/components/pages/GeneralPage/general-page.scss
weerd: Oh! Gotcha, yeah, that makes sense.

PR #1110, opened 2018-09-27
Explore Campaigns Page Updates

mendelB:
### What does this PR do?

This PR updates the Explore Campaigns page on Phoenix.

- [dea641f](https://github.com/DoSomething/phoenix-next/commit/dea641f44a0013b505e8a8feee3418175f7ce30e) for the `getAll` campaigns method - orders campaigns by `sys.updatedAt` field
- [abe9d6b](https://github.com/DoSomething/phoenix-next/commit/abe9d6b019ff94ab82e7acfffdac83680d0a7406) adds a new `CampaignRepository` method called `getAllCampaignsSorted`(?) which sorts campaigns as follows:
  1. Open campaigns that are staff picks
  2. Open campaigns that are not staff picks
  3. Rest of campaigns
- [085b87c](https://github.com/DoSomething/phoenix-next/commit/085b87c66c9745e417455ba12035134a09371bee) Ungate the campaigns index list on the page
- [5c588e0](https://github.com/DoSomething/phoenix-next/commit/5c588e0aa48e010179a19dea36c3ed921535d886) Update the text / title content on the page to align with Ashes

**UPDATES**:
- [e47b46b](https://github.com/DoSomething/phoenix-next/pull/1110/commits/e47b46bbb8b7227da7bef584624ae3ca26a2398c) adds the index campaigns method to the `QueriesContentful` Trait as a more abstract method to be applied to more types (pages, etc.)
- [945ba92](https://github.com/DoSomething/phoenix-next/pull/1110/commits/945ba9207dc9b5a2985f22529e77121fbaecd49b) Use the new method in `CampaignsRepository#getAll`
- tidy up some of the other methods -- we were finally able to implement @DFurnes [words](https://github.com/DoSomething/phoenix-next/pull/983#discussion_r196186478) for some of the methods due to our Contentful SDK update, as well as some general `mapInto` abstractions.

Refs [Pivotal ID #160603200](https://www.pivotaltracker.com/story/show/160603200)

image

### DIFF
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -52,6 +52,11 @@ class CampaignRepository
         return json_decode($campaigns);
     }
 
+    /**
+     * Get all campaigns sorted by open status, and staff pick status.
+     *
+     * @return \Illuminate\Support\Collection
+     */
     public function getAllCampaignsSorted()
     {
         $campaigns = $this->getAll();
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -2,6 +2,7 @@
 
 namespace App\Repositories;
 
+use Carbon\Carbon;
 use App\Entities\Campaign;
 use Contentful\Delivery\Query;
 use App\Services\PhoenixLegacy;
@@ -61,6 +62,21 @@ class CampaignRepository
         });
 
         return json_decode($flattenedCampaign);
+    }
+
+    public function getAllCampaignsSorted()
+    {
+        $campaigns = $this->getAll();
+
+        // Partition into list of open and closed campaigns for sorting.
+        list($openCampaigns, $closedCampaigns) = collect($campaigns)->partition(function($campaign) {
+            return ((!$campaign->endDate) || $campaign->endDate > Carbon::now());
+        });
+
+        // Sort open campaigns by staff pick.
+        $sortedOpenCampaigns = $openCampaigns->sortByDesc('staffPick');
+
+        return $sortedOpenCampaigns->merge($closedCampaigns);
     }
 
     /**
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -45,11 +45,11 @@ class CampaignRepository
      */
     public function getAll()
     {
-        $flattenedCampaigns = remember('campaigns', 15, function () {
+        $campaigns = remember('campaigns', 15, function () {
             return $this->getEntriesAsJson('campaign');
         });
 
-        return json_decode($flattenedCampaigns);
+        return json_decode($campaigns);
     }
 
     public function getAllCampaignsSorted()
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -152,11 +152,9 @@ class CampaignRepository
             ->where('fields.legacyCampaignId', $ids, 'in')
             ->setInclude(1);
 
-        $results = $this->contentful->getEntries($query);
+        $results = $this->contentful->getEntries($query)->getItems();
 
-        $contentfulCampaigns = collect($results->getIterator())->map(function ($campaign) {
-            return new TruncatedCampaign($campaign);
-        });
+        $contentfulCampaigns = collect($results)->mapInto(TruncatedCampaign::class);
 
         // List of IDs returned from Contentful.
         $foundIds = $contentfulCampaigns->pluck('legacyCampaignId')->all();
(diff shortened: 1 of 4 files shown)


app/Repositories/CampaignRepository.php
```diff
@@ -45,22 +46,39 @@ public function __construct(Contentful $contentful, PhoenixLegacy $phoenixLegacy
...
+
             $campaigns = $this->contentful->getEntries($query);
             $array = iterator_to_array($campaigns);
 
             // Transform & cast as JSON so we can cache this. One little gotcha -
             // we don't want full campaigns, that'd be a monstrous object!
-            $results = collect($array)->map(function ($entity) {
-                return new TruncatedCampaign($entity);
-            });
+            $results = collect($array)->mapInto(TruncatedCampaign::class);
```
mendelB: I was digging into collection methods for sorting, and bumped into this and thought it'd be cool to throw in here! https://laravel.com/docs/5.7/collections#method-mapinto

app/Repositories/CampaignRepository.php
```diff
@@ -45,22 +46,39 @@ public function __construct(Contentful $contentful, PhoenixLegacy $phoenixLegacy
...
 
         return json_decode($flattenedCampaign);
     }
 
+    public function getAllCampaignsSorted()
+    {
+        $campaigns = $this->getAll();
+
+        // Partition into list of open and closed campaigns for sorting.
+        list($openCampaigns, $closedCampaigns) = collect($campaigns)->partition(function ($campaign) {
```
mendelB: Bless your heart Taylor https://laravel.com/docs/5.7/collections#method-partition

app/Repositories/CampaignRepository.php
```diff
@@ -45,22 +46,39 @@ public function __construct(Contentful $contentful, PhoenixLegacy $phoenixLegacy
     public function getAll()
     {
         $flattenedCampaign = remember('campaigns', 15, function () {
-            $query = (new Query)->setContentType('campaign')->setInclude(0);
+            $query = (new Query)
+                ->setContentType('campaign')
+                ->setInclude(0)
+                ->orderBy('sys.updatedAt', true);
```
weerd: Not necessary for merge, but maybe we toss this into the `QueriesContentful` trait as well... could be something we use to get all pages in the future? 🤔

mendelB: Oh word, I can try moving this over there.

mendelB:
UPDATES (added to PR dscription as well):

- e47b46b adds the index campaigns method to the QueriesContentful Trait as a more abstract method to be applied to more types (pages, etc.)
- 945ba92 Use the new method in CampaignsRepository#getAll
- tidy up some of the other methods -- we were finally able to implement @DFurnes words for some of the methods due to our Contentful SDK update, as well as some general mapInto abstractions.

mendelB:
Having spent some more time with this, I encountered two points we might want to consider?
1. Contentful limits index queries to 100 entries by default. Right now this is not a problem, but as we import more campagns / add more pages, we should probably think about pagination (see that doc for the `skip` param). Or for short term we could also just add a `limit` param which has a max of `1000` which can keep us covered for a long while if need be.

2. We're caching the campaigns index query for 15 mins *always*. Do we want to:
    1. Add some toggle logic based on the `CONTENTFUL_CACHE` variable?
    2. And/Or think about adding a Cache button in a admin dashboard to the explore campaigns page? (In case e.g. someone needs to have the ordering switched or cover image updated in a pressing manner for a partner)

app/Repositories/CampaignRepository.php
```diff
@@ -44,21 +45,26 @@ public function __construct(Contentful $contentful, PhoenixLegacy $phoenixLegacy
...
-            $query = (new Query)->setContentType('campaign')->setInclude(0);
-            $campaigns = $this->contentful->getEntries($query);
-            $array = iterator_to_array($campaigns);
-
-            // Transform & cast as JSON so we can cache this. One little gotcha -
-            // we don't want full campaigns, that'd be a monstrous object!
-            $results = collect($array)->map(function ($entity) {
-                return new TruncatedCampaign($entity);
-            });
+        $flattenedCampaigns = remember('campaigns', 15, function () {
```
weerd: Can we just call this variable `$campaigns`? The whole "flattened" stuff makes it confusing and we already know it's coming in as json thanks to `getEntriesAsJson()` :)

app/Repositories/CampaignRepository.php
weerd: So awesome!

app/Repositories/CampaignRepository.php
```diff
@@ -44,21 +45,26 @@ public function __construct(Contentful $contentful, PhoenixLegacy $phoenixLegacy
...
-                return new TruncatedCampaign($entity);
-            });
+        $campaigns = remember('campaigns', 15, function () {
+            return $this->getEntriesAsJson('campaign');
+        });
+
+        return json_decode($campaigns);
+    }
+
+    public function getAllCampaignsSorted()
```
weerd: DocBlocks!

PR #1115, opened 2018-09-28
Using .length cause this is not PHP.

weerd:
### What does this PR do?

This PR switches to use `authors.length` for the ternary check if any authors have been added. @mendelB realized that just checking against `authors` is truthy and leads to an empty div with spacing. 

I was stuck in PHP world, where empty arrays are falsey 🤦‍♂️ 


weerd:
👍 😂

PR #1119, opened 2018-09-30
Page caching and Admin Dashboard

mendelB:
### What does this PR do?

This PR adds a Page Admin Dashboard to general content pages to enable staff to clear the page from the cache.

- New `PageDashboard` component with a link to the backend cache clearing route
- Added a `page_` prefix to the cache key for pages

I was working on the toggle for the Social Share tray in general pages, and realized that there wasn't ye a way to clear the Page cache so address this card we had in the icebox real quick.
@weerd had mentioned us potentially rethinking the Admin Dashboard in general, but figured this could work for now.

Refs [Pivotal ID #160696564](https://www.pivotaltracker.com/story/show/160696564)

### DIFF
--- a/app/Repositories/PageRepository.php
+++ b/app/Repositories/PageRepository.php
@@ -20,7 +20,7 @@ class PageRepository
         if (! config('services.contentful.cache')) {
             $page = $this->getEntryFromSlugAsJson('page', $slug);
         } else {
-            $page = remember($slug, 15, function () use ($slug) {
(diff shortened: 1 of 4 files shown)


weerd:
I think one reason I might have not thought to re-include the `campaign_` prefix on the cache keys is cause the keys are made with the slug which technically should already be unique to start with. But I can see this as being useful in case we ever want to clear all `campaign_*` or `page_*` caches!

PR #1128, opened 2018-10-11
TSA user gallery fix

weerd:
### What does this PR do?

This PR fixes an issue with how the TSA user gallery was displaying. Since the parent container was not expanding the full width the gallery was ending up compact. Did a couple fixes to resolve this while still keeping the TSA from being full width on large screen size, and also unifying the markup between it and the PSA.

**Before**:
screen shot 2018-10-11 at 12 54 00 pm

**After**:
screen shot 2018-10-11 at 12 50 48 pm

🌵 

### Checklist

### DIFF
--- a/resources/assets/components/actions/TextSubmissionAction/text-submission-action.scss
+++ b/resources/assets/components/actions/TextSubmissionAction/text-submission-action.scss
@@ -1,4 +1,11 @@
+@import '../../../scss/next-toolbox.scss';
+
 .text-submission-action {
+  @include media($large) {
+    padding-right: $half-spacing;
+    width: 66.6666666666666%;
+  }
+
   .field-label {
     height: auto;
   }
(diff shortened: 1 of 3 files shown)


resources/assets/components/actions/TextSubmissionAction/text-submission-action.scss
```diff
@@ -1,4 +1,11 @@
+@import '../../../scss/next-toolbox.scss';
+
 .text-submission-action {
+  @include media($large) {
+    padding-right: $half-spacing;
+    width: 66.6666666666666%;
```
DFurnes: How about using `calc(100%  / 3);` here?

weerd: Meh. I refrained since this is a common percentage we use in a few places (probs could us a variable, but holding off until the TailWind switcheroo), so calc doesn't really help _that_ much and just adds extra processing on the CSS side. Not a huge deal, and could be convinced to reconsider 😉

PR #1131, opened 2018-10-12
Gallery Block rendering - first pass

mendelB:
### What does this PR do?

This PR adds support and rendering for the new Gallery Block in Phoenix.

- adds new `GalleryBlock` Entity
- adds new case for `galleryBlock` type to `Entity#parseBlock`
- adds new case for `campaign` type to `Entity#parseBlock` (we now have Campaigns as reference blocks for the first time 🎊) using the `TruncatedCampaign` since we only need some top level fields.
- adds new `GalleryBlock` React component,and a case for rendering in the `ContentfulEntry`.

As per our Pivotal discussion, this is kind of a rough first pass at just getting this component up and running. 

I tested this with support for `Person` and `Campaign` references. I couldn't think of any other Content Types we wanted/needed to do this with as of yet.

Using `Figure` to help with Image/text layout was an absolute lifesaver.

I think the hairiest part of this is the `renderBlock` method in the `galleryBlock` component. Which addresses -I think- what is/going-to-be the most complex about setting this all up properly.

I went with -again, per the discussion- using a `switch` to set up specified `imageUrl` and `text` fields, pulling from the correct Entry fields based on the Content-Type. (for now, support for just Campaigns and Persons).

Question (beyond whether I did any of this right):
I had a thought while setting up the Entity, that that might be an interesting place to take care of formalizing the fields for the GalleryBlock nodes?

Would it be cool if the Entity had it's own `parseBlocks` method that based on the Content-Type, returned the `imageUrl` and `text` from the correct fields instead of the component taking care of that?

### Checklist

**CAMPAIGNS no title** 
image

image

image

**Persons with title**
image

**On a campaign page**
image

### DIFF
--- a/resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
+++ b/resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
@@ -37,7 +37,7 @@ const renderBlock = block => {
 
 const GalleryBlock = ({ title, blocks }) => (
   <div className="gallery-block">
-    <h1 className="text-centered">{title}</h1>
+    {title ? <h1 className="text-centered">{title}</h1> : null}
 
     <Gallery type="triad">{blocks.map(renderBlock)}</Gallery>
   </div>
--- a/resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
+++ b/resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
@@ -1,0 +1,56 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Figure } from '../../Figure';
+import Gallery from '../../utilities/Gallery/Gallery';
+import { contentfulImageUrl } from '../../../helpers';
+
+const renderBlock = block => {
+  let imageUrl;
+  let text;
+
+  switch (block.type) {
+    case 'person':
+      imageUrl = block.fields.photo;
+      text = block.fields.name;
+      break;
+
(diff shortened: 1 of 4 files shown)


weerd:
> we now have Campaigns as reference blocks for the first time 🎊

woot woot!

resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
```diff
@@ -0,0 +1,55 @@
...
+  let text;
+
+  switch (block.type) {
+    case 'person':
+      imageUrl = block.fields.photo;
+      text = block.fields.name;
+      break;
+
+    case 'campaign':
+      imageUrl = block.coverImage.url;
```
weerd: These cases work for now for this simplified version, but I could see us return JSX here, where each `type` dictates it's own markup for what to output... (similar to how the ContentfulEntry ends up doing). So in that case you wouldn't really need to unify the `imageUrl` or `text`, etc.

mendelB: Hmm, totally. Which would also make my question

>  had a thought while setting up the Entity, that that might be an interesting place to take care of formalizing the fields for the GalleryBlock nodes?

a moot point.

weerd: ya. I figured we'd answer that as we proceeded with development 💃

PR #1135, opened 2018-10-16
Gallery block updates - Person component.

mendelB:
### What does this PR do?

This PR updates the `GalleryBlock` to render type specific components.

- adds a `Person` component for rendering Person blocks 
- adds a `StaffTemplate` to the Person templates for `Person`s of type `staff`.
- adds a `BoardMemberTemplate` to the Person templates for `Person`s of type `board member`
- adds a few missing fields to the Person entity

Based on some discussion with @weerd this attempts to formalize the various `Person` rendered gallery items into their own templates. Beginning with Staff - having its own `StaffTemplate` for its specific rendering.

We're attempting to replicate the Our Team page on ashes:
image
 

With each Staff node rendering the `alternatePhoto`, `name` as a caption, `jobTitle` in italics, and link to twitter handle if set in `twitterId`. Giving us:

image

And for the Our Board page on ashes:
image

With each board node rendering the `alternatePhoto`, `name` as a caption, and `description`. Giving us:
image

The naming for these components is very poor: `StaffTemplate`, and `BoardMemberTemplate`. Trying to think of better naming for these that isn't so specific.

A thought: Would it be smart, or overkill to abstract away a component called `ImageWithCaption` or something which takes care of the repetition across the Staff and Board templates, but allows children as a prop, so we can add the appropriate specific fields across the two?

### DIFF
--- a/app/Entities/Person.php
+++ b/app/Entities/Person.php
@@ -18,13 +18,15 @@ class Person extends Entity implements JsonSerializable
             'type' => $this->getContentType(),
             'fields' => [
                 'name' => $this->name,
-                'type' => $this->type,
+                // $this->type returns 'Entry'.
+                'type' => $this->entry->type,
                 'active' => $this->active,
                 'jobTitle' => $this->jobTitle,
                 'email' => $this->email,
                 'photo' => get_image_url($this->photo),
                 'alternatePhoto' => get_image_url($this->alternatePhoto),
                 'description' => $this->description,
+                'twitterId' => $this->twitterId,
             ],
         ];
     }
--- a/app/Entities/Person.php
+++ b/app/Entities/Person.php
@@ -18,8 +18,7 @@ class Person extends Entity implements JsonSerializable
             'type' => $this->getContentType(),
             'fields' => [
                 'name' => $this->name,
-                // $this->type returns 'Entry'.
-                'type' => $this->entry->type,
+                'type' => $this->entry->type, // $this->type would result in 'Entry' (the `sys` type) being returned (https://git.io/fx2NH).
                 'active' => $this->active,
                 'jobTitle' => $this->jobTitle,
                 'email' => $this->email,
--- a/resources/assets/components/Person/Person.js
+++ b/resources/assets/components/Person/Person.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import StaffTemplate from './templates/StaffTemplate/StaffTemplate';
-import BoardTemplate from './templates/BoardTemplate/BoardTemplate';
+import BoardMemberTemplate from './templates/BoardMemberTemplate/BoardMemberTemplate';
 
 const Person = ({ json }) => {
   switch (json.fields.type) {
@@ -10,7 +10,7 @@ const Person = ({ json }) => {
       return <StaffTemplate {...json.fields} />;
 
     case 'board member':
-      return <BoardTemplate {...json.fields} />;
+      return <BoardMemberTemplate {...json.fields} />;
 
     default:
       return null;
--- a/resources/assets/components/Person/Person.js
+++ b/resources/assets/components/Person/Person.js
@@ -1,9 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import StaffTemplate from './templates/StaffTemplate/StaffTemplate';
 
 const Person = ({ json }) => {
   switch (json.fields.type) {
+    case 'staff':
+      return <StaffTemplate {...json.fields} />;
     default:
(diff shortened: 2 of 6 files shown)


mendelB:
**Update**:
Added `BoardTemplate` to deal with the Our Board page.

weerd:
> A thought: Would it be smart, or overkill to abstract away a component called ImageWithCaption or something which takes care of the repetition across the Staff and Board templates, but allows children as a prop, so we can add the appropriate specific fields across the two?

Possibly but stuff like this has bitten us in the past; for example eventually needing some extra markup in the image with caption to help a layout with a specific set children items... Not sure. I'd say leave it as is for now 🤔

app/Entities/Person.php
```diff
@@ -18,13 +18,15 @@ public function jsonSerialize()
             'type' => $this->getContentType(),
             'fields' => [
                 'name' => $this->name,
-                'type' => $this->type,
+                // $this->type returns 'Entry'.
```
weerd: Why is that though? Isn't the `type` field a radio select with an option like `staff` or `board` etc? I'm confused why it's `entry`...

mendelB:
> Possibly but stuff like this has bitten us in the past; for example eventually needing some extra markup in the image with caption to help a layout with a specific set children items... Not sure. I'd say leave it as is for now 🤔

Yeah, that does make sense. I was poking around some other galleries on Ashes, and figured the most classic one seems to be the Board style: photo, caption, description

Maybe rename `BoardTemplate` to `ClassicTemplate` and alter the prop naming a bit? That'd make it far less specific.

app/Entities/Person.php
mendelB: IKR? It seems to be a property on the updated `Entry` class, where `->type` will return `$this->sys->type` - I assume as a convenience method 🤷‍♂️

weerd: oh bummer. ok, whelp, good to know! Mind just maybe rephrasing the comment a bit, or starting it with a word or two. It initially just looked like commented out code and not an actual comment 😬

mendelB: @weerd I decided to actually 🕵️ a bit, and here's what I found:

First of all, this `type` business only affects Content Types with their own `type` field - which I found strange, since `$this->type` is returning `Entry` and not the proper field value anyways so why should it be contingent on a native field on the Content Type...

Turns out here's the flow:

Let's say `Page` calls `$this->entity`:
- We hit `Entity#__get` https://github.com/DoSomething/phoenix-next/blob/master/app/Entities/Entity.php#L120
- Run `if (! $this->offsetExists($property)) {
            return null;
        }` https://github.com/DoSomething/phoenix-next/blob/master/app/Entities/Entity.php#L122
  - since the entity itself has no `type` field, we indeed return `null`.

However if `Person` calls `$this->entity`:
- We hit `Entity#__get` https://github.com/DoSomething/phoenix-next/blob/master/app/Entities/Entity.php#L120
- `if (! $this->offsetExists($property)) {
            return null;
        }` https://github.com/DoSomething/phoenix-next/blob/master/app/Entities/Entity.php#L122 does not run since we do have a `type` field
- we call `$this->getType()` (`$value = $this->entry->{'get'.ucwords($property)}();`) https://github.com/DoSomething/phoenix-next/blob/master/app/Entities/Entity.php#L128
which runs on the `Entry` which (for our installed version of Contentful SDK) extends the `LocalizedResource` class which extends the `BaseResource` class which actually _has_ a declared method called `getType`! Which returns... `$this->sys->type`!

(Or in the case of the updated SDK - a slightly updated format: https://github.com/contentful/contentful.php/blob/master/src/Resource/BaseResource.php#L5)

However if we call `$this->entry->type`, then it goes directly to the Entry's `__get` method, which invokes it's own `get($prop)` method (), which faithfully retrieves the `type` field

weerd: Oh wow. Nice 🕵️ing!!

I wonder if we should just never name any fields as `type` 🤔 Maybe that was a bad call? Or this is all fine... Does make it seem a bit confusing.

mendelB: Another option could be to update the `$value = $this->entry->{'get'.ucwords($property)}` to simply call `this->entry->{$property}` since the SDK now supports virtual properties
https://github.com/contentful/contentful.php/blob/master/CHANGELOG.md#added-4

mendelB: This way, we wouldn't ever be calling `getType()` and accidentally invoking that method.

weerd: oh true. Yeah, mind adding an icebox ticket in Pivotal Tracker and referencing this convo there so we don't forget. It'll be a change we'll want to test thoroughly!

PR #1137, opened 2018-10-17
Updates to add support for page metadata.

weerd:
### What does this PR do?

This PR adds support on the Phoenix side for campaign and page `metadata`, which will help with SEO and also help simplify how we add social overrides to aforementioned content. Our current `socialOverrides` approach is a bit convoluted and not used much and/or not used well.

This new `metadata` support will ensure that a basic level of SEO is added to campaigns and pages, but will also allow for overriding certain items when necessary via an upcoming new `metadata` content type that can be referenced on a campaign or page.

This only adds support on the Phoenix side... I have an upcoming PR which will provide a migration to add the `metadata` content type to Contentful, as well as updating the `Metadata` entity to pull from Contentful. The items in the entity are currently just placeholder but end up providing suitable defaults for the meta tags on a page. That being said, lets hold off to merge this util the content type migration is also available to run 💃 

Refs [Pivotal ID #159468120](https://www.pivotaltracker.com/story/show/159468120)

### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -271,7 +271,7 @@ function get_metadata($entry)
 
     $image = $entry->metadata->fields->image;
 
-    if (!$image && data_get($entry, 'coverImage')) {
+    if (! $image && data_get($entry, 'coverImage')) {
         $image = $entry->coverImage->url.'?w=1200&h=1200&fm=jpg&fit=fill';
     }
 
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -224,6 +224,7 @@ function useOverrideIfSet($field, $base, $override)
  *
  * @param  \Contentful\Delivery\Resource\Entry|stdClass $entry
  * @return array|null
+ * @deprecated Will be removed; use get_metadata() instead.
  */
 function get_social_fields($entry)
 {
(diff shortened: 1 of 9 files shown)


app/helpers.php
```diff
@@ -256,6 +257,36 @@ function get_social_fields($entry)
...
 
+/**
+ * Get metadata associated with a page or campaign.
+ *
+ * @param  \Contentful\Delivery\Resource\Entry|stdClass $entry
+ * @return array
+ */
+function get_metadata($entry)
+{
+    if ($entry->type !== 'campaign') {
```
mendelB: Why is this different for non-campaigns?

weerd: Oh, because the structure is different and wanted to unify them. We've setup a `campaign` to be  a more initially flat structure while `page`s have the typical `['id', 'type', 'fields']` with the fields item containing all the fields for a page including the `metadata`.

Thus to make it easier when trying to grab data in this helper function, I try to unify what the structure of the `$entry` looks like ;)

weerd: It could be debated that we turn a campaign into a structure that matches all the other entities, but that's a bit of a refactor for a some future task if deemed worthy.

mendelB: Ahh totally! Yeah, I'd vote that it's worth thinking about down the pike. But def doesn't seem like too big of a deal in general.

PR #1145, opened 2018-10-18
Gallery Block - Items Per Row

mendelB:
### What does this PR do?

This PR adds support for setting the number of items per row for Gallery Blocks.

- Moves gallery block migration script to `contentful/content-types` in a new `gallery-block.js` file (you're welcome @weerd 😁)
- Adds `itemsPerRow` field and validations 
- Adds the field to the `GalleryBlock` Entity
- Adjusts the `GalleryBlock` component to use this prop to map to the `type` for the `Gallery` (`2 => 'duo', 3 => 'triad', 4 => 'quartet'`)
- Adds 'duo' as a supported `type` prop to `Gallery`. (The functionality is already supported via [Forge](https://github.com/DoSomething/forge/blob/dev/scss/_modules/_gallery.scss#L172)

~One interesting thing - while adding the new GalleryBlock `itemsPerRow` field to the migration script, I tried running the migration to ensure it had the correct syntax and got the following error:
`Error: "max" validation expected to be "number", but got "null"`
(This seems to have to do with it being of type `Integer` since `"max"` is one of the potential validations)~
~I confirmed that the syntax is indeed correct by adding and configuring the `itemsPerRow` field manually on Contentful and running a migration generation with the `contentful-cli` but no dice -- same error.~
~So this seems to be a Contentful issue and I did indeed find a similar issue referenced [here](https://github.com/contentful/contentful-migration/issues/106) from a while back, although that's closed out now. So while this field is valid, and works (I tested across the whole flow and everything is 👌) as of now at least - this is not a runnable migration. I can throw a question on the community forum to see what's up if you think this is worth investigating.~
I 👏 just 👏 needed 👏 to 👏 update 👏 my 👏 `contentful-cli` 👏 
The migration works wonders. That issue was closed bc it *was* indeed fixed.

Another separate thing --  I wasn't completely sure about how to gracefully map the numeric values set for `itemsPerRow` to the proper Gallery `types`. So I hope my approach is up to par! (Storing the types in an array and subtracting from the values and mapping them as indices)

### DIFF
--- a/contentful/content-types/galleryBlock.js
+++ b/contentful/content-types/galleryBlock.js
@@ -1,6 +1,6 @@
 module.exports = function(migration) {
   const galleryBlock = migration
-    .createContentType('galleryBlock')
+    .editContentType('galleryBlock')
     .name('Gallery Block')
     .description('Displays a gallery of referenced entries')
     .displayField('internalTitle');
--- a/contentful/content-types/galleryBlock.js
+++ b/contentful/content-types/galleryBlock.js
@@ -41,4 +41,21 @@ module.exports = function(migration) {
     })
     .required(true)
     .localized(false);
+
+  galleryBlock
+    .createField('itemsPerRow')
+    .name('Items Per Row')
+    .type('Integer')
+    .validations([
+      {
+        in: [2, 3, 4],
+      },
+    ])
+    .required(true)
+    .localized(false);
+
+  galleryBlock.changeEditorInterface('galleryBlock', 'radio', {
+    helpText:
+      'The maximum number of items in a single row when viewing the gallery in a large display. (small screens are limited to one per row).',
+  });
 };
--- a/contentful/content-types/galleryBlock.js
+++ b/contentful/content-types/galleryBlock.js
@@ -56,6 +56,6 @@ module.exports = function(migration) {
 
   galleryBlock.changeEditorInterface('itemsPerRow', 'radio', {
     helpText:
-      'The maximum number of items in a single row when viewing the gallery in a large display. (small screens are limited to one per row).',
+      'The maximum number of items in a single row when viewing the gallery in a large display.',
   });
 };
--- a/contentful/content-types/galleryBlock.js
+++ b/contentful/content-types/galleryBlock.js
@@ -1,6 +1,6 @@
 module.exports = function(migration) {
   const galleryBlock = migration
-    .editContentType('galleryBlock')
+    .createContentType('galleryBlock')
     .name('Gallery Block')
     .description('Displays a gallery of referenced entries')
     .displayField('internalTitle');
--- a/contentful/content-types/galleryBlock.js
+++ b/contentful/content-types/galleryBlock.js
@@ -54,7 +54,7 @@ module.exports = function(migration) {
     .required(true)
     .localized(false);
 
-  galleryBlock.changeEditorInterface('galleryBlock', 'radio', {
+  galleryBlock.changeEditorInterface('itemsPerRow', 'radio', {
     helpText:
       'The maximum number of items in a single row when viewing the gallery in a large display. (small screens are limited to one per row).',
   });
--- a/resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
+++ b/resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
@@ -20,6 +20,7 @@ const renderBlock = block => {
 const galleryTypes = ['duo', 'triad', 'quartet'];
 
 const GalleryBlock = ({ title, blocks, itemsPerRow }) => {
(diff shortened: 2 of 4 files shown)


contentful/content-types/galleryBlock.js
```diff
@@ -41,4 +41,21 @@ module.exports = function(migration) {
...
+  galleryBlock
+    .createField('itemsPerRow')
+    .name('Items Per Row')
+    .type('Integer')
+    .validations([
+      {
+        in: [2, 3, 4],
+      },
+    ])
+    .required(true)
```
mendelB: I made this a required field. Do we think it might be better as optional with some sensible default?

As of now, in entry drafts etc. if the field - and thus the `type` prop is left blank, the gallery defaults to 1 per row (with the usual component warning about a missing required prop). Which seems fine, but just wanted to raise the question.

contentful/content-types/galleryBlock.js
```diff
@@ -41,4 +41,21 @@ module.exports = function(migration) {
...
+    .localized(false);
+
+  galleryBlock.changeEditorInterface('galleryBlock', 'radio', {
+    helpText:
+      'The maximum number of items in a single row when viewing the gallery in a large display. (small screens are limited to one per row).',
```
mendelB: Should I get rid of the help text in this migration? I know we don't generally add them, but I can think of arguments as for why they'd be useful to have in our migration arsenal. 
(But also -- clutter  🤷‍♂️ )

weerd:
> Moves gallery block migration script to contentful/content-types in a new gallery-block.js file (you're welcome @weerd 😁)

Ahhhh 😌

contentful/content-types/galleryBlock.js
weerd: I think it's fine to leave it. I wasn't adding them before because I didn't know they are now supported via migrations! Huzzah! If we know the help text we want and can add so we don't need to copy paste across env, than do it! 💃

contentful/content-types/galleryBlock.js
weerd: 🤔 I'm potentially _for_ making it required, so there are no surprises on how it has decided to layout and thus needs to be explicitly set.

@DFurnes thoughts?

resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
```diff
@@ -17,17 +17,25 @@ const renderBlock = block => {
...
-  <div className="gallery-block">
-    {title ? <h1>{title}</h1> : null}
+const galleryTypes = ['duo', 'triad', 'quartet'];
 
-    <Gallery type="triad">{blocks.map(renderBlock)}</Gallery>
-  </div>
-);
+const GalleryBlock = ({ title, blocks, itemsPerRow }) => {
+  // Subtract 2 from itemsPerRow to map properly to the galleryTypes indices (see propTypes).
+  const galleryType = galleryTypes[itemsPerRow - 2];
```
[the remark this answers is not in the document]
weerd: Might not work if integers for the keys... but could also work fine if integers are strings and we just convert the selected integer to a string before grabbing from the `galleryTypes`.

contentful/content-types/galleryBlock.js
mendelB: That seems sensible 👌

contentful/content-types/galleryBlock.js
mendelB: Good call -- it's a bit extra. I'll 🗑 it

resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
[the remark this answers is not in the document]
mendelB: That feels like a better approach! I'll give this a go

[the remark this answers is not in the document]
mendelB: Works wonders! Also cool - the keys get typecast [automatically](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors#Property_names) so we don't need to do our own conversion and deal with `null`s etc. 😌

contentful/content-types/galleryBlock.js
mendelB: It just occurred to me -- should we add a `1` option which would just mean no `type` to allow adding Galleries with single columns? I reflexively thought we don't want those - but that _is_ the `Gallery` default. Unless we indeed specifically to not want single column galleries?

weerd: Lets not add it for now. I think we don't want to offer that as an option since we have other blocks that are better suited for that!

DFurnes: Yeah, I think those are smart calls! ➕

PR #1146, opened 2018-10-19
Fixing up the social metadata and associated meta tags.

weerd:
### What does this PR do?

This PR does a bit of tweaking and updates a few items in the `metadata` and corresponding outputted meta tags for pages and campaigns. We added a landscape version of the logo that can be linked to via Forge, and will work better for Facebook social shares.

This PR also adds the `og:image:width` and `og:image:height` meta tags, since according to Facebook:
> Using these tags will specify the image dimensions to the crawler so that it can render the image immediately without having to asynchronously download and process it.

🌵 

Refs [Pivotal ID #159468120](https://www.pivotaltracker.com/story/show/159468120)

### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -275,16 +275,18 @@ function get_metadata($entry)
         $image = $entry->coverImage->url;
     }
 
-    $data = [
+    return [
         'title' => data_get($entry, 'metadata.fields.title', $entry->title),
         'type' => 'article',
         'description' => data_get($entry, 'metadata.fields.description', null),
         'url' => config('services.phoenix.url').'/us/'.$entry->slug,
         'facebook_app_id' => config('services.analytics.facebook_id'),
-        'image' => $image ? $image.'?w=1200&h=1200&fm=jpg&fit=fill' : 'https://forge.dosomething.org/resources/ds-logo-highres.png',
+        'image' => [
+            'url' => $image ? $image.'?w=1200&h=1200&fm=jpg&fit=fill' : 'https://forge.dosomething.org/resources/ds-logo-landscape.png',
+            'width' => '1200',
+            'height' => $image ? '1200' : '630',
+        ],
     ];
-
-    return $data;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


app/helpers.php
```diff
@@ -275,16 +275,18 @@ function get_metadata($entry)
...
         'title' => data_get($entry, 'metadata.fields.title', $entry->title),
         'type' => 'article',
         'description' => data_get($entry, 'metadata.fields.description', null),
         'url' => config('services.phoenix.url').'/us/'.$entry->slug,
         'facebook_app_id' => config('services.analytics.facebook_id'),
-        'image' => $image ? $image.'?w=1200&h=1200&fm=jpg&fit=fill' : 'https://forge.dosomething.org/resources/ds-logo-highres.png',
+        'image' => [
+            'url' => $image ? $image.'?w=1200&h=1200&fm=jpg&fit=fill' : 'https://forge.dosomething.org/resources/ds-logo-landscape.png',
+            'width' => '1200',
+            'height' => $image ? '1200' : '630',
```
weerd: I can add a comment if you think it's useful, but the height will only differ if it resorts to using the `forge.dosomething.org` provided logo, otherwise it'll be `1200`.

resources/views/partials/metadata.blade.php
```diff
@@ -20,5 +20,7 @@
 @endif
 <meta property="og:url" content="{{ $metadata['url'] }}" />
 <meta property="og:site_name" content="DoSomething.org" />
-<meta property="og:image" content="{{ $metadata['image'] }}" />
-<meta property="fb:admins" content="{{ $metadata['facebook_app_id'] }}" />
+<meta property="og:image" content="{{ $metadata['image']['url'] }}" />
+<meta property="og:image:width" content="{{ $metadata['image']['width'] }}" />
+<meta property="og:image:height" content="{{ $metadata['image']['height'] }}" />
+<meta property="fb:app_id" content="{{ $metadata['facebook_app_id'] }}" />
```
weerd: Turns out `fb:admins` was an old meta tag property and no longer used in favor of `fb:app_id`.

app/helpers.php
mendelB: I _think_ it's pretty deducible 🤷‍♂️

DFurnes: It might be worth adding a comment explaining why the 1200x1200 default crop, but this looks straightforward enough to me otherwise!

weerd: Yeah, ok... I could see us forgetting 🤔 Will add and then merge 💃

PR #1150, opened 2018-10-22
Page Cover Image Field Contentful Migration

mendelB:
### What does this PR do?

This PR adds a Contentful Migration script which adds a `coverImage` field to the Page Content Type.

image

### DIFF
--- a/contentful/migrations/2018_10_22_001_add_cover_image_field_to_page.js
+++ b/contentful/migrations/2018_10_22_001_add_cover_image_field_to_page.js
@@ -13,6 +13,6 @@ module.exports = function(migration) {
   page.moveField('coverImage').beforeField('content');
 
   page.changeEditorInterface('coverImage', 'assetLinkEditor', {
-    helpText: 'The cover Image will display on the page before the content',
+    helpText: 'The cover image will display on the page before the content.',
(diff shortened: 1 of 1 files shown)


contentful/migrations/2018_10_22_001_add_cover_image_field_to_page.js
```diff
@@ -0,0 +1,18 @@
...
+    .type('Link')
+    .linkType('Asset')
+    .validations([{ linkMimetypeGroup: ['image'] }])
+    .required(false)
+    .localized(false);
+
+  page.moveField('coverImage').beforeField('content');
+
+  page.changeEditorInterface('coverImage', 'assetLinkEditor', {
+    helpText: 'The cover Image will display on the page before the content',
```
weerd: random uppercase in `Image`

weerd: Also add a period at end of the sentence in the help text; would like to try and be consistent with how we add that info 😬

mendelB: Ah, good catches - will fix!

PR #1151, opened 2018-10-22
Add support for Page Cover Image field

mendelB:
### What does this PR do?

This PR adds the new Page `coverImage` field to the Page Entity JSON serializer.

The format is copied from the Campaign [`coverImage` field](https://github.com/DoSomething/phoenix-next/blob/03a3b0200d1c26c1b005dad87cd225d7ea32fcb0/app/Entities/Campaign.php#L123) aside from the `landscapeUrl` which I figured we wouldn't need.

#1150


weerd:
> The format is copied from the Campaign coverImage field aside from the landscapeUrl which I figured we wouldn't need.

I think we actually _do_ want the `landscapeUrl`, since these images will show up at the top of an article page and landscape style is MUCH better for that kind of layout.

PR #1152, opened 2018-10-22
Add cover image

mendelB:
### What does this PR do?

This PR adds the optional Cover Image prop to content pages `GeneralPage` and outputs the Cover Image in landscape format.

A couple of things I did which are hopefully correct:
- I omitted to pass a format option to the `contentful_image_url` helper, allowing editors the discretion of choosing images that work well, without any cropping on our part.
- used a `LazyImage` here for a more graceful image load
- set the `alt` attribute to the assets `description`, defaulting to `'Page Cover Image'`
- placed the Cover Image *under* the horizontal rule. (should it be on top instead? As more of a heading to the article?)

### Checklist

* [ ] Documentation added for new features/changed endpoints. (*will add post merge*)

image

image

image

(Here's how it could look on top of the rule:)
image

### DIFF
--- a/resources/assets/components/pages/GeneralPage/GeneralPage.js
+++ b/resources/assets/components/pages/GeneralPage/GeneralPage.js
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
 import Enclosure from '../../Enclosure';
+import LazyImage from '../../utilities/LazyImage';
 import Byline from '../../utilities/Byline/Byline';
 import ContentfulEntry from '../../ContentfulEntry';
 import AuthorBio from '../../utilities/Author/AuthorBio';
@@ -26,6 +27,7 @@ const GeneralPage = props => {
     authors,
     title,
     subTitle,
+    coverImage,
     content,
     sidebar,
     blocks,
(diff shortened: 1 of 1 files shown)


weerd:
> placed the Cover Image under the horizontal rule. (should it be on top instead? As more of a heading to the article?)

Indeed, we want under the horizontal rule :)

resources/assets/components/pages/GeneralPage/GeneralPage.js
```diff
@@ -66,6 +68,14 @@ const GeneralPage = props => {
             ) : null}
           </div>
 
+          {coverImage.url ? (
+            <LazyImage
+              className="padding-vertical-md margin-horizontal-auto"
+              alt={coverImage.description || 'Page Cover Image'}
+              src={contentfulImageUrl(coverImage.url, 1440, 620)}
```
weerd: Yeah, I think this should work fine for now. We'll see if we need to change it at all depending on the images editors are adding, and maybe adjust, but seems good to me for now :)

resources/assets/components/pages/GeneralPage/GeneralPage.js
```diff
@@ -66,6 +68,14 @@ const GeneralPage = props => {
             ) : null}
           </div>
 
+          {coverImage.url ? (
+            <LazyImage
+              className="padding-vertical-md margin-horizontal-auto"
+              alt={coverImage.description || 'Page Cover Image'}
```
weerd: Would this not just be a `defaultProp`? Or is it coming in as `null` and thus not overriding w/ the default?

weerd: Ah wait, since this it a prop of a prop. Ok, nevermind!

mendelB: lol I had the exact same train of thought 🚂

weerd: synergy

PR #1154, opened 2018-10-23
Page gallery items

mendelB:
### What does this PR do?

This PR adds a new gallery item template component called `PageGalleryItem` for GalleryBlock items of type `page`.

Added `Page` as a valid reference for the Gallery Block `blocks` field.

~Also added a `link` prop to the `Figure` component to allow rendering the Figure media in a link.~

The [Ashes page gallery](https://www.dosomething.org/us/about/our-people) we're trying to replicate rendered the images as links as well as the caption. The way it seemed to be done there was by adding a `url` to the gallery node, and then the figure template would render the image within a link (https://git.io/fxPRZ). So I figured (ha!) it'd be cool to somewhat replicate this here. (There are some other gallery templates which might need similar functionality like Campaigns)

**Update:** We ended up rendering the entire Page gallery item as a link, instead of adding this functionality to `Figure` (https://github.com/DoSomething/phoenix-next/pull/1154#discussion_r227504553)

Question: Might we want to add a default image for this gallery since the `coverImage` field is optional? Perhaps just an image of the DoSomething logo? Otherwise, any page without a `coverImage` added to a gallery would render pretty badly. (Though the onus could be on the editor to ensure to set a Cover Image if using a page in a gallery).

### Checklist

### Screenshots:

- **[Desktop](https://user-images.githubusercontent.com/12417657/47365195-09617880-d6a9-11e8-82d2-ba4be5d476e3.png)**

- **[Tablet](https://user-images.githubusercontent.com/12417657/47365278-357cf980-d6a9-11e8-8e8b-0f3d0298d0c9.png)**

- **[Mobile](https://user-images.githubusercontent.com/12417657/47365354-64936b00-d6a9-11e8-9fc0-9b9ed677cded.png)**

### DIFF
--- a/resources/assets/components/Figure/index.js
+++ b/resources/assets/components/Figure/index.js
@@ -13,7 +13,6 @@ export const BaseFigure = ({
   media,
   size,
   className,
-  link,
   children,
 }) => (
   <article
@@ -23,19 +22,7 @@ export const BaseFigure = ({
       modifiers(alignment, verticalAlignment, size),
     )}
   >
-    <div className="figure__media">
-      {link ? (
-        <a
-          target={isExternal(link) ? '_blank' : '_self'}
-          rel="noopener noreferrer"
-          href={link}
-        >
-          {media}
-        </a>
-      ) : (
-        media
-      )}
-    </div>
+    <div className="figure__media">{media}</div>
     <div className="figure__body">{children}</div>
   </article>
 );
@@ -52,7 +39,6 @@ BaseFigure.propTypes = {
   verticalAlignment: PropTypes.oneOf(['center']),
   size: PropTypes.oneOf(['small', 'medium', 'large', 'one-third']),
   media: PropTypes.node,
-  link: PropTypes.string,
 };
 
 BaseFigure.defaultProps = {
@@ -62,7 +48,6 @@ BaseFigure.defaultProps = {
   verticalAlignment: null,
   size: null,
   media: null,
-  link: null,
 };
 
 export const Figure = props => {
--- a/resources/assets/components/Figure/index.js
+++ b/resources/assets/components/Figure/index.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
-import { modifiers } from '../../helpers';
+import { modifiers, isExternal } from '../../helpers';
 import LazyImage from '../utilities/LazyImage';
 
 import './figure.scss';
@@ -13,6 +13,7 @@ export const BaseFigure = ({
   media,
   size,
   className,
+  link,
   children,
 }) => (
   <article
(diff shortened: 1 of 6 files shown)


mendelB:
I should note that one caveat with this approach - wherein we use the `coverImage` as the gallery item photo - is that the Cover Image will always show up atop the page, even if you only wanted to set it for the sake of the gallery preview. So, for example, this Our People page gallery that we're replicating - on Ashes the gallery item photos do not show up on the actual pages themselves but will have to as part of the way we've implemented it.

Not saying this is necessarily an issue but just flagging.

mendelB:
^^
As discussed in our team Sprint Plan, we're comfortable with this arrangement (oh coupling the Cover Image display within the page and its function as a Gallery item photo) since it doesn't necessarily look bad, and keeps things simpler for content editors!

resources/assets/components/utilities/PageGalleryItem.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Figure } from '../Figure';
+import { contentfulImageUrl } from '../../helpers';
+
+const PageGalleryItem = ({ title, subTitle, coverImage, slug }) => (
+  <Figure
```
weerd: Instead of modifying the `Figure` block, could we not just wrap this in an `<a>` so the whole item can be clicked and go to the page?

mendelB: Do you mean wrapping the whole `Figure` in an `<a>`?

It causes the description text to be bolded (which I guess we can override for this component). Also in Ashes, the description text isn't clickable, it's just the image and caption as their own separate links, which was the pattern I was copying https://github.com/DoSomething/phoenix/blob/dev/lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/figure.tpl.php#L4...#L7

That being said - do you prefer it this way, leaving the `Figure` alone?

mendelB: Ok, can definitely do this. (We can always refactor if we end up seeing value in adding `Figure`)

> While your at it, thoughts on placing this template in a utilities/Gallery/templates/ directory?
Yes! I wasn't sure where to put this, I forgot that I meant to bring that up. I think that's a sensible solution, kinda jives with our `Person` setup.

resources/assets/components/Figure/index.js
```diff
@@ -22,7 +23,19 @@ export const BaseFigure = ({
       modifiers(alignment, verticalAlignment, size),
     )}
   >
-    <div className="figure__media">{media}</div>
+    <div className="figure__media">
```
weerd: This can be removed now right? Since wrapping `Figure` in `<a>`?

resources/assets/components/Figure/index.js
```diff
@@ -13,6 +13,7 @@ export const BaseFigure = ({
   media,
   size,
   className,
+  link,
```
weerd: This can be removed now right? Since wrapping `Figure` in `<a>`?

mendelB: Wow whoops, forgot to push that up!

resources/assets/components/utilities/Gallery/templates/PageGalleryItem/page-gallery-item.scss
```diff
@@ -0,0 +1,7 @@
+@import '../../../../../scss/next-toolbox.scss';
+
+.page-gallery-item {
```
weerd: Add a `display: block` to the `<a>` tag. It'll lead to less bugs if it's takes up the correct amount of area rather an thinking it's supposed to still default to `inline`. Either add as an item in the scss file or we might have a utility class for it 🤔

mendelB: We did indeed! Added.

PR #1162, opened 2018-10-26
Dynamic Source Token for Link Action

mendelB:
### What does this PR do?

This PR pulls the `state.user.source` field as a new `source` prop for the `LinkAction` and uses it as a valid dynamic token for `link`.

The `source` prop in the container defaults to `undefined` if `null`, allowing the default prop to be set as `'web'`, since that was the hardcoded value previously configured.

###  background context you want to provide?
The `user.source` field is set initially in the [`AppServiceProvider`](https://github.com/DoSomething/phoenix-next/blob/master/app/Providers/AppServiceProvider.php#L49), pulled from the `utm_source` query parameter.

This allows us to dynamically source link actions, instead of manually creating separate links, and thus content pages for SMS and WEB. (more deets in the Pivotal ticket.)

To test this, go to https://dosomething-phoenix-de-pr-1162.herokuapp.com/us/testing-dynamic-source and append any `utm_source=` param. The link on the page should have the correct source set based on your input.

Here's the entry https://app.contentful.com/spaces/81iqaqpfd8fy/environments/dev/entries/S5tdCOzMwSa6KI4eegUCG?previousEntries=1RAcQWG72cCCOQiweO0Qg2

Refs [Pivotal ID #161090964](https://www.pivotaltracker.com/story/show/161090964)

### DIFF
--- a/resources/assets/components/actions/LinkAction/LinkActionContainer.js
+++ b/resources/assets/components/actions/LinkAction/LinkActionContainer.js
@@ -1,17 +1,19 @@
 import { connect } from 'react-redux';
 
 import LinkAction from './LinkAction';
+import { withoutNulls } from '../../../helpers';
 import { getUserId } from '../../../selectors/user';
 
 /**
  * Provide state from the Redux store as props for this component.
  */
-const mapStateToProps = state => ({
-  userId: getUserId(state),
-  campaignId: state.campaign.legacyCampaignId,
-  campaignRunId: state.campaign.legacyCampaignRunId,
-  source: state.user.source || undefined,
-});
+const mapStateToProps = state =>
+  withoutNulls({
+    userId: getUserId(state),
+    campaignId: state.campaign.legacyCampaignId,
+    campaignRunId: state.campaign.legacyCampaignRunId,
+    source: state.user.source,
+  });
 
 // Export the container component.
 export default connect(mapStateToProps)(LinkAction);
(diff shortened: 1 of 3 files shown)


resources/assets/components/actions/LinkAction/LinkActionContainer.js
```diff
@@ -10,6 +10,7 @@ const mapStateToProps = state => ({
   userId: getUserId(state),
   campaignId: state.campaign.legacyCampaignId,
   campaignRunId: state.campaign.legacyCampaignRunId,
+  source: state.user.source || undefined,
```
DFurnes: I wish default props took effect with `null` as well as undefined. 😖 How about we use [`withoutNulls`](https://github.com/DoSomething/phoenix-next/blob/41c29caf536229b9219290acdec419e875169a8d/resources/assets/helpers/index.js#L727-L736) here?

```js
const mapStateToProps = state => withoutNulls({
   // ...
   source: state.user.source,
});
```

DFurnes: And not something we need to tackle for this PR, but it's a bummer to have to snag these same fields in each container that we use `dynamicString` in. I wonder if we should just make a generic `DynamicTokenContainer` that we can wrap any of these components in...

mendelB: That sounds so smart! I'll add an icebox 🎫 for this

PR #1163, opened 2018-10-29
Gallery Block Image Alignment field

mendelB:
### What does this PR do?

This PR updates the Gallery Block content type with a new field called `imageAlignment` 
- adds new field to the content type Contentful migration script
- adds new field to the Gallery Block Entity

This allows the editor to toggle between the two currently existing gallery types - with the image on top of the text or right aligned.

https://www.pivotaltracker.com/story/show/160603556/comments/195829585

### DIFF
--- a/contentful/content-types/galleryBlock.js
+++ b/contentful/content-types/galleryBlock.js
@@ -1,6 +1,6 @@
 module.exports = function(migration) {
   const galleryBlock = migration
-    .createContentType('galleryBlock')
+    .createContentType('galleryBlock2')
     .name('Gallery Block')
     .description('Displays a gallery of referenced entries')
     .displayField('internalTitle');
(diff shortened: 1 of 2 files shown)


contentful/content-types/galleryBlock.js
```diff
@@ -42,6 +42,14 @@ module.exports = function(migration) {
     .required(true)
     .localized(false);
 
+  galleryBlock
+    .createField('imageAlignment')
+    .name('Image Alignment')
+    .type('Symbol')
+    .validations([{ in: ['top', 'left'] }])
+    .required(true)
```
weerd: I can't remember, but is there a way to have `top` preselected in Contentful in the UI? I don't think so, but want to confirm.

mendelB: Pretty sure it's not possible 😞 
Though there seems to be a Contentful developed UI Extension for this!
https://www.contentful.com/developers/marketplace/default-field-value/

weerd: 🤔

PR #1167, opened 2018-11-01
Refactoring how we get signups for campaign.

weerd:
### What does this PR do?

This PR replaces the `checkForSignup()` method in the `signup` action with a new `getCampaignSignups()` method that is better integrated with some of the new improvements to the system, such as the API middleware.

Updates include:
- add the new `getCampaignSignups()` method in actions/signup.js which follows some of the patterns/approaches used in actions/post.js.
- Removing the need to pass a `token` within an `apiRequest()` action, thanks to the new `setRequestHeaders()` in [helpers/api.js](https://github.com/DoSomething/phoenix-next/pull/1167/files#diff-1d982c1cc2fd87afdf2617f9c4d7f964R13), which helps modify and/or add some default items when configuring API request headers.
- adds some [new action-type constants](https://github.com/DoSomething/phoenix-next/pull/1167/files#diff-94514e42e69d61ead37abc7dde379f08R6), specifically related to campaign signups.
- Officially puts into play the `getRequestAction()` in the [API middleware](https://github.com/DoSomething/phoenix-next/pull/1167/files#diff-252fe66a09d3de77eb4b779ca827a98dR43).
- Also cleans up the middleware, and adds a [new `tabularLog()` method](https://github.com/DoSomething/phoenix-next/pull/1167/files#diff-252fe66a09d3de77eb4b779ca827a98dR20), which I would eventually like to move to a dedicated Debugger or Logger service class.
- Updates the reducers/signup.js to use the [new campaign signup action-type constants](https://github.com/DoSomething/phoenix-next/pull/1167/files#diff-b062b29874e9e262b92c525f6358c79dR29) and affect the redux store accordingly.
- Cleans up selectors/user.js and adds a [new `getUserToken()` method](https://github.com/DoSomething/phoenix-next/pull/1167/files#diff-07edf9650ad6e8665db0acbddbf6c3b2R27).

This is part one of a multi-part approach to slowly removing Campaign Run IDs from Phoenix and fixing up signups across the system.

### Checklist


resources/assets/reducers/signups.js
```diff
@@ -11,15 +12,42 @@ import {
...
+        signups = [...state.data, data[0].campaign_id];
+
+        storageSet(action.meta.northstarId, SIGNUP_STORAGE_KEY, signups);
+      }
+
+      return {
+        ...state,
+        data: signups,
+        isPending: false,
+        thisCampaign: data.length ? true : false, // eslint-disable-line no-unneeded-ternary
```
weerd: Is there a better way I could write this? 🤔 I can't just use `data.length` cause currently w/ RunIDs it can be more than just `0` or `1`...

weerd: Oh maybe I wrap it in `Boolean(data.length)`?

DFurnes: Yep! `Boolean(data.length)` should do what you want here!

DFurnes: And actually – doesn't need to be in this pull request, but we should remove this `thisCampaign` boolean in favor of a selector that checks if the given signup ID is in the `signups` array (since we don't want computed state in our Redux store).

weerd: Ah, yeah, defs! Will include in a later PR as you mention, since trying to step-through the changes incrementally!

mendelB:
Just catching this. Straight up fire!

PR #1169, opened 2018-11-01
Removing unused method and constants.

weerd:
### What does this PR do?

This PR removes a vestige that signup action method and constant that is seemingly no longer used. I'm refraining from removing the `signups.total` property from the store until I can fully confirm that nothing is using it.

### Checklist


weerd:
😂

PR #1171, opened 2018-11-05
Homepage Content Type

mendelB:
### What does this PR do?

This PR adds a Contentful migration script and Entity for the new up and coming HomePage content type.

Should it be `Homepage` as opposed to `HomePage`? 🤔 


weerd:
I think `HomePage` works and follows well with `LandingPage` 😄

PR #1172, opened 2018-11-05
Home Page scaffolding

mendelB:
### What does this PR do?

This PR sets up the backend for our HomePage on Phoenix!

Adds
- a `HomePageController` as a basic [single action controller](https://laravel.com/docs/5.5/controllers#single-action-controllers).
- a `HomePageRepository` with a `getFirst` method to grab the first `HomePage` entry from Contentful (well, it grabs all of 'em and then returns the first, see below).
- backend routing for homepage as well as a basic template React route.

1) I was deliberating between adding a new method to the `QueriesContentful` trait to allow grabbing the first returned entry from Contentful for a given type, but figured we can just use the `getEntriesAsJson` to grab all the `HomePage`s, and then manually return the first element, since there shouldn't be more then one existing HomePage entry to begin with, and we're also caching these. Was this the right call, or is it worth adding a new query method?
UPDATE: *We added an optional `limit` parameter to the `getEntriesAsJson`*

2) Is it worth throwing in failsafes for when there are *no* homepage entries at all? (e.g. currently the controller is expecting one to be returned and calling `$homePage->fields->title` point blank).
UPDATE: *redirecting to explore campaigns if no homepage is found*

### Checklist

**It's beautiful!**
image

### DIFF
--- a/app/Http/Controllers/HomePageController.php
+++ b/app/Http/Controllers/HomePageController.php
@@ -33,6 +33,10 @@ class HomePageController extends Controller
     {
         $homePage = $this->homePageRepository->getFirst();
 
+        if (! $homePage) {
+            return redirect('/us/campaigns');
+        }
+
         return view('app', [
             'pageTitle' => $homePage->fields->title,
         ])->with('state', [
--- a/app/Http/Controllers/HomePageController.php
+++ b/app/Http/Controllers/HomePageController.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Repositories\HomePageRepository;
+
+class HomePageController extends Controller
+{
+    /**
+     * The home page repository.
+     *
+     * @var HomePageRepository
+     */
+    protected $homePageRepository;
+
+    /**
+     * Make a new HomePageController, inject dependencies,
+     * and set middleware for this controller's methods.
+     *
+     * @param HomePageRepository $homePageRepository
+     */
+    public function __construct(HomePageRepository $homePageRepository)
+    {
+        $this->homePageRepository = $homePageRepository;
+    }
+
+    /**
+     * Display the homepage.
+     *
+     * @return \Illuminate\View\View
+     */
+    public function __invoke()
+    {
+        $homePage = $this->homePageRepository->getFirst();
+
+        return view('app', [
+            'pageTitle' => $homePage->fields->title,
+        ])->with('state', [
+            'homePage' => $homePage,
+        ]);
+    }
+}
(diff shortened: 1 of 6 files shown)


weerd:
> I was debilitating between adding

I don't think you mean [this](https://www.google.com/search?q=define%3A+debilitating&oq=define%3A+debilitating&aqs=chrome..69i57j69i58.4846j0j7&sourceid=chrome&ie=UTF-8)... 😂

mendelB:
😆 I think I might just be 🤦‍♂️

weerd:
Regarding item 1, what if we add an optional param to set a `limit`: 
```php
getEntriesAsJson($type, $limit = null) {...}
``` 

And thus when calling it only ever grabs 1 (which I'm assuming would be the latest one, but defs something to test!)

Either that or doing a specific method. Not sure if there will be other requirements to would benefit having a dedicated method 🤔

mendelB:
That sounds really sensible, I'll give that a whirl! As of now I can't think of any requirement necessitating an individual method 🤷‍♂️

mendelB:
@weerd Works like a charm! And returns the most recently updated entry

weerd:
> And returns the most recently updated entry

Noice! And on Prod it'd be the most recent and _published_ entry which should be good enough! 👌🏽

weerd:
> Is it worth throwing in failsafes for when there are no homepage entries at all? (e.g. currently the controller is expecting one to be returned and calling $homePage->fields->title point blank)

Yeah, probably... maybe some smart default content or just grab like the first 6 or so most recently updated & published campaigns?

app/Repositories/QueriesContentful.php
```diff
@@ -13,14 +14,16 @@
...
      */
-    protected function getEntriesAsJson($type)
+    protected function getEntriesAsJson($type, $limit = null)
     {
         $query = (new Query)
                 ->setContentType($type)
                 ->setInclude(0)
-                ->orderBy('sys.updatedAt', true);
+                ->orderBy('sys.updatedAt', true)
+                ->setLimit($limit);
```
weerd: just want to confirm that passing `null` to `setLimit()` doesn't affect existing functionality. I wasn't sure if maybe use a conditional 🤷🏽‍♂️ If what you have works, that is defs preferred!

mendelB: Thankfully it does not affect existing functionality! e.g. explore campaigns still outputs the index of all campaigns

mendelB:
UPDATE:
Added a redirect for when no homepage entry is found.
Updated the React route to be `exact` and to use `render` instead of `component`

app/Http/Controllers/HomePageController.php
```diff
@@ -0,0 +1,46 @@
...
+     * Display the homepage.
+     *
+     * @return \Illuminate\View\View
+     */
+    public function __invoke()
+    {
+        $homePage = $this->homePageRepository->getFirst();
+
+        if (! $homePage) {
+            return redirect('/us/campaigns');
```
weerd: 👌

PR #1174, opened 2018-11-06
Refactoring signup action

weerd:
### What does this PR do?

This PR helps establish a new, simpler flow for the signup action process. It helps simplify a few of the actions in the `actions/signup.js` file and makes things a bit easier to follow.

It stubs some code out in the `CampaignSignupsController` and `SignupRepository` for upcoming PR to connect sending signups to Rogue v3 endpoint.

@DFurnes and I also worked together to address using a simpler approach for handling post-authentication signups on a campaign. It now uses a new browser storage package called `localforage` that more robustly stores data in the browser, and combining this with Laravel's sessions to know when a user was redirected back to a campaign after logging in via DS Identity (Northstar) and picking back up the rest of the signup process.

We are aiming to get rid of the `actions/events.js`, `reducers/events.js` and associated `queue` in the redux store which currently is a bit convoluted. We're also going to get rid of the `helpers/storage.js` in favor of the `localforage` package which handles storing data in the browser and provides lots of useful methods.

Refs [Pivotal ID #152003980](https://www.pivotaltracker.com/story/show/152003980)


resources/assets/actions/signup.js
```diff
@@ -111,10 +117,65 @@ export function getCampaignSignups(id = null, query = {}) {
...
-export function storeCampaignSignup() {}
+/**
+ * Dispatch action to store signup for a campaign.
+ *
+ * @param  {String} campaignId
+ * @param  {Object} data
+ * @return {Function}
+ */
+export function storeCampaignSignup(campaignId, data) {
+  // @TODO: if no campaignId or data let's throw an error!
```
DFurnes: Static typpppiiiiiing! 🧟‍♂️

weerd: via flow-types right? We should likely revisit post-Ashes migration 🤔

DFurnes: Yeah, Flow or Typescript would be the two front-runners here.

resources/assets/actions/signup.js
```diff
@@ -111,10 +117,65 @@ export function getCampaignSignups(id = null, query = {}) {
...
+  };
+
+  return dispatch => {
+    dispatch(
+      apiRequest('POST', {
+        body: { ...data },
+        failure: 'failure_message',
+        meta: {
+          analytics,
+          sixpackExperiments: { conversion: 'signup' },
```
mendelB: Would this be causing a second conversion on the `'signup'` in addition to the dispatched `convertOnSignup` action in the `clickedSignupAction`?
https://github.com/DoSomething/phoenix-next/pull/1174/files#diff-9e724eddc3f7ebfa3cb156868580310aR170

resources/assets/actions/signup.js
```diff
@@ -7,6 +7,7 @@ import { Phoenix } from '@dosomething/gateway';
 import apiRequest from './api';
 import { isCampaignClosed } from '../helpers';
 import { getUserId, isAuthenticated } from '../selectors/user';
+import { getCampaignAffiliateMessagingOptOutFlag } from '../selectors/campaign';
```
mendelB: 🥇

app/Http/Controllers/AuthController.php
```diff
@@ -46,10 +46,14 @@ public function getLogin(ServerRequestInterface $request, ResponseInterface $res
         if (! array_has($request->getQueryParams(), 'code')) {
             $intended = session()->pull('url.intended', url()->previous());
             session(['login.intended' => $intended]);
+
+            // If starting the login process, see if we have a "queued" action.
+            $actionId = array_get($request->getQueryParams(), 'actionId');
+            session()->flash('actionId', $actionId);
```
mendelB: Why do we need to store the `actionId` in the session -- would we ever be redirected back to this action after having already hit it in a previous request?
(I also think I just don't have a good grasp on this Controller action, to begin with, which is why I may be misunderstanding something here.)

resources/assets/actions/signup.js
weerd: Yeah, I was curious about that as well, but I don't think so. Sixpack only converts once for a user, regardless how many times a conversion happens, so I _think_ we're ok!

resources/assets/actions/signup.js
weerd: Naming..., amiright 😉

resources/assets/actions/signup.js
mendelB: Ah sweet!

app/Http/Controllers/AuthController.php
weerd: Oh ya. That is a good point. I believe it's still useful, but may not be 100% necessary. I'll confirm in follow up PR where I wrap up connecting this all with Rogue! But defs good question :)

PR #1176, opened 2018-11-08
11 Facts Ashes -> Contentful migration

katiecrane:
### What does this PR do?
It's a script to grab all the 11 Facts pages from Ashes and shove them into Contentful! It works like this:
1. Grabs the HTML from [this page where we list a link to every 11 Facts page](https://www.dosomething.org/us/about/here-are-all-11-facts)
2. Create an array of just the urls listed on that page
3. Visit each url and grab the HTML from it
4. Take that HTML and format the boilerplate, facts, and sources how we want them and save that as one chunk of markdown. Also save the title and slug.
5. For each of those chunks of markdown, create a page in Contentful using that markdown as the content, title as the title and internal title, and slug as the slug.

Log output will look like this:
```
1. Starting 11 Facts import script...
2. Getting raw HTML from 11 Facts list page
3. Got HTML from 11 Facts list page
4. Getting HTML from each 11 Facts page: 177 total
5. Got HTML from all 11 Facts pages
6. Starting to format markdown
finished formatting markdown for page number: 1
finished formatting markdown for page number: 2
finished formatting markdown for page number: 3
finished formatting markdown for page number: 4
finished formatting markdown for page number: 5
7. Creating Contentful pages for each 11 Facts page
-   Created Fact Page! 11 Facts About Animal Cruelty [ ID: pnn0YCvlfwOamU6YkYE4M]
-   Created Fact Page! 11 Facts About Animal Homelessness [ ID: 5xfSec1CTKgMssIKMQgQYa]
-   Created Fact Page! 11 Facts About Animal Testing [ ID: 8XaJ9jlmE0q6AaYgO0uAE]
-   Created Fact Page! 11 Facts About Animals and Factory Farms [ ID: vbPnS4ATsWygiWKsm4CuE]
-   Created Fact Page! 11 Facts About Animals in Entertainment [ ID: 12HWTkXn8e2acooso0eoau]
```

Slack - [context about some formatting](https://dosomething.slack.com/archives/C3ASB4204/p1540582140006000)

Refs [Pivotal ID #161320138](https://www.pivotaltracker.com/story/show/161320138)

### Checklist

cc: @DoSomething/team-bleed


contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
```diff
@@ -0,0 +1,142 @@
...
+    allSlugs.push(splitUrl[splitUrl.length - 1]);
+    factsHtml.push(html);
+  }
+  logger.info('5. Got HTML from all 11 Facts pages');
+
+  // Format the HTML into the markdown that we want for the page body for each 11 Facts page
+  var allMarkdown = [];
+  var allTitles = [];
+  var boilerplate =
+    'Welcome to [DoSomething.org](https://www.dosomething.org), a global movement of 6 million young people making positive change, online and off! The 11 facts you want are below, and the sources for the facts are at the very bottom of the page. After you learn something, Do Something! Find out how to [take action here](https://www.dosomething.org/us/campaigns).\n\n';
```
mendelB: I don't think our Markdown parser supports the `/n` for newlines 😞

Getting two consecutive newlines is probs a bit more tricky too.
I once tried using the tip in [this SO answer](https://stackoverflow.com/a/28633712) and it seemed to work for me.
Though not sure how this would work via dynamic creation as I've only tested this out manually in the Markdown field on Contentful.
(Maybe try something like the newline character `&nbsp;` followed by the carriage return character `&#13;` 🤔)

mendelB: But on the other hand - the list *is* showing up following at least one newline after the boiler plate, so not sure that I'm 100 percent correct, but would definitely be nice to have two newlines before the list, which potentially might work via those ASCII supported codes 🤷‍♂️

package.json
```diff
@@ -105,7 +105,8 @@
     "redux-logger": "^2.8.1",
     "redux-thunk": "^2.3.0",
     "regenerator-runtime": "^0.11.1",
-    "sixpack-client": "^2.1.0"
+    "sixpack-client": "^2.1.0",
+    "turndown": "5.0.1"
```
mendelB: I think we can just add the management API script specific requirements to `devDependencies` since we only run these scripts locally

contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
```diff
@@ -0,0 +1,142 @@
...
+  // Select all the 11 Facts link elements
+  const factsLinksHTML = [
+    ...rawFactsList.window.document.querySelectorAll(
+      '.container__block > ul > li > a',
+    ),
+  ];
+  logger.info('3. Got HTML from 11 Facts list page');
+
+  // Push the each 11 Facts url into an array
+  var factsLinks = [];
```
mendelB: Completely a JS nitpickery, and also out of curiosity -- can we get away with using `let`s instead of `var`s?

(I've seen arguments in favor of `let` due to its more clearly defined scope limitations, but definitely cannot authoritatively say that it's the better call!)

contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
```diff
@@ -0,0 +1,142 @@
...
+  logger.info(
+    '4. Getting HTML from each 11 Facts page: ' + factsLinks.length + ' total',
+  );
+  var factsHtml = [];
+  var allSlugs = [];
+  var html = '';
+  // @TODO: use factsLinks.length for i < _____  *****************************
+  for (i = 0; i < 5; i++) {
+    html = await JSDOM.fromURL(factsLinks[i], '');
+    splitUrl = factsLinks[i].split('/');
```
mendelB: I think this is a rogue global variable?

contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
```diff
@@ -0,0 +1,142 @@
...
+
+      // Put it all together!
+      markdown +=
+        counter +
+        '. ' +
+        turndownService.turndown(fact) +
+        '^[' +
+        plainSource +
+        ']' +
+        '\n';
```
mendelB: Hmm another `/n` that does seem to be affective! So maybe indeed what I [said above](https://github.com/DoSomething/phoenix-next/pull/1176/files#r232043617) only applies to double newlines when adding these dynamically 🤔

contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
katiecrane: @mendelB hmm responding to this and the other newline comment below - it shows up in the Contentful editor as:
```
Welcome to [DoSomething.org](https://www.dosomething.org), a global movement of 6 million young people making positive change, online and off! The 11 facts you want are below, and the sources for the facts are at the very bottom of the page. After you learn something, Do Something! Find out how to [take action here](https://www.dosomething.org/us/campaigns).

1. Only 1 out of every 10 dogs born will find a permanent home.^[The Mosby Foundation. "Truth of what happens to shelter animals everyday." Web Accessed February 23, 2015.]
```

So I believe the newlines _are_ being applied? Did you want to add another one?

contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
```diff
@@ -0,0 +1,142 @@
+const { contentManagementClient } = require('./contentManagementClient');
+const { attempt, createLogger, sleep, withFields } = require('./helpers');
+const { JSDOM } = require('jsdom');
+const TurndownService = require('turndown');
```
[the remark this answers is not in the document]
katiecrane: It's [an npm package](https://www.npmjs.com/package/turndown) that takes in straight HTML and converts it into markdown format instead! @DFurnes found and suggested it and it has worked great!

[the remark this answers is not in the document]
chloealee: So cool!!

contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
katiecrane: @mendelB whoops was thinking of another variable, added a `let` to this one

contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
katiecrane: @mendelB updated to `let`, everything seems to work!

contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
mendelB: Nice! I guess when it gets applied from the string into the field itself, COntentful does register at those `/n`. 

I dunno, do you think it's a little tight? it looks like on ashes there are two newlines between the boiler and the facts

katiecrane: @mendelB let's do a little before and after.
as it stands now:
image

with an additional newline:
image

I think it looks the same but the markdown has updated to:
```
Welcome to [DoSomething.org](https://www.dosomething.org), a global movement of 6 million young people making positive change, online and off! The 11 facts you want are below, and the sources for the facts are at the very bottom of the page. After you learn something, Do Something! Find out how to [take action here](https://www.dosomething.org/us/campaigns).


1. Only 1 out of every 10 dogs born will find a permanent home.^[The Mosby Foundation. "Truth of what happens to shelter animals everyday." Web Accessed February 23, 2015.]
```

katiecrane: @mendelB hmm I also tried using `&#13;` at the end of the string but it showed up in plain text

mendelB: Ah damn. I guess the catch here is that we're adding this Markdown dynamically, so the tricks I used in the field itself wouldn't necessarily work here 😞

katiecrane: @mendelB even through the editor in Contentful if I add more newlines it still shows up the same way on the frontend 😕

mendelB: The trick is the newlines need to have a double space -- so something like:
```markdown
We have lots of committees within our Advisory Board because we do lots of stuff -- and love lots of help. Each committee is comprised of people who let us draft off their talent, intelligence, and experience. (They have plenty of each.)
&nbsp;
[double space]
&nbsp;
[double space] 
hi!
```

renders:
image

But either way - sorry, I'm driving you crazy about this, and it's probably just fine how it is!

contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
DFurnes: Yeah, I think these are getting turned into real newlines when you use them in a JavaScript string (for example, you'll see actual newlines if you use `\n` in a `console.log`). My guess is that Contentful (rightfully) doesn't like editors typing special escape characters like this, so they get escaped in the UI as `\\n` and  just passed on along to the frontend when typed by an editor.

contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
DFurnes: `let`s do this!™

contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js
katiecrane: @mendelB oh hmmm let me try one more tinker with this

katiecrane: @mendelB okay I tried adding this to the end of `boilerplate`: `\n&nbsp;\n\n&nbsp;\n\n` but it came out like this in Contentful:
```
Welcome to [DoSomething.org](https://www.dosomething.org), a global movement of 6 million young people making positive change, online and off! The 11 facts you want are below, and the sources for the facts are at the very bottom of the page. After you learn something, Do Something! Find out how to [take action here](https://www.dosomething.org/us/campaigns).
&nbsp;
&nbsp;
1. Only 1 out of every 10 dogs born will find a permanent home.^[The Mosby Foundation. "Truth of what happens to shelter animals everyday." Web Accessed February 23, 2015.]
```

mendelB: Ah well, it's admittedly a pretty hacky solution anyways, and I'm assuming the way we have it currently is perfectly fine. Sorry for sending you down this rabbit hole!

mendelB: Although. And this is purely curiosity at this point -- 
I wonder if something like this might work:

`boilerplate: \n &nbsp; \n &#32;&#32; \n &nbsp; \n &#32;&#32; /n text`

katiecrane: Down to try it!

But do we for sure like this better?

image

They are so the same to me that I have no opinion haha

mendelB: lol tbh my design eye is pretty useless, but I'd only venture to say it's better since that's how it is on Ashes currently. (Kind of a more distinct visual separation between the boilerplate and the facts? I dunno, @lkpttn is probably laughing at us rn but would def have a qualified opinion here!)

lkpttn: Mendel is correct, it's best if we can separate out the facts from the boilerplate.

katiecrane: alright @mendelB giving `\n &nbsp; \n &#32;&#32; \n &nbsp; \n &#32;&#32; \n` at the end of the boilerplate a try!

katiecrane: @mendelB I think it worked!!! @lkpttn this passes muster?
image

mendelB: OMG yes!!

katiecrane:
@mendelB or @DFurnes I think this could use another once over since now I have added code to publish the pages after creating them (in my mind that is a non-trivial bit!)

PR #1183, opened 2018-11-14
Add Homepage Component

mendelB:
### What does this PR do?

This PR adds a basic stubbed `HomePage` component to be rendered from the `/us` route in our React application.

~Also added some basic setup for the homePage data - a `homePage` object in `initialState` along with a `homePage` reducer. ~
We render a `homePage` variable directly from the HomePageController, which the view `scriptify`s into `window.HOMEPAGE`, allowing the `HomePage` component to utilize that object directly instead of relying on our redux setup. (See [this Slack thread](https://dosomething.slack.com/archives/C3ASB4204/p1542236114226100) for more context)

### Checklist


weerd:
@mendelB what's the need for a `homePage` item in the store? Is it something just internal state on component would be sufficient?

DFurnes:
Ah, you read my mind @mendelB! Yeah, definitely worth rethinking how we do this. ➕

mendelB:
Oh sweet! Yeah that's a great idea! I'll try and refactor this one to use `window.STATE`

mendelB:
And actually, I think I can just turn this to a WIP and try and flesh out the full HomePage component and ditch all the redux noise.

mendelB:
Ahh but of course - complications!

So even after removing the container and directly pulling from `window.STATE`, we still get that reducer warning, since we are still rendering the component through [`init`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/init.js) and still compiling state ([as mentioned above](https://github.com/DoSomething/phoenix-next/pull/1183#issuecomment-438841500)) 😖 

Perhaps it's indeed worth going forward with the old pattern for now, and rethinking this whole jam at a later date?

mendelB:
@weerd @DFurnes How do we feel about merging this, and I can work on the actual implementation in a separate PR?

app/Http/Controllers/HomePageController.php
```diff
@@ -39,8 +39,6 @@ public function __invoke()
 
         return view('app', [
             'pageTitle' => $homePage->fields->title,
-        ])->with('state', [
-            'homePage' => $homePage,
-        ]);
+        ])->with('homePage', $homePage);
```
mendelB: I wasn't sure whether to add this variable directly to the defined view variable above, or use `->with`. I went with the latter since it seems like we use `->with` to separate the `scriptify`d 'stateful' variables from the local view ones. Was this correct?

resources/assets/components/pages/HomePage/HomePage.js
```diff
@@ -0,0 +1,18 @@
...
+
+const HomePage = () => <h1>Home Page</h1>;
+class HomePage extends React.Component {
+  constructor() {
+    super();
+    this.state = get(window.HOMEPAGE, 'fields', {});
+  }
+
+  render() {
+    return this.state.title ? <h1>{this.state.title}</h1> : null;
```
mendelB: I was unsure of how or if we should be dealing with the possibility of the state being empty (somehow the `window.HOMEPAGE` not being set.) 
This is an edge case, since we [redirect](https://github.com/DoSomething/phoenix-next/blob/master/app/Http/Controllers/HomePageController.php#L37) to `/campaigns` of there is no HomePage found in Contentful. But we still might have a HomePage with missing fields...

Would it make sense for this component to just hard return if there is no `title` (a required field in contentful) found?

resources/assets/reducers/index.js
```diff
@@ -2,11 +2,11 @@ export admin from './admin';
 export campaign from './campaign';
 export events from './events';
 export notifications from './notifications';
+export page from './page';
```
mendelB: Name sorting leftover 🤷‍♂️

resources/assets/components/pages/HomePage/HomePage.js
weerd: I think maybe hard return if there are no campaigns in the `blocks` field, but I think for title we can just some reasonable defaults there.

mendelB: Ah ok that makes sense! I'll leave this for now if that's ok, and implement that in the fleshed out component

app/Http/Controllers/HomePageController.php
weerd: This is a good question. I'm not actually 100% sure why we did that... might be to "visually" separate, but ultimately they all get attached as vars to the view. You can see here with the campaign view, they're all vars at the same level:
image

mendelB: I guess it can't hurt to have that visual distinction then? 🤷‍♂️

weerd: 🤷‍♂️

PR #1184, opened 2018-11-16
Fixing up how the OAuth redirect data hand-off works.

weerd:
### What does this PR do?

This PR does a bit of tweaking and some fixes for the OAuth login flow and with how metadata is passed around between Phoenix and Northstar.

It updates a couple things that weren't working correctly and causing Northstar to not receive any of the metadata options to customize the Northstar login for a user to provide context with info from the origin campaign and I also slightly cleans up the `AuthController@getLogin()` method.

There was a [question in the prior pull-request](https://github.com/DoSomething/phoenix-next/pull/1174#discussion_r233610852) regarding the need to flash the `actionId` in the session.

I did some further investigating and just needed to re-familiarize myself with how the Auth flow works. Here are the steps for an anonymous user on a campaign that clicks the campaign signup button:
1. User clicks button on the TFJ campaign.
2. The `requiresAuthenticationMiddleware` notices that the user is not logged in, so it does some prep work, then creates an `actionId`, stores it locally in the browser and also appends this `actionId` as a query parameter of the login URL, and then redirects them to this URL.
3. The login route activates the `AuthController@getLogin()` method, and user has no Auth `code`, so this method does some prep work, flashes the `actionId` into the session and uses `gateway` to redirect the user over to Northstar.
4. The user logs in via Northstar (Identity), and then is redirected back to the login route with their shiny, new Auth `code` and `state`.
5. Thus, we're now back to Phoenix at the login route, which again activates the `AuthController@getLogin()` method, but this time equipped with an Auth `code`. However, there's no `actionId` in the URL from Northstar, so we check the session, find an `actionId`, grab it and append it to the intended URL which is the URL for the TFJ campaign.
6. After we're redirected to the TFJ campaign, when initializing the App, we check to see if the URL has an `actionId`, and if so, use it's value as the key to find the item that was previously stored locally in the browser, which happens to be a `storeCampaignSignup` action, and then dispatch it!
7. User is signed up for the campaign and on the action page!

We also `flash()` the `actionId` since it doesn't really need to stick around in the session past the returning request to Phoenix from Northstar.

Refs [Pivotal ID #152003980](https://www.pivotaltracker.com/story/show/152003980)

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -75,8 +75,6 @@ class AuthController extends Controller
             $url .= str_contains($url, '?') ? '&'.$actionIdParam : '?'.$actionIdParam;
         }
 
-        \Illuminate\Support\Facades\Log::info('AuthController@getLogin', [$url]);
-
         return gateway('northstar')->authorize($request, $response, $url, $destination, $options);
     }
 
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -40,35 +40,35 @@ class AuthController extends Controller
      */
     public function getLogin(ServerRequestInterface $request, ResponseInterface $response)
     {
+        $queryParams = $request->getQueryParams();
+
         // Save the post-login redirect for when the user completes the flow: either to the intended
         // page (if logging in to view a page protected by the 'auth' middleware), or the previous
         // page (if the user clicked "Log In" in the top navigation).
-        if (! array_has($request->getQueryParams(), 'code')) {
+        if (! array_has($queryParams, 'code')) {
             $intended = session()->pull('url.intended', url()->previous());
+
             session(['login.intended' => $intended]);
 
             // If starting the login process, see if we have a "queued" action.
-            $actionId = array_get($request->getQueryParams(), 'actionId');
+            $actionId = array_get($queryParams, 'actionId');
+
             session()->flash('actionId', $actionId);
         }
 
-        $options = [];
-        $jsonOptions = array_get($request->getQueryParams(), 'jsonOptions');
+        // Set options containing metadata for Northstar login.
+        $options = array_get($queryParams, 'options', []);
 
-        // Check if the JS added auth options.
-        if ($jsonOptions) {
-            $options = (array) json_decode($jsonOptions);
+        if (is_string($options)) {
+            $options = (array) json_decode($options);
         }
 
-        // As a backup check the Blade template.
-        if (! $options) {
-            $options = array_get($request->getQueryParams(), 'options') ?: [];
-        }
+        // Set destination metadata for Northstar login; defaults to "title" in $options or null.
+        $destination = array_get($queryParams, 'destination', array_get($options, 'title'));
 
-        $destination = array_get($request->getQueryParams(), 'destination');
         $url = session('login.intended', $this->redirectTo);
 
-        // Do we have an action ID in the session? If so, set it on to the destination URL.
+        // If there is an actionId in the session, set it as a query param on to the intended URL.
         if (session('actionId')) {
             $url .= '?actionId=' . urlencode(session('actionId'));
         }
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -70,8 +70,12 @@ class AuthController extends Controller
 
         // If there is an actionId in the session, set it as a query param on to the intended URL.
         if (session('actionId')) {
-            $url .= '?actionId=' . urlencode(session('actionId'));
+            $actionIdParam = 'actionId='.urlencode(session('actionId'));
+
+            $url .= str_contains($url, '?') ? '&'.$actionIdParam : '?'.$actionIdParam;
         }
+
+        \Illuminate\Support\Facades\Log::info('AuthController@getLogin', [$url]);
 
         return gateway('northstar')->authorize($request, $response, $url, $destination, $options);
     }
--- a/resources/assets/helpers/index.js
+++ b/resources/assets/helpers/index.js
@@ -18,12 +18,15 @@ export const EMPTY_IMAGE =
 /**
  * Build login redirect URL with optional context data.
  *
- * @param  {Null|Object} jsonOptions
+ * @param  {Null|Object} options
  * @param  {Null|String} actionId
  * @return {String}
  */
-export function buildLoginRedirectUrl(jsonOptions = null, actionId = null) {
-  const params = queryString.stringify({ actionId, jsonOptions });
+export function buildLoginRedirectUrl(options = null, actionId = null) {
+  const params = queryString.stringify({
+    actionId,
+    options: JSON.stringify(options),
+  });
 
   return `${window.location.origin}/next/login?${params}`;
 }
--- a/resources/assets/middleware/requiresAuthentication.js
+++ b/resources/assets/middleware/requiresAuthentication.js
@@ -23,6 +23,8 @@ const requiresAuthenticationMiddleware = ({ getState }) => next => action => {
         actionId,
       );
     });
+
+    return null;
   }
 
   return next(action);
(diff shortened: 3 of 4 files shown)


resources/assets/helpers/index.js
```diff
@@ -18,12 +18,15 @@ export const EMPTY_IMAGE =
...
+ * @param  {Null|Object} options
  * @param  {Null|String} actionId
  * @return {String}
  */
-export function buildLoginRedirectUrl(jsonOptions = null, actionId = null) {
-  const params = queryString.stringify({ actionId, jsonOptions });
+export function buildLoginRedirectUrl(options = null, actionId = null) {
+  const params = queryString.stringify({
+    actionId,
+    options: JSON.stringify(options),
```
weerd: To properly stringify the query, we can't passed as a nested object since it expects everything to be a 1-level deep object. So instead we first need to stringify the passed `options` and then stringify using `queryString` to make a valid querystring with params.

resources/assets/middleware/requiresAuthentication.js
```diff
@@ -23,6 +23,8 @@ const requiresAuthenticationMiddleware = ({ getState }) => next => action => {
         actionId,
       );
     });
+
+    return null;
```
weerd: If we don't return `null` here, I could see in the console that the execution continued on, and would attempt to send the API request and fails as expected (since not auth'd). So no point in continuing and better to stop execution and let the redirect happen!

resources/assets/selectors/campaign.js
```diff
@@ -22,11 +22,14 @@ export function getCampaignDataForNorthstar(state) {
   const data = {};
 
   if (state.campaign) {
-    data.callToAction = encodeURIComponent(state.campaign.callToAction);
-    data.coverImage = encodeURIComponent(
-      contentfulImageUrl(state.campaign.coverImage.url, '800', '600', 'fill'),
+    data.callToAction = state.campaign.callToAction;
```
weerd: I had to remove all the `encodeURIComponent()` calls since they were breaking Northstar. It looks like we don't "decode URI component" and it was outputting the encoded info as the text in the interface and the image was broken.

app/Http/Controllers/AuthController.php
```diff
@@ -40,35 +40,35 @@ public function __construct()
...
+        $queryParams = $request->getQueryParams();
+
         // Save the post-login redirect for when the user completes the flow: either to the intended
         // page (if logging in to view a page protected by the 'auth' middleware), or the previous
         // page (if the user clicked "Log In" in the top navigation).
-        if (! array_has($request->getQueryParams(), 'code')) {
+        if (! array_has($queryParams, 'code')) {
             $intended = session()->pull('url.intended', url()->previous());
+
             session(['login.intended' => $intended]);
```
mendelB: Is there any reason we don't use `session()->flash()` here, or do we need this data for more than one request?

app/Http/Controllers/AuthController.php
```diff
@@ -40,35 +40,35 @@ public function __construct()
...
+        // Set destination metadata for Northstar login; defaults to "title" in $options or null.
+        $destination = array_get($queryParams, 'destination', array_get($options, 'title'));
 
-        $destination = array_get($request->getQueryParams(), 'destination');
         $url = session('login.intended', $this->redirectTo);
 
-        // Do we have an action ID in the session? If so, set it on to the destination URL.
+        // If there is an actionId in the session, set it as a query param on to the intended URL.
         if (session('actionId')) {
             $url .= '?actionId=' . urlencode(session('actionId'));
```
mendelB: Do we need to do any checking for existing query params on the intended `$url` to see if we need to use `&` instead of `?`?

resources/assets/selectors/campaign.js
mendelB: Is that something Northstar *should* be doing? Or are we just not worried about special chars in the campaign content?

app/Http/Controllers/AuthController.php
```diff
@@ -40,35 +40,35 @@ public function __construct()
...
+        if (! array_has($queryParams, 'code')) {
             $intended = session()->pull('url.intended', url()->previous());
+
             session(['login.intended' => $intended]);
 
             // If starting the login process, see if we have a "queued" action.
-            $actionId = array_get($request->getQueryParams(), 'actionId');
+            $actionId = array_get($queryParams, 'actionId');
+
             session()->flash('actionId', $actionId);
```
mendelB: Would it at all make sense to just append this to the `$intended` URL here, before storing that to the session? That way we only have to store/retrieve one item, and we don't have to append it later on

app/Http/Controllers/AuthController.php
```diff
@@ -40,35 +40,35 @@ public function __construct()
...
 
-        // Check if the JS added auth options.
-        if ($jsonOptions) {
-            $options = (array) json_decode($jsonOptions);
+        if (is_string($options)) {
+            $options = (array) json_decode($options);
         }
 
-        // As a backup check the Blade template.
-        if (! $options) {
```
mendelB: Is this just retired legacy stuff?

resources/assets/selectors/campaign.js
weerd: I think it's something we can look into. I tested for now and seems to work ok, but a little curious if there are characters that might mess things up. Potentially a ticket to add for Northstar and to re-add this back in Phoenix. We also don't properly do this in Phoenix, since we tried to do it here, but we don't do it in the `get_login_query()` helper method when just clicking on "Login" on a campaign page.

tldr: we'd encode URI when clicking signups but not when clicking login... so figured it best to remove for now.

app/Http/Controllers/AuthController.php
weerd: Nope. It relates to clicking the "login" link on a campaign page (not the signup link). The login link in the name has an options param appended. But it didn't really make sense to have a `jsonOptions` and an `options` when neither could ever exist simultaneously, so I merged them to just be `options` now :)

app/Http/Controllers/AuthController.php
weerd: I think this will depend on whether we decide to flash the `url.intended` like in your prior comment. If we do, then I think it's ok, but if we decide we don't want to we don't want the `actionId` lingering on the url for additional requests. I'll make a ticket for re-:eyes: this in the future and mention the suggestions you have that we can look into :)

app/Http/Controllers/AuthController.php
weerd: 🤔

weerd: @mendelB I added a check for existing query params and did a test. Here are the logs:

```
[2018-11-16 05:41:12] local.INFO: AuthController@getLogin ["http://phoenix.test/us/campaigns/test-teens-for-jeans/action?actionId=auth%3A1542383661423"] 

[2018-11-16 05:41:45] local.INFO: AuthController@getLogin ["http://phoenix.test/us/campaigns/test-teens-for-jeans/action?utm_example=babaganoosh&actionId=auth%3A1542383693909"] 
```

In the second example the URL already had a fake utm param and now the actionId is attached as a subsequent param :) Good call on this!

weerd:
> This has been so helpful to my understanding of this Controller action!!! Your PR description has just been silly useful.

It defs helped me too, to write it all out 💃

app/Http/Controllers/AuthController.php
```diff
@@ -70,9 +70,13 @@ public function getLogin(ServerRequestInterface $request, ResponseInterface $res
 
         // If there is an actionId in the session, set it as a query param on to the intended URL.
         if (session('actionId')) {
-            $url .= '?actionId=' . urlencode(session('actionId'));
+            $actionIdParam = 'actionId='.urlencode(session('actionId'));
+
+            $url .= str_contains($url, '?') ? '&'.$actionIdParam : '?'.$actionIdParam;
         }
 
+        \Illuminate\Support\Facades\Log::info('AuthController@getLogin', [$url]);
```
mendelB: This was just for testing right?

weerd: gah. damn. 🤦‍♂️ Got too excited it worked well, haha.

PR #1186, opened 2018-11-17
🚧 WIP: Sending signups to rogue v3

weerd:
# 🚧 Work In Progress

### What does this PR do?

This PR completes using the new signups approach to send signups to store in Rogue using the v3 endpoint.

...

Refs [Pivotal ID #]()

### Checklist


resources/assets/reducers/signups.js
```diff
@@ -45,6 +48,27 @@ const signupReducer = (state = {}, action) => {
...
+
+    case STORE_CAMPAIGN_SIGNUPS_PENDING:
+      return {
+        ...state,
+        isPending: true,
+      };
+
+    case STORE_CAMPAIGN_SIGNUPS_SUCCESSFUL:
+      if (data) {
+        signups = [...state.data, data.campaign_id];
```
weerd: The old approach used to pop this info into local storage as well, but that was cause there was a little bit of convoluted logic in the initialization of the app that would check if this data was in local storage maybe for convenience? I can't remember that well. I debated putting it back, but I figured it'd be simpler to leave out for now to keep things simpler/not confound source of truth in data and we can decide whether we need it in the future... maybe when we're in Progressive Web App land 🤔

weerd: The old case for `SIGNUP_CREATED` still exists in this file... will be removed in upcoming PR, but you can see where this is stored in local storage in old approach on line 75.

DFurnes: Yeah, the reason for this was so we wouldn't always have to make an additional API request for users returning to campaigns they're already signed up for. I'm good with removing & seeing whether or not this has a performance/UX impact. 👍

app/Http/Controllers/Api/CampaignSignupsController.php
```diff
@@ -52,7 +53,19 @@ public function index($id, Request $request)
      */
     public function store($id, Request $request)
     {
-        // @TODO: implement sending request to Rogue.
-        return response()->json([$id, $request->all()]);
+        $request->merge(['campaign_id' => $id]);
```
DFurnes: Would it make sense to send `campaign_id` in the payload of the request (rather than URL) so we don't need to adjust it here?

resources/assets/middleware/api.js
```diff
@@ -80,13 +81,16 @@ const postRequest = (payload, dispatch, getState) => {
     headers: setRequestHeaders({ token, contentType: 'multipart/form-data' }),
```
DFurnes: We don't need to fix this now, but I think we should only set the `contentType` to `multipart/form-data` if we provide a `FormData` instance as `payload.body` here.

weerd: Oh! So this is where it's happening 💡 I'll link this comment in the icebox ticket I created for future fix!

app/Http/Controllers/Api/CampaignSignupsController.php
weerd: I wanted to continue following the pattern established with `CampaignPostsController` which helps to signal that the `$id` should be the campaign ID; based on the influential presentation by sir Adam Wathan, "Cruddy By Design" which I really dug 😄

weerd: But defs something we can discuss and maybe it's not worth it and can pass along in the request AND as part of the URL, which didn't seem ideal to me at the time.

resources/assets/reducers/signups.js
```diff
@@ -45,6 +48,27 @@ const signupReducer = (state = {}, action) => {
         thisCampaign: Boolean(data.length),
       };
 
+    case STORE_CAMPAIGN_SIGNUPS_FAILED:
+      return state;
```
mendelB: Sorry, just catching up on this! Do we need to set `isPending` to `false` here?

weerd: oh ya! Nice catch 👍 I'll toss another PR w/ the quick fix 💃

PR #1188, opened 2018-11-29
Affirmation Legacy Author Photo Update

mendelB:
### What does this PR do?

This PR ensures old Affirmations author photo fields show up in the Campaign Affirmation

**UPDATE**: SEE https://github.com/DoSomething/phoenix-next/pull/1188#issuecomment-442924467

I'm not completely sure of the complete origin story (we've had several iterations on the Affirmation content type and Entity, so it's hard to pin down when this happened) - but we stopped sending over an `author->fields->photo` for affirmations which were not using a `Staff` (or now - `Person`) linked entry as the Affirmation author. Or rather we sent over the field hardcoded to `null` which means the author photo is not showing up in legacy affirmations:
image

This PR will just use the top level `photo` field as the `author->fields->photo` for these legacy affirmations.

Bumped into this issue today as an editor attempted to create an affirmation. We're also [making some changes on the Contentful side](https://dosomething.slack.com/archives/C2BPA7M8F/p1543502240156800) to ensure more coherency and consistency going forward.

### DIFF
--- a/app/Entities/Affirmation.php
+++ b/app/Entities/Affirmation.php
@@ -26,7 +26,7 @@ class Affirmation extends Entity implements JsonSerializable
                     'active' => true,
                     'jobTitle' => 'Campaign Lead',
                     'email' => 'help@dosomething.org',
-                    'photo' => null,
+                    'photo' => get_image_url($this->photo, 'square'),
                     'alternatePhoto' => null,
                     'description' => null,
                 ],
@@ -40,7 +40,6 @@ class Affirmation extends Entity implements JsonSerializable
             'type' => $this->getContentType(),
             'fields' => [
                 'header' => $this->header,
-                'photo' => get_image_url($this->photo, 'square'),
                 'quote' => $this->quote,
                 'author' => $author,
                 'callToActionHeader' => $this->callToActionHeader,
--- a/app/Entities/Affirmation.php
+++ b/app/Entities/Affirmation.php
@@ -26,7 +26,7 @@ class Affirmation extends Entity implements JsonSerializable
                     'active' => true,
                     'jobTitle' => 'Campaign Lead',
                     'email' => 'help@dosomething.org',
-                    'photo' => get_image_url($this->photo, 'square'),
+                    'photo' => null,
                     'alternatePhoto' => null,
                     'description' => null,
                 ],
@@ -40,6 +40,7 @@ class Affirmation extends Entity implements JsonSerializable
             'type' => $this->getContentType(),
             'fields' => [
                 'header' => $this->header,
+                'photo' => get_image_url($this->photo, 'square'),
                 'quote' => $this->quote,
                 'author' => $author,
                 'callToActionHeader' => $this->callToActionHeader,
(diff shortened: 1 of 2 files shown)


app/Entities/Affirmation.php
```diff
@@ -26,7 +26,7 @@ public function jsonSerialize()
                     'active' => true,
                     'jobTitle' => 'Campaign Lead',
                     'email' => 'help@dosomething.org',
-                    'photo' => null,
+                    'photo' => get_image_url($this->photo, 'square'),
```
weerd: I don't think the `photo` in the Affirmation content type corresponds to a photo for the author but is a photo to use as like an associated image in the affirmation, right?

weerd: I think it _may_ have been used as such but I was under the impression it was a content image. Maybe that's why I made it null? 🤷‍♂️

mendelB: Hmm, maybe originally, but I was looking through the old affirmations, and it seems to have been used for the author photo itself (maybe it was an accidental cannibalism?)

weerd: Checked a few out and I'm wrong. All seem to be author photos!

mendelB: Yeah, maybe something got lost in communication to editors and folks assumed it was for the author

weerd: Based on this, seems like the original intention was to use as a content image in affirmation body copy. I think maybe we used to output it as a large image, floated in body copy and happened to be an author, instead of placing inside of the byline.

https://github.com/DoSomething/phoenix-next/blame/master/app/Entities/Affirmation.php#L43

weerd: So defs a bit confusing. I'm kind of leaning towards keeping it as `null` so it just uses the DS Logo for the author. Since if someone does add a content image here, it'll be oddly placed as a circular author image in the byline 😬

mendelB: I figured we could just disable the field entirely in Contentful (along with legacy author) so older affirmations will still work and newer ones would need to use the new Author field. (In fact I'd already done so, assuming it was just an author field, but reversed once I saw your comment!)

mendelB: Also, it wasn't defaulting to the DS logo, it was just showing a blank photo, (I think since the Byline is being passed the `null` photo with no default)

weerd: Oh maybe an opportunity to run the props through (`withoutNulls()`) 💃

mendelB: haha that'd do the trick! 

So to confirm, you'd like us to keep the legacy `author->photo` field as `null`, relying on a DS logo for legacy affirmations?

What about the `photo` field in Contentful? Can we still disable that since folks are using it as an author photo field and it has no other use case anymore?

weerd: I _thiiiiiiiiink_ so on disabling the field. Not 100% sure if when we refactor the Affirmation field it comes into play, but might be worth disabling now, so we have less to deal with post a refactor... 🤔

weerd: > So to confirm, you'd like us to keep the legacy author->photo field as null, relying on a DS logo for legacy affirmations?

👍

mendelB:
Update: we've opted to leave the `photo` field as-is till we refactor this Content Type later on. We'll use a `withoutNulls` to ensure the default DS logo image is used for older affirmation authors
(https://github.com/DoSomething/phoenix-next/pull/1188#discussion_r237583410)

PR #1192, opened 2018-12-03
Update to main nav and signups reducer.

weerd:
### What does this PR do?

This PR makes a quick update to the nav so we don't keep linking to Thor DS when clicking on the Explore Campaigns link and instead go to the `/us/campaigns` of the current env that the user is on. It also makes an update to fix the state changes on the signup reducer based on @mendelB's [comment](https://github.com/DoSomething/phoenix-next/pull/1186#pullrequestreview-180852582).

🌵

### DIFF
--- a/resources/assets/reducers/signups.js
+++ b/resources/assets/reducers/signups.js
@@ -48,7 +48,10 @@ const signupReducer = (state = {}, action) => {
       };
 
     case STORE_CAMPAIGN_SIGNUPS_FAILED:
-      return state;
+      return {
+        ...state,
+        isPending: true,
+      };
 
     case STORE_CAMPAIGN_SIGNUPS_PENDING:
       return {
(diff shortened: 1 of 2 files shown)


resources/assets/reducers/signups.js
```diff
@@ -48,7 +48,10 @@ const signupReducer = (state = {}, action) => {
       };
 
     case STORE_CAMPAIGN_SIGNUPS_FAILED:
-      return state;
+      return {
+        ...state,
+        isPending: true,
```
mendelB: Wouldn't this be `false` since the request is complete at this point?

resources/views/partials/navigation.blade.php
```diff
@@ -4,7 +4,7 @@
     <div class="navigation__menu">
         <ul class="navigation__primary">
             <li>
-                <a href="{{ phoenixLink('us/campaigns') }}">
+                <a href="{{ config('app.url') }}/us/campaigns">
```
mendelB: 🎊 🙌

resources/assets/reducers/signups.js
weerd: Dang it! My brain is tired. I was staring at one of the other values and accidentally wrote true. Thanks goodness for peer-review 🙌

mendelB: 😄 ☕️

PR #1195, opened 2018-12-04
Fix to store and updating api requests.

weerd:
### What does this PR do?

This PR fixes a bug with the `store.js` which was dispatch a `null` action, because we were not checking if the `action` retrieved from localforage had a value or not. I decided to use the `isNull()` method from Lodash so the code reads clearer, that it shouldn't do anything if the value for `action` is `null`, but this may be overkill? Thoughts? Just felt it read better than `if (!action) {...}`, but let me know if the latter is preferred!

Also does some cleanup and refactoring to move `tabularLog()` helper method into the `helpers/api.js` file, and uses it in the `getRequest()` and `postRequest()` methods, while also cleaning up the `middleware/api.js` 🛁 

🌵

### DIFF
--- a/resources/assets/store/store.js
+++ b/resources/assets/store/store.js
@@ -1,5 +1,5 @@
-import merge from 'lodash/merge';
 import localforage from 'localforage';
+import { isNull, merge } from 'lodash';
 import { createStore, combineReducers, applyMiddleware, compose } from 'redux';
 
 import initialState from './initialState';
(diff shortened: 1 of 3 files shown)


resources/assets/store/store.js
```diff
@@ -72,13 +72,15 @@ export function initializeStore(store) {
     const actionId = decodeURIComponent(query('actionId'));
 
     localforage.getItem(actionId).then(action => {
-      store.dispatch(action);
+      if (!isNull(action)) {
```
mendelB: I think `isNull` reads nicely, so no hard preference! 

General question: do we prefer the pattern of hard returning if `isNull` instead of using the negative conditional and nesting the logic under the `if` statement? (Not functionally relevant but just curious as to your thoughts.)

resources/assets/store/store.js
```diff
@@ -72,6 +72,10 @@ export function initializeStore(store) {
     const actionId = decodeURIComponent(query('actionId'));
 
     localforage.getItem(actionId).then(action => {
+      if (isNull(action)) {
+        return;
+      }
+
```
weerd: And just to clarify, typically this will not be reached with an `action` of `null` due to the prior conditional `isAuthenticated(state) && query('actionId')`, where there needs to be a query `actionId`, but I did encounter this post signup and refreshing the page, since we don't clear the URL params (and don't think it's a big deal to do so).

PR #1196, opened 2018-12-05
Sprucing Up our 404 Page 

mendelB:
### What does this PR do?

This PR spruces up our 404 view to include the chrome, navigation bar, and footer.

This was as simple as extending our `master` layout instead of the `takeover`.

Refs [Pivotal ID #161694657](https://www.pivotaltracker.com/story/show/161694657)

**BEFORE**
image

**AFTER**
image

### DIFF
--- a/resources/views/errors/404.blade.php
+++ b/resources/views/errors/404.blade.php
@@ -1,4 +1,4 @@
-@extends('layouts.takeover')
+@extends('layouts.master')
 
 @section('content')
     <div class="chrome -plain">


weerd:
Sidenote: do we want to get rid of the `takeover` template file? Or is it used elsewhere as well?

mendelB:
> Sidenote: do we want to get rid of the `takeover` template file? Or is it used elsewhere as well?

The only other use case I could find was in the [`500` view](https://github.com/DoSomething/phoenix-next/blob/master/resources/views/errors/500.blade.php) 

Which come to think of it can make use of the master treatment as well?

PR #1197, opened 2018-12-05
Adding method to check if signup recently occurred and should show modal.

weerd:
### What does this PR do?

This PR adds a way to check against a signup's `created_at` date to see if the signup occurred recently or has existed for some time. If the signup _is within minutes_ of being created, then we change the status code in the response object passed to the store from the API Middleware to be `201` to indicate record was created. In the reducer we check against this status code and show the signup affirmation modal if it is `201`.

Also rename props in the `CampaignRouteContainer` and `CampaignRoute` to match the naming convention used elsewhere (making finding in code easier later!).

Ideally we _don't_ want to rely on a slightly hacky method of comparing `created_at` date for a response from Rogue, however our Gateway package currently only passes along the data in the response from Rogue instead of also including the additional status info. Initially we probably did this to help simplify things, but in the long run, seems like it's something we want to keep. So this approach is intended to be a hacky solution, with the intent that it will be easy to update the API Middleware to pass the actual response status, instead of relying on the `isWithinMinutes()` method. Once Gateway is updated it's a simple fix!

Refs [Pivotal ID #162418535](https://www.pivotaltracker.com/story/show/162418535)

### Checklist

### DIFF
--- a/resources/assets/middleware/api.js
+++ b/resources/assets/middleware/api.js
@@ -3,7 +3,7 @@
 import { get } from 'lodash';
 import { RestApiClient } from '@dosomething/gateway';
 
-import { report } from '../helpers';
+import { report, isWithinMinutes } from '../helpers';
 import { PHOENIX_URL } from '../constants';
 import { setFormData } from '../helpers/forms';
 import { API } from '../constants/action-types';
@@ -70,12 +70,16 @@ const postRequest = (payload, dispatch, getState) => {
   return client
     .post(payload.url, body)
     .then(response => {
+      console.log('👾 POST request response: ', response);
+
       tabularLog(get(response, 'data', null));
+
+      const dataCreatedAt = get(response, 'data.created_at', Date.now());
 
       response.status = {
         success: {
-          code: 201,
-          message: 'Thanks for your submission!',
+          code: isWithinMinutes(dataCreatedAt, 10) ? 201 : 200,
+          message: get(payload, 'meta.messaging.success', 'Thanks!'),
         },
       };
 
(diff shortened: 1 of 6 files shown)


resources/assets/middleware/api.js
```diff
@@ -72,10 +72,15 @@ const postRequest = (payload, dispatch, getState) => {
     .then(response => {
       tabularLog(get(response, 'data', null));
 
+      // @TODO: Not ideal. We would prefer to know the status code from response to know
+      // if data was created or not, but Gateway doesn't currently pass this to us. So for
+      // now we're resolving to check against the data's created_at value to decide time elapsed.
+      const dataCreatedAt = get(response, 'data.created_at', Date.now());
```
mendelB: Would we want this to default to `null` so that `isWithinMinutes ` will return `false` if there's no `data.created_at`?

resources/assets/middleware/api.js
```diff
@@ -72,10 +72,15 @@ const postRequest = (payload, dispatch, getState) => {
...
+
       response.status = {
         success: {
-          code: 201,
-          message: 'Thanks for your submission!',
+          code: isWithinMinutes(dataCreatedAt, 5) ? 201 : 200,
+          message: get(payload, 'meta.messaging.success', 'Thanks!'),
```
mendelB: Is the generic `'Thanks!'` so that it works with both signups and RBs?

resources/assets/middleware/api.js
weerd: oh yeah, I think that's a good call. I think I had that left in there from an earlier version of what I was thinking. 🌮

resources/assets/middleware/api.js
weerd: yep yep! And I have an update I have planned for RBs so they can pass something similar to [this](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/actions/signup.js#L138-L141) and have more customized messaging!

PR #1198, opened 2018-12-05
Prefer window.location.origin for app url.

weerd:
### What does this PR do?

This PR provides an update to remove the use of an old env variable for `PHOENIX_URL`, which then we used to set a JS `PHOENIX_URL` variable pulled from `window.ENV` which used this value. 

We now set the JS `PHOENIX_URL` using `window.location.origin`, and update a few places so they're all using this same constant when needing the default app url.

We saw this caused a problem on QA because we were using the hardcoded `www-qa.dosomething.org` env variable on the `qa.dosomething.org` instance and this was giving us a CORS error when @ngjo was testing submitting a text submission action. Nice find @ngjo!

It also does a slight bit of cleanup 🛁 

🌵 

### Checklist

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -30,7 +30,6 @@ class AppServiceProvider extends ServiceProvider
                 'NORTHSTAR_URL' => config('services.northstar.url'),
                 'NPS_SURVEY_ENABLED' => config('services.timed_modals.nps_survey.enabled'),
                 'PHOENIX_LEGACY_URL' => config('services.phoenix-legacy.url'),
-                'PHOENIX_URL' => config('services.phoenix.url'),
                 'PUCK_URL' => config('services.analytics.puck_url'),
                 'SIXPACK_BASE_URL' => config('services.sixpack.url'),
                 'SIXPACK_COOKIE_PREFIX' => config('services.sixpack.prefix'),
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -272,7 +272,7 @@ function get_metadata($entry)
         $entry = $entry->fields;
     }
 
-    $baseUrl = config('services.phoenix.url').'/us';
+    $baseUrl = config('app.url').'/us';
 
     $image = data_get($entry, 'metadata.fields.image', null);
 
--- a/config/services.php
+++ b/config/services.php
@@ -45,10 +45,6 @@ return [
         ],
     ],
 
-    'phoenix' => [
-        'url' => env('APP_URL'),
-    ],
-
     'phoenix-legacy' => [
         'url' => env('PHOENIX_LEGACY_URL', 'https://staging.dosomething.org'),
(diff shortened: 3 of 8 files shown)


weerd:
😊

app/Providers/AppServiceProvider.php
```diff
@@ -30,7 +30,6 @@ public function boot(Router $router)
                 'NORTHSTAR_URL' => config('services.northstar.url'),
                 'NPS_SURVEY_ENABLED' => config('services.timed_modals.nps_survey.enabled'),
                 'PHOENIX_LEGACY_URL' => config('services.phoenix-legacy.url'),
-                'PHOENIX_URL' => config('services.phoenix.url'),
```
mendelB: I think we still make use of this on the backend?
e.g. https://github.com/DoSomething/phoenix-next/blob/master/app/helpers.php#L275

mendelB: (I think that's the only place actually)

weerd: Oh I can update that to switch to use `app.url` which is the ideal config to be using.

mendelB: Oh sweet Where does the `app.url` get defined btw? I can't seem to find the source.

weerd: `.env` 💃

mendelB: Ahh see it, thanks! I was looking for a `services.app.url` 🤦‍♂️

weerd:
@DFurnes @mendelB I had to change things up slightly. Upon further testing I realized I can't have the `PHOENIX_URL` inside the `join()` method because it was changing `http://phoenix.test` to `http:/phoenix.test` and breaking the urls... it's a very subtle but significant different. Resulting urls were then `http://phoenix.test/phoenix.test/api/v2/...etc`.

Lemme know if the approach I resolved it with seems good to you.

Speaaaaaking of resolving... I also looked into the `path.resolve()` method which via StackOverflow sounded like it would solve the issue, but alas it's only really available on Node and not the browser since it is actually resolving from a node app root and I don't think meant for resolving web URLs.

app/Providers/AppServiceProvider.php
weerd: Oh and technically, in the `config/services.php` file the `phoenix` item in the array I changed a while back to use `APP_URL` so I think this would have still been fine, but best to just removed and avoid confusion!

weerd: And since we don't need it, just updated to remove it anyhoo 💃

mendelB: Ah yes, I confused this file with the `config/services.php` and mistakenly thought you were deleting the configuration of `services.phoenix.url` to begin with!

mendelB: I'm staring at Javascript for far too long 🚑

PR #1199, opened 2018-12-06
Remove PuckConnecter event tracking

mendelB:
### What does this PR do?

🏒 This PR removes all vestiges of the `PuckConnecter`s which were wrapping components to enable implicitly passing a `trackEvent` prop, and instead uses the `trackPuckEvent` analytics helper method.

We favor the more simple and digestible approach of using a plain ole helper function for explicit event tracking. This is also in preparation for some work we're going to be doing to the event tracking in general. Stay tuned!

### DIFF
--- a/resources/assets/actions/quiz.js
+++ b/resources/assets/actions/quiz.js
@@ -3,7 +3,7 @@
 import { find, get } from 'lodash';
 
 import {
-  clickedSignUpAction,
+  clickedSignupAction,
   PICK_QUIZ_ANSWER,
   COMPARE_QUIZ_ANSWER,
   VIEW_QUIZ_RESULT,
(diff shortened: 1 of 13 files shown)


resources/assets/actions/quiz.js
```diff
@@ -33,7 +33,7 @@ export function quizConvert(quizId, resultActionId) {
     const campaignId = getState().campaign.legacyCampaignId;
 
     dispatch(
-      clickedSignUpAction(campaignId, {
+      clickedSignupAction(campaignId, {
```
weerd: what got changed in this file?

mendelB: It's the upper case U in sign up - was `signUp` should be `signup`. Sneaky!

weerd: Oh damn! Yeah, I messed that up. I had it come up as a bug and couldn't figure it out for a while in another file until I eventually realized the typo. Must've missed one :( good catch!

weerd: We should have like a custom linter rule that flags any cases where the `U` in Signup is capitalized, lol 😂

mendelB: Haha that'd be nice!

PR #1202, opened 2018-12-07
Allowing more than two entries in affiliate credits

weerd:
### What does this PR do?

This PR updates the `AffiliateCredits` component to allow for more than a total of two items. Before this update, the credits would only show a single sponsor or partner, or show a sponsor and a partner, but no more than one of each total. So if there is a sponsor and 3 partners, it would only show the sponsor and the first partner in the list.

This also adds some logic to help separate each affiliate with a comma and if it is the last affiliate will prepend "and". It also ensures that if there are many affiliates which leads to a carriage return of the copy, the break doesn't happen in the middle of the affiliate's name, but breaks the whole name to the next line so it reads well.

It also does a bit of cleanup and moves the component to the utilities directory and removes the old approach of having an index file in the component directory.

Here's an example, with 4 affiliate credits.
image

But also looks fine with just one affiliate:
image

Refs [Pivotal ID #161694827](https://www.pivotaltracker.com/story/show/161694827)

### Checklist


resources/assets/scss/base.scss
```diff
@@ -392,3 +392,12 @@ p {
     font-weight: 400;
   }
 }
+
+// Tailwind-esk class names
+.w-1\/2 {
```
DFurnes: Woah. This is trippy.

resources/assets/components/utilities/AffiliateCredits/AffiliateCredits.js
```diff
@@ -0,0 +1,68 @@
...
+
+  return affiliates.length ? (
+    <div className="w-1/2 float-left">
+      In partnership with {affiliatesString}
+    </div>
+  ) : null;
+};
+
+AffiliateCredits.propTypes = {
+  affiliateSponsors: PropTypes.arrayOf(PropTypes.object).isRequired,
```
DFurnes: Do we want these to be optional and default to `[]` if not provided?

resources/assets/scss/base.scss
weerd: Yeah! Something I saw tailwind doing, but makes it nice to use fractions in classnames 💃

resources/assets/components/utilities/AffiliateCredits/AffiliateCredits.js
weerd: oh yeah, I dig that. I'll make that change and then mergy merge. For both `affiliateSponsors` and `affiliatePartners` right?

DFurnes: Yeah, that makes sense to me! 👍

PR #1205, opened 2018-12-11
Updated Analytics Tracking Method

mendelB:
### What does this PR do?

This PR adds an updated analytics tracking method (`trackAnalyticsEventBeta`) which provides a new event naming structure - an object consisting of `verb`, `noun` and the options `adjective` fields. As well as a new service dispatching methodology - it will dispatch the event to all services by default, and only to specific service if specified.

The method quickly grew to be pretty large, so I moved some of the logic into two helpers:
- `formatEventName` to generate a full snake cased event name from the `verb`, `noun`, and `adjective`
- `sendToServices` which dispatches the event to all services or the specified one

I also set up an `APP_PREFIX` constant to store the app prefix we use for event names.

### Checklist

### DIFF
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -141,7 +141,7 @@ export function trackAnalyticsEvent(name, data, service) {
  * Track an analytics event with a specified service. (Defaults to tracking with all services.)
  *
  * @param  {Object}      nameParams
- * @param  {Object}      data
+ * @param  {Object|Null} data
  * @param  {String|Null} service
  * @return {Object}
  */
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -15,6 +15,22 @@ import { get as getHistory } from '../history';
 const APP_NAME = 'phoenix';
 
 /**
+ * Parse analytics event name parameters into a snake cased string.
+ *
+ * @param  {String} verb
+ * @param  {String} noun
+ * @param  {String|Null} adjective
+ * @return {void}
+ */
+const parseEventName = (verb, noun, adjective) => {
+  let eventName = `${APP_NAME}_${verb}_${noun}`;
+  // Append adjective if defined.
+  eventName += adjective ? `_${adjective}` : '';
+
+  return eventName;
+};
+
+/**
  * Send event to analyze with Google Analytics.
  *
  * @param  {String} category
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -31,31 +31,6 @@ const parseEventName = (verb, noun, adjective) => {
 };
 
 /**
- * Dispatch analytics event to specified service, or all services by default.
- *
- * @param  {String}      category
- * @param  {String}      name
- * @param  {Object}      data
- * @param  {Object|Null} service
- * @return {void}
- */
-const dispatchToServices = (category, name, data, service) => {
-  switch (service) {
-    case 'ga':
-      analyzeWithGoogleAnalytics(category, name);
-      break;
-
-    case 'puck':
-      analyzeWithPuck(name, data);
-      break;
-
-    default:
-      analyzeWithGoogleAnalytics(category, name);
-      analyzeWithPuck(name, data);
-  }
-};
-
-/**
  * Send event to analyze with Google Analytics.
  *
  * @param  {String} category
@@ -94,6 +69,31 @@ export function analyzeWithPuck(name, data) {
 
   Puck.trackEvent(name, data);
 }
+
+/**
+ * Dispatch analytics event to specified service, or all services by default.
+ *
+ * @param  {String}      category
+ * @param  {String}      name
+ * @param  {Object}      data
+ * @param  {Object|Null} service
+ * @return {void}
+ */
+const dispatchToServices = (category, name, data, service) => {
+  switch (service) {
+    case 'ga':
+      analyzeWithGoogleAnalytics(category, name);
+      break;
+
+    case 'puck':
+      analyzeWithPuck(name, data);
+      break;
+
+    default:
+      analyzeWithGoogleAnalytics(category, name);
+      analyzeWithPuck(name, data);
+  }
+};
 
 /**
  * Watch the given parameters for changes in their state
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -113,6 +113,44 @@ export function trackAnalyticsEvent(name, data, service) {
 }
 
 /**
+ * Track an analytics event with a specified service. (Defaults to tracking with all services.)
+ *
+ * @param  {Object}      verb, noun, adjective
+ * @param  {Object}      data
+ * @param  {String|Null} service
+ * @return {Object}
+ */
+export function trackAnalyticsEventBeta(
+  { verb, noun, adjective },
+  data,
+  service,
+) {
+  if (!verb || !noun) {
+    console.error('The Verb or Noun is missing!');
+    return;
+  }
+
+  const eventName = parseEventName(verb, noun, adjective);
+
+  // Define category parameter for Google Analytics.
+  const category = `${APP_NAME}_${noun}`;
+
+  switch (service) {
+    case 'ga':
+      analyzeWithGoogleAnalytics(category, eventName);
+      break;
+
+    case 'puck':
+      analyzeWithPuck(eventName, data);
+      break;
+
+    default:
+      analyzeWithGoogleAnalytics(category, eventName);
+      analyzeWithPuck(eventName, data);
+  }
+}
+
+/**
  * Track an analytics event with Google Analytics.
  *
  * @param  {String} name
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -146,7 +146,7 @@ export function trackAnalyticsEvent(name, data, service) {
  * @return {Object}
  */
 export function trackAnalyticsEventBeta(nameParams, data, service) {
-  const { verb, noun, adjective } = nameParams;
+  const { verb, noun, adjective } = nameParams || {};
 
   if (!verb || !noun) {
     console.error('The Verb or Noun is missing!');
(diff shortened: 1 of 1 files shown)


resources/assets/helpers/analytics.js
```diff
@@ -11,6 +11,25 @@ import {
 import { PUCK_URL } from '../constants';
 import { get as getHistory } from '../history';
 
+// App name prefix used for event naming.
+const APP_NAME = 'phoenix';
```
weerd: Maybe rename to `APP_PREFIX`? Just don't want to cause confusion w/ the .env item `APP_NAME` which is set to `DoSomething`.

resources/assets/helpers/analytics.js
```diff
@@ -11,6 +11,25 @@ import {
...
+
+/**
+ * Parse analytics event name parameters into a snake cased string.
+ *
+ * @param  {String}      verb
+ * @param  {String}      noun
+ * @param  {String|Null} adjective
+ * @return {void}
+ */
+const parseEventName = (verb, noun, adjective) => {
```
weerd: should `adjective` default to `null? `(verb, noun, adjective = null)`

mendelB: Sure, is that better than the default `undefined`? I guess more explicit in showing the optional nature of the argument?

weerd: Yeah, just makes it clearer. And it's not something that relies on it being undefined for it to fill in a default like in react components :)

resources/assets/helpers/analytics.js
```diff
@@ -93,6 +137,30 @@ export function trackAnalyticsEvent(name, data, service) {
...
 
+/**
+ * Track an analytics event with a specified service. (Defaults to tracking with all services.)
+ *
+ * @param  {Object}      nameParams
+ * @param  {Object}      data
+ * @param  {String|Null} service
+ * @return {Object}
+ */
+export function trackAnalyticsEventBeta(nameParams, data, service) {
```
weerd: I wonder if this should be `trackAnalyticsEventBeta(data, nameParams={}, service=null)` 🤔 

This way, when pulling the variables out of the object, you don't get a breaking error if `nameParams` isn't an object.

resources/assets/helpers/analytics.js
```diff
@@ -51,6 +70,31 @@ export function analyzeWithPuck(name, data) {
   Puck.trackEvent(name, data);
 }
 
+/**
+ * Dispatch analytics event to specified service, or all services by default.
+ *
+ * @param  {String}      category
+ * @param  {String}      name
+ * @param  {Object}      data
+ * @param  {Object|Null} service
```
weerd: {String|Null} for `service`

resources/assets/helpers/analytics.js
mendelB: Ah, that makes sense! Though there are several instances where we don't pass a `data` argument at all to the tracker, this would force us to explicitly pass `null` in those scenarios

mendelB: What if we added a safeguard like so:
`const { verb, noun, adjective } = nameParams || {};`

(granted it's less elegant unfortunately...)

resources/assets/helpers/analytics.js
```diff
@@ -93,6 +137,30 @@ export function trackAnalyticsEvent(name, data, service) {
...
+    console.error('The Verb or Noun is missing!');
+    return;
+  }
+
+  const eventName = parseEventName(verb, noun, adjective);
+
+  // Define category parameter for Google Analytics.
+  const category = `${APP_PREFIX}_${noun}`;
+
+  dispatchToServices(category, eventName, data, service);
```
weerd: I wonder if we should use a different verb than `dispatch...` for this method name, since there's a bit of baggage related to the term and all the reduxing we do 🤔

mendelB: Hmm that's a great point. Perhaps `transmit` or `send`?

resources/assets/helpers/analytics.js
weerd: That could work. If you tested it and it works fine if `nameParams` isn't an object then I'm ok with that :)

resources/assets/helpers/analytics.js
weerd: `send` works for me :)

resources/assets/helpers/analytics.js
mendelB: Ok sweet!

mendelB: It did just occur to me that we'd have to explicitely pass `null` for `data` when specifying a `service` argument, but that generally should not be the case.

weerd: Yeah... a thought had passed me that we could change this to just pass a single object, like:

```js
{
  verb: 'someverb',
  noun: 'somenoun'
  adjective: 'someadjective',
  data: {a bunch of data},
  service: 'ga',
}
```
and then just use lodash `get()` to set defaults... 🤔

weerd: and then it becomes this: `trackAnalyticsEventBeta(options={})` and defaults to an object

mendelB: Hmm, that's a compelling thought -- I quite like this! Would this be the correct doc block approach for defining the desired fields? http://usejsdoc.org/tags-param.html#parameters-with-properties

mendelB: (also looks like, according to JSDoc, we're supposed to be lower casing our `String` definitions, not sure if we care about that?)

weerd: yep! and 😬 (we can do a later update to address capitalized string.... I wonder though since in JS `String` is an object... 🤔

resources/assets/helpers/analytics.js
```diff
@@ -11,6 +11,25 @@ import {
...
+
+/**
+ * Parse analytics event name parameters into a snake cased string.
+ *
+ * @param  {String}      verb
+ * @param  {String}      noun
+ * @param  {String|Null} adjective
+ * @return {void}
+ */
+const parseEventName = (verb, noun, adjective = null) => {
```
DFurnes: Would it make sense to name this `formatEventName` instead of `parseEventName`?

resources/assets/helpers/analytics.js
mendelB: It actually looks like they don't care! https://github.com/jsdoc3/jsdoc/issues/1046#issuecomment-126477791

(They mention their preferential distinction being primitives vs objects in that comment.)

resources/assets/helpers/analytics.js
mendelB: Oh sure, yeah, I think that's better!

resources/assets/helpers/analytics.js
mendelB: I used brackets to indicate the optional params for the `options` parameter as per http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values but we generally seem to use the `|` which seems to be meant for a param with multiple types http://usejsdoc.org/tags-param.html#multiple-types-and-repeatable-parameters. Do we want to be consistent with that, or are we fine with brackets?

mendelB: Not completely sure I'm correct about what I said about our general `Null` usage, but still confirming that I'm making the right move with the brackets!

resources/assets/helpers/analytics.js
```diff
@@ -93,6 +137,33 @@ export function trackAnalyticsEvent(name, data, service) {
...
 
+/**
+ * Track an analytics event with a specified service. (Defaults to tracking with all services.)
+ *
+ * @param  {Object} options
+ * @param  {String} options.verb
+ * @param  {String} options.noun
+ * @param  {String} [options.adjective]
+ * @param  {Object} [options.data]
+ * @param  {String} [options.service]
```
weerd: oh nice. I hadn't realized _optional_ props are placed in square brackets!

resources/assets/helpers/analytics.js
weerd: Let's be consistent!

resources/assets/helpers/analytics.js
```diff
@@ -93,6 +137,33 @@ export function trackAnalyticsEvent(name, data, service) {
...
+ * @return {void}
+ */
+export function trackAnalyticsEventBeta(options = {}) {
+  const { verb, noun, adjective, data, service } = options;
```
DFurnes: Totally subjective, but if you want you could move this destructuring into the function argument!

```js
export function trackAnalyticsEventBeta({ verb, noun, adjective, data, service }) {
```

weerd: Can you still have a default value though? For providing a default of `{}` it might read a little better if done in the function... 🤔

weerd: Actually, maybe that doesn't even matter anymore.

DFurnes: Yeah, you shouldn't need a default value, although you totally could if you wanted!

```js
export function trackAnalyticsEventBeta({ verb, noun, adjective, data, service } = {}) {
```

resources/assets/helpers/analytics.js
mendelB: Sorry, I may have confused this - but should I be using `|Null` for these optional props, or the brackets https://github.com/DoSomething/phoenix-next/pull/1205#discussion_r240763392?

mendelB: I also feel like it's sometimes unclear as to whether an argument is optional or if it's a type union and could be say an `Object` or `Null`, since these scenarios feel kind of interchagable to me... 

(maybe in instances where we *actually always expect* either something *or* `Null` for the function to function is the latter, and where we don't really care either way is the former? 🤔 )

mendelB:
Update: I added the destructured object as the options param (without the default, pls lemme know if you prefer adding that back in). I also added the use of `lodash#snakeCase` to ensure all the event name components are formatted properly.

resources/assets/helpers/analytics.js
weerd: Ok, I might be confused now.

Let's start again. Go with using the bracket syntax for optional params. _If_ a param can be multiple types then we do something like:

```js
@param  {Object|Null} [options.data]
```
How's that sound?

mendelB: Ah but of course, this actually makes perfect sense! I don't why I just decided it had to be either or.

PR #1206, opened 2018-12-13
Update All Analytics Events to Our New Format!

mendelB:
### What does this PR do?

This PR replaces the `trackAnalyticsEvent` method with the current Beta version and updates all of our puck tracking events to use the `trackAnalyticsEvent` method, with updated event naming based on our [established naming conventions](https://github.com/orgs/DoSomething/teams/tech-leads/discussions/3). 🥂 

We also remove some legacy analytics helpers specific to Puck and GA based on the older format.

This is a ton - but mostly very straightforward! 
I'll highlight anything I did which was beyond a simple event tracking update in the comments.

All events will be tracked to all services by default besides for the failed POST request event tracked in the [API middleware](https://github.com/DoSomething/phoenix-next/pull/1206/files#diff-252fe66a09d3de77eb4b779ca827a98dR103) since that seemed like it shouldn't be going to GA.

### DIFF
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -80,8 +80,8 @@ export function storeCampaignPost(campaignId, data) {
   const { action, body, id, type } = data;
 
   const analytics = {
-    name: `${type}-submission-action`,
-    service: 'puck',
+    verb: 'submitted',
+    noun: `${type}_submission_action`,
     payload: {
       action,
       campaignId,
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -119,8 +119,8 @@ export function getCampaignSignups(id = null, query = {}) {
 export function storeCampaignSignup(campaignId, data) {
   const path = join('api/v2/campaigns', campaignId, 'signups');
   const analytics = {
-    name: 'phoenix_clicked_signup',
-    service: 'puck',
+    verb: 'clicked',
+    noun: 'signup',
     payload: {
       campaignId,
     },
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -123,6 +123,7 @@ export function storeCampaignSignup(campaignId, data) {
     noun: 'signup',
     payload: {
       campaignId,
+      ...data.analytics,
     },
   };
 
--- a/resources/assets/components/LegacyQuiz/LegacyQuiz.js
+++ b/resources/assets/components/LegacyQuiz/LegacyQuiz.js
@@ -7,7 +7,7 @@ import Conclusion from './Conclusion';
 import Share from '../utilities/Share/Share';
 import ContentfulEntry from '../ContentfulEntry';
 import Markdown from '../utilities/Markdown/Markdown';
-import { trackAnalyticsEventBeta } from '../../helpers/analytics';
+import { trackAnalyticsEvent } from '../../helpers/analytics';
 
 import './legacy-quiz.scss';
 
@@ -64,7 +64,7 @@ const LegacyQuiz = props => {
   };
 
   if (shouldSeeResult) {
-    trackAnalyticsEventBeta({
+    trackAnalyticsEvent({
       verb: 'submitted',
       noun: 'quiz',
       data: {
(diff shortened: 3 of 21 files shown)


resources/assets/components/LegacyQuiz/LegacyQuiz.js
```diff
@@ -7,7 +7,7 @@ import Conclusion from './Conclusion';
 import Share from '../utilities/Share/Share';
 import ContentfulEntry from '../ContentfulEntry';
 import Markdown from '../utilities/Markdown/Markdown';
-import { trackPuckEvent } from '../../helpers/analytics';
+import { trackAnalyticsEventBeta } from '../../helpers/analytics';
```
weerd: Since this is going to switch out all the naming anyhoo, should we just use `trackAnalyticsEvent` and just rename `trackAnalyticsEventBeta` -> `trackAnalyticsEvent` and delete the old `trackAnalyticsEvent()` method?

resources/assets/helpers/index.js
```diff
@@ -672,7 +672,14 @@ export function showTwitterSharePrompt(href, quote = '', callback) {
...
  */
 export function handleFacebookShareClick(href, trackingData) {
-  trackPuckEvent('clicked facebook share', trackingData);
+  trackAnalyticsEventBeta({
+    verb: 'clicked',
+    noun: 'share',
+    adjective: 'facebook',
+    data: trackingData,
+  });
+  // @todo 12/13/2018: Use the showFacebookShareDialog to track
```
mendelB: I would have loved to knock this out in this PR, but it seems like it might get minorly complicated, so leaving this as a TODO

resources/assets/components/LegacyQuiz/LegacyQuiz.js
mendelB: I was going to leave that for a follow up PR, but happy to do so in this one if you prefer?

weerd: Changing all the names of events depends on using the new method, so do it in this one. Then you don't need another PR to rechange all the method names later.

mendelB: Updated! (Thank the gods for search and replace 😅)

resources/assets/actions/post.js
```diff
@@ -80,8 +80,8 @@ export function storeCampaignPost(campaignId, data) {
   const { action, body, id, type } = data;
 
   const analytics = {
-    name: `${type}-submission-action`,
-    service: 'puck',
+    verb: 'submitted',
+    noun: `${type}_submission_action`,
     payload: {
```
weerd: Should this property be renamed to `data` here, or doesn't matter?

resources/assets/components/SignupButton/SignupButton.js
```diff
@@ -26,11 +26,15 @@ const SignupButton = props => {
...
-      template,
-      legacyCampaignId: campaignId, // @TODO: confirm it's ok to send as campaignID and not legacyCampaignId
-      source,
-      sourceData: { text: buttonText },
+    trackAnalyticsEvent({
+      verb: 'clicked',
+      noun: 'signup',
+      data: {
+        template,
+        legacyCampaignId: campaignId, // @TODO: confirm it's ok to send as campaignID and not legacyCampaignId
```
weerd: should we also just address the `@TODO`? haha

resources/assets/actions/post.js
mendelB: It won't affect functionality per se, but do you think it'd be more clear as `data`?

resources/assets/actions/signup.js
```diff
@@ -133,7 +126,7 @@ export function storeCampaignSignup(campaignId, data) {
         failure: STORE_CAMPAIGN_SIGNUPS_FAILED,
         meta: {
           id: campaignId,
-          analytics,
+          analytics: data.analytics,
```
mendelB: @weerd Was this the right call? I moved the analytics to be added from wherever the action is dispatched, so we can add all the data previously fed to the manually invoked tracking event in `SignupButton`. (We could also keep this as was, and just allow a passed in `data.payload` argument and add that to the `analytics.payload` but I thought it might be nice to have it all declared in one place)

resources/assets/actions/post.js
weerd: Nah it's cool. Just wanted to make sure it wasn't an issue.

resources/assets/actions/signup.js
weerd: Meh, I think I prefer it how it was before. If the SignupButton needs to add any additional analytics data then it can and we can just append it.

mendelB: Aight cool, I'll update!

PR #1208, opened 2018-12-13
Explore campaigns pagination

weerd:
### What does this PR do?

This PR implements _very basic_ pagination for the Explore Campaigns page at `/us/campaigns`. 

To implement pagination properly, I had to circumvent the original method `getCampaignsSorted()`, which would gather ALL campaigns and sort them to remove closed campaigns, as well as sort the open campaigns by `staffPick` up top. As we continue adding more and more campaigns from Ashes into Phoenix, the complete list of campaigns will grow significantly and it will no longer be performant to gather ALL campaigns in one shot.

explore campaigns pagination

To allow for pagination and collecting campaigns by `36` at a time, we had to make a few concessions that we can address a bit better in follow up PRs. This solution is just v1!

- Both open and closed campaigns are displayed (will be addressed in follow up PR)
- List of campaigns is not preferentially sorted to display "Staff Pick" campaigns at the top (may have a potential solution for this in follow up PR).
- Will be increasing cache duration after adding a cache clearing option in upcoming PR.

Refs [Pivotal ID #162651211](https://www.pivotaltracker.com/story/show/162651211)

### Checklist

### DIFF
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -43,13 +43,37 @@ class CampaignRepository
      *
      * @return array
      */
-    public function getAll()
+    public function getAll($limit = null, $step = null)
     {
-        $campaigns = remember('campaigns', 15, function () {
-            return $this->getEntriesAsJson('campaign');
-        });
+        return json_decode($this->getEntriesAsJson('campaign', $limit, $step));
 
-        return json_decode($campaigns);
+        // $campaigns = remember('campaigns', 15, function () {
+        //     return $this->getEntriesAsJson('campaign');
+        // });
+
+        // return json_decode($campaigns);
+    }
+
+    /**
+     * Get all campaigns using pagination.
+     *
+     * @param  integer $count
+     * @param  integer $page
+     * @return \Illuminate\Support\Collection
+     */
+    public function getAllCampaignsPaginated($count = 24, $page = 1)
+    {
+        if (intval($page) <= 0) {
+            return collect();
+        }
+
+        $page = intval($page) - 1;
+
+        $skip = $page >= 1 ? $count * $page : null;
+
+        $campaigns = $this->getAll($count, $skip);
+
+        return collect($campaigns);
     }
 
     /**
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -72,10 +72,10 @@ class CampaignRepository
             return collect();
         }
 
-        // Calculate number to multiply count by, to get number to skip in collection query.
+        // Calculate number to multiply count by, to get number of items to skip in collection query.
         $multiplier = intval($page) - 1;
 
-        $skip = $multiplier >= 1 ? $count * $multiplier : null;
+        $skip = $count * $multiplier;
 
         $campaigns = $this->getAll($count, $skip);
 
--- a/app/Repositories/CampaignRepository.php
+++ b/app/Repositories/CampaignRepository.php
@@ -41,12 +41,16 @@ class CampaignRepository
     /**
      * Get all campaigns.
      *
+     * @param  int|null $limit
+     * @param  int|null $skip
      * @return array
      */
     public function getAll($limit = null, $skip = null)
     {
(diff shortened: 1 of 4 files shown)


app/Repositories/CampaignRepository.php
```diff
@@ -41,17 +41,47 @@ public function __construct(Contentful $contentful, PhoenixLegacy $phoenixLegacy
...
-    public function getAll()
+    public function getAll($limit = null, $skip = null)
     {
-        $campaigns = remember('campaigns', 15, function () {
-            return $this->getEntriesAsJson('campaign');
+        $cacheKey = 'campaigns';
+        $cacheKey = $limit ? $cacheKey.':limit='.$limit : $cacheKey;
+        $cacheKey = $skip ? $cacheKey.':skip='.$skip : $cacheKey;
+
+        $campaigns = remember($cacheKey, 15, function () use ($limit, $skip) {
```
mendelB: One concern with caching the pages individually - what happens if the pages desynchronize? For example, if a Campaign on page 3 is cached, but then is updated, so it moves to page one - it'll show up on both pages, and might cause another campaign which should shift to page 3 not to show up?

resources/views/campaigns/index.blade.php
```diff
@@ -24,6 +24,16 @@
                     </li>
                 @endforeach
             </ul>
+
+            <div class="pagination clearfix">
+                @if($previousPage > 0)
+                    <a class="button -tertiary float-left clear-none" href="?page={{$previousPage}}">Previous</a>
+                @endif
+
+                @if($campaigns->count() >= $count)
```
mendelB: super nitpicky - but am I understanding correctly that potentially, if the total campaigns are divisible by 36 - there'd be a Next Page button which would lead to a redirect back to `/us/campaigns` since the next paginated request would come back empty?

app/Repositories/CampaignRepository.php
```diff
@@ -41,17 +41,47 @@ public function __construct(Contentful $contentful, PhoenixLegacy $phoenixLegacy
...
+    {
+        // Return an empty collection if page is 0 or negative.
+        if (intval($page) <= 0) {
+            return collect();
+        }
+
+        // Calculate number to multiply count by, to get number to skip in collection query.
+        $multiplier = intval($page) - 1;
+
+        $skip = $multiplier >= 1 ? $count * $multiplier : null;
```
mendelB: A nitpick - but more curiosity -- is it a problem for us if the `$skip` is `0`  (which if not would allow us to avoid the ternary)?

app/Repositories/CampaignRepository.php
weerd: Yep! I think it's a small issue and eventually we should likely add a webhook that forces a cache clear on Phoenix when a campaign is updated for both the campaign show page and the index pages.

But also keep in mind this is for a bare bones v1 for now ;)

resources/views/campaigns/index.blade.php
weerd: Yep! Again this is bare-bones and basic and need to hit that exact scenario for that to be the case, which will ideally be rare and if it happens, it just goes back to page 1 as a failsafe.

Aiming to improve this over time as options get better!

app/Repositories/CampaignRepository.php
mendelB: Ahh integrating with Contentful Webhooks would be super neat!

resources/views/campaigns/index.blade.php
mendelB: Makes total sense, and this feels perfectly adequate! Just wanted to double check that we were ok with that.

app/Repositories/CampaignRepository.php
weerd: 🤔 hmm, yeah not sure. I think I may have required it not be zero in a prior approach but something I can test out and remove if it turns out it's fine to have `$skip = 0`...

weerd: Tested it! Looks like it's fine! Good call 💃 I'll update after lunch ✨

PR #1213, opened 2018-12-19
Push Analytics Events to Google Tag Manager

mendelB:
### What does this PR do?

This PR updates our analytics system to push Google Analytics events to the Google Tag Manager using it's `dataLayer`.

https://developers.google.com/tag-manager/devguide

This will -for any event tracked to Google Analytics- also push the event (using the 'action' as the event name) into the `dataLayer` array exposed by Google Tag Manager.

### DIFF
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -32,19 +32,6 @@ const formatEventName = (verb, noun, adjective = null) => {
 };
 
 /**
- * Analyze event with Google Tag Manager.
- * (pushes event into the window's Data Layer array.)
- *
- * @param  {String} event
- * @return {void}
- */
-export function analyzeWithGoogleTagManager(event) {
-  window.dataLayer = window.dataLayer || [];
-
-  window.dataLayer.push({ event });
-}
-
-/**
  * Send event to analyze with Google Analytics.
  *
  * @param  {String} category
@@ -63,7 +50,10 @@ export function analyzeWithGoogleAnalytics(category, action) {
   const identifier = `${category}:${action}:${label}`;
 
   analyze(identifier);
-  analyzeWithGoogleTagManager(action);
+
+  // Push event action to Google Tag Manager's data layer.
+  window.dataLayer = window.dataLayer || [];
+  window.dataLayer.push({ event: action });
 }
 
 /**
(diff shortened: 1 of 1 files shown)


resources/assets/helpers/analytics.js
```diff
@@ -31,6 +31,19 @@ const formatEventName = (verb, noun, adjective = null) => {
...
 };
 
+/**
+ * Analyze event with Google Tag Manager.
+ * (pushes event into the window's Data Layer array.)
+ *
+ * @param  {String} event
+ * @return {void}
+ */
+export function analyzeWithGoogleTagManager(event) {
```
mendelB: I wasn't quite sure about naming this since it's technically a bit different than the regular analyzeWith methods and we do call this from within the analyzeWithGoogleAnalytics but settled on this since it seemed consistent.

resources/assets/helpers/analytics.js
```diff
@@ -50,6 +63,7 @@ export function analyzeWithGoogleAnalytics(category, action) {
   const identifier = `${category}:${action}:${label}`;
 
   analyze(identifier);
+  analyzeWithGoogleTagManager(action);
```
mendelB: I figured since this is somewhat intrinsically tied to our GA work (not strictly speaking but functionally we're tracking these in conjunction with each other), so went with calling this from within this method directly. But def lmk if you have other thoughts

weerd: Yeah, I was debating this a bit too and kind of felt we're not doing enough for it to be a separate thing, and it's also tied to GA. Let's just do it in the `analyzeWithGoogleAnalytics()` method.

mendelB: Do you mean to get rid of the `analyzeWithGoogleTagManager ` method entirely and move the logic into this method?

weerd: Yeah I think so. There isn't much going on in that method and we can just add a comment saying `# Adding to data layer for Google Tag Manager` or whatever.

weerd: If we end up seeing more additions related to tag manager then we can split it up into a separate method, but for now this feels fine :)

mendelB: Updated!

Issue #1214, opened 2019-01-03
Social Share Action Type Field Bug


mendelB:
## _ISSUE OVERVIEW_

While digging through some Reportback validation error messages we noticed a bit of an issue introduced when updating the event naming for GA and Puck in #1206 :

https://git.io/fhYIG - changed the `type` field for Social Share RB submissions from `share-social` to `share_social`. This was to enable it to be formatted properly for the analytics event in the post request action https://git.io/fhYIl but without realizing that this would (obviously!) be changing the submitted `action` field which now fails validation in Rogue https://git.io/fhYIw

Hence, a ton of `type` validation failures since this change was introduced!

We're going to update the `type` to the correct format in an upcoming PR, and ensure the proper snake casing with a string replace in the Post request action analytics logic. 

And of course, any Social Shares which need to be backfilled into Rogue, will be adressed!

#### Additional Information
https://dosomething.slack.com/archives/C2BPA7M8F/p1546532053020600

mendelB:
We ran a backfill using exported Puck data to include the missing Social Share posts 🎉 
https://dosomething.slack.com/archives/CA0N5FXND/p1547046056003000

PR #1215, opened 2019-01-03
Social Share Action Type Field Fix

mendelB:
### What does this PR do?

This PR updates the `type` field used in the Social Share Action to the correct format.

- uses an imported constant from the new `post-types` file, with the updated (fixed) naming
- formats the analytics noun derived from the `type` field in the post action to use snake casing

This addresses an error introduced in #1206 where we mistakenly renamed the `type` field for the sake of the analytics formatting.

The constants file will introduce formality to the naming, and hopefully be less error-prone!

#1214

### DIFF
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -79,9 +79,11 @@ export function storeCampaignPost(campaignId, data) {
 
   const { action, body, id, type } = data;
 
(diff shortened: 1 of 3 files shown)


weerd:
> uses an imported constant from the new action-types

Updated description to post-types :)

PR #1219, opened 2019-01-07
Removing last bits of Ashes references 🔥 (part 2) 

weerd:
### What does this PR do?

This PR wraps up removing references to Ashes legacy IDs and legacy Run IDs, and updates code to only deal with numerical campaign IDs instead of mongo-like string IDs. It also removes the `LegacyCampaign` entity since we are no longer retrieving campaigns from Ashes and do not need this entity, as well as the `PhoenixLegacy` service class which was used for making requests to Ashes.

🔥 

Refs [Pivotal ID #162738678](https://www.pivotaltracker.com/story/show/162738678)

### Checklist


app/Http/Controllers/Api/CampaignsController.php
```diff
@@ -32,45 +32,33 @@ public function __construct(CampaignRepository $campaignRepository)
...
-        $idsArray = $ids ? explode(',', $ids) : [];
+        $ids = $idString ? explode(',', $idString) : [];
 
         // Resetting the filter[id] param value to an array so that we can properly validate.
-        $request->merge(['filter' => ['id' => $idsArray]]);
+        $request->merge(['filter' => ['id' => $ids]]);
 
         $this->validate($request, [
             'filter.id' => 'max:10',
+            // @TODO: potentially use a form request to do a bit more processing and
```
DFurnes: This can be done here with a `*` validation rule, I believe!

```php
'filter.id' => 'max:10',
'filter.id.*' => 'int',
```

weerd: ah perfect! Wasn't sure how to address the nested IDs, `*` wildcard makes sense! 👌 Will do a small PR update later today 💃 thanks @DFurnes

PR #1223, opened 2019-01-11
Updating middleware order of operations.

weerd:
### What does this PR do?

This PR updates and improves the middleware order of execution so that event data related middleware fire prior to authentication and other user related requests/executions.

Prior to this fix, we were intending to fire analytics events for when the signup button was clicked, but if the user was not logged in, the middleware to authenticate them would fire before the middleware to track the event, sending the user off to Northstar and never getting a chance to fire the click event!

### Checklist

### DIFF
--- a/resources/assets/store/middlewares.js
+++ b/resources/assets/store/middlewares.js
@@ -4,9 +4,14 @@ import sixpackExperimentsMiddleware from '../middleware/sixpackExperiments';
 import requiresAuthenticationMiddleware from '../middleware/requiresAuthentication';
 
 const middlewares = [
+  // Event data collection for analysis needs to execute first.
   analyticsMiddleware,
   sixpackExperimentsMiddleware,
+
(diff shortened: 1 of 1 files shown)


resources/assets/store/middlewares.js
```diff
@@ -4,10 +4,10 @@ import sixpackExperimentsMiddleware from '../middleware/sixpackExperiments';
 import requiresAuthenticationMiddleware from '../middleware/requiresAuthentication';
 
 const middlewares = [
-  requiresAuthenticationMiddleware,
   analyticsMiddleware,
-  apiMiddleware,
   sixpackExperimentsMiddleware,
+  requiresAuthenticationMiddleware,
```
DFurnes: It might be nice to add a comment above this line to explain why these are grouped at the bottom.

weerd: ya, that's a good idea!

PR #1224, opened 2019-01-11
Update getEntriesAsJson Contentful Query

### LINKED ISSUE #1222: Contentful Query Taking 4Ever!
### Current Behavior

What is the current behavior?
Something we've noticed about our Phoenix queries to Contentful (via the Contentful client and QueriesContentful trait in Laraval) is that it was taking an unusually long time for some pretty simple index queries.

@weerd raised this point specifically regarding the [CampaignController#index](https://github.com/DoSomething/phoenix-next/blob/master/app/Http/Controllers/CampaignController.php#L33) related query, specifically the [GetAllCampaignsPaginated query](https://github.com/DoSomething/phoenix-next/blob/master/app/Repositories/CampaignRepository.php#L58) which although the index is limiting the call to 36 entries, would still lag for a very long time, prompting [varnish cache timeouts via Fastly](https://www.pivotaltracker.com/story/show/162718264).

This could be nailed down more specifically to the [QueriesContentful#GetEntriesAsJson](https://github.com/DoSomething/phoenix-next/blob/master/app/Repositories/QueriesContentful.php#L22) method.

### Expected Behavior

Describe the behavior you are expecting.

We'd reasonably expect this query to be reasonably quick especially since it's being paginated in the campaigns index example. But we see it take -sometimes- a good 20-30 seconds to complete.

#### Additional Information

_Any other info or references to cards/tickets from other services (Trello, etc)._
Pivotal ID: [#162718264](https://www.pivotaltracker.com/story/show/162718264)

mendelB:
### What does this PR do?

This PR updates the `QueriesContentful#getEntriesAsJson` Contentful Query to be more efficient.

- utilize an `$options` array for all -optional- setting params
- include a `includeDepth` param to allow us to control the `setInclude` setting to heavilly optimize our querying, improving speed
- update the calls to this query from the `CampaignRepository#getAll`, and the `HomepageRepository#getFirst`

In our effort to solve the immense amount of time our campaigns index Contentful query was taking, and to preemptively understand if it was worth our while to pursue a basic site search feature - we ended up being able to easily optimize via the `setInclude` setting which will resolve Linked Entries up to the specified depth, and thus eager load the nested fields we were using in `TruncatedCampaign`, (as well as a bonus speed boost to our usage of `Homepage`)

**Q:** How do y'all feel about the docblock format for the `$options` array?

Refs 
[Pivotal ID #162718264](https://www.pivotaltracker.com/story/show/162718264) (Varnish cache issue)
[Pivotal ID #162972495](https://www.pivotaltracker.com/story/show/162972495) (Site search)
fixes #1222

### DIFF
--- a/app/Repositories/QueriesContentful.php
+++ b/app/Repositories/QueriesContentful.php
@@ -14,18 +14,20 @@ trait QueriesContentful
      * Get entries from a Contentful Query and return data as JSON.
      *
      * @param  string $type
-     * @param  int $limit
-     * @param  int $skip
+     * @param  array  $options
+     * @param  int    $options['includeDepth']
+     * @param  string $options['limit']
+     * @param  string $options['skip']
      * @return string
      */
-    protected function getEntriesAsJson($type, $limit = null, $skip = null)
+    protected function getEntriesAsJson($type, $options = [])
     {
         $query = (new Query)
                 ->setContentType($type)
-                ->setInclude(0)
+                ->setInclude(array_get($options, 'includeDepth', 0))
                 ->orderBy('sys.updatedAt', true)
-                ->setLimit($limit)
-                ->setSkip($skip);
+                ->setLimit(array_get($options, 'limit'))
+                ->setSkip(array_get($options, 'skip'));
 
         $entries = app('contentful.delivery')->getEntries($query)->getItems();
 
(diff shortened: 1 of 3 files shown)


app/Repositories/QueriesContentful.php
```diff
@@ -14,18 +14,20 @@
      * Get entries from a Contentful Query and return data as JSON.
      *
      * @param  string $type
-     * @param  int $limit
-     * @param  int $skip
+     * @param  array  $options
+     * @param  int    $options['includeDepth']
+     * @param  string $options['limit']
+     * @param  string $options['skip']
```
weerd: Yeah, I think I dig this just fine. No qualms from me!

DFurnes:
+1, nice sleuthing! 💯

PR #1225, opened 2019-01-11
Updates to differentiate signups found versus new completed signups.

weerd:
### What does this PR do?

This PR wraps up adding new analytics events that follow `phoenix_clicked_signup` so that we can differentiate if the user was signing in to a previously signed up campaign, or signing up for the campaign for the very first time.

It also adds an adjective to the clicked signup event IF the event was triggered AFTER a redirect to authenticate on identity.dosomething.org (Northstar). Even though the `phoenix_clicked_signup` event _will_ fire to log the event _before_ the redirection to Northstar, it fires again when returning from Northstar and dispatching the signup action again that it previously stored in localstorage. I found it might be helpful to differentiate between the duplicate events by specifying that the second was actually triggered post authentication to finalize the signup process on the selected campaign. That event is now named `phoenix_clicked_signup_triggered_post_auth`. Hopefully this is useful!

Additionally, we now have events after successful signup response from Rogue:
- `phoenix_completed_signup` for a `201` status response to a successfully created signup record
- `phoenix_found_signup` for a `200` status response to a found signup record

The above also apply for other successful post requests, and the API middleware uses the `payload.meta.analytics.noun` to determine what the post request type is and use that as the `noun`, defaulting to generic `post_request` if no noun is otherwise found. Examples:
- `phoenix_completed_photo_submission_action`
- `phoenix_completed_text_submission_action`

The middleware order sorting in #1223 helped pave the way for this PR 😄 I testing events on both authenticated and unauthenticated users and all seems to look good!

### Checklist

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -118,14 +118,6 @@ export function getCampaignSignups(id = null, query = {}) {
  */
 export function storeCampaignSignup(campaignId, data) {
   const path = join('api/v2/campaigns', campaignId, 'signups');
-  const analytics = {
-    verb: 'clicked',
-    noun: 'signup',
-    payload: {
-      campaignId,
-      ...data.analytics,
-    },
-  };
 
   return dispatch => {
     dispatch(
@@ -134,7 +126,6 @@ export function storeCampaignSignup(campaignId, data) {
         failure: STORE_CAMPAIGN_SIGNUPS_FAILED,
         meta: {
           id: campaignId,
-          analytics,
           sixpackExperiments: { conversion: 'signup' },
           messaging: {
             success: 'Thanks for signing up!',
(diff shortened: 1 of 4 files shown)


mendelB:
@weerd general question:

Am I understanding correctly that we are currently tracking two `phoenix_clicked_signup` events for unauthenticated users who successfully login?
If so, does that raise a question about the interdependence of the `storeCampaignSignup` action and clicking the signup button -- like I feel that it's somewhat strange that we're tracking a second clicked signup event when they haven't actually clicked it again. Would it make more sense, to decouple the clicked signup event from the subsequent action, and instead rely on the `completed_signup` or `found_signup` events which will now be triggered afterward? (I hope I'm making sense / understanding the flow correctly)

weerd:
Ok @mendelB! Updated and I definitely like your suggestion. Worked out better and feels a lot cleaner 👌

PR #1230, opened 2019-01-16
Basic Campaigns Search Functionality

mendelB:
### What does this PR do?

This PR adds basic campaigns search functionality. 🔍 

- new `/search` route supporting a `query` parameter
- `searchByFullText` method on the `CampaignRepository` to preform a Contentful [Full-Text](https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/search-parameters/full-text-search) search.
- new `SearchController` as a [Single Action Controller](https://laravel.com/docs/5.5/controllers#single-action-controllers) to handle the search requests

The format of the route, controller was heavily inspired by similar functionality in [Aurora](https://github.com/DoSomething/aurora/blob/master/app/Http/Controllers/UsersController.php#L106) and [Rogue](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Legacy/Web/UsersController.php#L59). (**Question**:) Though in those cases, since the searches are User-specific, the `search` controller action is in the `UserController` itself -- perhaps an argument for getting rid of the `SearchController` for now in favor of putting a `search` action in the [`CampaignController`](https://github.com/DoSomething/phoenix-next/blob/master/app/Http/Controllers/CampaignController.php)? (initially when making the controller I was still under the impression that the search would be across pages *and* campaigns)

Since we've opted to only perform searches on Campaigns, for now, I just added the search method to the CampaignRepository, instead of adding like a new SearchRepository - was this the correct move?

The controller is just returning the array for now, since implementing the view is going to take a bit of work so we can utilize some of the existing [`index.blade.php`](https://github.com/DoSomething/phoenix-next/blob/master/resources/views/campaigns/index.blade.php#L9-L28) logic

Refs [Pivotal ID #162972495](https://www.pivotaltracker.com/story/show/162972495)

### Checklist

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -39,8 +39,9 @@ $router->get('campaigns/{slug}/{clientRoute?}', function ($slug, $clientRoute = 
 
 // Search
 $router->get('us/search', 'SearchController');
-$router->get('search', function() {
+$router->get('search', function () {
     $queryString = request()->has('query') ? '?query='.request('query') : '';
+
     return redirect('/us/search'.$queryString);
 });
 
--- a/routes/web.php
+++ b/routes/web.php
@@ -37,6 +37,13 @@ $router->get('campaigns/{slug}/{clientRoute?}', function ($slug, $clientRoute = 
     return redirect('us/campaigns/'.$slug.'/'.$clientRoute);
 });
 
+// Search
+$router->get('us/search', 'SearchController');
+$router->get('search', function() {
+    $queryString = request()->has('query') ? '?query='.request('query') : '';
+    return redirect('/us/search'.$queryString);
+});
+
 // Categorized Pages (articles, facts)
 $categories = 'articles|facts|about';
 
--- a/routes/web.php
+++ b/routes/web.php
@@ -40,9 +40,7 @@ $router->get('campaigns/{slug}/{clientRoute?}', function ($slug, $clientRoute = 
 // Search
 $router->get('us/search', 'SearchController');
 $router->get('search', function () {
-    $queryString = request()->has('query') ? '?query='.request('query') : '';
-
-    return redirect('/us/search'.$queryString);
+    return redirect('/us'.request()->getRequestUri());
 });
 
 // Categorized Pages (articles, facts)
--- a/app/Http/Controllers/SearchController.php
+++ b/app/Http/Controllers/SearchController.php
@@ -25,7 +25,6 @@ class SearchController extends Controller
         $this->campaignRepository = $campaignRepository;
     }
 
-
     /**
      * Run a full-text search by given input. Display results.
      *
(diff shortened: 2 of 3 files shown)


DFurnes:
> (**Question**:) Though in those cases, since the searches are User-specific, the `search` controller action is in the `UserController` itself -- perhaps an argument for getting rid of the `SearchController` for now in favor of putting a `search` action in the [`CampaignController`](https://github.com/DoSomething/phoenix-next/blob/master/app/Http/Controllers/CampaignController.php)? (initially when making the controller I was still under the impression that the search would be across pages *and* campaigns)

I'm actually way more in-favor of having this in a separate controller, like you did (and limiting controllers to RESTful methods like `create`, `update`, `store`, etc). While a `search` method alone is probably fine, I think it makes it easy to start throwing the kitchen sink into a controller before you know it. 🚰

mendelB:
> I'm actually way more in-favor of having this in a separate controller, like you did (and limiting controllers to RESTful methods like `create`, `update`, `store`, etc). While a `search` method alone is probably fine, I think it makes it easy to start throwing the kitchen sink into a controller before you know it. 🚰

Ahh, would this be a good argument for a separate `SearchRepository` as well? (Or can that deviate from RESTfullness more flexibely?)

DFurnes:
Nah, I think it makes sense for repositories, when we use them, to be entity-specific (so all the `Campaign` actions in one place, all the `User` actions in another). (The same way that you wouldn't have a separate `UserSearch` model for searching users… you'd just query the existing `User` model a bit differently.)

routes/web.php
```diff
@@ -37,6 +37,14 @@
     return redirect('us/campaigns/'.$slug.'/'.$clientRoute);
 });
 
+// Search
+$router->get('us/search', 'SearchController');
+$router->get('search', function () {
+    $queryString = request()->has('query') ? '?query='.request('query') : '';
+
+    return redirect('/us/search'.$queryString);
```
weerd: Just curious, does doing `redirect('/us/search?query=gobbledygook')` not work? Not sure why the query is separated only to be re-added when hitting `/search` and redirecting to `/us/search` but probs something I'm missing? 🤔

mendelB: @weerd Updated this with -I think- a way better redirect solution! (I don't know why I wasn't able to find this method anywhere when initially building this out - hence the prior clumsiness)

routes/web.php
```diff
@@ -37,6 +37,12 @@
     return redirect('us/campaigns/'.$slug.'/'.$clientRoute);
 });
 
+// Search
+$router->get('us/search', 'SearchController');
+$router->get('search', function () {
+    return redirect('/us'.request()->getRequestUri());
```
weerd: 👏

PR #1231, opened 2019-01-18
Analytics events bug fix

weerd:
### What does this PR do?

This PR does a few updates to fix an issue with how `clicked_signup` events were being reported, along with `found_signup` and `completed_signup` and improves the analytics triggering for both the `signup` and `post` actions.

#1225 removed the `analytics` object from the action dispatched in `storeCampaignSignup()` for an alternate approach. However, this caused a bug because in the API middleware `postRequest()` the `noun` for the event that would be triggered in this method was looking for it in `meta.analyitcs.noun` which no longer existed, and thus defaulting to a generic `noun` of `post_request`. This was confusing in the GA results.

Screenshot of the correct, expected events in Google Analytics (local dev account):
image

### Checklist

### DIFF
--- a/resources/assets/middleware/api.js
+++ b/resources/assets/middleware/api.js
@@ -8,8 +8,8 @@ import { PHOENIX_URL } from '../constants';
 import { setFormData } from '../helpers/forms';
 import { API } from '../constants/action-types';
 import { getUserToken } from '../selectors/user';
-import { trackAnalyticsEvent } from '../helpers/analytics';
 import { getRequest, setRequestHeaders, tabularLog } from '../helpers/api';
+import { formatEventNoun, trackAnalyticsEvent } from '../helpers/analytics';
 
 /**
  * Send a GET request and dispatch actions.
@@ -61,9 +61,12 @@ const postRequest = (payload, dispatch, getState) => {
 
   const body =
     payload.body instanceof FormData ? payload.body : setFormData(payload.body);
+  const campaignContentfulId = get(payload, 'meta.id', 'post_request');
+  const campaignId = get(payload, 'meta.campaignId');
+  const postType = get(payload, 'meta.type', 'post_request');
 
   dispatch({
-    id: payload.meta.id,
+    id: payload.meta.id, // @TODO: rename to campaignContentfulId or campaignId
     type: payload.pending,
   });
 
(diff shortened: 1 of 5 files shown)


resources/assets/middleware/api.js
```diff
@@ -104,12 +110,13 @@ const postRequest = (payload, dispatch, getState) => {
 
       trackAnalyticsEvent({
         verb: 'failed',
-        noun: 'post_request',
+        noun: formatEventNoun(postType),
         data: {
-          url: payload.url,
```
mendelB: Would the URL of the request not be useful data?

resources/assets/middleware/api.js
```diff
@@ -104,12 +110,13 @@ const postRequest = (payload, dispatch, getState) => {
...
+          body,
```
DFurnes: Do we think `body` might be too large here to embed in individual analytics events?

weerd:
> TBH these action / middleware flows make my head spin sometimes 😵

Part of this eliminates calling these events via the middleware flow to make it a bit easier. Could be a middleware we end up removing if we no longer see a need for it 😄

resources/assets/middleware/api.js
```diff
@@ -86,10 +89,13 @@ const postRequest = (payload, dispatch, getState) => {
...
 
       trackAnalyticsEvent({
-        verb: statusCode === 200 ? 'found' : 'completed',
-        noun: get(payload, 'meta.analytics.noun', 'post_request'),
+        verb:
+          statusCode === 200 && postType === 'signup' ? 'found' : 'completed',
+        noun: formatEventNoun(postType),
         data: {
-          campaignId: payload.meta.id,
+          activityId: response.data.id,
```
DFurnes: What's `activityId` mean? The signup or post ID?

resources/assets/middleware/api.js
weerd: Only `signup` provided the URL in the payload and all it was is the same thing for those posts... a URL to `/api/v2/campaigns` haha. So not really that useful.

resources/assets/middleware/api.js
```diff
@@ -61,9 +61,12 @@ const postRequest = (payload, dispatch, getState) => {
 
   const body =
     payload.body instanceof FormData ? payload.body : setFormData(payload.body);
+  const campaignContentfulId = get(payload, 'meta.id', 'post_request');
```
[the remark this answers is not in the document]
weerd: oh b/c of silly copy-pasta! defs a typo 🍝 Good catch!

resources/assets/middleware/api.js
DFurnes: Got it! I was just confused at first what an "activity" ID was. I think it's fine as is.

resources/assets/middleware/api.js
DFurnes: Does `body` include file uploads? If so, I'd exclude here. Otherwise it probably doesn't matter either way.

weerd: It would for failed RB Post requests which was the only worry. Ok, I can exclude 🗡

PR #1232, opened 2019-01-23
Update Share Action id prop

mendelB:
### What does this PR do?

This PR updates the `id` prop passed from the `ShareAction` to the `storeCampaignPost` function with the new `campaignContentfulId` prop.

The `id` prop was the Share Action Contentful ID as opposed to the intended Campaign Contentful ID

https://dosomething.slack.com/archives/C3ASB4204/p1548261761165300

### DIFF
--- a/resources/assets/components/actions/ShareAction/ShareAction.js
+++ b/resources/assets/components/actions/ShareAction/ShareAction.js
@@ -31,12 +31,12 @@ class ShareAction extends React.Component {
   storeSharePost = puckId => {
     const action = get(this.props.additionalContent, 'action', 'default');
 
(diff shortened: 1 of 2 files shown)


weerd:
> The id prop was the Share Action Contentful ID as opposed to the intended Campaign Contentful ID

Do we want to start making it the `campaignId` moving forward though?

weerd:
Errr, wait, for posting we're using the correct `campaignId`. I think I'm just confusing myself.

mendelB:
Right, I think it's just for this field https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/actions/post.js#L100

PR #1239, opened 2019-01-29
Fix for bug in graphql query to show campaigns posts on profile page.

weerd:
### What does this PR do?

This PR fixes a bug in the `UserPostsQuery` component that is meant to show a user's posts in the Profile > Campaigns page. It was missing references to the components to use in the query

🐞  Before:
screen shot 2019-01-29 at 5 23 09 pm

🥾  After:
screen shot 2019-01-29 at 5 23 27 pm

Refs [Pivotal ID #163561914](https://www.pivotaltracker.com/story/show/163561914)

### DIFF
--- a/resources/assets/components/pages/AccountPage/UserPostsQuery.js
+++ b/resources/assets/components/pages/AccountPage/UserPostsQuery.js
@@ -21,6 +21,8 @@ const USER_POSTS_QUERY = gql`
       user {
         firstName
       }
+      ...PostCard
+      ...ReactionButton
     }
   }
 
--- a/resources/assets/components/pages/AccountPage/UserPostsQuery.js
+++ b/resources/assets/components/pages/AccountPage/UserPostsQuery.js
@@ -18,9 +18,6 @@ const USER_POSTS_QUERY = gql`
       text
       tags
       quantity
-      user {
-        firstName
-      }
       ...PostCard
       ...ReactionButton
     }


resources/assets/components/pages/AccountPage/UserPostsQuery.js
```diff
@@ -21,6 +21,8 @@ const USER_POSTS_QUERY = gql`
       user {
```
DFurnes: Oh, I think you should be safe to remove these fields since you have the fragments below.

DFurnes: Yup – if fields are included in the fragment, you don't need to repeat them here.

DFurnes: @weerd Sorry, just to clarify – I think you can remove _all_ these fields, not just `user { firstName }` since they're [included in the `PostCard` fragment](https://github.com/DoSomething/phoenix-next/blob/bc635bab2cd84730dbb734731e4fc44c7f68bd50/resources/assets/components/utilities/PostCard/PostCard.js#L15-L26).

DFurnes: So the final query could be simplified to:

```gql
query UserPostsQuery($userId: String!, $count: Int, $page: Int) {
  postsByUserId(id: $userId, count: $count, page: $page) {
    ...PostCard
    ...ReactionButton
  }
}
```

weerd: oh right right. ok. I'll toss up the update in another PR since I already merged this. For future reference though, does this technically hurt the query, or does GraphQL realize there's duplicate fields and just ignore them?

DFurnes: Nope, I don't believe this hurts performance. Just readability/maintainability.

PR #1243, opened 2019-02-01
Adding Signup Affirmation Toggle Option

mendelB:
### What does this PR do?

This PR adds logic to allow optionally toggling the Signup Affirmation Modal to be hidden upon successful signup. Quiz signup conversions will now toggle this to false.

For successful signups, by default, we'd set the redux state `shouldShowAffirmation` to `true`, this works fine except for Quiz conversions, where we don't want to show the affirmation modal (to allow results to be surfaced immediately and to prevent a Quiz re-render which would restart the quiz and hide the results).

This PR adds another `data` parameter field called `shouldShowAffirmation` (modeled after the existing field), which defaults to `true`, since that's our presumed case, but manually toggles this `false` for Quiz signups.

Refs [Pivotal ID #163655159](https://www.pivotaltracker.com/story/show/163655159)

### DIFF
--- a/resources/assets/actions/signup.js
+++ b/resources/assets/actions/signup.js
@@ -147,6 +147,7 @@ export function storeCampaignSignup(campaignId, data) {
             success: 'Thanks for signing up!',
             failure: 'Whoops! Something went wrong!',
           },
+          shouldShowAffirmation: get(data, 'shouldShowAffirmation', true),
         },
         pending: STORE_CAMPAIGN_SIGNUPS_PENDING,
         requiresAuthentication: true,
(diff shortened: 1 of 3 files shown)


weerd:
Quick thought.... what if we don't provide a default for the Affirmation if none added/linked to the `affirmation` field in Contentful and instead only show one if one is attached? And that decides whether a campaign gets an affirmation or not.

The current campaign this work is being done for currently has an affirmation attached... if no default and we just remove it, maybe that would be the less complicated and more intuitive fix for this...

thoughts 🤔

mendelB:
@weerd the issue is all of our Ashes imported database campaigns currently rely on the default Affirmation 😖

weerd:
> @weerd the issue is all of our Ashes imported database campaigns currently rely on the default Affirmation 😖

Yeah, ok. I was debating about how it would affect existing... 😞 

I'll make an icebox ticket to update and we can roll with this for now.

PR #1244, opened 2019-02-01
Signup Affirmation Toggle Fix

mendelB:
### What does this PR do?

This PR fixes a really clumsy mistake I made in #1243 grabbing the `shouldShowAffirmation` field in the signup reducer from the wrong `action` field. 

Also moving the evaluation of `shouldShowAffirmation` atop the signup action for clarity.

Refs [Pivotal ID #163655159](https://www.pivotaltracker.com/story/show/163655159)
#1243

### DIFF
--- a/resources/assets/reducers/signups.js
+++ b/resources/assets/reducers/signups.js
@@ -70,7 +70,7 @@ const signupReducer = (state = {}, action) => {
         data: signups,
         isPending: false,
         shouldShowAffirmation:
-          get(data, 'shouldShowAffirmation', true) &&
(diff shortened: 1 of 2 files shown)


resources/assets/reducers/signups.js
```diff
@@ -70,7 +70,7 @@ const signupReducer = (state = {}, action) => {
         data: signups,
         isPending: false,
         shouldShowAffirmation:
-          get(data, 'shouldShowAffirmation', true) &&
+          get(action, 'meta.shouldShowAffirmation', true) &&
```
weerd: meta

PR #1245, opened 2019-02-01
Send `utm_medium` to Northstar

katiecrane:
### What does this PR do?

✍️ Updates `.env.example` to provide a value (`http://experiments-qa.dosomething.org`) for `SIXPACK_BASE_URL`

🚦 Updates `getCampaignDataForNorthstar` to pass `utm_medium` to Northstar as `trafficSource`.

It's all in the Pivotal card!

Refs [Pivotal ID #163123198](https://www.pivotaltracker.com/story/show/163123198)

### Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -51,7 +51,7 @@ MAIL_PASSWORD=null
 MAIL_ENCRYPTION=null
 
 SIXPACK_ENABLED=false
-SIXPACK_BASE_URL=null
+SIXPACK_BASE_URL=http://experiments-qa.dosomething.org
 SIXPACK_COOKIE_PREFIX=null
 SIXPACK_TIMEOUT=null
 
--- a/.env.example
+++ b/.env.example
@@ -51,7 +51,7 @@ MAIL_PASSWORD=null
 MAIL_ENCRYPTION=null
 
 SIXPACK_ENABLED=false
-SIXPACK_BASE_URL=http://experiments-qa.dosomething.org
+SIXPACK_BASE_URL=https://experiments-qa.dosomething.org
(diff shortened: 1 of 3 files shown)


.env.example
```diff
@@ -51,7 +51,7 @@ MAIL_PASSWORD=null
 MAIL_ENCRYPTION=null
 
 SIXPACK_ENABLED=false
-SIXPACK_BASE_URL=null
+SIXPACK_BASE_URL=http://experiments-qa.dosomething.org
```
weerd: `https` 🤔

resources/assets/selectors/campaign.js
```diff
@@ -32,6 +32,11 @@ export function getCampaignDataForNorthstar(state) {
     data.title = state.campaign.title;
   }
 
+  // Pass along the utm_medium to Northstar
+  if (query('utm_medium')) {
```
weerd: _might_ be nicer to set a `const utmMedium = query('utm_medium');` under the `const data = {};` instead of calling `query()` twice. Thoughts? Also feel like there's a way to use a ternary for this with only calling the method once, but I can't recall offhand 🤔

weerd: Also not the biggest deal to call it twice, just felt a little cleaner if setting a variable 😄

katiecrane: @weerd yes I love that idea! I'm trying it out now

katiecrane: @weerd I cleaned it up even more because everything seems to work fine if we don't check that `query('utm_medium')` is defined, it just passes `null`

PR #1247, opened 2019-02-06
Adding migration script and entity for CompanyPage.

weerd:
### What does this PR do?

This PR adds a migration script file (exported from Contentful) for the new `companyPage` content type, and also adds a corresponding entity class.

🌵 

### Checklist

### DIFF
--- a/contentful/content-types/companyPage.js
+++ b/contentful/content-types/companyPage.js
@@ -1,0 +1,135 @@
+module.exports = function(migration) {
+  const companyPage = migration
+    .createContentType('companyPage')
+    .name('CompanyPage')
+    .description(
+      'A custom page for DoSomething static pages with company information.',
+    )
+    .displayField('internalTitle');
+  companyPage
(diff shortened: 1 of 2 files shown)


contentful/content-types/companyPage.js
```diff
@@ -0,0 +1,135 @@
...
+          ],
+        },
+      },
+    ])
+    .disabled(false)
+    .omitted(false);
+
+  // @TODO: remove linkAction from accepted types.
+  companyPage
+    .createField('blocks')
```
mendelB: If we use Rich Text, can we do away with the `blocks` field entirely, since those can just be added via the Rich Text field?

weerd: Ya, that's a good point, I was tempted to leave it, for Page types to allow for more flexibility, but might only be needed in certain pages and likely unnecessary for `CompanyPage`s. 🤔

PR #1248, opened 2019-02-06
Company pages setup continued

weerd:
### What does this PR do?

This PR does a few updates to support being able to properly query for `companyPage`s as a content type in Contentful. Since we cannot query Contentful with multiple content types in a query, I needed to find a reasonable way to determine what type of page the slug is referring to. The easiest spot seemed related to the `$category` for the slug. It all seems to work rather well 💃 

🌵 

### Checklist

### DIFF
--- a/app/Entities/Entity.php
+++ b/app/Entities/Entity.php
@@ -66,6 +66,8 @@ class Entity implements ArrayAccess, JsonSerializable
                 return new TruncatedCampaign($block->entry);
             case 'campaignUpdate':
                 return new CampaignUpdate($block->entry);
+            case 'companyPage':
+                return new CompanyPage($block->entry);
             case 'contentBlock':
                 return new ContentBlock($block->entry);
             case 'customBlock':
--- a/app/Repositories/QueriesContentful.php
+++ b/app/Repositories/QueriesContentful.php
@@ -2,8 +2,6 @@
 
 namespace App\Repositories;
 
-use App\Entities\Page;
-use App\Entities\Campaign;
 use App\Entities\HomePage;
 use Contentful\Delivery\Query;
 use App\Entities\TruncatedCampaign;
@@ -68,12 +66,11 @@ trait QueriesContentful
             return 'not_found';
         }
 
-        switch ($type) {
-            case 'campaign':
-                return json_encode(new Campaign($entry[0]));
+        // Eg: campaign -> Campaign
+        // Eg: companyPage -> CompanyPage
+        // Eg: page -> Page
+        $entityClass = '\\App\\Entities\\'.ucwords($type);
 
-            case 'page':
-                return json_encode(new Page($entry[0]));
-        }
+        return json_encode(new $entityClass($entry[0]));
     }
 }
--- a/contentful/content-types/companyPage.js
+++ b/contentful/content-types/companyPage.js
@@ -51,7 +51,7 @@ module.exports = function(migration) {
       },
       {
         regexp: {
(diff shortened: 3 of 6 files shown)


app/Entities/Entity.php
```diff
@@ -66,6 +66,8 @@ public function parseBlock($block)
                 return new TruncatedCampaign($block->entry);
             case 'campaignUpdate':
                 return new CampaignUpdate($block->entry);
+            case 'companyPage':
+                return new CompanyPage($block->entry);
```
weerd: We probably don't need this, but felt it might be useful to add anyway... Only really necessary if we foresee `CompanyPage`s being used in a reference field...

app/Repositories/QueriesContentful.php
```diff
@@ -68,12 +66,11 @@ protected function getEntryFromSlugAsJson($type, $slug)
...
-                return json_encode(new Campaign($entry[0]));
+        // Eg: campaign -> Campaign
+        // Eg: companyPage -> CompanyPage
+        // Eg: page -> Page
+        $entityClass = '\\App\\Entities\\'.ucwords($type);
 
-            case 'page':
-                return json_encode(new Page($entry[0]));
-        }
+        return json_encode(new $entityClass($entry[0]));
```
weerd: I wanted to try this out. Lemme know what y'all think! Most of the `switch` cases were basically the same, just calling a different class, which again was all based on a studly-cased version of the `$type`, so i figured we could simplify as I did above. But if you foresee issues with this tell me!

app/Entities/Entity.php
mendelB: I guess one scenario might be if we ever needed one of these pages in a gallery.

app/Repositories/QueriesContentful.php
mendelB: Super cool! Do you think it's worth sanitizing the `$type` before just in case, to ensure it's one of a valid list or is it fine since this is strictly internal?

contentful/content-types/companyPage.js
```diff
@@ -51,11 +51,11 @@ module.exports = function(migration) {
       },
       {
         regexp: {
-          pattern: '^[a-zA-Z0-9-/]+$',
+          pattern: '^(?!/)[a-zA-z0-9-/]+$',
```
mendelB: Is that downcase `z` in the `A-z` meant to be uppercase?

weerd: Ooof! Ya. Good catch :D Fixed it up on Contentful side but will update this migration script as well!

weerd:
> Super cool! Do you think it's worth sanitizing the `$type` before just in case, to ensure it's one of a valid list or is it fine since this is strictly internal?

@mendelB I _think_ it's fine. We'll see what the future hold 🔮

weerd:
> I guess one scenario might be if we ever needed one of these pages in a gallery.

Yep yep. I think that's the only scenario I could think of but seems unlikely to have a company page in a gallery, but who knows!

app/Repositories/QueriesContentful.php
weerd: @mendelB I _think_ it's fine. We'll see what the future hold 🔮

app/Entities/Entity.php
weerd: Yep yep. I think that's the only scenario I could think of but seems unlikely to have a company page in a gallery, but who knows!

PR #1249, opened 2019-02-07
Adding page type dispatcher

weerd:
### What does this PR do?

This PR updates to add a new `PageDispatcher` component that will help to render pages using the proper component, such as a `GeneralPage` or the new `CompanyPage`. It also updates the array in the `get_content_type_by_category()` helper method to continue to use `page` as the content type for `/about` pages for the time being until we're ready for a switcheroo.

🌵 

### Checklist

### DIFF
--- a/resources/assets/components/PageDispatcher/PageDispatcher.js
+++ b/resources/assets/components/PageDispatcher/PageDispatcher.js
@@ -1,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import CompanyPageContainer from '../pages/CompanyPage/CompanyPageContainer';
+import GeneralPageContainer from '../pages/GeneralPage/GeneralPageContainer';
+
+const PageDispatcher = ({ type }) => {
+  switch (type) {
+    case 'companyPage':
+      return <CompanyPageContainer />;
+
+    default:
+      return <GeneralPageContainer />;
+  }
+};
+
+PageDispatcher.propTypes = {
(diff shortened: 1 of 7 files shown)


resources/assets/components/PageDispatcher/PageDispatcher.js
```diff
@@ -0,0 +1,21 @@
...
+import CompanyPageContainer from '../pages/CompanyPage/CompanyPageContainer';
+import GeneralPageContainer from '../pages/GeneralPage/GeneralPageContainer';
+
+const PageDispatcher = ({ type }) => {
+  switch (type) {
+    case 'companyPage':
+      return <CompanyPageContainer />;
+
+    default:
+      return <GeneralPageContainer />;
```
weerd: One interesting note here, since we now have a `PageDispatcherContainer` we could potentially grab the data we need from the redux store in that container and basically eliminate the containers for both `GeneralPage` and `CompanyPage`, passing that info as props in the above component elements. 🤔 Something to ponder for upcoming PR...

mendelB: I think that makes sense! Kinda like how we do it with the [LedeBanner](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/LedeBanner/LedeBanner.js)?

weerd: Ya 💃 I can give it a whirl in an upcoming PR.

PR #1250, opened 2019-02-08
Add Action ID field to RB Action Content Types

mendelB:
### What does this PR do?

This PR adds a new Action ID field to the RB Action Contentful Content Types.

This affects the  `photoSubmissionAction`, `textSubmissionAction`, and `shareAction` content types.

- replaces migration scripts from `/contentful/migrations` with singular content type scripts using contentful-cli generated migrations in `/contentful/content-types` (It might be helpful to commit-by-commit the first few commits of this PR for the isolated migration script replacements)
- adds the new `actionId` field and helper text to the migration scripts (see (https://github.com/DoSomething/phoenix-next/commit/2ca3eea83bfefcd1fa8103e74b0ef4d553381aa3) for the isolated diff on the migration scripts)
- adds the new `actionId` field to the respective `Entity`s

- This is part of the new and improved process on Rogue for associating RB Posts with formal and detailed Action types, for enriched post metadata.

- The Action ID field is an integer in line with the [Rogue validation](https://github.com/DoSomething/rogue/blob/master/app/Http/Requests/PostRequest.php#L34) and is validated as unique since Actions are created individually for each RB action type on Rogue, which will correspond to the unique RB Action entry on Contentful. (Thanks @chloealee for clarification and context on this!) 

- Eventually, we're going to need to make this field required as well! (which I guess might require some discussion as to how to handle older RB Action entries on Contentful)

- The migration script migration is in line with our new approach to formalized content type migration scripts, as opposed to disperate evolving migration scripts. (a touch of context [in Slack](https://dosomething.slack.com/archives/C3ASB4204/p1549642926387200).)

Refs [Pivotal ID #163728937](https://www.pivotaltracker.com/story/show/163728937)


DFurnes:
Excited to see all this work coming together!! 👏

PR #1252, opened 2019-02-11
Refactoring markdown component

weerd:
### What does this PR do?

This PR refactors the `Markdown` component to start supporting both standard markdown and Contentful's new richtext-style markdown. It creates a new ` helpers/markdown.js` file to contain all the helper methods for rendering either standard or richtext markdown moving forward.

There's still a bit more to come with the richtext rendering and passing it an customized object to help with rendering embedded entities, etc, but this begins adding some basic support, without breaking prior standard markdown support!

🌵 

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "testURL": "http://localhost"
   },
   "dependencies": {
-    "@contentful/rich-text-html-renderer": "^13.0.0",
+    "@contentful/rich-text-react-renderer": "^13.0.1",
     "@dosomething/analytics": "^2.1.2",
     "@dosomething/forge": "^6.8.0",
     "@dosomething/gateway": "^1.4.0",
@@ -88,7 +88,7 @@
     "markdown-it": "^8.4.0",
     "markdown-it-footnote": "^3.0.1",
     "markdown-it-for-inline": "^0.1.1",
-    "prettier": "^1.15.3",
+    "prettier": "^1.16.4",
     "prop-types": "^15.6.2",
     "query-string": "^5.1.1",
     "react": "^16.6.3",
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "testURL": "http://localhost"
   },
   "dependencies": {
+    "@contentful/rich-text-html-renderer": "^13.0.0",
     "@dosomething/analytics": "^2.1.2",
     "@dosomething/forge": "^6.8.0",
     "@dosomething/gateway": "^1.4.0",
--- a/resources/assets/components/utilities/Markdown/Markdown.js
+++ b/resources/assets/components/utilities/Markdown/Markdown.js
@@ -1,27 +1,31 @@
-/* eslint-disable react/no-danger */
-
 import React from 'react';
+import { has } from 'lodash';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
-import { createMarkup } from '../../../helpers/markdown';
+import RichTextDocument from './RichTextDocument';
+import StandardMarkdown from './StandardMarkdown';
 
 import './markdown.scss';
 
 /**
  * Render Markdown as Markup prepared for a React Component
- * @see  https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml
  *
- * @param  {String} props.className
- * @param  {String|Array|Object} props.children
+ * @TODO: rename component to <TextContent>
+ * @param  {String} options.className
+ * @param  {String|Array|Object} options.children
  * @return {Object}
  */
-const Markdown = ({ className = null, children }) => (
-  <div
-    className={classnames('markdown', 'with-lists', className)}
-    dangerouslySetInnerHTML={{ __html: createMarkup(children) }}
-  />
-);
+const Markdown = ({ className = null, children }) =>
+  has(children, 'nodeType') ? (
+    <RichTextDocument className={classnames(className)}>
+      {children}
+    </RichTextDocument>
+  ) : (
+    <StandardMarkdown className={classnames(className)}>
+      {children}
+    </StandardMarkdown>
+  );
 
 Markdown.propTypes = {
   children: PropTypes.oneOfType([
--- a/resources/assets/components/utilities/Markdown/Markdown.js
+++ b/resources/assets/components/utilities/Markdown/Markdown.js
@@ -4,33 +4,30 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
-import { markdown, contentfulImageUrl } from '../../../helpers';
+import { createMarkup } from '../../../helpers/markdown';
 
 import './markdown.scss';
 
-const pattern = /\/\/images\.(ctfassets\.net|contentful\.com).+\.(jpg|png)/g;
-const contentfulImageFormat = url => contentfulImageUrl(url, '1000');
-const formatImageUrls = string =>
-  string.replace(pattern, contentfulImageFormat);
-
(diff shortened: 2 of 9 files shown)


resources/assets/helpers/markdown.js
```diff
@@ -0,0 +1,94 @@
...
+  return documentToHtmlString(markdown);
+}
+
+/**
+ * Parse Standard Markdown to Markup.
+ *
+ * @param  {String|Array} markdown
+ * @return {String}
+ */
+export function parseStandardMarkdown(markdown = '') {
```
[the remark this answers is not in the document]
weerd: 🤔 yeah, I could add some sort of fail-safe.

package.json
```diff
@@ -50,6 +50,7 @@
     "testURL": "http://localhost"
   },
   "dependencies": {
+    "@contentful/rich-text-html-renderer": "^13.0.0",
```
DFurnes: I was peeking to see whether this pulled in a Markdown parser of it's own (and what effect that'd have on bundle size). Good news - it doesn't (since text is formatted as objects a la `{ nodeType: 'text', value: 'Yo', marks: [{ type: 'bold' }]}`.

But then I stumbled upon… a [React renderer](https://github.com/contentful/rich-text/tree/master/packages/rich-text-react-renderer) that outputs React components instead of HTML!

resources/assets/components/utilities/Markdown/Markdown.js
```diff
@@ -4,33 +4,30 @@ import React from 'react';
...
-};
+/**
+ * Render Markdown as Markup prepared for a React Component
+ * @see  https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml
+ *
+ * @param  {String} props.className
+ * @param  {String|Array|Object} props.children
+ * @return {Object}
+ */
+const Markdown = ({ className = null, children }) => (
```
DFurnes: I had been under the assumption that Contentful's rich text was a superset of Markdown, but it's not! In that case, it might be clearer to name the top-level component a more generic `TextContent` or something instead.

resources/assets/helpers/markdown.js
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Format Standard Markdown to a standard markdown string.
+ *
+ * @param  {String|Array} markdown
+ * @return {String}
+ */
+function formatStandardMarkdown(markdown) {
+  // When directly writing content into a Markdown component, React may pass it as an
+  // array of strings; if so, combine them to get the Markdown source!
+  const formattedMarkdown = Array.isArray(markdown)
```
DFurnes: Maybe `flattenedMarkdown` (or keep as `sourceMarkdown`)? This isn't really "formatted" here.

resources/assets/helpers/markdown.js
```diff
@@ -0,0 +1,94 @@
...
+  return markdownIt;
+}
+
+/**
+ * Parse RichText Markdown to Markup.
+ *
+ * @param  {Object} markdown
+ * @return {String}
+ */
+export function parseRichTextMarkdown(markdown) {
```
DFurnes: As above, wonder if this should just be called `parseRichText` instead?

package.json
weerd: 🤔

resources/assets/helpers/markdown.js
weerd: Yeah, I wasn't super happy with the name for this... `sourceMarkdown` just sounded weird to me, but I kinda dig `flattenedMarkdown`. Will give that a whirl.

resources/assets/components/utilities/Markdown/Markdown.js
weerd: Yeah, that makes sense. I might save that for a future PR though since there's going to be a number of file changes to update to switch the name from `Markdown` to `TextContent` where this component is used.

weerd: @DFurnes in a similar regard, we'd probably want to rename the `helpers/markdown.js` file as well. Would it be `helpers/textContent.js` or something?

package.json
weerd: @dave based on your suggestion to use the contentful `rich-text-react-renderer` instead of the html one, I think I need to restructure the stuff in that PR a bit since I can’t pass the eventual output for either (markdown or richtext) all as a prop of `dangerouslySetInnerHTML`. So instead, thinking the of restructuring to:

```
/utilities/Markdown becomes /utilities/TextContent
  - Markdown.js becomes TextContent.js and will decide how to render and send to the other 2 components below
  - StandardMarkdown.js - component that renders markdown
  - RichTextDocument.js - component that renders richtext using the `rich-text-react-renderer` package
```

DFurnes: @weerd Yeah, I think that's a great approach! 👍

resources/assets/components/utilities/Markdown/Markdown.js
DFurnes: Yeah, that makes sense. For brevity, I'm kinda into `helpers/text.js`. 📃

resources/assets/components/utilities/Markdown/Markdown.js
```diff
@@ -1,36 +1,37 @@
...
-    />
+/**
+ * Render Markdown as Markup prepared for a React Component
+ *
+ * @TODO: rename component to <TextContent>
+ * @param  {String} options.className
+ * @param  {String|Array|Object} options.children
+ * @return {Object}
+ */
+const Markdown = ({ className = null, children }) =>
```
weerd: Will be renaming this component to `TextContent` in upcoming PR, but didn't want the numerous files that would affect to drown out the code changes for all this new behavior!

resources/assets/components/utilities/Markdown/RichTextDocument.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Return React Components for provided RichText.
+ * @see  https://github.com/contentful/rich-text/tree/master/packages/rich-text-react-renderer
+ *
+ * @param  {String} options.className
+ * @param  {Object} options.children
+ * @return {Object}
+ */
+const RichTextDocument = ({ className = null, children }) => (
+  <div className={classnames('markdown rich-text', 'with-lists', className)}>
```
weerd: Leaving the `markdown` class for now, so styling applies, but may end up renaming the class and style file to apply to both markdown and richtext in the future.

resources/assets/helpers/text.js
```diff
@@ -0,0 +1,81 @@
...
+  return markdownIt;
+}
+
+/**
+ * Parse RichText Document to React components.
+ *
+ * @param  {Object} document
+ * @return {String}
+ */
+export function parseRichTextDocument(document) {
```
DFurnes: It might be cleaner to keep this in the `RichTextDocument` component, since it likely won't (or even shouldn't) ever be used outside of that context.

weerd: Yeah, I'm waiting to see once we come back to this and implement configuring the options, which could be lengthy and convoluted, so it might be nice to keep that separated out and simplify the component. I was planning on reassessing once we get to the second part of this RichText work.

resources/assets/helpers/markdown.js
weerd: Made this `cleanupStandardMarkdown` 🤷‍♂️

resources/assets/helpers/markdown.js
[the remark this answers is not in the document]
weerd: Fixed!

PR #1253, opened 2019-02-12
Unsubmitted Quiz Event

mendelB:
### What does this PR do?

This PR adds Puck tracking for an incomplete or unsubmitted quiz.

It's called `phoenix_left_quiz` but I am not completely confident that that's the best name for this. (`abandoned`? `incomplete`?, `left_incomplete`, `hates_quizzes`?)

- https://github.com/DoSomething/phoenix-next/commit/6a404333d1815f25e1df6258a9cd4f96d34c8948 adds a new state property called `completedQuiz`. This allows us to determine completion or 'submission' status irrespective of the quiz being an autocomplete, gated, or what have you quiz.

- The `beforeunload` event was necessary since a user would primarilly be abandoning a quiz by exiting the page somehow, which would fail to trigger a `componentWillUnmount`. (In fact, I'm not currently sure if there's a regular user case of leaving a quiz with a `componentWillUnmount` being triggered, but I did add that secenario to be sure.)

Refs [Pivotal ID #163570667](https://www.pivotaltracker.com/story/show/163570667)

### Checklist

### DIFF
--- a/resources/assets/components/Quiz/Quiz.js
+++ b/resources/assets/components/Quiz/Quiz.js
@@ -55,7 +55,7 @@ class Quiz extends React.Component {
   }
 
   componentDidMount() {
-    window.addEventListener('beforeunload', this.trackUnfinishedQuiz);
+    window.addEventListener('beforeunload', this.trackAbandonedQuiz);
   }
 
   componentDidUpdate() {
@@ -69,16 +69,16 @@ class Quiz extends React.Component {
   }
 
   componentWillUnmount() {
-    window.removeEventListener('beforeunload', this.trackUnfinishedQuiz);
-    this.trackUnfinishedQuiz();
     /*
       Users would primarily be abandoning a quiz by exiting the page completely.
       This addresses the edge case, where the quiz is abandoned while remaining within the page
       (thus triggering the componentWillUnmount.)
     */
+    window.removeEventListener('beforeunload', this.trackAbandonedQuiz);
+    this.trackAbandonedQuiz();
   }
 
-  trackUnfinishedQuiz = () => {
+  trackAbandonedQuiz = () => {
     if (this.state.completedQuiz) {
       return;
     }
--- a/resources/assets/components/Quiz/Quiz.js
+++ b/resources/assets/components/Quiz/Quiz.js
@@ -54,6 +54,10 @@ class Quiz extends React.Component {
     setTimeout(() => this.setState({ renderScrollConcierge: false }), 500);
   }
 
+  componentDidMount() {
+    window.addEventListener('beforeunload', this.trackUnfinishedQuiz);
+  }
+
   componentDidUpdate() {
     if (!this.props.autoSubmit) {
       return;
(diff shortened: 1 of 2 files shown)


weerd:
I think I like `abandoned` better.

weerd:
I don't know why I have a tendency to to click the "close and comment" button when I just want to comment 🤦‍♂️ I think my mind reads the green button as the "merge" button and immediately goes to the one next to it since my intention is to comment and not merge. sigh.

resources/assets/components/Quiz/Quiz.js
```diff
@@ -47,22 +47,46 @@ class Quiz extends React.Component {
...
     }
 
-    if (!this.state.showResults && this.evaluateQuiz()) {
+    if (!this.state.completedQuiz && this.evaluateQuiz()) {
       setTimeout(this.completeQuiz, 300);
     }
   }
 
+  componentWillUnmount() {
+    window.removeEventListener('beforeunload', this.trackUnfinishedQuiz);
```
DFurnes: Could you add a comment explaining this? (Copy-pasting from your PR description is perf.)

resources/assets/components/Quiz/Quiz.js
```diff
@@ -47,22 +47,46 @@ class Quiz extends React.Component {
...
+    window.addEventListener('beforeunload', this.trackUnfinishedQuiz);
+  }
+
   componentDidUpdate() {
     if (!this.props.autoSubmit) {
       return;
     }
 
-    if (!this.state.showResults && this.evaluateQuiz()) {
+    if (!this.state.completedQuiz && this.evaluateQuiz()) {
```
DFurnes: I'm having a hard time figuring out this change. 😖 Could you outline one of the scenarios you mentioned in the PR description where `showResults` wouldn't have worked here?

mendelB: Yeah sorry, it's a confusing flow to begin with!
So if the quiz is gated, and the user clicks 'see results' after answering all the questions, and are redirected to Northstar, they'd be [redirected](https://github.com/DoSomething/phoenix-next/blob/quiz-abandonment-event/resources/assets/components/Quiz/Quiz.js#L137) without us [setting](https://github.com/DoSomething/phoenix-next/blob/quiz-abandonment-event/resources/assets/components/Quiz/Quiz.js#L148) the `showResults: true`, which'd trigger a faulty abandoned quiz event.

So adding this `completedQuiz` state allows us to [track](https://github.com/DoSomething/phoenix-next/blob/quiz-abandonment-event/resources/assets/components/Quiz/Quiz.js#L103) that they've technically completed the quiz even if we redirect before showing the results.

mendelB: In the above case, we'd [update the state](https://github.com/DoSomething/phoenix-next/blob/quiz-abandonment-event/resources/assets/components/Quiz/Quiz.js#L103) with `showResults: true` in the `completeQuiz` function, but then the `componentWillUpdate` will trigger again, so this ensures it won't call `completeQuiz` infinitely.

Sorry, I hope I'm articulating this clearly

DFurnes: Ahhhhh, that's so tricky!

mendelB: Totally. The multiple potential states of the quiz really bake complexity into it 😖

PR #1255, opened 2019-02-13
Favicons galore

weerd:
### What does this PR do?

This PR adds favicons of 2 flavors that browsers can choose from, as well as the apple touch icon for device screens. I included a `link` in the `master.blade.php` to help with Android and Chrome recognizing the touch icon. We'll see if that works or not, but otherwise this is also to help with reducing the number of 404s we were getting in Papertrail for both favicons and apple touch icons!

💅 

🌵


resources/views/layouts/master.blade.php
```diff
@@ -18,6 +18,7 @@
     @include('partials.script_loading')
 
     <link rel="icon" type="image/ico" href={{config('app.url') . "/favicon.ico?v1" }}>
+    <link rel="icon" sizes="144x144" href="/apple-touch-icon-precomposed.png">
```
weerd: Technically for iOS you no longer need to specify any info on where to look for a touch icon, instead Apple devices just scour the root directory to find one. However, Android and Chrome use the `link` reference, and here, I'm just referencing the same touch icon we otherwise use on Apple devices.

PR #1256, opened 2019-02-13
Action ID Field in RB Posts

mendelB:
### What does this PR do?

This PR adds the `action_id` field to our various RB action post's form data.

- I did not remove the `action` field, since it's presently still [required in Rogue](https://github.com/DoSomething/rogue/blob/master/app/Http/Requests/PostRequest.php#L33) and used to [auto assign Actions to Posts](https://github.com/DoSomething/rogue/blob/master/app/Repositories/PostRepository.php#L87)
- we needed some logic to only append the `action_id` field *if* it is not falsey, since, if it's sent over as a `null` value it'll [fail validation](https://github.com/DoSomething/rogue/blob/master/app/Http/Requests/PostRequest.php#L34)

**Question**:
Should I add a validation rule to our [`PostRequest#rules`](https://github.com/DoSomething/phoenix-next/blob/master/app/Http/Requests/PostRequest.php#L54) for `action_id`? 
It'd need to be for all post types, so it might require shifting the logic around a bit, or just repeating the rule for all three action types `case`s

Refs [Pivotal ID #163728937](https://www.pivotaltracker.com/story/show/163728937)

### Checklist

### DIFF
--- a/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
+++ b/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
@@ -399,6 +399,7 @@ class PhotoSubmissionAction extends React.Component {
 }
 
 PhotoSubmissionAction.propTypes = {
+  actionId: PropTypes.number,
   affirmationContent: PropTypes.string,
   additionalContent: PropTypes.shape({
     action: PropTypes.string,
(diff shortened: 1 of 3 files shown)


DFurnes:
> Should I add a validation rule to our [`PostRequest#rules`](https://github.com/DoSomething/phoenix-next/blob/master/app/Http/Requests/PostRequest.php#L54) for `action_id`? 
It'd need to be for all post types, so it might require shifting the logic around a bit, or just repeating the rule for all three action types `case`s

It looks like we've only been validating user-facing fields in there so far, so it makes sense to me to let Rogue's validation handle this one (since a "something went wrong" error and an `action_id is required` message are both equally unhelpful for a user).

mendelB:
Oh right, haha good point!

weerd:
Belated 👍 💃 🕺

PR #1258, opened 2019-02-15
Fixing bug previously introduced when refacorting pages.

weerd:
### What does this PR do?

This PR fixes a bug with the `PageRepository` that was leading to 500 errors on categorized pages. The bug wasn't immediately evident when testing since it only really applied when pages where being cached. The issue was a missing `$type` variable not being passed via `use` for the function's scope.

This issue also brought to mind whether we may want to set `CONTENTFUL_CACHE` to TRUE on QA instead of FALSE to fully represent how production works and so that this bug would have been found on QA instead of on Production.

🐞 

### Checklist

### DIFF
--- a/app/Repositories/PageRepository.php
+++ b/app/Repositories/PageRepository.php
@@ -11,16 +11,17 @@ class PageRepository
     /**
      * Find a page by its slug.
      *
+     * @param  string $type
      * @param  string $slug
      * @return \Contentful\Delivery\Resource\Entry
      * @throws \Illuminate\Database\Eloquent\ModelNotFoundException
      */
-    public function findBySlug($slug, $type)
+    public function findBySlug($type, $slug)
     {
         if (! config('services.contentful.cache')) {
             $page = $this->getEntryFromSlugAsJson($type, $slug);
         } else {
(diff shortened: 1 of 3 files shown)


app/Repositories/PageRepository.php
```diff
@@ -11,16 +11,17 @@ class PageRepository
     /**
      * Find a page by its slug.
      *
+     * @param  string $type
      * @param  string $slug
      * @return \Contentful\Delivery\Resource\Entry
      * @throws \Illuminate\Database\Eloquent\ModelNotFoundException
      */
-    public function findBySlug($slug, $type)
+    public function findBySlug($type, $slug)
```
weerd: I switched around the parameter signature here so it matches how the two are used in subsequent code and the `getEntryFromSlugAsJson()` method.

weerd:
Thanks to @DFurnes for helping 🕵️this one out 💃

mendelB:
> This issue also brought to mind whether we may want to set `CONTENTFUL_CACHE` to TRUE on QA instead of FALSE to fully represent how production works and so that this bug would have been found on QA instead of on Production.

I think this makes good sense!

PR #1260, opened 2019-02-15
Story page content type migration

weerd:
### What does this PR do?

This PR add a new content-type generated migration script for the new `StoryPage` content type.

I refrained from including a `blocks` field for the time being to keep things simple. I foresee likely needing to add it to this page type in the future.

I also limited the RichText field options to only markdown for now. It does not include support for Hyperlinks or Emebedded entries and assets for now. This is only temporary as I build out more robust support for the RichText field and how its content is parsed.

Refs [Pivotal ID #163913064](https://www.pivotaltracker.com/story/show/163913064)

### DIFF
--- a/contentful/content-types/storyPage.js
+++ b/contentful/content-types/storyPage.js
@@ -1,13 +1,13 @@
 module.exports = function(migration) {
-  const narrativePage = migration
-    .createContentType('narrativePage')
-    .name('NarrativePage')
+  const storyPage = migration
+    .createContentType('storyPage')
+    .name('StoryPage')
     .description(
-      'Narrative pages are used for helping craft a story or experience on a single web page.',
+      'Story pages are used for helping craft a story or experience on a single web page.',
     )
     .displayField('internalTitle');
 
-  narrativePage
+  storyPage
     .createField('internalTitle')
     .name('Internal Title')
     .type('Symbol')
@@ -16,7 +16,7 @@ module.exports = function(migration) {
     .validations([])
     .disabled(false)
     .omitted(false);
-  narrativePage
+  storyPage
     .createField('title')
     .name('Title')
     .type('Symbol')
@@ -25,7 +25,7 @@ module.exports = function(migration) {
     .validations([])
     .disabled(false)
     .omitted(false);
-  narrativePage
+  storyPage
     .createField('subTitle')
     .name('Subtitle')
     .type('Symbol')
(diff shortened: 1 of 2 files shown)


contentful/content-types/storyPage.js
```diff
@@ -0,0 +1,130 @@
...
+        unique: true,
+      },
+      {
+        size: {
+          max: 100,
+        },
+      },
+      {
+        regexp: {
+          pattern: '^(?!/)story/[a-zA-Z0-9-/]+$',
```
weerd: Made it so that the slug has to begin with `story/` 💃

contentful/content-types/storyPage.js
```diff
@@ -0,0 +1,130 @@
...
+    .omitted(false)
+    .linkType('Entry');
+
+  storyPage
+    .createField('content')
+    .name('Content')
+    .type('RichText')
+    .localized(false)
+    .required(false)
+    .validations([
```
mendelB: Why do we have these validations here unlike in CompanyPages? Are we excluding certain Markdown types?

contentful/content-types/storyPage.js
```diff
@@ -0,0 +1,130 @@
...
+    helpText:
+      'This title is used internally to help find this content. It will not be displayed anywhere on the rendered web page.',
+  });
+
+  storyPage.changeEditorInterface('title', 'singleLine', {});
+  storyPage.changeEditorInterface('subTitle', 'singleLine', {});
+
+  storyPage.changeEditorInterface('slug', 'slugEditor', {
+    helpText:
+      'If it applies, add an appropriate category prefix for the slug, e.g. "articles/", "about/", "facts/", etc.',
```
mendelB: Should we update this to be just `"story/"`?

contentful/content-types/storyPage.js
weerd: Yeah, I mention it in description. I'm keeping it to only simple markdown for now on these pages until I can further enhance how we handle the RichText field on Phoenix :)

contentful/content-types/storyPage.js
weerd: Oh ya.

contentful/content-types/storyPage.js
mendelB: Ohhh ok, I didn't realize that this is what's needed to invalidate embedded links

weerd: Yeah, this is all just temporary. As this is all getting built out 💃

PR #1262, opened 2019-02-19
Further favicon fix

weerd:
### What does this PR do?

This PR fixes a slightly confusing issue regarding where the site icons are stored. In #1255, I added the apple touch icon and favicons to the repository root directory instead of the `public/` directory which is the app document root, thus we were still getting a bunch of 404s.

🤦‍♂️ 

🌵 

### Checklist


resources/views/layouts/master.blade.php
```diff
@@ -18,7 +18,7 @@
     @include('partials.script_loading')
 
     <link rel="icon" type="image/ico" href={{config('app.url') . "/favicon.ico?v1" }}>
-    <link rel="icon" sizes="144x144" href="/apple-touch-icon-precomposed.png">
+    <link rel="icon" sizes="144x144" href={{config('app.url') . "/apple-touch-icon-precomposed.png"}}>
```
DFurnes: No need for `config('app.url')` here. Let's just use a root-relative URL.

weerd: Hokay. Added it since the favicon had it. I can remove from that as well!

PR #1266, opened 2019-02-21
Append location to post submissions.

DFurnes:
### What does this PR do?
This pull request reads the `X-Fastly-Country-Code` and `X-Fastly-Region-Code` headers, if present, and uses them to set the `location` field on any posts that a user submits.

I had thought we could read the headers directly from the front-end but [it turns out we can't](https://stackoverflow.com/a/4881836/811624)!

Refs [Pivotal ID #163728657](https://www.pivotaltracker.com/story/show/163728657).

### Checklist

### DIFF
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -1,4 +1,4 @@
-/* global FormData */
+/* global FormData, window */
 
 import { join } from 'path';
 
@@ -79,6 +79,11 @@ export function storeCampaignPost(campaignId, data) {
   }
 
   const { action, body, id, type } = data;
+
+  // Attach location information, provided by Fastly.
+  if (window.AUTH.location) {
+    body.append('location', window.AUTH.location);
+  }
 
   const sixpackExperiments = {
     conversion: 'reportbackPost',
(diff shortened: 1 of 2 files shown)


resources/assets/actions/post.js
```diff
@@ -80,6 +80,11 @@ export function storeCampaignPost(campaignId, data) {
 
   const { action, body, id, type } = data;
 
+  // Attach location information, provided by Fastly.
+  if (window.AUTH.location) {
```
mendelB: Unrelated to this PR specifically, but would it make sense at all to have a formalized way of interacting with our `window.AUTH` similar to [what we have](https://github.com/DoSomething/phoenix-next/blob/eb9639c57ff803759267163c0d45937f8aec2e31/resources/assets/helpers/index.js#L436-L438) with `window.ENV`? Or killing both birds with one helper?

DFurnes: Yeah, that's a good idea! 👍

weerd: 🤔 yeah, maybe. 

sidenote: I'd also like to just eventually move everything under the `window.DS` umbrella ☂️ and have that handle things better.

mendelB: Ok, added a Pivotal ticket to the icebox!
https://www.pivotaltracker.com/story/show/164146718

PR #1269, opened 2019-02-21
Update our API Post Requests FormData Formatting

mendelB:
### What does this PR do?

This PR updates our `middleware/api#postRequest` to format the request's content-type header based on the `data.body`'s type.

This enabled removing the `FormData`-fication of all post requests within that same method and allows removing that validation and formatting from `ShareAction` and `TextSubmissionAction` since those don't attach any files to the request's body.

❗️One catch here was that the helper method which took care of converting the form's body to `FormData`, [also took care](https://github.com/DoSomething/phoenix-next/blob/e3725ed887c1efd9e3d2cd75f5209848c1d69a73/resources/assets/helpers/forms.js#L79-L85) of properly serializing any `body.details` fields which get [validated as a string](https://github.com/DoSomething/phoenix-next/blob/e3725ed887c1efd9e3d2cd75f5209848c1d69a73/app/Http/Controllers/Api/CampaignSignupsController.php#L66).

I had initially wanted to add a new helper somewhere to abstract that logic from the existing form helper. But couldn't really think of an elegant way to do so, and felt like it even made more sense for the various components themselves to take care of this. Anyone have contrasting thoughts?

Refs [Pivotal ID #162349035](https://www.pivotaltracker.com/story/show/162349035)
#1265

### DIFF
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -70,7 +70,8 @@ export function resetPostSubmissionItem(id) {
  * @return {function}
  */
 export function storeCampaignPost(campaignId, data) {
-  const { action, body, id, type } = data;
+  const { action, id, type } = data;
+  let { body } = data;
 
   if (type === 'photo' && !(body instanceof FormData)) {
     throw Error(
@@ -82,7 +83,11 @@ export function storeCampaignPost(campaignId, data) {
 
   // Attach location information, provided by Fastly.
   if (window.AUTH.location) {
-    body.append('location', window.AUTH.location);
+    if (body instanceof FormData) {
+      body.append('location', window.AUTH.location);
+    } else {
+      body = { ...body, location: window.AUTH.location };
+    }
   }
 
   const sixpackExperiments = {
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -1,4 +1,4 @@
-/* global window */
+/* global FormData, window */
 
 import { join } from 'path';
 
@@ -71,6 +71,14 @@ export function resetPostSubmissionItem(id) {
  */
 export function storeCampaignPost(campaignId, data) {
   const { action, body, id, type } = data;
+
+  if (type === 'photo' && !(body instanceof FormData)) {
+    throw Error(
+      `The supplied data.body must be an instance of FormData, instead it is an instance of ${
+        body.constructor.name
+      }.`,
+    );
+  }
 
   // Attach location information, provided by Fastly.
   if (window.AUTH.location) {
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -1,4 +1,4 @@
-/* global FormData, window */
+/* global window */
 
 import { join } from 'path';
 
@@ -70,14 +70,6 @@ export function resetPostSubmissionItem(id) {
  * @return {function}
  */
 export function storeCampaignPost(campaignId, data) {
-  if (!(data.body instanceof FormData)) {
-    throw Error(
-      `The supplied data.body must be an instance of FormData, instead it is an instance of ${
-        data.body.constructor.name
-      }.`,
-    );
-  }
-
   const { action, body, id, type } = data;
 
   // Attach location information, provided by Fastly.
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -1,4 +1,4 @@
-/* global FormData, window */
+/* global FormData */
 
 import { join } from 'path';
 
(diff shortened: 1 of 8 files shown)


resources/assets/actions/post.js
```diff
@@ -70,14 +70,6 @@ export function resetPostSubmissionItem(id) {
  * @return {function}
  */
 export function storeCampaignPost(campaignId, data) {
-  if (!(data.body instanceof FormData)) {
```
weerd: Would it be useful to keep this for `photo` or whatever the string is for photo types?

```js
if (data.type === 'photo' && !(data.body instanceof FormData)) {
  throw Error(...);
}
```
🤔

DFurnes: I think that's a good idea! 👍

resources/assets/actions/post.js
```diff
@@ -70,14 +70,6 @@ export function resetPostSubmissionItem(id) {
...
-    throw Error(
-      `The supplied data.body must be an instance of FormData, instead it is an instance of ${
-        data.body.constructor.name
-      }.`,
-    );
-  }
-
   const { action, body, id, type } = data;
 
   // Attach location information, provided by Fastly.
```
DFurnes: Ah, I set you up with some trouble in #1266 – this conditional assumes `body` is a FormData. I think we'll need to have another if-else inside here to either directly set `body.location` or use `body.append()` if it's a FormData…

resources/assets/actions/post.js
mendelB: Ok, I'll add that!

resources/assets/components/SignupButton/SignupButton.js
```diff
@@ -22,7 +22,7 @@ const SignupButton = props => {
   // Decorate click handler for A/B tests & analytics.
   const onSignup = buttonText => {
     clickedSignupAction(campaignId, {
-      body: { details: { campaignContentfulId } },
+      body: { details: JSON.stringify({ campaignContentfulId }) },
```
weerd: I think stringifying in the respective location is fine for now. Not many places it needs to happen.

resources/assets/actions/post.js
```diff
@@ -70,19 +70,24 @@ export function resetPostSubmissionItem(id) {
...
       }.`,
     );
   }
 
-  const { action, body, id, type } = data;
-
   // Attach location information, provided by Fastly.
   if (window.AUTH.location) {
-    body.append('location', window.AUTH.location);
+    if (body instanceof FormData) {
```
mendelB: Nested `if`s ftw...

Wild thought -- would it at all make sense, to add a new cool helper method which can take care of either `FormData`-ifying the fields if necessary, stringifying `details` if necessary, and adding location if necessary?
Essentially one centralized place to add our RB post custom logic, so we don't have to do all the guesswork in various places?

mendelB: This all could be additions to the current Form helper
https://github.com/DoSomething/phoenix-next/blob/9d9cf4aaa4b2d3021d9a7d484a609359fc392a3b/resources/assets/helpers/forms.js#L76

weerd: I think I could be down for that, but weary to place it all within the `setFormData()`, since that's meant to be more specific for `FormData` stuff.

Maybe another general method, that calls `setFormData()` if it's meant actually be `FormData` otherwise return as simple object? 🤔

mendelB: That sounds wiser! I'll try and whip something up

mendelB: Ok! I added the new helper https://github.com/DoSomething/phoenix-next/pull/1269/commits/f652ababcd3cf52fc63a39e409676a73946f2bf8. Not sold on the naming, but I *think* the logic is sound and would help us clean up https://github.com/DoSomething/phoenix-next/pull/1269/commits/505fc61a8fc4ef01018c21224df9edd92904f9e2 the assorted formatting we were doing across RB posts

resources/assets/helpers/forms.js
```diff
@@ -110,3 +101,32 @@ export function getFormData(formData) {
...
+  }
+
+  // Attach location information, provided by Fastly.
+  if (window.AUTH.location) {
+    formattedData = { ...formattedData, location: window.AUTH.location };
+  }
+
+  // Photo RBs contain file uploads and thus need to be converted to a FormData object.
+  if (data.type === 'photo') {
+    formattedData = setFormData(formattedData);
```
DFurnes: I know this wasn't new in this pull request, but can we rename this method to `createFormData`? The `setFormData` naming makes me think it's modifying an existing thing.

weerd:
This is looking awesome! I agree with @DFurnes's recommendation to change `setFormData` to `createFormData`, and after that good to merge-y merge 👍

mendelB:
https://github.com/DoSomething/phoenix-next/pull/1269/commits/93027fbb625e815dfb886c1a06794d5917439a84 Updated the `setFormData` helper to `createFormData`.  (https://github.com/DoSomething/phoenix-next/pull/1269/commits/8a1797f61e95bcbf11eb0d6c0b28c1634bf64fa2 Snuck in adding the new `formatFormFields` to the -I believe defunct- `ReferralSubmissionAction`)

I'll merge this in!

PR #1270, opened 2019-02-21
Adding API route and related functionality for blocks.

weerd:
### What does this PR do?

This PR adds a new API `/blocks/{id}` route, which for only has a show method for now in the `Api\BlocksController`. I didn't add a `BlockRepository` because it seemed like overkill for now, but did add a new method to the `QueriesContentful` trait, called `getEntryFromIdAsJson()` which will grab a Contentful entry via a supplied entry ID.

This endpoint is needed so that we can render out entries added via RichText.

Refs [Pivotal ID #164092381](https://www.pivotaltracker.com/story/show/164092381)

### Checklist

### DIFF
--- a/app/Repositories/QueriesContentful.php
+++ b/app/Repositories/QueriesContentful.php
@@ -2,6 +2,7 @@
 
 namespace App\Repositories;
 
+use App\Entities\Entity;
 use App\Entities\HomePage;
 use Contentful\Delivery\Query;
 use App\Entities\TruncatedCampaign;
@@ -43,6 +44,23 @@ trait QueriesContentful
 
                 return $results->toJson();
         }
+    }
+
+    /**
+     * Get a single entry from Contentful by ID.
+     *
+     * @param  string $id
+     * @return mixed
+     */
+    protected function getEntryFromIdAsJson($id)
+    {
+        $entry = app('contentful.delivery')->getEntry($id);
+
+        $entity = new Entity($entry);
+
+        $block = $entity->parseBlock($entity);
+
+        return $block;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


app/Repositories/QueriesContentful.php
```diff
@@ -45,6 +46,23 @@ protected function getEntriesAsJson($type, $options = [])
...
+     * @param  string $id
+     * @return mixed
+     */
+    protected function getEntryFromIdAsJson($id)
+    {
+        $entry = app('contentful.delivery')->getEntry($id);
+
+        $entity = new Entity($entry);
+
+        $block = $entity->parseBlock($entity);
```
weerd: This was a bit weird. Since we don't know what content-type the retrieved entry will be, we need to create a generic new `Entity` and then use the `parseBlock()` method on itself. In hindsight not great, but I think fixing this up is going to likely happen when we reassess a lot of the other quirks with how we originally structured our architecture.

app/Repositories/QueriesContentful.php
```diff
@@ -45,6 +46,23 @@ protected function getEntriesAsJson($type, $options = [])
...
+
+        return $block;
```
weerd: Also, contrary to how we do this in the rest of the trait methods, I don't `json_encode()` the results, since this gets passed to the API `BlocksController`, which itself returns a json response and thus would get doubly encoded. Works fine for now, but may be an issue if aiming to cache these entries, and I'll aim to find a better solution then.

DFurnes: Makes sense! I think we'll definitely want to add caching here before Prom launches.

PR #1272, opened 2019-02-25
Getting the RichText field working on StoryPage.

weerd:
### What does this PR do?

This PR enables basic support for RichText on the StoryPage. It renders out the entry if found, otherwise it renders out a basic "not found" block. We can customize this further in the future.

_Important Note_:
We had to disable running our Jest JavaScript test suite due to an issue where we were running into a circular dependency when running our tests. Everything works fine when our JS is bundled in Webpack, but fails when the Jest test suite runs and uses Node's module loader. We're going to reassess our test, etc post-Prom launch and come back to address this, but needed to disable to unblock progress for now. [Pivotal Tracker Ticket](https://www.pivotaltracker.com/story/show/164282294)

**StoryPage with ImageBlock**:
image

**StoryPage with "not found" block**:
image

**StoryPage with other ImageBlock, using `SectionBlock`s with custom background/text colors**:
image

_Note: I am proud of my section custom color choices and stand by them... we need more of a Joker-esk color palette_ 😏 🃏

It uses a basic `Card` component for the "not found". We could use the `NotFound` component but it needs to be updated to be made more customizeable with different language depending on context.

Refs [Pivotal ID #163396477](https://www.pivotaltracker.com/story/show/163396477)

### Checklist

### DIFF
--- a/resources/assets/components/ContentfulEntry/ContentfulEntry.js
+++ b/resources/assets/components/ContentfulEntry/ContentfulEntry.js
@@ -143,13 +143,7 @@ class ContentfulEntry extends React.Component<Props, State> {
       case 'sectionBlock': {
         const fields = withoutNulls(json.fields);
 
-        return (
-          <SectionBlock
-            key={`section-block-${json.id}`}
-            id={json.id}
-            {...fields}
-          />
-        );
+        return <SectionBlock id={json.id} {...fields} />;
       }
 
       case 'shareAction':
--- a/resources/assets/components/ContentfulEntry/ContentfulEntry.js
+++ b/resources/assets/components/ContentfulEntry/ContentfulEntry.js
@@ -11,6 +11,7 @@ import PollLocator from '../PollLocator/PollLocator';
 import { CampaignUpdateContainer } from '../CampaignUpdate';
 import ImagesBlock from '../blocks/ImagesBlock/ImagesBlock';
 import GalleryBlock from '../blocks/GalleryBlock/GalleryBlock';
+import SectionBlock from '../blocks/SectionBlock/SectionBlock';
 import { parseContentfulType, report, withoutNulls } from '../../helpers';
 import CallToActionContainer from '../CallToAction/CallToActionContainer';
 import LandingPageContainer from '../pages/LandingPage/LandingPageContainer';
@@ -138,6 +139,18 @@ class ContentfulEntry extends React.Component<Props, State> {
 
       case 'referralSubmissionAction':
         return renderReferralSubmissionAction(json);
+
+      case 'sectionBlock': {
+        const fields = withoutNulls(json.fields);
+
+        return (
+          <SectionBlock
+            key={`section-block-${json.id}`}
+            id={json.id}
+            {...fields}
+          />
+        );
+      }
 
       case 'shareAction':
         return renderShareAction(json);
--- a/resources/assets/components/blocks/SectionBlock/SectionBlock.js
+++ b/resources/assets/components/blocks/SectionBlock/SectionBlock.js
@@ -1,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { withoutNulls } from '../../../helpers';
+import TextContent from '../../utilities/TextContent/TextContent';
+
+const SectionBlock = props => {
+  const { id, content, backgroundColor, textColor } = props;
+
+  const styles = {
+    backgroundColor,
+    color: textColor,
+  };
+
+  return (
+    <section id={id} className="story-section" style={withoutNulls(styles)}>
+      <TextContent>{content}</TextContent>
+    </section>
+  );
+};
+
+SectionBlock.propTypes = {
+  backgroundColor: PropTypes.string,
+  content: PropTypes.oneOfType([PropTypes.object]).isRequired,
+  id: PropTypes.string.isRequired,
+  textColor: PropTypes.string,
+};
+
+SectionBlock.defaultProps = {
+  backgroundColor: null,
+  textColor: null,
+};
+
+export default SectionBlock;
--- a/resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
+++ b/resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
@@ -18,18 +18,10 @@ class ContentfulEntryLoader extends React.Component {
   }
 
   componentDidMount() {
-    const request = getBlock(this.props.id);
-
-    request
-      .then(response => {
-        this.setState({
-          entryData: response,
-        });
-      })
+    getBlock(this.props.id)
+      .then(entryData => this.setState({ entryData }))
       .catch(error => {
-        this.setState({
-          hasError: true,
-        });
+        this.setState({ hasError: true });
 
         report(error);
       });
--- a/resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
+++ b/resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
@@ -1,0 +1,59 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { getBlock } from '../../../helpers/api';
+import ContentfulEntry from '../../ContentfulEntry';
+import Card from '../Card/Card';
+import TextContent from '../TextContent/TextContent';
+
+class ContentfulEntryLoader extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      entryData: null,
+      hasError: null,
+    };
+  }
+
+  componentDidMount() {
+    const request = getBlock(this.props.id);
+
+    request
+      .then(response => {
+        this.setState({
+          entryData: response,
+        });
+      })
+      .catch(error => {
+        this.setState({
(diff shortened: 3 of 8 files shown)


resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
```diff
@@ -0,0 +1,59 @@
...
+    super(props);
+
+    this.state = {
+      entryData: null,
+      hasError: null,
+    };
+  }
+
+  componentDidMount() {
+    const request = getBlock(this.props.id);
```
DFurnes: Saving the promise into a variable here feels a little odd to me since we only use it for chaining. What do you think about just directly calling `.then` and `.catch` on the result of `getBlock`?

Also wonder if we can sneak these into one-liners within our Prettier constraints...

```js
getBlock(this.props.id)
    .then(entryData => this.setState({ entryData });
    .catch(error => this.setState({hasError: true});
```

weerd: Yeah, defs. I was referencing approach from PSA, but sounds good to me 👍

resources/assets/components/blocks/SectionBlock/SectionBlock.js
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <section id={id} className="story-section" style={withoutNulls(styles)}>
+      <TextContent>{content}</TextContent>
+    </section>
+  );
+};
+
+SectionBlock.propTypes = {
+  backgroundColor: PropTypes.string,
+  content: PropTypes.oneOfType([PropTypes.object]).isRequired,
```
mendelB: Could this just be `PropTypes.object`, or do we anticipate adding support for more types?

wercker.yml
```diff
@@ -19,9 +19,13 @@ build:
     - script:
         name: run type checker
         code: npm run flow
-    - script:
-        name: test javascript code
-        code: npm run test:ci
```
mendelB: image

resources/assets/components/blocks/SectionBlock/SectionBlock.js
```diff
@@ -0,0 +1,34 @@
...
+const SectionBlock = props => {
+  const { id, content, backgroundColor, textColor } = props;
+
+  const styles = {
+    backgroundColor,
+    color: textColor,
+  };
+
+  return (
+    <section id={id} className="story-section" style={withoutNulls(styles)}>
```
weerd: I did some testing and technically we _don't_ need to use the `withoutNulls()` helper here, since sending through an object with say `color: null` (if no value supplied to the `textColor` field) to the `style` prop doesn't end up getting output, but I kind of felt it was a little cleaner to do so. Lemme know if this is overkill.

resources/assets/components/blocks/SectionBlock/SectionBlock.js
```diff
@@ -0,0 +1,34 @@
...
+
+import { withoutNulls } from '../../../helpers';
+import TextContent from '../../utilities/TextContent/TextContent';
+
+const SectionBlock = props => {
+  const { id, content, backgroundColor, textColor } = props;
+
+  const styles = {
+    backgroundColor,
+    color: textColor,
```
weerd: I also realize I'm going to probably need to expand on this, potentially append some `!important` and maybe target headings specifically. Just wanted to get the ball rolling with these basic style applications.

resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -139,6 +140,18 @@ class ContentfulEntry extends React.Component<Props, State> {
       case 'referralSubmissionAction':
         return renderReferralSubmissionAction(json);
 
+      case 'sectionBlock': {
+        const fields = withoutNulls(json.fields);
+
+        return (
+          <SectionBlock
+            key={`section-block-${json.id}`}
```
mendelB: Does this need a key, if it's being rendered with a key from the [initial iterator](https://github.com/DoSomething/phoenix-next/blob/02fd51ba0f63fe6d99f930a77fcd79cf41f3e2a2/resources/assets/components/pages/StoryPage/StoryPage.js#L15)? (I didn't _think_ I needed to add one to the PetitionSubmissionAction)

resources/assets/helpers/text.js
```diff
@@ -63,9 +66,15 @@ function getMarkdownItInstance() {
  * @return {String}
  */
 export function parseRichTextDocument(document) {
-  // @TODO: more to come here. Stay tuned!
+  const options = {
+    renderNode: {
+      [BLOCKS.EMBEDDED_ENTRY]: node => (
+        <ContentfulEntryLoader id={node.data.target.sys.id} />
```
mendelB: 💥

resources/assets/components/blocks/SectionBlock/SectionBlock.js
mendelB: I don't have objections! This did remind me that I could have used this for far cleaner logic when conditionally adding the `action_id` to RB submissions 🤦‍♂️  https://github.com/DoSomething/phoenix-next/blob/aa6d10d58d7b8f3a1314eab07a98e94ab476a141/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js#L185-L191

mendelB:
> _Note: I am proud of my section custom color choices and stand by them... we need more of a Joker-esk color palette_ 😏 🃏

image

resources/assets/components/blocks/SectionBlock/SectionBlock.js
weerd: I forget about a lot of the helper functions we've added over time. Luckily, refactoring is always an option, haha!

resources/assets/components/blocks/SectionBlock/SectionBlock.js
weerd: ESLint yells at us if we specify `PropTypes.object` alone, so we've found this approach works fine:

```js
24:3  error  Prop type `object` is forbidden  react/forbid-prop-types
```

weerd: One could argue we might want to edit our rules... meh. 🤷‍♂️

wercker.yml
weerd: I literally LOL'd when I scrolled down and saw this image @mendelB 🤣

resources/assets/components/blocks/SectionBlock/SectionBlock.js
DFurnes: I like using `withoutNulls` here! 👍

resources/assets/components/blocks/SectionBlock/SectionBlock.js
mendelB: Ohhhh right I forgot about that! Better then `PropTypes.shape`ing every potential block I suppose 😬

DFurnes: Yeah, I like the intent behind this linter rule but it's been a hard one for us to follow. It's probably better to just disable the rule rather than find clever ways to sneak around it. 😅

wercker.yml
DFurnes: 😂

resources/assets/components/ContentfulEntry/ContentfulEntry.js
weerd: Oh, hmmm... I thought it was yelling at me before about it, but I just tried removing it and wasn't anymore. Probs added the key to the initial iterator after and hadn't realized it fixed so I added here to 🤷‍♂️ Removing it!

wercker.yml
mendelB: lol there's admittedly a lot of buried angst in this meme

resources/assets/components/ContentfulEntry/ContentfulEntry.js
DFurnes: Nope, you only need [keys](https://reactjs.org/docs/lists-and-keys.html#keys) when we have an "array" of components that might change between renders (so yeah, we're okay here because `ContentfulEntry` only returns _one_ element, and the parent adds a key for the dynamic items it loops over where you linked).

wercker.yml
```diff
@@ -19,9 +19,13 @@ build:
...
+    # We are temporarily disabling running our JS test suite since we currently seem to have
+    # an issue with Jest and a circular dependency on ContentfulEntry. We are  planning on
+    # reassessing our tests and fixing things up post our current big initiative with
+    # StoryPages and ungated actions. (2019-02-27)
```
DFurnes: Would it be worth linking the Pivotal ticket here? 🤷‍♂️

resources/assets/components/blocks/SectionBlock/SectionBlock.js
weerd: I can add a ticket and make it a cleanup task. Once we disable the rule, just do a find replace.

wercker.yml
weerd: I linked the PT to the PR but ya, that sounds good. 👌

DFurnes: I know! Just thought it might be a handy thing to have easily available here.

PR #1273, opened 2019-02-25
Petition submission action migration

mendelB:
### What does this PR do?

This PR adds a Contentful content type migration script for the brand new `PetitionSubmissionAction`!

Also adds the Entity and the block parser case.

I was able to extrapolate most of this from the `PhotoSubmissionAction` and the `TextSubmissionAction` which had the fields we needed for this.

For reference, here's the Petition in the mock:
image

### Checklist

### DIFF
--- a/app/Entities/PetitionSubmissionAction.php
+++ b/app/Entities/PetitionSubmissionAction.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace App\Entities;
+
+use JsonSerializable;
+
+class PetitionSubmissionAction extends Entity implements JsonSerializable
+{
+    public function jsonSerialize()
+    {
+        return [
+            'id' => $this->entry->getId(),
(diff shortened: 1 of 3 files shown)


app/Entities/PetitionSubmissionAction.php
```diff
@@ -0,0 +1,27 @@
...
+    public function jsonSerialize()
+    {
+        return [
+            'id' => $this->entry->getId(),
+            'type' => $this->getContentType(),
+            'fields' => [
+                'actionId' => $this->actionId,
+                'title' => $this->title,
+                'content' => $this->content,
+                'textFieldPlaceholder' => $this->textFieldPlaceholder,
```
weerd: no `textFieldLabel`?

weerd: oh wait; there is no label... nevermind.

PR #1276, opened 2019-02-25
Refresh the page if token is expired.

DFurnes:
### What does this PR do?

This pull request adds an event which periodically checks whether or not the user's access token has expired. If it has, it refreshes the page so that the [RefreshTokenMiddleware](https://github.com/DoSomething/phoenix-next/blob/b9e40dc567997ff16b8865878396a5a8bbe1bdc2/app/Http/Kernel.php#L39) will kick in and fetch a new token for the user.

This should hopefully clear up most of the remaining "token expired" errors we see when users leave a tab open for more than an hour. 🤞 I've also been working on an experiment for a more graceful solution to this problem, which I'm hoping to wrap up and share this week.

Refs [Pivotal ID #157665698](https://www.pivotaltracker.com/story/show/157665698).

### Checklist

### DIFF
--- a/resources/assets/helpers/auth.js
+++ b/resources/assets/helpers/auth.js
@@ -1,0 +1,42 @@
+/* global window */
+
+/**
+ * Get the current UNIX timestamp (in seconds), optionally
+ * adjusted by the given clock skew.
+ *
+ * @param {Number} skew
+ * @return {Number}
+ */
+const now = (skew = 0) => Math.floor(+new Date() / 1000) + skew;
+
+/**
+ * Bind an event listener which will refresh the page
+ * if the current access token is expired.
+ */
+export function bindTokenRefreshEvent() {
+  const { token, expiresAt } = window.AUTH;
+
+  // Take into account clock skew between server & user's browser:
+  const skew = window.AUTH.now - now();
+
+  // Refresh the page if our authentication token is expired:
+  const authCheck = setInterval(() => {
+    if (token) {
+      // Read the current time, and adjust by our originally recorded
+      // clock "skew" – so we make sure we're not refreshing until the
+      // server expects this token to be expired.
+      const expiresIn = expiresAt - now(skew);
+
+      console.debug(`Token expires in ${Math.floor(expiresIn)} seconds.`);
+
+      if (expiresIn < 0) {
+        console.log('Token has expired! Refreshing...');
+
+        clearInterval(authCheck);
+        window.location.reload(true);
+      }
+    }
+  }, 5000);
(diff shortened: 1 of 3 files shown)


resources/assets/helpers/auth.js
```diff
@@ -0,0 +1,42 @@
...
+    if (token) {
+      // Read the current time, and adjust by our originally recorded
+      // clock "skew" – so we make sure we're not refreshing until the
+      // server expects this token to be expired.
+      const expiresIn = expiresAt - now(skew);
+
+      console.debug(`Token expires in ${Math.floor(expiresIn)} seconds.`);
+
+      if (expiresIn < 0) {
+        console.log('Token has expired! Refreshing...');
```
DFurnes: I wonder if this is something we want to track with a Puck event? If so, thoughts on naming?

```js
trackAnalyticsEvent({ verb: 'refreshed', noun: 'token'};
```

resources/assets/helpers/auth.js
```diff
@@ -0,0 +1,42 @@
...
+  // Take into account clock skew between server & user's browser:
+  const skew = window.AUTH.now - now();
+
+  // Refresh the page if our authentication token is expired:
+  const authCheck = setInterval(() => {
+    if (token) {
+      // Read the current time, and adjust by our originally recorded
+      // clock "skew" – so we make sure we're not refreshing until the
+      // server expects this token to be expired.
+      const expiresIn = expiresAt - now(skew);
```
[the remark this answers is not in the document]
DFurnes: Yeah, so the concern here is that a user's machine might not have the exact same "time" set as our servers. On most machines, this should be synced up via [NTP](https://en.wikipedia.org/wiki/Network_Time_Protocol) but in case it's not, we compare the server's self-reported time with the user's time.

For example, if a user's clock was 15 minutes ahead, it could result in their browser refreshing the page before the server's "refresh" logic would trigger. I'm not sure how likely this is in practice, but the fear is that someone could otherwise get stuck in a refresh loop since their computer would think the token was expired but our server wouldn't.

DFurnes:
@weerd Responded in a comment above! Let me know if you still have Qs.

weerd:
Ah! Sorry, I thought I refreshed GH and didn't see anything. Must've missed it! 👍

resources/assets/helpers/auth.js
DFurnes: Added an analytics event in b93e724! I also attached `skew` as a data property on the event, since I'm curious to see how much that matters in practice.

PR #1277, opened 2019-02-26
Section block and story page updates

weerd:
### What does this PR do?

This PR adds a migration for the new `sectionBlock` content type and supporting items, as well as updating the `storyPage` content-type migration to reference `sectionBlock` in use.

🌵 

Refs [Pivotal ID #164144958](https://www.pivotaltracker.com/story/show/164144958)

### Checklist

### DIFF
--- a/app/Entities/SectionBlock.php
+++ b/app/Entities/SectionBlock.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace App\Entities;
+
+use JsonSerializable;
+
+class SectionBlock extends Entity implements JsonSerializable
+{
+    public function jsonSerialize()
+    {
+        return [
+            'id' => $this->entry->getId(),
+            'type' => $this->getContentType(),
+            'fields' => [
+                'internalTitle' => $this->internalTitle,
+                'backgroundColor' => $this->backgroundColor,
+                'textColor' => $this->textColor,
+                'content' => $this->content,
+            ],
+        ];
+    }
+}
--- a/contentful/content-types/storyPage.js
+++ b/contentful/content-types/storyPage.js
@@ -57,7 +57,7 @@ module.exports = function(migration) {
         },
 
         message:
-          'Only alphanumeric, forward-slash, and hyphen characters are allowed in slugs! Entry cannot start with a forward-slash.',
+          'Only alphanumeric, forward-slash, and hyphen characters are allowed in slugs! Entry must begin with "stories/".',
       },
     ])
     .disabled(false)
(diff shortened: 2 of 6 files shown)


app/Entities/SectionBlock.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    public function jsonSerialize()
+    {
+        return [
+            'id' => $this->entry->getId(),
+            'type' => $this->getContentType(),
+            'fields' => [
+                'internalTitle' => $this->internalTitle,
+                'backgroundColor' => $this->backgroundColor,
+                'textColor' => $this->textColor,
```
weerd: Debating if maybe popping the style related items into a `styles` item and nesting the values within it?

weerd: Also debating if keeping the field named `textColor` but here we change it to `fontColor`...

weerd: And lastly, we chatted with Luke about having these be distinct fields to better control them... but I think there was also mention of `additionalContent` JSON usage. I kind of like these as distinct fields and feels more straightforward.

mendelB: Do we anticipate many/any more style fields being added?

I do feel like either way though, since it's fairly easy for us to extrapolate distinct fields into a new child entry if necessary down the road, it makes sense to keep these as distinct fields for now.

contentful/content-types/storyPage.js
```diff
@@ -79,38 +79,25 @@ module.exports = function(migration) {
...
 
   storyPage
-    .createField('content')
-    .name('Content')
-    .type('RichText')
+    .createField('blocks')
+    .name('Blocks')
+    .type('Array')
     .localized(false)
     .required(false)
```
mendelB: Should a `sectionBlock` require at least one block to be valid since it has no content of its own?

app/Entities/SectionBlock.php
DFurnes: Yeah, I like distinct fields too since we can validate their format & add good helper text. 👍 Down the road, it could be fun to look into figuring out a way to let us pass something with more flexibility here (like scoped CSS).

weerd: Potentially more fields? But not a lot is what I'm guessing.

I would also add, that while the `additionaContent` is useful on paper, we've been bad about going back and cleaning that "temporary" data/content/configuration out in the past, and has shot us in the foot in a couple instances 😖

mendelB: Hard truths. Though it really is neat on paper! I guess perhaps it's outlived some of it's usefulness now with multiple environments for testing, and with migration scripts for version controlling/documenting field experiments?

contentful/content-types/storyPage.js
weerd: Potentially, but I think it's fine for now, since it's going to be mainly curated by PMs, but definitely something to consider after it's in use for a bit :) Defs good to flag!

PR #1280, opened 2019-02-28
Add new text gallery styling.

DFurnes:
### What does this PR do?

This pull request adds the new text gallery styling, designed for the upcoming "Prom For All" campaign. This will replace the [old appearance](https://user-images.githubusercontent.com/583202/53599983-6d73dc00-3b76-11e9-9ba2-e61da66e9a58.png) for all galleries with text posts ([discussion](https://dosomething.slack.com/archives/C3ASB4204/p1551302391159200)).

screen shot 2019-02-28 at 4 34 04 pm

Some open questions:
- At the moment, we'll only show the state where a post was submitted from if the user is `null` (which for anonymous actions, they would be). It might make sense to set this on the gallery (or expose as a property on the post response) so we don't create confusion for staffers...
- Right now, we hardcode a min-height for these. It might be nice to see if we can update the gallery to support dynamic heights (but consistent within rows). That's a CSS Grid thing, yah? 🎨
- Is there a reason we omitted reactions here? We could add a toggle to the gallery content type, perhaps, but it'd be nice to keep this consistent everywhere if possible...
- Finally, I snagged `#7069d8` from @lkpttn's mocks. Do we have a name for that?

Refs [Pivotal ID #164214415](https://www.pivotaltracker.com/story/show/164214415).

### Checklist

### DIFF
--- a/apollo.config.js
+++ b/apollo.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  client: {
+    service: 'dosomething-graphql-dev-lambda',
+    includes: ['./resources/assets/**/*.js'],
+  },
+};
--- a/resources/assets/components/utilities/PostCard/PostCard.js
+++ b/resources/assets/components/utilities/PostCard/PostCard.js
@@ -20,6 +20,8 @@ export const postCardFragment = gql`
     text
     tags
     quantity
+    location(format: HUMAN_FORMAT)
+
     user {
       firstName
     }
@@ -27,7 +29,11 @@ export const postCardFragment = gql`
 `;
 
 const PostCard = ({ post, noun }) => {
-  const firstName = post.user ? post.user.firstName : 'A Doer';
+  // If this post isn't anonymous, show first name. Otherwise, state or fallback to "A Doer".
+  const displayName = post.user
+    ? post.user.firstName
+    : post.location || 'A Doer';
+
   const reactionElement = isAuthenticated() ? (
     <ReactionButton post={post} />
   ) : null;
@@ -37,13 +43,13 @@ const PostCard = ({ post, noun }) => {
   switch (post.type) {
     case 'text':
       media = (
-        <div className="chat-bubble -post-bubble flex-center-y margin-bottom-none rounded-top">
-          <p className="color-white">{post.text}</p>
+        <div className="chat-bubble -post-bubble margin-bottom-none rounded-top">
+          <p className="font-italic">{post.text}</p>
         </div>
       );
       break;
     case 'photo':
-      media = <LazyImage alt={`${firstName}'s photo`} src={post.url} />;
+      media = <LazyImage alt={`${displayName}'s photo`} src={post.url} />;
       break;
 
     default:
@@ -51,14 +57,14 @@ const PostCard = ({ post, noun }) => {
   }
 
   return (
-    <BaseFigure className="post" media={media}>
+    <BaseFigure className={`post post-${post.type}`} media={media}>
       <BaseFigure
         media={reactionElement}
         alignment="right"
         className="padded margin-bottom-none"
       >
         <h4>
-          {firstName}
+          {displayName}
           <PostBadge status={post.status} tags={post.tags} />
         </h4>
         {post.quantity ? (
--- a/resources/assets/components/utilities/PostCard/post.scss
+++ b/resources/assets/components/utilities/PostCard/post.scss
@@ -16,7 +16,7 @@
   }
 
   > .figure__media {
-    background-color: $med-gray;
+    background-color: $white;
 
     & > img {
       width: 480px; // HACK: Expand smaller images to fit in card.
@@ -32,32 +32,22 @@
   }
 }
 
+.post-text {
+  border-left: $half-spacing solid #7069d8;
+}
+
 .chat-bubble.-post-bubble {
-  font-family: $font-league-gothic;
-  min-height: 340px; // HACK: short text posts should be same height as photos.
-  background-color: $purple;
-
-  &:before {
-    // Offset the 'tail' from bottom with a 1px overlap to avoid
-    // sub-pixel rendering issue w/ slight gap between elements.
-    top: auto;
-    bottom: -19px;
-
-    border-color: $purple transparent transparent $purple;
-    left: 25%;
-  }
-
-  h4 {
-    @include media($only-medium) {
-      font-size: $font-small;
-    }
-  }
+  min-height: 332px; // HACK: short text posts should be same height as photos.
+  color: $black;
+  background-color: $white;
+  padding: $base-spacing $half-spacing;
 
   p {
-    font-size: 42px;
+    color: $black;
+    font-size: $font-medium;
 
     @include media($only-medium) {
-      font-size: $font-medium;
+      font-size: $font-regular;
     }
   }
 }
--- a/resources/assets/components/utilities/PostGallery/post-gallery.scss
+++ b/resources/assets/components/utilities/PostGallery/post-gallery.scss
@@ -1,13 +1,3 @@
 @import '../../../scss/next-toolbox';
 
-.post-gallery {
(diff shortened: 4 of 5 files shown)


DFurnes:
> At the moment, we'll only show the state where a post was submitted from if the user is `null` (which for anonymous actions, they would be). It might make sense to set this on the gallery (or expose as a property on the post response) so we don't create confusion for staffers..

Oh, submission galleries are a good example of why we'd probably want to expose `action` details on the post to toggle how this is displayed (since it might be confusing to a user to see their otherwise anonymous story with their name on it...!)

apollo.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  client: {
+    service: 'dosomething-graphql-dev-lambda',
+    includes: ['./resources/assets/**/*.js'],
```
mendelB: What's this config about?

mendelB:
> > At the moment, we'll only show the state where a post was submitted from if the user is `null` (which for anonymous actions, they would be). It might make sense to set this on the gallery (or expose as a property on the post response) so we don't create confusion for staffers..
> 
> Oh, submission galleries are a good example of why we'd probably want to expose `action` details on the post to toggle how this is displayed (since it might be confusing to a user to see their otherwise anonymous story with their name on it...!)

I think I totally agree with this! I know just from testing anonymous posts in the gallery, @chloealee and I were confused by the difference in display when we tested with our admin accounts vs with general test accounts.

> Right now, we hardcode a min-height for these. It might be nice to see if we can update the gallery to support dynamic heights (but consistent within rows). That's a CSS Grid thing, yah? 🎨

heck ye! I _think_ this is the [`grid-template-rows`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-rows) rule.

> Is there a reason we omitted reactions here? We could add a toggle to the gallery content type, perhaps, but it'd be nice to keep this consistent everywhere if possible..

➕

> Finally, I snagged `#7069d8` from @lkpttn's mocks. Do we have a name for that?

`lukurple` (didn't you get the memo?)



Just wanna shout out (maybe an obvious point) that as we inch towards custom gallery specs, we'll need to figure out a solution to deal with the current pattern of using the same `customBlock` across multiple campaigns. 
Though maybe this isn't really an issue, since for the most part these are simple Photo galleries?

apollo.config.js
DFurnes: This is used to configure the [Apollo Editor extension](https://www.apollographql.com/docs/platform/editor-plugins.html)! It tells the editor to use our "development" space for schema & performance hints (since that's where we're connected to in development), and manually "includes" our source files (since by default it looks in `./src/**/*.js`).

weerd:
> It might make sense to set this on the gallery (or expose as a property on the post response) so we don't create confusion for staffers...

I think I'm still a little confused by this statement. Setting _what_ on the gallery? Is this a setting in the proposed gallery content type or something on the actual gallery we are setting?

weerd:
> Just wanna shout out (maybe an obvious point) that as we inch towards custom gallery specs, we'll need to figure out a solution to deal with the current pattern of using the same `customBlock` across multiple campaigns. Though maybe this isn't really an issue, since for the most part these are simple Photo galleries?

We should formalize a submission gallery content type!!! We've been wanting to do it for a while. Is now the time?!? @DFurnes

resources/assets/components/utilities/PostCard/PostCard.js
```diff
@@ -20,14 +20,20 @@ export const postCardFragment = gql`
...
     }
   }
 `;
 
 const PostCard = ({ post, noun }) => {
-  const firstName = post.user ? post.user.firstName : 'A Doer';
+  // If this post isn't anonymous, show first name. Otherwise, state or fallback to "A Doer".
+  const displayName = post.user
+    ? post.user.firstName
+    : post.location || 'A Doer';
```
weerd: Or `'Somewhere, USA'` 😏

resources/assets/components/utilities/PostCard/PostCard.js
```diff
@@ -20,14 +20,20 @@ export const postCardFragment = gql`
...
     user {
       firstName
     }
   }
 `;
 
 const PostCard = ({ post, noun }) => {
-  const firstName = post.user ? post.user.firstName : 'A Doer';
+  // If this post isn't anonymous, show first name. Otherwise, state or fallback to "A Doer".
+  const displayName = post.user
```
weerd: totally not necessary, but maybe rename to `displayIdentifier`? so it's more generic and applies to either a name or a location? 🤔

apollo.config.js
mendelB: Ah so sublime! (see what I did there 😈) 
I guess we'll have to be sure to use VS code when working with Graphql

weerd: fozzy bear

DFurnes:
> I think I'm still a little confused by this statement. Setting _what_ on the gallery? Is this a setting in the proposed gallery content type or something on the actual gallery we are setting?

Oops, that was poorly phrased! The "what" is whether or not a post is made for an anonymous action. Since post owners & staffers have full "view all" privileges, they wouldn't have any indication that not _everyone_ sees the same authenticated response as they do.

resources/assets/components/utilities/PostCard/PostCard.js
DFurnes: Hm, `identifier` makes me think "ID"... but I totally get whatcha mean. Are there any other synonyms that might work better here? 🤔

resources/assets/components/utilities/PostCard/PostCard.js
DFurnes: Haha! 🥇

DFurnes:
> We should formalize a submission gallery content type!!! We've been wanting to do it for a while. Is now the time?!? @DFurnes

I'm down! Campaign gallery, though, right?

resources/assets/components/utilities/PostCard/PostCard.js
weerd: True... Maybe just `postLabel`?

DFurnes: Yeah, that works… let's keep mulling it over and I can swap in my next PR!

weerd:
> I'm down! Campaign gallery, though, right?

ya! `campaignGallery`, although with the idea of having galleries on a story page, etc... maybe something more like `postGallery` would make more sense? 

Also wondering if we do a `userSubmissionGallery` but I kind of feel those should just automatically always get included on Actions and not a content type instance an editor has to include on an experience. It should just appear when a user submits their first post RB or if they return to page and see all their previous posts.

DFurnes:
Yeah haha, I've always gotten those names mixed up. 😂 Agreed that "Post Gallery" makes more sense than "Campaign Gallery" (especially since these may often show just a particular type of post, or particular action, rather than a full campaign).

> Also wondering if we do a `userSubmissionGallery` but I kind of feel those should just automatically always get included on Actions and not a content type instance an editor has to include on an experience. It should just appear when a user submits their first post RB or if they return to page and see all their previous posts.

Yeah, I agree with trying to keep those bundled to the action blocks for now! 👍

ngjo:
> Is there a reason we omitted reactions here? We could add a toggle to the gallery content type, perhaps, but it'd be nice to keep this consistent everywhere if possible...

I don't 100% remember why we omitted the reactions here, but if I had to guess it might be three-fold:

1. Assumption that it'd add complexity to the prom page for unauthenticated folks
2. It might come off as insensitive with the "like this horrible story"
3. Possibly a stakeholder request that came up when reviewing wireframes

[Adding this to the "check with Luke" list](https://docs.google.com/spreadsheets/d/1lw6416L7N26SpyEMgjw5PefbcD7aMaaCE9SZKXwOnFY/edit#gid=1189638205)

DFurnes:
Alright cool! Let's leave 'em for now, and can always add a "toggle" setting later if necessary.

PR #1281, opened 2019-03-01
CSS grid on story page.

weerd:
### What does this PR do?

This PR updates the `StoryPage` and `SectionBlock` with a bunch of frontend structure and styling using CSS Grid to get it closer to the provided mocks. It also makes sure to use the specified background and text colors that an editor provides for a `SectionBlock` on the CMS within the output on the rendered page. To ensure that the specified color overrides the stylesheet color I ended up needing to add inline styles to the header tag, and utilizing Contentful's RichText parser package helper to insert the inline `style` attribute; turned out working pretty well! 🎉 

 - [Small](https://user-images.githubusercontent.com/105849/53666906-6fed3900-3c3d-11e9-8ef9-7105abadbd01.png)
 - [Medium](https://user-images.githubusercontent.com/105849/53666942-88f5ea00-3c3d-11e9-85bc-3be1272dc74c.png)
 - [Large](https://user-images.githubusercontent.com/105849/53666986-a62ab880-3c3d-11e9-9290-d2f2ce1f907e.png)

Refs [Pivotal ID #164092381](https://www.pivotaltracker.com/story/show/164092381)

### Checklist

### DIFF
--- a/resources/assets/components/pages/StoryPage/story-page.scss
+++ b/resources/assets/components/pages/StoryPage/story-page.scss
@@ -4,6 +4,18 @@
   .lede-banner {
     min-height: 500px;
 
+    > .wrapper {
+      grid-column: 1 / -1;
+
+      @include media($medium) {
+        grid-column: 2 / -2;
+      }
+
+      @include media($large) {
+        grid-column: 3 / -3;
+      }
+    }
+
     @include media($medium) {
       min-height: 610px;
     }
--- a/resources/assets/components/pages/StoryPage/story-page.scss
+++ b/resources/assets/components/pages/StoryPage/story-page.scss
@@ -2,10 +2,12 @@
 
 .story-page {
   .lede-banner {
+    align-items: center;
     min-height: 500px;
 
     > .wrapper {
       grid-column: 1 / -1;
+      padding: 0 $half-spacing;
 
       @include media($medium) {
         grid-column: 2 / -2;
@@ -25,6 +27,17 @@
       font-family: $secondary-font-family;
       font-size: 97px;
       font-weight: normal;
+      line-height: 1;
+      margin: $half-spacing 0;
+
+      @include media($medium) {
+        font-size: 123px;
+      }
+    }
+
+    .lede-banner__headline-subtitle {
+      font-size: 30px;
+      margin: $half-spacing 0;
     }
   }
 }
--- a/resources/assets/components/utilities/TextContent/RichTextDocument.js
+++ b/resources/assets/components/utilities/TextContent/RichTextDocument.js
@@ -10,21 +10,24 @@ import { parseRichTextDocument } from '../../../helpers/text';
  *
  * @param  {String} options.className
  * @param  {Object} options.children
+ * @param  {Object} options.styles
  * @return {Object}
  */
-const RichTextDocument = ({ className = null, children }) => (
+const RichTextDocument = ({ className = null, children, styles }) => (
   <div className={classnames('markdown rich-text', 'with-lists', className)}>
-    {parseRichTextDocument(children)}
+    {parseRichTextDocument(children, styles)}
   </div>
 );
 
 RichTextDocument.propTypes = {
   children: PropTypes.oneOfType([PropTypes.object]).isRequired,
   className: PropTypes.string,
+  styles: PropTypes.object, // eslint-disable-line react/forbid-prop-types
 };
 
 RichTextDocument.defaultProps = {
   className: null,
+  styles: {},
 };
 
 export default RichTextDocument;
--- a/resources/assets/components/utilities/TextContent/RichTextDocument.js
+++ b/resources/assets/components/utilities/TextContent/RichTextDocument.js
@@ -14,7 +14,7 @@ import { parseRichTextDocument } from '../../../helpers/text';
  * @return {Object}
  */
 const RichTextDocument = ({ className = null, children, styles }) => (
-  <div className={classnames('markdown rich-text', 'with-lists', className)}>
+  <div className={classnames('markdown rich-text', className)}>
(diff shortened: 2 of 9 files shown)


resources/assets/helpers/text.js
```diff
@@ -63,14 +64,41 @@ function getMarkdownItInstance() {
  * Parse RichText Document to React components.
  *
  * @param  {Object} document
+ * @param  {Object} styles
  * @return {String}
  */
-export function parseRichTextDocument(document) {
+export function parseRichTextDocument(document, styles) {
+  const textColor = pick(styles, 'color');
```
weerd: Only picking out the `color` info for now, but I could see eventually expanding this. But we should be careful to not make it too complex!

resources/assets/components/utilities/TextContent/RichTextDocument.js
```diff
@@ -10,21 +10,24 @@ import { parseRichTextDocument } from '../../../helpers/text';
...
  * @param  {String} options.className
  * @param  {Object} options.children
+ * @param  {Object} options.styles
  * @return {Object}
  */
-const RichTextDocument = ({ className = null, children }) => (
-  <div className={classnames('markdown rich-text', 'with-lists', className)}>
-    {parseRichTextDocument(children)}
+const RichTextDocument = ({ className = null, children, styles }) => (
+  <div className={classnames('markdown rich-text', className)}>
```
weerd: I needed to remove `with-lists` because it was awkwardly affecting components that use `<ul>`s to build out their structure... Since these components are getting rendered inside of a RichText field, the `.with-lists ul` style rules were mucking it up!

weerd: On another note, I had made a `rich-text` class that I was thinking could be like the `markdown` but only apply to RichText markup.

However, I also had left `markdown` in there and it for the most part is styling most of the markup in the RichText field. I feel we could potentially combine the two and just call it `text-content` and the styles it addresses for markup applies to both Markdown content and RichText content. Thoughts? Would be a bit of reworking for a later date.

Alternatively, we could have `text-content` apply generally to both and keep `markdown` and `rich-text` for any specific styling that needs to target just those. Also might not matter as we make a move to utility classes... (🤔 out loud)

resources/assets/components/pages/StoryPage/story-page.scss
```diff
@@ -2,8 +2,22 @@
 
 .story-page {
   .lede-banner {
+    align-items: center;
```
weerd: This works great! Except in Luke's mock he has the social share buttons show up, and they align differently depending on device size. So this might be something I need use Flexbox for in the lede-banner wrapper.

resources/assets/components/utilities/TextContent/RichTextDocument.js
mendelB: this is a sweet thought progression. I def agree through the final sentiment. Though I'm curious as to whether or why we'll need disparity between regular `markdown` and `rich-text`?

weerd: Yeah, not really sure if we will. Would prefer to just change both to a single class and then just see how well that works out at first.

mendelB:
Ayy @weerd two quick Qs that I didn't insert properly in the review:
1. What's the deal with those `package.lock` changes? Just innocent resolving updates? (lol I see the shrug emoji in the commit message) I'm just always confused about these diffs.
2. I definitely get that the `style` props being passed along through to the `
parseRichTextDocument` is necessary -as you aptly described- "To ensure that the specified color overrides the stylesheet color", that being said though, are we at all concerned with them being like deeply nested props? Or is this just the most elegant way to go about this? (this might be a perfectly superfluous point, but it just stood out to me)

weerd:
> 1. What's the deal with those `package.lock` changes? Just innocent resolving updates? (lol I see the shrug emoji in the commit message) I'm just always confused about these diffs.

Yeah, not sure what was up. I switched to my iMac while working at home, did an `npm install` and it should have referenced the lock file, but there were edits to some of the package's `resolved` properties 🤷‍♂️

weerd:
> 2\. I definitely get that the `style` props being passed along through to the ` parseRichTextDocument` is necessary -as you aptly described- "To ensure that the specified color overrides the stylesheet color", that being said though, are we at all concerned with them being like deeply nested props? Or is this just the most elegant way to go about this? (this might be a perfectly superfluous point, but it just stood out to me)

Potentially. There could definitely be room for refactoring this. It was just the best option I found at the time since we're using the parent `TextContent` component to decide which type to render and then from there the respective component called needs to do their own rendering via calling the helper methods.

resources/assets/components/utilities/TextContent/RichTextDocument.js
DFurnes: Ahhh, yeah that's no good! Is this something we could fix by attaching a `.list` class to the `BLOCKS.UL_LIST` and `BLOCKS.OL_LIST` elements?

DFurnes:
> Yeah, not sure what was up. I switched to my iMac while working at home, did an npm install and it should have referenced the lock file, but there were edits to some of the package's resolved properties 🤷‍♂️

Yeah, these are frustrating to get. I think it's from mis-matched `npm` versions?

resources/assets/components/utilities/TextContent/RichTextDocument.js
DFurnes: Oh I see you did that! 🎉

weerd:
> Yeah, these are frustrating to get. I think it's from mis-matched npm versions?

Yeah, I partly suspected that since there's likely a mismatch from my laptop to my iMac 😬

PR #1285, opened 2019-03-01
Differentiate anonymous posts & source nouns from Rogue.

DFurnes:
### What does this PR do?
This pull request makes it easier to tell when a post is for an "anonymous" action (since previously, post owners would still see their name & staffers would see for all posts). It also sources `noun` from Rogue actions, since that's the source of truth!

For review, here are some posts I've made. From left to right – anonymous without a location (uploaded my local, so no Fastly magic), anonymous with location & a photo post with quantity:

screen shot 2019-03-01 at 4 45 42 pm

We don't yet expose a convenient way to pluralize the noun, so it's possible to end up with `1 things`. We can make a card to revisit this (and return a properly pluralized version of this on the `Post`).

Refs [Pivotal ID #164214415](https://www.pivotaltracker.com/story/show/164214415)

### Checklist

### DIFF
--- a/resources/assets/components/utilities/PostCard/PostCard.js
+++ b/resources/assets/components/utilities/PostCard/PostCard.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { get } from 'lodash';
 import gql from 'graphql-tag';
 import { propType } from 'graphql-anywhere';
 
@@ -36,8 +37,8 @@ export const postCardFragment = gql`
 const PostCard = ({ post }) => {
   // For anonymous posts, label with state if available. Otherwise, the user's first name.
   const authorLabel = post.actionDetails.anonymous
-    ? post.location || 'Anonymous'
-    : post.user.firstName || 'A Doer';
+    ? get(post, 'location', 'Anonymous')
+    : get(post, 'user.firstName', 'A Doer');
 
   const reactionElement = isAuthenticated() ? (
     <ReactionButton post={post} />
--- a/resources/assets/components/utilities/PostCard/PostCard.js
+++ b/resources/assets/components/utilities/PostCard/PostCard.js
@@ -38,7 +38,7 @@ const PostCard = ({ post }) => {
   // If this post is for an anonymous action, label it with the state (if available).
   // For non-anonymous posts (default), label with the user's first name.
   const authorLabel = post.actionDetails.anonymous
-    ? get(post, 'location', 'Anonymous')
+    ? post.location || 'Anonymous'
     : get(post, 'user.firstName', 'A Doer');
 
   const reactionElement = isAuthenticated() ? (
--- a/resources/assets/components/utilities/PostCard/PostCard.js
+++ b/resources/assets/components/utilities/PostCard/PostCard.js
@@ -21,6 +21,11 @@ export const postCardFragment = gql`
     tags
     quantity
     location(format: HUMAN_FORMAT)
+    signupId
+
+    actionDetails {
+      anonymous
+    }
 
     user {
       firstName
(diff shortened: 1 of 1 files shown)


resources/assets/components/utilities/PostCard/PostCard.js
```diff
@@ -21,18 +20,24 @@ export const postCardFragment = gql`
...
-const PostCard = ({ post, noun }) => {
-  // If this post isn't anonymous, show first name. Otherwise, state or fallback to "A Doer".
-  const displayName = post.user
-    ? post.user.firstName
-    : post.location || 'A Doer';
+const PostCard = ({ post }) => {
+  // For anonymous posts, label with state if available. Otherwise, the user's first name.
+  const authorLabel = post.actionDetails.anonymous
+    ? post.location || 'Anonymous'
+    : post.user.firstName || 'A Doer';
```
mendelB: LMK if this is just a ridiculous edge case, but could we potentially have an issue with `post.user.firstname` if the user is deleted or missing? (or do we delete all user's posts when deleting the user?)

DFurnes: You're right, we should probably use `get()` here! I (think?) our process for deleting users is to anonymize the user's profile, signups, posts, etc. but leave the actual records intact for auditing & maintaining overall data integrity.

DFurnes: Updated in 171b6e5! This also gives us a good fallback if Northstar was temporarily offline, so the rest of the gallery could continue to function!

resources/assets/components/utilities/PostCard/PostCard.js
```diff
@@ -21,18 +21,24 @@ export const postCardFragment = gql`
...
   }
 `;
 
-const PostCard = ({ post, noun }) => {
-  // If this post isn't anonymous, show first name. Otherwise, state or fallback to "A Doer".
-  const displayName = post.user
-    ? post.user.firstName
-    : post.location || 'A Doer';
+const PostCard = ({ post }) => {
+  // For anonymous posts, label with state if available. Otherwise, the user's first name.
```
weerd: This comment reads a little funky, maybe rephrase to "Otherwise if not anonymous, use the user's first name or a default."

weerd: I was reading it as it would use the user's first name if the state wasn't available. Granted there's a period there, but still took me a little to realize what you meant 🤔

DFurnes: Yeah haha, I was hoping to squeeze it into one line. 😂 Reworded in 0bb5956.

resources/assets/components/utilities/PostCard/PostCard.js
```diff
@@ -21,25 +21,32 @@ export const postCardFragment = gql`
...
 `;
 
-const PostCard = ({ post, noun }) => {
-  // If this post isn't anonymous, show first name. Otherwise, state or fallback to "A Doer".
-  const displayName = post.user
-    ? post.user.firstName
-    : post.location || 'A Doer';
+const PostCard = ({ post }) => {
+  // If this post is for an anonymous action, label it with the state (if available).
+  // For non-anonymous posts (default), label with the user's first name.
```
weerd: 👌

DFurnes:
Added one last little fix in 68d1bd6. I had hoped to use [`get`](https://lodash.com/docs/4.17.11#get) for both `location` and `user.firstName` (for consistency), but forgot that `get` only triggers the fallback on `undefined`!

PR #1287, opened 2019-03-04
Text Submission Action Updates

mendelB:
### What does this PR do?

This PR updates the `TextSubmissionAction` in a couple of ways:
- new `informationTitle` and `informationContent` fields for displaying an optional 'Information block'
- updated footnotes. replacing current footnote with `500 character limit` messaging, and moving the 'disclaimer' footnote to the bottom of the action.

This is in the spirit of the new design for Text actions, specifically towards supporting the new 'Story' mocks, and generally to support the increased character limit, and parity with the Photo and Petition Action's information blocks.

I did not add the styling to have the information block appear to the side of the action on larger screens in the spirit of the Petition block as well, pending the Story Page.

❓ I did want to confirm the approach here to follow up on our [Slack discussion](https://dosomething.slack.com/archives/C2BPA7M8F/p1551715898025000?thread_ts=1551711619.018700&cid=C2BPA7M8F):
Actions have a disclaimer footnote within the action block, which will hopefully soon be dynamic based on the Action details. They'll additionally have the optional editor-editable Information Block for added context. 
Am I understanding this correctly? And do we want to implement this approach for Photo and Petition actions?

**Before**
image

**After**
image

**With Information Content**
image

Refs [Pivotal ID #164287413](https://www.pivotaltracker.com/story/show/164287413)
[Mocks](https://share.goabstract.com/e49949de-48f3-4772-8523-91c3fce110f1)
### Checklist

### DIFF
--- a/contentful/content-types/textSubmissionAction.js
+++ b/contentful/content-types/textSubmissionAction.js
@@ -80,6 +80,26 @@ module.exports = function(migration) {
     .omitted(false);
 
   textSubmissionAction
+    .createField('informationTitle')
+    .name('Information Title')
+    .type('Symbol')
+    .localized(true)
+    .required(false)
+    .validations([])
+    .disabled(false)
+    .omitted(false);
+
+  textSubmissionAction
+    .createField('informationContent')
+    .name('Information Content')
+    .type('Text')
+    .localized(true)
+    .required(false)
+    .validations([])
+    .disabled(false)
+    .omitted(false);
+
+  textSubmissionAction
     .createField('affirmationContent')
     .name('Affirmation Content')
     .type('Text')
@@ -120,6 +140,16 @@ module.exports = function(migration) {
         'The placeholder for the "text" field, providing an example of what a text submission should look like.',
     },
   );
+
+  textSubmissionAction.changeEditorInterface('informationTitle', 'singleLine', {
+    helpText:
+      'Title to display for the information block (defaults to "More Info").',
+  });
+
+  textSubmissionAction.changeEditorInterface('informationContent', 'markdown', {
+    helpText:
+      'Content to display for the information block (this block will be hidden by default unless content is provided).',
+  });
 
   textSubmissionAction.changeEditorInterface('buttonText', 'singleLine', {});
 
--- a/app/Entities/TextSubmissionAction.php
+++ b/app/Entities/TextSubmissionAction.php
@@ -22,6 +22,8 @@ class TextSubmissionAction extends Entity implements JsonSerializable
                 'textFieldLabel' => $this->textFieldLabel,
                 'textFieldPlaceholder' => $this->textFieldPlaceholder,
                 'buttonText' => $this->buttonText,
+                'informationTitle' => $this->informationTitle,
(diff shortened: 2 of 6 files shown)


weerd:
So I think we also want a `disclaimer` field on the content type.

I think this whole bit of text in screenshot below, makes sense to be able to customize:
image

If the editor leaves the field empty, then we can use the action type to infer something. 

My worry it that in the scenario you presented in the screenshot:
image

A user might read the part about it being in the public gallery and entirely stop there. Not continue to then read in the more info block about how it'll be anonymous.

So in that regard, I feel the More Info is context on an action, but disclaimer is all the legal language and shouldn't separate them between the two.

mendelB:
> My worry it that in the scenario you presented in the screenshot:
> image
> 
> A user might read the part about it being in the public gallery and entirely stop there. Not continue to then read in the more info block about how it'll be anonymous.

Lolll yes, this was a _terrible_ example to illustrate how the flow should look! The disclaimer text, in this case, should be the 'anonymous' messaging, and the More Info bar might just give some context re the map.

That being said though:

> So in that regard, I feel the More Info is context on an action (optional to use by editor), but disclaimer is all the legal language and shouldn't separate them between the two (can be overridden to be more specific to legal text for the action or defaults to something based on the action type) but is not optional whether we show it or not.

^ I feel like this could be a nice compromise to [@DFurnes and @ngjo points](https://dosomething.slack.com/archives/C2BPA7M8F/p1551722944046300?thread_ts=1551711619.018700&cid=C2BPA7M8F) re this not being edited often and the importance of having the correct legal messaging. This way, we always have the disclaimer tailored to the action, but, if necessary, have a formal way to override? (We can be very clear in the helper text regarding the purpose and importance of this `disclaimer` field).

contentful/content-types/textSubmissionAction.js
```diff
@@ -121,6 +141,16 @@ module.exports = function(migration) {
     },
   );
 
+  textSubmissionAction.changeEditorInterface('informationTitle', 'singleLine', {
```
DFurnes: Do we have a good example where "More Info" wouldn't work? I'm having a hard time thinking of one.

mendelB: Hmm, that's a good point! A query on our `master` environment against all `photoSubmissionAction`s shows that none of the entries have this field filled out. I can definitely understand if we want to remove this customization ability.

mendelB: @weerd do you have more context on why we'd need this field to be customizable back from the PSA work?

weerd: We did when we implemented it; I had defaulted it to "more info" but definitely was something that was desired to be customize-able. If it's something not used as much, it might just be in how that our resource docs for using this aren't super clear.

 This is meant to be an informative section that editors can provide context to things as they would prefer, so I don't really see it being a bad thing to have editable.

mendelB: Gotcha, ok, so I'll merge this in with the field! (We giveth, and we can always taketh away if need be 😀)

PR #1289, opened 2019-03-04
Add "top-level" block route for previewing & testing.

DFurnes:
### What does this PR do?

This adds a new top-level `/us/blocks/:id` route that fetches & renders a given block with `ContentfulEntryLoader`. It's intended as a handy way to test a particular block in isolation during development & could also be used to link directly to blocks for the "preview" button.

Here's an example with my fav campaign update block. I've also tested that this [does not break the old `campaigns/:slug/blocks/:id` route](https://user-images.githubusercontent.com/583202/53767523-15094b00-3ea4-11e9-83d8-e9f80865790b.png) (although I'm not sure that's even still in use).

screen shot 2019-03-04 at 5 35 40 pm

I whipped this up while working on the new Post Gallery block, since I needed an easy place to test it out while putting things together (and didn't want to mess with other folks by adding it to a page that we all use).

N/A

### Checklist


resources/assets/components/pages/BlockPage/index.js
```diff
@@ -1 +1 @@
-export default from './BlockPageContainer';
+export default from './BlockPage';
```
weerd: Do we want to just _not_ use the whole `index.js` file here and import the component directly `..../BlockPage/BlockPage`?

DFurnes: We do. I'd just kept this since it already existed.

weerd:
> Here's an example with my fav campaign update block.

This block is definitely not used nearly as much as it should be!

resources/assets/components/pages/BlockPage/index.js
weerd: Oh haha, I misread the import. 😖 Cleanup task then :)

mendelB:
Wait, this is like waaaaaay cool! My mind is spinning.
image

PR #1292, opened 2019-03-05
Campaign Independent Text Submission Action Posts

mendelB:
### What does this PR do?

This PR sets the `TextSubmissionAction` component free of its Campaign parental clutches, allowing it to be embedded anywhere, attempting to use its Action ID with our Campaign independent posting method to soar toward its glorious Rogue destiny 👼 

(https://github.com/DoSomething/phoenix-next/pull/1292/commits/80624636ac681ab957faa99ddc689bffe2ed75c0 snuck a little update in to ensure our Lazy Auth™ will work on actions outside of a campaign. The [`getCampaignDataForNorthstar`](https://github.com/DoSomething/phoenix-next/blob/80624636ac681ab957faa99ddc689bffe2ed75c0/resources/assets/selectors/campaign.js#L21) was always assuming a `campaign` was within `state` since we [`default`](https://github.com/DoSomething/phoenix-next/blob/80624636ac681ab957faa99ddc689bffe2ed75c0/resources/assets/store/initialState.js#L12)that property to an empty object. I copied over the logic used in the [`analytics` middleware](https://github.com/DoSomething/phoenix-next/blob/80624636ac681ab957faa99ddc689bffe2ed75c0/resources/assets/middleware/analytics.js#L15) and it now works 👌).

Testing this with the #1289 Block Route was a dream! This should work across the board, and will fall back to the Campaign ID for legacy actions which don't have that field filled out.

Refs [Pivotal ID #163728507](https://www.pivotaltracker.com/story/show/163728507)

### Checklist

### DIFF
--- a/resources/assets/selectors/campaign.js
+++ b/resources/assets/selectors/campaign.js
@@ -21,7 +21,7 @@ export function getCampaignAdditionalContent(state) {
 export function getCampaignDataForNorthstar(state) {
   const data = {};
 
-  if (state.campaign) {
+  if (get(state, 'campaign.type', null) === 'campaign') {
     data.callToAction = state.campaign.callToAction;
     data.coverImage = contentfulImageUrl(
       state.campaign.coverImage.url,
--- a/resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
+++ b/resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
@@ -80,9 +80,11 @@ class TextSubmissionAction extends React.Component {
 
     // Send request to store the text submission post.
     // (Use campaign ID independant post method if actionId is provided).
-    this.props.actionId
-      ? this.props.storePost(data)
-      : this.props.storeCampaignPost(this.props.campaignId, data);
+    if (this.props.actionId) {
+      this.props.storePost(data);
+    } else {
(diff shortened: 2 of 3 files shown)


resources/assets/selectors/campaign.js
```diff
@@ -21,7 +21,7 @@ export function getCampaignAdditionalContent(state) {
 export function getCampaignDataForNorthstar(state) {
   const data = {};
 
-  if (state.campaign) {
+  if (get(state, 'campaign.type', null) === 'campaign') {
```
mendelB: Seems like most of these fields would work with Story Pages too! I guess we can add that logic down the line if that proves necessary so that the signup experience can be tied to the Story Page.

DFurnes: Oh yeah, good flag! 🎏 I bet folks will definitely ask for this.

mendelB: I'll add a tentative row to the post-launch doc!

ngjo: Added here: https://docs.google.com/spreadsheets/d/1lw6416L7N26SpyEMgjw5PefbcD7aMaaCE9SZKXwOnFY/edit#gid=300314040

PR #1293, opened 2019-03-07
Phoenix admin dashboard revamp

weerd:
### What does this PR do?

This PR revamps the Phoenix Admin Dashboard™ as a dropdown that exists above the chrome of the page and thus doesn't interfere with the page design. This current iteration allows the page to be scrolled, but the dashboard is still accessible, by clicking on the toggle attached to the top of the browser window viewport.

We needed the ability to clear cache on Story Pages, and this provided an opportunity to attempt to fix the Admin Dashboard and move towards something more sustainable.

I tried to break up the commits in some logical order, so it might be useful to [go through those in order](https://github.com/DoSomething/phoenix-next/pull/1293/commits). 🤷‍♂️ 

To currently enable this admin dashboard and clear cache on a page, all that is needed is to pass a `$cacheUrl` variable when returning the page view in the respective controller. There is also now a new helper function `get_cache_url()` which will easily build it for you. 

It also provides a `get_cache_id()` (which the `get_cache_url()` method utilizes), to unify in one function how we build the Redis cache ID keys so we can ensure they're all created consistently the same. The prior approach was trickier and more prone to break due to human error when updating, since the "clear cache" button and thus the ID it referred to was generated in a JS React component, but the cache ID for Redis is generated in PHP on the server.

phoenix admin dashboard with scroll

When we originally added the Admin Dashboard, it was meant to be a hacky fix to allow toggling a few things, but has since continued to grow in an unsustainable and poorly-designed fashion. So this is an initial step towards making it better!™

With this approach we should be able to be more flexible in combining predefined settings when the page loads as well as injecting settings from specific components via `React.Portal`, so we get the best of both worlds.

I don't expect the "clear cache" button to remain as is in the future (for now I just copied the button as we had it in the old dashboard), but I could see this dashboard getting fleshed out with lots of useful information, settings, and even potential page-specific metrics like inserting Sixpack test data to visualize the progress of a test (charts/graphs) directly on the page instead of needing to switch to the Sixpack admin interface. There lots we can do!™

Refs [Pivotal ID #162482793](https://www.pivotaltracker.com/story/show/162482793)

### Checklist


app/Http/Controllers/CategorizedPageController.php
```diff
@@ -33,12 +33,15 @@ public function __construct(PageRepository $pageRepository)
      */
     public function show($category, $slug)
     {
-        $page = $this->pageRepository->findBySlug(get_content_type_by_category($category), $category.'/'.$slug);
+        $contentType = get_content_type_by_category($category);
+
+        $page = $this->pageRepository->findBySlug($contentType, $category.'/'.$slug);
```
weerd: I initial split out the `$contentType` variable because I thought I might use it in the `get_cache_url()` (ended up not making sense), but I left it split out as a separate variable because I thought it read a little better and seemed a bit cleaner when used in the `findBySlug()`. If this seems unnecessary, I can restore to prior approach.

resources/assets/components/pages/GeneralPage/GeneralPage.js
```diff
@@ -36,10 +34,6 @@ const GeneralPage = props => {
 
   return (
     <div>
-      <AdminDashboardContainer>
-        <PageDashboardContainer />
-      </AdminDashboardContainer>
```
weerd: 👋

resources/assets/helpers/index.js
```diff
@@ -773,3 +773,23 @@ export function withoutUndefined(data) {
...
+
+/**
+ * Toggle the specified class on the given target element
+ * when the button element is clicked or touched.
+ *
+ * @param  {Element} button
+ * @param  {Element} target
+ * @param  {String} toggleClass
+ */
+export function toggleHandler(button, target, toggleClass) {
```
weerd: I actually stole this code from the `/helpers/navigation.js` file, but since I needed it for the admin dashboard, I put it in the `helpers/index.js` file and it can now just be imported in both. I didn't make the switch to remove from `navigation.js` and import instead since this PR was already growing to over 10 file changes, so stay tuned for an upcoming PR with some cleanups.

resources/views/partials/admin-dashboard.blade.php
```diff
@@ -0,0 +1,6 @@
+<div id="admin-dashboard" class="admin-dashboard">
+    <div class="wrapper">
+        <a class="button -secondary" href="{{ $cacheUrl }}">Clear Cache</a>
+        <a id="js-admin-dashboard-toggle" class="toggle">@include('svg.bug-icon', ['class' => 'icon icon-bug'])</a>
```
weerd: This is a neat trick I learned from Adam Wathan when using SVG tags directly in html. You can save the svg as a `blade.php` file and then just `@include` it and even pass in custom classes. 💃

routes/web.php
```diff
@@ -60,7 +60,8 @@
     return redirect('us/'.$slug);
 });
 
-// Campaigns cache clear
+// Cache
+$router->get('cache/{cacheId}', 'CacheController');
```
weerd: I'd like to move away from `/next/` now that Ashes is gone, and figured that as we slowly move things to new Admin Dashboard approach we can eventually get rid of the `next/cache/` route.

resources/views/svg/bug-icon.blade.php
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" class="{{ $class ?? '' }}"><path d="M15.3 14.89l2.77 2.77a1 1 0 0 1 0 1.41 1 1 0 0 1-1.41 0l-2.59-2.58A5.99 5.99 0 0 1 11 18V9.04a1 1 0 0 0-2 0V18a5.98 5.98 0 0 1-3.07-1.51l-2.59 2.58a1 1 0 0 1-1.41 0 1 1 0 0 1 0-1.41l2.77-2.77A5.95 5.95 0 0 1 4.07 13H1a1 1 0 1 1 0-2h3V8.41L.93 5.34a1 1 0 0 1 0-1.41 1 1 0 0 1 1.41 0l2.1 2.1h11.12l2.1-2.1a1 1 0 0 1 1.41 0 1 1 0 0 1 0 1.41L16 8.41V11h3a1 1 0 1 1 0 2h-3.07c-.1.67-.32 1.31-.63 1.89zM15 5H5a5 5 0 1 1 10 0z"/></svg>
```
weerd: Icon is from http://www.zondicons.com/ 👌

routes/web.php
DFurnes: Aww yeah! 🔥 🐦

resources/views/svg/bug-icon.blade.php
DFurnes: Nice! Hadn't seen these before. 🙌

PR #1295, opened 2019-03-07
Petition auth signature

mendelB:
### What does this PR do?

This PR adds the authenticated state to the `PetitionSubmissionAction`, displaying the user's signature (first name). https://github.com/DoSomething/phoenix-next/commit/bb58ec95f6f92574d1b37005f6144b4ebb11e935

Also snuck in https://github.com/DoSomething/phoenix-next/commit/4ccd4cd3636b72a2da67e6f4f6326d3c45cd9db1 to disable the `textarea` field on successful submissions as part of the lite post-submission state.

https://github.com/DoSomething/phoenix-next/commit/39081c37c7be51b370cf4f75009817d0970c5187 was interesting. I was seeing that for Lazy Auth'd posts (regular text or petition), once the user autheticates and is redirected, the `pending` state of the submission which is initiated by the queued post action would be overriden with the call to [`initPostSubmissionItem`](https://github.com/DoSomething/phoenix-next/blob/e42812dd73aabfa1a27d20dc3c0a20c343afae48/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js#L23) in the component's constructors.
This isn't a very big deal at all, but it did mean that the button wouldn't be in a loading state, which for a slower connection could lead the viewer to attempt to resubmit the post etc.
I added the [conditional](https://github.com/DoSomething/phoenix-next/compare/petition-auth-signature?expand=1#diff-f9ba48856d8e39c6eba8fd1a4ce29d77R26) to ensure  this doesn't happen, but:
❓I actually figure that we can get rid of this call completely, since initiating the post in the store can just happen when the post request is made and the ['pending' reducer functionality](https://github.com/DoSomething/phoenix-next/blob/e42812dd73aabfa1a27d20dc3c0a20c343afae48/resources/assets/reducers/postSubmissions.js#L56-L65) is triggered, and we don't need the post to be in the store until then?

image

image

### Checklist

### DIFF
--- a/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js
+++ b/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js
@@ -89,8 +89,8 @@ class PetitionSubmissionAction extends React.Component {
 
     const formErrors = getFieldErrors(formResponse);
 
-    const FIRST_NAME_QUERY = gql`
-      query AccountQuery($userId: String!) {
+    const SIGNATURE_QUERY = gql`
+      query SignatureQuery($userId: String!) {
         user(id: $userId) {
           firstName
         }
@@ -126,7 +126,7 @@ class PetitionSubmissionAction extends React.Component {
 
               {userId ? (
                 <Query
-                  query={FIRST_NAME_QUERY}
+                  query={SIGNATURE_QUERY}
                   queryName="user"
                   variables={{ userId }}
                 >
--- a/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js
+++ b/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js
@@ -116,6 +116,7 @@ class PetitionSubmissionAction extends React.Component {
                   placeholder={textFieldPlaceholder}
                   value={this.state.textValue}
                   onChange={this.handleChange}
+                  disabled={this.state.showAffirmation}
                 />
                 <p className="footnote">500 character limit</p>
               </div>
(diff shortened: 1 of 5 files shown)


DFurnes:
> ❓I actually figure that we can get rid of this call completely, since initiating the post in the store can just happen when the post request is made and the ['pending' reducer functionality](https://github.com/DoSomething/phoenix-next/blob/e42812dd73aabfa1a27d20dc3c0a20c343afae48/resources/assets/reducers/postSubmissions.js#L56-L65) is triggered, and we don't need the post to be in the store until then?

Yeah, I think that's the way to do it.

resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js
```diff
@@ -83,6 +89,14 @@ class PetitionSubmissionAction extends React.Component {
 
     const formErrors = getFieldErrors(formResponse);
 
+    const FIRST_NAME_QUERY = gql`
+      query AccountQuery($userId: String!) {
```
DFurnes: Let's rename this! Maybe `SignatureQuery`?

mendelB: Ugh yikes, yes that's quite better!

mendelB:
> > ❓I actually figure that we can get rid of this call completely, since initiating the post in the store can just happen when the post request is made and the ['pending' reducer functionality](https://github.com/DoSomething/phoenix-next/blob/e42812dd73aabfa1a27d20dc3c0a20c343afae48/resources/assets/reducers/postSubmissions.js#L56-L65) is triggered, and we don't need the post to be in the store until then?
> 
> Yeah, I think that's the way to do it.

Ok, I can tackle removing these in a follow-up PR!

PR #1296, opened 2019-03-08
Remove Call to the initPostSubmission Action

mendelB:
### What does this PR do?

This PR removes the call to the `initPostSubmission` action in our various post submission action components.

This is a quick follow up to #1295 where we [deemed it unnecessary](https://github.com/DoSomething/phoenix-next/pull/1295#issuecomment-470667594) to be making this call for now.

### DIFF
--- a/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js
+++ b/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js
@@ -21,11 +21,6 @@ class PetitionSubmissionAction extends React.Component {
       showAffirmation: false,
       textValue: '',
     };
-
-    // Initialize post submission item in store if it doesn't yet exist.
-    if (!props.submissions.items[props.id]) {
-      this.props.initPostSubmissionItem(this.props.id);
-    }
   }
 
   static getDerivedStateFromProps(nextProps) {
@@ -157,7 +152,7 @@ class PetitionSubmissionAction extends React.Component {
               <Button
                 type="submit"
                 attached
-                loading={submissionItem ? submissionItem.isPending : true}
+                loading={submissionItem && submissionItem.isPending}
                 disabled={this.state.showAffirmation}
               >
                 {buttonText}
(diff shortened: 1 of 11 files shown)


weerd:
> This PR removes the call to the initPostSubmission action in our various post submission action components.

When was this deemed unnecessary?

mendelB:
> > This PR removes the call to the initPostSubmission action in our various post submission action components.
> 
> When was this deemed unnecessary?

Dang, sorry, I thought there was consensus. 
In the linked PR description #1295 
> 39081c3 was interesting. I was seeing that for Lazy Auth'd posts (regular text or petition), once the user autheticates and is redirected, the pending state of the submission which is initiated by the queued post action would be overriden with the call to initPostSubmissionItem in the component's constructors.
This isn't a very big deal at all, but it did mean that the button wouldn't be in a loading state, which for a slower connection could lead the viewer to attempt to resubmit the post etc.
I added the conditional to ensure this doesn't happen, but:
❓I actually figure that we can get rid of this call completely, since initiating the post in the store can just happen when the post request is made and the 'pending' reducer functionality is triggered, and we don't need the post to be in the store until then?

Followed By Dave's comment https://github.com/DoSomething/phoenix-next/pull/1295#issuecomment-470667594
I figured it would be a quick easy clean up to remove that logic.

weerd:
Reading through the decision you and Dave made regarding this.

Off the cuff, I'm just a little concerned because I remember having to thoroughly test this because it helped solve a very nuanced bug I was experiencing. I don't recall it completely, but it may have been due to multiple types of posts made on a single campaign or something.

I'd just be weary of removing this if every possible scenario hasn't been tested. This is also a very strong reason why we really need to get our tests back and working again in the system @_@

mendelB:
oooh yikes. I did test this for Petition/Text/Photo on dev's `test-teens-for-jeans` which has the assortment of actions on it. But, happy to table this for when we have less going on and more robust testing if you think that's a better idea!

DFurnes:
I'm fine leaving this as-is. I don't really trust our Redux implementation either.

That said, I'd hope this should be a pretty safe change & was a great flag by @mendelB for opportunistic refactoring - we're just replacing a place where we'd check for a `post.{id}.isPending` to be explicitly false with it being "implicitly" false.

DFurnes:
This was added in #937. From my reading of the bug this solved, I still think this is a safe change.

weerd:
Yeah, defs. It was probably just my approach to the logic there that lead to the need for it and I think it's a good refactor too.

Just want to make sure it's been thoroughly tested manually and with submitting multiple different types of reportback posts on a single campaign. 🚦😀

mendelB:
Ok awesome! I'll give this a merge, and test like an angry teacher

PR #1299, opened 2019-03-11
Embed Content Type

mendelB:
### What does this PR do?

This PR adds a very basic `Embed` Content type (migration script, Entity).

This is extremely bare bones for now, with only a `url` field to be embedded in an <iframe>.

This is also strictly validated as a DoSomething Carto map embed URL for now.

Refs [Pivotal ID #164541569](https://www.pivotaltracker.com/story/show/164541569)

### Checklist

### DIFF
--- a/contentful/content-types/embed.js
+++ b/contentful/content-types/embed.js
@@ -1,0 +1,48 @@
+module.exports = function(migration) {
+  const embed = migration
+    .createContentType('embed')
+    .name('Embed')
+    .description('Embed content from a URL onto a page.')
+    .displayField('internalTitle');
+
+  embed
+    .createField('internalTitle')
+    .name('Internal Title')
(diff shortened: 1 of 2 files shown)


contentful/content-types/embed.js
```diff
@@ -0,0 +1,48 @@
...
+    .validations([
+      {
+        regexp: {
+          pattern:
+            'https:\\/\\/dosomething\\.carto\\.com\\/builder\\/[a-z0-9-]+\\/embed',
+          flags: null,
+        },
+
+        message:
+          'Must be a valid Carto embeddable map URL from the DoSomething space. (https://dosomething.carto.com/dashboard).',
```
DFurnes: Smart! Definitely a good call to roll this out carefully.

PR #1304, opened 2019-03-11
Embed iframe rendering

mendelB:
### What does this PR do?

This PR adds a very basic `Iframe` component for rendering `Embed` entries.

It felt like a good idea to strictly enforce some hostname sanitizing on this `Iframe` even though we do already have the [validation rules](https://github.com/DoSomething/phoenix-next/blob/e42812dd73aabfa1a27d20dc3c0a20c343afae48/contentful/content-types/embed.js#L24-L34) in place on the Content-Type. 

Refs [Pivotal ID #164541569](https://www.pivotaltracker.com/story/show/164541569)

image

([Medium Screens](https://user-images.githubusercontent.com/12417657/54161587-fcabaa00-4428-11e9-90f2-e2f7d98984d4.png))

### Checklist

### DIFF
--- a/resources/assets/components/utilities/Iframe.js
+++ b/resources/assets/components/utilities/Iframe.js
@@ -1,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import ErrorBlock from '../ErrorBlock/ErrorBlock';
+
+const permittedHostnames = ['dosomething.carto.com'];
+
+const Iframe = ({ id, url }) => {
+  const hostname = url && new URL(url).hostname;
+
+  if (!permittedHostnames.includes(hostname)) {
+    console.warn(`Invalid URL ${url} supplied to Iframe component`);
+    return <ErrorBlock />;
+  }
+
+  return <iframe title={`embed ${id}`} src={url} width="100%" height="520" />;
+};
+
+Iframe.propTypes = {
+  id: PropTypes.string.isRequired,
+  url: PropTypes.string.isRequired,
+};
+
+export default Iframe;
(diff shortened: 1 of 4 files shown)


resources/assets/components/utilities/Iframe.js
```diff
@@ -0,0 +1,24 @@
...
+import ErrorBlock from '../ErrorBlock/ErrorBlock';
+
+const permittedHostnames = ['dosomething.carto.com'];
+
+const Iframe = ({ id, url }) => {
+  const hostname = url && new URL(url).hostname;
+
+  if (!permittedHostnames.includes(hostname)) {
+    console.warn(`Invalid URL ${url} supplied to Iframe component`);
+    return <ErrorBlock />;
```
mendelB: I wasn't perfectly sure what to do here, but a console warning and an error felt ok?

resources/assets/components/utilities/Iframe.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  return <iframe title={`embed ${id}`} src={url} width="100%" height="520" />;
```
mendelB: This `title` is not going to be super helpful for screen readers... It needs to be unique though and we don't have a `title` field on this content type. Maybe something slightly more descriptive? (`embedded block for ${url}`?)

resources/assets/components/utilities/Iframe.js
DFurnes: Yeah, I think this is fine for now! 👍

resources/assets/components/utilities/Iframe.js
DFurnes: Is that requirement for one of our accessibility linting rules? I agree that without an editorial title field it's hard to choose something that won't make for a bad screen-reader experience (imagine Siri reading out a Carto URL to you haha).

mendelB: Yeah, this was indeed a linting rule. My thought was it could be useful to know which URL is embedding, but lol def sounds like that might be abusive.

PR #1307, opened 2019-03-12
Update to add new text-content class and styles.

weerd:
### What does this PR do?

This PR adds a new stylesheet for the `TextContent` component. The intention is to begin deprecating the `.markdown` class in favor of the `.text-content` class which applies to both `RichText` and `StandardMarkdown` content. The `.markdown` and `.richtext` classes will exist, but they will only be used to override any `.text-content` styles that the `StandardMarkdown` and `RichText` components respectively, need to adjust.

Since we can now utilize entries inside of RichText fields, I found we could harness the power of the [*child combinator*](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator) selector to ensure that only specific tags that are immediate children of the tag containing the text content get styles applied, but the styles do not leak into embedded entries.

image

Refs [Pivotal ID #164584567](https://www.pivotaltracker.com/story/show/164584567)

### Checklist

### DIFF
--- a/resources/assets/components/utilities/TextContent/TextContent.js
+++ b/resources/assets/components/utilities/TextContent/TextContent.js
@@ -6,29 +6,32 @@ import classnames from 'classnames';
 import RichTextDocument from './RichTextDocument';
 import StandardMarkdown from './StandardMarkdown';
 
-import './markdown.scss';
+import './text-content.scss';
+import './markdown.scss'; // @deprecate
 
 /**
- * Render Markdown as Markup prepared for a React Component
+ * Render TextContent as Markup prepared for a React Component.
  *
- * @TODO: rename component to <TextContent>
  * @param  {String} options.className
  * @param  {String|Array|Object} options.children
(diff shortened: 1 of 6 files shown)


resources/assets/components/utilities/TextContent/TextContent.js
```diff
@@ -6,29 +6,32 @@ import classnames from 'classnames';
 import RichTextDocument from './RichTextDocument';
 import StandardMarkdown from './StandardMarkdown';
 
-import './markdown.scss';
+import './text-content.scss';
+import './markdown.scss'; // @deprecate
```
weerd: For now, I basically duplicated parts of the `markdown.scss` stylesheet for the `text-content.scss` styles. The intent is to eventually deprecate the `markdown.scss`, and basically make it only overrides that apply to the `StandardMarkdown` component (if needed).

weerd:
There's still more styling to deal with, like the yellow bar text alignment and spacing, but this is a first step!

PR #1308, opened 2019-03-13
Iframe State for Mobile View

mendelB:
### What does this PR do?

This PR adds state for mobile view to the Iframe component -- fetching and displaying an image preview via the Phoenix Embed API.

**Carto Map EMbed Example**:
screencapture-phoenix-test-us-blocks-cApWbYfL9DLEbrgkxvd29-2019-03-13-14_34_18

As [discussed](https://dosomething.slack.com/archives/C2BPA7M8F/p1552499163168500), this is somewhat specific for the Carto map embed. We'll definitely be updating this as we build our more extensively supported Embed functionality.

Refs [Pivotal ID #https://www.pivotaltracker.com/story/show/164541569](https://www.pivotaltracker.com/story/show/164541569)

### Checklist

### DIFF
--- a/resources/assets/components/utilities/Iframe.js
+++ b/resources/assets/components/utilities/Iframe.js
@@ -1,20 +1,69 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import LazyImage from './LazyImage';
+import { getRequest } from '../../helpers/api';
 import ErrorBlock from '../ErrorBlock/ErrorBlock';
+import { getFormattedScreenSize } from '../../helpers';
 
 const permittedHostnames = ['dosomething.carto.com'];
 
-const Iframe = ({ id, url }) => {
-  const hostname = url && new URL(url).hostname;
+class Iframe extends React.Component {
+  constructor(props) {
+    super(props);
 
-  if (!permittedHostnames.includes(hostname)) {
-    console.warn(`Invalid URL ${url} supplied to Iframe component`);
-    return <ErrorBlock />;
+    this.state = {
+      isMobile: getFormattedScreenSize() === 'small',
+      imagePreview: null,
+      hasError: false,
+    };
+
+    const hostname = new URL(this.props.url).hostname;
+    if (!permittedHostnames.includes(hostname)) {
+      console.warn(
+        `Invalid URL ${this.props.url} supplied to Iframe component`,
+      );
+      this.state = {
+        ...this.state,
+        hasError: true,
+      };
+    }
   }
 
-  return <iframe title={`embed ${id}`} src={url} width="100%" height="520" />;
-};
+  componentDidMount() {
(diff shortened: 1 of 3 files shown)


resources/assets/components/utilities/Iframe.js
```diff
@@ -1,20 +1,76 @@
...
-  const hostname = url && new URL(url).hostname;
+class Iframe extends React.Component {
+  constructor(props) {
+    super(props);
 
-  if (!permittedHostnames.includes(hostname)) {
-    console.warn(`Invalid URL ${url} supplied to Iframe component`);
-    return <ErrorBlock />;
+    this.state = {
+      isMobile: getFormattedScreenSize() === 'small',
```
DFurnes: We probably want to use something like [`react-media`](https://github.com/ReactTraining/react-media) instead of manually grabbing this and placing in state, since this won't re-render if a user resizes their browser (admittedly, probably an edge case).

mendelB: @DFurnes Ah ok, my bad, I'd figured we'd just ignore that edge case, but I'll add this in! 

In terms of usage with the `react-media` component, would you recommend always loading the image via the API and then rendering it conditionally for mobile, or should we only make that call  to begin with *if* we're on or change to mobile? (I'd assume the former since the latter might require some extra logic to ensure we only make that call once etc.?)

DFurnes: No worries! It's definitely an edge-case, and not a bad assumption to make! I'm good to ship this as-is and make a ticket to revisit later, depending on what else is still needed for launch.

If we do decide to dig into this & it doesn't get too hairy, I think we'd want to only kick off that API request if we're rendering the mobile UI – say, via a `IframePreview` component that handles its own data fetching when it mounts.

mendelB: Ok smart!
@ngjo how strapped for time are we? Do you feel like we're good to tackle this one now, or should we table this for a post-launch? (I don't estimate this refactor being more than a 2-3)

ngjo: hi @mendelB i added this to the post-launch list and im reviewing the current list right now to see what's left -- if it works for you, i'd recommend just closing this part of the work out to "finish" the map stuff.

mendelB:
Ahhh @DFurnes sorry, I had the refactor just about wrapped up, so @ngjo [gave the 👍](https://dosomething.slack.com/archives/C3ASB4204/p1552571864264500 )to push

resources/assets/components/utilities/Iframe.js
mendelB: Updated with a refactoring to utilize the `react-media` package and an `IframePreview` component to render the small vs big view responsively.

DFurnes:
Oh heck yeah, even better!! This looks great @mendelB! 🐎

PR #1309, opened 2019-03-13
Full reworking of Grid system.

weerd:
### What does this PR do?

This PR reworks a lot of structure in our story page and some components to work better with CSS grid and how we want content to layout on the page. We can now use named grid lines to define how content is laid out on the page at different media sizes (full, wide, main, and narrow).

- [Small Screenshot](https://user-images.githubusercontent.com/105849/54392273-966b9500-467d-11e9-9e77-b0b1ac8522f4.png)
- [Medium Screenshot](https://user-images.githubusercontent.com/105849/54392286-9ff4fd00-467d-11e9-9532-03e45a0a4bbf.png)
- [Large Screenshot](https://user-images.githubusercontent.com/105849/54392304-adaa8280-467d-11e9-8ed2-c498a32ad32e.png)

A few notes on some of the updates:

- To have the grid work correctly within RichText content and embedded entries, we needed to reconsider how the page outputs content and where the grid is applied. We now apply the grid to the containing div for RichText content, and apply grid classes that indicate where to start/end their column alignment to _each individual element_. We utilized Contentful's richtext react renderer package to help harness and output the correct class for each block level item.
- In the `ContentfulEntry` component, that outputs different contentful entries, we're going to move to _not_ wrap most of these in a div that has margin related classes attached. This should happen in the component that calls `ContentfulEntry` so that these classes are added only when needed. A specific case in point, is that now with the grid, we don't really need these margin classes.
- Speaking of classes, we now pass the `className` prop down through `ContentfulEntry` so we can pass classes to handle stuff like adding margins to the top level containing tag of the component.
- More components need to be updated. I've only really updated components that we _know_ are going to be used immediately on a Story Page and specifically for Prom. We can update the rest of the components over time (and we'll want to, so we can make sure they're not broken outside of Story Page).

Refs [Pivotal ID #164092392](https://www.pivotaltracker.com/story/show/164092392)

### Checklist

### DIFF
--- a/resources/assets/components/ContentfulEntry/renderers.js
+++ b/resources/assets/components/ContentfulEntry/renderers.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import classnames from 'classnames';
 import { PuckWaypoint } from '@dosomething/puck-client';
 
 import Affirmation from '../Affirmation';
@@ -80,28 +81,31 @@ export function renderLinkAction(step) {
  * @param {Object} data
  * @return {Component}
  */
-export function renderTextSubmissionAction(data) {
+export function renderTextSubmissionAction(data, className = null) {
   const contentfulId = data.id;
   const fields = withoutNulls(data.fields);
 
   return (
-    <div
-      key={`text-submission-action-${contentfulId}`}
-      className="margin-horizontal-md margin-bottom-lg"
-    >
+    <React.Fragment>
       <PuckWaypoint
         name="text_submission_action-top"
         waypointData={{ contentfulId }}
       />
-      <TextSubmissionActionContainer id={contentfulId} {...fields} />
-      <div className="margin-vertical-md">
-        <SubmissionGalleryBlockContainer type="text" />
-      </div>
+      <TextSubmissionActionContainer
+        className={classnames(className)}
+        id={contentfulId}
+        key={`text-submission-action-${contentfulId}`}
+        {...fields}
+      />
+      <SubmissionGalleryBlockContainer
+        className={classnames(className)}
+        type="text"
+      />
       <PuckWaypoint
         name="text_submission_action-bottom"
         waypointData={{ contentfulId }}
       />
-    </div>
+    </React.Fragment>
   );
 }
 
@@ -139,22 +143,26 @@ export function renderPhotoSubmissionAction(data) {
  * @param  {Object} data
  * @return {Component}
  */
-export function renderPetitionSubmissionAction(data) {
+export function renderPetitionSubmissionAction(data, className = null) {
   const contentfulId = data.id;
   const fields = withoutNulls(data.fields);
 
   return (
-    <div className="margin-horizontal-md margin-bottom-lg">
+    <React.Fragment>
       <PuckWaypoint
         name="petition_submission_action-top"
         waypointData={{ contentfulId }}
       />
-      <PetitionSubmissionActionContainer id={contentfulId} {...fields} />
+      <PetitionSubmissionActionContainer
+        className={classnames(className)}
+        id={contentfulId}
+        {...fields}
+      />
       <PuckWaypoint
         name="petition_submission_action-bottom"
         waypointData={{ contentfulId }}
       />
-    </div>
+    </React.Fragment>
   );
 }
 
@@ -206,14 +214,14 @@ export function renderContentBlock(data) {
  * @param {Object} data Embed
  * @return {Component}
  */
-export function renderEmbed(data) {
+export function renderEmbed(data, className = null) {
   const contentfulId = data.id;
   const fields = withoutNulls(data.fields);
 
   return (
     <React.Fragment>
       <PuckWaypoint name="embed-top" waypointData={{ contentfulId }} />
-      <Iframe id={contentfulId} {...fields} />
+      <Iframe className={classnames(className)} id={contentfulId} {...fields} />
       <PuckWaypoint name="embed-bottom" waypointData={{ contentfulId }} />
     </React.Fragment>
   );
--- a/resources/assets/components/ContentfulEntry/renderers.js
+++ b/resources/assets/components/ContentfulEntry/renderers.js
@@ -94,7 +94,6 @@ export function renderTextSubmissionAction(data, className = null) {
       <TextSubmissionActionContainer
         className={classnames(className)}
         id={contentfulId}
-        key={`text-submission-action-${contentfulId}`}
         {...fields}
       />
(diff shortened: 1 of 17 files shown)


resources/assets/components/ContentfulEntry/renderers.js
```diff
@@ -80,28 +81,31 @@ export function renderLinkAction(step) {
...
         waypointData={{ contentfulId }}
       />
-      <TextSubmissionActionContainer id={contentfulId} {...fields} />
-      <div className="margin-vertical-md">
-        <SubmissionGalleryBlockContainer type="text" />
-      </div>
+      <TextSubmissionActionContainer
+        className={classnames(className)}
+        id={contentfulId}
+        key={`text-submission-action-${contentfulId}`}
```
DFurnes: I don't believe this should need a key, since it's not dynamically generated.

resources/assets/components/ContentfulEntry/renderers.js
```diff
@@ -80,28 +81,31 @@ export function renderLinkAction(step) {
...
       <PuckWaypoint
         name="text_submission_action-top"
         waypointData={{ contentfulId }}
       />
-      <TextSubmissionActionContainer id={contentfulId} {...fields} />
-      <div className="margin-vertical-md">
-        <SubmissionGalleryBlockContainer type="text" />
-      </div>
+      <TextSubmissionActionContainer
+        className={classnames(className)}
```
DFurnes: Not really important either way, but we could just pass `className` directly to the `TextSubmissionAction` here (since we're not doing any special transformation or conditional classes with `classnames()`).

resources/assets/components/ContentfulEntry/renderers.js
```diff
@@ -192,11 +200,16 @@ export function renderAffirmation(data) {
...
-export function renderContentBlock(data) {
+export function renderContentBlock(data, className = null) {
   const fields = withoutNulls(data.fields);
 
   return (
-    <ContentBlock key={`content-block-${data.id}`} id={data.id} {...fields} />
+    <ContentBlock
+      className={classnames(className)}
+      id={data.id}
+      key={`content-block-${data.id}`}
```
DFurnes: Same deal - shouldn't need a `key` right?

resources/assets/components/blocks/SectionBlock/section-block.scss
```diff
@@ -1,30 +1,18 @@
...
-    @include media($medium) {
-      grid-column: 2 / -2;
-      padding: $base-spacing 0;
-    }
-
-    @include media($large) {
-      grid-column: 5 / -5;
-    }
+.section-block__content {
+  > *:not(ul):not(ol):not(blockquote):not(.component-entry) {
```
DFurnes: Eeek! Can we apply this to the paragraph & headings not matched by this, instead?

resources/assets/components/ContentfulEntry/renderers.js
weerd: Cool. Added prior to this PR but I can remove.

resources/assets/components/blocks/SectionBlock/section-block.scss
weerd: haha. Sure, I can try. I kind of got into trying out `:not()` when other solutions weren't working.

resources/assets/components/ContentfulEntry/renderers.js
weerd: Added prior to this PR but I can remove.

resources/assets/components/ContentfulEntry/renderers.js
weerd: Oh ya. I think I was doing more in earlier approaches, and then changed it up, or just residual copy/🍝

weerd: Did the same for all the other uses in the file and could basically remove `classnames` from importing. We can always reimport if we end up changing to do more unique things with classes in renderers.js

resources/assets/components/blocks/SectionBlock/section-block.scss
weerd: Ok, so now I remember. I intended the above to be temporary. The alternative would be:

```css
  > h1,
  > h2,
  > h3,
  > h4,
  > h5,
  > h6,
  > p {
    text-align: center;
  }
```

We could also add a class (like I think you mentioned to me IRL) to richtext headings, but I think we need to refrain from adding a `.heading` class since that carries some weight in Forge. It could be something else, or one of the two options above with the understanding that this is ideally temporary until we change fix up Forge and use Tailwind.

DFurnes: Could we apply the `text-center` utility class to these elements in the renderer? Or a `story-content` class shared between them with this single CSS property?

weerd: Also to clarify, I don't want to add like a `text-center` class to the richtext react rendering output, since moving forward we want RichtText on more things and don't want to force centering ALL text, just text on SectionBlocks in StoryPages.

DFurnes: Gotcha, that makes sense. In that case, I'd consider passing in a `story-content` class to apply to these elements I think, but this is fine too for now! 👍

weerd: Yeah, I think lets keep as is for now (I can switch to not using the crazy chained `:not()` rule, haha) and then as we move further with RichText on other things we figure out a better plan.

resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js
```diff
@@ -95,7 +95,14 @@ class PetitionSubmissionAction extends React.Component {
 
     return (
       <React.Fragment>
-        <div className="petition-submission-action margin-bottom-lg" id={id}>
+        <div
+          className={classnames(
+            'petition-submission-action',
+            'margin-bottom-lg',
```
mendelB: Quiiiick thing: I took a quick peek at the deploy preview, and it seems like this margin-bottom, together with the `margin-top: 12px`each `.component-entry` gets adds up to a bit more space then we need between the petition and the more info block
image

weerd: @mendelB yep yep. I'll be doing some follow-up cleanup with additional spacing. How we decided on using CSS Grid now nullifies margin collapsing.

PR #1311, opened 2019-03-14
Enable StoryPage block on Homepage Gallery

mendelB:
### What does this PR do?

This PR enables the Story Page to be added to the HomePage gallery.

- adds `storyPage` to the validations list for the HomePage `blocks` field in the Contentful migration script
- adds a `TruncatedStoryPage` Entity for the `parseBlocks` method so that we can render out the truncated necessary fields of the StoryPage for the HomePage `blocks` field.
- https://github.com/DoSomething/phoenix-next/commit/8e273d2a7cd14d33992b2212d5c7e4aff8e863a8 snuck in a quick change to add the new and improved server rendered admin dashboard to the homepage, which required a little update the `get_cache_url` helper method to support static cache Ids with no slug, like the Homepage. (https://github.com/DoSomething/phoenix-next/commit/b42fa03ab837d191fd3ad5d4bf58f494064f13ec)

Refs [Pivotal ID #164213954](https://www.pivotaltracker.com/story/show/164213954)

### Checklist

image


app/Entities/TruncatedStoryPage.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace App\Entities;
+
+use JsonSerializable;
+
+class TruncatedStoryPage extends StoryPage implements JsonSerializable
```
weerd: Not for this PR, but I wonder if we could eventually consolidate these to some other more generic class name, since these `Truncated*` classes mostly use the same series of fields... 🤔

app/helpers.php
```diff
@@ -21,13 +21,13 @@ function get_cache_id($prefix, $slug)
...
- * @param  string $slug
+ * @param  string      $prefix
+ * @param  string|null $slug
  * @return string
  */
-function get_cache_url($prefix, $slug)
+function get_cache_url($prefix, $slug = null)
 {
-    $cacheId = get_cache_id($prefix, $slug);
+    $cacheId = $slug ? get_cache_id($prefix, $slug) : $prefix;
```
weerd: 💃

app/Entities/TruncatedStoryPage.php
mendelB: iiiinteresting! like maybe some basic 'preview' type that pulls in the general fields we'd need for a gallery preview or something...

PR #1313, opened 2019-03-15
Updating to relocate PuckWaypoints on some components.

weerd:
### What does this PR do?

This PR does some margin cleanup but primarily moves some `PuckWaypoints` components _inside_ of some of the submission actions and the embed. I think in general it makes more sense to keep them in the component and helps actually remove some code cruft in the `renderers.js` (if we do this to all items we probably don't even need this file).

The `PuckWaypoints` also output empty divs in the markup, and while this is usually fine, it's not so great if those empty divs become child elements of the grid and can actually affect the layout of other items:

Screen Shot 2019-03-14 at 7 20 30 PM

By moving them inside the component it helps make sure they don't impact the grid layout and the markup just looks generally cleaner ✨ 

...

🌵 

### Checklist

### DIFF
--- a/resources/assets/components/ContentfulEntry/ContentfulEntry.js
+++ b/resources/assets/components/ContentfulEntry/ContentfulEntry.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import React from 'react';
+import * as React from 'react';
 
 import NotFound from '../NotFound';
 import Iframe from '../utilities/Iframe';
--- a/resources/assets/components/ContentfulEntry/ContentfulEntry.js
+++ b/resources/assets/components/ContentfulEntry/ContentfulEntry.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { React, Fragment } from 'react';
+import React from 'react';
 
 import NotFound from '../NotFound';
 import Iframe from '../utilities/Iframe';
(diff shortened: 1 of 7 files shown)


resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -1,8 +1,9 @@
 /* @flow */
 
-import React from 'react';
+import * as React from 'react';
```
mendelB: 💡

resources/assets/helpers/text.js
```diff
@@ -108,11 +108,12 @@ export function parseRichTextDocument(document, styles) {
           {children}
         </h6>
       ),
-      [BLOCKS.PARAGRAPH]: (node, children) => (
-        <p className="grid-main" style={textColor}>
-          {children}
-        </p>
-      ),
+      [BLOCKS.PARAGRAPH]: (node, children) =>
+        children[0] ? (
```
mendelB: Interesting... these render out with no children?

weerd: Oh! @mendelB I meant to add a comment in GH... this eliminates empty `<p>` tags from showing up if someone adds extra carriage returns in the RichText field.

resources/assets/components/ContentfulEntry/ContentfulEntry.js
weerd: Yeah.... Facebook Flow kept flagging errors in Wercker that it couldn't find `<React.Fragment>` and it was very annoying. I tried a few solutions and this one seemed to work. I think we need to assess if we want to continue with Flow or not and either double-down and use it everywhere or remove it.

PR #1315, opened 2019-03-15
Some More #Quick Design QA Updates

mendelB:
### What does this PR do?

This PR adds a couple of quick styling updates:
- Override the Forge font rule for `.text-field`s so we use `Source Sans Pro` (https://goo.gl/yuqiri)
- Bump the minimum width for the two column styling to kick in for the gallery grid system

Re the gallery grid width - totally open to debate of course, but while testing Prom on my (android 😲) phone, it felt a bit claustrophobic to have the two column grid on the gallery:
Exhibit A

As opposed to one:
Exhibit B

The only other place this would currently affect would be the Explore Campaigns gallery:
Exhibit C

Which addmittedly does not seem as claustrophobic currently:
Exhibit D

LMK whatcha think! CC @lkpttn 

Refs [Pivotal ID #164637008](https://www.pivotaltracker.com/story/show/164637008)

### DIFF
--- a/resources/assets/scss/base.scss
+++ b/resources/assets/scss/base.scss
@@ -57,6 +57,11 @@ p {
   }
 }
 
+// Override Forge rule for .text-field font-family (https://git.io/fjeFh).
+.text-field {
+  font-family: $primary-font-family;
+}
+
 // Utility classes.
 // @TODO: Move over to Forge 7 soon!
 .background-image-centered {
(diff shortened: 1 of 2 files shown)


lkpttn:
Hmmm a real pickle. I think we go ahead and make the gallery change here, but I'd like to find a way to main two 2-across approach for galleries where it makes sense.

weerd:
As we move to include grid for more things, the size adjustment could bite us. I think we should reassess post-prom and also find a unifying approach to all galleries.

I'm seeing some issues with gallery grid definitions for images block with some weird sizing rules:

image

image

mendelB:
Great points! Definitely sounds like a great idea to re-assess, and establish a wiser approach.

PR #1317, opened 2019-03-15
Add stricter location validation & fixes for particular country codes.

DFurnes:
### What does this PR do?

This pull request adds a quick fix to filter out "bad" geolocation data provided by Fastly. It checks that the country and region codes are the correct number of characters, and massages some specific countries that weren't provided in the correct format.

Mendel [noticed this morning](https://dosomething.slack.com/archives/C3ASB4204/p1552657729291600) that some users are getting validation errors due to bad `location` format. For example, users from Guam are receiving the region of `NO REGION` from Fastly.

From scanning the logs, it looks like only users from US Territories (Guam, Puerto Rico, and American Samoa) were affected, since Fastly provides the territory as the country code (so `GU-NO REGION` instead of [`US-GU`](https://en.wikipedia.org/wiki/ISO_3166-2:GU), and `AS-NO REGION` instead of [`US-AS`](https://en.wikipedia.org/wiki/ISO_3166-2:AS)).

Refs [Pivotal ID #164667891](https://www.pivotaltracker.com/story/show/164667891)

### Checklist

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -52,6 +52,12 @@ class AppServiceProvider extends ServiceProvider
                 $region = null;
             }
 
+            // Fix improper ISO-3601-2 format for American territories:
+            if (in_array($country, ['AS', 'GU', 'MP', 'PR'])) {
+                $region = $country;
+                $country = 'US';
+            }
+
             $view->with('auth', [
                 'isAuthenticated' => auth()->check(),
                 'id' => auth()->id() ?: request()->query('user_id'),
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -40,9 +40,17 @@ class AppServiceProvider extends ServiceProvider
         });
 
         View::composer('*', function ($view) {
-            // Read Fastly's geolocation headers for location-based features.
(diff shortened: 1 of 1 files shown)


app/Providers/AppServiceProvider.php
```diff
@@ -40,9 +40,23 @@ public function boot(Router $router)
...
+                $country = null;
+            }
+
             $region = request()->header('X-Fastly-Region-Code');
+            if (! in_array(strlen($region), [2, 3])) {
+                $region = null;
+            }
+
+            // Fix improper ISO-3601-2 format for American territories:
+            if (in_array($country, ['AS', 'GU', 'MP', 'PR'])) {
```
mendelB: I guess we may have to re-visit this list once we push back to the EU etc. (in case other countries have similar issues)?

weerd: Ya, good note. I can make a ticket and tag it #global so we don't forget this @_@

DFurnes: Indeed!

DFurnes:
Double-checked the full list of US territories, and added `US-UM` and `US-VI` in 4905bd4.

DFurnes:
And whoops, accidentally committed a trip to my favorite Canadian province when testing!

Fixed in d930b95. 🇨🇦

PR #1318, opened 2019-03-15
Update Text Submission Action Disclaimer Content

mendelB:
### What does this PR do?

This PR updates the `TextSubmissionAction`'s disclaimer text:

image

image
 

Does this seem like a lot for mobile? should we shrink the font size? (cc @lkpttn )
image

Refs [Pivotal ID #164287413](https://www.pivotaltracker.com/story/show/164287413)
https://www.pivotaltracker.com/story/show/164287413/comments/200590816

### DIFF
--- a/resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
+++ b/resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
@@ -148,8 +148,16 @@ class TextSubmissionAction extends React.Component {
                 {this.props.buttonText}
               </Button>
               <p className="footnote padding-horizontal-md padding-bottom-md">
-                Your submission will be reviewed by a DoSomething.org staffer
-                and added to our public gallery.
+                The data you submit in this form will be handled in accordance
+                with the DoSomething website{' '}
+                <a
+                  href="https://www.dosomething.org/us/about/privacy-policy"
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  privacy policy
+                </a>
+                . It will be reviewed by a DoSomething staff member.
               </p>
             </form>
           </Card>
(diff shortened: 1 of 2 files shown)


resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
```diff
@@ -148,8 +148,18 @@ class TextSubmissionAction extends React.Component {
                 {this.props.buttonText}
               </Button>
               <p className="footnote padding-horizontal-md padding-bottom-md">
-                Your submission will be reviewed by a DoSomething.org staffer
-                and added to our public gallery.
+                The data you submit in this form will be handled in accordance
+                with the DoSomething website{' '}
+                {/* Workaround for this jsx-a11y bug https://git.io/fN814 */}
+                {/* eslint-disable-next-line jsx-a11y/anchor-has-content */}
```
weerd: So weird!

weerd: Curious... what if you wrap all the text in a template literal? 🤔 Still gross, but maybe more readable if it works?

mendelB: This was officially [fixed](https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/437), I *think* we just need to update `es-lint` in our [`@dosomething/eslint-config`](@dosomething/eslint-config) package

mendelB: > Curious... what if you wrap all the text in a template literal? 🤔 Still gross, but maybe more readable if it works?

Haha that actually works and is *less* gross! (maybe worth keeping this though so we can easilly spot and update once it's updated on our eslint config?)

weerd: Yeah, fine either way. Mind adding a not to update once `@dosomething/eslint-config` is updated. And maybe throw a quick GH issue in that repo and link here?

PR #1321, opened 2019-03-18
Petition Submission Action Post-Submission State

mendelB:
### What does this PR do?

This PR adds updated post-submission state to the `PetitionSubmissionAction` component
- https://github.com/DoSomething/phoenix-next/commit/374fb3e7c0877e8fb6086e467507651f24376cb5 displays the affirmation modal post successful submissions
- https://github.com/DoSomething/phoenix-next/commit/fda0a1627eec04084c2aaeb4a475ca413c65f47f (the meat of the thing) adds an updated graphql query to fetch the users posts for this action and their first name for the signature and handles the various submission states accordingly.
- https://github.com/DoSomething/phoenix-next/commit/5f0ab066456a843926ab7c7e31f8ab3c040508cd disables the submission button if the text field length exceeds the character limit (as per [this Slack conversation](https://dosomething.slack.com/archives/C2BPA7M8F/p1552936127277200))
-  https://github.com/DoSomething/phoenix-next/commit/369c65d3af1a9c6887e6ec628844d8dc4e0aaeae (Suggestion! I can roll this back) displays the negative character count (a la Twitter) which I think makes it easier for the user to patch a faulty text submission.

Refs [Pivotal ID #164711433](https://www.pivotaltracker.com/story/show/164711433)

### Checklist

### DIFF
--- a/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js
+++ b/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js
@@ -27,6 +27,8 @@ const USER_POSTS_QUERY = gql`
   }
 `;
 
+const CHARACTER_LIMIT = 500;
+
 class PetitionSubmissionAction extends React.Component {
   constructor(props) {
     super(props);
@@ -152,7 +154,10 @@ class PetitionSubmissionAction extends React.Component {
                         onChange={this.handleChange}
                         disabled={submitted}
                       />
-                      <CharacterLimit limit={500} text={this.state.textValue} />
+                      <CharacterLimit
+                        limit={CHARACTER_LIMIT}
+                        text={this.state.textValue}
+                      />
                     </div>
 
                     <div className="padded">
@@ -171,7 +176,10 @@ class PetitionSubmissionAction extends React.Component {
                       type="submit"
                       attached
                       loading={submissionItem && submissionItem.isPending}
-                      disabled={submitted}
+                      disabled={
+                        submitted ||
+                        this.state.textValue.length > CHARACTER_LIMIT
+                      }
                     >
                       {buttonText}
                     </Button>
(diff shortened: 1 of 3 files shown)


weerd:
> (Suggestion! I can roll this back) displays the negative character count (a la Twitter) which I think makes it easier for the user to patch a faulty text submission.

Noice! I dig it 👏

DFurnes:
> > (Suggestion! I can roll this back) displays the negative character count (a la Twitter) which I think makes it easier for the user to patch a faulty text submission.
> 
> Noice! I dig it 👏

Me too! 👍 👍

resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js
```diff
@@ -107,70 +115,79 @@ class PetitionSubmissionAction extends React.Component {
...
-                disabled={this.state.showAffirmation}
-              >
-                {buttonText}
-              </Button>
-            </form>
-          </Card>
+          <Query
+            query={USER_POSTS_QUERY}
+            variables={{ userId, actionIds: String(actionId) }}
+            queryName="userPosts"
```
DFurnes: Y'know, I don't think this [is actually a prop](https://www.apollographql.com/docs/react/essentials/queries.html#props) for this component! No biggie, but maybe can remove.

mendelB: Oh right! Should we perhaps make use of the `displayName` prop instead?

DFurnes: Yeah, I haven't personally found it too disorienting with them all named `Query` so far (since it's always so close to the place it's being used), but it could be a good habit to get into!

PR #1329, opened 2019-03-20
Adds support for rendering assets in rich text.

DFurnes:
### What does this PR do?

This pull request adds support for rendering embedded assets in rich text fields. The asset information is fetched from Contentful using the `asset(id: "…")` query introduced in DoSomething/graphql#83, and then the image itself is loaded via our `LazyImage` component.

If an image is larger than 700x700 (a completely arbitrary size that seemed reasonable for the story page), it will be sized down by Contentful's Image API to fit within that constraint. Otherwise, it'll be loaded as is and given center stage. Remarkable.

Screen Shot 2019-03-20 at 5 37 01 PM

~This is blocked on DoSomething/graphql#83 being reviewed & deployed.~ No longer!

Refs [Pivotal ID #164712381](https://www.pivotaltracker.com/story/show/164712381).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/ContentfulAsset/ContentfulAsset.js
+++ b/resources/assets/components/utilities/ContentfulAsset/ContentfulAsset.js
@@ -47,7 +47,7 @@ const ContentfulAsset = ({ className, id, width, height }) => (
       }
 
       return (
-        <div className={className} style={{ width: '100%' }}>
+        <div className={className}>
           <LazyImage src={data.asset.url} alt={data.asset.description} />
         </div>
       );
--- a/resources/assets/components/utilities/ContentfulAsset/ContentfulAsset.js
+++ b/resources/assets/components/utilities/ContentfulAsset/ContentfulAsset.js
@@ -1,0 +1,71 @@
+import React from 'react';
+import gql from 'graphql-tag';
+import PropTypes from 'prop-types';
+import { Query } from 'react-apollo';
+
+import LazyImage from '../LazyImage';
+import Card from '../Card/Card';
+import TextContent from '../TextContent/TextContent';
+import ErrorBlock from '../../ErrorBlock/ErrorBlock';
+
+const CONTENTFUL_ASSET_QUERY = gql`
+  query ContentfulAssetQuery($id: String!, $width: Int!, $height: Int!) {
+    asset(id: $id) {
+      id
+      description
+      url(w: $width, h: $height)
+    }
+  }
+`;
+
+const ContentfulAsset = ({ className, id, width, height }) => (
+  <Query query={CONTENTFUL_ASSET_QUERY} variables={{ id, width, height }}>
+    {({ loading, error, data }) => {
+      if (loading) {
+        return (
+          <div className="grid-main spinner -centered margin-vertical-xlg" />
+        );
+      }
+
+      if (error) {
+        return (
+          <div className="component-entry grid-main margin-bottom-md">
+            <ErrorBlock />
+          </div>
+        );
+      }
+
+      if (!data.asset) {
+        return (
+          // @TODO: repurpose NotFound component to be customizeable; using basic Card component for now.
+          <Card className="component-entry grid-main margin-bottom-md rounded bordered">
+            <TextContent className="padded text-center">
+              Sorry! The specified asset was not found.
+            </TextContent>
+          </Card>
+        );
+      }
+
+      return (
+        <div className={className} style={{ width: '100%' }}>
+          <LazyImage src={data.asset.url} alt={data.asset.description} />
+        </div>
+      );
+    }}
+  </Query>
+);
+
+ContentfulAsset.propTypes = {
+  id: PropTypes.string.isRequired,
+  className: PropTypes.string,
+  width: PropTypes.number,
+  height: PropTypes.number,
+};
+
+ContentfulAsset.defaultProps = {
+  className: null,
+  width: 700,
+  height: 700,
+};
+
+export default ContentfulAsset;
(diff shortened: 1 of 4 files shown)


DFurnes:
@weerd @mendelB This is ready for review, and can be seen in action on the review branch [here](https://dosomething-phoenix-de-pr-1329.herokuapp.com/us/stories/page-test)!

I did a little cleanup in the `ContentfulEntryLoader` in this PR as well (to pull rich-text specific classes into the rich-text renderer, since we'll eventually be using these components in other contexts). Other than that, this should be fairly straightforward! :v:

resources/assets/components/utilities/ContentfulAsset/ContentfulAsset.js
```diff
@@ -0,0 +1,62 @@
...
+        return <div className="spinner -centered margin-vertical-xlg" />;
+      }
+
+      if (error) {
+        return <ErrorBlock />;
+      }
+
+      if (!data.asset) {
+        return (
+          // @TODO: repurpose NotFound component to be customizeable; using basic Card component for now.
```
DFurnes: One thing I was tossing around is whether we should just merge `error` and `!data.asset` together into the above `<ErrorBlock>` condition, since telling a user that "the specified content was not found" is not really helpful or actionable (and it's unlikely an editor embeds a blocks or asset that don't exist).

@weerd @mendelB What do y'all think?

resources/assets/helpers/text.js
```diff
@@ -125,10 +126,14 @@ export function parseRichTextDocument(document, styles) {
         </blockquote>
       ),
       [BLOCKS.EMBEDDED_ENTRY]: node => (
-        <ContentfulEntryLoader
-          id={node.data.target.sys.id}
-          className="component-entry"
-        />
+        <div className="grid-main component-entry">
```
weerd: Yeah, so I initially went this route and we can't really do this and need to pass the classes. For example, not all embedded entries are grid-main. In the review app link you'll see this broke the post gallery from going `grid-wide`:
image

resources/assets/components/utilities/ContentfulAsset/ContentfulAsset.js
mendelB: I think that's fair! 👍

weerd: > and it's unlikely an editor embeds a blocks or asset that don't exist

I think this could happen if it's an entry or asset used in multiple places and the actual resource gets deleted? Granted that doesn't happen much at all, but I think it's possible. Either way, just showing the error block for actual error or no data seems fine with me.

DFurnes: Yeah, totally! Un-publishing something would trigger this as well. Sadly, I think the average user wouldn't take much solace from knowing that. 😂

resources/assets/helpers/text.js
DFurnes: Ooooops, I can't believe I didn't notice that. Reverted that change in fd09ad1!

resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
```diff
@@ -88,7 +73,7 @@ const ContentfulEntryLoader = ({ id, className }) => (
 
       return (
         <ContentfulEntry
-          className={classnames(className, gridClass)}
+          className={classNames(className, gridClass)}
```
weerd: lowercase ⬇️ `classnames`

resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
```diff
@@ -54,30 +52,17 @@ const ContentfulEntryLoader = ({ id, className }) => (
   <Query query={CONTENTFUL_BLOCK_QUERY} variables={{ id }}>
     {({ loading, error, data }) => {
       if (loading) {
-        return (
-          <div className="grid-main spinner -centered margin-vertical-xlg" />
-        );
+        return <div className="spinner -centered margin-vertical-xlg" />;
```
weerd: no longer needs `grid-main`?

DFurnes: Fixed in cc73aea.

resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
DFurnes: Ok, can swap. For what it's worth, [that's how the package spells it](https://www.npmjs.com/package/classnames) in it's docs.

resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
```diff
@@ -2,12 +2,10 @@ import React from 'react';
 import { get } from 'lodash';
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
-import classnames from 'classnames';
+import classNames from 'classnames';
```
weerd: Oh it's capitalized here, but we don't use it like this in all other components 🤷‍♂️

resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
weerd: oh, haha, so we were wrong this whole time!

DFurnes: Changed to be lower-case in 2522f0ce.

mendelB: lolllll this always trips me up!

PR #1330, opened 2019-03-21
Update campaign page footer

ngjo:
### What does this PR do?
Removes `6` from the member count in the campaign footer!

Per staff meeting on 3/20, updating member count to "millions" across site & communications.

Requested in [this](https://docs.google.com/document/d/1kpWstZ18OTnqX5zW4s6m1tDFuvLodD52YSJrw8cAFUg/edit#bookmark=id.ezrnyuy09lpc) document.

Are there other places in the code that mention `6 million`?

### DIFF
--- a/resources/assets/components/pages/LandingPage/LandingPage.js
+++ b/resources/assets/components/pages/LandingPage/LandingPage.js
@@ -67,7 +67,7 @@ const LandingPage = props => {
 
       <div className="info-bar -dark">
         <div className="wrapper">
-          A DoSomething.org campaign. Join over 6 million members taking action.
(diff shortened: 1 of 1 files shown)


mendelB:
> Are there other places in the code that mention 6 million?

😨 [every 11 facts page](https://github.com/DoSomething/phoenix-next/blob/e42812dd73aabfa1a27d20dc3c0a20c343afae48/contentful/management-api-scripts/2018_10_30_001_import_11_facts_from_ashes.js#L51)

mendelB:
Ben also mentioned this IRL and will add this to the doc, we should be able to do this fairly easily with a  management API script

PR #1333, opened 2019-03-22
Post Gallery Block Waypoint Tracking

mendelB:
### What does this PR do?

This PR adds PuckWaypoint tracking to the `PostGalleryBlockQuery`.
Since we needed the ID for tracking data, it felt like a good time to add an `id` prop to the `PostGallery` component as well and pass that through, so we have anchor linking available for the prom gallery. (@weerd I don't _think_ this interferes with your gallery work but pls lemme know if this clashes at all).
Also, not sure if it's better to throw that `id` outside of the `PostGallery` and within the wrapping query?
 
This is one of the few tracking updates we need for the Prom page.

Refs [Pivotal ID #163910248](https://www.pivotaltracker.com/story/show/163910248)

### DIFF
--- a/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
+++ b/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
@@ -41,6 +41,7 @@ const PostGalleryBlockQuery = ({ id, actionIds, className, itemsPerRow }) => (
           waypointData={{ contentfulId: id }}
         />
         <PostGallery
+          id={id}
           className={classnames(className)}
           posts={result}
           loading={fetching}
--- a/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
+++ b/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
@@ -2,6 +2,7 @@ import React from 'react';
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import { PuckWaypoint } from '@dosomething/puck-client';
 
 import PaginatedQuery from '../../PaginatedQuery';
 import PostGallery from '../../utilities/PostGallery/PostGallery';
@@ -26,7 +27,7 @@ const POST_GALLERY_QUERY = gql`
 /**
  * Fetch results via GraphQL using a query component.
  */
-const PostGalleryBlockQuery = ({ actionIds, className, itemsPerRow }) => (
+const PostGalleryBlockQuery = ({ id, actionIds, className, itemsPerRow }) => (
   <PaginatedQuery
     query={POST_GALLERY_QUERY}
     queryName="posts"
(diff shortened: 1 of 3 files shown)


weerd:
> @weerd I don't think this interferes with your gallery work but pls lemme know if this clashes at all

We're defs gonna have some merge conflicts @_@ I think it should be ok for me to figure out though.

mendelB:
> > @weerd I don't think this interferes with your gallery work but pls lemme know if this clashes at all
> 
> We're defs gonna have some merge conflicts @_@ I think it should be ok for me to figure out though.

Ahh dang ok, thanks!

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import { PuckWaypoint } from '@dosomething/puck-client';
```
weerd: Could we place the waypoint inside the `<PostGallery/>` instead? Within the enclosing `<div>` that ends up applying the grid related class? This way the waypoints sit as children of the PostGallery and don't interfere with the grid layout.

mendelB: The reason I tried keeping the waypoint out of the `PostGallery` is that it's rendered from multiple different places, some static, some based on a Contentful entry. So for example in [the Account page](https://github.com/DoSomething/phoenix-next/blob/e42812dd73aabfa1a27d20dc3c0a20c343afae48/resources/assets/components/pages/AccountPage/UserPostsQuery.js#L46) -- I wasn't sure we'd need waypoint tracking since it's always statically rendered on that page and is the only content. So I figured by keeping it out of the component itself, we could track specific galleries, by name, from the containing component which renders the `PostGallery`. 

If we _are_ fine tracking it always though, (and the grid interference issue might be a good reason to be), we could add a `waypointName` prop to the `PostGallery` so we can get still get specifically labeled waypoints for the different galleries?

weerd: Yeah, go for the `waypointName` solution. I'd prefer to keep it as close to the component it's actually trying to track via the waypoint ⚡️

mendelB: Ok great, will do! This way we can also only conditionally render the waypoint so we're only tracking what we need to be 🎉

mendelB:
Update:
- https://github.com/DoSomething/phoenix-next/pull/1333/commits/c3b3c8899e5a0ce5e6c5b31dab1deaa61abb1ecb adds waypoint tracking ability to `PostGallery` controlled by a `waypointName` prop.
- https://github.com/DoSomething/phoenix-next/pull/1333/commits/fc6422fdc503a09f9d0e3ddba4b92ddc1ec12d94 removes waypoint tracking from `PostGalleryBlockQuery` in favor of passing the `waypointName` prop to `PostGallery`

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
```diff
@@ -26,32 +26,38 @@ const POST_GALLERY_QUERY = gql`
...
   >
     {({ result, fetching, fetchMore }) => (
-      <PostGallery
-        className={classnames(className)}
-        posts={result}
-        loading={fetching}
-        itemsPerRow={itemsPerRow}
-        loadMorePosts={fetchMore}
-      />
+      <React.Fragment>
```
weerd: You probs don't need the `React.Fragment` anymore, but I'm adding one in here in my updates for the gallery filtering, so up to you whether you want to remove or not 😄

mendelB: oh lol my Monday mind is fragmented. I'll leave it in here as ritual anticipation for the filter gallery!

PR #1336, opened 2019-03-25
Add (partial) support for more types of blocks via GraphQL.

DFurnes:
### What does this PR do?

This PR adds partial support for loading the Photo Submission, Link, Share, and Voter Registration blocks via `ContentfulEntryLoader`. These each still make assumptions that they're operating within a traditional campaign page (with a campaign ID, etc...), so they're **not fully functional yet**, but this gets us closer to unifying these two separate experiences.

I've used this as an opportunity to clean up some more of our renderers – moving more code into the `ContentfulEntry` itself & moving Puck waypoints into their respective components.

Finally (and hopefully not too controversially), I updated `ContentfulEntryLoader` to wrap the entire `ContentfulEntry` in the provided `className`, rather than requiring each possible component to implement its own custom `className` prop support.

This closes the loop on some work done in DoSomething/graphql#86.

N/A

### Checklist

### DIFF
--- a/resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js
+++ b/resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js
@@ -58,7 +58,8 @@ const DefaultTemplate = props => {
         {affiliateLogo ? (
           <SponsorPromotion
             className="affiliate-logo -padded"
-            imgUrl={affiliateLogo}
+            imgUrl={affiliateLogo.url}
+            title={affiliateLogo.description}
           />
         ) : null}
       </div>
@@ -80,7 +81,8 @@ const DefaultTemplate = props => {
         {affiliateLogo ? (
           <SponsorPromotion
             className="affiliate-logo -padded"
-            imgUrl={affiliateLogo}
+            imgUrl={affiliateLogo.url}
+            title={affiliateLogo.description}
           />
         ) : null}
 
@@ -105,7 +107,10 @@ DefaultTemplate.propTypes = {
   title: PropTypes.string.isRequired,
   content: PropTypes.string,
   link: PropTypes.string.isRequired,
-  affiliateLogo: PropTypes.string,
+  affiliateLogo: PropTypes.shape({
+    url: PropTypes.string,
+    description: PropTypes.string,
+  }),
   buttonText: PropTypes.string,
   campaignId: PropTypes.string,
   userId: PropTypes.string,
--- a/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
+++ b/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
@@ -4,6 +4,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { get, has, invert, mapValues } from 'lodash';
+import { PuckWaypoint } from '@dosomething/puck-client';
 
 import Card from '../../utilities/Card/Card';
 import Modal from '../../utilities/Modal/Modal';
@@ -249,6 +250,10 @@ class PhotoSubmissionAction extends React.Component {
 
     return (
       <React.Fragment>
+        <PuckWaypoint
+          name="photo_submission_action-top"
+          waypointData={{ contentfulId: this.props.id }}
+        />
         <div className="clearfix">
           <div className="photo-submission-action">
             <Card
@@ -394,6 +399,10 @@ class PhotoSubmissionAction extends React.Component {
             </Card>
           </Modal>
         ) : null}
+        <PuckWaypoint
+          name="photo_submission_action-bottom"
+          waypointData={{ contentfulId: this.props.id }}
+        />
       </React.Fragment>
     );
   }
--- a/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
+++ b/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
@@ -250,12 +250,12 @@ class PhotoSubmissionAction extends React.Component {
 
     return (
       <React.Fragment>
-        <PuckWaypoint
-          name="photo_submission_action-top"
-          waypointData={{ contentfulId: this.props.id }}
-        />
         <div className="clearfix">
           <div className="photo-submission-action">
+            <PuckWaypoint
+              name="photo_submission_action-top"
+              waypointData={{ contentfulId: this.props.id }}
+            />
             <Card
               className={classnames('bordered rounded', this.props.className)}
               title={this.props.title}
@@ -373,6 +373,10 @@ class PhotoSubmissionAction extends React.Component {
                 </Button>
               </form>
             </Card>
+            <PuckWaypoint
+              name="photo_submission_action-bottom"
+              waypointData={{ contentfulId: this.props.id }}
+            />
           </div>
 
           {this.props.informationContent ? (
@@ -399,10 +403,6 @@ class PhotoSubmissionAction extends React.Component {
             </Card>
           </Modal>
         ) : null}
-        <PuckWaypoint
-          name="photo_submission_action-bottom"
-          waypointData={{ contentfulId: this.props.id }}
-        />
       </React.Fragment>
     );
   }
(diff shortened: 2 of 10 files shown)


DFurnes:
I'm working on extracting this mega-query in `ContentfulEntryLoader` into fragments that are colocated with their components (like we do with `PostCardFragment` and `ReactionButtonFragment`). This makes it easier to change data dependencies in the future & saves us some extra work crafting prop-types

I'd also like to do another pass at clearing up `margin-horizontal-md` & friends this week.

This pull request is already a little hairy, though, so I think those are best tackled separately!

DFurnes:
Rebased to fix merge conflict in `PostGalleryBlockQuery`, no other changes.

resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js
```diff
@@ -80,7 +81,8 @@ const DefaultTemplate = props => {
         {affiliateLogo ? (
```
mendelB: Would this now always be truthy when rendered from the Phoenix backend, since we manually set the array as a property?

resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
```diff
@@ -75,10 +122,9 @@ const ContentfulEntryLoader = ({ id, className }) => (
       const gridClass = get(entryGridMapping, blockType, 'grid-main');
 
       return (
-        <ContentfulEntry
-          className={classnames(className, gridClass)}
-          json={data.block}
-        />
+        <div className={classnames(className, gridClass)}>
```
mendelB: will

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
```diff
@@ -52,6 +52,7 @@ const PostGalleryBlockQuery = ({ id, actionIds, className, itemsPerRow }) => (
 PostGalleryBlockQuery.propTypes = {
   id: PropTypes.string,
   actionIds: PropTypes.arrayOf(PropTypes.string),
+  actionIds: PropTypes.arrayOf(PropTypes.number),
```
mendelB: I think we've got two `actionIds` 👯👯

resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js
DFurnes: Oh gosh! It sure would.

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
DFurnes: Oh I did a boo-boo when I rebased I think. 🐙

DFurnes: Fixed in 029cd48!

resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js
weerd: Has this been resolved?

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -249,6 +250,10 @@ class PhotoSubmissionAction extends React.Component {
 
     return (
       <React.Fragment>
+        <PuckWaypoint
+          name="photo_submission_action-top"
+          waypointData={{ contentfulId: this.props.id }}
+        />
```
weerd: Can you move the PuckWaypoint inside the `<div className="photo-submission-action">` since it outputs an empty div and this will still mess with the grid once we shift to using PSAs in the grid as well.

weerd: The same way you did in the ShareAction :)

resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
weerd: Not sure if I'm super keen on this, given how much we end up customizing layout and concerned making a change like this so close to Prom release 😬

resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js
DFurnes: No, went to make coffee & then got sidetracked by some other pull requests.

DFurnes: Rest assured I won't merge without fixing this! 😸

resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
weerd: But I also see how it could make things easier... 🤔

DFurnes: Yeah, I don't think we'll ever want a `ContentfulEntry` component that _doesn't_ get wrapped by these classes so it feels like this is the most reasonable place to set it. By requiring each one to implement its own `className` support, we end up spreading that work all around.

weerd: Just debating if by having it a level up it will mess with layout spacing. For example the margin/padding classes that  get applied to the component will now be applied on the inside div as a child of the div that sets the grid, whereas now those classes can dictate aspects of margin/layout along with the grid class. I'll like to maybe hold off on this and test it out a bit more before we dive into this right before Prom launch and maybe end up with misalignments 😬

weerd: Screen Shot 2019-03-26 at 10 55 54 AM

As an example, alignment for the ImagesBlock depends on both the grid class AND the individual classes appended to the component to align properly.

DFurnes: Alrighty, I'll add `className` support to each of these components before merging this.

weerd: I think it's something we can reassess post-prom and as we shift everything over to grid. It's been a bit of a balancing act with components currently using grid and those without. Not ideal, but necessary for transition 😅

DFurnes: For what it's worth, I don't think `ImagesBlock` should be setting its own bottom margin there (and I thought we'd decided that we wanted to pull that out into something the story page handles). Regardless, I'm fine keeping things as they are and we can revisit later.

weerd: Yeah, I agree and we need to rethink things now with the magic of CSS Grid ✨ 👍

DFurnes: We chatted a wee bit in Slack & are going to move ahead with trying to clean this up now, since we have a bit more wiggle room until Prom launches on April 8th. 📆

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
DFurnes: Yeah, that's much better! Fixed in 345068d & confirmed waypoints fire when expected.

resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js
DFurnes: Fixed in 9502251!

resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
DFurnes: Removed the wrapping div in 20f4628; will re-visit in follow-up PR.

PR #1338, opened 2019-03-25
Filtering PostGallery by location.

weerd:
### What does this PR do?

This PR enables filtering a rendered PostGallery by location, and in this particular case, by US locations. I tried making the `SelectLocationDropdown` component versatile for other potential location-based selections (#global or potentially a limited sub-group of US states, for use in a voting related campaign or something), and only importing a package that provides the listing if needed.

Kapture 2019-03-26 at 14 49 32

Upcoming PRs will address showing something suitable if the selected location does not have any items (right now it just collapses the area since nothing is found), and also need to address passing a location via URL params.

Refs [Pivotal ID #164737236](https://www.pivotaltracker.com/story/show/164737236)

### Checklist

### DIFF
--- a/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
+++ b/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
@@ -58,7 +58,7 @@ class PostGalleryBlockQuery extends React.Component {
   };
 
   render() {
-    const { actionIds, className, filterType, itemsPerRow } = this.props;
+    const { actionIds, className, filterType, id, itemsPerRow } = this.props;
 
     return (
       <React.Fragment>
@@ -83,12 +83,14 @@ class PostGalleryBlockQuery extends React.Component {
         >
           {({ result, fetching, fetchMore }) => (
             <PostGallery
+              id={id}
               className={classnames(className)}
               posts={result}
               loading={fetching}
               itemsPerRow={itemsPerRow}
               filterType={filterType}
               loadMorePosts={fetchMore}
+              waypointName={'post_gallery_block'}
             />
           )}
         </PaginatedQuery>
--- a/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
+++ b/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
@@ -42,32 +42,30 @@ class PostGalleryBlockQuery extends React.Component {
   constructor(props) {
     super(props);
 
-    const filterType =
-      this.props.filterType === 'none' ? null : this.props.filterType;
-
     this.state = {
-      filterValue: null,
-      enableFiltering: Boolean(filterType),
+      filterValue: '',
+      filterType:
+        this.props.filterType === 'none' ? null : this.props.filterType,
     };
   }
 
   handleSelect = event => {
     this.setState({
-      filterValue: event.target.value || null,
+      filterValue: event.target.value,
     });
   };
 
   render() {
-    const { actionIds, className, filterType, id, itemsPerRow } = this.props;
+    const { actionIds, className, id, itemsPerRow } = this.props;
 
     return (
       <React.Fragment>
-        {this.state.enableFiltering ? (
+        {this.state.filterType === 'location' ? (
           <div className="grid-narrow margin-bottom-lg">
             <SelectLocationDropdown
               locationList="domestic"
               onSelect={this.handleSelect}
-              selectedOption={this.state.filterValue || ''}
+              selectedOption={this.state.filterValue}
             />
           </div>
         ) : null}
@@ -77,7 +75,7 @@ class PostGalleryBlockQuery extends React.Component {
           queryName="posts"
           variables={withoutNulls({
             actionIds,
-            location: this.state.filterValue,
+            location: this.state.filterValue || null,
           })}
           count={itemsPerRow * 3}
         >
@@ -88,7 +86,6 @@ class PostGalleryBlockQuery extends React.Component {
               posts={result}
               loading={fetching}
               itemsPerRow={itemsPerRow}
-              filterType={filterType}
               loadMorePosts={fetchMore}
               waypointName={'post_gallery_block'}
             />
--- a/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
+++ b/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
@@ -2,9 +2,11 @@ import React from 'react';
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import { UsaStates } from 'usa-states';
 
 import PaginatedQuery from '../../PaginatedQuery';
 import PostGallery from '../../utilities/PostGallery/PostGallery';
+import { withoutNulls } from '../../../helpers';
 import { postCardFragment } from '../../utilities/PostCard/PostCard';
 import { reactionButtonFragment } from '../../utilities/ReactionButton/ReactionButton';
 
@@ -12,8 +14,18 @@ import { reactionButtonFragment } from '../../utilities/ReactionButton/ReactionB
  * The GraphQL query to load data for this component.
  */
 const POST_GALLERY_QUERY = gql`
-  query PostGalleryQuery($actionIds: [Int], $count: Int, $page: Int) {
-    posts(actionIds: $actionIds, count: $count, page: $page) {
+  query PostGalleryQuery(
+    $actionIds: [Int]
+    $location: String
+    $count: Int
+    $page: Int
+  ) {
+    posts(
+      actionIds: $actionIds
+      location: $location
+      count: $count
+      page: $page
+    ) {
       ...PostCard
       ...ReactionButton
     }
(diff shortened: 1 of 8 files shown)


resources/assets/components/utilities/SelectLocationDropdown/SelectLocationDropdown.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class SelectLocationDropdown extends React.Component {
```
weerd: I made this a class component instead of a stateless functional component, because I found it helpful to setup everything in the constructor regarding importing the package to gather list of states and assigning it to the `locations` state property, so it doesn't try to do this every time the selected option changes (I don't actually know if I had made this a functional component, if it would actually instantiate `new UsaStates()` each time, but it did when I had it all inside the `PostGalleryBlockQuery` component. If y'all want me to try it as a functional component I can; just lemme know.

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
```diff
@@ -26,40 +38,80 @@ const POST_GALLERY_QUERY = gql`
...
-          loadMorePosts={fetchMore}
-          waypointName={'post_gallery_block'}
-        />
+        {this.state.enableFiltering ? (
+          <div className="grid-narrow margin-bottom-lg">
+            <SelectLocationDropdown
+              locationList="domestic"
+              onSelect={this.handleSelect}
+              selectedOption={this.state.filterValue || ''}
+            />
```
weerd: Also, I debated doing something like we do with the Quiz in ContentfulEntry:
https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/ContentfulEntry/ContentfulEntry.js#L219-L222

Only load the `SelectLocationDropdown` if `enableFiltering` is true. I'm already dynamically loading the `usa-states` package inside the `SelectLocationDropdown` component. I can try and use the quiz technique here as well if y'all think it'd be good to do so.

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
```diff
@@ -26,40 +38,80 @@ const POST_GALLERY_QUERY = gql`
...
-    variables={{ actionIds }}
-    count={itemsPerRow * 3}
-  >
-    {({ result, fetching, fetchMore }) => (
+class PostGalleryBlockQuery extends React.Component {
+  constructor(props) {
+    super(props);
+
+    const filterType =
+      this.props.filterType === 'none' ? null : this.props.filterType;
```
weerd: Side note, I could eventually see the options from Contentful changing from just `location` to `location:domestic`, `location:international`, `location:east_coast`, to differentiate and thus customize the listing retrieved more appropriately... 🤔

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
```diff
@@ -26,40 +38,80 @@ const POST_GALLERY_QUERY = gql`
...
+            />
+          </div>
+        ) : null}
+
+        <PaginatedQuery
+          query={POST_GALLERY_QUERY}
+          queryName="posts"
+          variables={withoutNulls({
+            actionIds,
+            location: this.state.filterValue,
```
mendelB: Bear with me here and maybe this is just nonsense: but would it at all make sense / make things simpler, to change this to `this.state.filterValue || null`, so that we don't need the conditional `null` when we chenge this value in state, and instead just set it to `''` to begin with?

mendelB: It just took me a minute to figure out how all these came together, not that that necessitates anything

resources/assets/components/utilities/SelectLocationDropdown/SelectLocationDropdown.js
mendelB: I feel like this was the right call 👍

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
weerd: I tried a few things depending on how the console was yelling at me... so when passing `this.state.filterValue` as the value for the `selectedOption`, React yells at you if you pass `null` as the selected value for the `<select>`.

Additionally, if the value for `location` here in the query ends up being `''` (empty string), then the query returns nothing because it's trying to find items by a location of "empty string", so it's helpful here to have it be `null` so using `withoutNull()` removes `locations` as a filter from the query and returns the original results without any filtering. For reference, the scenario here is if you load the gallery, and initially nothing is selected, then select a location like `US-NY`, gallery reloads filtered by that location. Now if you next select "Select a state", you end up passing an empty string as the `filterValue` but all you really want is the full list without location filtering.

I'll add some comments so it's easier to grok when coming back to this code in the future.

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
```diff
@@ -26,40 +38,80 @@ const POST_GALLERY_QUERY = gql`
...
+          count={itemsPerRow * 3}
+        >
+          {({ result, fetching, fetchMore }) => (
+            <PostGallery
+              id={id}
+              className={classnames(className)}
+              posts={result}
+              loading={fetching}
+              itemsPerRow={itemsPerRow}
+              filterType={filterType}
```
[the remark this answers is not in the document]
weerd: Oh good catch. I think it's a vestige from an earlier approach.

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
mendelB: I guess if we're only _rarely_ using the filter it could be a speed win? But it's a pretty low key component anyhoo 🤷‍♂️

resources/assets/components/utilities/SelectLocationDropdown/SelectLocationDropdown.js
```diff
@@ -0,0 +1,67 @@
...
+      prefix: '',
+    };
+
+    if (this.props.locationList === 'domestic') {
+      this.addDomesticOptions();
+    }
+  }
+
+  addDomesticOptions = (config = {}) => {
+    import('usa-states').then(({ UsaStates }) => {
```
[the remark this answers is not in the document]
weerd: It doesn't but it can based on passing in a `config`. Good call, forgot about that! Wondering now what a good string for this would be... `domestic` doesn't seem like it would be obvious it includes all US states and territories... lemme know if you have ideas!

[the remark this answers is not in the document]
mendelB: Hmmm `national`?

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
weerd: Yeah, that was my thinking too. At least we hold on importing the `usa-states` or eventually something like `countries` package unless it's actually needed, so figured that was _good enough for now_™.

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
mendelB: Gotcha, thanks for explaining this!!

I guess my thinking here is that if we only set it to `|| null` here, so when it's an empty string, it gets removed from the query by the `withoutNulls`, but this way it can always be an empty string in the rest of the component? (Since the only place it explicitly needs to be `null` is here to prevent an empty string being passed to the query)

weerd: Oh gotcha. You're basically saying reverse it. I can try that 🙆‍♂️

weerd: Excellent call @mendelB. 👌reads much better now! Thanks!

resources/assets/components/utilities/SelectLocationDropdown/SelectLocationDropdown.js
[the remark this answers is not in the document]
weerd: Gonna leave for now and re-tackle in upcoming PR for allowing filtering via URL params... 🤔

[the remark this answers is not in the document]
weerd: So, since US Customs and Border patrol indicates you don't need a passport to travel to most US territories, I think I might be fine just keeping this as domestic. haha 😄

resources/assets/components/utilities/SelectLocationDropdown/SelectLocationDropdown.js
```diff
@@ -23,6 +24,7 @@ class SelectLocationDropdown extends React.Component {
       });
 
       this.setState({
+        defaultText: 'Select a state',
```
mendelB: Ah is this to enable a different prompt for another location set, like global?

weerd: Yeah. I realized I hard coded it when I was discussing with you the empty string vs `null` and reviewing the code. I had made the `SelectLocationDropdown` component to be flexible based on locations provided, but ended up forgetting that I hardcoded that string, haha.

PR #1345, opened 2019-03-28
Allow filtering and scrolling to gallery via URL params.

weerd:
### What does this PR do?

This PR updates the `PostGalleryBlockQuery` to allow being filtered via URL params. If valid URL params are provided for filtering the gallery, the PostGallery will load entries based on that filter and also scroll to the location of the gallery, with the select dropdown filter in view so that the user has context of what the gallery is showing.

Only valid URL options will trigger the feature. The pattern for specifying the options is as follows:
`?options[post_gallery_contentful_id]=location:US-XX` where `XX` is the 2 character state abbreviation.

Eg:
`?options[7Jf7ViPBQEveJx0bpPDRpV]=location:US-NY`

Here it is in action:
PostGallery URL Param Filter   Scroll

This PR also includes an update to add a very basic "No Results Found" block to help let the user know that no entries were found for the selected state, which is better than showing nothing and having it seem like the experience is broken 😬  

Refs [Pivotal ID #164737236](https://www.pivotaltracker.com/story/show/164737236)

### Checklist

### DIFF
--- a/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
+++ b/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
@@ -1,10 +1,12 @@
 import React from 'react';
 import gql from 'graphql-tag';
+import { split } from 'lodash';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
-import { withoutNulls } from '../../../helpers';
 import PaginatedQuery from '../../PaginatedQuery';
+import ScrollConcierge from '../../ScrollConcierge';
+import { query, withoutNulls } from '../../../helpers';
 import PostGallery from '../../utilities/PostGallery/PostGallery';
 import { postCardFragment } from '../../utilities/PostCard/PostCard';
 import { reactionButtonFragment } from '../../utilities/ReactionButton/ReactionButton';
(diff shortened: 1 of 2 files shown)


weerd:
Also forgot to mention, the scrolling to the dropdown + gallery ended up working pretty great thanks to the **Scroll Concierge**!

Scroll Concierge

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
```diff
@@ -42,26 +44,83 @@ class PostGalleryBlockQuery extends React.Component {
...
       filterValue: event.target.value,
     });
   };
 
+  /**
+   * Callback for when PostGallery has rendered.
+   *
+   * @return {Void}
+   */
+  galleryReady = () => {
```
mendelB: I wonder if we can use this idea more generally to all components loaded via the `ContentfulEntryLoader` to allow anchor linking to work again?

Could perhaps be done when addressing https://www.pivotaltracker.com/story/show/164700372 and our general story page scrolling woes?

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
```diff
@@ -42,26 +44,83 @@ class PostGalleryBlockQuery extends React.Component {
...
+   * @param  {String} filterType
+   * @return {Boolean}
+   */
+  validQueryOptions = (options, filterType) => {
+    const validFilterType = options && options[0] === filterType;
+
+    let validFilterValue = false;
+
+    if (filterType === 'location') {
+      const regex = new RegExp(`^US-[A-Z]{2}$`, 'g');
```
mendelB: What happens if there's input that passes this validation but is not a valid state/territory code?

weerd: Yeah, I'd have to do a check against the list of valid state/territory codes and that felt a bit like overkill. Instead it just fails as potentially expected. It scrolls down to the post gallery, but shows a "No Results Found" block since something like `US-ZZ` is not a valid option in the select list:

image

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
weerd: Oh yeah, potentially! I'll link this comment of yours to the ticket for future reference 👌

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
mendelB: Oh amazing, no sweat!

PR #1349, opened 2019-03-29
Colocate GraphQL fragments with components.

DFurnes:
### What does this PR do?

This pull request colocates GraphQL fragments with the components that render that data, following the same pattern we established with [`PostCard`](https://github.com/DoSomething/phoenix-next/blob/1692e88b7c313be82fdaa6cd9e82b699eb40a780/resources/assets/components/utilities/PostCard/PostCard.js#L17-L40), [`ReactionButton`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/utilities/ReactionButton/ReactionButton.js), and [`PostGalleryBlockQuery`](https://github.com/DoSomething/phoenix-next/blob/1692e88b7c313be82fdaa6cd9e82b699eb40a780/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js#L17-L40).

This makes it easier to find and change data dependencies (since when adding or removing a field, both changes will now happen in one file), and makes `ContentfulEntryLoader` easier to read.

It'd also be smort to use [`graphql-anywhere`](https://github.com/apollographql/apollo-client/tree/master/packages/graphql-anywhere#utilities) to automatically start building prop-types for these components, but that'll get hairy and so probably makes more sense for another PR.

Finally, I think it could be a smart idea to rename relevant blocks to match the GraphQL type names for consistency across the board (e.g. `PhotoSubmissionAction` to `PhotoSubmissionBlock`). But another one that makes sense in its own PR!

N/A

### Checklist

### DIFF
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -9,3 +9,6 @@ global.open = jest.fn().mockReturnValue({ closed: true });
 
 // Mock the ContentfulEntry import to prevent circular dependency issues unique to the jest/node runtime.
 jest.mock('./resources/assets/components/ContentfulEntry/ContentfulEntry');
+jest.mock(
+  './resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader',
+);
--- a/resources/assets/components/actions/LinkAction/LinkAction.js
+++ b/resources/assets/components/actions/LinkAction/LinkAction.js
@@ -1,10 +1,26 @@
 import React from 'react';
 import { get } from 'lodash';
 import PropTypes from 'prop-types';
+import gql from 'graphql-tag';
 import { PuckWaypoint } from '@dosomething/puck-client';
 
 import CtaTemplate from './templates/CtaTemplate';
 import DefaultTemplate from './templates/DefaultTemplate';
+
+export const LinkBlockFragment = gql`
+  fragment LinkBlockFragment on LinkBlock {
+    title
+    content
+    link
+    buttonText
+    affiliateLogo {
+      url(w: 200, h: 100)
+      description
+    }
+    template
+    additionalContent
+  }
+`;
 
 const templates = {
   cta: CtaTemplate,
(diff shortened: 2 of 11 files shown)


mendelB:
> It'd also be smort to use graphql-anywhere to automatically start building prop-types for these components, but that'll get hairy and so probably makes more sense for another PR.

That looks super cool!

> Finally, I think it could be a smart idea to rename relevant blocks to match the GraphQL type names for consistency across the board (e.g. PhotoSubmissionAction to PhotoSubmissionBlock). But another one that makes sense in its own PR!

So we'd be fine with diverging from the Contentful content type names? (I mean, of course we already have with the graphql naming lol) -- Is the idea that these make more sense as 'blocks' on the front end, or/and that we should be suffixing most content types with `block` on Contentful as well?

Sorry for the disjointed question. I guess in a nutshell I'm asking, what is the idea behind the `block` suffix across the board here?

DFurnes:
>  I guess in a nutshell I'm asking, what is the idea behind the `block` suffix across the board here?

Yeah, the idea behind the GraphQL naming swap was to make these fit within our overall system a bit more cleanly (since `Action` already refers to a distinct thing in Rogue, so `PhotoSubmissionAction` could easily be confused for the action itself). Another option could have been to name these `PhotoSubmissionActionBlock`, but that's a mouthful!

I'm not sure we've actually ever formalized this now that you mention it, but I've always considered blocks to be any of these interchangeable interface elements that we use to build pages (like Lego blocks)! So if we accept blocks somewhere (say in the `blocks` field on an action page or as Rich Text embeds), we'd ideally accept _any_ of these elements (but not, say, a `Page`).

> So we'd be fine with diverging from the Contentful content type names? (I mean, of course we already have with the graphql naming lol) -- Is the idea that these make more sense as 'blocks' on the front end, or/and that we should be suffixing most content types with block on Contentful as well?

Yeah, that's for sure the bummer part. I think this will be less of an issue as we move more queries to GraphQL, but it'd be nice to programmatically rename these down the road if we can.

mendelB:
Gotcha, this makes good sense! Thanks for clarifying!!

weerd:
> Another option could have been to name these PhotoSubmissionActionBlock, but that's a mouthful!

I'm kinda fine with it being more verbose if needed, haha

PR #1351, opened 2019-04-01
Pass Contentful referrerD to Northstar. (& General Northstar Data Updates)

mendelB:
### What does this PR do?

This PR updates the data we pass to Northstar during authentication redirects to include the new `referrerId` (Contentful ID) field.

Since this is meant to apply to `StoryPages` as well, this required updating the `selector`s we use for this functionality to include a new general `getDataForNorthstar` selector, in addition to the campaign specific `getCampaignDataForNorthstar`, and brand new `getStoryPageDataForNorthstar` selectors.

Refs [Pivotal ID #164935945](https://www.pivotaltracker.com/story/show/164935945)

### Checklist

### DIFF
--- a/resources/assets/selectors/campaign.js
+++ b/resources/assets/selectors/campaign.js
@@ -19,20 +19,17 @@ export function getCampaignAdditionalContent(state) {
  * @return {Object}
  */
 export function getCampaignDataForNorthstar(state) {
-  const data = {};
-
-  if (get(state, 'campaign.type', null) === 'campaign') {
-    data.callToAction = state.campaign.callToAction;
-    data.coverImage = contentfulImageUrl(
-      state.campaign.coverImage.url,
-      '800',
-      '600',
-      'fill',
-    );
-    data.title = state.campaign.title;
+  if (get(state, 'campaign.type', null) !== 'campaign') {
+    return {};
   }
 
-  return data;
+  const { title, callToAction, coverImage } = state.campaign;
+
+  return {
+    title,
+    callToAction,
+    coverImage: contentfulImageUrl(coverImage.url, '800', '600', 'fill'),
+  };
 }
 
 export default null;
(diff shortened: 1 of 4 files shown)


resources/assets/selectors/storyPage.js
```diff
@@ -0,0 +1,25 @@
...
+export function getStoryPageDataForNorthstar(state) {
+  const data = {};
+
+  if (get(state, 'page.type', null) === 'storyPage') {
+    const { title, subTitle, coverImage } = state.page.fields;
+
+    data.title = title;
+    data.callToAction = subTitle;
+    data.coverImage = contentfulImageUrl(coverImage.url, '800', '600', 'fill');
+  }
```
weerd: Would it be cleaner to flip this logic around, so if it's NOT the `storyPage` then `return data;` otherwise continue w/ logic?

mendelB: Yeah totally, I think so! I was just copying the format from `selectors/campaign.js` but I'll refactor for both.

mendelB: Updated in https://github.com/DoSomething/phoenix-next/pull/1351/commits/50739d55aa1532aa8b31cf00344eff272bd5f154. 
This was a great call @weerd  -- so much cleaner!

resources/assets/selectors/campaign.js
```diff
@@ -19,22 +19,17 @@ export function getCampaignAdditionalContent(state) {
...
 
-  data.trafficSource = query('utm_medium');
+  const { title, callToAction, coverImage } = state.campaign;
 
-  return data;
+  return {
+    title,
+    callToAction,
+    coverImage: contentfulImageUrl(coverImage.url, '800', '600', 'fill'),
+  };
```
weerd: 🎉

PR #1352, opened 2019-04-01
Add Default Redirect Path for Auth Requests From Different Origin

mendelB:
### What does this PR do?

This PR ensures that authentication requests to `/next/login` originating from URL's of a different domain will be redirected post-auth to the default `$this->redirectTo` path and not the previous page from whence they came.

Over at Northstar, password reset form submissions are redirected to `/next/login` to authenticate but are then abruptly returned to the reset password form. Cases such as those (where the previous URL is not of the same domain as the Phoenix application) will be better redirected to a sensible Phoenix location (currently set to `/campaigns`).

Refs [Pivotal ID #164504959](https://www.pivotaltracker.com/story/show/164504959)

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -46,7 +46,9 @@ class AuthController extends Controller
         // page (if logging in to view a page protected by the 'auth' middleware), or the previous
         // page (if the user clicked "Log In" in the top navigation).
         if (! array_has($queryParams, 'code')) {
-            $intended = session()->pull('url.intended', url()->previous());
+            $defaultIntended = is_same_origin(url()->previous()) ? url()->previous() : $this->redirectTo;
+
+            $intended = session()->pull('url.intended', $defaultIntended);
 
             session(['login.intended' => $intended]);
 
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -42,12 +42,14 @@ class AuthController extends Controller
     {
         $queryParams = $request->getQueryParams();
 
-        // Save the post-login redirect for when the user completes the flow: either to the intended
-        // page (if logging in to view a page protected by the 'auth' middleware), or the previous
-        // page (if the user clicked "Log In" in the top navigation).
+        // Save the post-login redirect for when the user completes the flow.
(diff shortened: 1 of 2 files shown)


app/helpers.php
```diff
@@ -443,3 +443,17 @@ function generate_streamed_csv($columns, $records)
...
     return fclose($file);
 }
+
+/**
+ * Determine if URL is of the same origin as the app's URL.
+ *
+ * @param  string $url
+ * @return bool
+ */
+function is_same_origin($url)
```
mendelB: `is_same_domain`? `is_same_host`? Ah, naming... 😩

app/Http/Controllers/AuthController.php
```diff
@@ -42,11 +42,15 @@ public function getLogin(ServerRequestInterface $request, ResponseInterface $res
     {
         $queryParams = $request->getQueryParams();
 
-        // Save the post-login redirect for when the user completes the flow: either to the intended
-        // page (if logging in to view a page protected by the 'auth' middleware), or the previous
-        // page (if the user clicked "Log In" in the top navigation).
+        // Save the post-login redirect for when the user completes the flow.
```
mendelB: Rearranging these comments to have it make sense with the new flow was tricky... Is it any clearer now?

app/helpers.php
weerd: maybe `is_same_domain` is more straightforward? 🤔

mendelB: Yeah I think so! I was about to update that I just saw a similar helper in Northstar with `domain`
https://github.com/DoSomething/northstar/blob/84d387a9d83b5ed6e7bcf05899bd05d687b6808a/app/helpers.php#L203

mendelB: I'm going to add the safe hard return that method has as well.

aaronschachter:
Just checking -- do these changes need corresponding updates in [Northstar](https://github.com/DoSomething/northstar/blob/84d387a9d83b5ed6e7bcf05899bd05d687b6808a/app/Http/Controllers/Web/ResetPasswordController.php#L109)?

mendelB:
@aaronschachter this shouldn't require any updates in Northstar!

PR #1359, opened 2019-04-02
Preview Image Field for Embed Content Type

mendelB:
### What does this PR do?

This PR adds a `previewImage` field to the `Embed` content type migration script.

Refs [Pivotal ID #165039475](https://www.pivotaltracker.com/story/show/165039475)

### DIFF
--- a/contentful/content-types/embed.js
+++ b/contentful/content-types/embed.js
@@ -41,7 +41,7 @@ module.exports = function(migration) {
     .name('Preview Image')
     .type('Link')
     .localized(false)
-    .required(true)
+    .required(false)
     .validations([
       {
         linkMimetypeGroup: ['image'],
@@ -68,6 +68,6 @@ module.exports = function(migration) {
 
   embed.changeEditorInterface('previewImage', 'assetLinkEditor', {
     helpText:
-      'Preview image for the embed content. Replaces the embed on smaller screens.',
+      'If set, replaces the embed on smaller screens as a preview of the embed content.',
   });
 };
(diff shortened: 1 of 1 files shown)


contentful/content-types/embed.js
```diff
@@ -36,6 +36,26 @@ module.exports = function(migration) {
     .disabled(false)
     .omitted(false);
 
+  embed
+    .createField('previewImage')
+    .name('Preview Image')
+    .type('Link')
+    .localized(false)
+    .required(true)
```
weerd: Should it be required?

weerd: Might be embeds that end up fine on smaller screens and thus no preview image is required?

mendelB: I was thinking yes since it's part of the core functionality to be displayed on mobile?

mendelB: Oh good point 🤔

mendelB: Would that require an additional toggle field?

mendelB: Hmm, could we keep it required for now, and remove that if we encounter such a scenario?

weerd: Can we go with the opposite flow? We'll see if we need to make it required?

mendelB: Sure! Would this mean, we'd only conditionally show the image preview on mobile, if the field is set, and otherwise display the embed itself?

DFurnes: Yeah, I think keeping this field optional (with the behavior you just outlined) makes good sense. 👍

mendelB: Updated in https://github.com/DoSomething/phoenix-next/pull/1359/commits/310deebd731ba619c5bd6ec7e5e4f132c4ecafdc (also updated the helper text.)

PR #1360, opened 2019-04-03
Admin Dashboard updates and future possibilities 🔮

weerd:
### What does this PR do?

This PR makes some updates to the new Admin Dashboard to improve the design a tad bit, make it more obvious what it is by adding a heading, and include a new "edit on Contentful" button.

Here's a quick look at how it works!
Admin Dashboard Refactoring

🌵🤠 

### Checklist


app/helpers.php
```diff
@@ -112,6 +133,11 @@ function scriptify($json = [], $store = 'STATE')
     return new HtmlString('<script type="text/javascript">window.'.$store.' = '.json_encode($json).'</script>');
 }
 
+function readableTitle($string)
```
DFurnes: Should this be `readable_title` since it's a helper function?

app/helpers.php
```diff
@@ -35,6 +36,26 @@ function get_cache_url($prefix, $slug = null)
...
+    $contentful = config('contentful');
+    $spaceId = $contentful['delivery.space'];
+    $envronment = $contentful['delivery.environment'];
+
+    $environmentPath = $envronment !== 'master' ? 'environments/'.$envronment.'/' : '';
+
+    return 'https://app.contentful.com/spaces/'.$spaceId.'/'.$environmentPath.'entries/'.$contentfulId;
+}
+
+function get_page_settings($page, $prefix, $slug = null)
```
DFurnes: Want to add some docblocks to these functions before merging?

weerd:
Note, this doesn't apply to campaign pages just yet. I'd like to make an update that shifts a lots of the React based admin controls currently in the existing gray bar and shift them into the new Admin Dashboard via React Portals.

app/helpers.php
```diff
@@ -35,6 +36,26 @@ function get_cache_url($prefix, $slug = null)
...
 
+function get_contentful_edit_url($contentfulId)
+{
+    $contentful = config('contentful');
+    $spaceId = $contentful['delivery.space'];
+    $envronment = $contentful['delivery.environment'];
+
+    $environmentPath = $envronment !== 'master' ? 'environments/'.$envronment.'/' : '';
+
+    return 'https://app.contentful.com/spaces/'.$spaceId.'/'.$environmentPath.'entries/'.$contentfulId;
```
weerd: I was debating if I should put the base Contentful domain name into an env var...

DFurnes: I don't think it needs to be an environment variable (since it doesn't change per environment), but could be nice to pull out into a string in `config/services.php`.

resources/assets/scss/base.scss
```diff
@@ -438,7 +438,118 @@ p {
...
+| Tailwind-esh Class Names
+|-------------------------------------------------------------------------------
+|
+| Class names above this section will likely go to the chopping block once we
+| implement the use of TailWind in Phoenix, so when possible opt to use class
+| names indicated below.
+|
+*/
+
+.bg-gray-100 {
```
weerd: Just FYI, some of this naming for the colors `bg-gray-100` etc, are based on how colors are being defined in Tailwind 1.0.0 beta release.

app/helpers.php
weerd: oh! ya, good catch 👍

app/helpers.php
weerd: Ya! Got a bit excited to push up and forgot. Will do so now!

app/helpers.php
```diff
@@ -443,21 +469,3 @@ function generate_streamed_csv($columns, $records)
...
     return fclose($file);
 }
-
-/**
- * Determine if URL is of the same origin as the app's URL.
- *
- * @param  string $url
- * @return bool
- */
-function is_same_domain($url)
```
mendelB: Assuming this was deleted due to falling behind the latest git commit?

DFurnes: Also, was this intentionally deleted? 🗑

weerd: oh damn. I had to rebase because of some conflicts. Must've gotten lost in the shuffle. Good catch. I'll add back it! Sorry @mendelB 😬

mendelB: lol no offense taken 😁

weerd: Ya, I think so.

app/helpers.php
```diff
@@ -100,6 +136,24 @@ function is_legacy_id($id)
     return is_numeric($id);
 }
 
+/**
+ * Determine if URL is of the same origin as the app's URL.
+ *
+ * @param  string $url
+ * @return bool
+ */
+function is_same_domain($url)
```
weerd: Moved it up a bit in alphabetical ordering :)

PR #1362, opened 2019-04-03
Render preview image

mendelB:
### What does this PR do?

This PR updates the `Iframe` component to use the new `previewImage` prop when rendering the image for smaller screens. If the prop is not set, it will always render the `<iframe>`.

See this discussion https://github.com/DoSomething/phoenix-next/pull/1359#discussion_r271501378 for context for this update!

Refs [Pivotal ID #165039475](https://www.pivotaltracker.com/story/show/165039475)

### Checklist

**That's an image!**
image

### DIFF
--- a/app/Entities/Embed.php
+++ b/app/Entities/Embed.php
@@ -18,7 +18,7 @@ class Embed extends Entity implements JsonSerializable
             'type' => $this->getContentType(),
             'fields' => [
                 'url' => $this->url,
-            'previewImage' => $this->previewImage ? [
+                'previewImage' => $this->previewImage ? [
                     'description' => $this->previewImage->getDescription(),
                     'url' => get_image_url($this->previewImage, 'square'),
                 ] : null,
--- a/app/Entities/Embed.php
+++ b/app/Entities/Embed.php
@@ -18,6 +18,10 @@ class Embed extends Entity implements JsonSerializable
             'type' => $this->getContentType(),
             'fields' => [
                 'url' => $this->url,
+            'previewImage' => $this->previewImage ? [
+                    'description' => $this->previewImage->getDescription(),
+                    'url' => get_image_url($this->previewImage, 'square'),
+                ] : null,
             ],
         ];
     }
--- a/resources/assets/components/utilities/Iframe.js
+++ b/resources/assets/components/utilities/Iframe.js
@@ -3,7 +3,6 @@ import gql from 'graphql-tag';
 import Media from 'react-media';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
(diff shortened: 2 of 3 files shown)


weerd:
Wait, in the screenshot you provided, is it using the `previewImage`?

resources/assets/components/utilities/Iframe.js
```diff
@@ -14,54 +13,50 @@ const PERMITTED_HOSTNAMES = ['dosomething.carto.com'];
...
     url
-  }
-`;
-
-const EMBED_QUERY = gql`
-  query EmbedQuery($url: AbsoluteUrl!) {
-    embed(url: $url) {
-      thumbnailUrl
+    previewImage {
+      url(w: 700, h: 700)
```
mendelB: Is this ok? I went with these dimensions as it seemed suitable for mobile. I think the output here really largely depends on the preview image we're setting.

mendelB:
> Wait, in the screenshot you provided, is it using the `previewImage`?

Yes. Sorry, it's slightly confusing as I took the screenshot on a map rendered in mobile so it should be less wide.

resources/assets/components/utilities/Iframe.js
weerd: What if we only define a width and the height is dependent on the ratio of the image if resized to 700px wide? This way it's more adaptable depending on the size of the image supplied?

DFurnes: @weerd That's how this should work! By default, Contentful image assets will be resized to fit the given "boundaries" (so, say, a 2000x1000 image would resize down to 750x350 given this query). We'd have to specify a different resize strategy for it to crop.

weerd: Ah perfect! Forgot that was the case 🎉

app/Entities/Embed.php
```diff
@@ -18,6 +18,10 @@ public function jsonSerialize()
             'type' => $this->getContentType(),
             'fields' => [
                 'url' => $this->url,
+            'previewImage' => $this->previewImage ? [
```
DFurnes: Huh, I feel like the indentation looks a little wonky but StyleCI didn't complain. Am I seeing things?!

weerd: It does look wonky to me as well...

mendelB: yeah... I almost want to leave it here as a sign of victory over the styleci gods 🥊

mendelB: Reluctantly fixed in https://github.com/DoSomething/phoenix-next/pull/1362/commits/e0a80df900061e228f4f7792ab9ae8e1e4af4ccd

PR #1366, opened 2019-04-08
Scrolling WIP

mendelB:
### What does this PR do?

This PR is a WIP to enable us to scroll the user to a specific location or anchor on the page without worrying about asynchronously loaded content mucking up the scroll location.

It adds a `QueryObserver` service, (that name may need to be workshopped lol) which is meant to keep track of any content/spacing affecting pending graphql queries, or anything other async thing we'd need to track, and allows us to have a definitive idea of when we can start scrolling the user.

One caveat(?) with this approach -- it'd require any content consequential things e.g. `<Queries>` to register manually with this `QueryObserver` (ugh what a bad temp name) and then update it when complete. So it adds an assured amount of responsibility to wherever we render those out. So perhaps not as passive as we'd like? Or is that ok? I was hoping there would be someplace baked in where we could just observe those graphql batched queries from their end but wasn't able to find anything in the docs.

Refs [Pivotal ID #]()

### Checklist

### DIFF
--- a/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
+++ b/resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
@@ -6,9 +6,9 @@ import classnames from 'classnames';
 
 import PaginatedQuery from '../../PaginatedQuery';
 import ScrollConcierge from '../../ScrollConcierge';
-import { query, withoutNulls } from '../../../helpers';
 import PostGallery from '../../utilities/PostGallery/PostGallery';
 import { postCardFragment } from '../../utilities/PostCard/PostCard';
+import { query, withoutNulls, queryObserver } from '../../../helpers';
 import { reactionButtonFragment } from '../../utilities/ReactionButton/ReactionButton';
 import SelectLocationDropdown from '../../utilities/SelectLocationDropdown/SelectLocationDropdown';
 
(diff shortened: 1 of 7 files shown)


resources/assets/helpers/createHistoryHashObserver.js
```diff
@@ -74,7 +74,13 @@ export default (history, timeout = 1000) => {
 
       timeoutId = setTimeout(reset, timeout);
     });
+  };
+
+  history.listen(location => {
+    run(location);
   });
 
+  run(window.location);
```
mendelB: This would allow anchor links to work for initial page load as well. (the name `run` will be updated lol)

resources/assets/components/blocks/PostGalleryBlock/PostGalleryBlockQuery.js
```diff
@@ -55,6 +55,8 @@ class PostGalleryBlockQuery extends React.Component {
   constructor(props) {
     super(props);
 
+    queryObserver().add('posts');
```
mendelB: I've only had to add it for this one so far for this to be working since this gallery loads last on the page from graphql.

mendelB:
This was an ...interesting idea 🥃 

I think there are some bits and pieces here which might actually be sane and useful (https://github.com/DoSomething/phoenix-next/pull/1366#discussion_r273065832), and the general issue is still an issue. But I'll officially call an end to this particular experiment at this juncture.

PR #1372, opened 2019-04-08
Add Request ID and Other Info to Logs

mendelB:
### What does this PR do?

This PR adds some pertinent identifying information to logs to enable tracking across apps/lifecycle

This is a much belated follow up to #1218 (see https://github.com/DoSomething/phoenix-next/pull/1218#issuecomment-452379803 and that thread in general)

**Question**: Do we need to explicitly set the [`use ForwardsTransactionIds;`](https://github.com/DoSomething/gateway/blob/master/docs/traits/ForwardsTransactionIds.md) in our Client configurations? (I didn't see this set in Rogue anywhere).

### DIFF
--- a/app/Services/Bertly.php
+++ b/app/Services/Bertly.php
@@ -4,10 +4,12 @@ namespace App\Services;
 
 use DoSomething\Gateway\AuthorizesWithApiKey;
 use DoSomething\Gateway\Common\RestApiClient;
+use DoSomething\Gateway\ForwardsTransactionIds;
 
 class Bertly extends RestApiClient
 {
     use AuthorizesWithApiKey;
+    use ForwardsTransactionIds;
 
     /**
      * The header to provide the API key in.
--- a/app/Services/Bertly.php
+++ b/app/Services/Bertly.php
@@ -8,8 +8,7 @@ use DoSomething\Gateway\ForwardsTransactionIds;
 
 class Bertly extends RestApiClient
 {
-    use AuthorizesWithApiKey;
-    use ForwardsTransactionIds;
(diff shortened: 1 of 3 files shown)


DFurnes:
> **Question**: Do we need to explicitly set the [`use ForwardsTransactionIds;`](https://github.com/DoSomething/gateway/blob/master/docs/traits/ForwardsTransactionIds.md) in our Client configurations? (I didn't see this set in Rogue anywhere).

For custom-built clients like our `RogueClient` in this app, yes! Anything that's built-in with Gateway, like the Northstar client, should already have that trait included by default.

mendelB:
https://github.com/DoSomething/phoenix-next/pull/1372/commits/223bfc28116aa7ad9a76e4e78129c1d36eb40c28 Updated our custom-built Rogue and Bertly clients to use the `ForwardsTransactionIds` trait.

app/Services/Bertly.php
```diff
@@ -4,10 +4,12 @@
 
 use DoSomething\Gateway\AuthorizesWithApiKey;
 use DoSomething\Gateway\Common\RestApiClient;
+use DoSomething\Gateway\ForwardsTransactionIds;
 
 class Bertly extends RestApiClient
 {
     use AuthorizesWithApiKey;
+    use ForwardsTransactionIds;
```
DFurnes: If you want, these can be combined into one line! e.g:

```php
use AuthorizesWithApiKey, ForwardsTransactionIds;
```

mendelB: Oh neato! Imma update just for kicks!

PR #1376, opened 2019-04-11
Fix low resolution logos on retina displays.

weerd:
### What does this PR do?

This PR fixes a small bug with how sponsor logos are showing up on retina screens. The logos look fine on non-retina screens, but on retina screens they look a little blurry. So to fix this, we've doubled the resolution of the logo, while keeping the visual height the same to ideally make them look crisper on retina screens.

🌵 👁 

Refs [Pivotal ID #165263073](https://www.pivotaltracker.com/story/show/165263073)

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -292,7 +292,7 @@ function get_image_url($asset, $style = null)
 
     $options['logo'] = (new ImageOptions)
         ->setFormat('png')
-        ->setHeight(50)
+        ->setHeight(100)
         ->setResizeFit('scale');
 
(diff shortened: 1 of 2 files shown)


weerd:
cc @lkpttn @ddonizeth

weerd:
Also just noting here. We eventually want to remove the image processing we do on the PHP (entity) side of things and leave this for the client. But this was a last minute request and would prefer not to do the work involved for a cleaner fix in a rush!

For now, just went with this simple change. I checked and this only really affects logos also on the Campaign Update and maybe one other component, but still works as expected!

PR #1377, opened 2019-04-12
Report Failed api GET Requests to New Relic

mendelB:
### What does this PR do?

This PR ensures we report failed ('caught') API GET requests made using our API middleware to New Relic (and/or whichever analytics is configured) using the `report` helper.

I wasn't seeing these in our NR's browser dashboard and figured this would be good to be reporting (especially given our open issue re `500`s to the `.../signups` API endpoint)

I didn't add a general analytics event as we have for the [POST requests](https://github.com/DoSomething/phoenix-next/blob/80cc4393d180161076c2c6433619ea9623be3b4c/resources/assets/middleware/api.js#L114-L123) -- do you think that'd be helpful too?

Refs [Pivotal ID #165315689](https://www.pivotaltracker.com/story/show/165315689)

### Checklist

### DIFF
--- a/resources/assets/middleware/api.js
+++ b/resources/assets/middleware/api.js
@@ -31,6 +31,8 @@ const getRequestAction = (payload, dispatch) => {
       });
     })
     .catch(error => {
+      report(error);
+
       if (window.ENV.APP_ENV !== 'production') {
         console.log('🚫 failed response? caught the error!', error);
       }


weerd:
I think you made the right call in not adding a general analytics event to the GET. They make more sense for the POST since it could be related to failed Reportback posts which is a bigger deal.

PR #1379, opened 2019-04-12
Update to persist query params on campaign urls.

weerd:
### What does this PR do?

This PR fixes an issue where query parameters in URLs to campaign would be lost when redirecting the campaign to the `/action` page (there is a separate issue to address this behavior and eventually remove it). This update fixes the issue by passing along the `location.search` query params as a prop in the `Redirect` ReactRouter component.

Initially approached this by setting up a new helper function to handle relative URLs with query params (our `makeUrl()` helper only works well with absolute URLs), but then @mendelB and his very inquisitive mind found that the ReactRouter Redirect component can have a `search` prop passed to it to append query params in a string. Made the whole thing a lot easier.

This also cleans up the doc block in the `makeURL()` helper function.

Refs [Pivotal ID #165323440](https://www.pivotaltracker.com/story/show/165323440)

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5554,27 +5554,27 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+          "resolved": false,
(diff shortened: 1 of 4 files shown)


mendelB:
@weerd What's with the package.lock changes - is that just npm versions being different?

mendelB:
> @mendelB and his very inquisitive mind

gif

weerd:
> @weerd What's with the package.lock changes - is that just npm versions being different?

oh yeah, dang. I keep doing that with my iMac. It was missing some package, so needed to `npm install`

weerd:
Removed the `package-lock.json` changes!

PR #1380, opened 2019-04-15
Open external links from rich text in new tab

mendelB:
### What does this PR do?

This PR opens links in a new tab for external URI's provided in a Rich Text field on Contentful.

This ensures parity with the current functionality in [Markdown fields](https://github.com/DoSomething/phoenix-next/blob/b96abb3a0df3a61ee8caa05f54a9f5b8b8b34dac/resources/assets/helpers/text.js#L56-L58), where external URl's are set to open in a new tab.

You may notice we only conditionally set the `rel` to `noopener noreferrer` as well. It felt like this might be helpful so as to not block the `document.referrer` being set for internal links. (I noticed similar functionality on NYT articles.) 
I don't *think* it's an issue to have an empty `rel` attribute for internal links, right?

Refs [Pivotal ID #165226084](https://www.pivotaltracker.com/story/show/165226084)

### Checklist

### DIFF
--- a/resources/assets/helpers/text.js
+++ b/resources/assets/helpers/text.js
@@ -136,11 +136,20 @@ export function parseRichTextDocument(document, styles) {
           <ContentfulAsset id={node.data.target.sys.id} />
         </p>
       ),
-      [INLINES.HYPERLINK]: node => (
-        <a href={node.data.uri} style={{ color: hyperlinkColor }}>
-          {node.content[0].value}
-        </a>
-      ),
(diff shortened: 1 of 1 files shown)


weerd:
> I don't think it's an issue to have an empty rel attribute for internal links, right?

Not sure but I suspect it's fine 🤷‍♂️

weerd:
Looked into it and the [MDN docs on the `target` attribute and associated notes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target) was helpful!

PR #1383, opened 2019-04-17
creates SoftEdgeWidgetAction Entity

chloealee:
### What does this PR do?

This PR adds the SoftEdgeWidgetAction Entity. 

Refs [Pivotal ID first task in #165399561](https://www.pivotaltracker.com/n/projects/2019313/stories/165399561/comments/201731342)

### Checklist

### DIFF
--- a/app/Entities/SoftEdgeWidgetAction.php
+++ b/app/Entities/SoftEdgeWidgetAction.php
@@ -15,6 +15,7 @@ class SoftEdgeWidgetAction extends Entity implements JsonSerializable
     {
         return [
             'id' => $this->entry->getId(),
+            'type' => $this->getContentType(),
             'fields' => [
                 'title' => $this->title,
                 'softEdgeId' => $this->softEdgeId,
--- a/app/Entities/SoftEdgeWidgetAction.php
+++ b/app/Entities/SoftEdgeWidgetAction.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace App\Entities;
+
+use JsonSerializable;
+
+class SoftEdgeWidgetAction extends Entity implements JsonSerializable
+{
+    /**
+     * Convert the object into something JSON serializable.
+     *
+     * @return array
+     */
+    public function jsonSerialize()
+    {
+        return [
+            'id' => $this->entry->getId(),
+            'fields' => [
+                'title' => $this->title,
+                'softEdgeId' => $this->softEdgeId,
+            ],
+        ];
+    }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
     },
     "@babel/generator": {
       "version": "7.0.0-beta.44",
-      "resolved": "http://registry.npmjs.org/@babel/generator/-/generator-7.0.0-beta.44.tgz",
+      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.0.0-beta.44.tgz",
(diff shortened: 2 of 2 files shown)


chloealee:
@weerd @mendelB is there anything I should be testing here?

app/Entities/SoftEdgeWidgetAction.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    /**
+     * Convert the object into something JSON serializable.
+     *
+     * @return array
+     */
+    public function jsonSerialize()
+    {
+        return [
+            'id' => $this->entry->getId(),
```
mendelB: One important field we add to these Entities is the `type` field which is primarily used in React to figure out which component to load based on the Entity type.

Here's an example from the Photo Submission Action -- you can just copy paste!
https://github.com/DoSomething/phoenix-next/blob/1215c727d4168d07d2f329500a707e7c5020b246/app/Entities/PhotoSubmissionAction.php#L13

weerd: oh ya! Good call @mendelB 👍

chloealee: ah super helpful, thanks @mendelB ! I was wondering about that actually and thought it was the `type` that we chatted about potentially adding later if we use more SoftEdge tools beyond email. This is good to know the distinction though, thank you!

mendelB: Ohh right dang that can be confusing! Might even make sense to potentially name that field `actionType` or something down the pike to avoid confusion!

weerd: @chloealee one way to easily tell is if `$this->` is calling a property or a method. If it's a method then it's likely unrelated to the fields on the content type (keeping in mind sometimes we make our own methods in the entities to help manipulate the data extracted from a field). But general rule 😄

weerd:
> 💯 (Do you think it's worth git co-ing the package-lock.json file so we don't get those diffs? (I'm never quite sure if it actually makes a difference tbh 🤷‍♂️))

Ya, there's a mismatch in one of our Node versions. @chloealee either see if `git checkout package-lock.json` works to restore it back to what it was... or just git checkout master; copy the package-lock.json from there; checkout your branch again and paste it; then push up. I've had to take this approach in the past when the easier `git checkout` of the specific file didn't work out as well as I thought... 🤷‍♂️

app/Entities/SoftEdgeWidgetAction.php
chloealee: ah that is true, thank you @weerd ! So it seems like it is calling [this method](https://github.com/DoSomething/phoenix-next/blob/02196706243890df64f274c81a639b035b9286c6/app/Entities/Entity.php#L178) which is just
```
return $this->entry->getContentType()->getId();
```
is the above `getContentType()` something that comes with Contentful and it looks up the type by ID? Also, when creating the content type, is this "type" the name of the content type we created?

weerd: Yep yep. The "id" for the content type is different then the "id" for the entry. The ID for the entry is the long-ass mongo alphanumeric ID. The ID for the content type is what we called it, and in this case it would be `softEdgeWidgetAction`.

chloealee: got it, thank you!!

PR #1384, opened 2019-04-17
404 Page Admin Tools (Cache Clearing!)

mendelB:
### What does this PR do?

This PR adds the admin tooling dashboard (with cache clearing!) to 404 pages. 

https://github.com/DoSomething/phoenix-next/pull/1384/commits/0c127a3cb54ee04cdc964b108935956bea69b13e Also adds a fallback route declaration to ensure unknown routes which `404` run through our web middleware. 

This will only run for 404 pages rendered via the `ModelNotFoundException` -- usually induced by a Contentful query yielding a 'not_found' result e.g. https://github.com/DoSomething/phoenix-next/blob/5eea35378d94a12f16f562b8dd1db35c18a004c4/app/Repositories/CampaignRepository.php#L206-L208 

🎥 
[Rough video walkthrough](https://cl.ly/d06ee9e83e8c)

Refs [Pivotal ID #165249312](https://www.pivotaltracker.com/story/show/165249312)

### Checklist


app/Providers/AppServiceProvider.php
```diff
@@ -72,6 +72,36 @@ public function boot(Router $router)
...
+                // Otherwise, we presume a path of ':category/:slug'.
+                $content_type = get_content_type_by_category($path[0]);
+                // For campaigns, the :category will not be part of the actual entry's slug (used in the cache ID).
+                $slug = $content_type === 'campaign' ? $path[1] : $path[0].'/'.$path[1];
+            }
+
+            $view->with('admin', [
+                'page' => [
+                    'cacheUrl' => get_cache_url($content_type, $slug),
+                    'type' => request()->path(),
```
mendelB: We could change this to 'Not Found' or 'Unknown' unless showing the path is fine?

app/Providers/AppServiceProvider.php
```diff
@@ -72,6 +72,36 @@ public function boot(Router $router)
...
+                    'editUrl' => get_contentful_url().'entries?searchText='.$slug,
```
mendelB: This felt like a nice way to make the process of troubleshooting an entry 404-ing easier. It'll direct to the contentful search page using the slug as the search text.

app/Providers/AppServiceProvider.php
```diff
@@ -72,6 +72,36 @@ public function boot(Router $router)
             ]);
         });
 
+        // Attach admin tool variables to 404 pages rendered via the ModelNotFoundException.
```
mendelB: Iiiinteresting. I think I was very wrong about this only being rendered via a `ModelNotFoundException ` (and really via the [`NotFoundHttpException`](https://github.com/DoSomething/phoenix-next/blob/09491e5644ae3702b4085591eeeec094b39cdac0/app/Exceptions/Handler.php#L72-L75)).

I was working under the impression, that only non-routing related `404`s were hitting this View Composer, which I thought was strange, but assumed was just part of how routing `404`s were being handled. The reason I was so sure, was because the Admin tool bug was only showing up for entry related `404`s (e.g. `/us/campaigns/nonsense` and _not_ routing `404`s e.g. `/us/nonsense/nonsense`).

But it finally occurred to me that that was a lazy assumption, and I double checked, and *am* seeing this View Composer being hit even on routing related `404`s (which makes more sense).

[See this video walkthrough](https://cl.ly/75852a020951)

However, the reason the admin bug wasn't showing up is surprising:

For routing level `404`'s it seems even a logged in user would display as not logged in!
This means that on the front end when initializing the React app - we wouldn't show the admin tool since they're not authenticated https://github.com/DoSomething/phoenix-next/blob/09491e5644ae3702b4085591eeeec094b39cdac0/resources/assets/init.js#L75-L78

My hunch is that this is because a non-routing level `404` would indeed hit one of our controllers, and get the `AuthorizeRequest` treatment https://github.com/DoSomething/phoenix-next/blob/09491e5644ae3702b4085591eeeec094b39cdac0/app/Http/Controllers/Controller.php#L12 and thus would get authenticated while routing level `404`'s would not since they never hit any of our Controllers?
If this is correct -- is this something we'd want to correct?

weerd:
Still walking through the updates, but quick thought regarding the contentful URL and the ability to "edit on contentful". If it's a not found page that is cached, we should probably _not_ provide a link to edit it on contentful, since technically that just doesn't exist. Instead just clear the cache and if it then finds a page after cache clear, provide the link... 🤔 thoughts?

mendelB:
> Still walking through the updates, but quick thought regarding the contentful URL and the ability to "edit on contentful". If it's a not found page that is cached, we should probably _not_ provide a link to edit it on contentful, since technically that just doesn't exist. Instead just clear the cache and if it then finds a page after cache clear, provide the link... 🤔 thoughts?

@weerd addressed this here https://github.com/DoSomething/phoenix-next/pull/1384#discussion_r276446049 😄 
My thought was that linking to a contentful search could make it easier and more convenient to troubleshoot a page that we think should be working. (I'm basing this on the use cases that have come up for this tool in the past). Do you think it's too much? Or that it's unexpected for this link to be to a search and not to the actual entry?

weerd:
Ah, sorry. Missed that comment 👀 now :)

weerd:
> Or that it's unexpected for this link to be to a search and not to the actual entry?

I think more it's that it's unexpected for this link to head to search based on the phrasing of the link text. I think it's probably ok for now, and I _do_ dig the idea behind it in trying to be helpful, but may be nicer to rephrase the link text to something that more corresponds with what is actually happening. But not vital for now.

mendelB:
> > Or that it's unexpected for this link to be to a search and not to the actual entry?
> 
> I think more it's that it's unexpected for this link to head to search based on the phrasing of the link text. I think it's probably ok for now, and I _do_ dig the idea behind it in trying to be helpful, but may be nicer to rephrase the link text to something that more corresponds with what is actually happening. But not vital for now.

Ah very fair. Might be nice vLater to have an optional `editUrlLabel` variable to override that text.

app/Providers/AppServiceProvider.php
weerd: Mention this in the discussion. I dig it, but would be nice to somehow flag in the UI that you're not editing the page, but heading to try and find the intended page in Contentful's system. Again, not vital for now, since this 404ing isn't a regular occurrence.

app/Providers/AppServiceProvider.php
weerd: 🤔 oh interesting find. What's the proposal on attempting to correct it?

mendelB: https://github.com/DoSomething/phoenix-next/pull/1384/commits/0c127a3cb54ee04cdc964b108935956bea69b13e added fallback route solution which runs unknow routes through middleware (and avoids the view composer)

🎩 💁‍♂️ @weerd for finding https://themsaid.com/laravel-55-better-404-response-20170921

app/Providers/AppServiceProvider.php
weerd: yeah, maybe `Not Found`. We could also use this in the admin-dashboard.blade.php to limit what is see to just a "clear cache" and "search on Contentful" links and that's it. Instead of trying to use this info to fill out the Admin Dashboard view with stuff that doesn't make sense since the page doesn't exist 🤔

weerd: For example:

```php
@if($admin['page']['type'] === 'Not Found')
  // show limited links just for a not found page
@else 
  // show the usual admin dash stuff.
@endif
```

Might be a more elegant way to do it, but this is the general gist.

mendelB: Added in https://github.com/DoSomething/phoenix-next/pull/1384/commits/606f6fde1123ed47c2eb91be890126e6edd3e5ac

app/Providers/AppServiceProvider.php
```diff
@@ -72,6 +72,9 @@ public function boot(Router $router)
...
+        View::composer('errors::404', 'App\Http\ViewComposers\NotFoundComposer');
```
weerd: 🎉

routes/api.php
```diff
@@ -42,4 +42,10 @@
 
     // Signups
     $this->get('/signups', 'Api\SignupsController@index');
+
+    // Unknown Route Fallback
+    // Ensures JSON response. Complementing fallback in web.php).
```
weerd: single closing parenthesis `)`.  Might not even really need the second line anyhoo.

mendelB: oops lol (not quite sure how I missed that

PR #1389, opened 2019-04-23
Sixpack Experiment conversion documentation.

weerd:
### What does this PR do?

This PR expands on how the `SixpackExperiment` component converts a member on a specified test for a Sixpack A/B experiment within Phoenix.

Also decided to place all the Sixpack feature docs within a dedicated `features/sixpack` directory so it's easier to peruse the doc files in a code editor when updating.

[Current updates](https://dosomething.gitbook.io/phoenix-documentation/v/doc-sixpack-experiment-conversion/development/features/sixpack-a-b-testing) for review visually.

🌵 

Refs [Pivotal ID #165256862](https://www.pivotaltracker.com/story/show/165256862)


docs/development/features/sixpack/sixpack-a-b-testing.md
```diff
@@ -54,6 +54,26 @@ When a `SixpackExperiment` component is rendered on a page, upon mounting the fo
...
+{% endhint %}
+
+Within Phoenix, conversions can be triggered by a variety of actions, like a _campaign signup_, a _reportback post_ or clicking a button on a _call to action_ block.
+
+Thus, the following describes how a user is converted on a test in Phoenix within the context of a campaign signup action, but can be applied to other actions and their respective user interface element interaction.
+
+While on a Campaign landing page, with a Sixpack A/B test experiment running, upon clicking the button to signup for the campaign, the following series of steps occur:
+
+1. When clicked, the `SignupButton` component calls the `clickedSignupAction()` function from `/resources/assets/actions/signup.js`.
+2. Along with executing the signup process, the `clickedSignupAction()` function calls the `convertOnSignupAction()` function, which dispatches an action to trigger converting Sixpack experiments on signup; the action payload specifies the `conversion` as `signup`.
```
mendelB: One thing I just realized is that we dispatch two seperate `sixpackExperiments: {conversion: 'signup'}` payloads:
1. from the `convertOnSignupAction` indicated above
2. from the `storeCampaignSignup` method https://github.com/DoSomething/phoenix-next/blob/2e0e128239455af7b35458c88f1e9b23363b847a/resources/assets/actions/signup.js#L145

I'm not sure if this affects the conversion rate or not (I assume Sixpack won't convert you twice on the same test? 🤷‍♂️ ) but I imagine we shouldn't have this duplication anyways?

weerd: Oh right! This actually answers a question for me I noted to look into regarding why the console was showing 2 Sixpack logs for conversion simultaneously. Defs something I'll look into and clean up. I think we prefer only 1 per conversion type, but you're right that it doesn't necessarily harm the sixpack results. But defs something that is confusing and I'd prefer to clear up! Thanks for noting @mendelB

docs/development/features/sixpack/sixpack-a-b-testing.md
```diff
@@ -54,6 +54,26 @@ When a `SixpackExperiment` component is rendered on a page, upon mounting the fo
...
+3. Next, the `sixpackExperimentMiddleware` catches the dispatched action and it checks for a `payload.meta.sixpackExperiment.conversion` property within the action payload. If it finds a conversion specified on the property, it proceeds to run the `convertOnAction()` method from the `Sixpack` service class.
+4. The `convertOnAction()` method converts all available experiments within the `experiments` list of registered experiments for a page that match a specified convertable action. For example, this will convert all experiments on the page that specify conversion on a "signup" action.
```
mendelB: Could it be worth considering only allowing *one* test to convert per conversion type? (Maybe even not allowing additional tests to be activated if the selected conversion type is already activated on the page?) I'm not completely sure how this would work -- but might be helpful to ensure that there are no conflicting experiments being performed which could render the results faulty?

docs/development/features/sixpack/sixpack-a-b-testing.md
mendelB: Ah awesome, glad to clear up that mystery! 🎉

docs/development/features/sixpack/sixpack-a-b-testing.md
weerd: Yeah, I think we ended up deciding to leave it up to the discretion of the individual/PM involved with the campaign or testing initiative. I think it's ok to have a couple tests on a page and ideally we don't want tests conflicting but I think it would get a little crazy and be confusing to pops tests in and not see them converting because some other test is blocking it 😖 We'll probs need to see how well this works and later reassess and see if it's working as we intended.

PR #1394, opened 2019-04-25
Affiliates Content Type Migration Script (New UTM Label Field)

mendelB:
### What does this PR do?

This PR adds a Contentful Migration script for the Affiliate content type, and most importantly - adding a new UTM Label field (https://github.com/DoSomething/phoenix-next/pull/1394/commits/e485722b455a95ee169b19850c7b6ed72f42f243)!

Adding this field so that we can query affiliates responsibly using the UTM query parameter values for referring affiliate scholarship partners. (See [this Pivotal comment](https://www.pivotaltracker.com/story/show/165299110/comments/201990521) (as well as the attached card) for more context.)

Refs [Pivotal ID #165299110](https://www.pivotaltracker.com/story/show/165299110)

### Checklist

### DIFF
--- a/contentful/content-types/affiliate.js
+++ b/contentful/content-types/affiliate.js
@@ -84,7 +84,7 @@ module.exports = function(migration) {
         },
 
         message:
-          'The UTM Label field can only contain lower case letters or numbers and must be snake cased (http://bit.ly/2Pxkxv9).',
+          'The UTM Label field can only contain lower case, alphanumeric characters and must be snake cased (http://bit.ly/2Pxkxv9).',
       },
     ])
     .disabled(false)
--- a/contentful/content-types/affiliate.js
+++ b/contentful/content-types/affiliate.js
@@ -60,7 +60,47 @@ module.exports = function(migration) {
     .omitted(false)
     .linkType('Asset');
 
+  affiliates
+    .createField('utmLabel')
+    .name('UTM Label')
+    .type('Symbol')
+    .localized(false)
+    .required(false)
+    .validations([
+      {
+        unique: true,
+      },
+      {
+        size: {
+          max: 50,
+        },
+
+        message: 'The UTM Label field may not exceed 50 characters.',
+      },
+      {
+        regexp: {
+          pattern: '^[a-z0-9_]*$',
+          flags: null,
+        },
+
+        message:
+          'The UTM Label field can only contain lower case letters or numbers and must be snake cased (http://bit.ly/2Pxkxv9).',
+      },
+    ])
+    .disabled(false)
+    .omitted(false);
+
+  affiliates.changeEditorInterface('internalTitle', 'singleLine', {
+    helpText:
+      'This title is used internally to help find this content in the CMS. It will not be displayed anywhere on the website.',
+  });
+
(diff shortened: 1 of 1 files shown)


mendelB:
Initially, it felt nice to avoid uppercase characters in the UTM label, but found one example of a partner with upper case letters in the [UTM doc](https://docs.google.com/spreadsheets/d/18q06uD7lR7HMA_u9qPIph9_OuAYEb8cNp-MYRkK9BPs/edit#gid=27798981&range=71:71) so added that in the validation. https://github.com/DoSomething/phoenix-next/pull/1394/commits/82f108ac21b556d34f8c13eee1dee036dfd958be

contentful/content-types/affiliate.js
```diff
@@ -0,0 +1,106 @@
+module.exports = function(migration) {
+  const affiliates = migration
+    // Sigh at this very OG content type using a pluralized id :)
```
DFurnes: 😂

contentful/content-types/affiliate.js
```diff
@@ -0,0 +1,106 @@
...
+      {
+        size: {
+          max: 50,
+        },
+
+        message: 'The UTM Label field may not exceed 50 characters.',
+      },
+      {
+        regexp: {
+          pattern: '^[a-zA-Z0-9_]*$',
```
DFurnes: I wonder if it's better to require lower-case here for searchability? (So even if that one UTM you mentioned came in with an upper-case character, we'd treat it as if it were lower-case when looking for it on the backend?)

contentful/content-types/affiliate.js
weerd: oh, past us. 🤦‍♂️

contentful/content-types/affiliate.js
```diff
@@ -0,0 +1,106 @@
...
+    .validations([
+      {
+        unique: true,
+      },
+      {
+        size: {
+          max: 50,
+        },
+
+        message: 'The UTM Label field may not exceed 50 characters.',
```
weerd: Just curious if this is a specification of UTMs somewhere? Defs agree on having a max size!

contentful/content-types/affiliate.js
weerd: Yeah, I dig that. We can just lower-case the value before sending it as part of query on Contentful.

mendelB: Ah good call! I'll revert that commit.

contentful/content-types/affiliate.js
mendelB: I wasn't able to find anything definitive offhand, this was just my gut call!

contentful/content-types/affiliate.js
mendelB: Reverted, and updated the helper text in https://github.com/DoSomething/phoenix-next/pull/1394/commits/4ea714c8919dda39b4169df335e34ff1a84e11d2

contentful/content-types/affiliate.js
weerd: To be honest, I think I worked on it and likely my doing. Oh the shame... 😢

PR #1395, opened 2019-04-25
Cache anonymous requests for a little while.

DFurnes:
### What does this PR do?

This pull request adds a `Cache-Control: max-age=300, public` header to anonymous responses for the homepage, individual campaign page, and pages. This [tells Fastly that it should cache these responses](https://docs.fastly.com/guides/basic-concepts/how-caching-and-cdns-work) for up to 5 minutes (chosen as a hopefully nice "sweet spot" between no caching & a noticeable delay for editors).

It also extends the built-in `StartSession` middleware to _only_ start sessions for authenticated users. This is a nice scalability win since we don't need to persist sessions for users who don't log in & also makes it possible to easily vary whether or not to read a cached page from Fastly based on whether or not the request has an auth cookie.

When we tackle the "site speed" work down the road, I'd love to continue iterating on this & potentially use [surrogate keys](https://docs.fastly.com/guides/purging/getting-started-with-surrogate-keys) alongside Contentful webhooks to cache these for longer & automatically clear when any content on a rendered page changes.

References [#165515770](https://www.pivotaltracker.com/story/show/165515770) and DoSomething/internal#478.

### Checklist


app/Http/Controllers/CampaignController.php
```diff
@@ -60,7 +60,7 @@ public function show($slug, Request $request)
     {
         $campaign = $this->campaignRepository->findBySlug($slug);
 
-        return view('app', [
+        return response()->view('app', [
```
DFurnes: I had to change these to use the base `response()` factory in order for the macro to work. Otherwise it'd try to look for a macro on the `ViewFactory`, where I didn't have access to the actual response to attach a header.

app/Http/Controllers/CampaignController.php
```diff
@@ -69,8 +69,9 @@ public function show($slug, Request $request)
             // We render social metatags server-side because Facebook & Twitter
             // do not render JavaScript when crawling pages like Google does.
             'socialFields' => get_campaign_social_fields($campaign, $request->url()),
-        ])->with('state', [
-            'campaign' => $campaign,
-        ]);
+            'state' => [
```
DFurnes: This is subjective, but I inlined these in with the rest of the variables because I thought it made the chained `->cacheable()` method easier to see (and both approaches are effectively doing the same thing).

DFurnes: Another approach we might consider here would be to chain these `with` calls... maybe more readable?

```php
return response()
    ->view('app')
    ->with('campaign', $campaign)
    ->with('socialFields', get_campaign_social_fields($campaign, $request->url()))
    ->with('state', ['campaign' => $campaign])
    ->cacheable();
```

mendelB: 🤷‍♂️ I'm personally fine with these inlined (I guess the advantage was the distinctive 'Window' state being grouped separately?) but your readability idea looks nice!

mendelB:
> When we tackle the "site speed" work down the road, I'd love to continue iterating on this & potentially use surrogate keys alongside Contentful webhooks to cache these for longer & automatically clear when any content on a rendered page changes.

🤩

app/Http/Controllers/CampaignController.php
DFurnes: Ok, I think I'm going to leave it as-is unless @weerd has strong feels!

app/Providers/AppServiceProvider.php
```diff
@@ -22,6 +24,17 @@ public function boot(Router $router)
         // @see: https://laravel-news.com/laravel-5-4-key-too-long-error
         Schema::defaultStringLength(191);
 
+        // A simple $response->cacheable() macro:
+        Response::macro('cacheable', function ($seconds = 5 * 60) {
+            // If this is an anonymous request, this macro tells Fastly to cache for
+            // N seconds so we can handle sudden traffic spikes, e.g. https://git.io/fjs2W
+            if (Auth::guest()) {
+                $this->setPublic()->setMaxAge($seconds);
```
weerd: Just curious of the flow in a scenario like the following:

1. The homepage is accessed anonymously and cached in Fastly for 5 minutes based on above rule.
2. Editor changes order of items on homepage within Contentful and publishes the changes.
3. The head to homepage to see it and don't see the change.
4. They log in to use the Admin Dash to clear the cache...
5. ?

When the editor logs in and goes to homepage are they seeing the cached homepage served by Fastly or the cached homepage on Phoenix?

Does this question even make sense? 🤷‍♂Haha, not sure if I'm phrasing it correctly. Let me know if you need me to clarify further...

app/Http/Controllers/CampaignController.php
weerd: ooo, I kinda really dig chaining them all 🤔

app/Providers/AppServiceProvider.php
DFurnes: If we set things up correctly, logged-in users should not see (page) content that's cached in Fastly so it'd only be confusing if they were visiting the site in an incognito tab. I think that may require some tweaks in Laravel so we can vary based on whether or not there's a `laravel_session` cookie, though... still looking into it. 🤔

app/Http/Controllers/CampaignController.php
DFurnes: Cool! I'm good with that. 🔨

app/Providers/AppServiceProvider.php
weerd: K. Sounds good. Thanks for looking into it!

app/Http/Controllers/CampaignController.php
DFurnes: Ugh, maybe not. It seems like `with()` only exists on the actual `View` (from `view('app')`), not on the `Request` (from `request()->view('app')`). And I don't believe we can set headers on a `View`. Oh well. 😩

DFurnes:
My original thought was that we'd be able to apply a "force pass" rule in Fastly for authenticated sessions (if the `laravel_session` cookie exists), so that logged-in users wouldn't receive the cached anonymous response when visiting a campaign/article etc. (like we [do for backend requests](https://github.com/DoSomething/infrastructure/blob/aa01365eab80e891c429447ce9f077431d66b0cc/dosomething/fastly-backend/main.tf#L80-L90)).

This turned out to be less straightforward than I'd hoped since by default [Laravel always sets a `laravel_session` cookie, even for anonymous users](https://git.io/fjsFO), and so we have to override the `StartSession` middleware to change that. I went through quite a few iterations before settling on a fairly clean implementation, but on the bright side I am _quite_ familiar with how Laravel handles sessions now. 😂

DFurnes:
I've tested this on my local (inspecting whether the session cookie or appropriate `Cache-Control` header is set) and everything seems to work well. The next step is to deploy it to dev/QA & then add a "force pass" rule to our Fastly properties for users with the `laravel_session` cookie set.

app/Providers/AppServiceProvider.php
DFurnes: Ooof, this was tricker than I'd expected! I extended the built-in `StartSession` middleware so that the `laravel_session` (or equivalent) cookie is only set when a user is logged-in & now we should be able to easily vary caching rules based on this cookie in Fastly.

app/Http/Controllers/CampaignController.php
weerd: hokay. no worries.

app/Http/Controllers/CampaignController.php
```diff
@@ -69,8 +69,9 @@ public function show($slug, Request $request)
...
+                'campaign' => $campaign,
+            ],
+        ])->cacheableWhenAnonymous();
```
weerd: I like the more straightforward name 👌

app/Http/Middleware/StartSession.php
```diff
@@ -0,0 +1,36 @@
...
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, \Closure $next)
+    {
+        // If we don't have a session cookie & aren't on a route that requires keeping a
+        // persistent session between anonymous requests, use the 'array' driver.
+        $isAnonymousSession = ! $request->cookies->has(config('session.cookie'));
+        if ($isAnonymousSession && ! $request->is($this->mustPersistSession)) {
+            config(['session.driver' => 'array']);
```
weerd: @DFurnes just curious _why_ switching to the `array` driver over the `file` driver is best. Am I correct in understanding that using the `array` will keep it for the request cycle and not be persistent between requests (what your comment implies?)...

DFurnes: Correct! The idea is that we don't want a persistent session (either stored as a file or in Redis) for anonymous folks, so we choose the `array` driver which just keeps it in memory for the lifetime of that particular request.

This saves us some space (since we only persist a session for the subset of users who log in) & makes it easier to differentiate whether a request can be served from the Fastly cache (based on the `laravel_session` cookie).

weerd: Awesome! Sounds good and thanks for the added infoz!

PR #1387, opened 2019-04-22
Scholarship Card (AffiliateScholarshipBlock) component

mendelB:
### What does this PR do?

This PR adds a new `AffiliateScholarshipBlock` component to render scholarship and partner (affiliate) information.

(Some of this will be updated a bit once we introduce the actual Affiliate Contentful query).

📸 
- [**Large**](https://user-images.githubusercontent.com/12417657/56612682-f6d9e480-65e2-11e9-9ac9-eebc30d78cd9.png)

- [**Medium**](https://user-images.githubusercontent.com/12417657/56612720-09ecb480-65e3-11e9-84cc-1d93f9239607.png)

- [**Small**](https://user-images.githubusercontent.com/12417657/56612752-1cff8480-65e3-11e9-863e-d6f3d65d8631.png)

**Updated Mock**
Screen Shot 2019-04-22 at 1 58 13 PM

https://app.goabstract.com/projects/fadf0790-f1f9-11e6-9a30-b54ba7e8e705/branches/4a80d4a1-1f34-4bfb-b764-0295a6a71851/collections/0f68e651-cf6a-4005-a101-8fc83937c39d

### What are the relevant tickets/cards?

Refs [Pivotal ID #165299110](https://www.pivotaltracker.com/story/show/165299110)

### Checklist

* [ ] Documentation added for new features/changed endpoints.
* [ ] Added appropriate feature/unit tests.
* [x] Added screenshot to PR description of related front-end updates on **small** screens.
* [x] Added screenshot to PR description of related front-end updates on **medium** screens.
* [x] Added screenshot to PR description of related front-end updates on **large** screens.

PR #1405, opened 2019-05-01
Affiliate Scholarship Block (Querying & Rendering)

mendelB:
### What does this PR do?

This PR adds more fleshed out rendering and logic to display the Affiliate Scholarship card!

- https://github.com/DoSomething/phoenix-next/commit/6981b83b106c1c52a9961deb67d7fa9c54ced5b9 Adds an `AffiliateScholarshipBlockQuery` which attempts to find the corrosponding Affiliate entry from graphQL and renders the `AffiliateScholarshipBlock`.
- https://github.com/DoSomething/phoenix-next/commit/6bb685ccdac4661bc928cc436e921497ee88df4d, https://github.com/DoSomething/phoenix-next/commit/504d47ea03158e7044576b726fa068a0550cf619, https://github.com/DoSomething/phoenix-next/commit/504d47ea03158e7044576b726fa068a0550cf619, https://github.com/DoSomething/phoenix-next/commit/29b87d4b1fe689f9c19e72dbc88069719d858284 and https://github.com/DoSomething/phoenix-next/commit/de3af7ae26e15762c7f3028b7d5944005df2a2e0 update and fully flesh out the `AffiliateScholarshipBlock` component to support the proper affiliate `props`, scholarship props, and `className`
- https://github.com/DoSomething/phoenix-next/commit/5453d88c1bc7d6b92d9f7a99b22be1e39cd1039c adds a new helper to fetch the UTM Label - if applicable - from the referring scholarship affiliate via the UTM params.
- https://github.com/DoSomething/phoenix-next/commit/e941e01fa1d7c7b31bfea5e3e0e8e73f7d1f6518 Updates the `PitchPage` to render out the `AffiliateScholarshipBlockQuery` if we have the necessary information to do so.

Refs [Pivotal ID #165599975](https://www.pivotaltracker.com/story/show/165599975)
https://github.com/DoSomething/graphql/pull/104

### Checklist

Visually, this should be the same as it was in #1387

### DIFF
--- a/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
+++ b/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
@@ -38,7 +38,9 @@ const AffiliateScholarshipBlock = ({
         <div className="scholarship-information__amount">
           <p className="font-bold">AMOUNT</p>
           <p className="scholarship-amount league-gothic margin-top-none">
-            ${scholarshipAmount.toLocaleString()}
+            {scholarshipAmount
+              ? `$${scholarshipAmount.toLocaleString()}`
+              : 'N/A'}
           </p>
         </div>
 
@@ -46,9 +48,11 @@ const AffiliateScholarshipBlock = ({
           <div className="scholarship-deadline">
             <p className="font-bold">DEADLINE</p>
             <p className="margin-top-sm">
-              {format(scholarshipDeadline, 'MMMM do, YYYY', {
-                awareOfUnicodeTokens: true,
-              })}
+              {scholarshipDeadline
+                ? format(scholarshipDeadline, 'MMMM do, YYYY', {
+                    awareOfUnicodeTokens: true,
+                  })
+                : 'N/A'}
             </p>
           </div>
 
--- a/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
+++ b/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { format } from 'date-fns';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 
 import Card from '../../utilities/Card/Card';
 
@@ -9,10 +10,16 @@ import './affiliate-scholarship-block.scss';
 const AffiliateScholarshipBlock = ({
   affiliateLogo,
   affiliateTitle,
+  className,
   scholarshipAmount,
   scholarshipDeadline,
 }) => (
-  <Card className="rounded bordered padded affiliate-scholarship-block">
+  <Card
+    className={classNames(
+      'rounded bordered padded affiliate-scholarship-block',
+      className,
+    )}
+  >
     {affiliateLogo ? (
       <img
         className="affiliate-logo"
@@ -69,6 +76,7 @@ AffiliateScholarshipBlock.propTypes = {
     url: PropTypes.string.isRequired,
     description: PropTypes.string,
   }),
+  className: PropTypes.string,
   scholarshipAmount: PropTypes.number.isRequired,
   scholarshipDeadline: PropTypes.string.isRequired,
 };
@@ -76,6 +84,7 @@ AffiliateScholarshipBlock.propTypes = {
 AffiliateScholarshipBlock.defaultProps = {
   affiliateTitle: null,
   affiliateLogo: null,
+  className: null,
 };
 
 export default AffiliateScholarshipBlock;
--- a/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
+++ b/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
@@ -11,63 +11,57 @@ const AffiliateScholarshipBlock = ({
   affiliateTitle,
   scholarshipAmount,
   scholarshipDeadline,
-}) =>
-  affiliateLogo &&
-  affiliateTitle &&
-  scholarshipAmount &&
-  scholarshipDeadline ? (
-    <Card className="rounded bordered padded affiliate-scholarship-block">
-      {affiliateLogo ? (
-        <img
-          className="affiliate-logo"
-          src={affiliateLogo.url}
-          alt={affiliateLogo.description || 'Affiliate logo'}
-        />
-      ) : null}
+}) => (
+  <Card className="rounded bordered padded affiliate-scholarship-block">
+    {affiliateLogo ? (
+      <img
+        className="affiliate-logo"
+        src={affiliateLogo.url}
+        alt={affiliateLogo.description || 'Affiliate logo'}
+      />
+    ) : null}
 
-      <p className="margin-top-sm padding-bottom-md">
-        <strong>
-          Welcome to DoSomething.org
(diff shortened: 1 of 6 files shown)


resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
```diff
@@ -1,72 +1,90 @@
...
-      <div className="scholarship-information">
-        <div className="scholarship-information__amount">
-          <p className="font-bold">AMOUNT</p>
-          <p className="scholarship-amount league-gothic margin-top-none">
-            ${scholarshipAmount.toLocaleString()}
+    <div className="scholarship-information">
+      <div className="scholarship-information__amount">
+        <p className="font-bold">AMOUNT</p>
+        <p className="scholarship-amount league-gothic margin-top-none">
+          {scholarshipAmount ? `$${scholarshipAmount.toLocaleString()}` : 'N/A'}
```
mendelB: I wasn't sure if this was necessary, but I added some logic to ensure that if this component is rendered without these props it won't break. These props are required, but at least this way we have fallback safety. Is this overkill?

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * The GraphQL query to load data for this component.
+ */
+const AFFILIATE_QUERY = gql`
+  query AffiliateQuery($utmLabel: String!, $preview: Boolean!) {
+    affiliate(utmLabel: $utmLabel, preview: $preview) {
+      title
+      logo {
+        description
+        url(w: 300, h: 100)
```
mendelB: Any smaller width and the logo tends to look blurry on a big screen! I added a styling rule for this img to ensure it doesn't stretch out too much.

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js
```diff
@@ -0,0 +1,61 @@
...
+      preview: env('CONTENTFUL_USE_PREVIEW_API'),
+    }}
+  >
+    {({ loading, data }) => {
+      if (loading) {
+        return <div className="spinner -centered" />;
+      }
+
+      const title = get(data, 'affiliate.title');
+      const logo = get(data, 'affiliate.logo');
```
mendelB: These safety `get`s ensure that the Scholarship block can still be rendered with the scholarship info if we don't yield an affiliate result, or if the query errors out. Why not, I figured?

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
```diff
@@ -1,72 +1,90 @@
...
+        </p>
+      </div>
+
+      <div className="scholarship-information__details">
+        <div className="scholarship-deadline">
+          <p className="font-bold">DEADLINE</p>
+          <p className="margin-top-sm">
+            {scholarshipDeadline
```
mendelB: See https://github.com/DoSomething/phoenix-next/pull/1405#discussion_r280181293

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
weerd: Curious if it would be better to just not output the `.scholarship-information__amount` div if the `scholarshipAmount` is not available? Not sure if that's more confusing to the user or not, seeing no $$ info, versus seeing "N/A". 🤔

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
weerd: See [#1405 (comment reply)](https://github.com/DoSomething/phoenix-next/pull/1405/files#r280469344) 😏

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js
weerd: Can we just define the height of 100, do it automatically crops the width based on ratio with height of 100. I think forcing the height AND the width is going to bite us on logos that don't conform to that ratio.

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
mendelB: Yeah, that's a good point. I added this more as a safety fallback so the component shouldn't break completely if these values aren't provided since we operate on them both, but this technically really shouldn't display at all if there's no amount/deadline.

Do you think I should just revert to the previous pattern of only outputting the component _if_ these values are provided or refactor as you describe above (which might require a bunch of styling edits too...)?

resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
```diff
@@ -2,30 +2,55 @@ import React from 'react';
...
-      <Card title={sidebarCTA.title} className="rounded bordered">
-        <TextContent className="padded">{sidebarCTA.content}</TextContent>
-      </Card>
+const PitchTemplate = ({
+  content,
+  scholarshipAmount,
+  scholarshipDeadline,
+  sidebarCTA,
+}) => {
+  const scholarshipAffiliate = getScholarshipAffiliateLabel();
```
weerd: Maybe rename `scholarshipAffiliate` to `scholarshipAffiliateLabel` so it follows with what the helper function is named and doing?

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js
mendelB: Ah great call! Updated in https://github.com/DoSomething/phoenix-next/pull/1405/commits/e9a614458168e941a98ca96d76baed6acff47830

resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
mendelB: Smart! donezo! https://github.com/DoSomething/phoenix-next/pull/1405/commits/e7b36332d25df22fd2f59b2e99df7ecf2c4e7ecc

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlock.js
weerd: Let's try seeing how it looks outputting it with only the provided information. It might be a nice clue to editors that the campaign is missing an "amount" and "deadline" in the Campaign fields. I don't think it should affect the styles that much since the requirements will still output, so it'll be fine on small screen and still 2 50% columns on larger screens.

mendelB: Ok! It took some styling finagling, but I think I've got it to work! https://github.com/DoSomething/phoenix-next/pull/1405/commits/3c92303d1fe81ba0d76ce09505fe29a81887760a

image

image


image

image


image

image

weerd: Noice! 🎨

PR #1408, opened 2019-05-03
Pre-populated SoftEdge form

chloealee:
### What does this PR do?
Pre-populates the SoftEdge form with data we have on the user. 
- Adds `SoftEdgeWidgetActionContainer` to get the `userId` from the state.
- Adds the UserQuery` file to get user information based on the `userId` using GQL to pass to the `SoftEdgeWidgetAction` file. 
- Updates the SoftEdge URL based on what user data we have to pre-populate the fields in the form. 
Screen Shot 2019-05-06 at 3 17 08 PM

Refs [Pivotal ID #165399561](https://www.pivotaltracker.com/n/projects/2019313/stories/165399561)

### Checklist

### DIFF
--- a/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
+++ b/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
@@ -25,10 +25,15 @@ class SoftEdgeWidgetAction extends React.Component {
    */
   loadSoftEdgeWidget = () => {
     const softEdgeId = this.props.softEdgeId;
+    let url = '//www.congressweb.com/dosomething/' + softEdgeId;
+
+    console.log('🍎', this.props);
+
+    // + '?acceptAuthor=true&firstName=Chloe';
 
     window.$cweb(() => {
       $cweb(`#congressweb-action-${softEdgeId}`).congressweb({
-        url: `//www.congressweb.com/dosomething/${softEdgeId}`,
+        url: `${url}`,
         responsive: true,
       });
     });
--- a/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
+++ b/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
@@ -4,10 +4,6 @@ import PropTypes from 'prop-types';
 
 import Card from '../../utilities/Card/Card';
 
-// const SoftEdgeWidgetAction = props => (
-//   console.log('🔮', props);
-// );
-
 class SoftEdgeWidgetAction extends React.Component {
   componentDidMount() {
     this.initSoftEdgeScript();
@@ -28,12 +24,24 @@ class SoftEdgeWidgetAction extends React.Component {
    */
   loadSoftEdgeWidget = () => {
     const softEdgeId = this.props.softEdgeId;
-    // const url = '//www.congressweb.com/dosomething/' + softEdgeId;
-    // + '?acceptAuthor=true&firstName=Chloe';
-    console.log('🐶', this.props);
+    let url = `//www.congressweb.com/dosomething/${softEdgeId}?acceptAuthor=true`;
+
+    const prepopulatedFields = {
+      firstName: this.props.user.firstName,
+      lastName: this.props.user.lastName,
+      email: this.props.user.email,
+      mobile: this.props.user.mobile,
+    };
+
+    Object.keys(prepopulatedFields).forEach(key => {
+      if (prepopulatedFields[key] != null) {
+        url += `&${key}=${prepopulatedFields[key]}`;
+      }
+    });
+
     window.$cweb(() => {
       $cweb(`#congressweb-action-${softEdgeId}`).congressweb({
-        url: `//www.congressweb.com/dosomething/${softEdgeId}`,
+        url: `${url}`,
         responsive: true,
       });
     });
@@ -51,11 +59,13 @@ class SoftEdgeWidgetAction extends React.Component {
 SoftEdgeWidgetAction.propTypes = {
   title: PropTypes.string.isRequired,
   softEdgeId: PropTypes.number.isRequired,
+  user: PropTypes.object.isRequired,
 };
 
 SoftEdgeWidgetAction.defaultProps = {
   title: null,
   softEdgeId: null,
+  user: null,
 };
 
 export default SoftEdgeWidgetAction;
--- a/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
+++ b/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
@@ -72,7 +72,7 @@ SoftEdgeWidgetAction.propTypes = {
 SoftEdgeWidgetAction.defaultProps = {
   title: null,
   softEdgeId: null,
-  user: null,
+  user: {},
   userId: null,
 };
 
--- a/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
+++ b/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
@@ -64,12 +64,14 @@ SoftEdgeWidgetAction.propTypes = {
   title: PropTypes.string.isRequired,
   softEdgeId: PropTypes.number.isRequired,
   user: PropTypes.object.isRequired,
+  userId: PropTypes.string.isRequired,
 };
 
 SoftEdgeWidgetAction.defaultProps = {
   title: null,
   softEdgeId: null,
   user: null,
+  userId: null,
 };
 
 export default SoftEdgeWidgetAction;
--- a/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
+++ b/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
@@ -25,19 +25,20 @@ class SoftEdgeWidgetAction extends React.Component {
    */
   loadSoftEdgeWidget = () => {
     const softEdgeId = this.props.softEdgeId;
+    const user = this.props.user;
     let url = `//www.congressweb.com/dosomething/${softEdgeId}?acceptAuthor=true&memberId=${
       this.props.userId
     }`;
 
     const prepopulatedFields = withoutNulls({
-      firstName: this.props.user.firstName,
-      lastName: this.props.user.lastName,
-      email: this.props.user.email,
-      mobile: this.props.user.mobile,
-      address: this.props.user.addrStreet1,
-      city: this.props.user.addrCity,
-      state: this.props.user.addrState,
-      zip: this.props.user.addrZip,
+      firstName: user.firstName,
+      lastName: user.lastName,
+      email: user.email,
+      mobile: user.mobile,
+      address: user.addrStreet1,
+      city: user.addrCity,
+      state: user.addrState,
+      zip: user.addrZip,
     });
 
     Object.keys(prepopulatedFields).forEach(key => {
--- a/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
+++ b/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
@@ -47,7 +47,7 @@ class SoftEdgeWidgetAction extends React.Component {
 
     window.$cweb(() => {
       $cweb(`#congressweb-action-${softEdgeId}`).congressweb({
-        url: `${url}`,
+        url,
         responsive: true,
       });
     });
--- a/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
+++ b/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
@@ -3,6 +3,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import Card from '../../utilities/Card/Card';
+import { withoutNulls } from '../../../helpers';
 
 class SoftEdgeWidgetAction extends React.Component {
   componentDidMount() {
@@ -28,7 +29,7 @@ class SoftEdgeWidgetAction extends React.Component {
       this.props.userId
     }`;
 
-    const prepopulatedFields = {
+    const prepopulatedFields = withoutNulls({
       firstName: this.props.user.firstName,
       lastName: this.props.user.lastName,
       email: this.props.user.email,
@@ -37,12 +38,10 @@ class SoftEdgeWidgetAction extends React.Component {
       city: this.props.user.addrCity,
       state: this.props.user.addrState,
       zip: this.props.user.addrZip,
-    };
+    });
 
     Object.keys(prepopulatedFields).forEach(key => {
-      if (prepopulatedFields[key] != null) {
-        url += `&${key}=${prepopulatedFields[key]}`;
-      }
+      url += `&${key}=${prepopulatedFields[key]}`;
     });
 
     window.$cweb(() => {
--- a/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
+++ b/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
@@ -31,6 +31,10 @@ class SoftEdgeWidgetAction extends React.Component {
       lastName: this.props.user.lastName,
       email: this.props.user.email,
       mobile: this.props.user.mobile,
+      address: this.props.user.addrStreet1,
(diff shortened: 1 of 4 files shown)


chloealee:
Screen Shot 2019-05-03 at 1 52 45 PM

Hmm I have a file (`UserQuery.js`) that isn't getting uploaded and registered for some reason 🤔 Is there anything I'm doing incorrectly here?

chloealee:
thanks to @mendelB who noticed I importing `UserQuery` from the `SoftWidgetAction` file instead of from the `UserQuery` file!

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
```diff
@@ -0,0 +1,27 @@
...
+      mobile
+      email
+    }
+  }
+`;
+
+const UserQuery = ({ userId }) => (
+  <Query query={ACCOUNT_QUERY} queryName="user" variables={{ userId }}>
+    {({ data }) => {
+      return <SoftEdgeWidgetAction {...data} />;
```
chloealee: hmm this doesn't seem to be adding any data to the props and wipes the `title` and `softEdgeId` from props in the `SoftEdgeWidgetAction.js` file... am I doing this correctly? cc @mendelB 

Also, @DFurnes , @mendelB mentioned that you tend to like to keep the query in the component if possible. I chose to put it in a different component so I could pass as props to modify [this url](https://github.com/DoSomething/phoenix-next/pull/1408/files#diff-250047573525dc4217900353cb8b6052R36). Is this the correct approach?

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+
+import SoftEdgeWidgetAction from './SoftEdgeWidgetAction';
+
+const ACCOUNT_QUERY = gql`
```
chloealee: is this a specific query in GraphQL to run to find user data?

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
```diff
@@ -0,0 +1,27 @@
...
+  query AccountQuery($userId: String!) {
```
chloealee: is `query AccountQuery` a specific query in GraphQL to run to find user data?

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
chloealee: oh wait - I cleared cache and now getting the `title` and `softEdgeId` in the props again! But not getting any user data... maybe because the query isn't working in `UserQuery.js`?

mendelB: @chloealee Are you setting the condition around `loading` time to ensure the component is only rendered _if_ the `data` has been returned and there were no errors? (A nice example of this is the `Query` in the `Embed` component https://github.com/DoSomething/phoenix-next/blob/b96abb3a0df3a61ee8caa05f54a9f5b8b8b34dac/resources/assets/components/utilities/Embed/Embed.js#L37-L47)

mendelB: I _think_ what could be happening is that the component is being returned immediately while the query is still loading, and thus no user data is being passed through.

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
mendelB: @chloealee I'm not completely sure about this one and wasn't able to find anything explicitly about this in the docs, but my impression has always been that this is Apollo's way of defining and invoking the actual graphql query - the `user(`... bit from the `Query` component. (Sorry if this is not helpful at all, this is my elongated way of saying I have no idea! lol)

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
chloealee: thanks @mendelB ! Playing around with that now but also just realized I'm still getting the error mentioned [here](https://github.com/DoSomething/phoenix-next/pull/1408#discussion_r280891788) with the props being wiped out :-/

mendelB: @chloealee Ah, I think I see the issue! Since you're only passing through the `...data` as props to the `SoftEdgeWidgetAction` component, it's losing the other props from the Container!

I think you'd need to define the `Query` component to allow general `props` (`const UserQuery = (props) =>`) and _also_ pass through those props to the `SoftEdgeWidgetAction` component:
`return <SoftEdgeWidgetAction {...data} {...props} />;`

mendelB: I _think_ kinda like how it is in [this component](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js#L28-#L53)

chloealee: a ha! Thanks, @mendelB ! I added the props into the `SoftEdgeWidgetActionContainer` to pass to `UserQuery` and then to the `SoftEdgeWidgetAction` but now getting this error:
Screen Shot 2019-05-06 at 11 17 23 AM

It seems like the props aren't being passed still (by the URL, the action is `null`) and also says that `gql` isn't defined but looking at the other files, I think it is set up correctly? Any gotchas here that I am missing?

mendelB: @chloealee The props fed to the container in [`ContentfulEntry`](https://github.com/DoSomething/phoenix-next/pull/1408/files#diff-35056872f13a368c9e668973f09a237aR263) will be implicitly passed through to the component (the only issue previously was that we limited the prop available in `UserQuery` itself to be `userId` by defining it as the singular parameter). 

So you can undo the changes in the container and in the `gql` tag -- those should stay as they were defined before. But then with your new changes to the `UserQuery` component -- I'm betting it should work!

(pls LMK if this is unclear! Happy to hop on a quick call if that would be helpful!)

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
```diff
@@ -0,0 +1,43 @@
...
+      // const { embed } = data;
+
+      if (error) {
+        return <ErrorBlock />;
+      }
+
+      // If an <iframe> code snippet is provided, use that.
+      // Otherwise, fill in our "embed card".
+      // return isLoaded && embed && embed.html ? (
+      return <SoftEdgeWidgetAction {...data} {...props} />;
```
mendelB: @chloealee and lastly I think we need to add a case for when the query is still loading. You can add a `loading` param to this function definition (in addition to `error` and `data` params) and then place a conditional above this to output a loading icon:
```js
if (loading) {
  return <div className="spinner -centered" />;
}
```

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
chloealee: I'm a little confused actually now because I _think_ that the props still just get passed to the `SoftEdgeWidgetAction` and don't go through the `SoftEdgeWidgetActionContainer`?

chloealee: oh oops sorry @mendelB I didn't refresh the page and see your comments! I'll take a look at what you said now and keep playing around !

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
chloealee: no this makes sense, thanks @mendelB ! Tagging @weerd too in case he knows more :)

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
chloealee: awesome, thanks @mendelB !

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
```diff
@@ -25,10 +24,30 @@ class SoftEdgeWidgetAction extends React.Component {
    */
   loadSoftEdgeWidget = () => {
     const softEdgeId = this.props.softEdgeId;
+    let url = `//www.congressweb.com/dosomething/${softEdgeId}?acceptAuthor=true&memberId=${
+      this.props.userId
```
chloealee: The `memberId` part of this is to send SoftEdge the user's `northstar_id` in prep for the direct integration webhook to create a post in our system when a user sends an email.

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
```diff
@@ -0,0 +1,49 @@
...
+      addrState
+      addrZip
+    }
+  }
+`;
+
+const UserQuery = props => (
+  <Query
+    query={ACCOUNT_QUERY}
+    queryName="user"
```
mendelB: Fun fact -- this isn't even a prop! @DFurnes pointed this out a while back https://github.com/DoSomething/phoenix-next/pull/1321#discussion_r266646414 (and I subsequently completely forgot to remove this there 🤒 🤦‍♂ )

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
```diff
@@ -25,10 +24,30 @@ class SoftEdgeWidgetAction extends React.Component {
...
+      lastName: this.props.user.lastName,
+      email: this.props.user.email,
+      mobile: this.props.user.mobile,
+      address: this.props.user.addrStreet1,
+      city: this.props.user.addrCity,
+      state: this.props.user.addrState,
+      zip: this.props.user.addrZip,
+    };
+
+    Object.keys(prepopulatedFields).forEach(key => {
```
mendelB: We've got a pretty cool `withoutNulls` helper method that could be useful here to avoid the conditional!
https://github.com/DoSomething/phoenix-next/blob/6e7888712536b45fd454452edd5bdcc81664a3ca/resources/assets/helpers/index.js#L758-L767

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
```diff
@@ -25,10 +24,30 @@ class SoftEdgeWidgetAction extends React.Component {
...
+      if (prepopulatedFields[key] != null) {
+        url += `&${key}=${prepopulatedFields[key]}`;
+      }
+    });
 
     window.$cweb(() => {
       $cweb(`#congressweb-action-${softEdgeId}`).congressweb({
-        url: `//www.congressweb.com/dosomething/${softEdgeId}`,
+        url: `${url}`,
```
mendelB: Can this just be `url,` since it's already defined as a variable above?
```suggestion
        url,
```

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
```
mendelB: Import ordering 🙃 (we try to always order imports by line length -- shortest to longest)

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
```diff
@@ -25,10 +24,30 @@ class SoftEdgeWidgetAction extends React.Component {
...
+
+    const prepopulatedFields = {
+      firstName: this.props.user.firstName,
+      lastName: this.props.user.lastName,
+      email: this.props.user.email,
+      mobile: this.props.user.mobile,
+      address: this.props.user.addrStreet1,
+      city: this.props.user.addrCity,
+      state: this.props.user.addrState,
+      zip: this.props.user.addrZip,
```
mendelB: Toooootally an arbitrary nitpick - but you could define `const user = this.props.user` above to avoid all these nested invocations 🤷‍♂

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
```diff
@@ -46,11 +65,15 @@ class SoftEdgeWidgetAction extends React.Component {
...
   title: PropTypes.string.isRequired,
   softEdgeId: PropTypes.number.isRequired,
+  user: PropTypes.object.isRequired,
+  userId: PropTypes.string.isRequired,
 };
 
 SoftEdgeWidgetAction.defaultProps = {
   title: null,
   softEdgeId: null,
+  user: null,
```
mendelB: Do you think this should this default to an empty object so that we don't encounter errors in `loadSoftEdgeWidget` in the -edge- case `user` isn't passed to the component?

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
```diff
@@ -0,0 +1,49 @@
...
+      email
+      addrStreet1
+      addrCity
+      addrState
+      addrZip
+    }
+  }
+`;
+
+const UserQuery = props => (
```
mendelB: BTW re the second half of https://github.com/DoSomething/phoenix-next/pull/1408#discussion_r280891788 - the thought occurred to me that keeping this query component separate of the widget component is a good step towards having the widget decoupled from user info - which could allow us to display this on ungated pages if need be. (I mean, I guess that's kinda farfetched, especially since currently the widget does require a Northstar ID, but still... ?)

resources/assets/components/actions/SoftEdgeWidgetAction/UserQuery.js
chloealee: oohh interesting! Removed :)

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
chloealee: oh this is really handy, thanks @mendelB ! Also updated to utilize this :)

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
chloealee: hmm @mendelB when I had it like that, it threw this error:
Screen Shot 2019-05-06 at 4 57 15 PM

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
chloealee: ah you're so right, that is way better! Updated :)

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
chloealee: @mendelB do you mean something like `user: {},` instead?

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
mendelB: Hmm did you add it as just `url` or `url: url`? (I believe the latter would trigger that error since the former is the convenient shorthand for such a scenario).

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
mendelB: Yes totally! I *think* that'd be a nice safe fallback

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
chloealee: a-ha! I did the `url: url` but just tried with `url` and it worked - thank you! Didn't know that was shorthand for that and good to know for the future!

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
chloealee: cool that makes sense to me! Updated! :)

PR #1409, opened 2019-05-03
Updating SixpackExperiment component.

weerd:
### What does this PR do?

This PR includes some updates to the `SixpackExperiment` component. It removes the need for the `SixpackExperimentContainer` component by simplifying the logic for how the experiment name is generated. The name will now just be the supplied `title` prop if a code test, or the Contentful `internalTitle` field. We will be removing the `title` field from the `sixpackExperiment` content type on Contentful since it was a bit redundant.

Refs [Pivotal ID #165801497](https://www.pivotaltracker.com/story/show/165801497)

### Checklist


weerd:
> but I guess this comes with the caveat that experiments will need to be named responsibly to ensure distinction between similar experiments across campaigns

Yep yep! Reminds me that I forgot to mention that for code tests, that responsibility will be up to the implementor (making sure if it's a global test or page specific to include that in the name), and on Contentful we're going to make the `internalTitle` not only `required` but `unique` as well!

PR #1412, opened 2019-05-06
Updates Sixpack to be more flexible and a little simpler.

weerd:
### What does this PR do?

This PR updates the `SixpackExperiment` component to allow for "the absence of an alternative" to be an alternative. That is to say, an Sixpack A/B test can now experiment between having a block be present or missing and deciding which converts better. The component has also now be updated to specify a `control` when running an experiment. If no `control` is supplied, then the assumption is that the `control` is the empty alternative state.

This PR also adds a new `<Empty />` component. We utilize this component to specify the desire to have the lack of content as a test. Having it as a dedicated component makes it easier to insert in Sixpack code-based tests, or include dynamically if a Sixpack Contentful-based test. Overall it just allows us to keep the Sixpack logic within a set of limited options (a React component or a Contentful Entry).

As an example with a Code Test:

```js
// In CampaignPage.js

<SixpackExperiment
  title="Campaign Dashboard Experiment"
  convertableActions={['signup']}
  control={<Empty testName="Without Dashboard" />}
  alternatives={[
    <DashboardContainer
      endDate={{ date: '2019-07-04T04:00:00Z' }}
      content={dashboardContentA}
    />,
    <DashboardContainer
      endDate={{ date: '2019-08-04T04:00:00Z' }}
      content={dashboardContentB}
    />,
  ]}
/>
```
And the resulting experiment in the Sixpack Web UI:
image

Refs [Pivotal ID #165801497](https://www.pivotaltracker.com/story/show/165801497)

### Checklist

### DIFF
--- a/app/Entities/SixpackExperiment.php
+++ b/app/Entities/SixpackExperiment.php
@@ -28,7 +28,7 @@ class SixpackExperiment extends Entity implements JsonSerializable
             'type' => $this->getContentType(),
             'fields' => [
                 'alternatives' => $this->parseBlocks($this->alternatives),
-                'control' => $this->control ? $this->parseBlocks($this->control) : null,
+                'control' => $this->control ? $this->parseBlock($this->control) : null,
                 'convertableActions' => $this->convertableActions,
                 'kpi' => $this->kpi,
                 'title' => $this->internalTitle,
--- a/app/Entities/SixpackExperiment.php
+++ b/app/Entities/SixpackExperiment.php
@@ -28,6 +28,7 @@ class SixpackExperiment extends Entity implements JsonSerializable
             'type' => $this->getContentType(),
             'fields' => [
                 'alternatives' => $this->parseBlocks($this->alternatives),
+                'control' => $this->control ? $this->parseBlocks($this->control) : null,
                 'convertableActions' => $this->convertableActions,
                 'kpi' => $this->kpi,
                 'title' => $this->internalTitle,
--- a/resources/assets/components/utilities/Empty.js
+++ b/resources/assets/components/utilities/Empty.js
@@ -1,0 +1,5 @@
+import React from 'react';
+
+const Empty = () => <React.Fragment />;
+
+export default Empty;
--- a/resources/assets/components/utilities/SixpackExperiment/SixpackExperiment.js
+++ b/resources/assets/components/utilities/SixpackExperiment/SixpackExperiment.js
@@ -30,7 +30,7 @@ class SixpackExperiment extends React.Component {
     const controlAlternative = control || <Empty testName="No Alternative" />;
     const alternativeOptions = [controlAlternative, ...alternatives];
 
-    // Get names of  all test alternatives
+    // Get names of all test alternatives
     const alternativeOptionNames = alternativeOptions.map((item, index) => {
       let testAlternativeName = `Test Alternative ${index + 1}`;
 
@@ -74,7 +74,7 @@ class SixpackExperiment extends React.Component {
    * Get the name for the provided test alternative.
    *
    * @param  {Object} alternative
-   * @return {String}
+   * @return {String|Null}
    */
   getTestAlternativeName = alternative => {
     let testAlternativeName;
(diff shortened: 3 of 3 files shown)


resources/assets/components/utilities/Empty.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const Empty = () => <React.Fragment />;
```
weerd: This seemed like a reasonable way to approach this. Nice thing is, you can still pass props as a regular React component, etc.

weerd:
My main debate is with how to handle things on the Contentful side. Currently, the `control` field is not required, which would mean if only the `alternatives` are filled out, then the experiment would end up with an empty test control. However, if we make the field required, and the `control` has to be explicitly linked, than that would mean we would need to create an `emptyBlock` content type to allow editors to specify they want an empty test as the control, instead of just leaving the field blank. This could mean lots of instances of an empty block and potentially the use of a empty block in unexpected places... 

thoughts? 🤔

DFurnes:
I really like the idea of making `control` optional, so that an experiment where the control is empty is as simple as just leaving that field blank. The issue you raised the other day about one of the alternatives being the "empty" option seems tougher to handle in Contentful... 

Do we have good examples of where an editor would want an "empty" alternative in a Contentful A/B test? If so, would they be okay with the "empty" variation just being labeled as the control? (I _think_ that shouldn't affect results since it's an even traffic split, just makes things a bit harder to read in the Sixpack admin panel.)

resources/assets/components/utilities/Empty.js
DFurnes: Love it!

weerd:
>Do we have good examples of where an editor would want an "empty" alternative in a Contentful A/B test? If so, would they be okay with the "empty" variation just being labeled as the control? (I think that shouldn't affect results since it's an even traffic split, just makes things a bit harder to read in the Sixpack admin panel.)

Yeah, it could be something we try out for now and see how well it works. If it turns out the editors don't want the "control" to be the "no block" test in the experiment, then we may have to resort to providing an "empty" content type.

We actually already have a potential test on board to test the Campaign Dashboard, and seeing if more conversions happen with or without it. Other tests are also being planned!

weerd:
> Ooh tough. So even if control is optional, wouldn't that mean that any unfilled control would result in an 'Empty' alternative, whether that's intended or not? (so even an A/B test between the currently displayed content block and an altered content block would contain an 'Empty' control alternative?) Or would we only add the 'Empty' alternative if only one alternative is provided and the control is empty? 

Well, in a test without an empty block, the "control" would be the default block you actually want to test against. We've _always_ had a "control" in Sixpack experiments; it was just always the first item in the alternatives array. Now we formally define a "control", so instead of putting all alternatives, including the original in the `alternatives` reference array, the original would go in the `control` field.

Empty would only happen when you want to test against the lack of a block and thus would need to leave the `control` empty. The caveat as @DFurnes mentioned is that it would mean moving the "default" original block into the `alternatives` array when wanting to test against an empty block.

mendelB:
Ahaaaa understood, thanks for clarifying! 

If you'll forgive a svedka-esque idea -- how crazy would it be to include a toggle field to 'include empty alternative'? We could completely ignore this toggle if the `control` is empty so we don't add two empty alternatives, but otherwise, this would allow the `control` alternative to be more accurate?

weerd:
Yeah, I was debating something similar... just debating if this would be at all confusing on the editor side. Toggle or Empty content type you can toss in wherever (the control or the alternatives) but potentially muddies up our content instances... which is less confusing/easier? 🤔

mendelB:
Ugh, totes... I feel like the toggle is slightly less editor friendly but far more system friendly?

mendelB:
That sounds like a very reasonable approach!

weerd:
@DFurnes @mendelB do either of you deem this PR as 👍 ? 😬

app/Entities/SixpackExperiment.php
```diff
@@ -28,6 +28,7 @@ public function jsonSerialize()
             'type' => $this->getContentType(),
             'fields' => [
                 'alternatives' => $this->parseBlocks($this->alternatives),
+                'control' => $this->control ? $this->parseBlocks($this->control) : null,
```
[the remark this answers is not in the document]
weerd: Doh! Nice catch @mendelB 🙌 thanks!!

PR #1423, opened 2019-05-13
Update to latest ESLint version & plugins.

DFurnes:
### What does this PR do?

This pull request updates our code to match new & updated linting rules in DoSomething/eslint-config#12. Many of the changes were automatically performed with `eslint --fix`, and the ones I made manually are split into commits based on the rule the change addresses. In a few cases, I've left TODOs where a fix would have been more involved than makes sense for me to add here.

These changes all make sense to me, but of course linting rules are only as useful as we find them! If rules seem like more pain than they're worth, let's discuss & make changes! 😸

Follow-up to [#151738638](https://www.pivotaltracker.com/story/show/151738638).

### Checklist


resources/assets/actions/admin.js
```diff
@@ -1,4 +1,4 @@
-import { SHOW_LANDING_PAGE, SHOW_ACTION_PAGE } from '../actions';
+import { SHOW_LANDING_PAGE, SHOW_ACTION_PAGE } from ".";
```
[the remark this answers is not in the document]
mendelB: And wow, this 'implicit lookup'(?) import style is super new to me! And while certainly cleaner do you think it's less clear -- could the path jumping be confusing? Or I guess the idea is that this is just understood based on the style?

resources/assets/components/Quiz/QuizChoice.js
```diff
@@ -22,10 +22,10 @@ const QuizChoice = props => {
   });
 
   return (
-    <a
+    <button
```
mendelB: Would switching this to a `<button>` have any styling implications? (EDIT: I did a quick check - I think it's affecting the `font-family` ("Source Sans" to "system-ui"))

image

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
```diff
@@ -63,15 +63,14 @@ class SoftEdgeWidgetAction extends React.Component {
 }
 
 SoftEdgeWidgetAction.propTypes = {
-  title: PropTypes.string.isRequired,
   softEdgeId: PropTypes.number.isRequired,
-  user: PropTypes.object.isRequired,
-  userId: PropTypes.string.isRequired,
+  title: PropTypes.string,
+  user: PropTypes.object,
+  userId: PropTypes.string,
```
mendelB: I thought `userId` was indeed a valid required prop for this Component? (Would it make sense to get rid of the default prop instead?)

resources/assets/actions/admin.js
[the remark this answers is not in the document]
DFurnes: Oh, isn't that funny! I wonder if that's just a quirk with the auto-formatter...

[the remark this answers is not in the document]
DFurnes: Mmm, seems to be a [known issue](https://git.io/fjW1Y) with the fixer. Addressed in 6e010425.

resources/assets/components/Quiz/QuizChoice.js
DFurnes: Woah, good catch! 🦅 👁  I think it seems safe to apply our `$base-font-family` to buttons across the board (like we do to other base elements), so I gave that a try in f419b989 & will poke around to see if there are any side-effects.

resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetAction.js
DFurnes: Yeah, I think it depends if we want this to function for anonymous users or not, since [`getUserId`](https://github.com/DoSomething/phoenix-next/blob/d5a276bcf59c55a642303fa5da2f59cc9c7080fc/resources/assets/components/actions/SoftEdgeWidgetAction/SoftEdgeWidgetActionContainer.js#L10) will return `null` if this is viewed anonymously (say, on a story page). I'm fine with making this required if that's the intent!

mendelB: Ah, right, I totally just assumed was that this wasn't just yet geared to work anonymously. @chloealee any thoughts on the gravity of this `userId` prop?

DFurnes: I think you're totally right Mendel! I just tried sending `null` for `userId` and everything exploded. I've updated this component to require this prop in 169837da.

DFurnes: And I just remembered reading [in this thread](https://dosomething.slack.com/archives/C782XBKDM/p1557265137017800?thread_ts=1557264580.016000&cid=C782XBKDM) that this component only works for logged-in folks by design. So defs required!

mendelB: Aha, this makes sense then -- thanks for confirming this!

.eslintrc.js
```diff
@@ -4,30 +4,11 @@ module.exports = {
...
-      {
-        'newlines-between': 'always',
-        groups: [
-          ['builtin', 'external'],
-          ['parent', 'sibling', 'internal', 'index'],
-        ],
-      },
-    ],
+    // For now, disable the following linters:
+    'jsx-a11y/click-events-have-key-events': 'off',
```
DFurnes: Just for future reference – I left [this rule](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/click-events-have-key-events.md) disabled because we haven't really thought much about keyboard interaction on our existing components & so it'd take a bit more work to fix this in a scalable way (for example, when multiple components with keyboard event handlers are on the page).

I definitely think it's worth revisiting at some point, though, which is why I didn't disable it in our shared config!

.eslintrc.js
```diff
@@ -4,30 +4,11 @@ module.exports = {
...
+    'import/no-cycle': 'off',
```
DFurnes: This is another one that I left disabled since [we have a decent workaround](https://www.pivotaltracker.com/story/show/164282294) for issues we'd faced with cyclic dependencies in tests, and it sounds like some of the browser issues may have been [fixed upstream in Babel](https://www.pivotaltracker.com/n/projects/2019313/stories/151384966).

DFurnes:
Going to merge this in! If anyone else notices anything else awry in here, I'm more than happy to fix things up in follow-up pull requests. :v:

PR #1427, opened 2019-05-15
Campaign dashboard migration

weerd:
### What does this PR do?

This PR simply adds the Contentful migration for the `campaignDashboard` content type. It also updates the `CampaignDashboard` component to add a ternary around the `shareHeader` since it is not required.

Refs [Pivotal ID #165795039](https://www.pivotaltracker.com/story/show/165795039)

### DIFF
--- a/contentful/content-types/campaignDashboard.js
+++ b/contentful/content-types/campaignDashboard.js
@@ -1,0 +1,111 @@
+module.exports = function(migration) {
+  const campaignDashboard = migration
+    .createContentType('campaignDashboard')
+    .name('Campaign Dashboard')
+    .description('Override the content in the Campaign Dashboard')
+    .displayField('internalTitle');
+
+  campaignDashboard
+    .createField('internalTitle')
+    .name('Internal Title')
+    .type('Symbol')
+    .localized(false)
+    .required(true)
+    .validations([
+      {
+        unique: true,
+      },
+    ])
+    .disabled(false)
+    .omitted(false);
+
+  campaignDashboard
+    .createField('shareHeader')
+    .name('Share Header')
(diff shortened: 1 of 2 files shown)


contentful/content-types/campaignDashboard.js
```diff
@@ -0,0 +1,111 @@
...
+      },
+    ])
+    .disabled(false)
+    .omitted(false);
+
+  campaignDashboard
+    .createField('shareHeader')
+    .name('Share Header')
+    .type('Symbol')
+    .localized(false)
```
mendelB: Don't we generally set content fields like these to be `localized`? (I mean not that it's likely to matter anytime soon. hashtag global)

contentful/content-types/campaignDashboard.js
```diff
@@ -0,0 +1,111 @@
+module.exports = function(migration) {
+  const campaignDashboard = migration
+    .createContentType('campaignDashboard')
+    .name('Campaign Dashboard')
+    .description('Override the content in the Campaign Dashboard')
```
mendelB: Ah, is the idea that we're going to be outputting a dashboard for all campaigns with default copy and this would serve as the optional override?

contentful/content-types/campaignDashboard.js
weerd: ah, right. yeah, most of these I was inheriting from when this content type was created a long time ago... I'll update. Good 👁 !

contentful/content-types/campaignDashboard.js
weerd: Nope, haha. Lingering description from when first created! 🤦‍♂

PR #1429, opened 2019-05-16
Survey modal component updates

weerd:
### What does this PR do?

This PR makes some incremental updates to the `SurveyModal` component. This is one of a few updates coming up to allow showcasing a Typeform survey on other pages apart from the campaign page. This update allows passing in the typeform survey url as a prop instead of hardcoding it as a constant in the component. It also allows for only passing variables in URL params to the survey if available and doesn't make the Northstar and Campaign IDs a requirement.

There's a bunch of updates I have planned for this component (like removing the container, renaming it, not making it a modal but a component that gets placed inside the modal portal, etc). But taking it in baby steps so that I don't break anything!

### DIFF
--- a/resources/assets/components/pages/SurveyModal/SurveyModal.js
+++ b/resources/assets/components/pages/SurveyModal/SurveyModal.js
@@ -3,7 +3,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { makeUrl } from '../../../helpers';
+import { makeUrl, withoutNulls } from '../../../helpers';
 
 class SurveyModal extends React.Component {
   componentDidMount() {
@@ -19,7 +19,7 @@ class SurveyModal extends React.Component {
       location_pathname: window.location.pathname,
     };
 
-    const url = makeUrl(typeformUrl, typeformQuery);
+    const url = makeUrl(typeformUrl, withoutNulls(typeformQuery));
 
     return (
       <div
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/SurveyModal/SurveyModal.js
```diff
@@ -3,39 +3,43 @@
...
 
   render() {
-    const { northstarId, campaignId } = this.props;
+    const { campaignId, northstarId, typeformUrl } = this.props;
 
     const typeformQuery = {
       northstar_id: northstarId,
       campaign_id: campaignId,
-      origin: window.location.pathname,
+      location_pathname: window.location.pathname,
```
weerd: It was confusing to me that this was called `origin` but was actually referring to the `pathname` of the current url. Talked with Sohaib and it wasn't currently being used by data, so it was ok to rename it. I figured this name made sense since it's in a url for the Typeform modal, and it would specify the location's pathname on where it was placed.

mendelB: Not sure if my response in the ticket got lost (https://www.pivotaltracker.com/story/show/165397335/comments/202693585) but we use this param to redirect the user back to the page. This new name makes more sense, we'd just need to make sure to update it in the survey settings on typeform when we deploy!

weerd: Thanks for the flag @mendelB! Based on this feedback I'm going to rename the property to `redirect_pathname` so it's purpose is even clearer. I'll make sure the change on Typeform side of things happens before deploying 👍

PR #1430, opened 2019-05-17
ModalLauncher component updates

weerd:
### What does this PR do?

This PR updates the `ModalLauncher` component a bit to simplify it. It no longer cares about receiving props for the `userId` or `isAuthenticated`. Instead, if we care that the modal launcher only show up for authenticated users, then we should just wrap it in a conditional that checks against the user's authentication status. This has allows removing the `ModalLauncherContainer` 🎉 

It also moves the `ModalLauncher` into the `utilities/Modal` directory which makes more sense!

There's also a bit of additional cleanup 🛁 

...


weerd:
I removed the `ModalLauncher.test.js` for the time being. Will be restoring it and updating it once this component is fully reworked. A few more upcoming changes are planned.

resources/assets/components/utilities/Modal/ModalLauncher.js
```diff
@@ -29,23 +33,16 @@ class ModalLauncher extends React.Component {
...
 
   shouldSeeModal = () => {
-    const { userId, type, isAuthenticated } = this.props;
-
-    if (!isAuthenticated) {
-      return false;
-    }
+    const { type } = this.props;
 
-    let shouldNotSee = getStorage(`${userId}_hide_${type}`, 'boolean');
```
mendelB: @weerd one notable consequence of this change is that this will effectively reset the currently stored status for users for these modals. So anyone who's local storage is currently set to hide the survey would see it once again (since their traffic distribution setting is presumably still active).

mendelB: I'm not sure how big of a deal we consider that (if at all?), but potentially, one compromise (I don't know if this idea is insane) might be adding some temporary logic to reset any stored legacy value to the new format. We could have this code planted for some reasonable time period until we remove and consider everything clear?

weerd: Yeah, I think it's inevitable if we want to remove the `userId` to allow for anonymous survey responses for the Homepage.

Would rather not need to set ones with and without UserID cause that could potentially expose the survey more frequently to people if it shows before and after logging in 😬

weerd: Maybe in upcoming PR for switching custom storage solution and moving to using `localForage` I can maaaaaybe do a regex-y check where it ignores the userID and mostly looks for a key with the rest of the string name? I'll see how feasible that is, but also don't want to go too crazy... users can just close it if they don't want to answer it.

mendelB: Gotcha, understood, sounds good! I think then we can also get away with removing the support for legacy nps survey below https://github.com/DoSomething/phoenix-next/pull/1430/files#diff-2907c0f5f5bf5922becc1083da0d5807R39

weerd: Nice. Will do in upcoming PRs. What was the legacy NPS survey?

mendelB: Looks like this was the PR where we overhauled the structure a bit and added this legacy check https://github.com/DoSomething/phoenix-next/pull/730/files#diff-f41cf28c2ca512f6ffd7b3d79aa2b239R19

weerd:
Rebased since my last PR that simplified the `SurveyModal` changed a few things that conflicted with this PR.

PR #1432, opened 2019-05-20
Fix bug with missing parameter in redirect url.

weerd:
### What does this PR do?

This PR fixes a bug I introduced in #1431. The `hide_nps_survey` url parameter is being added to the main Typeform URL instead of the redirect URL that the user is sent to after completing the survey. This would mean they get redirected and the survey potentially pops up again due to the missing parameter.

The approach in this PR fixes the bug while ensuring that other URL parameters originally in the URL are preserved.

🌵 

### DIFF
--- a/resources/assets/components/pages/SurveyModal/SurveyModal.js
+++ b/resources/assets/components/pages/SurveyModal/SurveyModal.js
@@ -3,7 +3,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { makeUrl, withoutNulls } from '../../../helpers';
+import { appendToQuery, makeUrl, withoutNulls } from '../../../helpers';
 
 class SurveyModal extends React.Component {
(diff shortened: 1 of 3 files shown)


resources/assets/components/pages/SurveyModal/SurveyModal.js
```diff
@@ -3,18 +3,20 @@
...
 class SurveyModal extends React.Component {
   componentDidMount() {
     window.typeformInit();
   }
 
   render() {
-    const { typeformUrl, queryParameters } = this.props;
+    const { queryParameters, redirectParameters, typeformUrl } = this.props;
+
+    const redirectUrl = appendToQuery(redirectParameters, window.location.href);
```
weerd: While the second param in `appendToQuery()` defaults to `window.location.href`, felt a little clearer to specifically pass it in.

PR #1433, opened 2019-05-21
Adding NPS Survey to Homepage.

weerd:
### What does this PR do?

This PR renames the `SurveyModal` component to `TypeFormSurvey`, so it makes a bit more sense; not necessarily a modal (can potentially appear in other contexts) and very specifically just for TypeForm based surveys.

It also more importantly, adds a modal pop-up on the Homepage for a new NPS survey.

If the user submits or closes either the Homepage NPS survey OR the Campaign NPS survey, it will stop any other NPS surveys from potentially showing; better user experience 😉 

### DIFF
--- a/resources/assets/components/pages/HomePage/HomePage.js
+++ b/resources/assets/components/pages/HomePage/HomePage.js
@@ -81,11 +81,6 @@ class HomePage extends React.Component {
           <section className="container container--sponsors">
             <div className="wrapper">
               <div className="container__block">
-                {/*
-                Workaround for this jsx-a11y bug https://git.io/fN814.
-                @TODO: Update once the eslint-config package is updated (https://git.io/fjejY).
-               */}
-                {/* eslint-disable-next-line jsx-a11y/heading-has-content */}
                 <h4>Sponsors</h4>
                 <ul>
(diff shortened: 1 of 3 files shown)


resources/assets/components/pages/HomePage/HomePage.js
```diff
@@ -62,42 +65,62 @@ class HomePage extends React.Component {
...
-        <section className="container container--sponsors">
-          <div className="wrapper">
-            <div className="container__block">
-              {/*
+          <section className="container container--sponsors">
+            <div className="wrapper">
+              <div className="container__block">
+                {/*
                 Workaround for this jsx-a11y bug https://git.io/fN814.
                 @TODO: Update once the eslint-config package is updated (https://git.io/fjejY).
```
DFurnes: We updated our ESLint plugins in https://github.com/DoSomething/phoenix-next/pull/1423, including the latest version of `jsx-a11y` with this bugfix, so this can likely be removed!

weerd: Existing from prior, but will remove! Thanks for the heads up!

Issue #1437, opened 2019-05-28
Experiment: Front-End Testing


DFurnes:
**tl;dr:** This is a hefty write-up. If you're just interested in the takeaways, I'd recommend skimming the introduction & goals sections and then jumping right down to "recommendation" section at the end.

------------------------------------------------------------------

Phoenix currently uses [Jest](https://jestjs.io) for front-end unit & snapshot testing, covering 22 of about 100 components. Unfortunately, we've not kept pace with tests as we've added more and more functionality to this application & the "time" cost of writing the remaining tests is high.

Further, these unit tests actually don't cover large portions of our flow (such as things that happen in Redux actions or reducers), which means we still rely on manual testing and monitoring, like [Ghost Inspector](https://ghostinspector.com), to find issues.

## Goals

Our "high level" goal is to be able to rapidly iterate & deploy changes to Phoenix without fearing that we'll introduce bugs. Since we don't want to put too much feature work on hold while we write tests, we also want to make sure we get the most "bang for our buck" for the time we invest.

With that in mind, I think our time is best spent adding a core suite of _functional tests_ that run through our core flows (such as signing up & reporting back) as a user would. In particular:

- We want to test **whole website flows**, not just functions or components in isolation.
- We need to be able to **mock any state or functionality in other apps or third-party services** (e.g. Northstar, Rogue or Contentful), so that issues in other systems don't cause these tests to fail. If the tests aren't rock solid, it's tempting to just ignore them when they fail.
- We want it to be **easy and quick to run tests** during development & on a CI service (like Wercker or CircleCI) so that we limit the chance of merging bugs into `master`.

This isn't our first attempt at writing a functional test suite. We'd previously written tests with [Behat](http://behat.org/en/latest/) and [CasperJS](http://casperjs.org) for Ashes, and [Laravel Dusk](https://laravel.com/docs/5.8/dusk) for Rogue. These all suffered from the inability to mock API calls to external services & general flakiness which eroded developer trust.

We want to avoid repeating that scenario!

## Process

[...]

PR #1438, opened 2019-05-28
Add Cypress browser tests.

DFurnes:
### What does this PR do?

**This pull request adds [Cypress](https://www.cypress.io) and a signup flow test suite.** This was my favorite of the tools I evaluated, and the most "full featured" test suite I implemented during my R&D sprint. It's also the tool I'd recommend we use going forward!

cypress-example

Feel free to pull down this branch to test it out on your machine:

```
git fetch origin
git checkout -b experimental-cypress origin/experimental-cypress

npm install
npm run cypress
```

See the "Cypress" section in #1437!

Refs #1437 and [Pivotal #165505563](https://www.pivotaltracker.com/story/show/165505563).

### Checklist


cypress/integration/campaign-signup.js
```diff
@@ -0,0 +1,73 @@
+/// <reference types="Cypress" />
```
DFurnes: This something I first noticed in Cypress's examples, and turns out to be a [triple-slash directive](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) that tells TypeScript (or in my case, my editor) where to pull implicit "types" for this file so [you get nice IntelliSense autocompletes](https://docs.cypress.io/guides/tooling/intelligent-code-completion.html#Features)!

resources/assets/middleware/requiresAuthentication.js
```diff
@@ -18,10 +18,19 @@ const requiresAuthenticationMiddleware = ({ getState }) => next => action => {
...
     localforage.setItem(actionId, action).then(() => {
-      window.location.href = buildLoginRedirectUrl(
+      const redirect = buildLoginRedirectUrl(
         getDataForNorthstar(state),
         actionId,
       );
+
+      // If we're running our test suite, don't automatically initiate
+      // the login redirect flow & leave something to assert on.
+      if (window.Cypress) {
```
DFurnes: This `window.Cypress` object is [injected by Cypress when running tests](https://docs.cypress.io/api/commands/window.html#Start-tests-when-app-is-ready) and so we can use it to conditionally skip the cross-domain redirect that would otherwise break our tests (and instead give us something to "hook onto" in our `handleLogin` test helper).

DFurnes:
Alright, this is ready for review!

I've configured CircleCI to build the application ([context](https://github.com/DoSomething/infrastructure/issues/139)), run our existing PHPUnit & Jest tests, and start up a server & run Cypress tests. These tests will then "report" to Cypress' [Dashboard service](https://www.cypress.io/dashboard), which lets us view screenshots & video recordings of tests for debugging issues.

Once we've merged this, I'll disable our Wercker integration since it's no longer needed.

DFurnes:
Oops! 🦋

.env.example
```diff
@@ -2,6 +2,7 @@ APP_NAME=DoSomething
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
+APP_DEBUGBAR=false
```
weerd: Cool. I thought only `APP_DEBUG` controlled whether Debugbar showed up or not, but this env var defs does work as well!

app/helpers.php
```diff
@@ -178,7 +178,7 @@ function is_same_domain($url)
  */
 function scriptify($json = [], $store = 'STATE')
 {
-    return new HtmlString('<script type="text/javascript">window.'.$store.' = '.json_encode($json).'</script>');
+    return new HtmlString('<script type="text/javascript">window.'.$store.' = window.'.$store.' || '.json_encode($json).'</script>');
```
weerd: Might just be confused, but not 100% sure what's going on here....

weerd: wouldn't this always end up making `window.STATE` equal to `window.STATE`? Since `$store` has a default value? I may just be confused and misreading 🤔

weerd: Oh wait, if `window.STATE` is empty/undefined then it sets it to the `$json` data?

DFurnes: Yep, exactly! It's a little awkward to read in PHP, but this evaluates to `window.STATE = window.STATE || { ...json}`, which lets us "mock" that `window.STATE` variable in tests (whereas in the real world, it'll be undefined and so take whatever the server sets).

DFurnes: I think there are better ways we can probably provide & mock these in the future, but this gets us what we need for now (specifically actually, for mocking `window.AUTH` [here](https://github.com/DoSomething/phoenix-next/pull/1438/files#diff-9593faa52e480476d0f5eef99f197fbbR22)).

weerd: Gotcha, right. Makes more sense now... I think reading it within the PHP was throwing me a bit, which is what I had assumed was messing me up, haha. Thanks @DFurnes

weerd:
Very excited to move forward with all this! Great work @DFurnes 🏆

PR #1442, opened 2019-05-30
updates SoftEdgeWidget to include actionId

chloealee:
### What does this PR do?

This PR adds the `actionId` to the SoftEdgeWidget for the user to add this when creating a new SoftEdge action to the page and to send this information in the URL to SoftEdge.

Refs first task in [Pivotal ID #165974086](https://www.pivotaltracker.com/n/projects/2019313/stories/165974086)

### Checklist

### DIFF
--- a/contentful/content-types/softEdgeWidgetAction.js
+++ b/contentful/content-types/softEdgeWidgetAction.js
@@ -1,16 +1,16 @@
 module.exports = function(migration) {
   const softEdgeWidgetAction = migration
-    .createContentType('softEdgeWidgetAction')
-    .name('SoftEdgeWidgetAction')
+    .createContentType("softEdgeWidgetAction")
+    .name("SoftEdgeWidgetAction")
     .description(
-      'Actions made via our third party partner, SoftEdge. This can be an email, Tweet, Facebook post, call, etc. to specific targets. ',
+      "Actions made via our third party partner, SoftEdge. This can be an email, Tweet, Facebook post, call, etc. to specific targets. "
     )
-    .displayField('internalTitle');
+    .displayField("internalTitle");
 
   softEdgeWidgetAction
-    .createField('internalTitle')
-    .name('Internal Title')
-    .type('Symbol')
+    .createField("internalTitle")
+    .name("Internal Title")
+    .type("Symbol")
     .localized(false)
     .required(true)
     .validations([])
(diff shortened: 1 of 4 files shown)


contentful/content-types/softEdgeWidgetAction.js
```diff
@@ -1,64 +1,82 @@
 module.exports = function(migration) {
   const softEdgeWidgetAction = migration
-    .createContentType('softEdgeWidgetAction')
-    .name('SoftEdgeWidgetAction')
+    .createContentType("softEdgeWidgetAction")
```
mendelB: Do ya know why these single quotes got replaced with doubles? 🤔I feel like our other scripts use singles tho obviously not a massive deal lol

contentful/content-types/softEdgeWidgetAction.js
```diff
@@ -1,64 +1,82 @@
...
-  softEdgeWidgetAction.changeEditorInterface('internalTitle', 'singleLine', {
+  softEdgeWidgetAction
+    .createField("actionId")
+    .name("Action Id")
+    .type("Integer")
+    .localized(false)
+    .required(true)
+    .validations([
+      {
+        unique: true
```
mendelB: I know we've traditionally added the `unique` validation to our Action ID fields, but I *feel* like I recall @DFurnes or @weerd making the case for these not to be unique 🤔 I think the argument was something along the lines of potentially having the ability to make two separate UI representations of the same Action on say a Campaign and a separate article page or the like. But might be worth being consistent with this until we remove the validation across actions if necessary?

contentful/content-types/softEdgeWidgetAction.js
```diff
@@ -1,64 +1,82 @@
...
 
-  softEdgeWidgetAction.changeEditorInterface('softEdgeId', 'numberEditor', {
+  softEdgeWidgetAction.changeEditorInterface("softEdgeId", "numberEditor", {
     helpText:
-      'The ID of the SoftEdge action, given upon creation in SoftEdge. When you click "Publish Now", a URL will appear in the "See the Action" box. The SoftEdge ID is the last digit in the URL. e.g. SoftEdge ID = 3 in http://www.congressweb.com/dosomething/3',
+      'The ID of the SoftEdge action, given upon creation in SoftEdge. When you click "Publish Now", a URL will appear in the "See the Action" box. The SoftEdge ID is the last digit in the URL. e.g. SoftEdge ID = 3 in http://www.congressweb.com/dosomething/3'
+  });
+
+  softEdgeWidgetAction.changeEditorInterface("actionId", "numberEditor", {
+    helpText: "The Action ID associated with this action in Rogue."
```
mendelB: Would it be worth adding the link to actions in the help text as we did here? https://github.com/DoSomething/phoenix-next/blob/273a35eb8011e173333ab6744ee3db2151a93e63/contentful/content-types/postGallery.js#L81-L84

contentful/content-types/softEdgeWidgetAction.js
chloealee: I don't know why that happened! I can update those back to change to single quotes though!

weerd: When generating the migration, it generates them with double quotes. But then when saving the file Prettier goes in the turns them all into single quotes. @chloealee is your Prettier extension working correctly? Does it modify your code when you save the file?

contentful/content-types/softEdgeWidgetAction.js
weerd: Seems useful! Let's add that 💃

contentful/content-types/softEdgeWidgetAction.js
weerd: Yeah... hmmm, maybe lets be consistent for now so if we decide to update we will know to do it to all the actions.

weerd: But defs good note and something to swing back to...

contentful/content-types/softEdgeWidgetAction.js
chloealee: @weerd hm it must not be working! I just manually changed all these back though - lmk if there is anything that I missed!

contentful/content-types/softEdgeWidgetAction.js
chloealee: love that idea, thanks @mendelB ! Added!

contentful/content-types/softEdgeWidgetAction.js
weerd: Gah! We definitely need to get it working on your editor. It'll save you from lots of little headaches! I can help you next week with it. Sometimes it's an issue with the Prettier package in Sublime knowing the path to NodeJS. I've had to fix it in the past. If you bring up the Sublime console (Control + `), you might see an error every time you save a file...

DFurnes: Hmm, it seems like the git hooks may not be working anymore... let me take a look. (But also absolutely grab the [Prettier extension](https://prettier.io/) for your editor if you haven't yet!)

chloealee: will do thank you both!

PR #1444, opened 2019-05-30
Updates to allow using additionalContent JSON to override affiliate credit text.

weerd:
### What does this PR do?

This PR adds an update to support using `additionalContent` JSON to enabled overriding the default affiliate credit text to something custom for the campaign. If an override is provided it will override the text in the lede banner and the info bar above the footer with the specified text.

**Default text:**
default copy

Adding the following into the `additionalContent` JSON field for a campaign updates the text.

```json
{
  "affiliateCreditText": "Sponsored by",
}
```

**Custom text:**
custom copy

Refs [Pivotal ID #166249781](https://www.pivotaltracker.com/story/show/166249781)


resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
```diff
@@ -91,6 +94,7 @@ CampaignRoute.propTypes = {
 };
 
 CampaignRoute.defaultProps = {
+  affiliateCreditText: undefined,
```
weerd: I wasn't super keen on needing to set these as `undefined` but it was necessary to allow the ending components to use their set prop defaults.

PR #1446, opened 2019-05-31
Analytics reconfiguring

weerd:
### What does this PR do?

This PR updates our analytics system to conform to the new requirements the data team has defined moving forward.

One of the bigger changes happening to help structure calls to `trackAnalyticsEvent()` a little better is that it now only accepts 3 params:
 - `metaData` (required) // all data around describing the event (noun, verb, adjective, category, etc)
- `contextData` (defaults to empty object) // all additional contextual data (campaignId, actionId, etc)
- `service` (optional) // specifying to only send to a specific service (rarely used but still seems useful; for example if we restore waypoints in the future maybe only Snowplow cares about them)

Most of the PR is actually related to updating ALL calls to `trackAnalyticsEvent()` to pass through the new argument structure.

Refs [Pivotal ID #]()

### Checklist

### DIFF
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -92,13 +92,18 @@ export function storeCampaignPost(campaignId, data) {
 
   // Track post submission event.
   trackAnalyticsEvent({
-    verb: 'submitted',
-    noun: formatEventNoun(type),
-    data: {
+    metaData: {
+      category: 'campaign_action',
+      target: 'form',
+      label: type,
+      noun: formatEventNoun(type),
+      verb: 'submitted',
+    },
+    contextData: {
       action,
       actionId,
-      campaignId,
       campaignContentfulId,
+      campaignId,
     },
   });
 
@@ -144,9 +149,14 @@ export function storePost(data) {
 
   // Track post submission event.
   trackAnalyticsEvent({
-    verb: 'submitted',
-    noun: formatEventNoun(type),
-    data: {
+    metaData: {
+      category: 'campaign_action',
+      target: 'form',
+      label: type,
+      noun: formatEventNoun(type),
+      verb: 'submitted',
+    },
+    contextData: {
       actionId,
     },
   });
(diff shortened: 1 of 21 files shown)


resources/assets/components/actions/ShareAction/ShareAction.js
```diff
@@ -76,54 +76,65 @@ class ShareAction extends React.Component {
...
+  /**
+   * Component helper method for tracking analytics events.
+   *
+   * @param  {String} service
+   * @param  {String} target
+   * @param  {String} verb
+   * @param  {Object} data
+   * @return {Void}
+   */
+  trackEvent = (service, target, verb, data = {}) => {
```
weerd: For this and the `SocialShareTray` component, since there were multiple calls to `trackAnalyticsEvent()`, it felt better to provide a component-wise helper method to handle most of the props with similar values among the events.

resources/assets/actions/post.js
```diff
@@ -92,13 +92,18 @@ export function storeCampaignPost(campaignId, data) {
...
-    noun: formatEventNoun(type),
-    data: {
+    contextData: {
       action,
       actionId,
-      campaignId,
       campaignContentfulId,
+      campaignId,
+    },
+    metaData: {
```
DFurnes: This is sorta nitpicky, but I think this should read `metadata` since it's one word.

resources/assets/components/SignupButton/SignupButton.js
```diff
@@ -50,7 +50,7 @@ const SignupButton = props => {
   const buttonCopy = text || sourceOverride || campaignActionText;
 
   return (
-    <Button className={className} onClick={() => onSignup(buttonCopy)}>
+    <Button className={className} onClick={() => onSignup()}>
```
DFurnes: Since we're not overloading an argument here, this can just be `onClick={onSignup}`. I think our convention elsewhere is to name handlers `handleX` as well, so `onClick={handleSignup}`.

resources/assets/actions/post.js
weerd: I debated this... since `contextData` is 2 words it felt weird to keep `metadata` lowercase... 🤷‍♂

resources/assets/components/SignupButton/SignupButton.js
weerd: Will update in upcoming cleanup PR!

PR #1447, opened 2019-06-03
Renaming trackAnalyticsEvent properties.

weerd:
### What does this PR do?

This PR address recommendation in [#1446](https://github.com/DoSomething/phoenix-next/pull/1446#discussion_r290037530) regarding the naming of `metaData` and `contextData` properties. It changes these to be `metadata` and `context` to simplify things and keep `metadata` as one word! 

It also does a little bit of [cleanup with the `SignupButton` component](https://github.com/DoSomething/phoenix-next/pull/1447/commits/f7c7aba217a8dd28e19eb3164c4a1c0d7cdb208f) and [renames the `eventName` data layer property to just `event`](https://github.com/DoSomething/phoenix-next/pull/1447/commits/702820db3325cf2ae3137adff50a616e14d35105) since this is what GTM prefers.

🛁 

Refs [Pivotal ID #166243697](https://www.pivotaltracker.com/story/show/166243697)

### DIFF
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -180,13 +180,11 @@ export function trackAnalyticsEvent({ metadata, context = {}, service }) {
   }
 
   const { adjective, category, target, noun, verb } = metadata;
-  let { label } = metadata;
+  const label = metadata.label || noun;
 
   const name = formatEventName(verb, noun, adjective);
 
   const action = snakeCase(`${target}_${verb}`);
-
-  label = label || noun;
 
   sendToServices(name, category, action, label, context, service);
 }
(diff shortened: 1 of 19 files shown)


resources/assets/helpers/analytics.js
```diff
@@ -164,28 +164,29 @@ export function googleAnalyticsInit(history) {
...
+export function trackAnalyticsEvent({ metadata, context = {}, service }) {
+  if (!metadata) {
+    console.error('The metadata object is missing!');
     return;
   }
 
-  const { adjective, category, target, noun, verb } = metaData;
-  let { label } = metaData;
+  const { adjective, category, target, noun, verb } = metadata;
+  let { label } = metadata;
```
DFurnes: Would it make more sense to have this as a single `const`?

```js
const label = metadata.label || noun;
```

weerd: oh ya, I like that 👍

PR #1452, opened 2019-06-05
Format socialPlatform field from ShareActionContainer

mendelB:
### What does this PR do?

This PR format the `ShareAction.socialPlatform` field from the `ShareActionContainer` instead of from the `ShareAction` Entity.  

The `ShareAction` component expects a String value:
 https://github.com/DoSomething/phoenix-next/blob/f3864e63dd720d9011544daaf6454bad4857821f/resources/assets/components/actions/ShareAction/ShareAction.js#L226

On Contentful, the field is returned as an Array value:
https://github.com/DoSomething/phoenix-next/blob/f3864e63dd720d9011544daaf6454bad4857821f/contentful/content-types/shareAction.js#L45
(I believe the single Short Text radio button option was not available when we added this field, and this was the only way to allow multiple predefined selections).

We format this on the Entity:
https://github.com/DoSomething/phoenix-next/blob/98856187db42eb366b639dcb835a522f71775705/app/Entities/ShareAction.php#L27

But this formatting wouldn't apply to entries fetched directly from GraphQL via the [`ContentfulEntryLoader`](https://github.com/DoSomething/phoenix-next/blob/f3864e63dd720d9011544daaf6454bad4857821f/resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js#L69).

ideally, we'll want to run a script to transform this field to use the plain Short Text, with a radio button interface (I left a TODO), but for now, I figured moving the formatting over to the Container would be a nice way to apply this formatting across the board without extra finageling in the component logic itself.

https://dosomething.slack.com/archives/C09ANFQLA/p1559687038006100

### DIFF
--- a/resources/assets/components/actions/ShareAction/ShareActionContainer.js
+++ b/resources/assets/components/actions/ShareAction/ShareActionContainer.js
@@ -7,11 +7,14 @@ import { storeCampaignPost } from '../../../actions/post';
 /**
  * Provide state from the Redux store as props for this component.
  */
-const mapStateToProps = state => ({
+const mapStateToProps = (state, ownProps) => ({
   userId: getUserId(state),
   campaignId: state.campaign.campaignId,
   isAuthenticated: isAuthenticated(state),
   campaignContentfulId: state.campaign.id,
+  // Value comes through as Array, but component expects a String value.
+  // @TODO: Update this Contentful field to be a String value.
+  socialPlatform: ownProps.socialPlatform[0] || 'facebook',
 });
 
 /**
--- a/resources/assets/components/actions/ShareAction/ShareActionContainer.js
+++ b/resources/assets/components/actions/ShareAction/ShareActionContainer.js
@@ -1,3 +1,4 @@
+import { isArray } from 'lodash';
 import { connect } from 'react-redux';
 
 import ShareAction from './ShareAction';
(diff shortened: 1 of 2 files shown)


resources/assets/components/actions/ShareAction/ShareActionContainer.js
```diff
@@ -7,11 +7,14 @@ import { storeCampaignPost } from '../../../actions/post';
...
  */
-const mapStateToProps = state => ({
+const mapStateToProps = (state, ownProps) => ({
   userId: getUserId(state),
   campaignId: state.campaign.campaignId,
   isAuthenticated: isAuthenticated(state),
   campaignContentfulId: state.campaign.id,
+  // Value comes through as Array, but component expects a String value.
+  // @TODO: Update this Contentful field to be a String value.
+  socialPlatform: ownProps.socialPlatform[0] || 'facebook',
```
mendelB: I double checked that when this field is _not_ set (it's a required field, but for preview env etc.), this will still come through as an Array, so it won't break when we try to grab the first value.

DFurnes: Let's make the `[0]` conditional on it being an array so there's no blocker on fixing this field!

```js
import { isArray } from 'lodash';

// ...

const socialPlatform = isArray(ownProps.socialPlatform) ? ownProps.socialPlatform[0] : ownProps.socialPlatform;

// ...

socialPlatform: socialPlatform || 'facebook',
```

mendelB: Smarrrrrt! Updated in https://github.com/DoSomething/phoenix-next/pull/1452/commits/5ca8f551d1e085e85ce7ad7131e006480c8c7a13

weerd:
@mendelB I added a Pivotal Ticket for the script update: https://www.pivotaltracker.com/story/show/166501288

mendelB:
> @mendelB I added a Pivotal Ticket for the script update: https://www.pivotaltracker.com/story/show/166501288

🙌 Thanks, @weerd!

PR #1454, opened 2019-06-06
Send all the UTMs to Northstar and Rogue!

DFurnes:
### What does this PR do?

This pull request sends along _all the UTMs(!!!)_ in `source_details` to Rogue for signups & posts, and to Northstar for registrations (pending a little work over there). Since we had existing JSON data (or inconsistent strings & JSON) for `source_details` in Rogue, I opted to start sending as JSON now.

I removed `campaignContentfulId` from details [per Sohaib's confirmation that it's not used](https://dosomething.slack.com/archives/C6J24ADQW/p1559839322010200?thread_ts=1559835260.008300&cid=C6J24ADQW). We can use the `referrer_id` field now included on `source_details` for the same purpose in the future.

I renamed the `formatFormFields` to `formatPostPayload` because it has post-specific logic in it (such as using a FormData when the type field is "photo").

Finally, I added `utm_source` and `utm_campaign` fields to Northstar's `&options=` payload, and renamed `trafficSource` to `utm_medium` so it's a little bit clearer how that's used & easier to search for. I'll remove the old fields once I've made and deployed the corresponding change in Northstar.

Refs [Pivotal ID #166250944](https://www.pivotaltracker.com/story/show/166250944).

### Checklist

### DIFF
--- a/resources/assets/components/SignupButton/SignupButton.js
+++ b/resources/assets/components/SignupButton/SignupButton.js
@@ -22,8 +22,7 @@ const SignupButton = props => {
 
   // Decorate click handler for A/B tests & analytics.
   const handleSignup = () => {
-    // @TODO: Do we need this if we store refferer_id on source_details?
-    const details = { campaignContentfulId: referrerId };
+    const details = {};
 
     // Set affiliate opt in field if user has opted in.
     if (affiliateMessagingOptIn) {
--- a/resources/assets/components/SignupButton/SignupButton.js
+++ b/resources/assets/components/SignupButton/SignupButton.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { get } from 'lodash';
 
+import { query, withoutNulls } from '../../helpers';
 import Button from '../utilities/Button/Button';
 
 const SignupButton = props => {
@@ -30,7 +31,17 @@ const SignupButton = props => {
     }
 
     storeCampaignSignup(campaignId, {
-      body: { details: JSON.stringify(details) },
+      body: {
+        details: JSON.stringify(details),
+        source_details: JSON.stringify(
+          withoutNulls({
+            referrer_id: referrerId,
+            utm_source: query('utm_source'),
+            utm_medium: query('utm_medium'),
+            utm_campaign: query('utm_campaign'),
+          }),
+        ),
+      },
       analytics: {
         context: {
           campaignContentfulId: referrerId,
--- a/resources/assets/components/SignupButton/SignupButton.js
+++ b/resources/assets/components/SignupButton/SignupButton.js
@@ -11,9 +11,9 @@ const SignupButton = props => {
     campaignActionText,
     campaignId,
     campaignTitle,
+    contentfulId,
     className,
     disableSignup,
-    referrerId,
     sourceActionText,
     storeCampaignSignup,
     text,
@@ -34,7 +34,7 @@ const SignupButton = props => {
         details: JSON.stringify(details),
         source_details: JSON.stringify(
           withoutNulls({
-            referrer_id: referrerId,
+            contentful_id: contentfulId,
             utm_source: query('utm_source'),
             utm_medium: query('utm_medium'),
             utm_campaign: query('utm_campaign'),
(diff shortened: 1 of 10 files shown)


resources/assets/components/SignupButton/SignupButton.js
```diff
@@ -21,18 +22,28 @@ const SignupButton = props => {
...
+            utm_source: query('utm_source'),
+            utm_medium: query('utm_medium'),
+            utm_campaign: query('utm_campaign'),
+          }),
+        ),
+      },
       analytics: {
         context: {
-          campaignContentfulId,
+          campaignContentfulId: referrerId,
```
DFurnes: Would `contentfulId` make more sense (since it may not always be a campaign page that has a signup button)?

resources/assets/components/SignupButton/SignupButton.js
```diff
@@ -21,18 +22,28 @@ const SignupButton = props => {
...
       details.affiliateOptIn = true;
     }
 
     storeCampaignSignup(campaignId, {
-      body: { details: JSON.stringify(details) },
+      body: {
+        details: JSON.stringify(details),
+        source_details: JSON.stringify(
+          withoutNulls({
+            referrer_id: referrerId,
```
DFurnes: I'm curious if `referrer_id` makes as much sense here as it does on a user. I guess it _is_ technically the content page that "refers" a user to making a signup, but would `contentful_id` be more straightforward?

resources/assets/helpers/forms.js
```diff
@@ -110,14 +112,25 @@ export function getFormData(formData) {
...
   // JSON serialize details field which gets validated as JSON.
   if (data.details) {
     formattedData.details = JSON.stringify(data.details);
   }
 
+  // Attach 'source_details' based on referring page/block & UTMs:
+  formattedData.source_details = JSON.stringify(
+    withoutNulls({
+      // @TODO: Pass in 'referrer_id' parameter here w/ the containing page ID.
+      // referrer_id: referrerId,
```
DFurnes: Since we don't pass the page ID along to the `PhotoSubmissionAction` et al, I decided to leave this as a TODO. If I have time leftover this sprint, I'm happy to revisit this so that all of these are consistent!

weerd:
> This pull request sends along all the UTMs(!!!)

image

resources/assets/components/SignupButton/SignupButton.js
weerd: I think so.

resources/assets/components/SignupButton/SignupButton.js
weerd: Yeah, I dig that. I think before I did the more verbose name to help make it obvious it was the contentful ID on the campaign... 

Separate point, we have talked about considering anything that contains actions and signup to be "campaigns" in our system... similar to how in Rogue we create a campaign ID even for a Story Page, that doesn't technically have a signup button...

resources/assets/helpers/forms.js
weerd: I might be already working on something like this related to getting more data passed along events for analytics...

resources/assets/components/SignupButton/SignupButton.js
DFurnes: > Separate point, we have talked about considering anything that contains actions and signup to be "campaigns" in our system... similar to how in Rogue we create a campaign ID even for a Story Page, that doesn't technically have a signup button...

Yup! Campaigns are just a way to group user activity. They're distinct from the interface used to submit those posts or signups (campaign or story page, SMS flow, standalone block, etc).

DFurnes: Renamed this to `contentfulId` in 1d2bbc7.

resources/assets/components/SignupButton/SignupButton.js
DFurnes: Cool! Renamed this to `contentful_id` in 1d2bbc7.

resources/assets/helpers/forms.js
DFurnes: Synergy! ⚡️

weerd: synergy

PR #1455, opened 2019-06-06
Snowplow integration

weerd:
### What does this PR do?

This PR wraps up the configuration to send analytics events to Snowplow. We tested this and it is correctly sending events to the QA Snowplow setup.

It also separates out analytics related scripts into their own blade partials. We're eventually going to remove the `google_analytics_script.blade.php` so it made more sense to separate them all out 😄

One thing to note, based on the docs, the `schema` that is provided seems to need to come from a URL. The [Fivetran/Snowplow docs (item #3 in their list if you scroll down a bit)](https://fivetran.com/docs/events/snowplow#uaparsercontext) suggest referencing a JSON schema definition from within a repo. We're temporarily using a public gist for now, and will switch it out to a GitHub link to the schema file _in_ this repo after it merges, but we needed a reference for now to also test this on QA.

We're not 100% sure when the link to the JSON schema is requested... whether it happens on the client side when sending events to Snowplow, or if it happens on the Fivetran/Snowplow side once the event is received on their end. This is something we need to look into, because we'd prefer not having the client make those requests with ever event trigger. Ideally, we'd like to just import the JSON schema within the `helpers/analytics.js` file, and pass it into the `schema` property, but the docs do not seem to indicate if that is possible.

Refs [Pivotal ID #166194848](https://www.pivotaltracker.com/story/show/166194848)


sheyd:
image

resources/assets/helpers/analytics.js
```diff
@@ -118,9 +118,10 @@ export function analyzeWithSnowplow(name, category, action, label, data) {
     return;
   }
 
-  window.snowplow('trackStructEvent', category, action, label, null, null, [
+  window.snowplow('trackStructEvent', category, action, label, name, null, [
     {
-      schema: '', // @TODO: add schema
+      schema:
+        'https://gist.githubusercontent.com/weerd/f4395e328ee16a398739f626a4b91b4c/raw/c1f7825eb7d2e4f7b7620a6a4fd1e599cb7476ff/snowplow_schema_string.json',
```
DFurnes: It looks like these are [just for documentation](https://github.com/snowplow/iglu/wiki/Self-describing-JSONs) from what I can tell! If so, it could be nice to plop `snowplow.json` into the `public/` directory and just reference www.dosomething.org/snowplow.json (or something like that).

resources/assets/schemas/snowplow.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "object",
+  "title": "snowplow_event",
+  "description": "Schema for Snowplow event data payload.",
+  "properties": {
+    "payload": {
+      "type": "string"
```
DFurnes: How come we're stringifying everything into a `payload` string? Would having a defined schema be handy?

resources/views/partials/snowplow_script.blade.php
```diff
@@ -0,0 +1,14 @@
+@if (config('services.analytics.snowplow_url'))
+    <script type='text/javascript'>
+        (function(p,l,o,w,i,n,g){if(!p[i]){p.GlobalSnowplowNamespace=p.GlobalSnowplowNamespace||[];
+          p.GlobalSnowplowNamespace.push(i);p[i]=function(){(p[i].q=p[i].q||[]).push(arguments)
+          };p[i].q=p[i].q||[];n=l.createElement(o);g=l.getElementsByTagName(o)[0];n.async=1;
+          n.src=w;g.parentNode.insertBefore(n,g)}}(window,document,'script','//d1fc8wv8zag5ca.cloudfront.net/2.5.3/sp.js','snowplow'));
+          window.snowplow('newTracker', 'cf', '{{config('services.analytics.snowplow_url')}}', {
+            appId: 'dosomething-web' ,
```
DFurnes: There's a sneaky little space before the comma here! 👾

weerd:
@DFurnes, @sheyd is going to contact Fivetran and ask them if it's possible to pass a JSON object or stringified JSON to the `schema` property instead of passing a URL to a raw JSON file. If not possible, we're also asking if the call to request the raw JSON file happens on the client in the Snowplow analytics script or if it happens on the Fivetran/Snowplow server side so it's not a ton of requests happening for each event on our user's browser.

resources/views/partials/snowplow_script.blade.php
weerd: gah!

DFurnes:
Are any requests to that Gist URL showing up in dev tools when running this branch?

resources/assets/helpers/analytics.js
weerd: Oh interesting. Didn't think of having it as a link in our domain. Good call, we could try that!

resources/assets/schemas/snowplow.json
sheyd: @DFurnes Unfortunately, this is a current [Fivetran Snowplow limitation](https://fivetran.com/docs/events/snowplow#uaparsercontext). In step 4, the documentation references the collector only accepts values and homogenous arrays. @weerd and I attempted to send an `object` type which in [JSON Schema](https://json-schema.org/understanding-json-schema/reference/type.html) is the equivalent of a dict/json object, but it broke the connector on the Fivetran side since it couldn't handle creating the correct column type in the data warehouse table. 😨 

A string was the compromise we came up with so the schema isn't needlessly complicated with a `properties` per event type we would send, which would make lots of empty columns in the data warehouse, and with a single string value, we can extract the object as a JSON type on the data warehouse side for analysis and parsing.

weerd:
> Are any requests to that Gist URL showing up in dev tools when running this branch?

Checked the network tab and not that I could tell 🤔 but would like confirmation from Fivetran.

sheyd:
> Checked the network tab and not that I could tell 🤔 but would like confirmation from Fivetran.

Questions have been sent to Fivetran. Usually takes about a day for them to respond.

resources/assets/helpers/analytics.js
```diff
@@ -118,9 +118,9 @@ export function analyzeWithSnowplow(name, category, action, label, data) {
     return;
   }
 
-  window.snowplow('trackStructEvent', category, action, label, null, null, [
+  window.snowplow('trackStructEvent', category, action, label, name, null, [
     {
-      schema: '', // @TODO: add schema
+      schema: `${window.ENV.PHOENIX_URL}/snowplow_schema.json`,
```
weerd: One caveat is that with local development, events would fails since Fivetran would try to hit my local machine to grab the schema... 

I could maybe switch this to:

```js
window.ENV.APP_ENV === 'local' ? 'https://www.dosomething.org/snowplow_schema.json' : `${window.ENV.PHOENIX_URL}/snowplow_schema.json`
```

thoughts? it'd still be broken on local until it deploys to prod.

DFurnes: Should we be sending events to Snowplow from our locals? If not, probably a moot point!

I like your `${window.ENV.PHOENIX_URL}/snowplow_schema.json` solution for loading the right schema URL on dev/QA! It seems like our schema won't likely ever change, but it'd be a nice way to make it easy to test new fields if it did!

weerd: Cool cool! I guess if we ever need to test again on local, we can just use a quick gist reference to the schema for local testing as needed.

resources/assets/schemas/snowplow.json
weerd: @sheyd maybe we try again? It might have been something else with our setup that was messing up and maybe it will accept an object just fine, just not extracted into Redshift which I think you mentioned is not a concern for us?

weerd: Looking into it further and it seems like we probably _could_ pass an object, but it would mean defining each property it could have and what type it would be, which would be a bit too much overhead. I think we likely found the right middle-ground and data can just parse the string as JSON to turn it back into something more easy to work with.

PR #1457, opened 2019-06-07
Adding UTM data to analytics events.

weerd:
### What does this PR do?

This PR updates our analytics events payloads to include UTM related data in each event if it is available. It also does some cleanup/updates to the new `withoutNullsOrUndefined()` to include also removing items with empty strings, and renames the helper function to `withoutValueless()` so it's a more terse yet understandable name! 💃 Additionally it adds a new `isEmptyString()` helper function!

Refs [Pivotal ID #166243697](https://www.pivotaltracker.com/story/show/166243697)

### DIFF
--- a/resources/assets/helpers/analytics.js
+++ b/resources/assets/helpers/analytics.js
@@ -10,7 +10,7 @@ import {
 } from '@dosomething/analytics';
 
 import { PUCK_URL } from '../constants';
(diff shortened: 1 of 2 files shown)


weerd:
@sohaibhasan this adds any UTMs available to every analytics event!

PR #1460, opened 2019-06-11
Add badges to user profile [pr]

katiecrane:
### What does this PR do?

This PR adds badges to the user profile!

💳 Access to the badge areas are gated and are depended upon the user having the `feature_flag` of `badges => true` in Northstar.

🏸 It computes via GraphQL whether the badge is "earned" or "locked" and displays the appropriate image.

🌜 Surfaces last name in the user profile.

It looks like this (large):
Screen Shot 2019-06-11 at 2 12 12 PM

or this (medium):
phoenix test_us_account_profile_badges

or this (small):
Screen Shot 2019-06-11 at 2 12 28 PM

No one will have the feature flag that gives them badge access until `DS_BADGES_TEST` is `true` in Northstar. Currently this flag is `false` in production and on QA.

[Pivotal ID #165864572](https://www.pivotaltracker.com/story/show/165864572)

### Checklist


resources/assets/components/pages/AccountPage/BadgesTab.js
```diff
@@ -0,0 +1,246 @@
...
+  }
+`;
+
+const BadgesTab = ({ userId }) => (
+  <div className="bg-gray padding-bottom-lg wrapper">
+    <h2 className="caps-lock league-gothic -sm">Your Badges</h2>
+    <div className="margin-top-lg float-left">
+      <div className="margin-top-lg">
+        <ul className="gallery-grid-sextet">
+          <Query query={SIGNUP_COUNT_BADGE} variables={{ userId }}>
```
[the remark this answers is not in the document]
katiecrane: Thanks! @DoSomething/team-rocket wanted to see what you thought about this, I had actually tried to split out each of these `<Query>`s into a `<BadgeQuery>` component. That ended up with all the queries and `<Query>`s themselves in `<BadgeQuery>`, but still a decent amount of code in this `<BadgesTab>` in order to pass through: the query name, the user id, the earned criteria for the badge, the badge name, and the badge text. 

I didn't find it that much simpler so I went the route we see here, but does anyone have a strong preference for using another component like `<BadgeQuery>?

resources/assets/components/pages/AccountPage/Badge.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Figure } from '../../Figure';
+import badgeImages from './BadgeImages';
+
+class Badge extends React.Component {
```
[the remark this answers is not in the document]
katiecrane: Thank you! @DoSomething/team-rocket what do you all thin about this? I haven't seen a function component like this before so not sure if it's something we don't generally use vs. just something that doesn't come up that much.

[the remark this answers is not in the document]
mendelB: I like this idea! We def make use of functional components e.g. for the [Content Block](https://github.com/DoSomething/phoenix-next/blob/b96abb3a0df3a61ee8caa05f54a9f5b8b8b34dac/resources/assets/components/blocks/ContentBlock/ContentBlock.js). I believe our rule of thumb is that any component which doesn't need to maintain its own state or otherwise doesn't *need* a class, we prefer the functional component for simplicity (& I *think* it's slightly more performant but I could be wrong).

resources/assets/components/pages/AccountPage/BadgesTab.js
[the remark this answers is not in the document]
mendelB: I personally do like that it would DRY up the general setup for the queries (the loading/error state etc.) I kinda feel like that's a nice win even if we still need to be somewhat explicit about the query and badge info? (tho tbh I'm never quite certain about perfectly balancing DRYness with superfluous abstraction so def add salt to my opinion).

resources/assets/components/pages/AccountPage/Profile.js
```diff
@@ -13,7 +13,10 @@ const Profile = props => (
     <div className="container__block -half">
       <h3>Profile Info</h3>
 
-      <FormItem title="Name" value={props.user.firstName} />
+      <FormItem
+        title="Name"
+        value={`${props.user.firstName} ${props.user.lastName}`}
```
mendelB: Do you think it's worth separating the First and Last Name into individual form elements since that's how they appear on the edit profile page?

resources/assets/components/pages/AccountPage/BadgesTab.js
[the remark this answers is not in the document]
katiecrane: Alright I'm counting that as 2 vote for another component here so I'm going to update!

resources/assets/components/pages/AccountPage/Badge.js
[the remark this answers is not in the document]
DFurnes: I could've sworn we actually used to have a linting rule for this (that if you weren't using state or lifecycle hooks, would tell you to re-write class components as functions)... +1 from me!

resources/assets/components/pages/AccountPage/BadgesTab.js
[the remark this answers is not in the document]
DFurnes: I think using our [`PaginatedQuery`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/PaginatedQuery.js) component as a template (but minus all the pagination logic) would make great sense here!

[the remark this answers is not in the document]
katiecrane: @DFurnes I was looking at that too, minus all the paginated logic it's pretty much what I already have here yeah? Busting that out into its own component is still the way to go right?

resources/assets/components/pages/AccountPage/Profile.js
katiecrane: Chatted with @ngjo and Leah and we're good to leave it as it is here! Good eye though!

resources/assets/components/pages/AccountPage/Badge.js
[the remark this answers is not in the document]
katiecrane: Now that I've come back to this I don't know what I was talking about before, I have definitely seen components like this!

resources/assets/components/pages/AccountPage/BadgesTab.js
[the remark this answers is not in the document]
DFurnes: Yeah, I think so!

resources/assets/components/QueryResults.js
```diff
@@ -0,0 +1,40 @@
...
+import PropTypes from 'prop-types';
+import { Query } from 'react-apollo';
+
+import ErrorBlock from './ErrorBlock/ErrorBlock';
+import { NetworkStatus } from '../constants';
+
+/**
+ * Fetch results via GraphQL using a query component.
+ */
+const QueryResults = ({ query, queryName, variables, children }) => (
```
DFurnes: Is `queryName` something we could parse from the `query` object, perhaps? (Not at all blocking, but would be a nice way to reduce some overhead when using this component!

DFurnes: Would plain `Query` work for this component name? At first glance, `QueryResults` makes me think the query would have already "happened" by the time we include this component.

resources/assets/components/pages/AccountPage/BadgeImages.js
```diff
@@ -0,0 +1,41 @@
...
+import oneStaffFaveBadgeLocked from './Badges/StaffFave1Locked.svg';
+import twoStaffFavesBadge from './Badges/StaffFave2.svg';
+import twoStaffFavesBadgeLocked from './Badges/StaffFave2Locked.svg';
+import threeStaffFavesBadge from './Badges/StaffFave3.svg';
+import threeStaffFavesBadgeLocked from './Badges/StaffFave3Locked.svg';
+import voterBadge from './Badges/VoterBadge.svg';
+import voterBadgeLocked from './Badges/VoterBadgeLocked.svg';
+
+const BadgeImages = {
+  signupBadge,
```
DFurnes: Are all these images being included in the application bundle (since it looks like they're generally below 8kb)? If so, it might be a good idea to lazy-load this as a separate bundle (like we do [for Quizzes](https://github.com/DoSomething/phoenix-next/blob/8149680de8dafbd29242fbd00f0ff8026cfddf2e/resources/assets/components/ContentfulEntry/ContentfulEntry.js#L224-L225)) so it doesn't impact load times for folks on the rest of the site.

resources/assets/components/QueryResults.js
katiecrane: I think `Query` for sure makes more sense than `QueryResults` but I was worried about it being confusing since there is also the apollo `Query` component that we use here and elsewhere. I did struggle with naming this though, do you think `QueryRunner` or `QueryWrapper` could make more sense? Or do you not share my concern with having two `Query`s?

DFurnes: Personally, I'm okay with having our own `Query`!

resources/assets/scss/gallery-grid.scss
```diff
@@ -60,6 +61,26 @@
   }
 }
 
+.gallery-grid-sextet {
+  @include media('(min-width: 425px)') {
+    grid-template-columns: repeat(3, 1fr);
+  }
+
+  @include media($medium) {
+    grid-template-columns: repeat(3, 1fr);
```
weerd: Would `4` columns not work here? Considering the jump from `3` on small to `6` on large and these are fairly smaller square elements?

katiecrane: @weerd I only have wireframes/prototypes that have 3 and 6, none of them have 4. I'll double check with Leah

weerd: Ah ok! Did you have wireframes for `medium` size? Or just `small` and `large`?

katiecrane: Just desktop and mobile I believe! I tagged you in Slack where I asked Leah to confirm 😃

katiecrane: This 3 columns is approved from Leah!

resources/assets/components/QueryResults.js
katiecrane: Ooo good idea, I'll look into this!

katiecrane: Then our own `Query` we shall have!

resources/assets/scss/gallery-grid.scss
weerd: Aight. Mind then just adding a screenshot in the description for medium size as well as the ones you have for small and large :)

katiecrane: @weerd updated!

weerd: awesome thanks!

resources/assets/components/pages/AccountPage/AccountNavigation.js
```diff
@@ -30,4 +40,10 @@ const AccountNavigation = () => (
   </div>
 );
 
+AccountNavigation.propTypes = {
+  user: PropTypes.shape({
+    hasFeatureFlag: PropTypes.bool,
```
weerd: Not crucial for merge, but should `hasFeatureFlag` be more specific? Like `hasFeatureFlagBadges`... we don't have other feature flags but as a precedent moving forward would be nice to start setting a standard here. If profiles had other feature flags, this would end up being confusing as to specifically what test feature it's targeting? Thoughts 🤔

resources/assets/components/pages/AccountPage/AccountRoute.js
```diff
@@ -2,12 +2,19 @@ import React from 'react';
...
 import PropTypes from 'prop-types';
 
+import BadgesTab from './BadgesTab';
 import Profile from './Profile';
 import Subscriptions from './Subscriptions';
 import UserPostsQuery from './UserPostsQuery';
 
 const AccountRoute = props => (
   <Switch>
+  { props.user.hasFeatureFlag ?
```
weerd: Similar question to https://github.com/DoSomething/phoenix-next/pull/1460/files#r297353234

resources/assets/components/pages/AccountPage/Badge.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Figure } from '../../Figure';
+import badgeImages from './BadgeImages';
+
+const Badge = props => {
+  const { name, text, earned } = props;
+  const badgeImageIndex = earned ? name : `${name}Locked`;
```
weerd: ha! Clever approach with the file naming 👌

resources/assets/components/pages/AccountPage/AccountNavigation.js
katiecrane: Yes! Great suggestion, thank you! I didn't realize you could re-name the fields you get from GraphQL! Updated [here](https://github.com/DoSomething/phoenix-next/pull/1460/commits/5cc9074892d89b470dec84cecab8c2dfe3f9d500).

resources/assets/components/pages/AccountPage/AccountRoute.js
katiecrane: Updated [here](https://github.com/DoSomething/phoenix-next/pull/1460/commits/5cc9074892d89b470dec84cecab8c2dfe3f9d500).

resources/assets/components/pages/AccountPage/Badge.js
katiecrane: thanks!! 😊

resources/assets/components/QueryResults.js
katiecrane: Updated [here](https://github.com/DoSomething/phoenix-next/pull/1460/commits/1cf7b380889d8a72c6ff6eb9f9baff8aa1fbcf2e)

resources/assets/components/pages/AccountPage/BadgeImages.js
katiecrane: Would this happen in this file or elsewhere where `BadgeImages` is used?

If in this file, would it be something like this for each image?
```
        const signupBadge = Loader(import('./Badges/Signup.svg'));
        const signupBadgeLocked = Loader(import('./Badges/SignupBadgeLocked.svg'));
```

DFurnes: I think our best option might actually be to work our way "up" and lazy-load either the whole `BadgesTab` component (and anything it relies on), or even the whole `Account` interface (since most users don't end up on any of these tabs during normal usage).

In either case, you'd want to head to wherever that component is used & replace the `import Thing from '…'` with an inline `import('…')` call and then wrap that in a `Loader()` component before rendering it into the React tree.

So, for example, if we wanted to lazy-load the whole `Account` we could update [this](https://github.com/DoSomething/phoenix-next/blob/e1941cc0c7f9772558183940aac20192a1ea506b/resources/assets/components/pages/AccountPage/AccountQuery.js#L23-L33) to:

```js
import Loader from '../../utilities/Loader';

const AccountQuery = ({ userId }) => {
  const Account = Loader(import('./Account'));
  return (
    <Query query={ACCOUNT_QUERY} queryName="user" variables={{ userId }}>
      {result => <Account {...data} userId={userId} />}
    </Query>
  );
};
```

resources/assets/components/QueryResults.js
katiecrane: @DFurnes the `query` objects are pretty gnarly and as far as I can tell there's not straightforward way to grab just the name unfortunately. I still think it's helpful to pass because otherwise we can only print out a generic "Query failed!" error message

resources/assets/components/pages/AccountPage/BadgeImages.js
katiecrane: Ah and now you're using our hot off the presses `<Query>` component there too? I like this, I'm going to go this route!

resources/assets/components/pages/AccountPage/AccountNavigation.js
weerd: Haha, I didn't either, but it was worth questioning 🕵

resources/assets/components/QueryResults.js
DFurnes: Oh, y'know actually I think we already report these (with more information) in this global [`onError` link](https://github.com/DoSomething/phoenix-next/blob/8149680de8dafbd29242fbd00f0ff8026cfddf2e/resources/assets/graphql.js#L34-L47), so this `console.error` here might be unnecessary...

resources/assets/components/pages/AccountPage/BadgeImages.js
katiecrane: Update [here](https://github.com/DoSomething/phoenix-next/pull/1460/commits/6b2a66d3a9609b5885858f3ea6725d0b30647910)

resources/assets/components/QueryResults.js
DFurnes: Hmm, so I'd originally configured that by [following this example](https://www.apollographql.com/docs/react/features/error-handling/) in the docs, which isn't really all that helpful:

```
[GraphQL error]: Message: Cannot query field "dog" on type "User"., Location: [object Object], Path: undefined
```

I just messed around a bit on my local & was able to get it to output much more helpful messages like this:

```
[GraphQL: AccountQuery] Cannot query field "dog" on type "User".
```

That's coming from the following change to the [`onError` link](https://github.com/DoSomething/phoenix-next/blob/8149680de8dafbd29242fbd00f0ff8026cfddf2e/resources/assets/graphql.js#L34-L47):

```js
const errorLink = onError(({ operation, graphQLErrors, networkError }) => {
  if (graphQLErrors) {
    graphQLErrors.map(({ message }) =>
      console.error(`[GraphQL: ${operation.operationName}] ${message}`),
    );
  }
```

katiecrane: OMG AMAZING! This works super well:
```
graphql.js:37 [GraphQL: SubscriptionTopicsBadgeQuery] Cannot query field "dog" on type "User".
```

katiecrane: Updated [here](https://github.com/DoSomething/phoenix-next/pull/1460/commits/f536f30dcb896e705ad03a42d7628b17d101b42f) and [here](https://github.com/DoSomething/phoenix-next/pull/1460/commits/f629e66467cb650c27da18b1543576b54db6e731)

katiecrane:
A note from @weerd to keep in mind for the future:
> it might make sense to house the badges in a directory outside of the `AccountPage` so it’s easier to use them in other spots. It could be something we want to showcase as they receive a badge etc, and that would be the campaign page, so maybe somewhere more generic :slightly_smiling_face: Defs not a thing now since still feature flag, but something to consider in the future. Would be weird to reference the badges to show on campaign page but ref’d from account page directory.

PR #1462, opened 2019-06-13
Analytics payload updates

weerd:
### What does this PR do?

This PR mostly updates the payload data from events to be more consistent across events and simplify/unify the naming of properties in the payload. We're renaming `campaignContentfulID` to `pageId` and the very generic `id` on components (which would correspond to the content type instance on Contentful) to `blockId` to make it more evident what it is.

It also corrects a typo in the `userId` payload prop (was `userID`), and renames the `app-id` Snowplow config setting from `dosomething-web` to `phoenix`.

🌵 

Refs [Pivotal ID #166595074](https://www.pivotaltracker.com/story/show/166595074)

### DIFF
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -65,12 +65,12 @@ export function resetPostSubmissionItem(id) {
 /**
  * Store posts for the specified campaign.
  *
- * @param {String} campaignId
+ * @param  {String} campaignId
  * @param  {Object} data
  * @return {function}
  */
 export function storeCampaignPost(campaignId, data) {
-  const { action, actionId, body, campaignContentfulId, id, type } = data;
+  const { action, actionId, blockId, body, pageId, type } = data;
 
   if (type === 'photo' && !(body instanceof FormData)) {
     throw Error(
@@ -94,9 +94,9 @@ export function storeCampaignPost(campaignId, data) {
   trackAnalyticsEvent({
     context: {
       actionId,
-      blockId: id,
+      blockId,
       campaignId,
-      pageId: campaignContentfulId,
+      pageId,
     },
     metadata: {
       category: 'campaign_action',
@@ -116,8 +116,8 @@ export function storeCampaignPost(campaignId, data) {
           action,
           actionId,
           campaignId,
-          campaignContentfulId,
-          id,
+          id: blockId, // @TODO: rename property to blockId
+          pageId,
           sixpackExperiments,
           type,
         },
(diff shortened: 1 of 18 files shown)


resources/assets/components/actions/ShareAction/ShareAction.js
```diff
@@ -104,7 +109,7 @@ class ShareAction extends React.Component {
 
     let trackingData = { url: link };
 
-    this.trackEvent('facebook', 'button', 'clicked', { url: link });
+    this.trackEvent('facebook', 'button', 'clicked', { ...trackingData });
```
DFurnes: Can we just pass `trackingData` directly (rather than spreading it into a new object)?

weerd: Oh, ya. It's spread in other ones, but only cause items are added, but defs not needed here.

resources/assets/components/actions/TextSubmissionAction/TextSubmissionActionContainer.js
```diff
@@ -11,8 +11,8 @@ import {
  * Provide state from the Redux store as props for this component.
  */
 const mapStateToProps = state => ({
+  pageId: state.campaign.id,
```
mendelB: I think this directly relates to your [Pivotal ticket](https://www.pivotaltracker.com/n/projects/2019313/stories/166687284) and applies to all actions -- what happens when these actions are in a Story Page since the ID would be in `state.page` and not `state.campaign`?

resources/assets/components/actions/ShareAction/ShareAction.js
```diff
@@ -50,12 +50,12 @@ class ShareAction extends React.Component {
...
 
-    const { actionId, campaignContentfulId, campaignId, link } = this.props;
+    const { actionId, pageId, campaignId, link } = this.props;
 
     const formFields = withoutNulls({
-      action,
+      action, // @TODO: deprecate
       type: SOCIAL_SHARE_TYPE,
-      id: campaignContentfulId,
+      id: pageId, // @TODO: rename property to pageId? Other actions use the blockId?
```
mendelB: lol I was so confused by this for a second. I def agree that it makes sense to switch this to be the `blockId` as the other actions do. (I think that also avoids a potential collision of two Share Actions on the same page!)

resources/assets/actions/post.js
```diff
@@ -116,8 +116,8 @@ export function storeCampaignPost(campaignId, data) {
           action,
           actionId,
           campaignId,
-          campaignContentfulId,
```
mendelB: I think we need to update the [`middleware/api`](https://github.com/DoSomething/phoenix-next/blob/b96abb3a0df3a61ee8caa05f54a9f5b8b8b34dac/resources/assets/middleware/api.js#L66) to reflect the `campaignContentfulId` change?

resources/assets/components/actions/TextSubmissionAction/TextSubmissionActionContainer.js
weerd: End up being `undefined` and then removed from being sent as part of payload for analytics events. But yeah, this is part of the reason I brought it up in the PT ticket 👍

resources/assets/components/actions/ShareAction/ShareAction.js
weerd: Left it as a TODO for cleanup week, since felt like it could end up being more involved.

resources/assets/actions/post.js
weerd: 🤔

weerd: Good catch @mendelB! Updating in a new PR right now 💃

PR #1468, opened 2019-06-17
Snowplow page view tracking updates.

weerd:
### What does this PR do?

This PR updates how we track page views with Snowplow to ensure that when a user clicks on a campaign subpage, those actions are tracked as distinct page views.

This PR also removes a bit of the old, now deprecated Google Analytics code.

Lastly, it also adds a nice visual console output when there is no `SNOWPLOW_URL` environment variable provided. For local development, this is great for seeing when and what data is getting tracked by Snowplow, since Snowplow will not work locally (specifically with any methods that rely on context data schemas):

image

Screen Shot 2019-06-19 at 10 51 05 AM

Refs [Pivotal ID #166709478](https://www.pivotaltracker.com/story/show/166709478)


resources/assets/helpers/analytics.js
```diff
@@ -178,20 +165,45 @@ export function formatEventNoun(string) {
...
 /**
- * Watch the given parameters for changes in their state
- * and record it to Google Analytics.
+ * Get additional context data.
+ *
+ * @return {Object}
  */
-export function googleAnalyticsInit(history) {
-  init('track');
+export function getAdditionalContext() {
```
weerd: Only getting UTM data for now, but is ideal spot for adding anything extra for both user triggered events and page view events.

resources/assets/helpers/analytics.js
```diff
@@ -45,18 +39,11 @@ const formatEventName = (verb, noun, adjective = null) => {
  * @return {void}
  */
 export function analyzeWithGoogle(name, category, action, label, data) {
-  if (!category || !action) {
-    console.error('The Category or Action is missing!');
+  if (!name || !category || !action || !label) {
+    console.error('Some expected data is missing!');
```
weerd: 🤷‍♂

resources/assets/helpers/analytics.js
```diff
@@ -178,20 +165,45 @@ export function formatEventNoun(string) {
...
+ * @param  {Object} history
+ * @return {void}
+ */
+export function trackAnalyticsPageView(history) {
+  const context = {
+    ...getAdditionalContext(),
+    campaignId: get(window, 'STATE.campaign.campaignId', null),
+    pageId:
+      get(window, 'STATE.campaign.id', null) ||
+      get(window, 'STATE.page.id', null),
```
weerd: There may be a better way to write this statement 🤔

resources/assets/helpers/analytics.js
```diff
@@ -45,18 +39,11 @@ const formatEventName = (verb, noun, adjective = null) => {
...
+    console.error('Some expected data is missing!');
     return;
   }
 
-  // @DEPRECATE
-  // Format the event parameter as expected by the analyze method.
-  const identifier = `${category}:${action}:${label}`;
-
-  // @DEPRECATE
-  analyze(identifier);
```
mendelB: 😢 👋

resources/assets/helpers/analytics.js
mendelB: buzz

resources/assets/helpers/analytics.js
mendelB: Can't think of one personally (though that's not proof of anything)...

resources/views/partials/snowplow_script.blade.php
```diff
@@ -4,12 +4,51 @@
...
           n.src=w;g.parentNode.insertBefore(n,g)}}(window,document,'script','//d1fc8wv8zag5ca.cloudfront.net/2.5.3/sp.js','snowplow'));
+
           window.snowplow('newTracker', 'cf', '{{config('services.analytics.snowplow_url')}}', {
             appId: 'phoenix',
             cookieDomain: null,
             discoverRootDomain: true
         });
+    </script>
+@else
+    {{-- Custom script for logging Snowplow events to the console when no ENV variable provided. --}}
```
mendelB: 🤸‍♂ 🚀 🥇

resources/assets/helpers/analytics.js
```diff
@@ -178,20 +165,45 @@ export function formatEventNoun(string) {
...
+/**
+ * Track page views on initial load and for any changes in History interface.
+ *
+ * @param  {Object} history
+ * @return {void}
+ */
+export function trackAnalyticsPageView(history) {
+  const context = {
+    ...getAdditionalContext(),
+    campaignId: get(window, 'STATE.campaign.campaignId', null),
```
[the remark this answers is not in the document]
weerd: I was debating this, and I thiiiink we may prefer to have it on the individual events and not potentially override any data. Grabbing this from the `window.STATE` is not the most ideal method to begin with, which is why I opted to only do it this way for page view tracking since trying to reach into the Redux store from here would be trickier and more involved. Defs something we could improve on in the future though!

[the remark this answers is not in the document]
mendelB: Ah, true that - fair point!

PR #1469, opened 2019-06-18
Mock Contentful data when running browser tests.

DFurnes:
### What does this PR do?

This pull request adds the ability to replace `window.STATE` values with mocked data in tests, like we do with `window.AUTH`. Using that, we can now freeze a "snapshot" of an [example campaign](https://dev.dosomething.org/us/campaigns/test-example-campaign) so that content changes wouldn't cause tests to fail unexpectedly.

In the future, it'd be sweet to remove the dependency on making that initial Contentful request from [the server-side controller](https://github.com/DoSomething/phoenix-next/blob/40b19e91baba111b5aa2bfd034dc7451f7aec905/app/Http/Controllers/CampaignController.php#L61) (since without a valid Contentful API key, that route 500s).

Refs [Pivotal ID #166290295](https://www.pivotaltracker.com/story/show/166290295).

### Checklist

### DIFF
--- a/cypress/integration/campaign-signup.js
+++ b/cypress/integration/campaign-signup.js
@@ -1,10 +1,11 @@
 /// <reference types="Cypress" />
 
 import { userFactory } from '../fixtures/user';
+import { exampleCampaign } from '../fixtures/contentful';
 import { emptyResponse, existingSignup, newSignup } from '../fixtures/signups';
 
-const API = `/api/v2/campaigns/9001`;
-const campaignId = '9001';
+const campaignId = '9002';
+const API = `/api/v2/campaigns/${campaignId}`;
 
 describe('Campaign Signup', () => {
   // Configure a new "mock" server before each test:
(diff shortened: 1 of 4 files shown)


cypress/integration/campaign-signup.js
```diff
@@ -61,12 +65,13 @@ describe('Campaign Signup', () => {
     );
 
     // Log in & visit the campaign pitch page:
-    cy.login(user).visit('/us/campaigns/test-teens-for-jeans');
+    cy.login(user).visit('/us/campaigns/test-example-campaign');
```
mendelB: Why do we not need to add the `.withState(exampleCampaign)` for this one?

DFurnes: Oh gosh! We sure should. Without that, it's just using the actual Contentful API request (and so,  works the same for now, but could break unexpectedly down the line).

DFurnes: Added in f05f78d. Thanks for catching that @mendelB!

weerd:
image

So good 👌

PR #1474, opened 2019-06-25
Manually setting Snowplow cookie domain to ds root domain.

sheyd:
### What does this PR do?

This PR removes `discoverRootDomain` function which isn't valid in Snowplow Tracker 2.5.3 and manually adds the root domain.

@mendelB Tracked down that the version of the Snowplow Tracker we use from Fivetran doesn't support the `discoverRootDomain` functionality, and this was validated by Megha in Quasar that we're not seeing cross-domain sessions and `domain_userid` between Northstar and Phoenix. 😡 

It is possible to set the cookie domain for cross-domain tracking in Snowplow Tracker version 2.5.3 as noted [here](https://github.com/snowplow/snowplow/wiki/1-General-parameters-for-the-Javascript-tracker-v2.5#223-configuring-the-cookie-domain). 

This PR removes the root domain discovery and manually sets the domain to be wildcard as [described](https://github.com/snowplow/snowplow/wiki/1-General-parameters-for-the-Javascript-tracker-v2.5#223-configuring-the-cookie-domain) in the 2.5.3 documentation.

https://dosomething.slack.com/archives/CJU7WCQG4/p1561414915100800

[Northstar corollary PR](https://github.com/DoSomething/northstar/pull/886).

### DIFF
--- a/resources/views/partials/snowplow_script.blade.php
+++ b/resources/views/partials/snowplow_script.blade.php
@@ -7,8 +7,7 @@
 
           window.snowplow('newTracker', 'cf', '{{config('services.analytics.snowplow_url')}}', {
             appId: 'phoenix',
-            cookieDomain: null,
-            discoverRootDomain: true
+            cookieDomain: '.dosomething.org'
         });
     </script>
 @else


resources/views/partials/snowplow_script.blade.php
```diff
@@ -7,8 +7,7 @@
 
           window.snowplow('newTracker', 'cf', '{{config('services.analytics.snowplow_url')}}', {
             appId: 'phoenix',
-            cookieDomain: null,
-            discoverRootDomain: true
+            cookieDomain: '.dosomething.org'
```
weerd: > WARNING: Changing the cookie domain will reset all existing cookies. As a result, it might be a major one-time disruption to data analytics because all visitors to the website will receive a new domain_userid.

From the Snowplow docs. Just FYI

weerd: If we ever upgrade the tracker, do we care about this <https://github.com/snowplow/snowplow/wiki/1-General-parameters-for-the-Javascript-tracker#2219-automatically-discover-and-set-the-root-domain> and if adding or not adding `.`?

weerd:
Given that we are now setting this manually, have we reviewed all the other related settings and if they need to be set as well? `cookieName`? cookie lifetime?

resources/views/partials/snowplow_script.blade.php
sheyd: Thanks for the double-validation! Data team discussed in the Slack thread, and since we're not using the data for analytics yet, we're okay generating new cookies based on changing the domain name and using the new data post-deployment going forward.

mendelB: Sohaib confirmed that we're OK with this https://dosomething.slack.com/archives/CJU7WCQG4/p1561472592002600?thread_ts=1561414915.100800&cid=CJU7WCQG4

sheyd: Based on my [read of the docs](https://github.com/snowplow/snowplow/wiki/1-General-parameters-for-the-Javascript-tracker-v2.5#223-configuring-the-cookie-domain) for v2.5 of the tracker, we have to add the pre-pended `.` to enable subdomain tracking, so not sure we have a choice.

sheyd:
@weerd Reading thru the rest of the settings, I don't think we want to change any of the other ones unless we have to, and aside from the domain change, we don't have enough time to test each change or as a bunch if it introduces potentially new artifacts into our analytics layer.

PR #1479, opened 2019-06-27
Sixpack base env

jocoso:
### What does this PR do?

The default environment after cloning phoenix-next shows env variable for SIXPACK_BASE_URL as https://experiments-qa.dosomething.org instead of https://experiments-qa.dosomething.org/api what causes the testing environment to constantly ask for authentication while SIXPACK_ENABLED is true. 

Screen Shot 2019-06-27 at 1 16 35 PM

### Checklist


mendelB:
This is great, @jocoso!

This highlighted a few questions:
1. Is the `SIXPACK_BASE_URL` name a misnomer since it's got the `/api` route?
2. It seems that the `SIXPACK_ENABLED` env variable, even when set to `false` would still prompt this auth pop up -- is it actually controlling anything? We don't seem to be using it anywhere aside from passing it to the front-end https://github.com/DoSomething/phoenix-next/blob/b96abb3a0df3a61ee8caa05f54a9f5b8b8b34dac/app/Providers/AppServiceProvider.php#L36

weerd:
> It seems that the SIXPACK_ENABLED env variable, even when set to false would still prompt this auth pop up -- is it actually controlling anything?

The auth pop up is not related to this at all.

I think we used to use it but after some of the rework last summer it got ignored. I did want to bring it back in and use it with some of my upcoming Sixpack work since I sometimes use a local Sixpack server and would be a nice way to just keep any Sixpack stuff from working if not enabled.

mendelB:
> > It seems that the SIXPACK_ENABLED env variable, even when set to false would still prompt this auth pop up -- is it actually controlling anything?
> 
> The auth pop up is not related to this at all.
> 
> I think we used to use it but after some of the rework last summer it got ignored. I did want to bring it back in and use it with some of my upcoming Sixpack work since I sometimes use a local Sixpack server and would be a nice way to just keep any Sixpack stuff from working if not enabled.

Ah ok that makes sense!

mendelB:
@jocoso Feel free to merge this Pull Request! 🎉

PR #1482, opened 2019-06-28
Fixed displacement of yellow line and line appearing when there was n…

jocoso:
### What does this PR do?

This PR deletes the margin modifying the Forge class. It also fixes an issue in which a yellow line would be present even without a pre-title. 

Screen Shot 2019-06-28 at 12 26 34 PM

Screen Shot 2019-06-28 at 12 27 32 PM

Refs [Pivotal ID #166937791](https://www.pivotaltracker.com/n/projects/2019313/stories/166937791)

### Checklist

### DIFF
--- a/resources/assets/components/SectionHeader/SectionHeader.js
+++ b/resources/assets/components/SectionHeader/SectionHeader.js
@@ -7,7 +7,7 @@ import './section-header.scss';
 
 const SectionHeader = ({ preTitle, title }) => (
   <div className="flex__cell -two-thirds section-header">
-    {preTitle != null ? (
+    {preTitle ? (
       <span className="heading -emphasized section-header__pre-title">
         {preTitle}
       </span>
--- a/resources/assets/components/SectionHeader/SectionHeader.js
+++ b/resources/assets/components/SectionHeader/SectionHeader.js
@@ -7,11 +7,11 @@ import './section-header.scss';
 
 const SectionHeader = ({ preTitle, title }) => (
   <div className="flex__cell -two-thirds section-header">
-    {preTitle != null && (
+    {preTitle != null ? (
       <span className="heading -emphasized section-header__pre-title">
         {preTitle}
       </span>
-    )}
+    ) : null}
 
     {title ? <h1 className="section-header__title">{title}</h1> : null}
   </div>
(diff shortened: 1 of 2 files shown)


resources/assets/components/SectionHeader/SectionHeader.js
```diff
@@ -7,9 +7,12 @@ import './section-header.scss';
 
 const SectionHeader = ({ preTitle, title }) => (
   <div className="flex__cell -two-thirds section-header">
-    <span className="heading -emphasized section-header__pre-title">
-      {preTitle}
-    </span>
+    {preTitle != null && (
```
mendelB: @jocoso I think it'd be optimal to keep the formatting for conditionally outputting markup consistent with using a [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator). (Check out the way we output the `title` [below](https://github.com/DoSomething/phoenix-next/pull/1482/files#diff-125cb0eadb64a3a4ff5e8f31e1413063R16) for reference.)

resources/assets/components/SectionHeader/section-header.scss
```diff
@@ -14,7 +14,6 @@
 
   .-emphasized:after {
     margin-top: -12px;
```
mendelB: @lkpttn Did you want the top negative margin displacement removed as well or was it just the nudge to the right?

resources/assets/components/SectionHeader/SectionHeader.js
jocoso: @mendelB Okay, working on it.

resources/assets/components/SectionHeader/section-header.scss
jocoso: It was a nudge. Should I remove the complete .-emphasized: after?

mendelB: Let's wait to hear from @lkpttn as to whether we want the entire thing removed.

resources/assets/components/SectionHeader/SectionHeader.js
jocoso: @mendelB Fixed

resources/assets/components/SectionHeader/SectionHeader.js
```diff
@@ -7,11 +7,11 @@ import './section-header.scss';
 
 const SectionHeader = ({ preTitle, title }) => (
   <div className="flex__cell -two-thirds section-header">
-    {preTitle != null && (
+    {preTitle != null ? (
```
mendelB: Sorry - one more thing - we also don't need to explicitly check if `preTitle != null` since the ternary operator will evaluate the variable as a Boolean, so just `preTitle ? ...` will evaluate to `false` if the value is `null`. (Just like in the `title` below 😀)

PR #1486, opened 2019-07-03
Signup badge in Affirmation modal

katiecrane:
### What does this PR do?

This PR adds the welcome badge to the affirmation modal after signup. The badge will only appear to users who:
- are opted in to the badge experience
- have exactly one signup

Props have been added to the Badge component to allow for differences of styling and additional information without changing the appearance of the badges on the user profile.

Small Screen
image

Medium Screen
image

Large Screen
image

I think it's mostly in the card! Prototypes/wireframes are linked from there.

Refs [Pivotal ID #165719621](https://www.pivotaltracker.com/story/show/165719621)

### Checklist

### DIFF
--- a/resources/assets/components/Figure/figure.scss
+++ b/resources/assets/components/Figure/figure.scss
@@ -84,3 +84,19 @@
     }
   }
 }
+
+// @TODO: temporary for badges, will need clean-up later - added on 07/09/19
+.figure.badge {
+  @include media($small) {
+    > .figure__media {
+      width: 150px;
+    }
+  }
+
+  @include media($medium) {
+    > .figure__body {
+      margin: 0 auto;
+      max-width: 80%;
+    }
+  }
+}
--- a/resources/assets/components/pages/AccountPage/Badge.js
+++ b/resources/assets/components/pages/AccountPage/Badge.js
@@ -5,13 +5,14 @@ import { Figure } from '../../Figure';
 import badgeImages from './BadgeImages';
 
 const Badge = props => {
-  const { name, text, earned, boldText } = props;
+  const { name, text, earned, boldText, explainerText, size } = props;
   const badgeImageIndex = earned ? name : `${name}Locked`;
 
   return (
     <div>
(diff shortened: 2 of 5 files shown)


resources/assets/components/pages/AccountPage/Badge.js
```diff
@@ -5,13 +5,14 @@ import { Figure } from '../../Figure';
 import badgeImages from './BadgeImages';
 
 const Badge = props => {
-  const { name, text, earned } = props;
+  const { name, text, earned, boldText, explainerText, size } = props;
   const badgeImageIndex = earned ? name : `${name}Locked`;
 
   return (
     <div>
```
weerd: So instead of wrapping this in a `div` which you don't really need (only added since React complains that there's no wrapping element). You can instead use a `<React.Fragment>` and closing `</React.Fragment>` which doesn't end up outputting a useless `div`. Try this and see if the figure better sizes itself. If not, then maybe allow passing a `className` prop and we can see if passing a class when called in affirmation will help with the sizing issue 😄

katiecrane: @weerd oh wait yeah! that must be left over from something, I think it can just be wrapped in `<Figure>` and doesn't need the `<React.Fragment>` if I remove the `<div>`

weerd: Oh true!

katiecrane: @weerd I removed the `<div>` and it looks the same 😩 What class should I try passing?

resources/assets/components/Figure/figure.scss
```diff
@@ -84,3 +84,19 @@
     }
   }
 }
+
+// @TODO: temporary for badges, will need clean-up later - added on 07/09/19
```
weerd: 🥇

katiecrane:
@mendelB OMG nope you're the first, you win the trophy 🏆 !!!

PR #1487, opened 2019-07-03
Enabling Sixpack landing page gallery experiments.

weerd:
### What does this PR do?

This PR enables select campaigns to run a Sixpack A/B experiment on their respective landing pages. I've also included a Dev and QA campaign to help with testing after merging this PR.

You can enable this test on a Campaign's landing page, by adding the following JSON to the `additionalContent` field on the `LandingPage` attached to a campaign:

```json
{
	"sixpackLandingPageGallery": true
}
```

Refs [Pivotal ID #165970311](https://www.pivotaltracker.com/story/show/165970311)

### Checklist

### DIFF
--- a/resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
+++ b/resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
@@ -4,10 +4,15 @@ import PropTypes from 'prop-types';
 import Card from '../../../utilities/Card/Card';
 import { getScholarshipAffiliateLabel } from '../../../../helpers';
 import TextContent from '../../../utilities/TextContent/TextContent';
+import SixpackExperiment from '../../../utilities/SixpackExperiment/SixpackExperiment';
+import PostGalleryBlockQuery from '../../../blocks/PostGalleryBlock/PostGalleryBlockQuery';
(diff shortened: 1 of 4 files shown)


resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
```diff
@@ -16,6 +21,14 @@ const PitchTemplate = ({
   const displayAffiliateScholarshipBlock =
     scholarshipAffiliateLabel && scholarshipAmount && scholarshipDeadline;
 
+  const testableCampaigns = [
+    '6LQzMvDNQcYQYwso8qSkQ8', // Dev
+    '6tgfJeVQIQXw8xfsgyNkHI', // QA
+    '6tgfJeVQIQXw8xfsgyNkHI', // Prod
+    '6ATBgGEQEeJoIcxs1qbwaC', // Prod
+    '5Jp8I6l0e9GVZzuHSgZ9Qp', // Prod
+  ];
```
weerd: I was debating with using `additionalContent` field or this approach. Since it's a code test and would help with reminding us to clean this up after the experiments have run their course I figured this would be fine and less to deal with passing the `additionalContent` field prop around. 🤷‍♂

weerd: Chatted with @DFurnes in Slack and we agreed I shouldn't be lazy and instead use `additionalContent` for this test flag instead!

PR #1489, opened 2019-07-03
Fixed styling issues

jocoso:
### What does this PR do?

This PR removes the top margin found on top of card elements affected by the p+p tag.

Screen Shot 2019-07-03 at 5 21 38 PM

Screen Shot 2019-07-03 at 5 22 08 PM

Refs [Pivotal ID #167056640](https://www.pivotaltracker.com/n/projects/2019313/stories/167056640)

### Relevant Notes

Changed line 7 on package.json `"cypress:ci": "cypress run -c baseUrl=http://127.0.0.1:8000 --record", ` to `"cypress:ci": "cypress run -c baseUrl=http://127.0.0.1:8000", `  to avoid error relating cypress.

Screen Shot 2019-07-15 at 11 01 29 AM

### Checklist


weerd:
hey @jocoso can you link to the Pivot Ticket by adding the URL to the ticket in the `()`:

```
Refs [Pivotal ID #167056640](https://url_here...)
```

resources/assets/components/utilities/Card/card.scss
```diff
@@ -4,6 +4,10 @@
   background-color: $white;
   overflow: hidden;
   width: 100%;
+
+  p + p {
+    margin-top: 8px;
```
weerd: This will affect ALL cards that have multiple paragraphs. I think you may want to specifically target these embeds and where it outputs the source in bold.

jocoso: Fixed it by changing p+p to .footnote.

resources/assets/components/utilities/Card/card.scss
```diff
@@ -4,6 +4,12 @@
   background-color: $white;
   overflow: hidden;
   width: 100%;
+
+  .footnote {
```
mendelB: Since this concerns the Embed component specifically, it might make sense to add this rule within its respective scss file https://github.com/DoSomething/phoenix-next/blob/b96abb3a0df3a61ee8caa05f54a9f5b8b8b34dac/resources/assets/components/utilities/Embed/Embed.js#L16

jocoso: @mendelB  It is done. Although I am unsure if the place I put it was the right one.

mendelB: Hmm, I'm thinking that since we do probably want a little bit of margin between the embed description and the embed provider name, it might make sense to override this margin with a utility class on the element itself. Maybe by adding a `margin-top-md` class to the `p.footnote` in the Embed component?
https://github.com/DoSomething/phoenix-next/blob/17756a54d0ef11e9d85461824230f166ee49cb3b/resources/assets/components/utilities/Embed/Embed.js#L83

jocoso: What about replacing the footnote class with margin-top-md?

mendelB: I think we do use the `footnote` class for font related styling so we need to keep it. But I believe just adding the `margin-top-md` should get us what we need!

jocoso: I think the override of footnote also overrides margin-top-md for some reason.

weerd:
Something to note with these design changes, please scour the code and find additional references where a particular component that is getting updated is used. While a fix may address a particular issue in a specific location, it can have affects in other areas where it's being used. This is what happened with the edits we had to revert prior to last week's deploy. Just helps to be safe! Eventually once we completely switch over to Tailwind, this will ideally be less of a problem 😄

resources/assets/components/utilities/Card/card.scss
mendelB: Ah, right it totally will! you'll need to get rid of the override since we won't be needing it anymore with the utility class.

jocoso: Okay, works like a charm.

package.json
```diff
@@ -4,7 +4,7 @@
     "test": "jest --watch",
     "test:ci": "jest",
     "cypress": "cypress open",
-    "cypress:ci": "cypress run -c baseUrl=http://127.0.0.1:8000 --record",
+    "cypress:ci": "cypress run -c baseUrl=http://127.0.0.1:8000",
```
mendelB: Also, can you please add some context for this change to the PR description? (You can link to your Slack question as well)

jocoso: David told me to do it so it will not generate this error:
Screen Shot 2019-07-15 at 11 01 29 AM

jocoso:
> Something to note with these design changes, please scour the code and find additional references where a particular component that is getting updated is used. While a fix may address a particular issue in a specific location, it can have affects in other areas where it's being used. This is what happened with the edits we had to revert prior to last week's deploy. Just helps to be safe! Eventually once we completely switch over to Tailwind, this will ideally be less of a problem 😄

@weerd I checked some other campaigns and everything seems to be fine, the only changes I see were relevant to the issue.

weerd:
@jocoso awesome! thanks for double-checking 👌

PR #1490, opened 2019-07-03
Button-Title-Case & Removed-White-Background

aktersabina122:
### What does this PR do?

This PR Changed all the button into title case and removed white background from -tertiary class

Screen Shot 2019-07-03 at 5 21 30 PM

Refs [Pivotal ID #66849537]()
https://www.pivotaltracker.com/n/projects/2019313/stories/166849537

### Checklist

* [] ### **Documentation added for new features/changed endpoints.**
* [x ] Added screenshot to PR description of related front-end updates on **small** screens.

### DIFF
--- a/resources/assets/components/utilities/Button/button.scss
+++ b/resources/assets/components/utilities/Button/button.scss
@@ -32,7 +32,3 @@
 .button__attached-container {
   padding: $half-spacing;
 }
-
-.button.-tertiary {
-  background: none;
-}
--- a/resources/assets/components/utilities/Button/button.scss
+++ b/resources/assets/components/utilities/Button/button.scss
@@ -1,6 +1,8 @@
 @import '../../../scss/next-toolbox.scss';
 
 .button {
+  text-transform: capitalize;
+
   &.-inline {
     clear: none;
   }
@@ -32,5 +34,5 @@
 }
 
 .button.-tertiary {
-  background: #fff;
+  background: none;
 }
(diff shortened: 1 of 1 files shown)


resources/assets/components/utilities/Button/button.scss
```diff
@@ -32,5 +34,5 @@
 }
 
 .button.-tertiary {
-  background: #fff;
+  background: none;
```
mendelB: I think we might be able to just get rid of this entire `.button.-tertiary` rule completely. Unless there's some other Forge rule that we're overriding? (When you inspect the element, do you see some other rule adding a background other than the default?)

aktersabina122: Let me check. Thanks

aktersabina122: > I think we might be able to just get rid of this entire `.button.-tertiary` rule completely. Unless there's some other Forge rule that we're overriding? (When you inspect the element, do you see some other rule adding a background other than the default?)

No, I do not see any other rule adding a background other than the default. Do you want me to remove the .button.-tertiary class?

mendelB: Ok perfect! Then yes we can get rid of the whole rule 🗑

aktersabina122: I removed the .button.-tertiary class from the button.scss. Please let me know if everything looks good. Thanks

PR #1492, opened 2019-07-05
Updating to collect all content into PitchTemplate.

weerd:
### What does this PR do?

This PR updates the `LandingPage` and `PitchTemplate` components to collect all the content outputted in the `LandingPage` into the `PitchTemplate`. This will help with an upcoming A/B test on the landing page to test an entirely different template which showcases content and layout differently. The `PitchTemplate` only handled the middle set of content on the page, but with this upcoming A/B test we also want to rework the lede banner and other elements, so it makes more sense for each template file to contain all the content on the page.

### Checklist

### DIFF
--- a/resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
+++ b/resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
@@ -1,10 +1,14 @@
 import React from 'react';
 import { get } from 'lodash';
 import PropTypes from 'prop-types';
+import { PuckWaypoint } from '@dosomething/puck-client';
 
+import Enclosure from '../../../Enclosure';
 import Card from '../../../utilities/Card/Card';
 import { getScholarshipAffiliateLabel } from '../../../../helpers';
 import TextContent from '../../../utilities/TextContent/TextContent';
+import LedeBannerContainer from '../../../LedeBanner/LedeBannerContainer';
+import CallToActionContainer from '../../../CallToAction/CallToActionContainer';
 import SixpackExperiment from '../../../utilities/SixpackExperiment/SixpackExperiment';
 import PostGalleryBlockQuery from '../../../blocks/PostGalleryBlock/PostGalleryBlockQuery';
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
```diff
@@ -75,18 +115,19 @@ PitchTemplate.propTypes = {
   campaignId: PropTypes.string.isRequired,
   campaignTitle: PropTypes.string.isRequired,
   content: PropTypes.string.isRequired,
-  scholarshipAmount: PropTypes.number,
```
mendelB: Aren't the `scholarshipAmount` & `scholarshipDeadline` fields optional on campaigns?

weerd: Yep, and the `LandingPage` that calls this template is already providing the defaults. To me it seemed strange to have to provide defaults in both locations that are exactly the same. Thus, the `LandingPage` sets the defaults (even if the value is `null`) and passes that to the `PitchTemplate`.

weerd:
> I imagine more 🚧 to come as you add the new templates, but as of now does it feel like the LandingPage itself is a bit superfluous? Or I guess that's precisely the point and the idea is to eventually mimic the LedeBanner?

Yeah, I think so... it's leaning more towards that, and while we're aiming to avoid template toggles in Contentful.

resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
mendelB: Ah but wouldn't we still get a warning for a required prop being `null`?

weerd: Oh since it needs to be a string or number depending on prop... 🤔 I guess this is more a facet of the templating approach and thus needs to have those default props repeated or do the whole one of type (string or null) which feels more gross. Ok! Good call @mendelB 👍

PR #1505, opened 2019-07-15
Revamping affiliate promotion component

weerd:
### What does this PR do?

This PR removes the `SponsorPromotion` component by revamping it as the `AffiliatePromotion` component. Renamed it since it seemed to mostly be a promotion of not just sponsors but also partners and all it really is is a very specific component for showcase promotions of affiliates with some copy and a logo 😉 

I went about simplifying a lot of the styles. Most of the styles seemed to be related to scenarios that were specific to the legacy lede banner template or how we displayed sponsors on the legacy campaign. Also enabled the ability to customize the design by relying on more utility classes that can be passed in to customize it based on the context of where it's used.

It also deletes the Lede Banner `LegacyTemplate` since it turns out we don't need it or use it anymore!

This work was necessary to allow more flexibility in this component for use in the new Marquee Landing Page template.

[AffiliatePromotion in the Lede Banner](https://user-images.githubusercontent.com/105849/61247853-de7bf080-a71f-11e9-8ade-a0d99e33a8ba.png)

[AffiliatePromotion in a CampaignUpdate](https://user-images.githubusercontent.com/105849/61247820-cdcb7a80-a71f-11e9-8c1c-77857ef75db0.png)

[AffiliatePromotion in a Link Action](https://user-images.githubusercontent.com/105849/61247907-010e0980-a720-11e9-950d-d4aef30c0b95.png)

### Checklist


package.json
```diff
@@ -4,7 +4,7 @@
     "test": "jest --watch",
     "test:ci": "jest",
     "cypress": "cypress open",
-    "cypress:ci": "cypress run -c baseUrl=http://127.0.0.1:8000",
+    "cypress:ci": "cypress run -c baseUrl=http://127.0.0.1:8000 --record",
```
weerd: 👋 @DFurnes

weerd:
15 files but a few of them are deletions and others have straightforward changes to refer to the new component. So the main changes aren't that big 😬

resources/assets/components/LedeBanner/templates/MosaicTemplate.js
```diff
@@ -95,11 +95,11 @@ const MosaicTemplate = props => {
           {isAffiliated ? actionButton : signupButton}
 
           {affiliateSponsors.length ? (
-            <SponsorPromotion
+            <AffiliatePromotion
               className="mosaic-lede-banner__sponsor padding-top-lg clear-both"
               imgUrl={affiliateSponsors[0].fields.logo.url}
               title={affiliateSponsors[0].fields.logo.title}
-              text={affiliateCreditText}
+              text={{ value: affiliateCreditText, className: 'text-gray-400' }}
```
weerd: One thing to note, I needed a way to pass specific classes to the `<p>` that contains the text inside the `AffiliatePromotion` component. Passing the `text-gray-400` to the `AffiliatePromotion`'s `className` attribute wasn't enough since there's some Forge styling that targets `p` tags specifically. So I needed to be able to pass a class to the `p` and this was my idea. Maybe not ideal, but it works 🤷‍♂ Lemme know if y'all have a better idea!

resources/assets/components/utilities/AffiliatePromotion/AffiliatePromotion.js
```diff
@@ -0,0 +1,35 @@
...
+  }),
+  title: PropTypes.string,
+};
+
+AffiliatePromotion.defaultProps = {
+  className: null,
+  text: {
+    value: 'Powered by',
+    className: null,
+  },
```
weerd: Only caveat I saw with having the `text` be a "shaped" object was that if a `className` was passed and a `value` was not, then it would not trigger the default props to fill in `Powered by` as the value. Not sure why not (it might be a thing that only really works to fill in a default prop if it's a single item, and doesn't do it for nested items of a prop... 🤷‍♂

DFurnes: Yeah, I think React just naively checks "is this value defined" for default props. Because of that drawback, I think it could make sense to make these into two separate `text` and `textClassName` props. (It also has a side-benefit of working better with static analysis & editor autocompletes!)

weerd: Gotcha, ok! Yeah, I initially went for `textClassName` but it felt a little gross to me for whatever reason and the nested approach seemed more "collected". But since it doesn't help with default values I'll separate.

PR #1507, opened 2019-07-17
Marquee template landing page

weerd:
### What does this PR do?

This PR adds a new Marquee template for the Landing Page. This template will be used to run a Sixpack test on select campaigns to see how it performs versus the current Pitch template. 

While developing the new Marquee template it also provided an opportunity to revisit the CSS Grid we developed for the Story Page and revamp it. With @lkpttn help, we decided to move to a 12 column grid instead of a 16 column grid, so that our content can better fit within a grid that is divisible by 2, 3, and 4, allowing for those respective sized content areas.

This PR also includes some small tweaks to some utility classes and other small changes that came up due to the work on adding this new template.

**Responsive resizing of Marquee template**
Responsive Marquee Template

[Small](https://user-images.githubusercontent.com/105849/61408679-93dcae80-a8ae-11e9-9507-55e17e8091e7.png)

[Medium](https://user-images.githubusercontent.com/105849/61408677-93dcae80-a8ae-11e9-8a6f-9f654f7f5483.png)

[Large](https://user-images.githubusercontent.com/105849/61408676-93dcae80-a8ae-11e9-9b7f-dd3e21781a9b.png)

### Checklist

### DIFF
--- a/resources/assets/components/pages/LandingPage/landing-page.scss
+++ b/resources/assets/components/pages/LandingPage/landing-page.scss
@@ -2,6 +2,14 @@
 
 .marquee-landing-page {
   background-color: fuchsia;
+
+  // .primary {
+  // }
+
+  // .secondary {
+  //   @include media($large) {
+  //   }
+  // }
 }
 
 .pitch-landing-page {
--- a/resources/assets/components/pages/LandingPage/landing-page.scss
+++ b/resources/assets/components/pages/LandingPage/landing-page.scss
@@ -1,5 +1,9 @@
 @import '../../../scss/next-toolbox.scss';
 
+.marquee-landing-page {
+  background-color: fuchsia;
+}
+
 .pitch-landing-page {
   .container__block {
     // When inside a container wrapper, increase top/bottom padding
--- a/resources/assets/components/pages/LandingPage/landing-page.scss
+++ b/resources/assets/components/pages/LandingPage/landing-page.scss
@@ -1,6 +1,12 @@
 @import '../../../scss/next-toolbox.scss';
 
 .marquee-landing-page {
+  .cover-image {
+    img {
+      width: 100%;
+    }
+  }
+
   .marquee-banner {
     margin-bottom: $half-spacing;
     grid-column: full-start / full-end;
@@ -17,7 +23,7 @@
   .marquee-banner__headline-title {
     text-transform: uppercase;
     font-size: $font-large;
-    color: #141493;
+    color: $purple-500;
   }
 
   .marquee-banner__headline-subtitle {
--- a/resources/assets/components/pages/LandingPage/landing-page.scss
+++ b/resources/assets/components/pages/LandingPage/landing-page.scss
@@ -1,11 +1,7 @@
 @import '../../../scss/next-toolbox.scss';
 
 .marquee-landing-page {
-  background-color: fuchsia;
-
   .marquee-banner {
-    // background-color: maroon;
-
     margin-bottom: $half-spacing;
     grid-column: full-start / full-end;
 
@@ -32,12 +28,6 @@
   }
 
   .primary {
-    h1,
-    h2,
-    h3 {
-      display: none; // @REMOVE
-    }
-
     margin-bottom: $base-spacing;
   }
 
--- a/resources/assets/components/pages/LandingPage/landing-page.scss
+++ b/resources/assets/components/pages/LandingPage/landing-page.scss
@@ -3,13 +3,37 @@
 .marquee-landing-page {
   background-color: fuchsia;
 
+  .marquee-banner {
+    // background-color: maroon;
+
+    grid-column: full-start / full-end;
+
+    @include media($large) {
+      grid-column: wide-start / span 7;
+    }
+  }
+
+  .marquee-banner__headline-title {
+    text-transform: uppercase;
+    font-size: $font-large;
+    color: #141493;
+  }
+
+  .marquee-banner__headline-subtitle {
+    text-transform: uppercase;
+    font-size: 47px; // @TODO: not a standard pattern library size.
+    font-family: $font-league-gothic;
+    font-weight: $weight-normal;
+  }
+
   // .primary {
   // }
 
-  // .secondary {
-  //   @include media($large) {
-  //   }
-  // }
+  .secondary {
+    @include media($large) {
+      grid-row: 1 / span 2;
+    }
+  }
 }
 
 .pitch-landing-page {
--- a/resources/assets/components/pages/LandingPage/landing-page.scss
+++ b/resources/assets/components/pages/LandingPage/landing-page.scss
@@ -6,9 +6,14 @@
   .marquee-banner {
     // background-color: maroon;
 
+    margin-bottom: $half-spacing;
     grid-column: full-start / full-end;
 
     @include media($large) {
+      grid-column: wide-start / wide-end;
+    }
+
+    @include media($largest) {
       grid-column: wide-start / span 7;
     }
   }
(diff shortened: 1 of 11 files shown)


resources/assets/components/pages/LandingPage/landing-page.scss
```diff
@@ -1,5 +1,95 @@
...
+
+    @include media($largest) {
+      grid-column: wide-start / span 7;
+    }
+  }
+
+  .marquee-banner__headline-title {
+    text-transform: uppercase;
+    font-size: $font-large;
+    color: #141493;
```
DFurnes: Would it make sense to pull this out into a local `$purple` variable for readability?

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -0,0 +1,134 @@
...
+  title,
+}) => {
+  // @TODO: If this experiment is successful we should turn generating the series urls for
+  // the cover image photo at different sizes into a helper function!
+  const coverImageUrls = {
+    extraLarge: contentfulImageUrl(coverImage.url, '2232', '1000', 'fill'), // 2880 x 1620
+    large: contentfulImageUrl(coverImage.url, '1116', '500', 'fill'), // 1440 x 810
+    medium: contentfulImageUrl(coverImage.url, '720', '350', 'fill'), // 720 x 405
+    small: contentfulImageUrl(coverImage.url, '360', '200', 'fill'),
+  };
```
weerd: One nice aspect of this new design is that it has provided an opportunity to try out responsive images and the `srcSet` attribute, allowing the browsers to decide which image to download based on the size it needs to fill. In comparison to largest size versus the smallest size used, there's a 0.5MB in size savings for the small size!

resources/assets/components/pages/LandingPage/landing-page.scss
```diff
@@ -1,5 +1,95 @@
...
+
+.marquee-signup-button {
+  display: block;
+  margin: 0 auto $half-spacing;
+  max-width: 450px;
+  padding: $base-spacing;
+}
+
+// @TODO: this could become its own component (KeyValueInformationCard or some better name)
+// but since only an experiment now, leaving as is.
```
DFurnes: Sounds good. 👍

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -0,0 +1,134 @@
...
+  scholarshipAmount,
+  subtitle,
+  title,
+}) => {
+  // @TODO: If this experiment is successful we should turn generating the series urls for
+  // the cover image photo at different sizes into a helper function!
+  const coverImageUrls = {
+    extraLarge: contentfulImageUrl(coverImage.url, '2232', '1000', 'fill'), // 2880 x 1620
+    large: contentfulImageUrl(coverImage.url, '1116', '500', 'fill'), // 1440 x 810
+    medium: contentfulImageUrl(coverImage.url, '720', '350', 'fill'), // 720 x 405
```
DFurnes: What's the relationship between the requested dimensions and the ones listed in the comments?

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -0,0 +1,134 @@
...
+                      <dt>Deadline</dt>
+                      <dd>
+                        {format(endDate, 'MMMM do, yyyy', {
+                          awareOfUnicodeTokens: true,
+                        })}
+                      </dd>
+                    </React.Fragment>
+                  ) : null}
+                  {timeCommitment ? (
+                    <React.Fragment>
```
DFurnes: This is totally subjective, but now that we're on Babel 7 we can use [shorthand syntax](https://reactjs.org/docs/fragments.html#short-syntax) for these!

```js
<>
  <dt>Time</dt>
  <dd>5-10 hours</dd>
</>
```

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
DFurnes: Heck yeah!! I'm really excited about this. 🎉

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
weerd: Oh nice! Didn't realize or otherwise forgot! Good call 😄

resources/assets/components/pages/LandingPage/landing-page.scss
weerd: I'll probs do it as a Tailwind-esk variable, so `$purple-500`, middle of the road, or maybe higher since it's dark. Moving forward with Tailwind, it's nice that it uses a number scale for colors, since you rarely _just_ use the one color everywhere but usually want access to tints and shades. Instead of using `light, lighter, dark, darkest` etc.

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
weerd: May need to leave it in for now. My editor doesn't seem to have the same update and I lose all syntax highlighting when I use the shorthand syntax. Will look into an update for editor but in the meantime will leave as is 😬

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
weerd: Nothing really. I forgot to remove them. The ones in the comments are more our "old" largest size and then divided in half at each smaller size. But based on the mocks for the Marquee template this wasn't helpful and actually needed different sizes. I just had them there for reference.

weerd:
image
Found a small bug with the cover image not expanding the full width in the grid at a specific breakpoint. Not entirely sure why, but might be related to the ratio of the image reaching it's full size and not needing to spread, even though the browser _has_ chosen the larger image based on the `srcSet` declaration... 🤷‍♂ 

Setting the `img` to `width: 100%` solves it 👌

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
DFurnes: Ah bummer! Yeah, definitely no rush.

resources/assets/components/pages/LandingPage/landing-page.scss
```diff
@@ -1,5 +1,95 @@
...
+      grid-column: wide-start / wide-end;
+    }
+
+    @include media($largest) {
+      grid-column: wide-start / span 7;
+    }
+  }
+
+  .marquee-banner__headline-title {
+    text-transform: uppercase;
```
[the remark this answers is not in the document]
weerd: Yeah, so we're currently in kind of a weird zone in between really using utility classes to their fullest extent and still relying on collection of styles in specific named component classes.

There are a couple caveats at the moment, until we get Tailwind fully implemented:
- No way to deal with responsiveness in utility classes until we have Tailwind running and can do stuff like `class="s:inline m:block m:uppercase"`.
- The nice thing about tailwind is when we do make the switch we can still use component classes where it makes sense (very repeated set of utility classes) and use the `@apply` to basically include the declarations a utility class provides, but within a component class, so everything still works with expected rules but can reduce maintenance burdens. (see [extracting components](https://tailwindcss.com/docs/extracting-components/))

So for the time being, I'm kind of juggling between the two until we get Tailwind fully transitioned in! Hopefully soooooon! 👀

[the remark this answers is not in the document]
weerd: Also, don't use `.caps-lock`, while @DFurnes loves it, the standard in Tailwind is `.uppercase`. Although we could always just keep it, haha 😛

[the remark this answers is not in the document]
mendelB: Ahhhh ok way way cool -- I haven't looked too deeply into Tailwind yet but it's sounding like a dream! Thanks for clarifying!

> Also, don't use .caps-lock, while @DFurnes loves it, the standard in Tailwind is .uppercase. Although we could always just keep it, haha 😛
🍵

PR #1508, opened 2019-07-18
Relocating the info bar to internal components.

weerd:
### What does this PR do?

This PR relocates the `InfoBar` component to allow better customization on campaign pages. The new Marquee Template for the landing page did not require showing it, but because it was being added at the `CampaignRoute` parent component level, it would be included in every campaign related page. By moving it deeper, the children components can now decide whether they need to show it or not.

I also renamed the `InforBar` to `CampaignInfoBar` since it is more specific to Campaigns (outputting affiliates, etc) and moved the component into the `utilities` directory. 🎉 

Lastly, I also took this opportunity to do some cleanup, use `React.Fragment` to reduce some "divitis" and a little re-organization.

Also moved the `InfoBar` above the darker "Join millions of young people..." bar, to give it a nice gradient effect as it gets darker approaching the black footer. Prior to this it looked real odd with the dark bar, then lighter info bar and then black footer 😬 
image

### Checklist

### DIFF
--- a/resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
+++ b/resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
@@ -9,18 +9,14 @@ import { getScholarshipAffiliateLabel } from '../../../../helpers';
 import TextContent from '../../../utilities/TextContent/TextContent';
 import LedeBannerContainer from '../../../LedeBanner/LedeBannerContainer';
 import CallToActionContainer from '../../../CallToAction/CallToActionContainer';
-import CampaignInfoBar from '../../../utilities/CampaignInfoBar/CampaignInfoBar';
+import CampaignInfoBarContainer from '../../../CampaignInfoBar/CampaignInfoBarContainer';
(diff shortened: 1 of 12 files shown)


resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
```diff
@@ -128,6 +147,10 @@ PitchTemplate.propTypes = {
 
 PitchTemplate.defaultProps = {
   additionalContent: null,
+  affiliateCreditText: undefined,
+  affiliatePartners: [],
+  affiliateSponsors: [],
+  campaignLead: {},
```
DFurnes: Would it be cleaner to set up a Redux container component for the `InfoBar` that reads these from the store, rather than having to pass it in to the component each place it's used?

weerd: Hmmm, yeah potentially. Would mean popping it back out of the `utilities` directory if so.

weerd: But it would simplify things. I think I'll do that. Makes sense...

PR #1512, opened 2019-07-18
Run (another :) Sixpack Experiment for Source Overridden Signup Button Copy

mendelB:
### What does this PR do?

This PR moves the existing Signup Button feature allowing custom button copy overrides based on user's traffic source into a Sixpack Experiment so we can test the conversion metrics around this feature.

Also updated some of the comments for -hopefully- clarity.

We have this test before (see Pivotal ID#](https://www.pivotaltracker.com/n/projects/2019313/stories/155625023)). 

This will effectively deprecate this existing feature (albeit not completely since it'll just split it into an AB test). I've confirmed that we do not have any live campaigns utilizing this feature currently so we should be 👍

Our goal with this ticket was to run this on the Shred Hate campaign specifically, but I've opted to allow the test generally for any campaign using the `additionalContent.sourceActionText` (similiarly to what we've recently done with #1487 and #1509), ~instead of adding a separate 'run the source copy experiment' feature flag attribute.~ (I misunderstood -- and we _did_ add an explicit 'feature flag' field to control running the experiment in https://github.com/DoSomething/phoenix-next/pull/1512/commits/861797b15712dcaf40fd47aec16e822359c74b88)

Refs [Pivotal ID #165972814](https://www.pivotaltracker.com/story/show/165972814)

image

image

### DIFF
--- a/resources/assets/components/SignupButton/SignupButton.js
+++ b/resources/assets/components/SignupButton/SignupButton.js
@@ -51,6 +51,12 @@ const SignupButton = props => {
     });
   };
 
+  const signupButton = copy => (
+    <Button className={className} onClick={handleSignup}>
+      {copy}
+    </Button>
+  );
+
   // Have signups been disabled for this campaign?
   if (disableSignup) {
     return null;
@@ -60,11 +66,7 @@ const SignupButton = props => {
   const sourceOverride = get(sourceActionText, trafficSource);
   const buttonCopy = text || sourceOverride || campaignActionText;
 
-  return (
-    <Button className={className} onClick={handleSignup}>
-      {buttonCopy}
-    </Button>
-  );
+  return signupButton(buttonCopy);
 };
 
 SignupButton.propTypes = {
--- a/resources/assets/components/SignupButton/SignupButton.js
+++ b/resources/assets/components/SignupButton/SignupButton.js
@@ -2,8 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { get } from 'lodash';
 
-import { query, withoutNulls } from '../../helpers';
 import Button from '../utilities/Button/Button';
+import { query, withoutNulls } from '../../helpers';
 
 const SignupButton = props => {
   const {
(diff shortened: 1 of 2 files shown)


mendelB:
> This looks great! One question is whether we want a separate additional content item that is a boolean to indicate to participate in the test instead of the `sourceActionText` making that decision. I think it would make it more straightforward that a campaign is participating in the Sixpack test if there's an item like `"sixpackSignupButtonSourceOverride": true`, and that is only used as the conditional to decide whether to do the test or not. It would make it very clear with one look at the additional JSON content that this campaign is using it. This is kind of the pattern I've been trying with the past couple Sixpack experiments that you referenced in the description :)

Ahh you know what I see that I misinterpreted those attributes! For some reason, the `sixpack...` prefix escaped me - and I just viewed these as a plain feature toggle, instead of an explicit sixpack toggle. yikes. I think that is reasonable!

I _also_ opted not to add the additional attribute, since no one is using this feature (and I don't think it's documented) so it felt safe to do so. But if you think it's safer to indeed add that attribute - I can do so!

weerd:
> I also opted not to add the additional attribute, since no one is using this feature (and I don't think it's documented) so it felt safe to do so. But if you think it's safer to indeed add that attribute - I can do so!

Yeah, I think not necessarily "safer", but more makes it **immediately** obvious what it is for and also a good reminder that it needs to be removed when the experiment is done and formalized if it's something to keep or otherwise discard 😄

mendelB:
> > I also opted not to add the additional attribute, since no one is using this feature (and I don't think it's documented) so it felt safe to do so. But if you think it's safer to indeed add that attribute - I can do so!
> 
> Yeah, I think not necessarily "safer", but more makes it **immediately** obvious what it is for and also a good reminder that it needs to be removed when the experiment is done and formalized if it's something to keep or otherwise discard 😄

Gotcha, yeah, that does sound wise. Imma add that in real quick!

mendelB:
Ok, added a new sixpack field toggling the feature experiment in https://github.com/DoSomething/phoenix-next/pull/1512/commits/861797b15712dcaf40fd47aec16e822359c74b88

PR #1515, opened 2019-07-19
Fixed signup button on small screen devices for Marquee template.

weerd:
### What does this PR do?

This PR updates the Marquee Template for the Landing Page to make the signup button behave as a fixed button on the bottom of the viewport, only for small screen devices. From medium sized devices and up, it returns to the normal flow of the page. 

Example of behavior below:
Fixed Signup Button Marquee Template

Refs [Pivotal ID #166250301](https://www.pivotaltracker.com/story/show/166250301)

### Checklist

### DIFF
--- a/resources/assets/scss/base.scss
+++ b/resources/assets/scss/base.scss
@@ -51,10 +51,8 @@ p {
 // @TODO: Added for the Marquee template experiment, so the full footer
 // can be viewed when fixed signup button reaches the bottom of the page.
 .footer {
-  padding-bottom: 105px;
-
-  @include media($medium) {
-    padding-bottom: 0;
+  @include media($small) {
+    padding-bottom: 105px;
   }
 }
 
--- a/resources/assets/scss/base.scss
+++ b/resources/assets/scss/base.scss
@@ -48,6 +48,16 @@ p {
   }
 }
 
+// @TODO: Added for the Marquee template experiment, so the full footer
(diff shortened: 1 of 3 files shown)


resources/assets/scss/base.scss
```diff
@@ -48,6 +48,16 @@ p {
   }
 }
 
+// @TODO: Added for the Marquee template experiment, so the full footer
+// can be viewed when fixed signup button reaches the bottom of the page.
+.footer {
+  padding-bottom: 105px;
```
weerd: I needed to add this rule for the footer so that when a user scrolls to the bottom of the page with the fixed signup button, the button goes past the end of the footer so that the social icons, terms of service and privacy policy links can all be accessed. This will apply the extra padding to the footer for all small screen devices, since I couldn't specifically target it just for the Marquee Template. For the most part I figured it's not a big deal for there to be extra padding; we can address this better if needed once the Marquee Template experiment is completed.

mendelB: Ah fair enough! I'm sure there's some interesting way we can deal with this, but agreed that it's ok to table that until/if we formalize and complete this template!

It doesn't look too bad to me 
image


Curious though if this is a good case for using the `$small` media query? I know we like to style for mobile by default, but since this is a bit of a specific hack, would it be clearer, and help avoid needing the explicit `$medium` query below?

weerd: Yeah, I think that's a good call 👌

PR #1521, opened 2019-07-25
Starts on ReferralPage

aaronschachter:
### What does this PR do?

This PR begins on a Referral Page per [the Refer A Friend technical spec](https://docs.google.com/document/d/1hzw8eBNLTwFFG2KQFdEHdljJbDpgW1awjG3wYjy_A_Q/edit#).
* Defines referral page route as `/referral/:signupId`
* Repurposes `ReferralController` to display Referral Page,  removing the CSV routes and adding a TODO to 🔪  the rest that it seems we are no longer supporting
* Queries GraphQL by SignupID -- displays user name, still needs to query for campaign info

Need to find the new world equivalent of a campaign in GraphQL (or add if they don't exist) per [designs](https://projects.invisionapp.com/share/GYSBOGRSDUA#/screens/373205658)

* Cover Image
* Tagline

https://www.pivotaltracker.com/story/show/166457198

### Checklist


routes/web.php
```diff
@@ -51,6 +51,9 @@
     return redirect('us/'.$category.'/'.$slug);
 })->where('category', $categories);
 
+// Referral Pages
+$router->get('referral/{signupId}', 'ReferralController@show');
```
weerd: In the tech spec doc I misspoke, and /referral should have been plural. It’s something I’ve been trying to have us stick too but most endpoints should be plural /referrals?signupId=10660286 Also I think I prefer the query param version than passing the signup id as a url param, since the ID is not a referralID and thus could be confusing (i.e. it’s not /referrals/:referralID)

weerd: I could see us potentially wanting to show a user the status of their referral thus we would need /referrals/:id to point to the actual referral record if that makes sense 🙂

aaronschachter: Holding off changing this URL for now, I continued the thread in the tech spec.

aaronschachter:
I've pushed up some changes that query for campaign title and slug, and link out to the campaigns with the `referral_user_id` query string. I've been testing against this dev signup: https://dosomething-phoenix-de-pr-1521.herokuapp.com/referral/668

Note -- there aren't any plans for this to go live any time soon. This page still needs images, as well as the tagline (which I'm suspecting require GraphQL additions).

routes/web.php
weerd: Cool. I commented in the tech spec. But the gist is I really dig your suggestion for using `/join/:id` 💃

aaronschachter:
Updated this branch with the new `join/:signupId` route: https://dosomething-phoenix-de-pr-1521.herokuapp.com/join/668

I think this is at a good stopping point to merge if approved -- this won't be user facing until we add the link to the post-signup modal, which is quite a ways away. Outstanding tasks for a future PR:

- Display campaign cover image, tagline
- Display [envelope graphic](https://dosomething.slack.com/archives/C3ASB4204/p1564086503007100?thread_ts=1564085528.004400&cid=C3ASB4204) per design
- Using a config variable for the secondary campaign ID (as well as the description, which has a hardcoded signup count)
- Add any additional analytics, SEO info if required

app/Http/Controllers/ReferralController.php
```diff
@@ -9,11 +9,14 @@
...
 {
     /**
-     * ReferralController constructor.
+     * Show a referral page for given signup id.
+     * TODO: Remove all other functions in this class.
+     *
+     * @param  int $signupId
      */
-    public function __construct()
+    public function show($signupId, Request $request)
```
weerd: You can remove the injected `$request` since you don't end up using it.

app/Http/Controllers/ReferralController.php
```diff
@@ -9,11 +9,14 @@
...
 {
     /**
-     * ReferralController constructor.
+     * Show a referral page for given signup id.
+     * TODO: Remove all other functions in this class.
+     *
+     * @param  int $signupId
      */
-    public function __construct()
+    public function show($signupId)
```
weerd: This still feels a little weird to me that the ID specified for the Referral resource is not a referral ID but a signup ID. Buuut I can't think of a better solution. Fine for now and maybe we'll come up with something better later :)

resources/assets/components/pages/ReferralPage/ReferralPage.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import CampaignLink from './ReferralPageCampaignLink';
+
+// TODO: Allow override via config variable.
```
weerd: Pop an `@` in front of TODO so it'll show up when searching `@TODO` with all the rest of our growing list, haha.

app/Http/Controllers/ReferralController.php
aaronschachter: I hear that -- that's why I was looking to add this into a new `SignupController` at first. Seems easy enough to change out later.

routes/web.php
```diff
@@ -64,19 +67,8 @@
 $router->get('cache/{cacheId}', 'CacheController');
 $router->get('next/cache/{cacheId}', 'CacheController');
 
-// Referrals CSV export
```
weerd: Awesome. I think in a followup PR you should remove the rest of the old Referrals stuff we're no longer using. There's a few pieces including a content type in Contentful.

aaronschachter: Definitely going to remove the referrals stuff in a separate PR 🔥

aaronschachter:
This branch has been updated per our latest discussions and iteration of the [technical spec](https://docs.google.com/document/d/1hzw8eBNLTwFFG2KQFdEHdljJbDpgW1awjG3wYjy_A_Q/edit#).

The Referral Page now expects a `user_id` query parameter, and optional `campaign_id` query parameter. Examples:  

* With campaign: https://dosomething-phoenix-de-pr-1521.herokuapp.com/join?user_id=5d30ef993885ef00b01d1fe2&campaign_id=9004

* Without campaign: https://dosomething-phoenix-de-pr-1521.herokuapp.com/join?user_id=5d30ef993885ef00b01d1fe2

I've reverted the changes I made to the `ReferralController`, as it doesn't seem we need it in order to render the ReferralPage component in React, at least within the scope of this PR.

Again want to note that images and styles will be added at a later time when we're dev complete without final styles -- this page won't be user facing for quite some time.

routes/web.php
```diff
@@ -51,6 +51,9 @@
     return redirect('us/'.$category.'/'.$slug);
 })->where('category', $categories);
 
+// Referral Pages
+$router->view('join', 'app');
```
weerd: We should probs prefix with `us/`.

weerd: I think we'll be adding something in the future where it'll append `/us` to the main domain on any route that's missing it.

resources/assets/components/App.js
```diff
@@ -37,6 +38,7 @@ const App = ({ store, history }) => {
               <Route path="/us/blocks/:id" component={BlockPage} />
               <Route path="/us/campaigns/:slug" component={CampaignContainer} />
               <Route path="/us/:slug" component={PageDispatcherContainer} />
+              <Route path="/join" component={ReferralPage} />
```
weerd: Same as the route definition, add prefix, so `/us/join`.

resources/assets/components/pages/ReferralPage/ReferralPage.js
```diff
@@ -0,0 +1,70 @@
...
+import PropTypes from 'prop-types';
+
+import CampaignLink from './ReferralPageCampaignLink';
+
+// @TODO: Allow override via config variables.
+const SECONDARY_CAMPAIGN_ID = 7951;
+const SECONDARY_CAMPAIGN_PROMPT =
+  'In less than 5 minutes, you can join 193,242 young people putting an end to gun violence.';
+
+const ReferralPage = props => {
```
weerd: Since you're not doing an implicit return here, you could pop out the props into their own variables:

```js
const { firstName, primaryCampaignId, userId } = props;
```
And then just use the variables without the dot syntax. We do that in a lot of our components and makes it a bit more readable :)

resources/assets/components/pages/ReferralPage/ReferralPageCampaignLink.js
```diff
@@ -0,0 +1,56 @@
...
+      slug
+      callToAction
+      coverImage {
+        url(w: 150)
+      }
+    }
+  }
+`;
+
+const ReferralPageCampaignLink = props => (
```
weerd: Could this just technically be a `CampaignLink` to make it a more generic component that could be used elsewhere?

resources/assets/components/App.js
weerd: If using a container we prefer to name it as the component exists so there's less confusion, so:

```jsx
<Route path="/join" component={ReferralPageContainer} />
```

resources/assets/components/pages/ReferralPage/ReferralPageContainer.js
```diff
@@ -0,0 +1,41 @@
...
+const REFERRAL_PAGE_USER = gql`
+  query ReferralPageUserQuery($id: String!) {
+    user(id: $id) {
+      id
+      firstName
+    }
+  }
+`;
+
+const ReferralPageContainer = () => {
```
weerd: "Containers" in Phoenix components have a very specific function in relation to Redux, for helping map state items into props for the referenced component. I'm hesitant to name this in a similar fashion when it's not related to Redux... It defs confused me when looking at it and expecting something else...

resources/assets/components/pages/ReferralPage/ReferralPageCampaignLink.js
aaronschachter: I added this component into the `ReferralPage` directory because it does seem very specific to the Referral Page:

Screen Shot 2019-08-01 at 11 07 42 AM

It queries GraphQL for a given `campaignId`, appends the `referrer_user_id` query parameter to the URL, that hyperlink right column, and that DOSOMETHING.ORG footer as well.

We could make this more generic by passing various properties to a `CampaignLink` component:
* `queryVars` (Query string to append to URL)
* `displayDoSomethingFooter`, `displayHyperlinkColumn` - can't imagine we'd want these globally

I'd think this should be ok refactor in a separate pull request, as it doesn't seem like we're using blocks like this anywhere else to DRY.

aaronschachter:
Updated URL with the `/us/` prefix: https://dosomething-phoenix-de-pr-1521.herokuapp.com/us/join?user_id=5d30ef993885ef00b01d1fe2&campaign_id=9004

Also renamed: 
ReferralPageContainer -> ReferralPage
ReferralPage -> ReferralPageContent

resources/assets/components/pages/ReferralPage/ReferralPageCampaignLink.js
weerd: @aaronschachter thoughts on this?

weerd: Gotcha, cool cool. Yeah I think a separate refactor would be fine since it seems like this part:
image

is already very similar to the `LinkAction` component. But that component would maybe need some massaging to allow for Campaign link in this vain. Defs fine for a future refactor :)

resources/assets/components/pages/ReferralPage/ReferralPage.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import gql from 'graphql-tag';
+
+import ReferralPageContent from './ReferralPageContent';
+import Query from '../../Query';
+import ErrorBlock from '../../ErrorBlock/ErrorBlock';
+import { query } from '../../../helpers';
```
weerd: Something we've been trying to do in all our components is organize imports in line-length order, same as our PHP. Small nit-pick!

resources/assets/components/pages/ReferralPage/ReferralPageCampaignLink.js
mendelB: @aaronschachter @weerd  Could we get away with just using the [`Embed`](https://github.com/DoSomething/phoenix-next/blob/b96abb3a0df3a61ee8caa05f54a9f5b8b8b34dac/resources/assets/components/utilities/Embed/Embed.js#L1) component to render this Campaign Link? We default campaign `og:[content]` to be based on Title, CTA, and Cover Image, and the provider name will be `dosomething.org`. Here's an example https://dev.dosomething.org/us/blocks/WtJjirK4aAZP2eWGQHlkf

I guess we'd still need to wrap it in a graphql query for the `slug`, but once you have the URL, you should be able to simply pass it to an Embed with a `badged` prop for the link icon, and it'll do the rest of the work!

(Ooh and maybe even just pass the `slug` over as another param from the referral link to avoid another graphql query altogether?)

aaronschachter: This makes a lot of sense to re-purpose, it's got all of the styles we need. I'll look into this in a future PR (we've scheduled all styling work to be done at the end once all basic components have been added)

PR #1524, opened 2019-07-29
Adding Debug class to help with visualizing console output.

weerd:
### What does this PR do?

This PR adds a new `Dedug` service to help toggle showing/hiding of app specific console logs within any of our environments. Currently, we only really log useful event information to the console if we're on our local development environment, but it would be nice to be able to toggle whether to show or hide console logs on both QA and Production to easily be able to see what is firing. Now we can!

DS Logs Toggle

This new `Debug` class and associated `toggleLogs()` method can only be triggered from any Phoenix environment. 

The toggle value is stored in local storage and once toggled the Debug class will provide visual logs in the console for pre-setup custom event logs. To turn them back off, just toggle the setting back off. This setting only affects a user's personal browser client.

To help better track when analytics events are firing and coordinate that they are firing as expected and in what sequence, we would like to be able to see the events fired in the console, not only locally, but also on QA and Prod

I initially used `localForage` for the toggle value setting, but it's Promise-based approach proved to not be super helpful when events are firing and the page reloads. Since this is _only_ for debugging purposes, I figured it was alright to directly use the `localStorage` API.

Additionally, to be able to log when Google or Snowplow events fire, I needed to wrap the specific method each uses when triggering events. It allowed me to log and then continue the native event as expected. I'll be doing a bit of testing to ensure this all works as expected.

Lastly, the aim is to follow-up with an additional PR that adds a toggle to the Admin Dashboard dropdown that calls the same `Debug` class method you can otherwise call manually in the console!

:cactus:

### Checklist

### DIFF
--- a/resources/assets/init.js
+++ b/resources/assets/init.js
@@ -47,7 +47,8 @@ import { bindFlashMessageEvents } from './helpers/flash-message';
 import { bindAdminDashboardEvents } from './helpers/admin-dashboard';
 
 ready(() => {
-  debug().setupTrackerWrappers();
+  // Enable Debug tools.
+  debug();
 
   // Configure store & history.
   const history = historyInit();
--- a/resources/assets/init.js
+++ b/resources/assets/init.js
@@ -38,7 +38,7 @@ import './scss/gallery-grid.scss';
 import App from './components/App';
 
 // DOM Helpers
-import { ready } from './helpers';
+import { ready, debug } from './helpers';
 import { init as historyInit } from './history';
 import { bindTokenRefreshEvent } from './helpers/auth';
 import { trackAnalyticsPageView } from './helpers/analytics';
@@ -47,6 +47,8 @@ import { bindFlashMessageEvents } from './helpers/flash-message';
 import { bindAdminDashboardEvents } from './helpers/admin-dashboard';
 
 ready(() => {
+  debug().setupWrappers();
+
   // Configure store & history.
   const history = historyInit();
   const middleware = [thunk];
@@ -61,6 +63,13 @@ ready(() => {
 
   // Add event listeners for top-level navigation.
   bindNavigationEvents();
+
+  // const nativeSnowplow = window.snowplow;
+
+  // window.snowplow = function() {
+  //   nativeSnowplow.call(window, ...arguments);
+  //   debug().log('snowplow', ...arguments);
+  // };
 
   // If available, set User ID for Snowplow analytics.
   if (typeof window.snowplow === 'function' && window.AUTH.id) {
--- a/resources/assets/init.js
+++ b/resources/assets/init.js
@@ -47,7 +47,7 @@ import { bindFlashMessageEvents } from './helpers/flash-message';
 import { bindAdminDashboardEvents } from './helpers/admin-dashboard';
 
 ready(() => {
-  debug().setupWrappers();
+  debug().setupTrackerWrappers();
 
   // Configure store & history.
   const history = historyInit();
@@ -63,13 +63,6 @@ ready(() => {
 
   // Add event listeners for top-level navigation.
   bindNavigationEvents();
-
-  // const nativeSnowplow = window.snowplow;
-
-  // window.snowplow = function() {
-  //   nativeSnowplow.call(window, ...arguments);
-  //   debug().log('snowplow', ...arguments);
-  // };
 
   // If available, set User ID for Snowplow analytics.
   if (typeof window.snowplow === 'function' && window.AUTH.id) {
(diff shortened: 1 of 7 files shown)


weerd:
One nice note is that the _wrapping_ of the third-party script functions `window.dataLayer.push()` and `window.snowplow()` only happens to a staff member's specific browser client and if they choose to toggle to show the logs; otherwise the native third-party functions aren't touched, which is great for when debugging and entirely untouched for our member users 😄

resources/assets/services/Debug.js
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  /**
+   * Wrap specified event trackers in a custom fu nction to allow logging
+   * when events fire into the console for debugging.
+   *
+   * @return {void}
+   */
+  setupTrackerWrappers() {
+    if (typeof window.snowplow !== 'function') {
```
mendelB: Do we need a similar fail-safe for the Data Layer?

weerd: oh true. probs would be useful, haha

resources/assets/init.js
```diff
@@ -47,6 +47,8 @@ import { bindFlashMessageEvents } from './helpers/flash-message';
 import { bindAdminDashboardEvents } from './helpers/admin-dashboard';
 
 ready(() => {
+  debug().setupTrackerWrappers();
```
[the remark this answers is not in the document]
weerd: Oh, hmmm. That might be ideal. I can look into it. I can likely do a similar check like I do in the `log()` method at the very start.

resources/assets/services/Debug.js
```diff
@@ -0,0 +1,135 @@
+/* global window */
+
+import { isStaff } from '../helpers';
+import { googleLog, snowplowLog } from '../helpers/loggers';
+
+const KEY = 'DS_SHOW_LOGS';
```
DFurnes: > This new Debug class and associated toggleLogs() method can only be successfully triggered by Staff members when they are logged into Phoenix (on any environment):

Since this could theoretically be set by just manually adding this value to local storage via devtools, I think it's fine to remove any role-based authorization requirements when using the `toggle` function.

resources/assets/services/Debug.js
```diff
@@ -0,0 +1,135 @@
...
+
+    // We do not use rest params for these wrapper functions below, since it would convert the
+    // prototype object from an Object to an Array and potentially disable the native functionality!
+    // @see https://eslint.org/docs/rules/prefer-rest-params#suggest-using-the-rest-parameters-instead-of-arguments-prefer-rest-params
+
+    // Wrap the Snowplow function to allow logging events.
+    const nativeSnowplow = window.snowplow;
+
+    window.snowplow = function() {
+      nativeSnowplow.call(window, ...arguments); // eslint-disable-line prefer-rest-params
```
DFurnes: Let's use [`Function.prototype.apply`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply) for these wrapper functions so we can provide the arguments as an array (and then won't need to disable the linting rule here).

resources/views/partials/google_tag_manager_script.blade.php
```diff
@@ -6,6 +6,12 @@
         'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
         })(window,document,'script','dataLayer','GTM-{{config('services.analytics.google_tag_manager_id')}}');
     </script>
+@else
+    {{-- Mocked function to allow showing console logs. --}}
+    <script type='text/javascript'>
+        window.dataLayer = {};
+        window.dataLayer.push = function() {};
```
DFurnes: We can just initialize this as an array with `window.dataLayer = []` (that's what the snippet is doing above).

resources/assets/init.js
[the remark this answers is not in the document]
weerd: Yais! Works much better now and only runs the setup wrapper functions when Debug Logs are toggled on! Thanks @mendelB !

resources/assets/services/Debug.js
```diff
@@ -0,0 +1,159 @@
...
+  wrapGoogleTracker() {
+    if (!Array.isArray(window.dataLayer)) {
+      return;
+    }
+
+    const debug = this;
+
+    const nativeGtmDataLayerPush = window.dataLayer.push;
+
+    window.dataLayer.push = function(...args) {
```
DFurnes: Oh, did you get a chance to try using an arrow function here (to preserve the outer `this` context)?

weerd: Oh forgot about that. I can try quickly and see.

weerd: Woot woot! It works! Must've been something I did before with structure that was keeping the arrow functions from working correctly!

PR #1525, opened 2019-07-29
Badge modals on user profile

katiecrane:
### What does this PR do?

This PR completes work for badges in the user profile. A user is able to click on a badge to get more information on how they earned it, or information on how to earn it if they haven't yet.

This involves a decent amount of refactoring on the `<Badge>` component:
- All the text displayed with a badge is now passed as children. 
- `<BaseFigure>` is now used instead of `<Figure>`

#### Small
image

#### Medium
image

#### Large
image

#### Opening and closing modal
modal_profile

Copy and wireframes are linked in the card!

Refs [Pivotal ID #165719236](https://www.pivotaltracker.com/story/show/165719236)

### Checklist


resources/assets/components/Affirmation/Affirmation.js
```diff
@@ -54,15 +54,19 @@ const Affirmation = ({
...
                   earned
-                  boldText
                   className="badge padded"
                   size="medium"
                   name="signupBadge"
-                  text="1 Sign-Up"
-                  explainerText="Congratulations! You signed up for your first campaign...and
+                >
+                  <p>
+                    <strong>1 Sign-Up</strong>
```
DFurnes: Would it make sense to turn this into a `h4`? (It should have pretty much the same appearance as bolded paragraph text, but better semantic meaning for screen readers and the like).

resources/assets/components/pages/AccountPage/Badge.js
```diff
@@ -1,49 +1,48 @@
...
-    earned,
-    boldText,
-    explainerText,
-    size,
-    className,
-  } = props;
+  const { name, earned, size, className, children, showLock } = props;
   const badgeImageIndex = earned ? name : `${name}Locked`;
 
+  const media = showLock ? (
```
DFurnes: This is kinda subjective, but it's probably more readable to move the `showLock` ternary around to the lock image itself, and keep the `position: relative` wrapper in both cases (since we DRY up the otherwise duplicated badge image).

DFurnes: Another subjective suggestion – it might make more sense to have the `showLock` parameter determine whether the lock should be displayed for unearned badges (so `showLock && !earned` here). Then it can just be a constant and there's only the one `earned` prop that's being flipped.

resources/assets/components/pages/AccountPage/BadgesTab.js
```diff
@@ -3,165 +3,380 @@ import PropTypes from 'prop-types';
...
 
-const NEWSLETTER_BADGE = gql`
-  query SubscriptionTopicsBadgeQuery($userId: String!) {
-    user(id: $userId) {
-      emailSubscriptionTopics
-    }
+  showModal(name, earned, title, text) {
+    this.setState({
+      showModal: true,
+      modalName: name,
```
DFurnes: Instead of providing these four different state variables in order to toggle the modal, it'd be nice to have an object we could reference (so, for example, we might just set `{ showModal: true, modalName: 'signupBadge' }` and we'd reference the title, text, and "earned" criteria from a centralized place.

(And actually, we could even simplify that to `{ modal: 'signupBadge' } or ` { modal: null }` then to toggle!)

resources/assets/components/pages/AccountPage/BadgesTab.js
```diff
@@ -3,165 +3,380 @@ import PropTypes from 'prop-types';
...
-                  name="voterBadge"
-                  text="Registered Voter"
-                />
-              </li>
-            )}
-          </Query>
-        </ul>
+
+  render() {
+    const SIGNUP_COUNT_BADGE = gql`
```
DFurnes: Let's pull these queries out of the render function since they're constants.

resources/assets/components/Affirmation/Affirmation.js
katiecrane: Can confirm it looks _exactly_ the same!

resources/assets/components/pages/AccountPage/Badge.js
katiecrane: @DFurnes I'm a bit confused about what you mean - we don't want to display the lock for all `unearned` badges, that's why it's a separate parameter right now so I don't think that can be collapsed into one value?

resources/assets/components/pages/AccountPage/BadgesTab.js
katiecrane: So we would have several objects like this (one for each badge):
```
const signupBadge = {
  title: '1 SIGN-UP',
  earnedText:
    'Hey, congrats! You earned a badge for signing up for your first campaign. More badges ahead…!',
  unearnedText: 'Unlock this badge by signing up for a campaign.',
};
```
And then state would just be:
```
 this.state = {
      modalName: '',
      modalEarned: false,
    };
```
And if `this.state.modalName` then we could pull in the correct fields from the corresponding object?

resources/assets/components/pages/AccountPage/Badge.js
DFurnes: Sorry that wasn't clear at all! 🙃Let me try again!

I was thinking it might be easier to set `showLock={true}` for any badge where it _would_ show a lock if unearned (and if so, then calculate whether or not to show the lock _inside_ the Badge), rather than `earned={earned} showLock={!earned}`.

resources/assets/components/pages/AccountPage/BadgesTab.js
DFurnes: I was thinking like a big object!!

```js
const badges = {
  signupBadge: {
    title: '1 SIGN-UP',
    earnedText: 'Hey, congrats! .....',
    unearnedText: 'Unlock this badge by…',
  },
  onePostBadge: {
    title: '1 ACTION',
    // and so on...
};
```

Then, we could grab the appropriate title/text like `badges[this.state.modalName].title`!

DFurnes: And actually, if we included the `query` and `earned` callback in the object, we might even be able to build the grid of icons as a loop over the object?!

```js
const badges = {
  signupBadge: {
    title: '1 SIGN-UP',
    earnedText: 'Hey, congrats! .....',
    unearnedText: 'Unlock this badge by…',
    query: SIGNUP_COUNT_BADGE,
    earned: data => data.signupsCount > 0,
  },
  onePostBadge: {
    title: '1 ACTION',
    // and so on...
};
```

Definitely not a merge blocker, but a cool way to DRY this interface up if it works out.

resources/assets/components/pages/AccountPage/Badge.js
katiecrane: OH okay I follow now! It would simplify what we're getting when we create the badge in `<BadgeModal>`. Thanks!

resources/assets/components/pages/AccountPage/BadgesTab.js
katiecrane: AH one big object awesome! I'll go with that part and save the looping over the object for another time

resources/assets/components/Affirmation/Affirmation.js
katiecrane: Fixed in https://github.com/DoSomething/phoenix-next/pull/1525/commits/bd38600ce37fe1a368d35520512bfd0f3f7b3b94

resources/assets/components/pages/AccountPage/Badge.js
katiecrane: Fixed in https://github.com/DoSomething/phoenix-next/pull/1525/commits/bbf89383be6568bc2f2330aaff7f427d4124960c

resources/assets/components/pages/AccountPage/BadgesTab.js
katiecrane: Fixed in https://github.com/DoSomething/phoenix-next/pull/1525/commits/1d5bb57510eed3c0a8c70adb3920562b31d72405

resources/assets/components/pages/AccountPage/Badge.js
katiecrane: Fixed in https://github.com/DoSomething/phoenix-next/pull/1525/commits/0525ff6d6ddd0677b8b64dd4ffab867a5f443327

resources/assets/components/pages/AccountPage/BadgesTab.js
katiecrane: Fixed in https://github.com/DoSomething/phoenix-next/pull/1525/commits/c14165b78b4fa8bc5fe01d38c09bb65c6e7801bd

katiecrane:
@DFurnes suggestions have all been addressed, some bigger refactoring so I think this could benefit from 👀 again, thank you!

katiecrane:
Tested on review app woohoo here we go!

PR #1529, opened 2019-08-02
ReferralPage templates

aaronschachter:
### What does this PR do?

This PR continues the ReferralPage work started in #1521,  adding a minimal Alpha view ("Refer your friends and get a $5 gift card") of the Referral Page at `/us/refer-friends`

Example:  `/us/refer-friends?user_id=5d30ef993885ef00b01d1fe2&campaign_id=9003`

* Refactors `ReferralPageContent` component as `templates/BetaTemplate` to display content for the `/us/join` route introduced in #1521 

* Adds `templates/AlphaTemplate` to display content for the `/us/refer-friends` route 

Using `ReferralPage/templates` to display the similar yet different Referral Page views was a great suggestion made by @weerd in the [tech spec](https://docs.google.com/document/d/1hzw8eBNLTwFFG2KQFdEHdljJbDpgW1awjG3wYjy_A_Q/edit?ts=5d433511#heading=h.qj9jzfxbhzrl).

* https://www.pivotaltracker.com/story/show/167602877 - Alpha Template
* https://www.pivotaltracker.com/story/show/167501573 - Beta Template


aaronschachter:
Example page: https://dosomething-phoenix-de-pr-1529.herokuapp.com/us/refer-friends?user_id=5d30ef993885ef00b01d1fe2&campaign_id=9004

resources/assets/components/pages/ReferralPage/ReferralPage.js
```diff
@@ -15,27 +17,64 @@ const REFERRAL_PAGE_USER = gql`
...
 `;
 
-const ReferralPage = () => {
+const ReferralPage = props => {
   const userId = query('user_id');
   if (!userId) {
     return <ErrorBlock />;
   }
+  const campaignId = query('campaign_id');
+  const isAlphaTemplate = props.location.pathname === '/us/refer-friends';
```
mendelB: Do you think a nice alternative to needing the location & path here for determining the template, might be something similar way we render the template in places like the LedeBanner, with a single determining prop?

https://github.com/DoSomething/phoenix-next/blob/a0ebcc772c14e114026eda27083383627fcb8023/resources/assets/components/LedeBanner/LedeBanner.js#L11-L20

So maybe here we add a `template` prop, which we could set from the Router itself (`<ReferralPage template="alpha"/>`) so that within here we could simply do what we please based on the assigned `template`? (I _feel_ like that might be cleaner but I could be so wrong).

resources/assets/components/pages/ReferralPage/templates/AlphaTemplate.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const AlphaTemplate = props => {
+  const { firstName, primaryCampaignId, userId } = props;
+  // @TODO: This URL should be shortened via Bertly.
```
mendelB: You probably are aware already, but just in case -- I think we'll be able to repurpose the (magnificently name) [`SocialDriveAction` component](https://github.com/DoSomething/phoenix-next/blob/a0ebcc772c14e114026eda27083383627fcb8023/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js) for this template. The mock is based on it, and it's got the Bertly shortening, etc. built-in!

resources/assets/components/pages/ReferralPage/templates/BetaTemplate.js
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <React.Fragment>
+      <p>
+        {firstName} just signed up for this campaign from DoSomething.org. Once
+        you and {props.firstName} complete the campaign, you’ll both earn a $5
+        gift card!
+      </p>
+      {props.primaryCampaignId ? (
+        <div>
+          <CampaignLink campaignId={primaryCampaignId} userId={userId} />
```
mendelB: I had a thought about potentially using the `Embed` component for this `CampaignLink` functionality https://github.com/DoSomething/phoenix-next/pull/1521#discussion_r309872224 (may have gotten buried in that thread). Do you think it might make sense?

resources/assets/components/pages/ReferralPage/templates/AlphaTemplate.js
aaronschachter: Yup, taking a look in a next PR! Thanks for linking to it, would've taken me a bit to find 😸

resources/assets/components/pages/ReferralPage/ReferralPage.js
aaronschachter: Great feedback - I agree it'd be cleaner to set a prop vs check the path. Pushed up an update I made by following [this tutorial](https://tylermcginnis.com/react-router-pass-props-to-components/) to figure out how to pass props to a component rendered by the React Router.

resources/assets/components/App.js
```diff
@@ -37,7 +37,14 @@ const App = ({ store, history }) => {
...
               <Route path="/us/campaigns/:slug" component={CampaignContainer} />
-              <Route path="/us/join" component={ReferralPage} />
+              <Route
+                path="/us/join"
+                render={props => <ReferralPage {...props} template="beta" />}
+              />
+              <Route
+                path="/us/refer-friends"
+                render={props => <ReferralPage {...props} template="alpha" />}
+              />
```
weerd: Reviewing these a little late, but wondering if we could simplify these by just checking the slugs/path in the component to decide on the template. If slug is `join` then template is `beta` otherwise `alpha` 🤔

resources/assets/components/pages/ReferralPage/ReferralPage.js
```diff
@@ -15,27 +17,58 @@ const REFERRAL_PAGE_USER = gql`
...
 `;
 
-const ReferralPage = () => {
+const ReferralPage = props => {
   const userId = query('user_id');
   if (!userId) {
     return <ErrorBlock />;
   }
+  const campaignId = query('campaign_id');
+  const isAlphaTemplate = props.template === 'alpha';
```
weerd: If checking the slug like I mentioned in the prior comment, then you could probably get rid of this `isAlphaTemplate ` variable and do the check in the ternary below that currently uses the `isAlphaTemplate` variable for the logic.

resources/assets/components/pages/ReferralPage/ReferralPage.js
```diff
@@ -15,27 +17,58 @@ const REFERRAL_PAGE_USER = gql`
...
+                ) : (
+                  <BetaTemplate
+                    firstName={firstName}
+                    primaryCampaignId={campaignId}
+                    userId={userId}
+                  />
+                )}
+              </div>
+            </div>
+          </div>
```
weerd: I might also recommend just moving all of this JSX markup into the respective templates. Then this file is mostly just setup and deciding which template to show. There will be some duplication in the markup but it give each template the flexibility to change over time/look different from each other and simplify the template structure so you don't have a few `isAlphaTemplate` ternaries dispersed throughout.

resources/assets/components/pages/ReferralPage/templates/AlphaTemplate.js
```diff
@@ -0,0 +1,41 @@
...
+  let url = `/us/join?user_id=${userId}`;
+  if (primaryCampaignId) {
```
weerd: Small nit-pick but seeing it in a few places, add a carriage return between creating variables and following statements like `if` conditionals, etc 😬

resources/assets/components/pages/ReferralPage/ReferralPage.js
weerd: Oh I missed seeing this... looks like you and @mendelB already discussed it's more ideal to pass the prop than to check the path. I think I'm cool with that and makes sense, but I still think we should do something like the `switch` example Mendel linked and keep ALL the JSX/HTML inside the respective templates.

aaronschachter: That sounds good, I can make those changes in a new branch once #1538 is ready for merge.

PR #1530, opened 2019-08-02
Rethinking Debug class approach

weerd:
### What does this PR do?

This PR updates the `Debug` class to work a little better and not have issues with `window.snowplow()`. It takes a different approach specifically for Snowplow and instead of overriding the existing service function, `window.snowplow()`, it just wraps calls to this function within an `analyze()` helper.

While this should catch the majority of triggered Snowplow analytics events within Phoenix, it won't catch events where the script third-party Snowplow script makes calls to `window.snowplow()` on its own, since they won' be wrapped in the `analyze()` helper. It was a caveat that was hard to avoid given that the `window.snowplow()` would break if we tried to override it 😢, but for the most part, Phoenix directly triggers the majority of Snowplow events, so this tool is still useful in understanding whether they're firing as expected.

While reworking, it provided an opportunity to tackle an approach to also specifying _which_ logs to actually show!  I've removed the `toggleLogs()` method in favor of two distinct methods that are clearer (and easier to control), called `enableLogs()` and `disableLogs()`. The `enabledLogs()` method requires an array with the specified logs you want to display in the console. This will help reduce the noise in the console when wanting to only show a particular set of event logs that the developer cares about in the moment.

```js
window.DS.Debug.enableLogs(['google', 'snowplow']);
```

image

🌵 

### Checklist

### DIFF
--- a/resources/assets/services/Debug.js
+++ b/resources/assets/services/Debug.js
@@ -14,6 +14,7 @@ class Debug {
     window.DS = window.DS || {};
     window.DS.Debug = this;
 
+    // @TODO: call this once once document is ready?
     this.logExistingDataLayerEvents();
   }
 
@@ -49,7 +50,7 @@ class Debug {
 
     switch (type) {
       case 'google':
-        googleLog(data[0]);
+        googleLog(data);
         break;
 
       case 'snowplow':
--- a/resources/assets/services/Debug.js
+++ b/resources/assets/services/Debug.js
@@ -74,7 +74,7 @@ class Debug {
         break;
 
       default:
-        console.error('😢 No custom log formatter found.');
+        console.error(`😢 No ${type} log formatter found.`);
     }
   }
 
--- a/resources/assets/services/Debug.js
+++ b/resources/assets/services/Debug.js
@@ -11,8 +11,9 @@ class Debug {
     window.DS = window.DS || {};
     window.DS.Debug = this;
 
-    // @TODO: call this once once document is ready?
-    // this.logExistingDataLayerEvents();
+    // Set up any supported trackers that allow being wrapped in an override
+    // function, to help easily output logs whenever native event called.
+    this.setupTrackerWrappers();
   }
 
   /**
@@ -73,7 +74,7 @@ class Debug {
         break;
 
       default:
-        console.error('No custom log formatter found.');
+        console.error('😢 No custom log formatter found.');
     }
   }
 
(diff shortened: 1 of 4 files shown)


resources/assets/services/Debug.js
```diff
@@ -46,109 +74,96 @@ class Debug {
         break;
 
       default:
-        console.error('No custom log formatter specified.');
+        console.error('😢 No custom log formatter found.');
```
DFurnes: Should we include the missing `type` in this message?

resources/assets/services/Debug.js
```diff
@@ -46,109 +74,96 @@ class Debug {
...
-   * @return {void}
+   * @param  {Array} specifiedLogs
+   * @return {undefined}
    */
-  setupTrackerWrappers() {
-    if (Debug.getToggleValue()) {
-      this.logExistingDataLayerEvents();
-
-      this.wrapGoogleTracker();
+  enableLogs(specifiedLogs) {
```
DFurnes: It'd be nice to default this to including both `['google', 'snowplow']` for convenience!

resources/assets/services/Debug.js
weerd: Ah, good call 👌

resources/assets/services/Debug.js
weerd: I was wondering about this a bit... have also been thinking about how it would work as a UI item in the Admin Dashboard... I don't know if we'd want to assume which loggers to show off the bat 🦇 🤔

weerd:
## Notes 
Ended up needing to use 2 different approaches based on the analytics tracker, due to differences in their implementations.

For logging Snowplow events, it turned out to be best to use the `analyze()` helper function to wrap the calls to `window.snowplow()`. However, for Google Tag Manager, it ended up working best to actually override the `window.dataLayer.push()` method, which would allow us to automatically harness any call to the method and do our own logging whenever it was triggered, without having to manually wrap each call.

The former helped allow us to log Snowplow events without breaking Snowplow event requests to Fivetran, but it _will not_ capture any non-manual (system executed) calls to `window.snowplow()`. The latter lets us capture pretty much all non-manual _and_ manual GTM events triggered by adding items to the `window.dataLayer`. This lets us be more confident that we're capturing every analytics event for Google.

Just something to note when reviewing the logs for each service within this debug console output.

PR #1536, opened 2019-08-08
Badges in the modal after action submission

katiecrane:
### What does this PR do?

This PR adds action badges to the modal that comes up after a user completes an action. Whenever a user completes a share, photo, petition, or text action, if it was their 1st, 2nd, or 3rd action, the corresponding badge shows up in the modal that comes up after the action (only if the user is in the badge test group). In the modal is also a link to the badge area of the user profile.

I started extracting all the repeated badge code into a separate component, but ultimately I think it doesn't make sense to have a component that renders an action badge if a user has just earned one. Also, the way things stand now we can adjust copy across the 4 different actions if we end up needing to do that.

### Screenshots
Small
badge_modal_small

Medium
image

Large
image

Wireframes are linked in the card.

Refs [Pivotal ID #165721183](https://www.pivotaltracker.com/story/show/165721183)

### Checklist

### DIFF
--- a/resources/assets/components/actions/ShareAction/ShareAction.js
+++ b/resources/assets/components/actions/ShareAction/ShareAction.js
@@ -249,7 +249,7 @@ class ShareAction extends React.Component {
                               </Badge>
                             );
                           }
-                          if (postData.postsCount > 0) {
+                          if (postData.postsCount === 2) {
                             return (
                               <Badge
                                 earned
--- a/resources/assets/components/actions/ShareAction/ShareAction.js
+++ b/resources/assets/components/actions/ShareAction/ShareAction.js
@@ -20,6 +20,8 @@ import {
   showTwitterSharePrompt,
   withoutNulls,
 } from '../../../helpers';
+import Badge from '../../pages/AccountPage/Badge';
+import Query from '../../Query';
 
 export const ShareBlockFragment = gql`
   fragment ShareBlockFragment on ShareBlock {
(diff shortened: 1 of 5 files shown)


DFurnes:
> I started extracting all the repeated badge code into a separate component, but ultimately I think it doesn't make sense to have a component that renders an action badge if a user has just earned one. Also, the way things stand now we can adjust copy across the 4 different actions if we end up needing to do that.

That's fair! If badges are successful, though, I definitely think it's worth thinking about how we could DRY some of the up down the line. 🍃

resources/assets/components/actions/ShareAction/ShareAction.js
```diff
@@ -200,6 +216,87 @@ class ShareAction extends React.Component {
                 title="Thanks for sharing!"
                 className="modal__slide bordered rounded"
               >
+                <Query
```
mendelB: I wonder if it's worth thinking about removing the [`affirmationBlock` option](https://github.com/DoSomething/phoenix-next/pull/1536/files#diff-65c5c2b2e909188264fa961f7747ac4aL196) for this since it's the only Action we have it in, and it'd cause the Badge stuff to not show up.

Maybe worth thinking about post the badges test run!

katiecrane: @mendelB I actually meant to ask about that! Putting the badge in the `<Card>` is how I got it to show up, but I wasn't sure if that's because of the campaign I was on or if that's really the more common option?

mendelB: I _think_ It's the 'optimal' option, but it seems like there's quite a fair amount of Share Actions that use an Affirmation Block and thus would not show the card and badge. I can try and dig into the context for why we added this, and if it'd be reasonable for us to strip this feature entirely at this point.

Otherwise, we can at least update the field description on Contentful to warn that using the `Affirmation` field would hide the Badge info

katiecrane: I can add the badge to come up in that option as well, but I'd need to figure out a campaign that uses that option for testing, is there a way to find that out?

PR #1537, opened 2019-08-09
Use Embed in Referral Page beta template

aaronschachter:
### What does this PR do?

This PR uses the `Embed` component in the ReferralPage beta template to render campaign links per https://github.com/DoSomething/phoenix-next/pull/1521#discussion_r309872224.

It also ensures we don't display duplicate campaign links if the referral page happens to be shared from the secondary campaign, fixing https://www.pivotaltracker.com/story/show/167680643

Holding off on @mendelB's suggestion about adding additional logic into the `Embed` component https://github.com/DoSomething/phoenix-next/pull/1521#discussion_r309872224 for now:

> I guess we'd still need to wrap it in a graphql query for the slug, but once you have the URL, you should be able to simply pass it to an Embed with a badged prop for the link icon, and it'll do the rest of the work!
>
>(Ooh and maybe even just pass the slug over as another param from the referral link to avoid another graphql query altogether?)

We may revisit this if there's time before launch, but I think the 2 extra GraphQL queries we need to get the slug for a campaignId should be fine for our MVP. 

Fixes https://www.pivotaltracker.com/n/projects/2019429
Refs https://www.pivotaltracker.com/story/show/167633963

### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -400,7 +400,7 @@ function get_metadata($entry)
     return [
         'title' => data_get($entry, 'metadata.fields.title', $entry->title),
         'type' => 'article',
-        'description' => data_get($entry, 'metadata.fields.description', $entry->callToAction ? str_limit($entry->callToAction, 150) : null),
+        'description' => data_get($entry, 'metadata.fields.description', isset($entry->callToAction) ? str_limit($entry->callToAction, 150) : null),
         'url' => $entryType === 'campaign' ? $baseUrl.'/campaigns/'.$entry->slug : $baseUrl.'/'.$entry->slug,
         'facebook_app_id' => config('services.analytics.facebook_id'),
         'image' => [
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -400,7 +400,7 @@ function get_metadata($entry)
     return [
         'title' => data_get($entry, 'metadata.fields.title', $entry->title),
         'type' => 'article',
-        'description' => data_get($entry, 'metadata.fields.description', str_limit($entry->callToAction, 150)),
+        'description' => data_get($entry, 'metadata.fields.description', $entry->callToAction ? str_limit($entry->callToAction, 150) : null),
         'url' => $entryType === 'campaign' ? $baseUrl.'/campaigns/'.$entry->slug : $baseUrl.'/'.$entry->slug,
         'facebook_app_id' => config('services.analytics.facebook_id'),
         'image' => [
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -400,7 +400,7 @@ function get_metadata($entry)
     return [
         'title' => data_get($entry, 'metadata.fields.title', $entry->title),
         'type' => 'article',
-        'description' => data_get($entry, 'metadata.fields.description', null),
+        'description' => data_get($entry, 'metadata.fields.description', str_limit($entry->callToAction, 150)),
         'url' => $entryType === 'campaign' ? $baseUrl.'/campaigns/'.$entry->slug : $baseUrl.'/'.$entry->slug,
         'facebook_app_id' => config('services.analytics.facebook_id'),
         'image' => [
--- a/resources/assets/components/pages/ReferralPage/ReferralPage.js
+++ b/resources/assets/components/pages/ReferralPage/ReferralPage.js
@@ -55,6 +55,7 @@ const ReferralPage = props => {
                   />
                 ) : (
                   <BetaTemplate
+                    blockClassName={blockClassName}
                     firstName={firstName}
                     primaryCampaignId={campaignId}
                     userId={userId}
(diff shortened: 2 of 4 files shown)


aaronschachter:
Example with campaign 9004:

https://dosomething-phoenix-de-pr-1537.herokuapp.com/us/join?user_id=5d30ef993885ef00b01d1fe2&campaign_id=9004

Example with 7951 (only one campaign block should appear):

https://dosomething-phoenix-de-pr-1537.herokuapp.com/us/join?user_id=5d30ef993885ef00b01d1fe2&campaign_id=7951

@mendelB  - What doesn't seem to be working as expected per the suggestion is rendering the call to action in description... checking in GraphQL and I can see the `description` property is set to null.

Example request:
```
  embed(url:"https://dosomething-phoenix-de-pr-1537.herokuapp.com/us/campaigns/do-something-about-gun-violence") {
    thumbnailUrl
    title
    description
  }
```
Response:
```
{
  "data": {
    "embed": {
      "thumbnailUrl": "https://images.ctfassets.net/81iqaqpfd8fy/2GUGkcVAXme6W0qOIkgg6K/c8f1d809d79ef8e982158b8bb0fbc474/Screen_Shot_2018-05-07_at_12.16.20_PM.png?w=1200&h=1200&fm=jpg&fit=fill",
      "title": "Do Something: Gun Violence",
      "description": null
    }
  }...
```

mendelB:
> @mendelB - What doesn't seem to be working as expected per the suggestion is rendering the call to action in description... checking in GraphQL and I can see the description property is set to null.

Oh no! Looks like I lied. Looks like we try and pull the description from the `metadata` field, but don't default to setting it to the campaigns CTA https://github.com/DoSomething/phoenix-next/blob/b96abb3a0df3a61ee8caa05f54a9f5b8b8b34dac/app/helpers.php#L384

But I think we should be!

Can you update that line to default to `$entry->callToAction` instead of `null`? I think that'd do the trick here.

(My only reservation is if @weerd intentionally excluded the Call to action default, but I can't think of why (maybe character count concern if the CTA runs long for a campaign 🤔- so maybe we could truncate after 150 chars which seems to be the recommended length))

aaronschachter:
Added the `$entry->callToAction` as default if it exists, working as expected.

app/helpers.php
```diff
@@ -400,7 +400,7 @@ function get_metadata($entry)
     return [
         'title' => data_get($entry, 'metadata.fields.title', $entry->title),
         'type' => 'article',
-        'description' => data_get($entry, 'metadata.fields.description', null),
+        'description' => data_get($entry, 'metadata.fields.description', isset($entry->callToAction) ? str_limit($entry->callToAction, 150) : null),
```
mendelB: 🎉

resources/assets/components/pages/ReferralPage/ReferralPage.js
```diff
@@ -31,7 +31,8 @@ const ReferralPage = props => {
         if (!data.user) {
           return <ErrorBlock />;
         }
-
+        const blockClassName =
+          'general-page__block margin-vertical margin-horizontal-md';
```
mendelB: D'you think it might be easier to just add a top-level `<div>` with a `className="margin-horizontal-md"` and add the `margin-vertical-md` to each block individually within the template itself? This could allow us to kill this prop (albeit introducing an assumption to the template about its styling - though I think that's fair?)

(apologies if I'm micro analyzing styling that's gonna be updated later anyway 😬 )

aaronschachter: I'm open to making this change, my hope is that these will be the final styles... I think after this all we need is adding the envelope graphic. I was looking at the Who We Are for example markup to copy:

Screen Shot 2019-08-12 at 1 21 53 PM

mendelB: Oh great! The `general-page__block` doesn't actually have any styling purpose. I'm not sure why we add the `horizontal` individually, we probs could have wrapped the blocks in a div as well 🤷‍♂ 

It might've just been easier because we're mapping there anyways https://github.com/DoSomething/phoenix-next/blob/4b31b231618eead9a8cd46d0e32521a45afea6f1/resources/assets/components/pages/GeneralPage/GeneralPage.js#L97-L104

weerd:
> (My only reservation is if @weerd intentionally excluded the Call to action default, but I can't think of why (maybe character count concern if the CTA runs long for a campaign 🤔- so maybe we could truncate after 150 chars which seems to be the recommended length))

Catching up on all this... I believe there was a specific reason but don't remember offhand. Lemme think on it and see if I remember.

app/helpers.php
weerd: This kind of defeats the purpose of the `data_get()` a little... I'd recommend simplifying this by setting a variable outside the return and using the following:

```php
// Prefer metadata description otherwise the campaign call to action if available:
$description = data_get($entry, 'metadata.fields.description', null) ?: data_get($entry, 'callToAction', null);

// Then in the return:
return [
  ...
  'description' => str_limit($description, 160),  // The string limit on metadata description on Contentful is 160 chars
  ...
];
```

weerd: I'm a big fan of PHP's Elvis Operator `?:`  🕺

PR #1539, opened 2019-08-13
Progress on updating flow to separate login from registration flow.

weerd:
### What does this PR do?
This PR updates the flow to separate out the login flow from the registration flow. It includes a new route for `/register` that will be specifically for users registering via Northstar and a separate one for the login flow.

I ended up breaking the `getLogin()` method in the `AuthController.php` file into smaller protected methods in the class that now get called by both the `getLogin()` and `getRegistration()` methods. By splitting up the logic into smaller methods it helps make it a little bit more understandable instead of having a large block of logic, and also allows using the logic in both flows. The `getLogin()` method is only really different in that it adds an additional item to the `$options` array to specify overriding the authorization page so it heads to the `/login` on Northstar rather than `/register`.

To refrain from the PR getting to big I'll be doing the styling updates and fixes along with showing the new navigation link in an upcoming PR. For now I just hid it by commenting it out in Blade.

### Checklist

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -23,6 +23,78 @@ class AuthController extends Controller
     protected $redirectAfterLogout;
 
     /**
+     * Get destination metadata for Northstar login.
+     * Defaults to "title" in $options or null.
+     *
+     * @param  array  $queryParams
+     * @param  array  $options
+     * @return string
+     */
+    protected function getDestination($queryParams, $options = []) {
+        return array_get($queryParams, 'destination', array_get($options, 'title'));
+    }
+
+    /**
+     * Get options containing metadata for Northstar login.
+     *
+     * @param  array $queryParams
+     * @return array
+     */
+    protected function getOptions($queryParams) {
+        $options = array_get($queryParams, 'options', []);
+
+        if (is_string($options)) {
+            $options = (array) json_decode($options);
+        }
+
+        return $options;
+    }
+
+    /**
+     * Get the destination URL to redirect to on a successful login.
+     *
+     * @return string
+     */
+    protected function getUrl() {
+        $url = session('login.intended', $this->redirectTo);
+
+        // If there is an actionId in the session, set it as a query param on to the intended URL.
+        if (session('actionId')) {
+            $actionIdParam = 'actionId='.urlencode(session('actionId'));
+
+            $url .= str_contains($url, '?') ? '&'.$actionIdParam : '?'.$actionIdParam;
+        }
+
+        return $url;
+    }
+
+    /**
+     * Set session data used post-login and once user returns, redirected back to intended page.
+     *
+     * @param array $queryParams
+     */
+    protected function setSessionData($queryParams = []) {
+        if (array_has($queryParams, 'code')) {
+            return;
+        }
+
+        // The default post-login redirect will be to the previous page (if the user clicked
+        // "Log In" in the top navigation), or to the path defined in the $redirectTo property.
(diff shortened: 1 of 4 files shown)


resources/views/partials/navigation.blade.php
```diff
@@ -33,7 +33,8 @@
                         <li><a href="{{ route('logout') }}" class="secondary-nav-item" id="link--logout">Log Out</a></li>
                     </ul>
                 @else
-                    <a href="{{ route('login', isset($campaign) ? get_login_query($campaign) : null) }}">Log In</a>
+                    {{-- <a href="{{ route('login', isset($campaign) ? get_login_query($campaign) : null) }}">Log In</a> --}}
```
weerd: Commented this out for now, but will be included in upcoming PR that addresses the styling!

app/Http/Controllers/AuthController.php
```diff
@@ -92,4 +142,26 @@ public function getLogout(ResponseInterface $response)
...
     }
+
+    /**
+     * Handle a registration request to the application.
+     *
+     * @param  ServerRequestInterface $request
+     * @param  ResponseInterface      $response
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function getRegistration(ServerRequestInterface $request, ResponseInterface $response)
```
weerd: Debated whether to call this `getRegistration()` or `getRegister()` 🤔

app/Http/Controllers/AuthController.php
```diff
@@ -79,6 +68,67 @@ public function getLogin(ServerRequestInterface $request, ResponseInterface $res
...
+        $this->setSessionData($queryParams);
+
+        $url = $this->getUrl();
+
+        $options = $this->getOptions($queryParams);
+
+        $destination = $this->getDestination($queryParams, $options);
+
+        // Override the default Northstar authorization /register page and view /login instead.
+        $options['mode'] = 'login';
```
mendelB: My only reservation about all of this is having to repeat all this logic word for word in both actions just for the `$options` override for `'login'` requests. But I can't think of any alternate suggestions offhand!

weerd: True, but since there's a [`@TODO`](https://github.com/DoSomething/gateway/blob/master/src/AuthorizesWithOAuth2.php#L123) to merge both `$url` and `$destination` into the `$options` array, eventually this could be simplified and remove those two private methods in favor of just the `getOptions()` method.

mendelB: Ah, nice! That'd def be cleaner

app/Http/Controllers/AuthController.php
mendelB: Hmm yeah, I guess I can think of arguments for both... I personally think `getRegistration` is fine here!

resources/views/partials/navigation.blade.php
```diff
@@ -1,3 +1,9 @@
+@php
+    // @TODO: temporary until we can universalize campaign, pages, story-page entities into
+    // a single variable passed to the view.
+    $entity = isset($campaign) ? $campaign : null;
+@endphp
```
weerd: @mendelB @aaronschachter how do y'all feel about this? After https://github.com/DoSomething/phoenix-next/pull/1541 I wasn't super digging the repeated calls to `get_login_query()`. I could see us unifying the variable passed to the view and renaming it to something more generic in the future so it can apply to different page types and allow us to remove this bit. Figured a `@TODO` would suffice as a nice reminder for now.

mendelB: I dig this! I'm def looking forward to having this work across the site, so we can also capture the `contentful_id` etc. for everything.

PR #1540, opened 2019-08-14
ReferralPage cleanup

aaronschachter:
### What does this PR do?

This PR makes changes suggested per reviews post merge, see commits for linked change requests:
- Simplifies default description metadata
- Removes outer JSX from `ReferralPage` and into each component in the templates dir

Will link to alpha/beta template URL's once review app is created.

Still not styled, so no need for screenshots.

### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -397,13 +397,11 @@ function get_metadata($entry)
         $image = $entry->coverImage->url;
     }
 
-    // Prefer metadata description otherwise the campaign call to action if available:
     $description = data_get($entry, 'metadata.fields.description', null) ?: data_get($entry, 'callToAction', null);
 
     return [
         'title' => data_get($entry, 'metadata.fields.title', $entry->title),
         'type' => 'article',
-        // The string limit on metadata description on Contentful is 160 chars.
         'description' => str_limit($description, 160),
         'url' => $entryType === 'campaign' ? $baseUrl.'/campaigns/'.$entry->slug : $baseUrl.'/'.$entry->slug,
         'facebook_app_id' => config('services.analytics.facebook_id'),
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -397,10 +397,14 @@ function get_metadata($entry)
         $image = $entry->coverImage->url;
     }
 
+    // Prefer metadata description otherwise the campaign call to action if available:
+    $description = data_get($entry, 'metadata.fields.description', null) ?: data_get($entry, 'callToAction', null);
+
     return [
         'title' => data_get($entry, 'metadata.fields.title', $entry->title),
         'type' => 'article',
-        'description' => data_get($entry, 'metadata.fields.description', isset($entry->callToAction) ? str_limit($entry->callToAction, 150) : null),
+        // The string limit on metadata description on Contentful is 160 chars.
+        'description' => str_limit($description, 160),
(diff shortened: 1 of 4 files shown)


aaronschachter:
Alpha template: https://dosomething-phoenix-de-pr-1540.herokuapp.com/us/refer-friends?user_id=5d30ef993885ef00b01d1fe2&campaign_id=9004

Beta template: https://dosomething-phoenix-de-pr-1540.herokuapp.com/us/join?user_id=5d30ef993885ef00b01d1fe2&campaign_id=9004

app/helpers.php
```diff
@@ -397,10 +397,14 @@ function get_metadata($entry)
         $image = $entry->coverImage->url;
     }
 
+    // Prefer metadata description otherwise the campaign call to action if available:
```
weerd: Not sure if we need the comments here unless you think it's helpful; I feel like it's understandable enough if you know what `data_get()` does.

resources/assets/components/pages/ReferralPage/ReferralPage.js
```diff
@@ -34,34 +34,18 @@ const ReferralPage = props => {
 
         const firstName = data.user.firstName;
 
+        if (props.template === 'alpha') {
```
weerd: Does the `switch` style statement I mentioned in original PR not work here? I prefer how it reads plus makes it really easy to add additional templates in the future if needed.

aaronschachter: A switch felt like overkill when there are only two conditions. I think it'd make sense to refactor as one if there were more, but it seems very unlikely.

weerd: Maybe, but it does feel like it reads better. Plus the default could be an error block if the wrong or mistyped template is passed. Otherwise it's always the beta one.

weerd: Not vital for merge. Could try it in upcoming PR. Just something to keep in mind.

aaronschachter: I tried this out, and agree it does read a bit better -- but it caused an `consistent-return` error since line 32 returns an `ErrorBlock` isn't found. Can think about refactoring this for a future PR.

weerd: @aaronschachter I did a quick test and this seemed to work fine:

```js
<Query query={REFERRAL_PAGE_USER} variables={{ id: userId }}>
      {data => {
        if (!data.user) {
          return <ErrorBlock />;
        }

        const firstName = data.user.firstName;

        switch (props.template) {
          case 'alpha':
            return (
              <AlphaTemplate primaryCampaignId={campaignId} userId={userId} />
            );

          case 'beta':
            return (
              <BetaTemplate
                firstName={firstName}
                primaryCampaignId={campaignId}
                userId={userId}
              />
            );

          default:
            return <ErrorBlock />;
        }
      }}
    </Query>
```

 I think the linter was just mad that there was no `default` case and that also affected the `consistent-return` linting rule. Once added (`<ErrorBlock />` seemed appropriate for the default case, it seemed like it was happy and the linting passed 😄

weerd: Could also probs get rid of the `firstName` variable since it's only used in one spot and just inline it there.

PR #1538, opened 2019-08-13
Use SocialActionDrive in Referral Page alpha template

aaronschachter:
### What does this PR do?

This PR refactors the alpha template to render a `SocialActionDrive` component with the user's beta template URL, getting it a whole lot closer to the [design](https://projects.invisionapp.com/share/GYSBOGRSDUA#/screens/372969111):

* Adds optional `shareCardTitle` prop to `SocialActionDrive` for overriding the "Your Online Drive" Card header

* Adds optional `shareCardDescription` prop for intro text required by design

### Any background context you want to provide?

This seemed like a good spot to open for review. There are two biggies left, but shouldn't necessarily block merge as this isn't going live for a bit. 

1 - CSS. The `.social-drive-action` has a width set to 66.6666666666666%, so this looks terrible. Ideally we need to override this on the Referral Page to full width

~2 - Per design, the `Embed` should have title "Check Out Your Referral Page" and a custom description. I doubt we'd want the SEO title/description to read as this, so I've asked about this in our [product team Slack](https://dosomething.slack.com/archives/C09ANFQLA/p1565655219094400) to determine next steps on this item. I was thinking initially that we could pass optional `title` and `description` props to the `Embed`, which would entail another two extra additional props on a `SocialDriveActionContainer` in order to pass these along to the `Embed` within (feels kinda yucky and defeating of an `Embed`)~ Non-issue per https://github.com/DoSomething/phoenix-next/pull/1538#issuecomment-521033247

### What are the relevant tickets/cards?

* https://www.pivotaltracker.com/story/show/167602877
* https://www.pivotaltracker.com/story/show/167633963

### Checklist

* [ ] Documentation added for new features/changed endpoints.
* [ ] Added appropriate feature/unit tests.
* [ ] Added screenshot to PR description of related front-end updates on **small** screens.
* [ ] Added screenshot to PR description of related front-end updates on **medium** screens.
* [ ] Added screenshot to PR description of related front-end updates on **large** screens.

PR #1543, opened 2019-08-16
ReferralPageController

aaronschachter:
## What does this PR do?

This PR adds a`ReferralPageController` to provide metadata for the beta template of the Referral Page. The title and description will be used by the `Embed` in the `SocialActionDrive` on the alpha template per https://github.com/DoSomething/phoenix-next/pull/1538#issuecomment-521033247

Screen Shot 2019-08-15 at 5 58 21 PM

One of last tasks left for the Referral Page (to be done in a future PR) is to add the [envelope graphic](https://dosomething.slack.com/files/UGAB76GQZ/FLEUUA96F/refer_a_friend-01.svg) to the beta template, which we will want to use instead of the placeholder DS logo. 

* https://www.pivotaltracker.com/story/show/167602877
* https://trello.com/c/jVOPwcf2 

### Checklist

### DIFF
--- a/app/Http/Controllers/ReferralPageController.php
+++ b/app/Http/Controllers/ReferralPageController.php
@@ -28,7 +28,6 @@ class ReferralPageController extends Controller
         $firstName = $user->first_name;
         $title = 'DoSomething Good with '.$firstName;
         $callToAction = $firstName.' wants to make an impact with you, and bonus, you\'ll get $5 gift certificate for completing a DoSomething campaign.';
-        $facebookAppId = config('services.analytics.facebook_id');
 
         return response()->view('app', [
             'headTitle' => $title,
@@ -37,7 +36,7 @@ class ReferralPageController extends Controller
             'metadata' => [
               'title' => $title,
               'description' => $callToAction,
-              'facebook_app_id' => $facebookAppId,
+              'facebook_app_id' =>  config('services.analytics.facebook_id'),
               // @TODO: Replace with envelope graphic once added to beta template.
               'image' => [
                 'url' => 'https://forge.dosomething.org/resources/ds-logo-landscape.png',
@@ -45,11 +44,6 @@ class ReferralPageController extends Controller
                 'height' => '630',
               ],
               'url' => $request->fullUrl(),
-            ],
-            'socialFields' => [
-              'title' => $title,
-              'callToAction' => $callToAction,
-              'facebookAppId' => $facebookAppId,
             ],
         ])->cacheableWhenAnonymous();
     }
--- a/app/Http/Controllers/ReferralPageController.php
+++ b/app/Http/Controllers/ReferralPageController.php
@@ -29,6 +29,7 @@ class ReferralPageController extends Controller
         $title = 'DoSomething Good with '.$firstName;
         $callToAction = $firstName.' wants to make an impact with you, and bonus, you\'ll get $5 gift certificate for completing a DoSomething campaign.';
         $facebookAppId = config('services.analytics.facebook_id');
+
         return response()->view('app', [
             'headTitle' => $title,
             // @TODO: Create an $entity object to pass to the get_metadata helper
--- a/app/Http/Controllers/ReferralPageController.php
+++ b/app/Http/Controllers/ReferralPageController.php
@@ -29,9 +29,10 @@ class ReferralPageController extends Controller
         $title = 'DoSomething Good with '.$firstName;
         $callToAction = $firstName.' wants to make an impact with you, and bonus, you\'ll get $5 gift certificate for completing a DoSomething campaign.';
         $facebookAppId = config('services.analytics.facebook_id');
-
         return response()->view('app', [
             'headTitle' => $title,
+            // @TODO: Create an $entity object to pass to the get_metadata helper
+            // (and refactor get_metadata helper to expect an $entity instead of $campaign)
             'metadata' => [
               'title' => $title,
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/ReferralPageController.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $user = gateway('northstar')->getUser('id', $request->query('user_id'));
+
+        if (! $user) {
+            abort(404);
+        }
+
+        $firstName = $user->first_name;
+        $title = 'DoSomething Good with '.$firstName;
+        $callToAction = $firstName.' wants to make an impact with you, and bonus, you\'ll get $5 gift certificate for completing a DoSomething campaign.';
```
aaronschachter: This is placeholder -- in the [content request](https://trello.com/c/jVOPwcf2), I noted this should be 160 chars max.

app/Http/Controllers/ReferralPageController.php
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Display the beta Referral page.
+     *
+     * @return \Illuminate\View\View
+     */
+    public function show(Request $request)
+    {
+        $user = gateway('northstar')->getUser('id', $request->query('user_id'));
+
+        if (! $user) {
```
weerd: I wonder if the abort should happen earlier if there isn't even a `user_id` query param, so then the request to Northstar via Gateway doesn't even need to happen 🤔 

```php
public function show(Request $request)
{
  $userId = $request->query('user_id');

  if (! $userId) {
    abort(404);
  }

  // ...
}
```

aaronschachter: We would still need to abort if we don't find a user though..

aaronschachter: .. but yes makes sense to add that check/abort before! Can add

app/Http/Controllers/ReferralPageController.php
```diff
@@ -0,0 +1,48 @@
...
+                'url' => 'https://forge.dosomething.org/resources/ds-logo-landscape.png',
+                'width' => '1200',
+                'height' => '630',
+              ],
+              'url' => $request->fullUrl(),
+            ],
+            'socialFields' => [
+              'title' => $title,
+              'callToAction' => $callToAction,
+              'facebookAppId' => $facebookAppId,
```
weerd: Just confirming, the `metadata` and `socialFields` here are being set manually since this isn't technically a page on Contentful with associated Metadata content type, right?

Might be a future `@TODO` to allow passing custom object/array to the `get_metadata()` that returns expected output for open graph stuff, etc along with the current approach of grabbing it from the `$entry` object 🤔

app/Http/Controllers/ReferralPageController.php
```diff
@@ -0,0 +1,48 @@
...
+              'facebook_app_id' => $facebookAppId,
+              // @TODO: Replace with envelope graphic once added to beta template.
+              'image' => [
+                'url' => 'https://forge.dosomething.org/resources/ds-logo-landscape.png',
+                'width' => '1200',
+                'height' => '630',
+              ],
+              'url' => $request->fullUrl(),
+            ],
+            'socialFields' => [
```
weerd: Oh, I need to double-check, but I think you don't even need to provide this `socialFields` item and data. We only do it in other controllers because some of those pages have old components that rely on that info, but moving forward we're aiming to deprecate and stick with just the `metadata` property that covers all social sharing needs.

app/Http/Controllers/ReferralPageController.php
aaronschachter: Yes - manually setting because this all Refer A Friend content is hardcoded.

app/Http/Controllers/ReferralPageController.php
aaronschachter: I was following by example, and per this comment:
https://github.com/DoSomething/phoenix-next/blob/56cb3f375f8f074b386a9c449326d8734bdc177c/app/Http/Controllers/CampaignController.php#L71-L74

A few other controllers are setting both (`CampaignController`, `CategorizedPageController`, `PageController`, etc)

aaronschachter:
https://dosomething-phoenix-de-pr-1543.herokuapp.com/us/refer-friends?user_id=5d30ef993885ef00b01d1fe2 is an example showing the `Embed` of the `SocialActionDrive` picking up the title and description (and also an example of the CSS rule we'll eventually need to change in the `SocialActionDrive` before this is ready for prime-time

app/Http/Controllers/ReferralPageController.php
weerd: Yep yep. It's just that those controllers haven't been updated to remove it. If you compare `metadata.blade.php` and `social.blade.php` you'll see there's overlap and we should just remove the use of the old Blade partial. I think we've left it in those other controllers because we need to confirm the set variable isn't used elsewhere. But since Referrals is a totally new thing I think it's safe to not include :)

aaronschachter: Ok! ✂️

app/Http/Controllers/ReferralPageController.php
weerd: Ah! Valid point, haha. Not necessary for merge, but maybe `redirect()`-ing would be more ideal for one or both of the scenarios? Instead of ending the experience outright? 🤔

PR #1544, opened 2019-08-16
Prefixing register endpoint to resolve issue with prior 301 redirect.

weerd:
### What does this PR do?

This PR updates the new `/register` route endpoint to `/us/register` since a while back a vanity URL was set up as a `301` redirect from `/register` to `vote.dosomething.org`. Being a `301` this redirect is permanently cached in a user's browser unless we set up a redirect back or potentially try to bust the browser cache with some cache control headers 🤷‍♂ For now this should work fine an allow us to move forward.

Some context on `301` browser caching via [stack overflow](https://stackoverflow.com/questions/9130422/how-long-do-browsers-cache-http-301s/21396547#21396547)

### Checklist

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -78,6 +78,7 @@ class AuthController extends Controller
      */
     protected function setSessionData($queryParams = [])
     {
+        // @see Northstar Authorization Code Grant: https://git.io/fjd8N
         if (array_has($queryParams, 'code')) {
             return;
         }
--- a/routes/web.php
+++ b/routes/web.php
@@ -12,7 +12,7 @@ $router->redirect('/', '/us');
 $router->get('/us', 'HomePageController');
 
 // Authentication
-$router->get('/register', 'AuthController@getRegistration')->name('register');
+$router->get('us/register', 'AuthController@getRegistration')->name('register');
 $router->get('next/login', 'AuthController@getLogin')->name('login');
 
 $router->get('next/logout', 'AuthController@getLogout')->name('logout');


app/Http/Controllers/AuthController.php
```diff
@@ -78,6 +78,7 @@ protected function getUrl()
      */
     protected function setSessionData($queryParams = [])
     {
+        // @see Northstar Authorization Code Grant: https://git.io/fjd8N
```
mendelB: 🙌

aaronschachter:
Looks ok to me, but I don't have ability to approve.

Should the navigation be updated as well?  https://github.com/DoSomething/phoenix-next/blob/56cb3f375f8f074b386a9c449326d8734bdc177c/resources/views/partials/navigation.blade.php#L43

routes/web.php
```diff
@@ -12,7 +12,7 @@
 $router->get('/us', 'HomePageController');
 
 // Authentication
-$router->get('/register', 'AuthController@getRegistration')->name('register');
+$router->get('us/register', 'AuthController@getRegistration')->name('register');
```
mendelB: There are some #global implications here right? Though I guess since this is primarily an internal endpoint for now, and since this experience shouldn't need to be segmented by language anyways (and since this is a 'for now' solution!) this isn't an issue 🤔

weerd:
> Should the navigation be updated as well?

@aaronschachter the nav item is using a named route to `'register'` declared in the `routes/web.php` file. Named routes are nice since the underlying route can change but always keep the same name throughout the codebase :)

app/Http/Controllers/AuthController.php
weerd: oh yeah, my memory sucks and I appreciated @aaronschachter linking me to the NS docs to remind me, haha. Figured it'd help for when I forget again in the future 🔮 👴

routes/web.php
weerd: Yep totally. I thought about that, but hopefully by the time we go #global again, we'll have figured out how to bust-that_cache™!

PR #1545, opened 2019-08-16
CTA Register Popover (design implementation / v1)

mendelB:
### What does this PR do?

This PR adds a CTARegisterPopover component as a base design implementation.

Also adds a `redirectToNorthstar` helper method to gather the necessary data from `state` and query params, generate a Northstar redirect URL, and actually redirect.

### Any background context you want to provide?
This will show up on the `GeneralPage` (limited to `/articles/:slug` & `/facts/:slug` pages), prompting visitors to register. It will eventually pop up after some specified time and be dismissable. Copy also will be updated based on the page taxonomy.

We tried doing this using `position: sticky` but ran into a host of issues -- I think the primary one being that philosophically, this doesn't necessarily quite fit into the `position: sticky` since it doesn't occupy a logical place _within_ the container, but rather exists in specific position _on top_ of it.

🎩 💁‍♂This is a joint programming effort with the great @lkpttn 🎉 

One notable current bug -- The popover will be fixed all the way down through the footer on [mobile (📸)](https://user-images.githubusercontent.com/12417657/63187332-9d337500-c02c-11e9-9322-fe1372b39438.png) which is obviously not optimal. It is something we can potentially fix with JS - and which we can address later.

### What are the relevant tickets/cards?

Refs [Pivotal ID #167881873](https://www.pivotaltracker.com/story/show/167881873)

### Checklist

* [ ] Documentation added for new features/changed endpoints.
* [ ] Added appropriate feature/unit tests.
* [x] Added screenshot to PR description of related front-end updates on **small** screens.
* [x] Added screenshot to PR description of related front-end updates on **medium** screens.
* [x] Added screenshot to PR description of related front-end updates on **large** screens.

📸 
- [Desktop](https://user-images.githubusercontent.com/12417657/63186633-f4d0e100-c02a-11e9-8b08-bf181df02583.png)
- [Tablet](https://user-images.githubusercontent.com/12417657/63186661-07e3b100-c02b-11e9-9b35-ae0e8e4a728a.png)
- [Mobile](https://user-images.githubusercontent.com/12417657/63186717-2a75ca00-c02b-11e9-9e6f-bfc46b335d67.png)

[Screen Grab](https://p65.f3.n0.cdn.getcloudapp.com/items/NQudwb48/Kapture+2019-08-16+at+13.41.50.gif?v=ede876ec96ef38659ca480b17b9c1fc1)

PR #1546, opened 2019-08-16
Require authentication for the Alpha ReferralPage 

aaronschachter:
### What does this PR do?

This PR modifies the `/us/refer-friends` Alpha Referral Page ("Want free stuff?") to require authentication, and uses the logged in user's ID instead of relying on a `user_id` query parameter. With that in place, it makes sense to refactor a single `ReferralPage` component with templates instead as separate `AlphaPage` and `BetaPage` components, where only the `BetaPage` needs to query GraphQL by the `user_id` query parameter. 

There are no visible front-end changes, this is just moving around the existing code to simplify it.

 I think this is much cleaner, and also makes sense for the Alpha page to be limited to the logged-in user's ID. Will add sample URL's once review app is built.

https://www.pivotaltracker.com/story/show/167602877

### Checklist


aaronschachter:
Hmm, this review app keeps returning a `Client authentication failed` error when I try to login 😢 so is the one over in #1545 - https://dosomething-phoenix-de-pr-1545.herokuapp.com/us

weerd:
Ah, that's one of the current caveats with the review apps. Since the domains are generated based on branch ID, we can't specify a Redirect URI for OAuth flow in the `dev-oauth` client on Aurora. I think you could either edit the client for dev and add `https://dosomething-phoenix-de-pr-1545.herokuapp.com/us/next/login`, remembering to remove it after merging... or just wait to test this on dev.ds.org after merge.

Known issue, but not sure if we have a sufficient proposed solution to resolve.

aaronschachter:
Ah yes - I'll fix the order of the import statements.

I did consider what you mention about the Alpha page no longer being able to be hyperlinked... but I think for the most part we'd always be prompting someone over email or text to share the Beta page (here's an example: https://app.goabstract.com/share/e74d31e1-2e56-49c7-a385-5bb98ab41313?collectionLayerId=13e05ce7-d3db-41a0-80ae-d688f81dd017&mode=design)

It's unlikely it'd happen but it'd also be somewhat strange to access another user's alpha page by changing the `user_id` parameter (whereas on the beta page, i could have multiple friends who are referring me to join DS)

mendelB:
> Ah yes - I'll fix the order of the import statements.
> 
> I did consider what you mention about the Alpha page no longer being able to be hyperlinked... but I think for the most part we'd always be prompting someone over email or text to share the Beta page (here's an example: https://app.goabstract.com/share/e74d31e1-2e56-49c7-a385-5bb98ab41313?collectionLayerId=13e05ce7-d3db-41a0-80ae-d688f81dd017&mode=design)
> 
> It's unlikely it'd happen but it'd also be somewhat strange to access another user's alpha page by changing the `user_id` parameter (whereas on the beta page, i could have multiple friends who are referring me to join DS)

Gotcha, ok that makes good sense! We'd just be sending them their own share link directly!

PR #1551, opened 2019-08-21
A minimalist Affirmation Refer Friends Link

aaronschachter:
### What does this PR do?

This PR branches off of #1546 (which still needs review) to add a minimal link to the Alpha Referral Page in the Affirmation when a `refer-friends` query parameter is present. Once this and the other Referral Pages are styled, we can remove this query parameter check -- as this feature will be rolled out for all users (see [Pivotal thread](https://www.pivotaltracker.com/n/projects/2019429/stories/167656050/comments/205807100))

e.g. http://phoenix.test/us/campaigns/test-teens-for-jeans/action?refer-friends=true

Screen Shot 2019-08-21 at 11 29 55 AM

I'm feeling a bit over my head on how to get started styling this per [the design](https://projects.invisionapp.com/share/GYSBOGRSDUA#/screens/372969109)- but figured I'd open this PR as is to get reviewed on the general approach, e.g:

*  is it ok to require a `campaignId` prop in the `Affirmation`?
* How can the "Continue to campaign" link (not rendered yet) close the modal?
* Are there places in the codebase I can look to for guidance on how to style this block, and how to add the envelope image for both this block and the [Beta page](https://projects.invisionapp.com/share/GYSBOGRSDUA#/screens/373205658)? 

Refs https://www.pivotaltracker.com/story/show/167501569

### Checklist


mendelB:
> is it ok to require a campaignId prop in the Affirmation?

I believe the only other place this Affirmation can technically be rendered, is from the Share Action's [`affirmationBlock` field](https://github.com/DoSomething/phoenix-next/blob/bf9d0170dfdee4da6fb8f12a87dd823405009915/contentful/content-types/shareAction.js#L113) which would pop up in a modal once a user successfully shares. This field, in general, is one big headache (we had the same exact issue with badges).

I ran a script across Contentful and I don't see _any_ active `ShareAction` with an `Affirmation` referenced in this field, so I'd venture to say we can remove this type from the allowed list of references and call it a day.

One thought -- should we rename this component to `CampaignSignupAffirmation` so that it's scoped functionality is more explicit?

mendelB:
> I'm feeling a bit over my head on how to get started styling this per the design

#1548 (though not yet approved so take it with a grain of salt) might be a good reference for similar design (wherein we have some content and a button set against a backdrop of an svg image).

In this case, since the image will be referenced from two places, maybe worth storing the svg in the images directory? https://github.com/DoSomething/phoenix-next/blob/bf9d0170dfdee4da6fb8f12a87dd823405009915/resources/assets/images/

mendelB:
> How can the "Continue to campaign" link (not rendered yet) close the modal?

OOf, this might be tricky, because this Affirmation is rendered within a modal which uses a `onClose` prop for dismissal, but it's nested a few levels deep... 🤔 

https://github.com/DoSomething/phoenix-next/blob/bf9d0170dfdee4da6fb8f12a87dd823405009915/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js#L40-L44

resources/assets/components/Affirmation/Affirmation.js
```diff
@@ -90,13 +91,27 @@ const Affirmation = ({
         className="padding-bottom-md padding-horizontal-md"
       />
     ) : null}
+
+    {query('refer-friends') ? (
```
mendelB: Do you think it's worth throwing this behind an `env` variable feature flag instead?

aaronschachter: As is, I think we'd be better off using the secluded query string parameter until this is styled, to avoid unnecessary stress / confusion for why the Affirmation Refer A Friend link looks so craigslist on QA. Ideally the other pages styles need to be finished as well.

mendelB: Ah ok I feel you -- a more scoped feature flag if you will.

> so craigslist

😂

aaronschachter:
I'm going to hold on any big renames, I don't think they're within scope of this PR... and I don't know enough about the history to have strong opinions here. One thought is making the `campaignId` property optional, and only display the Refer A Friend link if the `refer-friends` query string parameter exists as well as a `campaignId` prop.

mendelB:
> I'm going to hold on any big renames, I don't think they're within scope of this PR... and I don't know enough about the history to have strong opinions here. One thought is making the `campaignId` property optional, and only display the Refer A Friend link if the `refer-friends` query string parameter exists as well as a `campaignId` prop.

I think that's a fair concession until we formalize this as a Campaign affirmation component specifically. (Though I do feel like we've made that choice with badges, and think it's the right call based on how this component is used within Phoenix). But agreed - any renaming is out of scope of this PR -- just something I think we should keep in mind.

PR #1552, opened 2019-08-21
Fixing registration authorization issue and updating auth route.

weerd:
### What does this PR do?

This PR provides a fix for both the issue we were experiencing on QA when testing registration, where the server would throw a `500` error and indicating that `The OAuth state field did not match`. Turns out there's a custom middleware we use that I missed updating which controls which routes to persist session data on if the flow redirects outside the app. It was specified as `next/login` but the new route for authentication was `/register`. Thus on the return trip from Northstar during the OAuth flow, the session state data did not match because it did not exist!

However, this PR gave us an opportunity to rethink the routes as well, and within the OAuth flow established, it makes more sense to have a single route for both login and register, so we've renamed it to `/authorize` which overall just makes a bit more sense.

Additionally, during this process that was full of surprises, we also found out that the defined route redirects in Laravel prior to 5.7 would set redirects as `301` which are permanently cached in a user's browser and nigh impossible to clear. As of Laravel 5.7 this has been updated so redirects default to `302`.

Moving forward we're changing the defined redirects in the routes file to `302`, and being very prudent as to which should be `301` 😬 

### Checklist

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -45,8 +45,14 @@ class AuthController extends Controller
     {
         $options = array_get($queryParams, 'options', []);
 
+        $mode = array_get($queryParams, 'mode', null);
+
         if (is_string($options)) {
             $options = (array) json_decode($options);
+        }
+
+        if ($mode) {
+            $options['mode'] = $mode;
         }
 
         return $options;
@@ -109,31 +115,6 @@ class AuthController extends Controller
     }
 
     /**
-     * Handle a login request to the application.
-     *
-     * @param ServerRequestInterface $request
-     * @param ResponseInterface      $response
-     * @return \Illuminate\Http\RedirectResponse
-     */
-    public function getLogin(ServerRequestInterface $request, ResponseInterface $response)
-    {
-        $queryParams = $request->getQueryParams();
-
-        $this->setSessionData($queryParams);
-
-        $url = $this->getUrl();
-
-        $options = $this->getOptions($queryParams);
-
-        $destination = $this->getDestination($queryParams, $options);
-
-        // Override the default Northstar authorization /register page and view /login instead.
-        $options['mode'] = 'login';
-
-        return gateway('northstar')->authorize($request, $response, $url, $destination, $options);
-    }
-
-    /**
      * Handle a logout request to the application.
      *
      * @param ResponseInterface $response
@@ -145,13 +126,13 @@ class AuthController extends Controller
     }
 
     /**
-     * Handle a registration request to the application.
+     * Handle an auth request to the application.
      *
      * @param  ServerRequestInterface $request
      * @param  ResponseInterface      $response
      * @return \Illuminate\Http\RedirectResponse
      */
-    public function getRegistration(ServerRequestInterface $request, ResponseInterface $response)
+    public function getAuthorization(ServerRequestInterface $request, ResponseInterface $response)
     {
         $queryParams = $request->getQueryParams();
 
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -23,6 +23,48 @@ class AuthController extends Controller
     protected $redirectAfterLogout;
 
     /**
+     * Make a new AuthController, inject dependencies,
+     * and set middleware for this controller's methods.
+     */
+    public function __construct()
+    {
+        $this->redirectAfterLogout = config('app.url');
+    }
+
+    /**
+     * Handle a logout request to the application.
+     *
+     * @param ResponseInterface $response
+     * @return ResponseInterface
+     */
+    public function getLogout(ResponseInterface $response)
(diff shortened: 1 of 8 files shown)


config/services.php
```diff
@@ -40,7 +40,7 @@
             'client_id' => env('NORTHSTAR_AUTHORIZATION_ID'),
             'client_secret' => env('NORTHSTAR_AUTHORIZATION_SECRET'),
             'scope' => ['user', 'activity', 'write', 'openid', 'role:staff', 'role:admin'],
-            'redirect_uri' => 'next/login',
+            'redirect_uri' => 'authorize',
```
weerd: @DFurnes should this include the `/`? `/authorize`?

DFurnes: This one _shoooouldn't_ matter because it's being passed through Laravel's `url()` helper IIRC, but it also shouldn't hurt and kinda feels nicer. Want to give it a try?

app/Http/Controllers/AuthController.php
```diff
@@ -108,31 +114,6 @@ public function __construct()
         $this->redirectAfterLogout = config('app.url');
```
DFurnes: I know this isn't new in this pull request, but would you mind moving the constructor up to the top of this class? So it reads properties, constructor, public functions, protected functions?

resources/views/partials/navigation.blade.php
```diff
@@ -39,8 +39,8 @@
                         <li><a href="{{ route('logout') }}" class="secondary-nav-item" id="link--logout">Log Out</a></li>
                     </ul>
                 @else
-                    {{-- <a href="{{ route('login', get_login_query($entity)) }}">Log In</a> --}}
-                    <a href="{{ route('register', get_login_query($entity)) }}">Log In</a>
+                    {{-- <a href="{{ route('authorize', array_merge(['mode' => 'login'], get_login_query($entity))) }}">Log In</a> --}}
```
DFurnes: Assuming that you're keeping this around for the separated links in a follow-up PR? If so, I wonder if it makes sense to handle this `array_merge()` business inside this helper function (and maybe rename to `get_authorization_query()` while we're at it?) Something like:

```blade
<a href="{{ route('authorize', get_authorization_query($entity, 'login')) }}">Log In</a>
```

app/Http/Controllers/AuthController.php
weerd: Sounds good. Always assumed public funcs go at the bottom, but not a thing I read anywhere, just an assumption, haha.

resources/views/partials/navigation.blade.php
weerd: ya, I dig that 👍

app/Http/Controllers/AuthController.php
DFurnes: Yeah, I'm not sure if there's actually a suggested order in PSR or not... my gut says that it's nice to start with the public functions since that makes it easy to skim what a class can do without wading through the internals.

app/Http/Middleware/StartSession.php
```diff
@@ -12,7 +12,7 @@ class StartSession extends Middleware
      * @var array
      */
     protected $mustPersistSession = [
-        'next/login',
+        'authorize',
```
mendelB: 🕵️‍♂️AHA!

routes/web.php
```diff
@@ -12,11 +12,11 @@
 $router->get('/us', 'HomePageController');
 
 // Authentication
-$router->get('us/register', 'AuthController@getRegistration')->name('register');
-$router->get('next/login', 'AuthController@getLogin')->name('login');
+$router->get('/authorize', 'AuthController@getAuthorization')->name('authorize');
+$router->redirect('/auth/login', '/next/login', 302); // Fix for hard-coded redirect in Gateway! <goo.gl/2VPxDC>
```
mendelB: Wouldn't these indeed be permanent `301`s though, or is the idea to play it safe in case we bring back this route?

weerd: Yeah, we want to play it safe until we define some better rules around when exactly we want to use `301`. This won't change the existing `301`s on user's browsers, but any new ones would be updated. 

We also need to see if there's any sort of way to bust `301` caches in some fashion... otherwise we might have problems in the future, because as of right now, there is no easy way to repurpose `www.dosomething.org/vote` for future 2020 use 😬

config/services.php
weerd: Initially it seemed to give me an issue, but might have been a fluke from not having refreshed page or something. Did a bunch of testing and adding the `/` didn't seem to affect functionality 😄

routes/web.php
mendelB: Oof yeah I think that SO answer had some ideas for cache busting which we can play around with

weerd: Yep yep. I linked it over to @DFurnes to add to a Platforms ticket regarding addressing `301`s.

PR #1575, opened 2019-08-28
Final Beta Referral Page styles

aaronschachter:
### What does this PR do?

This PR adds an image to the Beta Referral Page to finish styling, and adds desired copy changes.

The one place I'm stuck is referencing the new image when setting the Beta Page metadata. Using `asset('/resources/assets/components/pages/ReferralPage/Beta/gift-card.svg')` results in a 404 ... from what I can tell it looks like we'd need to add the image into `/storage/app/public`, but there's nothing in there currently, so not sure about that.

Screenshots:

* [Desktop](https://user-images.githubusercontent.com/1236811/63880880-6ee65b80-c983-11e9-8505-95f866d1bc88.png)
* [Tablet](https://user-images.githubusercontent.com/1236811/63880886-73127900-c983-11e9-8c43-06de55b4865f.png)
* [Mobile](https://user-images.githubusercontent.com/1236811/63880913-845b8580-c983-11e9-99b0-1449074e6241.png)

Thinking ahead to the last much-larger chunk of styling work ahead: there will be [2 new additional gift card background images (1 for mobile, 1 for desktop)](https://www.pivotaltracker.com/n/projects/2019429/stories/167633963/comments/206082314) that we'll want to add for the Affirmation Refer Friends link introduced in #1551 . As is, those images seem like they could live within the `Affirmation` directory, but would it make sense to move the Refer Friends div and images into a new component in `resources/assets/components/blocks/ReferFriendsBlock`?

https://www.pivotaltracker.com/story/show/167633963

### Checklist


mendelB:
> from what I can tell it looks like we'd need to add the image into /storage/app/public

➕I believe this is correct re [the docs](https://laravel.com/docs/5.5/filesystem#the-public-disk)!

> would it make sense to move the Refer Friends div and images into a new component in resources/assets/components/blocks/ReferFriendsBlock?

Is this more of a call to action type block? Maybe like `CtaReferFriends` or something? Also If we're using it only within other components like Affirmation, might it be a `/utilities` type thing?

aaronschachter:
Thanks for reviewing @mendelB ! I have some followup questions to your comments:

Re: adding the gift card image, I will add this in a separate PR and merge this while approved. I wasn't sure about whether I need to create the symbolic link that [the docs mention](https://laravel.com/docs/5.5/filesystem#the-public-disk), but I'll look into it know that I know that's the path I should go down.

> Is this more of a call to action type block? Maybe like CtaReferFriends or something? Also If we're using it only within other components like Affirmation, might it be a /utilities type thing?

I think that's probably a yes to both -- it's a call to action to visit the alpha referral page, and will only ever be used within other components -- but for now, it's only used within the Affirmation component. I may just add the styles/images into the `Affirmation` directory to keep things super basic, but we can see how it goes in the next PR!

mendelB:
> Re: adding the gift card image, I will add this in a separate PR and merge this while approved. I wasn't sure about whether I need to create the symbolic link that the docs mention, but I'll look into it know that I know that's the path I should go down.

@aaronschachter I'm thinking now that I may have misspoken and we'd perhaps just want to put this image directly in the `/public` directory, or a new `/public/images` directory! Since we'd be fine with this image being publically accessible, and there is some precedent of us storing files/images there. 

@weerd any intel on this?

aaronschachter:
Ah, I can maybe try that. I keep getting the 404 when trying to visit the image path, and figured it had something to do with our fallback route. Was looking into something like [this SO thread](https://stackoverflow.com/questions/15836443/laravel-how-to-route-to-public-file) about adding a route for public files, which seems to make this harder than it should be...

mendelB:
>  I keep getting the 404 when trying to visit the image path

Are you adding `/public` to the path? I think it needs to be from app_url/file-name directly without `/public`. Kinda like how we reference analytics event schema [here](https://github.com/DoSomething/phoenix-next/blob/5fce41fa20f5f6f662fe86a5e4cf95d90e7aa0be/resources/assets/helpers/analytics.js#L104)

aaronschachter:
thanks, got it working!

PR #1577, opened 2019-08-29
Adds Referral Page metadata image

aaronschachter:
### What does this PR do?

This PR adds a `public/images` directory to house the gift card image to be used for Referral Page metadata.

This approach was discussed in #1575. I'll add the review app that this PR creates to spot check the image appears correctly in the Alpha page that embeds the Beta page.

https://www.pivotaltracker.com/story/show/167633963

### Checklist


aaronschachter:
Added review app to the `phoenix` dev client to confirm the image shows up in https://dosomething-phoenix-de-pr-1577.herokuapp.com/us/refer-friends:

alpha-complete

app/Http/Controllers/ReferralPageController.php
```diff
@@ -37,11 +37,10 @@ public function show(Request $request)
               'title' => $title,
               'description' => $callToAction,
               'facebook_app_id' =>  config('services.analytics.facebook_id'),
-              // @TODO: Replace with envelope graphic once added to beta template.
               'image' => [
-                'url' => 'https://forge.dosomething.org/resources/ds-logo-landscape.png',
-                'width' => '1200',
-                'height' => '630',
+                'url' => url('/').'/images/gift-card.svg',
+                'width' => '685',
```
mendelB: Is there a specific reason to deviate from the standards [in the helper](https://github.com/DoSomething/phoenix-next/blob/a0b55033ecfd1f1b8e7729d6295447ffac6178d7/app/helpers.php#L409-L413)? I _think_ the 1200x1200 recommendation is so it works well in Facebook/Twitter/Social embed cards

aaronschachter: I went with image dimensions, but updated to the 1200x1200.

app/Http/Controllers/ReferralPageController.php
```diff
@@ -37,11 +37,10 @@ public function show(Request $request)
               'title' => $title,
               'description' => $callToAction,
               'facebook_app_id' =>  config('services.analytics.facebook_id'),
-              // @TODO: Replace with envelope graphic once added to beta template.
               'image' => [
-                'url' => 'https://forge.dosomething.org/resources/ds-logo-landscape.png',
+                'url' => url('/').'/images/gift-card.svg',
```
mendelB: Ah shoot sorry -- one other thing which I'm not sure about but just caught my eye -- should we use `url('app.url')` here to get the base URL like in the [helper](https://github.com/DoSomething/phoenix-next/blob/a0b55033ecfd1f1b8e7729d6295447ffac6178d7/app/helpers.php#L392)?

aaronschachter: Sure? I can make this change, though this seemed to work fine.

mendelB: 🤷‍♂ lol I honestly dunno, if you don't think it's worth the change then probs better not

mendelB: General question cause now I'm curious: what's better practice -- relative or absolute URL's? I feel like the main argument I've seen for relative is so that it works across environments, but if we're using an ENV variable that's set per environment, is there any advantage in the absolute definition? 🤔 

(Sorry, I don't mean to derail this straightforward PR with my ignorance 🙃)

weerd: Ya, please use the `config('app.url')`!

weerd: I actually think we may want to be using `asset()` here instead.

weerd: `asset('images/gift-card.svg')` === `http://phoenix.test/images/gift-card.svg`

PR #1578, opened 2019-08-29
Update Authorization Exception Handler To Use New Authorize Route For Authentication

mendelB:
### What does this PR do?

This PR updates our `unauthenticated` exception handler - which functions to redirect an unauthed user to authorize with Northstar instead of serving an error - to use the new `authorize` endpoint established in #1552 

[Here's](https://dosomething.slack.com/archives/C3ASB4204/p1567121435041900) some Slack messages for context. 

But TL;DR: Any gated request by an un-authed user will result in a `500` since we'll attempt to serve the `login` route to prompt them to authenticate which no longer exists!

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -94,7 +94,7 @@ class Handler extends ExceptionHandler
             return response()->json(['error' => 'Unauthenticated.'], 401);
         }
 
-        return redirect()->guest(route('login'));
+        return redirect()->guest(route('authorize'));
     }
 
     /**


app/Exceptions/Handler.php
```diff
@@ -94,7 +94,7 @@ protected function unauthenticated($request, AuthenticationException $exception)
             return response()->json(['error' => 'Unauthenticated.'], 401);
         }
 
-        return redirect()->guest(route('login'));
+        return redirect()->guest(route('authorize'));
```
weerd: gah! Missed this one! Good catch!

PR #1579, opened 2019-08-30
Remove Register from CTA Register Banner 

mendelB:
### What does this PR do?

This PR abstracts the `CtaRegisterBanner` component to be a `CtaBanner`, agnostic about its specific link and copy by using `props`.

This refactor is per the discussion here https://github.com/DoSomething/phoenix-next/pull/1554#discussion_r319125407 https://github.com/DoSomething/phoenix-next/pull/1548#discussion_r318261969

If this seems ok, I'll apply this to the `CtaRegisterPopover` component as well!

Refs [Pivotal ID #167881877](https://www.pivotaltracker.com/story/show/167881877)

### Checklist


resources/assets/components/utilities/CtaBanner/CtaBanner.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './cta-banner.scss';
+
+const CtaBanner = ({ buttonText, content, link, title }) => (
```
weerd: Noice! Defs dig this better. We could eventually even have a prop for the background pattern/color to use. vNext ;)

resources/assets/components/utilities/CtaBanner/CtaBanner.js
```diff
@@ -0,0 +1,29 @@
...
+  <div className="cta-banner base-12-grid">
+    <div className="grid-narrow margin-md">
```
weerd: I wonder if we should move away from our specific `margin-md` to tailwind style classes? Maybe make the class in the bottom tailwind section of `base.scss` if not there already, so when we switch there's less components to edit?

resources/assets/components/utilities/CtaBanner/CtaBanner.js
```diff
@@ -0,0 +1,29 @@
...
+      <h3 className="text-m color-yellow font-bold uppercase">{title}</h3>
+      <p className="text-white mt-4">{content}</p>
+      <a
+        className="cta-banner__button button padded mt-4"
```
weerd: Same same with `padded`. Also not crucial if you want to merge and maybe do the class switches in upcoming PR.

resources/assets/components/utilities/CtaBanner/CtaBanner.js
mendelB: > We could eventually even have a prop for the background pattern/color to use. vNext ;)

💡

resources/assets/components/utilities/CtaBanner/CtaBanner.js
mendelB: Yeah that's a great idea! I'll tackle that and  https://github.com/DoSomething/phoenix-next/pull/1579#discussion_r319564005 in a follow-up PR so I can kill both components with one stone

PR #1580, opened 2019-08-30
Navigation event analytics

weerd:
### What does this PR do?

This PR adds event analytics for links and the search bar in the main Navigation. It also adds some updates/cleanup within the `helpers/analytics.js` to help DRY some things up.

Example Google Event:
image

Example Snowplow Event:
image

Refs [Pivotal ID #168175576](https://www.pivotaltracker.com/story/show/168175576)

### Checklist

### DIFF
--- a/resources/assets/helpers/navigation.js
+++ b/resources/assets/helpers/navigation.js
@@ -1,4 +1,4 @@
-/* global window, document */
+/* global document */
 
 import { snakeCase } from 'lodash';
 
@@ -23,11 +23,12 @@ export function linkAnalyticsHandler(link) {
           ...getUtmContext(),
           ...getPageContext(),
           referrer: document.referrer,
-          pathname: window.location.pathname,
         },
         metadata: {
+          adjective: snakeCase(event.target.textContent),
           category: 'navigation',
-          noun: snakeCase(event.target.textContent),
+          label: snakeCase(event.target.textContent),
+          noun: 'nav_link',
           target: 'link',
           verb: 'clicked',
         },
--- a/resources/assets/helpers/navigation.js
+++ b/resources/assets/helpers/navigation.js
@@ -1,4 +1,4 @@
-/* global document */
+/* global window, document */
 
 import { snakeCase } from 'lodash';
 
@@ -23,7 +23,7 @@ export function linkAnalyticsHandler(link) {
           ...getUtmContext(),
           ...getPageContext(),
           referrer: document.referrer,
-          pathname: document.pathname,
+          pathname: window.location.pathname,
         },
         metadata: {
           category: 'navigation',
--- a/resources/assets/helpers/navigation.js
+++ b/resources/assets/helpers/navigation.js
@@ -1,48 +1,84 @@
 /* global document */
 
-import { findKey } from 'lodash';
+import { snakeCase } from 'lodash';
 
+import {
+  getPageContext,
+  getUtmContext,
+  trackAnalyticsEvent,
+} from './analytics';
 import { toggleClassHandler } from '.';
-import { trackAnalyticsEvent } from './analytics';
-
-export const campaignPaths = {
-  community: '/community',
-  action: '/action',
-  blocks: '/blocks/',
-  quiz: '/quiz/',
-};
 
 /**
- * Get the display name of the given route,
+ * Handle collecting Navigation link data and triggering analytics event.
  *
- * @param  {String} route
- * @return {String}
+ * @param {Object} link
  */
-export function getRouteName(route) {
-  // When doing path comparisons, we want the least specific
-  // (eg: '/') paths at the end of the array.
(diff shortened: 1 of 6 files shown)


resources/assets/helpers/navigation.js
```diff
@@ -1,79 +1,114 @@
...
 
 /**
- * Get the display name of the given route,
+ * Handle collecting Navigation link data and triggering analytics event.
  *
- * @param  {String} route
- * @return {String}
+ * @param {Object} link
  */
-export function getRouteName(route) {
```
weerd: Did some extensive searching and this method was no longer in use so I removed it!

resources/assets/helpers/navigation.js
```diff
@@ -1,79 +1,114 @@
 /* global document */
 
-import { findKey } from 'lodash';
+import { snakeCase } from 'lodash';
 
-export const campaignPaths = {
```
weerd: No longer needed since `getRouteName()` method was removed!

resources/assets/helpers/navigation.js
```diff
@@ -1,79 +1,114 @@
...
-    (pathA, pathB) => pathB.length - pathA.length,
+export function linkAnalyticsHandler(link) {
+  link.addEventListener(
+    'mousedown',
+    event => {
+      trackAnalyticsEvent({
+        context: {
+          ...getUtmContext(),
+          ...getPageContext(),
+          referrer: document.referrer,
```
mendelB: Do `document.referrer` and `document.pathname` not get tracked automatically by Snowplow?

resources/assets/helpers/navigation.js
```diff
@@ -1,79 +1,114 @@
...
+ * @param {Object} link
  */
-export function getRouteName(route) {
-  // When doing path comparisons, we want the least specific
-  // (eg: '/') paths at the end of the array.
-  const pathValues = Object.values(campaignPaths).sort(
-    (pathA, pathB) => pathB.length - pathA.length,
+export function linkAnalyticsHandler(link) {
+  link.addEventListener(
+    'mousedown',
```
mendelB: Would the `'mousedown'` event be triggered if the nav link was clicked via the Tab highlighter and clicking Enter? (I'm thinking like an accessibility browser or something). Or would `'click'` perhaps account for that as well?

resources/assets/helpers/navigation.js
weerd: They might, but not sure w/ Google. I'm running this stuff by the Data Team to see if it's useful or not.

resources/assets/helpers/navigation.js
weerd: It was using `mousedown` before, which is why I stuck with it... I had initially assumed we'd be using `click` but deferred to what was there already 🤷‍♂ But defs a good note to investigate.

weerd: I believe the `mousedown` is potentially used to help interaction be more responsive on mobile, since it triggers when a button is pressed but before it is released, whereas `click` will only fire once it is released.

resources/assets/helpers/navigation.js
```diff
@@ -1,79 +1,114 @@
...
+      trackAnalyticsEvent({
+        context: {
+          ...getUtmContext(),
+          ...getPageContext(),
+          referrer: document.referrer,
+          pathname: document.pathname,
+        },
+        metadata: {
+          category: 'navigation',
+          noun: snakeCase(event.target.textContent),
```
mendelB: @weerd Are you thinking of the link text itself as the `noun` as opposed to it being the `adjective`? I feel like the `noun` could also be something like `link` or `nav_link` in this case, no?

resources/assets/helpers/navigation.js
weerd: Yeah, I did some testing and noticed that too. While the link does go to where it's intending to go, the JS doesn't trigger and nothing is tracked. I can make a ticket to consider changing it, but not a call I want to make right now since outside the scope of this PR.

mendelB: Gotcha, that sounds reasonable!

resources/assets/helpers/navigation.js
weerd: Yeah, I can try that!

weerd: Works better, but also need to include a `label` property otherwise the label pulls from the `noun` and it's too basic to work well as a label :)

resources/assets/helpers/navigation.js
weerd: Ticket made! https://www.pivotaltracker.com/story/show/168197550

resources/assets/helpers/navigation.js
weerd: Got input from Data Team, leaving in `referrer` but got rid of `pathname` :)

PR #1581, opened 2019-08-30
Embed block refactor - Update File locations

lindsayrmaher:
### What does this PR do?

This PR creates an EmbedBlock directory that will house both the Carto and TypeForm Embed Templates and a 'dispatcher' to help distinguish which template editors will be using.

First of a few PRs for Quiz updates, but trying to utilize already created code as much as possible.

### Checklist


resources/assets/components/Campaign/Campaign.js
```diff
@@ -8,8 +8,8 @@ import Modal from '../utilities/Modal/Modal';
 import NotificationContainer from '../Notification';
 import ModalRoute from '../utilities/ModalRoute/ModalRoute';
 import DelayedElement from '../utilities/DelayedElement/DelayedElement';
-import TypeFormSurvey from '../utilities/TypeFormSurvey/TypeFormSurvey';
 import CampaignRouteContainer from './CampaignRoute/CampaignRouteContainer';
+import TypeFormTemplate from '../blocks/EmbedBlock/templates/TypeFormTemplate';
```
weerd: So I think since here the Typeform is being used independently of being in a Contentful block, it doesn't make sense to be just grabbing the template. The idea with the templates should be that they only get called within the parent component they reside in.

So only the `EmbedBlock` would call a template when it's being rendered.

The templates themselves can pull in the `TypeFormSurvey` component, but in this scenario the TypeFormSurvey is a standalone utility and should stay that way.

mendelB: Maybe we can rename to `TypeFormEmbed` instead of `TypeFormSurvey` since it's technically just embedding any specified TypeForm form?

lindsayrmaher: @weerd @mendelB I like the idea of merging your two suggestions. so, renaming to `TypeFormEmbed`, but moving it back into the utilities folder and importing into the `TypeFormTemplate` in `EmbedBlock`. Does that make sense?

weerd: I think what we want is a structure like this:

```js
/components/blocks
  |__> /EmbedBlock
        |__> /templates
              |__> TypeFormTemplate.js // this uses the TypeFormSurvey utility component
              |__> CartoTemplate.js // this uses the Iframe utility component
...
/components/utilities
  |__> /TypeFormSurvey/TypeFormSurvey.js
  |__> /Iframe/Iframe.js
```

weerd: I dig the renaming from `TypeFormSurvey` to `TypeFormEmbed` 👍 Maybe also rename `Iframe` to `IframeEmbed` to follow a similar pattern? Also pop that one into it's own directory instead of a standalone file that is a child of `/utilities`.

lindsayrmaher: 💯 ok i like this! will do.

resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
```diff
@@ -7,7 +7,7 @@ import { Query } from 'react-apollo';
 
 import { env } from '../../../helpers';
 import ContentfulEntry from '../../ContentfulEntry';
-import { EmbedBlockFragment } from '../Iframe';
+import { EmbedBlockFragment } from '../IframeEmbed/IframeEmbed';
```
weerd: Lower this line since it just became longer than the one after it :)

resources/assets/components/pages/HomePage/HomePage.js
```diff
@@ -7,7 +7,7 @@ import sponsorList from './sponsor-list';
 import Modal from '../../utilities/Modal/Modal';
 import { contentfulImageUrl, env } from '../../../helpers';
 import DelayedElement from '../../utilities/DelayedElement/DelayedElement';
-import TypeFormSurvey from '../../utilities/TypeFormSurvey/TypeFormSurvey';
+import TypeFormEmbed from '../../utilities/TypeFormEmbed/TypeFormEmbed';
```
weerd: Move this line up since it's shorter than the one before it now. :)

resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -3,17 +3,17 @@
...
 import NotFound from '../NotFound';
-import Iframe from '../utilities/Iframe';
 import Loader from '../utilities/Loader';
-import AffirmationContainer from '../Affirmation/AffirmationContainer';
 import StaticBlock from '../StaticBlock';
 import { ContentfulEntryJson } from '../../types';
 import PollLocator from '../PollLocator/PollLocator';
 import ErrorBlock from '../blocks/ErrorBlock/ErrorBlock';
 import ImagesBlock from '../blocks/ImagesBlock/ImagesBlock';
+import IframeEmbed from '../utilities/IframeEmbed/IframeEmbed';
```
weerd: Small note, when import line lengths are the same length, then we go in alphabetical order, so pop this one under `GalleryBlock`!

resources/assets/components/Campaign/Campaign.js
```diff
@@ -8,8 +8,8 @@ import Modal from '../utilities/Modal/Modal';
 import NotificationContainer from '../Notification';
 import ModalRoute from '../utilities/ModalRoute/ModalRoute';
 import DelayedElement from '../utilities/DelayedElement/DelayedElement';
-import TypeFormSurvey from '../utilities/TypeFormSurvey/TypeFormSurvey';
 import CampaignRouteContainer from './CampaignRoute/CampaignRouteContainer';
+import TypeFormEmbed from '../utilities/TypeFormEmbed/TypeFormEmbed';
```
weerd: Same on this line, move it up since now shorter length.

weerd:
Are `EmbedBlock.js`, `CartoTemplate.js` and `TypeFormTemplate.js` empty files? Maybe stub something out in them for now. Not ideal to just add empty files. Even if a basic stubbed out component to get an idea of intent is fine. Can have `@TODO`s in them or whatever... or hold off on adding them until you build out those components :)

PR #1583, opened 2019-08-30
Starts on CtaReferralPageBanner

aaronschachter:
### What does this PR do?

This PR refactors the minimalist `Affirmation` link added in #1551 as a `CtaReferralPageBanner`, and (heaven, help us) starts on styling it per the [desktop design](https://projects.invisionapp.com/share/GYSBOGRSDUA#/screens/372969109). This is not ready for production, but opening for review now to make sure this is looking 👍 

This can be tested on any campaign action page by appending a `refer-friends=true` query string to the URL, and viewing the affirmation. I will post a link to the review app (and update dev admin with the review app URL) so it's easily visible.

* Will need to add a different styles for [mobile screens](https://projects.invisionapp.com/share/QCSXLUH4NGJ#/screens/368867492): different background image, centered text

* I set the hardcoded widths by measuring the [desktop design](https://projects.invisionapp.com/share/GYSBOGRSDUA#/screens/372969109) with a Chrome pixel ruler plugin. Feels a bit like a hack but also kind of looks ok? I don't get a sense of what the banner should do on wider screens, so guessed there too.

* Per https://github.com/dosomething/phoenix-next/pull/1551#issuecomment-523627655, I referenced the `CtaRegisterPopover` component as a guide (which has now been refactored/renamed in pull requests from yesterday)

Refs https://www.pivotaltracker.com/n/projects/2019429/stories/167633963

### Checklist

Note -- not including screenshots as this is not ready for primetime since we need separate CSS for desktop/mobile.


aaronschachter:
Added this review app to admin-dev for a working demo: https://dosomething-phoenix-de-pr-1583.herokuapp.com/us/campaigns/test-teens-for-jeans/action?refer-friends=true (must be logged in)

resources/assets/components/utilities/CtaReferralPageBanner/cta-referral-page-banner.scss
```diff
@@ -0,0 +1,23 @@
...
+
+.cta-register-banner {
+  background: url('./cta-referral-page-banner-desktop.svg');
+  background-size: cover;
+  height: 270px;
+}
+
+.cta-register-banner h3,
+p {
+  color: white;
```
aaronschachter: This is 🙅‍♂ - this style is carrying over to the Alpha and Beta Referral Pages (I've highlighted it to demonstrate)

DFurnes: This is a bit of a "gotcha": these two selectors are actually completely separate `.cta-register-banner h3` and `p`. And because CSS is a global namespace, that second `p` rule just gets applied to every paragraph! 😨

A quick fix would be prefixing the same `.cta-register-banner` class to the `p` selector, but it might also be a good idea to add a specific named class to these elements, or use [our `.color-white` helper class](https://github.com/DoSomething/phoenix-next/blob/ac0ec68e4de15ab4aa5e1e561e03fc8ce6d084ec/resources/assets/scss/base.scss#L86-L88) instead.

resources/assets/components/utilities/CtaReferralPageBanner/cta-referral-page-banner.scss
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.cta-register-banner .button {
+  text-transform: uppercase;
```
DFurnes: This is subjective, but you could use our [`.uppercase`](https://github.com/DoSomething/phoenix-next/blob/ac0ec68e4de15ab4aa5e1e561e03fc8ce6d084ec/resources/assets/scss/base.scss#L681-L683) helper class to avoid having to write extra CSS for this!

resources/assets/components/utilities/CtaReferralPageBanner/CtaReferralPageBanner.js
```diff
@@ -0,0 +1,44 @@
...
+    return (
+      <div className="padding-top-md padding-bottom-md padding-horizontal-md">
+        <div className="cta-register-banner padding-top-md padding-bottom-md padding-horizontal-md">
+          <div>
+            <h3>Benefits With Friends</h3>
+            <p className="padding-bottom-md">
+              Refer a friend to this campaign, and you’ll *both* earn a $5 gift
+              card!
+            </p>
+            <button
```
mendelB: @aaronschachter I think if you use an `<a>` tag here instead of a button, (retaining the class names so it gets styled as a button), you could just add the `href` directly as an attribute, which would allow removing the `redirectToReferralPage` helper and turning this component into a functional component!

(I ended up replacing the <button> with an <a> tag in the cta components I was working on as well)

mendelB: ...oh unless you did this because of potential analytics events 🤔

resources/assets/components/utilities/CtaReferralPageBanner/CtaReferralPageBanner.js
```diff
@@ -0,0 +1,44 @@
...
+    this.redirectToReferralPage = this.redirectToReferralPage.bind(this);
+  }
+
+  redirectToReferralPage() {
+    window.location.href = `/us/refer-friends?campaign_id=${this.props.campaignId}`;
+  }
+
+  render() {
+    return (
+      <div className="padding-top-md padding-bottom-md padding-horizontal-md">
```
mendelB: I think you could get away with replacing these classes with the single [`.p-4` utility class](https://github.com/DoSomething/phoenix-next/blob/9e2d2176fa26a098730a69218bd846513ce65a65/resources/assets/scss/base.scss#L567-L569)

resources/assets/components/utilities/CtaReferralPageBanner/CtaReferralPageBanner.js
aaronschachter: Ah, that'd be way cleaner! Per the [analytics spreadsheet](https://docs.google.com/spreadsheets/d/1hNZdkcS7orp7xU0mOHSRKPodNoif0EP5UC2YwOnDmPI/edit#gid=1098747526), it doesn't look like we need to track visits to the Alpha page (which is where this link is directing)

resources/assets/components/utilities/CtaReferralPageBanner/cta-referral-page-banner.scss
aaronschachter: Ah, thanks! 🤦‍♂ Amateur hour 🍸

resources/assets/components/utilities/CtaReferralPageBanner/cta-referral-page-banner.scss
aaronschachter: Actually it looks like the `share -blue` classes I added later after finding them adds the uppercase, so I think this can be removed entirely.

aaronschachter:
Thank you for the helpful reviews @DFurnes and @mendelB. I've made all changes and tested out the review app, https://dosomething-phoenix-de-pr-1583.herokuapp.com/us/campaigns/test-teens-for-jeans/action?refer-friends=true looks pretty decent to my 👀 . Next up in a future PR will be conditionally styling background/alignment per mobile.

resources/assets/components/utilities/CtaReferralPageBanner/CtaReferralPageBanner.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './cta-referral-page-banner.scss';
+
+const CtaReferralPageBanner = props => (
+  <div className="p-4">
+    <div className="cta-register-banner p-4">
+      <div>
+        <h3 className="color-white">Benefits With Friends</h3>
```
mendelB: In the spirit of trying to get into the habit of using more Tailwind CSS-esque classes where we can instead of the legacy utility ones we've defined -- could we replace `color-white` with it's [`text-white`](https://github.com/DoSomething/phoenix-next/blob/9e2d2176fa26a098730a69218bd846513ce65a65/resources/assets/scss/base.scss#L640-L642) replacement?

PR #1584, opened 2019-09-03
Added Direct Link to NorthStar's Edit Page on Phoenix Profile Page

Vmack:
# What does this PR do?

This PR directly links north star's edit page to the 'edit profile' button on the Phoenix profile page 

This new link will remove the additional prompting for editing the user's profile page and render a more smoother page transition  

Refs [Pivotal ID #167742718](https://www.pivotaltracker.com/story/show/167742718)

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Profile.js
+++ b/resources/assets/components/pages/AccountPage/Profile.js
@@ -24,7 +24,7 @@ const Profile = props => (
       <FormItem title="Phone Number" value={props.user.mobile} />
       <div className="margin-top-lg">
         <a
-          href={`${env('NORTHSTAR_URL')}/users/${props.userId}/edit`}
+          href={`${env('NORTHSTAR_URL')}/users/${props.user.id}/edit`}
           target="_blank"
           rel="noopener noreferrer"
           className="button"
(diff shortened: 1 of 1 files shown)


resources/assets/components/pages/AccountPage/Profile.js
```diff
@@ -26,7 +24,7 @@ const Profile = props => (
       <FormItem title="Phone Number" value={props.user.mobile} />
       <div className="margin-top-lg">
         <a
-          href={env('NORTHSTAR_URL') + login}
+          href={`${env('NORTHSTAR_URL')}/users/${props.user.id}/edit`}
```
katiecrane: Were you not able to use `props.userId`? That's the one we saw showing up in dev tools before.

katiecrane:
I added this review app to the Redirect URIs in Aurora so we're able to log in on the review app and test it out! It's working WONDERFULLY!!! Added one comment about which prop you're using though 😃

resources/assets/components/pages/AccountPage/Profile.js
Vmack: Yes! So initially the plan was to use the userId prop, however, I was prompted to create a propType for userId which existed in AccountRoute.js but not in Profile.js. So we figured by just adding the id propType to user, it would make Profile.propTypes look cleaner but I can switch it back.

katiecrane: Oh I see! It had the user id in two places! I like using the one under the user too, like you have it here

PR #1585, opened 2019-09-03
Rendering CTA's on Articles & Facts pages (Feature Flagged)

mendelB:
### What does this PR do?

This PR renders the `CtaBanner` and `CtaPopover` on the `GeneralPage`, with copy based on the page category in the URL for unauthenticated users.

It also adds a 'close' or 'ex' button to the `CtaPopover` to handle user dismissal. I ended up utilizing the pre-existing styling from the [Modal component](https://github.com/DoSomething/phoenix-next/blob/fc5cdfcf5d6edea99ab07966d2ec5b0568b16941/resources/assets/components/utilities/Modal/modal.scss#L46-L62) and duplicating its rendering of the button, but it does feel like we can extract that into a utility component to DRY this up a bit? (I needed to add some modifier classes to get the right color and sizing https://github.com/DoSomething/phoenix-next/commit/f32d4a042b2512595d0ce86dc4ad3471885c5f8d). I added a TODO for the above - but lemme know if you disagree! https://github.com/DoSomething/phoenix-next/commit/475d237b5993cde41fa28b786c77388326c2b327

https://github.com/DoSomething/phoenix-next/commit/dac0d806a3d1aa4ed0eb9429d3cdb6aaf01dbf4c I also updated the HACK to snap the popover to the left of the page to prevent chrome overflow to be more generous about when it does so, since it should optimally be on the right, and it was snapping a pixel too early to begin with. 

~Finally, I feature flagged the CTA's under the `show_register_cta` query param.~ 🎩 💁‍♂ @aaronschachter

UPDATE: I added the sixpack experiment conversion for any `ctaButtonClick` tests to the CTA components in https://github.com/DoSomething/phoenix-next/pull/1585/commits/7c6d0ea6f66234e464ae67fe717fb942f8716fc1 and rendered the CTA's in a Sixpack Experiment on the `GeneralPage` in https://github.com/DoSomething/phoenix-next/pull/1585/commits/a559de459c0046c198e866d864474376db984cb2 which means this should be launch-ready once approved!

Refs [Pivotal ID #167527419](https://www.pivotaltracker.com/story/show/167527419), [Pivotal ID #167878581](https://www.pivotaltracker.com/story/show/167878581)

### Checklist

image

image

### DIFF
--- a/resources/assets/components/PageDispatcher/PageDispatcher.js
+++ b/resources/assets/components/PageDispatcher/PageDispatcher.js
@@ -25,10 +25,14 @@ const PageDispatcher = props => {
 };
 
 PageDispatcher.propTypes = {
-  authUrl: PropTypes.string.isRequired,
+  authUrl: PropTypes.string,
   isAuthenticated: PropTypes.bool.isRequired,
   type: PropTypes.string.isRequired,
   fields: PropTypes.object.isRequired,
+};
+
+PageDispatcher.defaultProps = {
+  authUrl: null,
 };
 
 export default PageDispatcher;
--- a/resources/assets/components/PageDispatcher/PageDispatcher.js
+++ b/resources/assets/components/PageDispatcher/PageDispatcher.js
@@ -14,11 +14,19 @@ const PageDispatcher = props => {
       return <StoryPage {...props.fields} />;
 
     default:
-      return <GeneralPage {...props.fields} />;
+      return (
+        <GeneralPage
+          {...props.fields}
+          authUrl={props.authUrl}
+          isAuthenticated={props.isAuthenticated}
+        />
+      );
   }
 };
 
 PageDispatcher.propTypes = {
+  authUrl: PropTypes.string.isRequired,
+  isAuthenticated: PropTypes.bool.isRequired,
   type: PropTypes.string.isRequired,
   fields: PropTypes.object.isRequired,
 };
--- a/resources/assets/components/pages/GeneralPage/GeneralPage.js
+++ b/resources/assets/components/pages/GeneralPage/GeneralPage.js
@@ -8,12 +8,21 @@ import Enclosure from '../../Enclosure';
 import LazyImage from '../../utilities/LazyImage';
 import Byline from '../../utilities/Byline/Byline';
 import ContentfulEntry from '../../ContentfulEntry';
+import { REGISTER_CTA_COPY } from '../../../constants';
 import AuthorBio from '../../utilities/Author/AuthorBio';
+import CtaBanner from '../../utilities/CtaBanner/CtaBanner';
+import CtaPopover from '../../utilities/CtaPopover/CtaPopover';
 import TextContent from '../../utilities/TextContent/TextContent';
-import { contentfulImageUrl, withoutNulls } from '../../../helpers';
+import DelayedElement from '../../utilities/DelayedElement/DelayedElement';
+import { contentfulImageUrl, withoutNulls, query } from '../../../helpers';
 import SocialShareTray from '../../utilities/SocialShareTray/SocialShareTray';
+import DismissableElement from '../../utilities/DismissableElement/DismissableElement';
 
 import './general-page.scss';
+
+// Grab the CTA copy based on the current page category.
+const pageCategory = window.location.pathname.split('/')[2];
+const ctaCopy = REGISTER_CTA_COPY[pageCategory];
 
 /**
  * Render a general page
@@ -30,6 +39,8 @@ const GeneralPage = props => {
     sidebar,
     blocks,
     displaySocialShare,
+    isAuthenticated,
+    authUrl,
   } = props;
 
   return (
@@ -125,6 +136,32 @@ const GeneralPage = props => {
           ) : null}
         </Enclosure>
       </div>
+
+      {ctaCopy && !isAuthenticated && query('show_register_cta') ? (
+        <>
+          <DismissableElement
+            testName="CTA Popover"
+            name="cta_register_popover"
+            render={handleClose => (
+              <DelayedElement delay={3}>
+                <CtaPopover
+                  title={ctaCopy.title}
+                  content={ctaCopy.content}
+                  link={authUrl}
+                  buttonText={ctaCopy.buttonText}
+                  handleClose={handleClose}
+                />
+              </DelayedElement>
+            )}
+          />
+          <CtaBanner
+            title={ctaCopy.title}
+            content={ctaCopy.content}
+            link={authUrl}
+            buttonText={ctaCopy.buttonText}
+          />
+        </>
+      ) : null}
     </div>
   );
 };
@@ -141,6 +178,8 @@ GeneralPage.propTypes = {
   sidebar: PropTypes.arrayOf(PropTypes.object),
   blocks: PropTypes.arrayOf(PropTypes.object).isRequired,
   displaySocialShare: PropTypes.bool,
+  isAuthenticated: PropTypes.bool.isRequired,
+  authUrl: PropTypes.string.isRequired,
 };
 
 GeneralPage.defaultProps = {
--- a/resources/assets/components/pages/GeneralPage/GeneralPage.js
+++ b/resources/assets/components/pages/GeneralPage/GeneralPage.js
@@ -147,7 +147,7 @@ const GeneralPage = props => {
               testName="CTA Popover"
               name="cta_register_popover"
               render={handleClose => (
-                <DelayedElement delay={0}>
+                <DelayedElement delay={3}>
                   <CtaPopover
                     title={ctaCopy.title}
                     content={ctaCopy.content}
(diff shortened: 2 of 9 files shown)


resources/assets/components/PageDispatcher/PageDispatcher.js
```diff
@@ -14,11 +14,19 @@ const PageDispatcher = props => {
...
+          {...props.fields}
+          authUrl={props.authUrl}
+          isAuthenticated={props.isAuthenticated}
+        />
+      );
   }
 };
 
 PageDispatcher.propTypes = {
+  authUrl: PropTypes.string.isRequired,
```
weerd: Debating if we really need to have the `authUrl` be required 🤔

mendelB: Hmm, do you mean like since there are other page templates that don't make use of this prop it's a bit of an overextended requirement?

weerd: Yep yep. Feels like a bit much to have to require it. But not sure if there's a situation that otherwise results in having to require it.

weerd:
Not specifically something done in this PR but playing around with the CSS I think it gets confusing that you're using both `right` and `margin-left` to designate adding spacing on the left and right sides. I would recommend just using margins for that:

image

resources/assets/components/utilities/CtaPopover/cta-popover.scss
```diff
@@ -18,9 +18,9 @@
     max-width: 400px;
   }
 
-  // Snap back to the left once the wrapper width maxes out so the
-  // popover isn't drawn past the edge of the chrome.
-  @media (min-width: 1440px) {
+  // @HACK: Snap back to the left a bit after the wrapper width maxes out
+  // so the popover isn't drawn too far past the edge of the chrome.
+  @media (min-width: 1500px) {
```
weerd: Found an alternative to this hack and works pretty precisely. Instead of approaching it like this, just change the positioning context from `fixed` to `absolute`. Then you can position relative to the parent container and behaves as needed. You can even keep it pinned to the right side, but within the parent container :)

weerd: Not 100% necessary but I think I'd recommend also adding `position: relative` to the `<div id="app" />` but maaaaybe also test a bunch of pages to make sure it doesn't break other layout things. But if you can add this, then you can also adjust the `top` value from `120px` or whatever and just keep it `0` or `12px` to give it a little margin from top of app container.

mendelB:
UPDATES:
https://github.com/DoSomething/phoenix-next/pull/1585/commits/4ec090d86000ce3bc097158e555170d72a4eea7d removes the `-small` modifier class for modal close buttons, since we're happy to have the X button be the default size based on @lkpttn feedback.

https://github.com/DoSomething/phoenix-next/pull/1585/commits/92b65799bfc554ef75a8da692aee0ca575b5a6a5 updates to render a different experiment (updating the `title` so it'll be a separate experiment) for mobile vs larger screens. This is based on feedback from Lilly being that the mobile CTA Popover experience is different -- it deserves a contained experiment.

I initially tried to use the `Media` component for a dynamicly running query as we did in the [`IFrame`](https://github.com/DoSomething/phoenix-next/blob/92b65799bfc554ef75a8da692aee0ca575b5a6a5/resources/assets/components/utilities/Iframe.js#L40-L55) but the Sixpack experiment won't actually update when the screen size changes. (This might be due to the way the `SixpackExperiment` component gets rendered). I figured we don't need to address the edge case of a user changing screen sizes, since 1. it's quite an edge case AFAIK, and 2. even if they do, I think we'd be fine with them converting on the wrong test. It'd also just be statistically insignificant.

LMK if I'm making the wrong call here though!

resources/assets/components/utilities/CtaPopover/cta-popover.scss
mendelB: I think I'm confused as to how we can achieve the scroll effect with `absolute` positioning -- wouldn't the user just be able to scroll past the element unless it's `fixed`?

mendelB:
> Not specifically something done in this PR but playing around with the CSS I think it gets confusing that you're using both `right` and `margin-left` to designate adding spacing on the left and right sides. I would recommend just using margins for that (`margin: 0 12px`) and then specifying `0` for the right positioning so that on next breakpoint it sticks to the right:
> 
> image

👌 Works brilliantly! Updated in https://github.com/DoSomething/phoenix-next/pull/1585/commits/cd44037e717e06fc92668ae96dbbadb16c9b4b09

resources/assets/components/utilities/CtaPopover/cta-popover.scss
weerd: Yep. I missed realizing that it also needs to continue to scroll, haha! But I did find an alternative that is maybe less hacky and uses one of my favorite CSS methods `calc()`.

```css
@media (min-width: 1440px) {
    right: calc((100vw - 1440px)/2);
}
```
I initially used `100%` but by using `vw` units it ensure that the calculation is performed with 100% of the viewport width and not 100% of some container.

Example with solution using calc()

resources/assets/components/pages/GeneralPage/GeneralPage.js
```diff
@@ -8,13 +8,23 @@ import Enclosure from '../../Enclosure';
...
 import { contentfulImageUrl, withoutNulls } from '../../../helpers';
+import DelayedElement from '../../utilities/DelayedElement/DelayedElement';
 import SocialShareTray from '../../utilities/SocialShareTray/SocialShareTray';
+import SixpackExperiment from '../../utilities/SixpackExperiment/SixpackExperiment';
+import DismissableElement from '../../utilities/DismissableElement/DismissableElement';
 
 import './general-page.scss';
 
+// Grab the CTA copy based on the current page category.
+const pageCategory = window.location.pathname.split('/')[2];
```
weerd: I wonder if in the future when we have more specific page types in our content models we could just refer to those? Also, would the page `slug` be helpful to use here instead? Less cruft (`/us/...`) to deal with? Not crucial for merge, just a side ponder... :)

resources/assets/components/utilities/CtaPopover/cta-popover.scss
mendelB: Straight genius! 🎩 updated in https://github.com/DoSomething/phoenix-next/pull/1585/commits/2588646f6fc51d915f380ad2303088b7abe67423

resources/assets/components/pages/GeneralPage/GeneralPage.js
mendelB: > Also, would the page slug be helpful to use here instead? Less cruft (/us/...) to deal with?

Oh hold up, yes it would be! I'll update real quick.

resources/assets/components/pages/GeneralPage/GeneralPage.js
```diff
@@ -125,6 +137,41 @@ const GeneralPage = props => {
...
+            <CtaBanner
+              testName="CTA Banner"
+              title={ctaCopy.title}
+              content={ctaCopy.content}
+              link={authUrl}
+              buttonText={ctaCopy.buttonText}
+            />,
+          ]}
+        />
+      ) : null}
```
weerd: Love seeing some Sixpack action! 💪 🍻

resources/assets/components/utilities/CtaBanner/CtaBanner.js
```diff
@@ -1,14 +1,20 @@
...
 const CtaBanner = ({ buttonText, content, link, title }) => (
   <div className="cta-banner base-12-grid">
     <div className="grid-narrow m-4">
       <h3 className="text-m text-yellow font-bold uppercase">{title}</h3>
       <p className="text-white mt-4">{content}</p>
-      <a className="cta-banner__button button p-4 mt-4" href={link}>
+      <a
+        className="cta-banner__button button p-4 mt-4"
+        href={link}
+        onClick={() => sixpack().convertOnAction('ctaButtonClick')}
```
weerd: We should make a ticket to add button clicks as convertable actions that Sixpack looks for to avoid having to pop this directly into a component. Ok for now since these components are not used elsewhere, but good that it highlights that we're starting to need more convertable actions 💃

resources/assets/components/utilities/CtaPopover/cta-popover.scss
```diff
@@ -18,10 +18,10 @@
     max-width: 400px;
   }
 
-  // Snap back to the left once the wrapper width maxes out so the
-  // popover isn't drawn past the edge of the chrome.
+  // Ensure the popover isn't drawn past the edge of the chrome
+  // once the wrapper maxes out at 1440px.
   @media (min-width: 1440px) {
-    right: inherit;
+    right: calc((100vw - 1440px) / 2);
```
weerd: 🎉

resources/assets/components/utilities/CtaBanner/CtaBanner.js
mendelB: Yeah, I wasn't completely sure how to go about formalizing this since button clicks are a very different thing then signups or posts which are more centralized and use middleware to convert the tests. 

This did end up feeling pretty good to me since as you mention we don't use these anywhere else (since the test itself will determine which one we should keep!)

I'll throw a ticket in the icebox for us to ideate a more formalized conversion!

mendelB: https://www.pivotaltracker.com/story/show/168348332

weerd: Perfect! Thanks @mendelB 👍

PR #1586, opened 2019-09-03
Refactor Carto Embed Logic using EmbedBloc

lindsayrmaher:
### What does this PR do?

This PR builds out logic to run all embed blocks through the same component (carto and typeform). Specifically, this PR addresses updating the logic for the Carto embeds.

There will be a follow up PR pending the approval here to complete adding typeform. This PR addresses existing logic and is more focused on sustaining engineering than specifically the typeform component that will be used to render the quiz.

### Checklist


resources/assets/components/blocks/EmbedBlock/EmbedBlock.js
```diff
@@ -0,0 +1,59 @@
...
+
+  switch (domainCleanUp(hostname)) {
+    case 'typeform':
+      return <TypeFormTemplate {...props} />;
+
+    case 'carto':
+      return <CartoTemplate {...props} />;
+
+    default:
+      return <ErrorBlock />;
```
mendelB: What if by default we rendered the `Embed` component? 🤔 It might set us up to eventually remove the validations on the `EmbedBlock` in Contentful and allow editors to embed things without needing to resort to the `LinkAction`! This way, if it's a recognized domain it gets the `Iframe` or specific treatment, but otherwise, it'll safely embed as a link. CC @weerd

resources/assets/components/blocks/EmbedBlock/EmbedBlock.js
```diff
@@ -0,0 +1,59 @@
...
+      description
+    }
+  }
+`;
+
+const EmbedBlock = props => {
+  const { url } = props;
+  const hostname = new URL(url).hostname;
+
+  const domainCleanUp = baseURL => {
```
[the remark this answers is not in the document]
lindsayrmaher: I like this! probably a faster lookup time in the end and still the same level of readability 🤸‍♀

resources/assets/components/blocks/EmbedBlock/templates/TypeFormTemplate.js
```diff
@@ -0,0 +1,11 @@
...
+
+import TypeFormEmbed from '../../../utilities/TypeFormEmbed/TypeFormEmbed';
+
+const TypeFormTemplate = props => {
+  return <TypeFormEmbed {...props} />;
+};
+
+export default TypeFormTemplate;
+
+// @TODO create a template to utilize the typeformembed utility function
```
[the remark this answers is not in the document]
lindsayrmaher: yah, I forgot to delete them out before committing for this PR

resources/assets/components/blocks/EmbedBlock/EmbedBlock.js
lindsayrmaher: That could be a smart alternative, would we then handle error handling individually in each of the templates? I don't know as much about how the `Embed` component is rendered logically

mendelB: I think the error here is (or was) just indicative that an unpermitted domain was added to the block (which shouldn't really happen since we only render this via Contentful which has validation).

But, since the `Embed` will just render out an embedded link (or youtube vid), it should be able to accept any valid URL! So I think we would just do away with the `ErrorBlock` here.

lindsayrmaher: @weerd What do you think here?

lindsayrmaher: @mendelB I'm going to make a ticket in the Icebox for this as a v2 update of the EmbedBlock! cc @weerd

lindsayrmaher:
@mendelB / @weerd Lemme know if this is good to go!

resources/assets/components/blocks/EmbedBlock/EmbedBlock.js
weerd: Yar, we chatted IRL. I dig the idea but to keep on task we figured best for a v2. Can likely sneak it in after the main stuff to get this all in shape for the Quiz is all set. But defs a good suggestion!

PR #1587, opened 2019-09-04
Adds mobile-specific styles for Referral Page Banner

aaronschachter:
### What does this PR do?

This PR gets the mobile view of the Referral Page Banner looking pretty prettay close to the [mobile design](https://projects.invisionapp.com/share/QCSXLUH4NGJ#/screens/368867492) -- without the "Continue to Campaign" link that I'll attempt in a future PR.

Also modifies existing CSS to to match the [desktop design](https://projects.invisionapp.com/share/GYSBOGRSDUA#/screens/372969109).

The dimensions of the [mobile background image](https://www.pivotaltracker.com/story/show/167501569/comments/206081706) are slightly off -- you'll notice the top of the graphic gets cut off. If I make the size any smaller, it doesn't stretch across the full width. I'll flag this for Leah upon deploying the review app, and can update with a different BG image in a future PR if this needs to be changed.

~~Also, I'm able to get this looking weird if I mess around with the browser width to around 674 px or so. Not sure how much of an issue this is.. or what the best fix should be?~~ **UPDATE** This was fixed by adding a background color as well as a background image

Screen Shot 2019-09-04 at 11 13 08 AM

Refs https://www.pivotaltracker.com/story/show/167633963

### Checklist

* [desktop](https://user-images.githubusercontent.com/1236811/64280075-3a374e80-cf05-11e9-92ee-12f5004cbdd1.png)
* [tablet](https://user-images.githubusercontent.com/1236811/64280078-3acfe500-cf05-11e9-98e8-28e22f47f25b.png)
* [mobile](https://user-images.githubusercontent.com/1236811/64280077-3acfe500-cf05-11e9-8d42-7e80085bf43b.png)


aaronschachter:
Here's an example on the review app: https://dosomething-phoenix-de-pr-1587.herokuapp.com/us/campaigns/escape-vape?refer-friends=true, but the CSS doesn't work as I'd expect when I'm shrinking my browser down to mobile (the mobile background image appears, but the mobile specific inner div styles are not applied) 😦. On my local, the inner div looked ok (per screenshot in checklist)

I also forgot to mention in the PR that in working on this branch, I found any time I changed the CSS -- the Referral Page Banner gift card background image would disappear (mentioned this in [Slack yesterday](https://dosomething.slack.com/archives/C3ASB4204/p1567530370004400). I'd work around this by restarting `npm`, which was a little bit 😿

aaronschachter:
The reason this looks so broken on the review app is that the CSS rules are different than my local... still trying to research why this could be happening.



On my local, the inner div is styled with the rule inside the `@include media($small) {` definition (line 15)

Screen Shot 2019-09-04 at 2 56 13 PM


But on the review app, this rule is overridden by the rule on line 27:

Screen Shot 2019-09-04 at 2 59 49 PM

mendelB:
> The reason this looks so broken on the review app is that the CSS rules are different than my local... still trying to research why this could be happening.

@aaronschachter Could this have to do with the ordering of the styling rules? The rules for `$small` are getting defined _before_ the default rules for the `div`, so I think they're getting overwritten by the later rules no matter what the viewport size. 

e.g. I'm seeing the `background-size` & `height` mobile specific rules are getting applied because they're _not_ overwritten in the later `div` styling since they're defined earlier. (Though why this would work on local and not dev I can't say -- maybe something to do with how the `scss` is compiled and built? 🤔)

One other thing -- I know @weerd always stresses to style for mobile-first, and to add any overrides for larger screens in a media query as opposed to using the media query for `$small` to begin with. Perhaps re-writing the rules that way -- so that the mobile rules are the default, and the other rules are only applied for `@include media($medium)` would ~kill two birds with one stone~ (violence 🙅‍♂) fix this in one fell swoop?

mendelB:
> I also forgot to mention in the PR that in working on this branch, I found any time I changed the CSS -- the Referral Page Banner gift card background image would disappear (mentioned this in Slack yesterday. I'd work around this by restarting npm, which was a little bit 😿

This one beats me 😖That's gotta be so frustrating!!

aaronschachter:
> Perhaps re-writing the rules that way -- so that the mobile rules are the default, and the other rules are only applied for @include media($medium) would kill two birds with one stone (violence 🙅‍♂) fix this in one fell swoop?

@mendelB You are a genius, this did the trick. Thank you for the help. I've also added a background color to fix the weird view I reported in the "Any background context you want to provide?" section.

resources/assets/components/utilities/CtaReferralPageBanner/cta-referral-page-banner.scss
```diff
@@ -1,18 +1,33 @@
 @import '../../../scss/next-toolbox';
 
 .cta-register-banner {
-  background: url('./cta-referral-page-banner-desktop.svg');
-  background-size: cover;
-  height: 270px;
+  background: #983ac9 url('./cta-referral-page-banner-mobile.svg');
```
mendelB: I think this is a clever hack for that issue you mentioned! I wonder though if there's an alternative way to prevent the somewhat abrupt end of the SVG with those background icons and stuff at either end:

image

Would using `background-size: cover;` here be an OK concession? I guess the tradeoff is the slightly more cut off envelope vs the cut off icons. 

image


(Or maybe a resized SVG (slightly wider??) would do the trick?)

resources/assets/components/utilities/CtaReferralPageBanner/cta-referral-page-banner.scss
```diff
@@ -1,18 +1,33 @@
...
     text-transform: uppercase;
   }
 
-  div {
-    float: left;
-    margin-left: 340px;
-    margin-top: 50px;
-    margin-right: 20px;
+  @include media($medium) {
+    background: #983ac9 url('./cta-referral-page-banner-desktop.svg');
```
mendelB: Do larger screens need the color as well?

resources/assets/components/utilities/CtaReferralPageBanner/cta-referral-page-banner.scss
weerd: If you define the position for where the background attaches it would stretch a little differently. My guess is it's defaulting to `center center` but if you specified something like `center top` then it would attach to the top and spread down. Only concern there is if it stretches the white envelope onto the white text, but that could be solved with some `text-shadow`.

weerd: As for an alternative solution, you could separate the background from the envelope/hand. Have the falling money background be a tileable background, which might be useful for use in scholarship or other rewards context, and it can just tile across the background, and then have the envelope and hand be a separate image that is attached to an element that sits on top.

weerd:
@aaronschachter when reviewing [here](https://dosomething-phoenix-de-pr-1587.herokuapp.com/us/campaigns/escape-vape?refer-friends=true) I see the closed campaign page but no RFA stuff. Am I doing something wrong?

resources/assets/components/utilities/CtaReferralPageBanner/cta-referral-page-banner.scss
```diff
@@ -1,18 +1,33 @@
...
 
 .cta-register-banner {
-  background: url('./cta-referral-page-banner-desktop.svg');
-  background-size: cover;
-  height: 270px;
+  background: #983ac9 url('./cta-referral-page-banner-mobile.svg');
+  background-size: 490px;
+  background-position: center;
+  background-repeat: no-repeat;
+  height: 382px;
```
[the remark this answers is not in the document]
aaronschachter: It looks like this one wasn't necessary, but the fixed height used for larger screens has been added because the inner div is set to `float: left`.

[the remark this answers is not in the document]
weerd: Ah yeah, then the solution for a fix is to include a `clearfix` on the parent container instead of fixing the height. Pretty much _never_ set a fixed height on anything with varied text; it always leads you to trouble, unless it's an icon or fixed image.

resources/assets/components/utilities/CtaReferralPageBanner/cta-referral-page-banner.scss
aaronschachter: I added the color to avoid tiling the background image, taking a guess at what should happen. Going to review these designs with Leah to better define how to avoid the situation above (whether to keep the background color, or split out the money and envelope graphics per https://github.com/DoSomething/phoenix-next/pull/1587#discussion_r321305573)

resources/assets/components/utilities/CtaReferralPageBanner/cta-referral-page-banner.scss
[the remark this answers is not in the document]
weerd: Meant to follow up. You can either use the `.clearfix` class in the parent container or use the `@include clearfix;` Sass method to do the same thing within the style definitions :)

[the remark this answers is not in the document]
aaronschachter: This overflow has been resolved by removing all fixed heights.

resources/assets/components/utilities/CtaReferralPageBanner/cta-referral-page-banner.scss
aaronschachter: [I've asked about this in Slack](https://dosomething.slack.com/archives/C2BLZGBPH/p1567721436010600) to determine whether this is OK as is, or whether we need different images (separate per breakpoints, or separate moneys/envelope graphic). I think for now this should be okay to merge, and we can revisit per feedback.

PR #1588, opened 2019-09-04
Fix nav link analytics and add analytics for DS logo link.

weerd:
### What does this PR do?

This PR fixes an issue with the analytics events in the Navigation added in #1580.

Turns out that since the nav items for "Explore Campaigns" and "What is DoSomething.org?" have subtext that shows on large size screens, sometimes what is clicked is the subtext in a `<span />` and thus the text obtained for the analytics `adjective` and `label` via `event.target.textContent` would vary (which we don't want to happen and prefer analytics consistency in event naming).

To simplify this, it seemed like a better approach to attach a `data-label` on all anchor links in the navigation and use the value assigned as the `adjective` and `label`, which makes it way more consistent and also much cleaner in naming that pulling from the elements text content.

This also adds a missing binding to the DS Logo link to trigger an analytics event as well!

Example:
image

Refs [Pivotal ID #168175576](https://www.pivotaltracker.com/story/show/168175576)

### Checklist

### DIFF
--- a/resources/assets/helpers/navigation.js
+++ b/resources/assets/helpers/navigation.js
@@ -18,6 +18,7 @@ export function linkAnalyticsHandler(link) {
   link.addEventListener(
     'mousedown',
     event => {
+      // If element is not an anchor tag, find the parent which should be an anchor tag!
       const element =
         event.target.tagName !== 'A'
           ? event.target.parentElement
--- a/resources/assets/helpers/navigation.js
+++ b/resources/assets/helpers/navigation.js
@@ -18,6 +18,11 @@ export function linkAnalyticsHandler(link) {
   link.addEventListener(
     'mousedown',
     event => {
+      const element =
+        event.target.tagName !== 'A'
+          ? event.target.parentElement
+          : event.target;
+
       trackAnalyticsEvent({
         context: {
           ...getUtmContext(),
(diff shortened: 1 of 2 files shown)


resources/views/partials/navigation.blade.php
```diff
@@ -33,15 +33,15 @@
 
             @if (Auth::user())
                 <li class="navigation__dropdown">
-                    <a id="js-account-toggle" class="navigation__dropdown-toggle">My Profile</a>
+                    <a id="js-account-toggle" class="navigation__dropdown-toggle" data-label="account-toggle">My Profile</a>
                     <ul>
-                        <li><a href="{{ url('us/account/profile') }}">Profile</a></li>
-                        <li><a href="{{ route('logout') }}" class="secondary-nav-item" id="link--logout">Log Out</a></li>
+                        <li><a href="{{ url('us/account/profile') }}" data-label="profle">Profile</a></li>
```
mendelB: *`profile` 👾

resources/assets/helpers/navigation.js
```diff
@@ -18,16 +18,21 @@ export function linkAnalyticsHandler(link) {
   link.addEventListener(
     'mousedown',
     event => {
+      const element =
```
mendelB: Do you think this could merit from a comment? It took me a sec to get what we were doing here (tho that could just be me 🙃)

weerd: Yeah, I was debating that. I'll add one in :)

resources/views/partials/navigation.blade.php
weerd: 👌

PR #1590, opened 2019-09-05
Footer link to close Affirmation modal

aaronschachter:
### What does this PR do?

In today's episode of "Hey let's try it", this PR adds a "Continue to campaign" footer link to Affirmation modals that close the modal. This feature has been requested per the Refer A Friend [designs](https://projects.invisionapp.com/share/GYSBOGRSDUA#/screens/372969109), however this really doesn't look anything like said designs. I'll be reviewing it with Leah shortly, as well as updating the `phoenix` dev client with the review app to make this testable.

I am feeling like it'd be a much bigger lift to style the footer link to look like the design (within the modal content). Refs https://github.com/DoSomething/phoenix-next/pull/1551#issuecomment-523628327

* https://www.pivotaltracker.com/story/show/167501569
* https://www.pivotaltracker.com/story/show/167633963

### Checklist

### DIFF
--- a/resources/assets/components/utilities/Modal/ModalContent.js
+++ b/resources/assets/components/utilities/Modal/ModalContent.js
@@ -11,7 +11,11 @@ const ModalContent = ({ children, onClose, className, closeFooterText }) => (
     {children}
 
     {closeFooterText ? (
-      <button type="button" className="text-white" onClick={onClose}>
+      <button
+        type="button"
+        className="modal__footer text-white"
+        onClick={onClose}
+      >
         {closeFooterText}
       </button>
     ) : null}
--- a/resources/assets/components/utilities/Modal/ModalContent.js
+++ b/resources/assets/components/utilities/Modal/ModalContent.js
@@ -2,17 +2,24 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
-const ModalContent = ({ children, onClose, className }) => (
+const ModalContent = ({ children, onClose, className, closeFooterText }) => (
   <div className={classnames('modal', className)}>
     <button type="button" className="modal__close" onClick={onClose}>
       &times;
     </button>
 
     {children}
+
+    {closeFooterText ? (
+      <button type="button" className="text-white" onClick={onClose}>
+        {closeFooterText}
+      </button>
+    ) : null}
   </div>
 );
 
 ModalContent.propTypes = {
+  closeFooterText: PropTypes.string,
   className: PropTypes.string,
   children: PropTypes.element.isRequired,
   onClose: PropTypes.func.isRequired,
@@ -20,6 +27,7 @@ ModalContent.propTypes = {
 
 ModalContent.defaultProps = {
   className: null,
+  closeFooterText: null,
 };
 
 export default ModalContent;
--- a/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
+++ b/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
@@ -38,7 +38,10 @@ const CampaignRoute = props => {
   return (
     <React.Fragment>
       {shouldShowAffirmation ? (
-        <Modal onClose={clickedHideAffirmation}>
+        <Modal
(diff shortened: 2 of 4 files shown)


aaronschachter:
Added the review app to our dev `phoenix` client for testing: https://dosomething-phoenix-de-pr-1590.herokuapp.com/us/campaigns/escape-vape (click on "Show Affirmation" button to easily find)

resources/assets/components/utilities/Modal/ModalContent.js
```diff
@@ -2,24 +2,36 @@ import React from 'react';
...
+
+    {closeFooterText ? (
+      <button
+        type="button"
+        className="modal__footer text-white"
+        onClick={onClose}
+      >
+        {closeFooterText}
+      </button>
+    ) : null}
```
katiecrane: Does this end up outside the modal box because it's not part of `{children}`?

aaronschachter: Yep. The `.modal__slide` div is what's rendering the inner modal box contents of `children`. We won't know how those children are styled, which is why moving the footer outside of the box (where we know the background is the dark background color defined by `.modal-portal`) seemed like a safe move. Again - this is pending design approval, but just seemed like an easy way to provide the functionality we're looking for.

Screen Shot 2019-09-05 at 4 22 15 PM

mendelB: An alternative idea that just occurred to me which _might_ be cleaner and would allow us to style this link within the affirmation itself -- is to yield the [`onClose` prop](https://github.com/DoSomething/phoenix-next/blob/cd44037e717e06fc92668ae96dbbadb16c9b4b09/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js#L41) from the Modal, through to the [`PostSignupModal`](https://github.com/DoSomething/phoenix-next/blob/cd44037e717e06fc92668ae96dbbadb16c9b4b09/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js#L42) all the way to the [`ContentfulEntry`](https://github.com/DoSomething/phoenix-next/blob/cd44037e717e06fc92668ae96dbbadb16c9b4b09/resources/assets/components/pages/PostSignupModal/PostSignupModal.js#L10), which will pass it to the Affirmation.

Once the Affirmation has access to this method as a prop, it can just add the link to itself and style it as we wish, and invoke the method `onClick` to close the modal!

The downside is the multi-level prop passing. But, I don't think it's unreasonable for the Affirmation to expect this prop since it's very much a modal specific component. (And I also wonder if this multi-level prop passing isn't as much an anti-pattern here but more of an indication that we can simplify this a bit! But that's beyond the scope of this feature of course).

Does this make sense? What do y'all think?

aaronschachter: I like this idea... by sticking with the designs, we'd sure get closer to shipping this thing!

aaronschachter: I suppose one aspect of this that doesn't feel the cleanest is that our `ContentfulEntry` component now will have an optional `onClose` prop, along with the `json` and `className` props. It doesn't really bother me, but makes me think I should work on this in a separate branch in case it gets we'd prefer this approach.

mendelB: Ah yeah fair point! 🤔 I wonder if we could prevent that by appending the `onClose` prop to the `json`:
```
const PostSignupModal = ({ affirmation, onClose }) => (
  <div className="modal__slide">
    <SlideshowContainer slideshowId="post-signup-modal" hideCloseButton>
      <ContentfulEntry json={{...affirmation, onClose}} />
    </SlideshowContainer>
  </div>
);
```


But maybe _that's_ too hacky?

aaronschachter: > But maybe _that's_ too hacky?

Coming up on today's episode of "Hey let's try it" 📺 💻 ⚡️

aaronschachter:
Closed in favor of #1595

PR #1591, opened 2019-09-05
Check existence of badgesData.user

aaronschachter:
### What does this PR do?

This PR fixes a bug we're seeing in the Affirmation per this [Slack thread](https://dosomething.slack.com/archives/C3ASB4204/p1567720528075700?thread_ts=1567712576.065400&cid=C3ASB4204)

The "Show Affirmation" modal closes for me after a few seconds, bc I don't have the badges feature flag. Seems introduced by https://github.com/DoSomething/graphql/pull/128

### Checklist

### DIFF
--- a/resources/assets/components/Affirmation/Affirmation.js
+++ b/resources/assets/components/Affirmation/Affirmation.js
@@ -1,6 +1,7 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import { has } from 'lodash';
 import gql from 'graphql-tag';
+import PropTypes from 'prop-types';
 
 import Query from '../Query';
 import { Flex, FlexCell } from '../Flex';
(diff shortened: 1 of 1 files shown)


resources/assets/components/Affirmation/Affirmation.js
```diff
@@ -49,7 +49,7 @@ const Affirmation = ({
 
     <Query query={BADGE_QUERY} variables={{ userId }} hideSpinner>
       {badgeData =>
-        badgeData.user.hasBadgesFlag ? (
+        badgeData.user && badgeData.user.hasBadgesFlag ? (
```
weerd: We could use lodash `has()` here:

```js
has(badgeData, 'user. hasBadgesFlag`) ? ...
```

Returns a boolean and can check for nested properties. If `user` not available, then it returns `false` without bothering to check for `hasBadgesFlag` :)

resources/assets/components/Affirmation/Affirmation.js
```diff
@@ -49,7 +50,7 @@ const Affirmation = ({
 
     <Query query={BADGE_QUERY} variables={{ userId }} hideSpinner>
       {badgeData =>
-        badgeData.user.hasBadgesFlag ? (
+        has(badgeData, 'user.hasBadgesFlag') ? (
```
katiecrane: Won't this see if we got a `hasBadgesFlag`? I think we also need to check and see if the value of it is `true`?

weerd: Oh, right. Didn't realize that. @aaronschachter maybe just switch it to `get()` and you can default to `false`.

```js
get(badgeData, 'user.hasBadgesFlag', false) ? ...
```

PR #1595, opened 2019-09-06
Affirmation close button

aaronschachter:
### What does this PR do?

This PR adds a close footer link in Affirmation per [Refer A Friend designs](https://projects.invisionapp.com/share/GYSBOGRSDUA#/screens/372969109) and per  https://github.com/DoSomething/phoenix-next/pull/1590#discussion_r321747054.

As @mendelB mentioned in https://github.com/DoSomething/phoenix-next/pull/1590#discussion_r321747054, this might be a bit too hacky -- but I think appending an `onClose` property into the `json` of the Affirmation component beats passing an optional `onClose` prop to the `ContentfulEntry` component, since it's only used by the Affirmation.

### Screenshots

* [mobile](https://user-images.githubusercontent.com/1236811/64453348-363e3480-d09d-11e9-8363-8ebd7bf67a9d.png)
* [tablet](https://user-images.githubusercontent.com/1236811/64453371-3dfdd900-d09d-11e9-94a4-2eb9329c3e6e.png)
* [desktop](https://user-images.githubusercontent.com/1236811/64453067-8ff22f00-d09c-11e9-99ab-95cff320eb5c.png)

### DIFF
--- a/resources/assets/components/Affirmation/affirmation.scss
+++ b/resources/assets/components/Affirmation/affirmation.scss
@@ -10,4 +10,11 @@
       font-weight: 800;
     }
   }
+
+  .close-button {
+    color: $blue;
+    cursor: pointer;
+    font-size: $font-regular;
+    font-weight: 800;
+  }
 }
--- a/resources/assets/components/Affirmation/affirmation.scss
+++ b/resources/assets/components/Affirmation/affirmation.scss
@@ -14,7 +14,5 @@
   .close-button {
     color: $blue;
     cursor: pointer;
-    font-size: $font-regular;
-    font-weight: 800;
   }
 }
--- a/resources/assets/components/Affirmation/Affirmation.js
+++ b/resources/assets/components/Affirmation/Affirmation.js
@@ -96,9 +96,11 @@ const Affirmation = ({
 
     {query('refer-friends') ? <CtaReferralPageBannerContainer /> : null}
 
-    <button type="button" onClick={onClose}>
-      Continue to Campaign
-    </button>
+    <div className="p-4">
+      <button type="button" className="close-button" onClick={onClose}>
+        Continue to Campaign
+      </button>
+    </div>
   </Card>
 );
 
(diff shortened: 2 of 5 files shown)


aaronschachter:
Added this review app to our dev client for testing. Worth noting this footer will show up regardless of whether the `refer-friends` query parameter is set - I don't think this should be a biggie.

https://dosomething-phoenix-de-pr-1595.herokuapp.com/us/campaigns/test-teens-for-jeans/action

https://dosomething-phoenix-de-pr-1595.herokuapp.com/us/campaigns/test-teens-for-jeans/action?refer-friends=true

If we go this route I will close #1590 without merging.

resources/assets/components/Affirmation/affirmation.scss
```diff
@@ -10,4 +10,11 @@
       font-weight: 800;
     }
   }
+
+  .close-button {
```
mendelB: We could avoid some of these rules by using utility classes on the markup itself: [`.font-bold`](https://github.com/DoSomething/phoenix-next/blob/a0ebcc772c14e114026eda27083383627fcb8023/resources/assets/scss/base.scss#L506-L508) (though that's 700 and not 8) & [`.text-base`](https://github.com/DoSomething/phoenix-next/blob/a0ebcc772c14e114026eda27083383627fcb8023/resources/assets/scss/base.scss#L656-L658)

mendelB:
>  Worth noting this footer will show up regardless of whether the refer-friends query parameter is set - I don't think this should be a biggie.

I think it's a good thing!

weerd:
I saw this convo happening, but in the middle of a bunch of stuff. I can try and chime in a bit later with any suggestions I might have. One side note, debating if we should get rid of the modal slide crap, since we stopped using that a while back and it's been annoyingly sitting there; but likely unrelated to this, haha.

mendelB:
> One side note, debating if we should get rid of the modal slide crap, since we stopped using that a while back and it's been annoyingly sitting there; but likely unrelated to this, haha.

That'd be quite nice I think! It'd allow us to get rid of the [`PostSignupModal`](https://github.com/DoSomething/phoenix-next/blob/21762b802f0de641ba96172219e65ad6a612d7c2/resources/assets/components/pages/PostSignupModal/PostSignupModal.js) altogether, which would remove a nested layer here.

PR #1598, opened 2019-09-06
Track analytics event for an Embed click

aaronschachter:
### What does this PR do?

This PR tracks a new analytics event for when a user clicks on the link in an `Embed` component. This can potentially resolve tracking the two [Refer A Friend events we wish to track](https://docs.google.com/spreadsheets/d/1hNZdkcS7orp7xU0mOHSRKPodNoif0EP5UC2YwOnDmPI/edit#gid=1098747526):

* Track clicks on the Beta page links to campaigns.
* Track clicks on alpha page link to view their beta page

I started a discussion in the [Pivotal story](https://www.pivotaltracker.com/n/projects/2019429/stories/168201845) but it may be easier to hash things out in this PR, where we can comment per line. Here are some outstanding questions:

* What category should this fall into? In Dev RT, the idea of a new category was discussed - but we need a name for said category if so. I picked `campaign_action` thinking that these are mainly used in campaign pages (although questionable if Refer A Friend clicks should toward that). Also we might use `Embed` components elsewhere, which may warrant refactoring this PR to send the track event payload as a prop to an `Embed` component...  

* Will using the `context.url` and the page the event was fired from give us enough information to track the desired clicks on the Alpha / Beta Referral Pages?

* `adjective` - I wonder whether it'd be useful to save something here like `internal` / `external` based on `isExternal(url)`

Refs https://www.pivotaltracker.com/n/projects/2019429/stories/168201845

### Checklist

* ~[ ] Added screenshot to PR description of related front-end updates on **small** screens.~
* ~[ ] Added screenshot to PR description of related front-end updates on **medium** screens.~
* ~[ ] Added screenshot to PR description of related front-end updates on **large** screens.~

### DIFF
--- a/resources/assets/components/utilities/Embed/Embed.js
+++ b/resources/assets/components/utilities/Embed/Embed.js
@@ -38,8 +38,9 @@ const Embed = props => {
         url,
       },
       metadata: {
-        category: 'campaign_action',
-        noun: 'embed',
+        category: 'site_action',
+        adjective: 'embed',
+        noun: 'link',
         verb: 'clicked',
       },
     });
--- a/resources/assets/components/utilities/Embed/Embed.js
+++ b/resources/assets/components/utilities/Embed/Embed.js
@@ -12,6 +12,7 @@ import linkIcon from './linkIcon.svg';
 import { isExternal } from '../../../helpers';
 import ErrorBlock from '../../blocks/ErrorBlock/ErrorBlock';
 import PlaceholderText from '../PlaceholderText/PlaceholderText';
+import { trackAnalyticsEvent } from '../../../helpers/analytics';
 
 import './embed.scss';
 
@@ -30,6 +31,19 @@ const EMBED_QUERY = gql`
 
 const Embed = props => {
   const { url, badged, className } = props;
+
+  const onClick = () => {
+    trackAnalyticsEvent({
+      context: {
+        url,
+      },
+      metadata: {
+        category: 'campaign_action',
+        noun: 'embed',
+        verb: 'clicked',
+      },
+    });
+  };
 
   return (
     <div className={classnames('bordered', 'rounded', 'bg-white', className)}>
@@ -55,6 +69,7 @@ const Embed = props => {
               className="embed__linker"
               target={isExternal(url) ? '_blank' : '_self'}
               rel="noopener noreferrer"
+              onClick={onClick}
             >
               <div className="embed">
                 <LazyImage
--- a/resources/assets/components/utilities/Embed/Embed.js
+++ b/resources/assets/components/utilities/Embed/Embed.js
@@ -41,6 +41,7 @@ const Embed = props => {
         category: 'site_action',
         adjective: 'embed',
         noun: 'link',
+        target: 'link_embed',
         verb: 'clicked',
       },
     });
--- a/resources/assets/components/utilities/Embed/Embed.js
+++ b/resources/assets/components/utilities/Embed/Embed.js
@@ -41,7 +41,7 @@ const Embed = props => {
         category: 'site_action',
         adjective: 'embed',
         noun: 'link',
-        target: 'link_embed',
+        target: 'link',
         verb: 'clicked',
       },
     });


mendelB:
> What category should this fall into?

I [commented](https://www.pivotaltracker.com/story/show/168201845/comments/206520329) on the Pivotal card, but am totally unsure.

> Also we might use Embed components elsewhere, which may warrant refactoring this PR to send the track event payload as a prop to an Embed component...

I think this is a super solid point -- especially since we already do! Within `LinkAction`s, [we often](https://github.com/DoSomething/phoenix-next/blob/d48fe11a5259aa53203bd2d0c24ac9317052bc58/resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js#L57-L76) render the link using an `Embed`, and [we wrap it](https://github.com/DoSomething/phoenix-next/blob/d48fe11a5259aa53203bd2d0c24ac9317052bc58/resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js#L63) in the `LinkAction` specific analytics event, which would cause two events to fire if we added an additional one to the `Embed` itself.

So it might indeed be prudent to formalize a prop for the event. Perhaps we have a local default `phoenix_clicked_link_embed` event, which can be overwritten by the prop? And so the `SocialShareAction` and the `LinkAction` can overwrite it, but other embeds will default to the general one? 🤔 

On the other hand, this could also complicate things with an assortment of events tracking essentially the same user action, so maybe we just have the one local event, with an additional optional `adjective` or `context` prop which would identify the specific rendering parent? Maybe this prop is the `Label`?

resources/assets/components/utilities/Embed/Embed.js
```diff
@@ -31,6 +32,20 @@ const EMBED_QUERY = gql`
...
+  const onClick = () => {
+    trackAnalyticsEvent({
+      context: {
+        url,
+      },
+      metadata: {
+        category: 'site_action',
+        adjective: 'embed',
+        noun: 'link',
+        verb: 'clicked',
```
aaronschachter: Am I missing something here? I've added the SNOWPLOW_URL and GOOGLE_TAG_MANAGER_ID vars from dev.dosomething.org on my local, and also into this review app to make sure this looks ok, but the event action reads as "Undefined Clicked"

Screen Shot 2019-09-13 at 9 46 09 AM

Note: The "Site Action" category came from this [Pivotal discussion](https://www.pivotaltracker.com/story/show/168201845/comments/206521237).

mendelB: @aaronschachter I think you're just missing the `metadata.target` field, (which can often seem quite similar to the `noun` confusingly, but is a distinct concern and can vary.)

I think in this case it might be `link_embed`?

aaronschachter: Thanks, that was it! I was following https://github.com/orgs/DoSomething/teams/tech-leads/discussions/3, which didn't mention `target` at all - so I thought it was optional.

weerd: Sorry about that @aaronschachter. The analytics work was such a whirlwind of stuff that happened entirely unplanned and we didn't have an opportunity to get better docs/update docs. Mostly it's a combo between that tech doc and the spreadsheet I talked about during DevRT. I'll spend a bit of time this week and either update that GH discussion, or find a better place for documenting this 😬

weerd: I think the best reference for now is just the [`trackAnalyticsEvent()` function](https://github.com/DoSomething/phoenix-next/blob/dbd98fc6450fc134bc578118748bcf1929182303/resources/assets/helpers/analytics.js#L262-L291).

weerd: And to just clarify a bit further, the info in <https://github.com/orgs/DoSomething/teams/tech-leads/discussions/3> is meant _just_ for the event naming structure, and the spreadsheet is for further info on the structures for the third-party (GA & Snowplow) event data (target, category, etc)

Defs a bit confusing and we can aim to fix this.

resources/assets/components/utilities/Embed/Embed.js
```diff
@@ -31,6 +32,21 @@ const EMBED_QUERY = gql`
...
+  const onClick = () => {
+    trackAnalyticsEvent({
+      context: {
+        url,
+      },
+      metadata: {
+        category: 'site_action',
+        adjective: 'embed',
+        noun: 'link',
+        target: 'link_embed',
```
weerd: I think the `noun` and the `target` should be flipped here. The target is typically just "what item was acted upon", which in this case is the link.

For the GA & Snowplow events the `label` provides the extra context, which if not provided defaults to the noun, and in this case it should be `link_embed`.

aaronschachter: I'm having a hard time following this logic per https://github.com/DoSomething/phoenix-next/pull/1598#discussion_r324293434 and the [Pivotal thread](https://github.com/DoSomething/phoenix-next/pull/1598#discussion_r324293434) - which served as the spec for sourcing these values. Per this logic you suggest and in Mendel's comment I linked to, then the action would be `Link Embed Clicked` ?

resources/assets/components/utilities/Embed/Embed.js
aaronschachter: I think the analytics docs that Mendel is working on in #1613 will be a good home for a one-stop shop for devs (vs the  Github discussion or Google sheet)

weerd: Yep yep. I saw that after commenting here. I think that will be a good place to start with it, but I think eventually would need to be a communal doc somewhere since these rules apply to Northstar too.

resources/assets/components/utilities/Embed/Embed.js
weerd: Provided some clarification in [Slack](https://dosomething.slack.com/archives/C3ASB4204/p1568660448004200)

I think I would actually amend it to this:

```
category: 'site_action',
noun: 'embed_link',
target: 'link',
verb: 'clicked',
```

The trigger for the event is a link in the embed component. The specific target is a link that was clicked.

I tackled something similar with the link clicks in the nav recently if you want to use it [as an example](https://github.com/DoSomething/phoenix-next/blob/dbd98fc6450fc134bc578118748bcf1929182303/resources/assets/helpers/navigation.js#L27-L41).

weerd: You could add the `adjective` if you'd like, but it'd have to get passed in... could be something that specifies what this was for, but not a big deal if not. The `context.url` has the link details anyhoo.

weerd: This would also work:

```
adjective: 'embed',
category: 'site_action',
noun: 'link',
target: 'link',
verb: 'clicked',
```

Based on the PT story. Sorry I tried clicking on the link you popped in this thread but I think it's incorrect and just links to GH so didn't see it.

aaronschachter: I defer to you on what would be best, as you've definitely spent a lot more time in this world than me! I'm now a little bit confused on whether to use the `adjective` - but leaning towards excluding if we don't need it (less is more 🌎). Sorry about the typo on the [Pivotal link](https://www.pivotaltracker.com/story/show/168201845/comments/206520329), I updated it there  -- this is what steered the current direction

weerd: Use the adjective as `embed`, I think your instinct on that was correct!

mendelB:
@weerd @aaronschachter
I wanted to highlight what I mentioned in https://github.com/DoSomething/phoenix-next/pull/1598#issuecomment-529982977

Specifically this point:
>> Also we might use Embed components elsewhere, which may warrant refactoring this PR to send the track event payload as a prop to an Embed component...

> I think this is a super solid point -- especially since we already do! Within `LinkAction`s, [we often](https://github.com/DoSomething/phoenix-next/blob/d48fe11a5259aa53203bd2d0c24ac9317052bc58/resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js#L57-L76) render the link using an `Embed`, and [we wrap it](https://github.com/DoSomething/phoenix-next/blob/d48fe11a5259aa53203bd2d0c24ac9317052bc58/resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js#L63) in the `LinkAction` specific analytics event, which would cause two events to fire if we added an additional one to the `Embed` itself.

Since now, clicking on a Link Action as an embed would trigger _two_ analytics events:
- `phoenix_clicked_link_action`
- `phoenix_clicked_link_embed`

I figure this could potentially confuse our analytics?

aaronschachter:
@mendelB @weerd - Are there currently any Link Actions live on the site? Should we revert this? It hasn't been deployed to production yet.

 Getting in analytics was a nice to have for the Refer A Friend launch but not critical. Not sure when https://www.pivotaltracker.com/story/show/168302139 will get in, but seems like reverting may be a safe move if there are Link Actions live that would confuse analytics..

mendelB:
> @mendelB @weerd - Are there currently any Link Actions live on the site? Should we revert this? It hasn't been deployed to production yet.
> 
> Getting in analytics was a nice to have for the Refer A Friend launch but not critical. Not sure when https://www.pivotaltracker.com/story/show/168302139 will get in, but seems like reverting may be a safe move if there are Link Actions live that would confuse analytics..

It's hard to know how many are live but I'd assume not many. 

Hmmm in order not to hold up RAF analytics, and until we do tackle that ticket, what do you think about an interim hack where we add an `Embed` component prop to `blockAnalyticsEvent` or something which the `LinkAction` can pass through to conditionally just `return;` from the top of the `onClick` callback in the embed?

aaronschachter:
We dug deeper into the duplicate events in today's session: they occur when a `LinkAction` has a `template` field value not equal to CTA. Doing a Contentful search for this results in about 255 entries, many of which are live on article pages.

We thought that our best bet for now is to remove tracking events in the `Embed`. Megha had questions about `phoenix_clicked_link_action` vs `phoenix_clicked_link_embed`, and I'd feel more comfortable not changing/testing other components that don't touch RAF since we're hoping to go live this week. I'll open up a pull request that comments out tracking this `Embed` click event and refers to this pull request as reference.

mendelB:
> We dug deeper into the duplicate events in today's session: they occur when a `LinkAction` has a `template` field value not equal to CTA. Doing a Contentful search for this results in about 255 entries, many of which are live -- e.g. /facts/11-facts-about-cyber-bullying

Sorry, just to clarify - this only would happen for LinkActions where the `template` field is not `cta` _and_ where the `content` field is empty (which would render them as an embedded link using the `Embed` component as opposed to a card with content and a button link). I doubt there are too many on live campaigns which aren't youtube links (which would not trigger the event, since it'd render a youtube video and not an embedded link).

But that aside -- I understand if you want to table the embed analytics event for now!

PR #1602, opened 2019-09-09
updates embed content-type in contentful config

lindsayrmaher:
### What does this PR do?

This PR updates the embed configuration for contentful to align with the changes we made ahead of the Quiz implementation.

Needs to be updated in order to finalize Quiz implementation.

Sample URLS:

**TypeForm**
```
https://dosomething.typeform.com/to/samplepass
https://dosomething.typeform.org/to/sample
http://dosomething.typeform.com/to/sample
```

**Carto**
```
https://dosomething.carto.com/builder/samplepass/embed
https://dosomething.carto.org/builder/sample/embed
http://dosomething.carto.com/builder/sample/embed
```

### Checklist


lindsayrmaher:
@weerd added some samples! let me know if you need more or edits

contentful/content-types/embed.js
```diff
@@ -25,12 +24,12 @@ module.exports = function(migration) {
       {
         regexp: {
           pattern:
-            'https:\\/\\/dosomething\\.carto\\.com\\/builder\\/[a-z0-9-]+\\/embed',
+            'https:\\/\\/dosomething\\.(?:carto|typeform)\\.com\\/(?:builder|to)\\/[a-zA-Z0-9-]+',
```
weerd: What's the `?:` do?

lindsayrmaher: It makes it a non capturing group. If you make it a capturing group you can back reference it elsewhere, but that didn't feel like a use case we needed.

mendelB: Do you think it might make sense to split the URL regex into two distinct groups for the two URLs? Since hypothetically now, `https://dosomething.carto.com/to/1` or `https://dosomething.typeform.com/builder/1` would be valid. 🤔

weerd: Ah yeah, that's a good point... also might be useful to ensure it ends after the second segment, cause this works too:

```
https://dosomething.typeform.com/to/formid/something/extra/here/for/a/super/long/url
```

lindsayrmaher: yeah i was struggling with how to handle the differences in the two urls past the second '/', would it be possible in contentful to split the regex validations for the two URLS?

mendelB: I _think_ you could do it somewhat manually with regular ole Regex groups. e.g. something like: 
```
'https:\\/\\/dosomething\\.(carto\\.com\\/builder\\/[a-z0-9-]+\\/embed)|(typeform\\.com\\/to\\/[a-zA-Z0-9-]+)'
```

It's slightly repetitive but I think that's ok for our purposes!

lindsayrmaher: that makes sense! i'll test with some samples.

weerd: If we end up adding more _allowable_ URLs we may want to separate with dedicated (optional) fields 🤔 Otherwise these regexes could get a little crazy, haha!

lindsayrmaher: agreed! for now i tweaked the example Mendel wrote and tested in regexr. let me know if those changes make sense to you both.

weerd:
```
^https:\/\/dosomething\.((carto\.com\/builder\/[a-z0-9-]+\/embed)|(typeform\.com\/to\/[a-zA-Z0-9-]+))\/?$
```

Looks good, I just added the beginning carot `^` to indicate the start of the string (otherwise you could accidentally add some characters before the `http`, and also added `\/?$` to indicate an optional `/` at the end of the second segment in the URL and a `$` to indicate the end of the string. This keeps it from having any more URL segments appended at the end 💃 


_*Disclaimer: Regexr gets mad at me if I use the double escape `\\` so the above example is amended to the single escape `\`. Just something to note!_

weerd:
@lindsayrmaher I think my comment got lost since since the response was updated...

```
^https:\/\/dosomething\.((carto\.com\/builder\/[a-z0-9-]+\/embed)|(typeform\.com\/to\/[a-zA-Z0-9-]+))\/?$
```

Looks good, I just added the beginning carot `^` to indicate the start of the string (otherwise you could accidentally add some characters before the `http`, and also added `\/?$` to indicate an optional `/` at the end of the second segment in the URL and a `$` to indicate the end of the string. This keeps it from having any more URL segments appended at the end 💃 


_*Disclaimer: Regexr gets mad at me if I use the double escape `\\` so the above example is amended to the single escape `\`. Just something to note!_

lindsayrmaher:
@weerd good call! I'll update and push up that final change so we can merge 🤸‍♀

lindsayrmaher:
@weerd i think this test failure is the same one we talked about last week that i wasn't sure how to fix. Wondering whether we can look into it before mergin

mendelB:
> @weerd i think this test failure is the same one we talked about last week that i wasn't sure how to fix. Wondering whether we can look into it before mergin

Ya know, I've seen this failure a few times now. Usually something like 
```
 jest: failed to cache transform results in: /tmp/jest_2ne/jest-transform-cache-...
    Failure message: ENOMEM: not enough memory, open '/tmp/jest_2ne/jest-transform-cache-cfbff7f8ba39fa6c314c...
```

The tests seem to pass on a manual rebuild.

It seems like others have encountered this with Jest on CircleCI, and it looks like the suggested fix is to run the tests with [` --runInBand`](https://jestjs.io/docs/en/cli.html#runinband)
e.g.
https://discuss.circleci.com/t/babel-7-jest-results-in-timeout/25383
https://vgpena.github.io/jest-circleci/

Perhaps we can try adding this flag to the `jest` command for the ci, and see if that helps fix this?
https://github.com/DoSomething/phoenix-next/blob/b42097394b5a76011de8ffabc42e2038b47352c1/package.json#L5

weerd:
@lindsayrmaher maybe try @mendelB's suggestion w/ the setting?

DFurnes:
Killer sleuthing, @mendelB! I like the suggestion of trying `--max-Workers=2`. 👍 Also, ummm, did anyone else know that you can [SSH into failed builds to debug them](https://circleci.com/docs/2.0/ssh-access-jobs/) (mentioned in the forum post you linked)!??! Game changer!!

lindsayrmaher:
SUCCESS! @DFurnes @weerd @mendelB I'm squashin and mergin!

mendelB:
> Also, ummm, did anyone else know that you can [SSH into failed builds to debug them](https://circleci.com/docs/2.0/ssh-access-jobs/) (mentioned in the forum post you linked)!??! Game changer!!

IKR?? I had no idea!

PR #1606, opened 2019-09-10
Enable more flexibility with grid classes in RichText rendered components.

weerd:
### What does this PR do?

This PR updates an issue we've had since the original `StoryPage` was completed. At the time we had hard-coded the `grid-narrow` class for any elements rendered using RichText. It worked great at the time when the `StoryPage` was the only component using RichText, and most of the components rendered on that page adhered to the `grid-narrow` layout.

However, now that we're expanding use and looking to implement the CSS Grid across more components, not all of them need to layout within the `grid-narrow`. So this PR provides an alternative approach when rendering components, which allows passing a `classNameByEntry` object that defines `defaults` for classes to always add to each component and the option to supply specific classes for dedicated components that get rendered in a particular context.

As an example, the `StoryPage` passes the following `classNameByEntry` object prop:

```js
classNameByEntry={{
  default: 'grid-narrow',
  EmbedBlock: 'grid-main',
  PostGalleryBlock: 'grid-main',
}}
```
Which indicates that by default, all components rendered in the `StoryPage` get a `grid-narrow` class, and the `EmbedBlock` and `PostGalleryBlock` get the `grid-main` class.

Refs [Pivotal ID #167297536](https://www.pivotaltracker.com/story/show/167297536)

### Checklist

### DIFF
--- a/resources/assets/components/ContentfulEntry/ContentfulEntry.js
+++ b/resources/assets/components/ContentfulEntry/ContentfulEntry.js
@@ -39,6 +39,7 @@ const DEFAULT_BLOCK: ContentfulEntryJson = { fields: { type: null } };
 type Props = {
   json: ContentfulEntryJson,
   className: String,
+  classNameByEntry: Object,
 };
 type State = { hasError: boolean };
 
@@ -59,7 +60,11 @@ class ContentfulEntry extends React.Component<Props, State> {
     }
 
     // Otherwise, find the corresponding component & render it!
-    const { json = DEFAULT_BLOCK, className = null } = this.props;
+    const {
+      json = DEFAULT_BLOCK,
+      className = null,
+      classNameByEntry = {},
+    } = this.props;
     const type = parseContentfulType(json);
 
     switch (type) {
@@ -248,7 +253,7 @@ class ContentfulEntry extends React.Component<Props, State> {
         return (
           <SectionBlock
             className={className}
-            classNameByEntry={this.props.classNameByEntry}
+            classNameByEntry={classNameByEntry}
             id={json.id}
             {...withoutNulls(json.fields)}
           />
--- a/resources/assets/components/ContentfulEntry/ContentfulEntry.js
+++ b/resources/assets/components/ContentfulEntry/ContentfulEntry.js
@@ -248,6 +248,7 @@ class ContentfulEntry extends React.Component<Props, State> {
         return (
           <SectionBlock
             className={className}
+            classNameByEntry={this.props.classNameByEntry}
             id={json.id}
             {...withoutNulls(json.fields)}
           />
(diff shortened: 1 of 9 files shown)


resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -59,7 +60,11 @@ class ContentfulEntry extends React.Component<Props, State> {
     }
 
     // Otherwise, find the corresponding component & render it!
-    const { json = DEFAULT_BLOCK, className = null } = this.props;
+    const {
+      json = DEFAULT_BLOCK,
+      className = null,
+      classNameByEntry = {},
```
weerd: Is this fine as a default? Also can we just remove the use of Flow... I don't know if the experiment really worked and just makes this and maybe a couple other files confusing for people jumping into Phoenix.

resources/assets/components/utilities/TextContent/TextContent.js
```diff
@@ -46,6 +53,9 @@ TextContent.propTypes = {
 
 TextContent.defaultProps = {
   className: null,
+  classNameByEntry: {
+    default: 'grid-main',
+  },
```
weerd: I wasn't 100% sure about defaulting some of these components w/ this as the default prop. Alternative could be to default to empty object, and then entirely letting the `ContentfulEntryLoader` handle defaulting to `grid-main`? 🤔

resources/assets/components/ContentfulEntry/ContentfulEntry.js
DFurnes: I think that's a fine default, and yup 👍 to removing Flow.

resources/assets/components/blocks/SectionBlock/SectionBlock.js
```diff
@@ -60,6 +63,9 @@ SectionBlock.propTypes = {
 SectionBlock.defaultProps = {
   backgroundColor: null,
   className: null,
+  classNameByEntry: {
+    default: 'grid-main',
```
DFurnes: Instead of a "magic" named property, I wonder if this should be a separate `classnameByEntryDefault` prop? That might be a bit more discoverable, and reduces the (arguably very unlikely) chance that this would conflict with a content-type name.

resources/assets/components/utilities/TextContent/TextContent.js
DFurnes: Yeah, I think defaulting to an empty object makes the most sense. 👍

resources/assets/components/ContentfulEntry/ContentfulEntry.js
weerd: K. We can remove in a future PR. I'll make a ticket.

resources/assets/components/blocks/SectionBlock/SectionBlock.js
weerd: 🤔 yeah, I could dig that.

weerd: I might also pluralize these `classNamesByEntry` and `classNamesByEntryDefault` since you can technically add multiple in a single string 🤔

PR #1613, opened 2019-09-12
Add Analytics Waypoint to Phoenix

mendelB:
### What does this PR do?
🎯This PR adds a dead simple Analytics Waypoint component to Phoenix.

I renamed the component to `AnalyticsWaypoint` -- LMK of any preferred alternatives!

🏒As we retire our legacy analytics package [Puck](https://github.com/DoSomething/puck), we need to add an in-house replacement for its [`Waypoint`](https://github.com/DoSomething/puck-client#react-waypoints) functionality, which serves to analyze scroll depth and feature impressions.

This takes a slight detour from [Pucks usage](https://github.com/DoSomething/puck-client/blob/9d3a4cb8a47d8e154dd52c3828d1656529b9fe66/lib/Waypoint.js#L3) of the [`@researchgate/react-intersection-observer`](https://www.npmjs.com/package/@researchgate/react-intersection-observer) package in favor of [this one](https://www.npmjs.com/package/react-intersection-observer) for reasons mentioned [here](https://www.pivotaltracker.com/story/show/167630420/comments/206535494) (It also has a nifty Hooks implementation!)

🥘I followed the packages recipe for ['Track Impressions'](https://github.com/thebuilder/react-intersection-observer/blob/master/docs/Recipes.md#track-impressions), since it follows our use case to the T!

🔧If you're curious to peek under the hood, we're making use of the packages [`useInView`](https://github.com/thebuilder/react-intersection-observer/blob/master/src/useInView.tsx) Hooks-esque method. It's a pretty straightforward one!

#### Event Naming:
The Pivotal Ticket has a description of the agreed-upon naming scheme, and you can see our conversation in [Slack#topic-web-analytics](https://dosomething.slack.com/archives/CJU7WCQG4/p1568217409027400).

I tested placing a `<AnalyticsWaypoint name="photo_submission_action-[start/end] context..."/>` at the start and end of the `PhotoSubmissionAction` and here's the console output for scrolling past it on a campaign:
Screen Shot 2019-09-12 at 12 30 49 PM

(I updated the `/analytics.md` documentation a bit in https://github.com/DoSomething/phoenix-next/pull/1613/commits/379321978bdb758762a5813ea4f33629f218bd5c while adding documentation of this component. It still could use some work, but I just wanted to get it a bit more up to par with our current implementation without spending too much time on it beyond the scope of this feature!)

Refs [Pivotal ID #167630420](https://www.pivotaltracker.com/story/show/167630420)

### Checklist

### DIFF
--- a/resources/assets/components/utilities/AnalyticsWaypoint/AnalyticsWaypoint.js
+++ b/resources/assets/components/utilities/AnalyticsWaypoint/AnalyticsWaypoint.js
@@ -1,0 +1,19 @@
+import React, { useEffect } from 'react';
+import { useInView } from 'react-intersection-observer';
+
+const AnalyticsWaypoint = () => {
+  const [ref, inView] = useInView({
+    triggerOnce: true,
+  });
+
+  useEffect(() => {
+    if (inView) {
+      // @TODO: Track analytics event here!
+      console.log('ANALYTICS WAYPOINT!');
+    }
+  }, [inView]);
+
+  return <div ref={ref} />;
+};
+
+export default AnalyticsWaypoint;
--- a/resources/assets/components/utilities/AnalyticsWaypoint/AnalyticsWaypoint.js
+++ b/resources/assets/components/utilities/AnalyticsWaypoint/AnalyticsWaypoint.js
@@ -1,19 +1,42 @@
+import PropTypes from 'prop-types';
 import React, { useEffect } from 'react';
 import { useInView } from 'react-intersection-observer';
 
-const AnalyticsWaypoint = () => {
+import { trackAnalyticsEvent } from '../../../helpers/analytics';
+
+const AnalyticsWaypoint = ({ name, context }) => {
   const [ref, inView] = useInView({
     triggerOnce: true,
   });
 
   useEffect(() => {
     if (inView) {
-      // @TODO: Track analytics event here!
-      console.log('ANALYTICS WAYPOINT!');
+      trackAnalyticsEvent({
+        metadata: {
+          verb: 'reached',
+          noun: 'waypoint',
+          target: 'waypoint',
+          category: 'waypoint',
+          label: name,
+        },
+        context: {
+          ...context,
+          name,
+        },
+      });
     }
   }, [inView]);
 
   return <div ref={ref} />;
+};
+
+AnalyticsWaypoint.propTypes = {
+  name: PropTypes.string.isRequired,
+  context: PropTypes.object,
+};
+
+AnalyticsWaypoint.defaultProps = {
+  context: {},
 };
 
 export default AnalyticsWaypoint;
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -32,6 +32,7 @@
   - [Dismissable Element](development/features/delayed-element.md)
   - [Affiliate Opt In](development/features/affiliate-opt-in.md)
   - [Affiliate Scholarship Block](development/features/affiliate-scholarship-block.md)
+  - [Analytics Waypoint](development/features/analytics-waypoint.md)
 - [Contentful](development/contentful/README.md)
   - [Migration Scripts](development/contentful/migration-scripts.md)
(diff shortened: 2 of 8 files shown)


weerd:
> (I updated the /analytics.md documentation a bit in 3793219 while adding documentation of this component. It still could use some work, but I just wanted to get it a bit more up to par with our current implementation without spending too much time on it beyond the scope of this feature!)

Noice! 💃

resources/assets/components/utilities/AnalyticsWaypoint/AnalyticsWaypoint.js
```diff
@@ -0,0 +1,42 @@
...
+        },
+        context: {
+          ...context,
+          name,
+        },
+      });
+    }
+  }, [inView]);
+
+  return <div ref={ref} />;
```
weerd: Curious... we can't use a React Fragment for this right? Wondering since as we implement Grid, if we can't use a fragment, if these components that use the Waypoint have a dedicated css grid, it might affect how the items work. If not, it's ok and it's something we can work around, but mostly wondering if it's at all possible!

mendelB: Very good point! It's something I tested while implementing, but unfortunately, nothing would actually be rendered to the DOM, so there was no waypoint to be tracked 😞 

Is there another tag that you think would be better here for styling concerns? A `<span>` perhaps?

mendelB: 🤔 An interesting thing we could actually try is to `display:none;` this - or better yet - render `null` from this component once it intersects so that by the time the user can actually see the affected markup, it'd unmount, and hopefully restore any styling it affected?

mendelB: Tested the above idea and it works! Rendering a `<span>` also works so lemme know if either is preferred!

(or both! `return !inView ? <span ref={ref} /> : null;`)

weerd: I'm a little weary about changing from rendering a div to not rendering a div when it comes into view. I think we could just use the `hidden` Tailwind class to make the `div` `display: none`, and I _think_ that takes it out of consideration for the grid, which should be fine and feels like less of flurry of changes if scrolling quickly.

weerd: Also if when it comes into view and renders the element it would still be a problem with the grid.... no?

mendelB: Unfortunately, the Waypoint won't actually trigger if the element has `display: none;` to begin with. I think the element needs to visually affect the viewport somehow in order to register with the Intersection Observer.

image
https://www.smashingmagazine.com/2018/01/deferring-lazy-loading-intersection-observer-api/#things-to-keep-in-mind

Our only bet would be to add `display: none` once the element came into view and the observer was triggered.

mendelB: > Also if when it comes into view and renders the element it would still be a problem with the grid.... no?

I think you're right, and it could potentially blip for a sec, but I guess the win would be that it wouldn't permanently affect the layout? But I dunno how big of a win that is instead of figuring out how to address such as issue head-on if it becomes a problem!

weerd: Yeah. Ok, lets shelve this convo for now. We don't _actually_ know it's going to affect grid layout of stuff, since grid will mostly be affecting how these blocks that _contain_ the waypoints will be positioned. So may just be a moot point... Let's leave as div with no styling for now and we can come up w/ some fancy solution if need be :)

weerd: But excellent find! 🕵

PR #1616, opened 2019-09-17
Adds Design Updates to Badge Modal Close Button

Vmack:
This PR changes the 'x' on the badge modals to white

An `-inverted` class was added to the .modal selector and added to the `Modal` container. This will only change the badge modal's 'x' button to white
Screen Sizes
**Small** 
Screen Shot 2019-09-19 at 10 40 14 AM

**Medium**
Screen Shot 2019-09-19 at 10 40 00 AM

**Large**
Screen Shot 2019-09-19 at 10 39 39 AM

Refs [Pivotal ID #168273935](https://www.pivotaltracker.com/story/show/168273935)

### Checklist

### DIFF
--- a/resources/assets/components/utilities/Modal/modal.scss
+++ b/resources/assets/components/utilities/Modal/modal.scss
@@ -49,13 +49,13 @@
   cursor: pointer;
   display: block;
   font-size: 34px;
-  height: 30px;
+  height: 50px;
   line-height: 34px;
   position: absolute;
   right: 0;
   text-align: center;
   top: 0;
-  width: 30px;
+  width: 50px;
   z-index: 9999; // Ensure "x" is above all other elements, like iframes.
 
   &:focus {
--- a/resources/assets/components/utilities/Modal/modal.scss
+++ b/resources/assets/components/utilities/Modal/modal.scss
@@ -76,6 +76,6 @@
 
 .modal.-inverted {
   .modal__close {
-    color: white;
+    color: $white;
   }
 }
--- a/resources/assets/components/utilities/Modal/modal.scss
+++ b/resources/assets/components/utilities/Modal/modal.scss
@@ -44,7 +44,7 @@
 }
 
 .modal__close {
-  color: $white;
+  color: $black;
   cursor: pointer;
   display: block;
   font-size: 34px;
--- a/resources/assets/components/utilities/Modal/modal.scss
+++ b/resources/assets/components/utilities/Modal/modal.scss
@@ -44,17 +44,18 @@
 }
 
 .modal__close {
-  color: $black;
+  color: $white;
+  background-color: #731daa;
   cursor: pointer;
   display: block;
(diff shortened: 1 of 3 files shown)


weerd:
@Vmack since these are design updates can you add 3 screenshots for small, medium and large sizes and link them in the description? :)

Vmack:
@weerd thanks! Added! Let me know if this looks good.

resources/assets/components/utilities/Modal/modal.scss
```diff
@@ -44,17 +44,18 @@
 }
 
 .modal__close {
-  color: $black;
+  color: $white;
+  background-color: #731daa;
```
weerd: I don't think we should make the background purple. This will affect all modals where sometimes the background is the yellow bar and thus would have a weird purple square around the close button!

Vmack: Oh ok good point @weerd

resources/assets/components/utilities/Modal/modal.scss
```diff
@@ -44,17 +44,18 @@
...
   font-size: 34px;
-  height: 50px;
+  height: 30px;
   line-height: 34px;
   position: absolute;
   right: 0;
   text-align: center;
   top: 0;
-  width: 50px;
+  width: 30px;
```
weerd: I think we should maybe keep the size at 50px for the height and width because on mobile devices it's a nice size for a user's finger to click.

Vmack:
@weerd let me know if it's good to go!

resources/assets/components/utilities/Modal/modal.scss
weerd: @Vmack I think you still need to remove this background color rule!

resources/assets/components/utilities/Modal/modal.scss
weerd: And still restore the button to 50px!

resources/assets/components/utilities/Modal/modal.scss
```diff
@@ -44,7 +44,7 @@
 }
 
 .modal__close {
-  color: $black;
+  color: $white;
```
weerd: We want to keep this one as `$black` by default and only change to white in the rule below when using the `-inverted` class! So just switch this back to black and confirm it's all looking as intended in the badge modal 😀

resources/assets/components/utilities/Modal/modal.scss
```diff
@@ -73,3 +73,9 @@
     min-width: 600px;
   }
 }
+
+.modal.-inverted {
+  .modal__close {
+    color: white;
```
weerd: Also make sure to use the `$white` variable instead of the HTML `white` color name :)

Vmack: good catch! Thanks!

PR #1619, opened 2019-09-19
Display Referral Page Banner per feature flags

aaronschachter:
### What does this PR do?

This PR modifies the `Affirmation` logic to display the Referral Page Banner if the user has a `refer-friends` value of `true` set in the `feature_flags` field (introduced in https://github.com/DoSomething/northstar/pull/915). There are no design changes, so screenshots not necessary.

Because we need to wait for the GraphQL response to render the top Badges component and newly added bottom Referral Page banner component, it felt like it made sense to move the `Flex` and `Byline` components inside the Query, to avoid elements shifting around as the request loads. It also felt like it made sense to add the spinner to the Query, to better indicate we're waiting for a request to complete when the modal appears.

I'll update the dev `phoenix` client so this app can be tested via review app. I've turned on `DS_REFER_FLAGS_TEST` in Northstar dev, so the easiest way to test this will be to create a new account and then sign up for a campaign to see the Affirmation -- your new user will have the `refer-friends` feature flag set and should see the banner. Will open a PR in Aurora eventually to make the `refer-flags` feature flag editable to admins (so it's a lot easier to test).

Next up: restrict the campaigns the Referral Page Banner appears on to a specific list of campaign ID's: https://www.pivotaltracker.com/n/projects/2019429/stories/168489148/comments/206819678

https://www.pivotaltracker.com/story/show/168489148

### Checklist

### DIFF
--- a/resources/assets/components/Affirmation/Affirmation.js
+++ b/resources/assets/components/Affirmation/Affirmation.js
@@ -15,18 +15,13 @@ import CtaReferralPageBannerContainer from '../utilities/CtaReferralPageBanner/C
 
 import './affirmation.scss';
 
-const SIGNUP_COUNT_BADGE = gql`
-  query SignupsCountQuery($userId: String!) {
-    signupsCount(userId: $userId, limit: 2)
-  }
-`;
-
 const USER_QUERY = gql`
-  query AccountQuery($userId: String!) {
+  query UserAccountAndSignupsCountQuery($userId: String!) {
     user(id: $userId) {
       hasBadgesFlag: hasFeatureFlag(feature: "badges")
       hasReferFriendsFlag: hasFeatureFlag(feature: "refer-friends")
     }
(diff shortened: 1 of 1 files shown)


mendelB:
>  It also felt like it made sense to add the spinner to the Query, to better indicate we're waiting for a request to complete when the modal appears.

Now that some more of the core modal content loads pending the query result I think this def makes more sense to do!

One thing I noticed re the spinner which has nothing to do with this PR but could be a nice quick fix is that there's an awkward lack of space under the spinner:

Screen Shot 2019-09-19 at 2 22 47 PM

Maybe worth throwing like a `.p-4` on the [`<spinner>`](https://github.com/DoSomething/phoenix-next/blob/a0b55033ecfd1f1b8e7729d6295447ffac6178d7/resources/assets/components/Query.js#L16) for some vertical spacing?

PR #1620, opened 2019-09-19
Only display Referral Page Banner for certain campaigns

aaronschachter:
### What does this PR do?

This PR hardcodes a list of campaigns that we should display the Referral Page Banner for in an Affirmation. A future iteration of this could save the list of campaigns to a config variable, but for the sake of going live with this hopefully tomorrow -- hardcoding to keep it super simple.

Some of the complexity in using a Heroku config variable as a list of id's bring up question of sending over to React app. This could potentially be handled by adding a new property into the campaigns repository, but that can be for vLater. We may end up wanting to run this for all campaigns and removing the check completely.

https://www.pivotaltracker.com/story/show/168489148

### Checklist

### DIFF
--- a/resources/assets/components/Query.js
+++ b/resources/assets/components/Query.js
@@ -13,7 +13,7 @@ const Query = ({ query, variables, children, hideSpinner }) => (
     {result => {
       // On initial load, just display a loading spinner.
       if (result.networkStatus === NetworkStatus.LOADING) {
-        return hideSpinner ? null : <div className="spinner -centered" />;
+        return hideSpinner ? null : <div className="spinner -centered p-8" />;
       }
 
       if (result.error) {
--- a/resources/assets/components/utilities/CtaReferralPageBanner/CtaReferralPageBanner.js
+++ b/resources/assets/components/utilities/CtaReferralPageBanner/CtaReferralPageBanner.js
@@ -3,24 +3,45 @@ import PropTypes from 'prop-types';
 
 import './cta-referral-page-banner.scss';
 
-const CtaReferralPageBanner = props => (
-  <div className="p-4">
-    <div className="cta-register-banner p-4 clearfix">
+// For now, only display Referral Pages for certain campaigns.
+// @see https://www.pivotaltracker.com/n/projects/2019429/stories/168489148/comments/206819678
+const REFERRAL_CAMPAIGN_IDS = [
+  '7',
+  '2932',
+  '3271',
+  '3302',
+  '3590',
+  '7951',
+  '9026',
+  '9027',
+  '9030',
+  '9031',
+  '9032',
+];
+
+const CtaReferralPageBanner = ({ campaignId }) => (
+  <React.Fragment>
+    {REFERRAL_CAMPAIGN_IDS.includes(campaignId) ? (
       <div className="p-4">
-        <h3 className="text-white">Benefits With Friends</h3>
-        <p className="text-white padding-bottom-md">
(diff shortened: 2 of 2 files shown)


resources/assets/components/Query.js
```diff
@@ -13,7 +13,7 @@ const Query = ({ query, variables, children, hideSpinner }) => (
     {result => {
       // On initial load, just display a loading spinner.
       if (result.networkStatus === NetworkStatus.LOADING) {
-        return hideSpinner ? null : <div className="spinner -centered" />;
+        return hideSpinner ? null : <div className="spinner -centered p-8" />;
```
aaronschachter: Refs https://github.com/DoSomething/phoenix-next/pull/1619#issuecomment-533256830

mendelB: 🙌

resources/assets/components/utilities/CtaReferralPageBanner/CtaReferralPageBanner.js
```diff
@@ -3,24 +3,45 @@ import PropTypes from 'prop-types';
...
+  '7951',
+  '9026',
+  '9027',
+  '9030',
+  '9031',
+  '9032',
+];
+
+const CtaReferralPageBanner = ({ campaignId }) => (
+  <React.Fragment>
```
mendelB: I don't think you should need to wrap this in a fragment since it's rendering either content wrapped in a parent `<div>`, or `null` -- which should be fine for wherever this component is rendered.

aaronschachter: If I make this change, I think then we'd need a `return` statement somewhere, e.g.:
```
const CtaReferralPageBanner = ({ campaignId }) => {
  return REFERRAL_CAMPAIGN_IDS.includes(campaignId) ? (
      <div className="p-4">
        <div className="cta-register-banner p-4 clearfix">
          <div className="p-4">
            <h3 className="text-white">Benefits With Friends</h3>
            <p className="text-white padding-bottom-md">
              Refer a friend to this campaign, and you’ll *both* earn a $5 gift
              card!
            </p>
            <a
              href={`/us/refer-friends?campaign_id=${campaignId}`}
              className="button padded -attached"
            >
              Refer A Friend
            </a>
          </div>
        </div>
      </div>
    ) : null;
};
```

Before adding the `React.Fragment` to get the conditional inline, I had initially defined this component as:
```
const CtaReferralPageBanner = ({ campaignId }) => {
  if (!REFERRAL_CAMPAIGN_IDS.includes(campaignId) { 
   return null:
  }
  return (
      <div className="p-4">
        <div className="cta-register-banner p-4 clearfix">
          <div className="p-4">
            <h3 className="text-white">Benefits With Friends</h3>
            <p className="text-white padding-bottom-md">
              Refer a friend to this campaign, and you’ll *both* earn a $5 gift
              card!
            </p>
            <a
              href={`/us/refer-friends?campaign_id=${campaignId}`}
              className="button padded -attached"
            >
              Refer A Friend
            </a>
          </div>
        </div>
      </div>
    );
}
```
which I slightly prefer to the first.

I went with the current approach because it seemed similar to most other components (and reviews in other PR's), but open to changing this around if you think it'd be cleaner?

aaronschachter: I can clean this up in a future PR - gonna merge now to start testing on QA (where it's easier to test these campaign ID's vs dev in the review app)

PR #1621, opened 2019-09-20
Adds FAQ to Referral Pages

aaronschachter:
### What does this PR do?

This PR adds/edits copy per https://www.pivotaltracker.com/story/show/168666619/comments/206904718. I'll add this review app to the phoenix dev client to make these changes reviewable to product/design.

Per [copy guide](https://docs.google.com/document/d/1CQrceatp9cbQN9uMtkm2VYtmhZP0qXDNVHh5AUq7Bp0/edit), the official rules Contentful entry is still in draft (so will 404 for now).

https://www.pivotaltracker.com/story/show/168666619

### DIFF
--- a/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
+++ b/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
@@ -67,7 +67,11 @@ class SocialDriveAction extends React.Component {
       <div className="clearfix padding-bottom-lg">
         <div className="social-drive-action">
           <Card title={shareCardTitle} className="rounded bordered">
-            {shareCardDescription}
+            {shareCardDescription ? (
+              <div className="padded">
+                <p>{shareCardDescription}</p>
+              </div>
+            ) : null}
 
             <div className="padded">
               <Embed url={link} />
--- a/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
+++ b/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
@@ -67,11 +67,7 @@ class SocialDriveAction extends React.Component {
       <div className="clearfix padding-bottom-lg">
         <div className="social-drive-action">
           <Card title={shareCardTitle} className="rounded bordered">
-            {shareCardDescription ? (
-              <div className="padded">
-                <p>{shareCardDescription}</p>
-              </div>
-            ) : null}
+            {shareCardDescription}
 
             <div className="padded">
               <Embed url={link} />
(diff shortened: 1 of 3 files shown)


ngjo:
@aaronschachter can you remove `If you already have a DoSomething account, you can earn the $5 gift card by referring new people to DoSomething!` from the first question so it only says `Unfortunately, if another DoSomething member sends you a referral link, and you already have a DoSomething account, you won’t get the reward when you sign up for the shared campaign.`

Realizing that if they're an existing member, they won't get into the RAF experience for this first go.

aaronschachter:
Updated: https://dosomething-phoenix-de-pr-1621.herokuapp.com/us/join?user_id=55767606a59dbf3c7a8b4571

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -67,11 +67,7 @@ class SocialDriveAction extends React.Component {
       <div className="clearfix padding-bottom-lg">
         <div className="social-drive-action">
           <Card title={shareCardTitle} className="rounded bordered">
-            {shareCardDescription ? (
-              <div className="padded">
```
mendelB: What's the rationale for moving out the markup surrounding the description? Wouldn't it be more of a concern of _this_ component? (Also, the `PropTypes.string` requirement will trigger a warning since we're now feeding it a `node` instead of a `string`)

aaronschachter: Initially I misread the copy guide, and added all of the FAQ into this intro `div` -- which is how this refactoring got started. When I realized I needed to add the Alpha FAQ's at the bottom of the page instead, it still seemed legit to pass HTML here -- say if we wanted to render an image in a certain `SocialDriveAction`. I do see what you mean now (I hadn't noticed)
```
checkPropTypes.js:20 Warning: Failed prop type: Invalid prop `shareCardDescription` of type `object` supplied to `SocialDriveAction`, expected `string`.
```
Because we still only need a string, I'm fine with changing this back to the way it was.

mendelB: Ah ok, gotcha! My vote would be to indeed change it back since for now, it's still simply a description String. I _think_ if we needed more complexity in the description (for e.g. images) we could allow the component to except `children` nodes as a more elegant way of yielding to custom markup/JSX.

aaronschachter: Thanks - will update now!

PR #1628, opened 2019-10-01
Fix Referral Page share image

aaronschachter:
### What does this PR do?

This PR replaces the gift card svg with a png, aiming to [fix the Open Graph image](https://stackoverflow.com/questions/21636503/use-svg-as-ogimage) on the Beta Referral Page. 

I'll use the [Sharing Debugger](https://developers.facebook.com/tools/debug/sharing) to see if this fixes the image upon review app creation. Having a tough time trying to get my local public via ngrok.

https://www.pivotaltracker.com/n/projects/2019429/stories/168845017


aaronschachter:
Looking good!

Screen Shot 2019-10-01 at 12 12 28 PM

app/Http/Controllers/ReferralPageController.php
```diff
@@ -38,9 +38,9 @@ public function show(Request $request)
               'description' => $callToAction,
               'facebook_app_id' =>  config('services.analytics.facebook_id'),
               'image' => [
-                'url' => asset('images/gift-card.svg'),
-                'width' => '1200',
-                'height' => '1200',
+                'url' => asset('images/gift-card.png'),
+                'width' => '1368',
+                'height' => '484',
```
katiecrane: Why did the dimensions change?

weerd: Curious as well. The sizes were based on recommended sizes for Open Graph. The crops actually change from landscape on some platforms like  Facebook and Twitter but are square on platforms like Pinterest. The 1200px square is meant to account for the largest variety of platforms... There is a cropping guide I set up to help with these crops, that show the safe zones.

aaronschachter: I thought changing this to the image's actual dimensions would keep the same proportions, but I've made the change back and it seems to look the same either way in the [Sharing Debugger](https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fdosomething-phoenix-de-pr-1628.herokuapp.com%2Fus%2Fjoin%3Fuser_id%3D55767606a59dbf3c7a8b4571).

weerd: Yeah, it took me a bit to find the ideal size to fit across different social shares when I revamped how the metadata/open-graph stuff worked in Phoenix.

Seems like it being an SVG was the main culprit!

PR #1638, opened 2019-10-10
Meganav js behavior

weerd:
### What does this PR do?

This PR adds JS functionality to the `SiteNavigation` component. It handles a few scenarios that ended up being a bit of a surprise that only really highlighted themselves whilst playing around with it.

_Note: in the clips below, I activated click highlighting in Kap to better visualize when a click is happening_

**Small screen behavior**
Small screen capture

**Large screen behavior**
Large screen capture

This PR also introduces a new `<CloseButton>` component that is versatile enough to allow use in numerous situations; something we've been hoping to have for a while now. It also includes an initial suite of analytics events logging.

Additionally, it pulls the DS Logo SVG out and into its own component, that ends up outputting an `` with the SVG as the source. This will also be happening to the Search Icon SVG in upcoming PR.

Refs [Pivotal ID #168552873](https://www.pivotaltracker.com/story/show/168552873)

### Checklist


resources/assets/components/utilities/CloseButton/CloseButton.js
```diff
@@ -0,0 +1,56 @@
...
+        height: `${size}`,
+        pointerEvents: 'none',
+      }}
+    >
+      <path
+        fill={color}
+        fillRule="evenodd"
+        d="M17.24 0L20 2.76 12.76 10 20 17.24 17.24 20 10 12.76 2.76 20 0 17.24 7.24 10 0 2.76 2.76 0 10 7.24 17.24 0z"
+      />
+    </svg>
```
weerd: I ended up nesting the SVG in the component to make it easier to use props to manipulate some aspects of the SVG.

resources/assets/components/utilities/CloseButton/CloseButton.js
```diff
@@ -0,0 +1,56 @@
...
+  color,
+  dataLabel,
+  dataNoun,
+  size,
+}) => (
+  <button
+    type="button"
+    className={classnames('cursor-pointer', className)}
+    data-label={dataLabel}
+    data-noun={dataNoun}
```
DFurnes: Hmm, it feels like this is going to mean we'll be adding `dataLabel` and `dataNoun` props to a lot of components if we follow this pattern moving forward... not a blocker on merging this, but doesn't feel like the ideal solution. 🤔

weerd: Yeah, it felt alright initially when I took this approach for links, but I could see this approach within this separate component not being super ideal. That being said, we do likely want to track these things anywhere else a close button is used (ie: closing the modal). Lemme know if you have any alternative ideas! Defs not set on this approach :D

PR #1643, opened 2019-10-15
Adds Opt-In button to Marquee Template

Vmack:
### What does this PR do?

This PR adds the affiliate opt-in content to the marquee template, therefore other sponsored campaigns will have the opt-in option. The `affilitateOptIn` is passed in as a prop in the LandingPageContainer and then passed into the MarqueeTemplate as a prop from the LandingPage. 

The default color appears as white, therefore the MosaicTemplate needs to have `textColor` prop that will have a default white color while the MarqueeTemplate will be optional.

Refs [Pivotal ID #167674087](https://www.pivotaltracker.com/story/show/167674087)

### Checklist

### DIFF
--- a/resources/assets/components/AffiliateOptInToggle/AffiliateOptInToggle.js
+++ b/resources/assets/components/AffiliateOptInToggle/AffiliateOptInToggle.js
@@ -9,6 +9,7 @@ const AffiliateOptInToggle = ({
   affiliateOptInContent,
   affiliateMessagingOptIn,
   clickedOptIn,
+  textColor,
 }) => (
   <div className="form-wrapper affiliate-opt-in">
     <label className="option -checkbox" htmlFor="affiliate_opt_in">
@@ -21,7 +22,7 @@ const AffiliateOptInToggle = ({
         className="form-checkbox opt_in_checkbox"
         onClick={clickedOptIn}
       />
-      <TextContent styles={{ textColor: '#fff' }}>
+      <TextContent styles={{ textColor: `${textColor}` }}>
         {affiliateOptInContent}
       </TextContent>
     </label>
@@ -32,6 +33,9 @@ AffiliateOptInToggle.propTypes = {
   clickedOptIn: PropTypes.func.isRequired,
   affiliateMessagingOptIn: PropTypes.bool.isRequired,
   affiliateOptInContent: PropTypes.object.isRequired,
+  textColor: PropTypes.string,
 };
+
+AffiliateOptInToggle.defaultProps = { textColor: '#222' };
 
 export default AffiliateOptInToggle;
--- a/resources/assets/components/LedeBanner/templates/MosaicTemplate.js
+++ b/resources/assets/components/LedeBanner/templates/MosaicTemplate.js
@@ -24,7 +24,6 @@ const MosaicTemplate = props => {
     affiliateOptInContent,
     affiliateSponsors,
     displaySignup,
-    textColor,
     title,
     subtitle,
     blurb,
@@ -47,7 +46,7 @@ const MosaicTemplate = props => {
       {affiliateOptInContent ? (
         <AffiliateOptInToggleContainer
           affiliateOptInContent={affiliateOptInContent}
-          textColor={textColor}
+          textColor="#fff"
         />
       ) : null}
 
@@ -132,7 +131,6 @@ MosaicTemplate.propTypes = {
     url: PropTypes.string,
   }).isRequired,
   displaySignup: PropTypes.bool,
-  textColor: PropTypes.string,
   isAffiliated: PropTypes.bool.isRequired,
   subtitle: PropTypes.string.isRequired,
   title: PropTypes.string.isRequired,
@@ -146,7 +144,6 @@ MosaicTemplate.defaultProps = {
   affiliateOptInContent: null,
   blurb: null,
   displaySignup: true,
-  textColor: null,
   signupArrowContent: null,
 };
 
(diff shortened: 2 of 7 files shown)


resources/assets/components/LedeBanner/templates/MosaicTemplate.js
```diff
@@ -42,6 +43,7 @@ const MosaicTemplate = props => {
       {affiliateOptInContent ? (
         <AffiliateOptInToggleContainer
           affiliateOptInContent={affiliateOptInContent}
+          textColor="#ffffff"
```
lindsayrmaher: since we're pulling the textColor in on `props`, you should be able to pass that here rather than explicitly defining the hex color. `textColor={textColor}` I am also wondering whether the MosaicTemplate has access to `textColor` prop, but I wasn't sure where it's getting passed from!

Vmack: Thanks @lindsayrmaher! Yea I'm not completely sure but I know that textColor is getting passed from TextContent

weerd: @Vmack `TextContent` is a component used inside of `MosaicTemplate` so it's not a prop that would get passed from `TextContent` to `MosaicTemplate`. You'd _only_ pass `textColor` as part of the `styles` object prop into `TextContent`.

weerd: It would need to get passed into `MosaicTemplate` from the `LandingPage` component, not from the `TextContent`. But also not sure why we're aiming to pass in change in color. I think you'd just pass a `styles` prop to the `TextContent` in each respective template. I don't think it's something that needs to get passed into the template files.

lindsayrmaher: Oh gotcha, I see it here: https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/utilities/TextContent/TextContent.js

I don't think that TextContent is a parent component to Mosaic Template so maybe it's not getting a textColor prop passed to it. But I think if you update line 46 to reference the `textColor` rather than `#ffffff` it will at least use the prop you defined

weerd: True, but we shouldn't pass this as a prop to the templates (mosaic/marquee), it should just be a prop used wherever `TextContent` is used. Specific styles to a template shouldn't be things that get modified outside of it, otherwise not much point in having separate templates :)

lindsayrmaher: you right, you right :) I didn't read your comments til i posted mine 😎

weerd: no worries 💃 ye olde github no-auto-refresh-on-new-comments-issue, haha

Vmack: Thanks @weerd! I spoke to @mendelB last week about it because the checkbox displayed but because the text is white and the background was white, we were unable to see the text. So we wanted to make the default color for the mosaic template remains white since it already has a black background and allow the marquee template to have a optional textColor.

weerd: Yep yep, but this would be on the `AffiliateOptInToggleContainer` component, and not a text color that gets passed into the template.

So passing it into `AffiliateOptInToggleContainer` is correct, but the `textColor` in `propTypes` and `defaultProps` won't do anything, since it's not something you pass _into_ the Marquee/Mosaic templates. You need to edit the `AffiliateOptInToggleContainer` component to have it accept and take the `textColor` into account, which right now it's not doing!

resources/assets/components/AffiliateOptInToggle/AffiliateOptInToggle.js
```diff
@@ -21,7 +22,7 @@ const AffiliateOptInToggle = ({
         className="form-checkbox opt_in_checkbox"
         onClick={clickedOptIn}
       />
-      <TextContent styles={{ textColor: '#fff' }}>
+      <TextContent styles={{ textColor: `${textColor}` }}>
```
DFurnes: This doesn't need to be a template string!

```suggestion
      <TextContent styles={{ textColor }}>
```

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -27,6 +29,8 @@ const MarqueeTemplate = ({
 }) => {
   // @TODO: If this experiment is successful we should turn generating the series urls for
   // the cover image photo at different sizes into a helper function!
+
+  console.log(affiliateOptInContent);
```
DFurnes: Let's remove this before merging. 👾 

```suggestion

```

resources/assets/components/AffiliateOptInToggle/AffiliateOptInToggle.js
```diff
@@ -32,6 +33,9 @@ AffiliateOptInToggle.propTypes = {
   clickedOptIn: PropTypes.func.isRequired,
   affiliateMessagingOptIn: PropTypes.bool.isRequired,
   affiliateOptInContent: PropTypes.object.isRequired,
+  textColor: PropTypes.string,
 };
 
+AffiliateOptInToggle.defaultProps = { textColor: '#222' };
```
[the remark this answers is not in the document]
weerd: I like @mendelB's suggestion here. Maybe update in a follow-up PR @Vmack?

PR #1644, opened 2019-10-16
Meganav login and join behavior

weerd:
### What does this PR do?

This PR updates the *Login* and *Join Now* links in the `SiteNavigation` component to work as expected with the right series of URL params, etc.

...

Refs [Pivotal ID #]()

### Checklist


resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -3,16 +3,18 @@
 
 import React from 'react';
 import { get } from 'lodash';
+import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
-import CloseButton from '../CloseButton/CloseButton';
-import searchIcon from '../../../images/search-icon.svg';
-import DoSomethingLogo from '../DoSomethingLogo/DoSomethingLogo';
+import searchIcon from '../../images/search-icon.svg';
```
weerd: I have one more style related PR upcoming and will address switching this to be a component then.

Also, I'm wondering if components that are just really icons should maybe go in a different directory than the `/utilities` directory (see `ProfileIcon` below). It makes sense in utilities, but I could also see it in a different one that implies more it's just a simple icon asset. 🤔

resources/assets/helpers/auth.js
```diff
@@ -51,4 +54,24 @@ export function bindTokenRefreshEvent() {
...
+ */
+export function buildAuthRedirectUrl(options = null, actionId = null) {
+  let params = queryString.stringify(
+    withoutValueless({
+      actionId,
+      options: !isEmpty(options) ? JSON.stringify(options) : null,
+    }),
+  );
+
+  params = params ? `?${params}` : '';
```
weerd: Ok, so I _hated_ how this function would sometimes output `/authorize?` when no params were present, so I decided to change this. I always did a bit of a double take, wondering if the URL was broken and thus outputting the `?` alone, but it really was just cause there were no params to output.

Now it will just output `/authorize` with no `?` if no params present and I am finally at peace ☮️ Just wanted to provide context for the change 🍵

resources/assets/components/SiteNavigation/SiteNavigation.js
mendelB: I think that does make sense! One could argue that we've got a lot lumped into `/utilities` as-is. Maybe a new `/icons` directory? (Probably outside of `/utilities` so it's not too deeply nested...)

resources/assets/components/SiteNavigation/site-navigation.scss
```diff
@@ -211,6 +211,28 @@ $extraSmall: '(min-width: 360px)';
   }
 }
 
+.utility-nav__account-profile {
+  border-left: 1px solid $gray-300;
+
+  > a {
```
mendelB: I may have missed some discussion in the earlier PR's here, but do we prefer class names to raw tag/hierarchy selectors? (and/or is this something you mean to address in the follow-up styling PR, or is it just not an issue here?)

resources/assets/components/SiteNavigation/SiteNavigation.js
mendelB: btw is there a reason the search icon is imported as an SVG vs the `profileIcon` which is a component?

resources/assets/helpers/auth.js
mendelB: 🙌 Bless your heart.

Small fry nitpick -- Do you think it might be more 'convention-ey' to construct a `URL` which'd take care of adding/removing the `?` so that we don't need this custom logic?

e.g. 
```js
const url = new URL(`${window.location.origin}/authorize`);
url.search = params;
return url.href;
```

resources/assets/selectors/index.js
```diff
@@ -1,20 +1,20 @@
...
 
 export function getDataForNorthstar(state) {
-  return {
+  return withoutValueless({
     ...getCampaignDataForNorthstar(state),
     ...getStoryPageDataForNorthstar(state),
+
     // For 'source_details':
     contentful_id: state.campaign.id || state.page.id,
+    mode: query('mode') || null,
```
mendelB: Could it be helpful to add some 'auth' indiciation to this query param name, since it's not yet in in the context of authentication when set on the top level URL? (e.g. something like `auth-mode`?)

resources/assets/components/SiteNavigation/SiteNavigation.js
weerd: Might not have been clear; mentioned it in the first comment:

> I have one more style related PR upcoming and will address switching this to be a component then.

Planning on making it a component icon in follow-up PR.

resources/assets/helpers/auth.js
weerd: oh! Good suggestion! I always forget about `URL`. Will give this a whirl, thanks @mendelB 💃

resources/assets/components/SiteNavigation/site-navigation.scss
weerd: Yeah, I was kind of striking a balance and having difficulty coming up with appropriate class names. I think once I implement TailwindCSS in Phoenix properly, the plan is to replace most of these styles with utility classes anyhoo, so this is meant to be (ideally) temporary.

resources/assets/components/SiteNavigation/SiteNavigationContainer.js
```diff
@@ -0,0 +1,24 @@
...
+import SiteNavigation from './SiteNavigation';
+import { getDataForNorthstar } from '../../selectors';
+import { isAuthenticated } from '../../selectors/user';
+import { buildAuthRedirectUrl } from '../../helpers/auth';
+
+/**
+ * Provide state from the Redux store as props for this component.
+ */
+const mapStateToProps = state => {
+  const northStarData = getDataForNorthstar(state);
```
DFurnes: This is so nitpicky, but should this be `northstarData` since "Northstar" is a single proper noun?

resources/assets/selectors/index.js
weerd: Yeah, I could dig that. @DFurnes any thoughts on this? It would only be on the Phoenix side, but then get sent through as `mode`.

One thought though...  maybe one caveat is that it then doesn't match what's sent to Northstar and maybe that could be a point of confusion. It could be confusing if someone ends up trying to pass just a `mode` query param instead of `auth-mode` and needing to troubleshoot why it's not working. Just not sure how big of a deal that'd be... 🤔

resources/assets/components/SiteNavigation/SiteNavigation.js
DFurnes: Makes sense, and I like the idea of an `icons/` directory! I wonder if there's a good name for these sort of logic-less "visual" components? (Since maybe things that basically just string together classes onto an element, like `Button`, would also live here?)

weerd: `basics/` or `elements/` or `pieces/` or `fragments/`? 🤷‍♂

resources/assets/selectors/index.js
mendelB: That's a good point! I guess it's a question of which one is _more_ confusing to a dev? lol I can't say for sure myself.

resources/assets/components/SiteNavigation/SiteNavigation.js
mendelB: `basicElementalFragments` 

k sorry that's not helpful.

PR #1654, opened 2019-10-23
Meganav cleanup and analytics

weerd:
### What does this PR do?

This PR does a bit of cleanup and some updates which include:

- Switching analytics data from using a `data-*` attribute on the element and instead passing an object to whatever handler method is being used, which [simplifies things](https://github.com/DoSomething/phoenix-next/pull/1638#discussion_r334175670).
- Adds an analytics event for the Logo Homepage link.
- Adds an analytics event for the Meganav search form, clicking on the input and submitting a search query (including the query in the context of the analytics event payload).

🛁 

Refs [Pivotal ID #]()

### Checklist


resources/assets/components/pages/AccountPage/AccountNavigation.js
```diff
@@ -1,7 +1,15 @@
+/* eslint-disable id-length */
```
mendelB: I can't spot any unreasonably short identifiers on this page, what was it complaining about here?

resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -23,32 +23,38 @@ class SiteNavigation extends React.Component {
...
+        ...getUtmContext(),
         referrer: document.referrer,
+        ...context,
       },
       metadata: {
-        adjective: event.target.dataset.label,
-        category: 'navigation',
-        label: event.target.dataset.label,
-        noun: event.target.dataset.noun || 'nav_link',
+        adjective: get(analytics, 'label', null),
```
mendelB: 👾 This needs to be `'adjective'` right?

weerd: gah! copy-pasta 🍝 accident! thanks!

weerd:
> I think we can also get rid of the get_login_query php helper now as well?

@mendelB I was having trouble finding this helper function and it's cause it got renamed when I updated the auth stuff to `get_authorization_query()`! But yes, adding to a cleanup task to remove :)

PR #1660, opened 2019-10-25
Cause Page - Contentful content type migration script

mendelB:
### What does this PR do?

This PR adds a new Contentful migration script for the `CausePage` entity.

Some of these fields are subject to change once we start implementing, but this feels like a pretty solid start.

Refs [Pivotal ID #169097150](https://www.pivotaltracker.com/story/show/169097150)

### DIFF
--- a/contentful/content-types/cause-page.js
+++ b/contentful/content-types/cause-page.js
@@ -34,12 +34,12 @@ module.exports = function(migration) {
       },
       {
         regexp: {
-          pattern: '^causes\\/[a-z\\-]+$',
+          pattern: '^[a-z\\-]+$',
           flags: null,
         },
 
         message:
-          'Only alphanumeric and hyphen characters are allowed in slugs! Entry must begin with "causes/".',
+          'Slug should match the cause type e.g. "environment". Only alphanumeric and hyphen characters are allowed in slugs! The URL for the Cause Page will be prefixed with "causes/" e.g. "https://dosomething.org/us/causes/environment".',
       },
     ])
     .disabled(false)
--- a/contentful/content-types/cause-page.js
+++ b/contentful/content-types/cause-page.js
@@ -1,0 +1,197 @@
+module.exports = function(migration) {
+  const causePage = migration
+    .createContentType('causePage')
+    .name('Cause Page')
+    .description(
+      'A collection of content (information, campaigns, articles, reportbacks) revolving around a particular cause space',
+    )
+    .displayField('internalTitle');
+
+  causePage
+    .createField('internalTitle')
+    .name('Internal Title')
+    .type('Symbol')
+    .localized(false)
+    .required(true)
+    .validations([])
+    .disabled(false)
+    .omitted(false);
+
+  causePage
+    .createField('slug')
+    .name('Slug')
+    .type('Symbol')
+    .localized(false)
+    .required(true)
+    .validations([
+      {
+        unique: true,
+      },
+      {
+        size: {
+          max: 100,
+        },
+      },
+      {
+        regexp: {
+          pattern: '^causes\\/[a-z\\-]+$',
+          flags: null,
+        },
+
+        message:
+          'Only alphanumeric and hyphen characters are allowed in slugs! Entry must begin with "causes/".',
+      },
+    ])
+    .disabled(false)
+    .omitted(false);
+
+  causePage
+    .createField('coverImage')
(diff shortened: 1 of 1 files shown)


contentful/content-types/cause-page.js
```diff
@@ -0,0 +1,197 @@
...
+  causePage
+    .createField('superTitle')
+    .name('Super Title')
+    .type('Symbol')
+    .localized(true)
+    .required(true)
+    .validations([
+      {
+        size: {
+          max: 100,
```
mendelB: I gave this and the `title` a max size of 100 characters completely arbitrarily. @lkpttn, did you happen to have any maximum char length in mind? Perhaps 150? We can even remove the validation altogether and allow editorial judgment.

contentful/content-types/cause-page.js
```diff
@@ -0,0 +1,197 @@
...
+        unique: true,
+      },
+      {
+        size: {
+          max: 100,
+        },
+      },
+      {
+        regexp: {
+          pattern: '^causes\\/[a-z\\-]+$',
```
mendelB: I wonder if we could diverge from our usual pattern of adding the taxonomic prefix and instead just use the slug directly without the `causes/` prefix? 🤔 

It could make querying slightly less awkward with us being able to use the `/:slug` wildcard directly?

contentful/content-types/cause-page.js
mendelB: On second thought -- went ahead and just removed these validations in https://github.com/DoSomething/phoenix-next/pull/1660/commits/cbd9b114c5b2debcc0bcfe06ec2289be7568cfb6 relying on the built in 256 chars validation for Short Text

contentful/content-types/cause-page.js
```diff
@@ -0,0 +1,184 @@
...
+          width: {
+            min: 1440,
+            max: null,
+          },
+
+          height: {
+            min: 610,
+            max: null,
+          },
+        },
```
mendelB: I copied these dimensions over from [Story Page](https://github.com/DoSomething/phoenix-next/blob/cbd9b114c5b2debcc0bcfe06ec2289be7568cfb6/contentful/content-types/storyPage.js#L92-L103) since we're using a similarly large cover image banner

contentful/content-types/cause-page.js
```diff
@@ -0,0 +1,184 @@
...
+    .name('Content')
+    .type('RichText')
+    .localized(true)
+    .required(true)
+    .validations([
+      {
+        nodes: {
+          'embedded-entry-block': [
+            {
+              linkContentType: ['galleryBlock'],
```
mendelB: Limiting this to Gallery Blocks for now, since that's what's on the design spec.

contentful/content-types/cause-page.js
weerd: Potentially... the pattern was initially established because one page type handled multiple sub-types (articles, facts, about... etc). I think if the description was clear that it'd be prefixed with `causes/` I think it would be ok. Could also be an ideal way to keep us from trying to use `CausePage` for a similar type page that is not cause specific... saving us from ourselves ;)

contentful/content-types/cause-page.js
```diff
@@ -0,0 +1,184 @@
...
+    .name('Title')
+    .type('Symbol')
+    .localized(true)
+    .required(true)
+    .validations([])
+    .disabled(false)
+    .omitted(false);
+
+  causePage
+    .createField('description')
```
weerd: What is this field?

contentful/content-types/cause-page.js
mendelB: Lol nothing like a little forced self-discipline 😈
Ok, I'll go ahead and give it a go!

contentful/content-types/cause-page.js
mendelB: Exactly! Here it is in the specs (ultimate vision & MVP):
Screen Shot 2019-10-28 at 1 04 19 PM

Screen Shot 2019-10-28 at 1 02 31 PM

contentful/content-types/cause-page.js
mendelB: And updated in https://github.com/DoSomething/phoenix-next/pull/1660/commits/5ef4616d33f1288e008465b51c22368118eaeb34!

PR #1663, opened 2019-10-28
Meganav dropdown styles

weerd:
### What does this PR do?

This PR wraps up most of the styling for the Causes dropdown so it'll be ready when the cause pages launch! It also adds a new `SiteNavigationFeature` component that handles the featured item in the dropdown. It's a basic component for now and we can chat about maybe having some solid default content to show/link to if added with no props, but it's not in use yet, so we can hold off for now.

Responsive Styled Causes Dropdown

🌵 

### Checklist

* [x] Added screen-grab to PR description of related front-end updates on **small** screens.
* [x] Added screen-grab to PR description of related front-end updates on **medium** screens.
* [x] Added screen-grab to PR description of related front-end updates on **large** screens.

### DIFF
--- a/resources/assets/components/SiteNavigation/SiteNavigationFeature.js
+++ b/resources/assets/components/SiteNavigation/SiteNavigationFeature.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SiteNavigationFeature = ({
+  callback,
+  imageSrc,
+  imageAlt,
+  moreLinkText,
+  text,
+  title,
+  url,
+}) => (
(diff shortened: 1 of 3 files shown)


Vmack:
Looks great to me!

resources/assets/components/SiteNavigation/SiteNavigationFeature.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SiteNavigationFeature = ({
```
weerd: It's possible this component could become something more generic that we also use elsewhere, but felt it appropriate to contain it within the `SiteNavigation` for the time being.

PR #1664, opened 2019-10-28
Add ButtonLink component and update actions

lkpttn:
### What does this PR do?

This PR creates a new component, ButtonLink, that better handles things we want styled as a button but need to link somewhere instead of submitting a form. We also use implement the new  component in two actions

Currently this looks identical, but sets us up to better manage and style all button types in the future with Tailwind.

Refs [Pivotal ID #169045693](https://www.pivotaltracker.com/story/show/169045693)

### DIFF
--- a/resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js
+++ b/resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js
@@ -4,7 +4,7 @@ import classnames from 'classnames';
 
 import Card from '../../../utilities/Card/Card';
 import Embed from '../../../utilities/Embed/Embed';
-import { isExternal, dynamicString } from '../../../../helpers';
+import { dynamicString } from '../../../../helpers';
 import ButtonLink from '../../../utilities/ButtonLink/ButtonLink';
 import { trackAnalyticsEvent } from '../../../../helpers/analytics';
 import TextContent from '../../../utilities/TextContent/TextContent';
@@ -91,15 +91,10 @@ const DefaultTemplate = props => {
           />
         ) : null}
 
-        {/* <Button attached onClick={() => onLinkClick(href, context)}>
-          {buttonText}
-        </Button> */}
-
         <ButtonLink
           attached
           link={href}
           onClick={() => analyzeClick(href, context)}
-          targetBlank={isExternal(href)}
         >
           {buttonText}
         </ButtonLink>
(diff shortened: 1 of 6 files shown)


resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js
```diff
@@ -95,9 +91,18 @@ const DefaultTemplate = props => {
           />
         ) : null}
 
-        <Button attached onClick={() => onLinkClick(href, context)}>
+        {/* <Button attached onClick={() => onLinkClick(href, context)}>
+          {buttonText}
+        </Button> */}
```
DFurnes: We can get rid of this! 👾

resources/assets/components/actions/LinkAction/templates/DefaultTemplate.js
```diff
@@ -95,9 +91,18 @@ const DefaultTemplate = props => {
...
+
+        <ButtonLink
+          attached
+          link={href}
+          onClick={() => analyzeClick(href, context)}
+          targetBlank={isExternal(href)}
```
DFurnes: I wonder if we'd always want this for external links, actually... and if so, should we just run `isExternal` on the `link` prop inside of `ButtonLink`?

resources/assets/components/actions/VoterRegistrationAction/VoterRegistrationAction.js
```diff
@@ -61,17 +62,9 @@ const VoterRegistrationAction = props => {
...
-            onClick={handleClick}
-            target="_blank"
-            rel="noopener noreferrer"
-          >
-            Start Registration
-          </a>
-        ) : null}
+        <ButtonLink link={parsedLink} onClick={handleClick} targetBlank>
+          Start Registration
+        </ButtonLink>
```
DFurnes: So much cleaner! 😌

resources/assets/components/utilities/ButtonLink/ButtonLink.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import '../Button/button.scss';
+import './buttonlink.scss';
```
DFurnes: We can just remove this file since it's empty. 🔥

PR #1666, opened 2019-10-30
Experimenting with react-select

aaronschachter:
### What does this PR do?

This PR is a WIP, but opening for review so a review app gets created. I'm trying to get [React Select](https://react-select.com/home) to work, in hopes we can use it for the [School Finder component](https://docs.google.com/document/d/1c11vXT-nu5TGR4B8LyPPApQDYiTUcjgRMKaSP9nQ20M/edit). The library supports [populating the options via async operations](https://jedwatson.github.io/react-select/), which would be ideal for querying our `lofichools` service, as we've used for previous iterations of TFJ, (or GraphQL, if we integrate the lofi MongoDB)

It hardcodes a start of a School Finder component to appear in the dev TFJ campaign. The React Select component isn't working -- it doesn't let you select any options at all.

I've created a sandbox that has the same `Select` component working: https://codesandbox.io/s/react-codesandboxer-example-krit7.  (see the example.js file)

Opening this review app to see if maybe this is a problem only on my localhost (like how my [CSS background images die when making changes](https://github.com/DoSomething/phoenix-next/pull/1587#issuecomment-528132276)).

Using this React Select library out of the box feels like it'd definitely help speed up development of building this feature.

R&D For Tech Spec in [Pivotal](https://www.pivotaltracker.com/n/projects/2401401/stories/169294661)

### DIFF
--- a/resources/assets/components/utilities/SchoolFinder/school-finder.scss
+++ b/resources/assets/components/utilities/SchoolFinder/school-finder.scss
@@ -1,0 +1,7 @@
+@import '../../../scss/next-toolbox';
+
+.school-finder {
+  > .card.rounded {
+    overflow: visible;
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -224,7 +224,6 @@
       "version": "7.0.0",
       "resolved": "https://registry.npmjs.org/@babel/helper-module-imports/-/helper-module-imports-7.0.0.tgz",
       "integrity": "sha512-aP/hlLq01DWNEiDg4Jn23i+CXxW/owM4WpDLFUbpjxe4NS3BhLVZQ5i7E0ZrxuQ/vwekIeciyamgB1UIYxxM6A==",
-      "dev": true,
       "requires": {
         "@babel/types": "^7.0.0"
       }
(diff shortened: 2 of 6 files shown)


aaronschachter:
Added the review app to the dev phoenix client and still no luck: https://dosomething-phoenix-de-pr-1666.herokuapp.com/us/campaigns/test-teens-for-jeans/action

The console doesn't give any clues...  although the library does output a warning `[Violation] 'setTimeout' handler took 135ms` ?

aaronschachter:
Also just to clarify - this code isn't trying to fetch options via sync, it's the most basic example from the [Getting Started](https://react-select.com/home#getting-started) section.

aaronschachter:
I also removed the check for campaign 9001 locally to verify the same behavior happens on other campaigns (was thinking the Carto embed could have maybe caused conflicts)

mendelB:
Hmm I can't see why this isn't showing up 😖
FWIW, It looks like the HTML for the options _do_ render onto the page when you click the dropdown icon, but don't seem to be rendering properly, or are being hidden by some CSS 🤔 
Screen Shot 2019-10-30 at 1 42 36 PM
Same as the HTML here 
image

weerd:
Is there an option for outputting the markup as an actual `<select>` with `<options>`? It feels really gross to me that it's all a bunch of `spans` and `divs` instead of using html elements that are specifically made for this type of behavior @_@

We used `<select>` and `<options>` in the [`SelectionSubmissionAction.js`](https://github.com/DoSomething/phoenix-next/blob/abecd18eb47ea9bfddd492efb71de275514cec48/resources/assets/components/actions/SelectionSubmissionAction/SelectionSubmissionAction.js#L156-L171) as an example. There'd be less potentially finicky JS to handle showing/hiding dropdown when it's a native `<select>` HTML element that has all the expected behavior supported in all browsers out of the box.

Thoughts?

weerd:
@aaronschachter ah gotcha! Thanks for the added context; much appreciated 🙌 

I think I misunderstood that this was for the state selection part of the school finder, which I think makes sense as a `<select>`; you're right in it needing to be a different component as a filterable list where the user is typing in custom text which the `<select>` doesn't support.

This react select component does seem like a mess of nested `<div>`s but I guess it can't be avoided, although nice to know you _can_ customize the component. Maybe something for a v2 of this :)

Anyhoo, thanks for clarifying!

aaronschachter:
Ideally we'd use React Select both for state, but more importantly for selecting the school -- because it supports [fetching options from an API](https://react-select.com/async), which is what we need for the autocomplete.

I can muck around with trying to customize components, but I'm at this point -- just trying to get a working prototype vs anything ready for production. I'll see if I can get those options I have visible by hitting up some css

resources/assets/components/utilities/SchoolFinder/school-finder.scss
```diff
@@ -0,0 +1,7 @@
+@import '../../../scss/next-toolbox';
+
+.school-finder {
+  > .card.rounded {
+    overflow: visible;
```
aaronschachter: CSS was the trick! Thanks @mendelB for pointing out the menu HTML was hidden, what threw me off was it disappearing when the text input wasn't active.
Screen Shot 2019-10-30 at 2 51 09 PM

aaronschachter:
I'm going to close this for now, but still continue on this route. Markup aside, this is looking promising functionally.

PR #1671, opened 2019-10-31
Campaign info card pt 1

lindsayrmaher:
### What does this PR do?

This PR creates a new component for our campaign info card on the right side of the Hero template! The component is using a graphql query to pull the info from rogue and dynamically display the information accordingly. This still needs testing and a double checking on styling and responsiveness!

This was previously being hardcoded into the Marquee template component. PS, do we want to rename the Marquee Template file to Hero so as not to confuse ourselves?

Refs [167923472](https://www.pivotaltracker.com/story/show/167923472)

Campaign Card

### Checklist


resources/assets/components/blocks/CampaignInfoBlock/CampaignInfo.js
```diff
@@ -0,0 +1,95 @@
...
+                      awareOfUnicodeTokens: true,
+                    })
+                  : 'Evergreen'}
+              </dd>
+              {timeCommitment ? (
+                <React.Fragment>
+                  <dt>Time</dt>
+                  <dd>{timeCommitment}</dd>
+                </React.Fragment>
+              ) : null}
```
aaronschachter: I think the `defineCampaignActionInfo` function makes this code a little verbose than it needs it to be, when all it is does is return a given property in the first element of an array.
What do you think about saving the first element as a const, and just checking its properties? e.g.
```
const action = actions.length ? actions[0] : null;
```

```
{action && action->timeCommitmentLabel ? (
   <React.Fragment>
     <dt>Time</td>
     <dt>{action->timeCommitmentLabel}</dt>
  </React.Fragment>
  : null}

File: resources/assets/components/blocks/CampaignInfoBlock/CampaignInfo.js [hunk 2]
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+`;
+
+const CampaignInfo = ({ campaignId, scholarshipAmount }) => (
+  <Card className="bordered padded rounded campaign-info">
+    <h1 className="mb-4 text-m uppercase">Campaign Info</h1>
+    <dl className="clearfix">
+      <Query query={CAMPAIGN_INFO_QUERY} variables={{ campaignId }}>
+        {/* is scholarship end date always the same as campaign end date???? */}
```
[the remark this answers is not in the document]
lindsayrmaher: ok! yeah i was talking about this with Mendel yesterday and i just wanted to double check.

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfo.js
lindsayrmaher: that definitely makes sense, i think this is a good adjustment!

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfo.js
```diff
@@ -0,0 +1,91 @@
...
+          const endDate = res.campaign.endDate;
+          const actions = res.campaign.actions
+            ? res.campaign.actions.filter(
+                action => action.scholarshipEntry && action.reportback,
+              )
+            : [];
+          const action = actions.length ? actions[0] : null;
+
+          return (
+            <>
```
aaronschachter: 🐛 ?

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfo.js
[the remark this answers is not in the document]
aaronschachter: Can we remove this comment then?

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfo.js
```diff
@@ -0,0 +1,91 @@
...
+        actionLabel
+        timeCommitmentLabel
+        scholarshipEntry
+        reportback
+      }
+    }
+  }
+`;
+
+const CampaignInfo = ({ campaignId, scholarshipAmount }) => (
```
aaronschachter: I missed this before, looking at the convention used in other block components -- I'd expect this to be `CampaignInfoBlock`, with the outlier being `CampaignUpdate`. I see @weerd was the author of  `CampaignUpdate` - might have a better feel for whether to keep/drop the `Block` suffix?

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfo.js
lindsayrmaher: it's a shortcut for `<React.fragment>` ! @mendelB pointed it out to me when we were pairing.

https://reactjs.org/docs/fragments.html#short-syntax

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfo.js
lindsayrmaher: Good call, I was wondering about that as well. I had `CampaignInfoBlock` initially, fine with it either way. 

@weerd any thoughts?

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfo.js
aaronschachter: Whoa, wild! 🦁

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfo.js
weerd: Sure. I like to refer to the helpful [Phoenix Architectural Approach & Best Practices](https://github.com/orgs/DoSomething/teams/team-rocket/discussions/29) info I set up a while back when debating stuff like this. Not all components follow this because they're older but I think this is the guide to use moving forward (and to amend when we make new decisions, etc).

@aaronschachter is correct and this should be named `CampaignInfoBlock` 👍

lindsayrmaher: roger that!

PR #1677, opened 2019-11-05
Scholarship card on hero template

lindsayrmaher:
### What does this PR do?

This PR creates a new affiliate scholarship component for us to use in the new hero template!! I added logic when using the marquee template component to render the correct scholarship card, styling, etc. If there's a more elegant way to implement this check let a girl know :) 

Updated styling and an update to the graphql for the card are coming in some subsequent PRs.

This is part of ongoing work with the hero template.

Refs [168907906](https://www.pivotaltracker.com/story/show/168907906)

### Checklist

### DIFF
--- a/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockBeta.js
+++ b/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockBeta.js
@@ -32,7 +32,7 @@ const AffiliateScholarshipBlockBeta = ({
       Ready to earn scholarships for doing good? Just follow the simple
       instructions below for the chance to win. Let’s Do This!
     </p>
-    <dl className="clearfix">
+    <dl className="clearfix campaign-info__scholarship-beta">
       {scholarshipAmount ? (
         <>
           <dt className="campaign-info__scholarship-beta">Win A Scholarship</dt>
@@ -43,7 +43,7 @@ const AffiliateScholarshipBlockBeta = ({
       ) : null}
       {scholarshipDeadline ? (
         <>
-          <dt>Deadline</dt>
+          <dt className="campaign-info__scholarship-beta">Deadline</dt>
           <dd className="campaign-info__scholarship-beta-info">
             {format(scholarshipDeadline, 'MMMM do, YYYY', {
               awareOfUnicodeTokens: true,
@@ -52,7 +52,7 @@ const AffiliateScholarshipBlockBeta = ({
         </>
       ) : null}
       <>
-        <dt>Requirements</dt>
+        <dt className="campaign-info__scholarship-beta">Requirements</dt>
         <dd className="campaign-info__scholarship-beta-info">
           <ul className="margin-top-sm list -compacted">
             <li>Under 26 years old</li>
--- a/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockBeta.js
+++ b/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockBeta.js
@@ -35,7 +35,7 @@ const AffiliateScholarshipBlockBeta = ({
     <dl className="clearfix">
       {scholarshipAmount ? (
         <>
-          <dt className="campaign-info__scholarship">Win A Scholarship</dt>
+          <dt className="campaign-info__scholarship-beta">Win A Scholarship</dt>
           <dd className="campaign-info__scholarship">
             ${scholarshipAmount.toLocaleString()}
           </dd>
@@ -44,7 +44,7 @@ const AffiliateScholarshipBlockBeta = ({
       {scholarshipDeadline ? (
         <>
           <dt>Deadline</dt>
-          <dd>
+          <dd className="campaign-info__scholarship-beta-info">
             {format(scholarshipDeadline, 'MMMM do, YYYY', {
               awareOfUnicodeTokens: true,
             })}
@@ -53,11 +53,13 @@ const AffiliateScholarshipBlockBeta = ({
       ) : null}
       <>
         <dt>Requirements</dt>
-        <ul className="margin-top-sm list -compacted">
-          <li>Under 26 years old</li>
-          <li>No minimum GPA</li>
-          <li>No essay</li>
-        </ul>
+        <dd className="campaign-info__scholarship-beta-info">
+          <ul className="margin-top-sm list -compacted">
+            <li>Under 26 years old</li>
+            <li>No minimum GPA</li>
+            <li>No essay</li>
+          </ul>
+        </dd>
       </>
     </dl>
   </Card>
--- a/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockBeta.js
+++ b/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockBeta.js
@@ -7,7 +7,7 @@ import Card from '../../utilities/Card/Card';
 
 import './affiliate-scholarship-block.scss';
 
-const AffiliateScholarshipBlock = ({
+const AffiliateScholarshipBlockBeta = ({
   affiliateLogo,
   affiliateTitle,
   className,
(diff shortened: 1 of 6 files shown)


resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js
```diff
@@ -25,36 +26,50 @@ const AFFILIATE_QUERY = gql`
...
-          affiliateTitle={title}
-          affiliateLogo={logo}
-          {...props}
-        />
-      );
-    }}
-  </Query>
-);
+        return (
+          <>
```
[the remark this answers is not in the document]
lindsayrmaher: so i thought so too, but when i try to return just the bracketed ternary, i get a linting error! it doesn't seem to be able to parse out `props.isScholarshipBeta`

resources/assets/components/pages/LandingPage/LandingPage.js
```diff
@@ -65,8 +66,10 @@ const LandingPage = props => {
               coverImage={coverImage}
               endDate={endDate}
               scholarshipAmount={scholarshipAmount}
+              scholarshipDeadline={scholarshipDeadline}
               subtitle={subtitle}
               testName="Marquee Template"
+              isScholarshipBeta={isScholarshipBeta}
```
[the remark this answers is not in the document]
lindsayrmaher: if we can reduce the level of nesting to pass this down i'm all for it!

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js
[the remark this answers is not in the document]
mendelB: iiinteresting 🤔 hmm, ya know -- one nice cleanup opportunity that might be neat here once we're in this component, would be replacing our usage of the apollo `Query` with our [in-house one](https://github.com/DoSomething/phoenix-next/blob/7a940381f1f3ebe9cca933ab92c8d3dceecb1cc1/resources/assets/components/Query.js). I wonder if as a side benefit, we'd be able to simplify the `return` statement here!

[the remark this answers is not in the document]
lindsayrmaher: oh snap! i didn't even realize we weren't using it here, i'll investigate 🕵️‍♀️

lindsayrmaher:
medium:large
small

screenshots for current styling pre action type

resources/assets/components/pages/LandingPage/landing-page.scss
```diff
@@ -104,6 +104,20 @@
...
+
+  .campaign-info__scholarship-beta {
+    dd,
+    dt {
+      @include media('(min-width: 350px)') {
+        float: none;
+      }
+      padding-bottom: $half-spacing;
+    }
+  }
```
mendelB: I have a slight reservation about coupling the `AffiliateScholarshipBlock` and the `CampaignInfo` cards unless we do indeed want to consolidate them into one at some point? (Which kinda does make sense to me in a way since they're pretty much serving the same end at this point.)

Perhaps when we address this comment here https://github.com/DoSomething/phoenix-next/pull/1677/files#diff-64d86d0f790aa28ad123460f5a1ecca0R101-R102, to pull this out into its own component, we can work on some form of consolidation 🤔

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockBeta.js
```diff
@@ -0,0 +1,85 @@
...
+      />
+    ) : null}
+    <p>
+      <strong>
+        Welcome to DoSomething.org
+        {affiliateTitle ? ` via ${affiliateTitle.toUpperCase()}` : null}!
+      </strong>{' '}
+      Ready to earn scholarships for doing good? Just follow the simple
+      instructions below for the chance to win. Let’s Do This!
+    </p>
```
mendelB: Do you think it'd make sense to throw a touch of padding bottom on this message? (perhaps a new Tailwind-esque `.pb-2` class in [base.scss](https://github.com/DoSomething/phoenix-next/blob/f4e762488824085da695e74ee314001212f12d31/resources/assets/scss/base.scss#L604-L647)?) the message seems to be hugging the scholarship amount label a bit on the card.

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
-import { get } from 'lodash';
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
-import { Query } from 'react-apollo';
 
+import Query from '../../Query';
```
mendelB: 🙌 💯

resources/assets/components/pages/LandingPage/landing-page.scss
lindsayrmaher: mmmm good point! @weerd and i were talking about this yesterday as well, i think it would make sense to consolidate the components into one since they are being used in such a similar way. in the mean time it would make sense to pull this css into it's own file maybe?

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockBeta.js
lindsayrmaher: yes! i like that, i'll add before i merge

resources/assets/components/pages/LandingPage/landing-page.scss
mendelB: Ahh, great minds! And hmm, on the one hand, it's always nice to have the styling for the component within the respective components `scss` file, but on the other -- this styling seems to be connected to the `.campaign-info`s `dt, dd` rules which are defined in _this_ file. hrmmmm, maybe worth throwing it into the `affiliate-scholarship-block.scss` with a clarifying comment?

lindsayrmaher: my only struggle is that because the styling needs to be so similar for both components i'm re using `campaign-info` a lot in the scholarship-beta component. so it might just feel bulky to then repeat all those styles on the `affiliate-scholarship-block` stylesheet. 😫 @weerd any insights on this?

mendelB: Oof that's a good point. I was referring to the `.campaign-info__scholarship-beta` rules specifically, since those are unique to this component. Yeah, would love @weerds thoughts on this -- maybe worth throwing _that_ rule set into the local `scss` file with a comment addressing `.campaign-info`? (we've done this before, where we've borrowed another components styling by using the same class name but I feel like it's bit us in the foot like with [.campaign-page](https://github.com/DoSomething/phoenix-next/blob/92155e4716a0da4e086214e93645568634d08e31/resources/assets/components/pages/CampaignPage/campaign-page.scss#L3))

weerd: Yeah. Ultimately I could see this as a single component with some customizations since both are very similar, which is why I mentioned a while back possibly [generalizing this into a component](https://github.com/DoSomething/phoenix-next/blob/f4e762488824085da695e74ee314001212f12d31/resources/assets/components/pages/LandingPage/landing-page.scss#L101-L102).

Not sure what the best call is here... but given this is a "beta" it could just make sense to keep it isolated for now (copy the campaign info styles and change as needed within the beta component itself). If it comes out of beta then we can discuss universalizing it as a single component (and by then maybe we'll have Tailwind ready in Phoenix 😉). If it doesn't come out of beta and decide not to move forward with it (unlikely) we can easily just delete the component directory for the beta component with it's own stylesheet etc and less stuff to cleanup in disparate files.

mendelB:
> @lindsayrmaher Ah dang, I didn't see this change in my review, I think we'll need to get rid of this new `$campaignId` variable, since the graphQL query is going to fail without it which will cause the scholarship block to fail on production. (🤔 I'm also not sure this action query belongs in the `ScholarshipAffiliateBlockQuery` since it's related to the campaign info itself, but I guess that's part of the reason to consolidate the two!)

Oop and I missed your follow up PR! https://github.com/DoSomething/phoenix-next/pull/1684 I'll take a look!

lindsayrmaher:
just to close the loop on the discussion around consolidating the `campaignInfo` and `scholarship` components: @lkpttn is looking into pulling the scholarship info out of this card completely, so we aren't going to go forward with consolidating in the meantime.

PR #1680, opened 2019-11-05
Make the Gallery Block GraphQL Query-able!

mendelB:
### What does this PR do?

This PR... looks scary but really is pretty straightforward!!

We're setting up the `GalleryBlock` component/entity to be query-able via GraphQL. In order to do that, we need to update the formatting of the fields, make some things 'Showcasable', and of course add the query. I'll explain:

- https://github.com/DoSomething/phoenix-next/commit/6fdb39bcd58f754b1dda1a779136dbe30ff45053 updates the various gallery template nodes to use the standardized 'Showcasable' fields. For more context, check out this PR in GraphQL https://github.com/DoSomething/graphql/pull/120 where we introduced this interface. TL;DR: This gives us a convenient way to think about showcasing contentful entries like Campaigns, Pages, and the like in a synchronized consistent manner.
- The issue is, we still need to support the ole field names, since when querying via the Phoenix backend for most Pages, we won't have these resolved showcasable fields! So https://github.com/DoSomething/phoenix-next/commit/75a891a20c4100b48acdbb9b730cf6d0f7c02d92 does the job of finessing support for both GraphQL formatted, and legacy formatted JSON.
 - Then of course, we need the actual GraphQL query! https://github.com/DoSomething/phoenix-next/commit/7f238f1c16dd488d74dd11bbe288f65cd610e5b8, https://github.com/DoSomething/phoenix-next/commit/a649f127371ca8805ff2a061d069977e5ae89f36, and https://github.com/DoSomething/phoenix-next/commit/938e472226f56106b9f4142e3ff3e1a4eaca73e7 do the job of defining and exporting a `GalleryBlockFragment` from the `GalleryBlock`, and adding it to the list in the `ContentfulEntryLoader`. You might notice some specifically queried fields for things like a PersonBlock within the galleries `blocks`. This is because our gallery template nodes expect some additional specific fields for certain types (`twitterId` for `PersonBlocks` of type `staff` for example).
 - Then, two quick misc improvements: https://github.com/DoSomething/phoenix-next/commit/d48a71eaf37516a6479e4e2e9af8a32236e670b1 to safely retrieve contentful image URLs when the image is not required on the Entity, so we don't see any annoying explosions (mostly an editor/developer convenience). And https://github.com/DoSomething/phoenix-next/commit/f5b516e4a4c6405d3a5b984c53ef73d325268e28 to add better `alt` descriptions for our person gallery node templates, inspired by the [`Byine`](https://github.com/DoSomething/phoenix-next/blob/f4e762488824085da695e74ee314001212f12d31/resources/assets/components/utilities/Byline/Byline.js#L21).

That's it! I told you, this is way less scary than it seems, it just is bloated mostly because of updated prop naming. and misc improvements.

This is towards the effort to implement Cause Pages in Phoenix, wherein we'll be querying RichText field embedded GalleryBlocks via GraphQL.

A major major hat tip to @jocoso, since he did most of the leg work for this in #1522 ...and oh goody! This actually closes #1522  Thank you Joshua!! Your work has been invaluable towards this initiative 🙌

Some of the queried fields (`itemsPerRow` for GalleryBlocks, `twitterId` for PersonBlocks) rely on an open PR in GraphQL land https://github.com/DoSomething/graphql/pull/151

Refs [Pivotal ID #168182024](https://www.pivotaltracker.com/story/show/168182024) and [#168182083](https://www.pivotaltracker.com/story/show/168182083)

### DIFF
--- a/resources/assets/components/Affirmation/Affirmation.js
+++ b/resources/assets/components/Affirmation/Affirmation.js
@@ -80,7 +80,7 @@ const Affirmation = ({
               author={author.fields.name}
               {...withoutNulls(author.fields)}
               photo={contentfulImageUrl(
-                author.fields.photo.url,
+                get(author, 'fields.photo.url'),
                 175,
                 175,
                 'fill',
--- a/resources/assets/components/ContentfulEntry/ContentfulEntry.js
+++ b/resources/assets/components/ContentfulEntry/ContentfulEntry.js
@@ -165,6 +165,9 @@ class ContentfulEntry extends React.Component<Props, State> {
       case 'galleryBlock':
         return <GalleryBlock {...json.fields} />;
 
+      case 'GalleryBlock':
+        return <GalleryBlock {...withoutNulls(json)} />;
+
       case 'imagesBlock':
         return (
           <ImagesBlock className={className} images={json.fields.images} />
--- a/resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
+++ b/resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import { withoutNulls } from '../../../helpers';
 import Person from '../../utilities/Person/Person';
 import Gallery from '../../utilities/Gallery/Gallery';
 import PageGalleryItem from '../../utilities/Gallery/templates/PageGalleryItem/PageGalleryItem';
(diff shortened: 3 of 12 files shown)


katiecrane:
gah @mendelB sorry I ended up commenting on the commits instead of the overall changes and sorry GitHub makes that sort of annoying sometimes!

mendelB:
> gah @mendelB sorry I ended up commenting on the commits instead of the overall changes and sorry GitHub makes that sort of annoying sometimes!

NP! That always trips me up when I'm commit-commit reviewing a PR, and it's awkward to switch back and forth and put the comment in the main file diff view 😖 

This was not inconvenient at all though!

PR #1684, opened 2019-11-06
Add ActionType to Scholarship Page

lindsayrmaher:
### What does this PR do?

This PR adds the action type field to our scholarship campaign info card! I also did a little clean up on the marquee template and campaignInfoBlock component. 

We're using a query I added in the last sprint here to pull the actions information!

Refs [168907906](https://www.pivotaltracker.com/story/show/168907906)

[Large/Medium Screen Card](https://user-images.githubusercontent.com/15236023/68334251-39a76980-00a8-11ea-8fab-1a0994ee3b47.png)

### Checklist


resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -21,13 +21,15 @@ const MarqueeTemplate = ({
...
   affiliateOptInContent,
+  campaignId,
   content,
   coverImage,
   scholarshipAmount,
   scholarshipDeadline,
   subtitle,
   title,
 }) => {
+  const numCampaignId = parseInt(campaignId, 10);
```
mendelB: Small fry -- would `Number(campaignId)` be a slightly simpler way to get an integer here?

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js
```diff
@@ -41,23 +41,24 @@ const AffiliateScholarshipBlockQuery = props => (
...
-      //   ? res.campaign.actions.filter(
-      //       action => action.scholarshipEntry && action.reportback,
-      //     )
-      //   : [];
-      // const action = actions.length ? actions[0] : null;
+      const actions = res.actions
+        ? res.actions.filter(
+            action => action.scholarshipEntry && action.reportback,
+          )
+        : [];
```
mendelB: This is also arbitrary, but It might be to do away with the ternary here by using lodash `get` to just default to an Array `get(res, 'actions', [])` and calling `filter` on it?

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js
```diff
@@ -41,23 +41,24 @@ const AffiliateScholarshipBlockQuery = props => (
...
+      const action = actions.length ? actions[0] : null;
```
mendelB: I think we can do away with this ternary logic here -- If `actions` is an empty array, `actions[0]` would return `undefined` which I think is fine for the same price as the `null` default value here?

resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js
```diff
@@ -41,23 +41,24 @@ const AffiliateScholarshipBlockQuery = props => (
...
 
       return props.isScholarshipBeta ? (
         <AffiliateScholarshipBlockBeta
           affiliateTitle={title}
           affiliateLogo={logo}
+          actionType={action.actionLabel}
```
mendelB: Beyond all my finicking above 🙃 -- this might actually be a very good place for a lodash `get` (`get(action, 'actionLabel')`) since otherwise, if we don't have a valid filtered action returned from graphql, this will error out when we try to access the `actionLabel` field

lindsayrmaher: oh good call! i think with the other changes you suggested this will also be a cleaner way to write this.

package.json
```diff
@@ -59,98 +59,98 @@
...
   },
   "devDependencies": {
-    "@babel/core": "^7.4.5",
+    "@babel/core": "^7.7.2",
     "@babel/preset-flow": "^7.0.0",
     "@dosomething/babel-preset": "^3.2.0",
     "@dosomething/eslint-config": "^5.0.1",
     "@dosomething/webpack-config": "^5.2.0",
-    "apollo": "^2.13.1",
+    "apollo": "^2.21.0",
```
weerd: Howdy! Was this change intentional or from running `npm update` instead of maybe `npm install`? My tailwind PR had a few merge conflicts I believe from the package update changes here and in the lock file. Just want to make sure this was intended!

lindsayrmaher: i was having issues earlier today with circleci and i did run `npm update` locally. 

https://dosomething.slack.com/archives/C3ASB4204/p1573143515139700

weerd: Ah gotcha. I think we usually want to be a bit more weary on running `npm update` and better to just first try and delete `node_modules` and rerunning `npm install` if having issues. The reason being that `npm update` could potentially introduce some patches in a package we're not ready for or surprise us; at least I think that's the reasoning right @DFurnes @mendelB?

weerd: That's not to say we _never_ run `npm update` but I remember in the past being more conservative when running it, haha

lindsayrmaher: that's totally fair! i did also delete my `node_modules` and re install, but i'm not sure if i did that before or after i ran the `update` should we revert back to the older version of apollo?

weerd: ah, no worries! Sounds like you did all the right steps though in trouble-shooting. I think I was just surprised by the number of updated packages.

In hindsight maybe we _should_ do `npm update` more often, so that when we do run it it's not a ton of updates, haha.

DFurnes: Nothing wrong with running `npm update` in my opinion! It's good to keep things up to date. 😇

One gotcha with `npm update` that we've dealt with in the past is that it's actually a little too conservative, and won't update sub-dependencies unless it absolutely has to. So, sometimes I'll run a `rm -rf node_modules package-lock.json && npm install` to get those all up to the newest versions that fit their constraints.

PR #1685, opened 2019-11-07
Adding initial setup for Tailwind CSS.

weerd:
### What does this PR do?

This PR updates Phoenix to prepare it for the use of Tailwind CSS framework. It updates [@dosomething/webpack-config](https://github.com/DoSomething/webpack-config/releases/tag/v6.0.0) package to v6.0.0 which removed configuration of PostCSS from the package to allow projects to define their own PostCSS configuration.

It also adds a few other packages, including `tailwindcss` and also sets up the initial Tailwind CSS configuration file (copied over from [Northstar](https://github.com/DoSomething/northstar/blob/master/tailwind.config.js), with some edits).

None of the Tailwind classes are getting output yet, so no styles on Phoenix are affected!

I plan on enabling their output in upcoming PRs and as I start to replace old Phoenix classes with the newly generated Tailwind ones, but wanted to keep this setup PR small to just the basics!


tailwind.config.js
```diff
@@ -0,0 +1,122 @@
...
+      gray: {
+        '100': '#f6f6f6',
+        '200': '#d7d7d7',
+        '300': '#c8c8c8',
+        '400': '#b9b9b9',
+        '500': '#a9a9a9',
+        '600': '#9a9a9a',
+        '700': '#8a8a8a',
+        '800': '#7b7b7b',
+        '900': '#626262',
```
weerd: @lkpttn added the new grays you defined.

tailwind.config.js
```diff
@@ -0,0 +1,122 @@
...
+      blue: {
+        '100': '#5145ff',
+        '200': '#4a3ef7',
+        '300': '#4337de',
+        '400': '#3b31c4',
+        '500': '#332baa',
+        '600': '#2c2491',
+        '700': '#241e78',
+        '800': '#1c185e',
+        '900': '#151145',
```
weerd: @lkpttn if you have a newly defined `500` value for blue, please let me know along with the rest of the `100 - 900` values if they need to change as well.

tailwind.config.js
```diff
@@ -0,0 +1,122 @@
...
+    fontSize: {
+      xs: '0.5rem',
+      sm: '0.75rem',
+      base: '1.125rem',
+      lg: '1.375rem',
+      xl: '1.688rem',
+      '2xl': '2.531rem',
+      '3xl': '3.797rem',
+      '4xl': '5.695rem',
+      '5xl': '8.543rem',
```
weerd: @lkpttn do these all look correct? Should be based on that conversation we had a while back when implementing Tailwind in Northstar but would like to double-check!

tailwind.config.js
```diff
@@ -0,0 +1,122 @@
+const { colors } = require('tailwindcss/defaultTheme');
+
+module.exports = {
+  theme: {
+    screens: {
+      xsm: '360px',
```
DFurnes: Should this be `xs` for consistency with Tailwind's built-in "extra small" classes (`.max-w-xs` and `.text-xs`)?

tailwind.config.js
```diff
@@ -0,0 +1,122 @@
...
+      sm: '480px',
+      md: '760px',
+      lg: '960px',
+      xl: '1060px',
+      xxl: '1280px',
+    },
+    colors: {
```
DFurnes: Instead of replacing the entire `colors` array (and having to re-enable `black: colors.black`, etc), would it make sense to extend it below & specify just the ones we want to change to fit our brand?

tailwind.config.js
weerd: oh true! Good catch! I copied this over from NS where it was edited to add this, so we'll need to update there too. Once this all lives in Forge it'll be easier to manage 😬

tailwind.config.js
weerd: When developing the Tailwind config for Northstar I considered this. There's a difference between overriding (which is what we're doing here) versus extending. 

When overriding the colors, Tailwind will _only_ output classes based on our limited series of defined colors; which would just be our specified brand colors. If we extended and add our colors, then Tailwind will still output classes with the all the colors it defines by default (including the ones we override).

Granted, with a tool like Purge CSS we could end up removing all those classes that don't get used because they're not part of the DS brand palette, BUT it still leaves the opportunity for a developer to potentially use the color as a class and thus Purge CSS does not remove it and we introduce colors into designs that shouldn't be used.

Ultimately I felt overriding and re-specifying black, white, and gray in the single config file is better than including all the colors, many of which we won't be using (accidental usage or otherwise).

If we end up deciding to add our own values for some of those other colors we could always revise the approach (I could see potentially needing some reds for alerts/errors on an as needed basis).

DFurnes: Thanks for explaining the difference between overriding and extending!

The reason I suggested using `extend` was because, when experimenting with this in Rogue the other day, I got tripped up on why `.border-green` and `.border-red` weren't working for me even though I was using the classes from the docs. It took me a while to figure out that we had entirely removed them by overriding the entire array here.

I definitely see the value in "banning" the use of non-brand colors, though, so let's continue with this approach and see how often we end up needing these in practice.

weerd: Ah gotcha. Yeah, I think part of the process with how developers use Tailwind is to also reference the config file, or be familiar with it to know how _we_ use it. If it looks like we need green and red, etc for buttons in Rogue and elsewhere, extending might be the better option :) Just hadn't encountered that in Northstar so it made sense at the time!

tailwind.config.js
lkpttn: Here's the blue values:

100 - #86DFFF
200 - #6BD6FF
300 - #52CCFF
400 - #3AC2FF
500 - #23B7FB
600 - #1A9ADF
700 - #127FBF
800 - #0C659F
900 - #074C80

Blue

lkpttn:
Also some smaller scales for red and green to be used in system messaging.

Red 300 - #FF6D69
Red 500 - #FF4540
Red 700 - #BF3929

Red

Green 300 - #42D86F
Green 500 - #0CC242
Green 700 - #02983A

Green

PR #1691, opened 2019-11-08
Cause page style updates

lkpttn:
### What does this PR do?

This PR updates the Cause Page to prepare for taxonomy and collection pages. There's a couple set of changes:

**Banner**
- Update the banner to a new style
- Set the banner on the grid

**Galleries**
- All galleries use the `SectionHeader` component for titles now
- `CampaignGalleryItem` now use the card component and style

**Downstream effects**
- `SectionHeader` components now use the single underline style everywhere, with a new prop
- `ContentBlock` uses this new prop, affecting campaign pages
- Some new tailwindy css classes are available

Refs [Pivotal ID #169325037](https://www.pivotaltracker.com/story/show/169325037)

Cause Page S
Cause Page M
Cause Page L

### DIFF
--- a/resources/assets/components/pages/CausePage/CausePage.js
+++ b/resources/assets/components/pages/CausePage/CausePage.js
@@ -26,7 +26,7 @@ const CausePage = ({ coverImage, superTitle, title, description, content }) => {
           className="lede-banner base-12-grid"
           style={withoutNulls(styles)}
         >
-          <div className="title-lockup pt-8">
+          <div className="title-lockup my-8">
             <h2 className="my-4 uppercase color-white text-lg">{superTitle}</h2>
             <h1 className="lede-banner__headline-title my-4 font-normal font-secondary color-white caps-lock">
               {title}
--- a/resources/assets/components/pages/CausePage/CausePage.js
+++ b/resources/assets/components/pages/CausePage/CausePage.js
@@ -32,10 +32,12 @@ const CausePage = ({ coverImage, superTitle, title, description, content }) => {
               {title}
             </h1>
             <TextContent styles={{ textColor: '#FFF' }}>
+              {/* @TODO: update font size of discription */}
               {description}
             </TextContent>
           </div>
         </header>
+        {/* @TODO: fix gallery header/title underline*/}
         <TextContent
           className="base-12-grid"
           classNameByEntry={{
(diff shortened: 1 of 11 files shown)


resources/assets/components/pages/CausePage/CausePage.js
```diff
@@ -2,22 +2,57 @@ import React from 'react';
...
-    <TextContent>{description}</TextContent>
+import './cause-page.scss';
 
-    <TextContent>{content}</TextContent>
-  </div>
-);
+const CausePage = ({ coverImage, superTitle, title, description, content }) => {
+  const backgroundImage = coverImage
+    ? `url(${contentfulImageUrl(coverImage.url, '1440', '610', 'fill')})`
+    : null;
```
mendelB: Can we add a `@TODO` here to add some logic to set the image dimensions relative to screen size somewhat akin to what we do in `MarqueeTemplate`?
https://github.com/DoSomething/phoenix-next/blob/092dba41b721b4630738245750d1dd19589ab61b/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js#L35-L39

Obviously we can't exactly emulate the `srcSet`, but at least, if we set one static size initially dynamically based on the screen width, we won't be making a huge request for mobile screens!
Also this looks cool potentially for the future, for `srcSet` on background images! cc @weerd  https://aclaes.com/responsive-background-images-with-srcset-and-sizes/
(another idea is to somehow integrate the `<Media>` utility!) https://github.com/DoSomething/phoenix-next/blob/092dba41b721b4630738245750d1dd19589ab61b/resources/assets/components/utilities/IframeEmbed/IframeEmbed.js#L19

lkpttn: Done!

resources/assets/components/pages/CausePage/cause-page.scss
```diff
@@ -0,0 +1,30 @@
...
+  .lede-banner {
+    background-position: center center;
+    background-repeat: no-repeat;
+    background-size: cover;
+
+    > .title-lockup {
+      grid-column: 1 / -1;
+
+      @include media($medium) {
+        grid-column: 2 / -2;
```
lindsayrmaher: Are we dividing by negative numbers here or is this a syntax thing to replicate TW?

lkpttn: Grid column lets you measure forward from say 1..10 and backwards like -1..-10. When we span an area from 1 to -1, we're actually saying go from the first column to the last column, regardless of how many columns are in between them.

PR #1682, opened 2019-11-06
Beta Referral Page tests

aaronschachter:
###  What does this PR do?

This PR adds some minimal testing for the Beta Referral Page, checking for expected content per query string parameters. Also adds a `constants` fixture to DRY hardcoding real user and campaign ID's to avoid getting 404 responses back from our dev API's, per chat in today's dev round table.

### Any background context you want to provide?

Future PR to add mocks and tests for verifying referring user's first name, campaign information, etc

### What are the relevant tickets/cards?

https://www.pivotaltracker.com/n/projects/2401401/stories/169604403

### Checklist

* [x] Added appropriate feature/unit tests.

PR #1694, opened 2019-11-09
School Finder + GraphQL schools

aaronschachter:
### What does this PR do?

This PR continues development on the School Finder by querying GraphQL for school information (per https://github.com/DoSomething/graphql/pull/152 and https://github.com/DoSomething/graphql/pull/154) instead of the legacy [lofi-schools](https://github.com/DoSomething/lofi-schools-flask) app.

Still a few features left to add before this is ready for screenshots/visual review:

* Adding  "My school isn't available" checkbox + saving `school-not-available` to user's `school_id` field 
* Modal for searching tips

Writing Cypress tests for school finder is a bit tricky. Our best bet could be to create a new user for each test, to avoid needing to hardcore user ID's (and continuously unset / set their `school_id` values)

* https://www.pivotaltracker.com/story/show/169549547
* https://www.pivotaltracker.com/story/show/169549488

### Checklist

### DIFF
--- a/resources/assets/components/utilities/SchoolFinder/SchoolFinder.js
+++ b/resources/assets/components/utilities/SchoolFinder/SchoolFinder.js
@@ -41,15 +41,13 @@ const SchoolFinder = ({ campaignId, userId }) => {
     <div className="school-finder margin-bottom-lg margin-horizontal-md clear-both primary">
       <Query query={USER_SCHOOL_QUERY} variables={{ userId }}>
         {res => {
-          const schoolId = res.user.schoolId;
-
-          if (schoolId) {
+          if (res.user && res.user.school) {
             setSchool(res.user.school);
           }
 
           return (
             <Card
-              title={schoolId ? 'Your School' : 'Find Your School'}
+              title={school.id ? 'Your School' : 'Find Your School'}
               className="rounded bordered overflow-visible"
             >
               {school.id ? (
--- a/resources/assets/components/utilities/SchoolFinder/SchoolFinder.js
+++ b/resources/assets/components/utilities/SchoolFinder/SchoolFinder.js
@@ -1,5 +1,4 @@
-import React from 'react';
-import { get } from 'lodash';
+import React, { useState } from 'react';
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
 
@@ -11,6 +10,7 @@ import CurrentSchool from './CurrentSchool';
 const USER_SCHOOL_QUERY = gql`
   query UserSchoolQuery($userId: String!) {
     user(id: $userId) {
+      id
       schoolId
       school {
         id
@@ -29,25 +29,41 @@ const SchoolFinder = ({ campaignId, userId }) => {
     return null;
   }
 
+  // TODO: We shouldn't need a school state variable, but instead update Apollo Cache.
+  const [school, setSchool] = useState({
+    id: null,
+    name: null,
+    city: null,
+    state: null,
+  });
+
   return (
     <div className="school-finder margin-bottom-lg margin-horizontal-md clear-both primary">
       <Query query={USER_SCHOOL_QUERY} variables={{ userId }}>
-        {res => (
-          <Card
-            title={res.user.schoolId ? 'Your School' : 'Find Your School'}
-            className="rounded bordered overflow-visible"
-          >
-            {res.user.schoolId ? (
-              <CurrentSchool
-                name={get(res, 'user.school.name', null)}
-                city={get(res, 'user.school.city', null)}
-                state={get(res, 'user.school.state', null)}
-              />
-            ) : (
-              <UpdateSchool userId={userId} />
-            )}
-          </Card>
-        )}
+        {res => {
+          const schoolId = res.user.schoolId;
+
+          if (schoolId) {
+            setSchool(res.user.school);
+          }
+
+          return (
+            <Card
+              title={schoolId ? 'Your School' : 'Find Your School'}
+              className="rounded bordered overflow-visible"
+            >
+              {school.id ? (
+                <CurrentSchool school={school} />
+              ) : (
+                // TODO: Remove onSubmit once Apollo cache is updated via mutation in UpdateSchool.
+                <UpdateSchool
+                  userId={userId}
+                  onSubmit={selected => setSchool(selected)}
+                />
+              )}
+            </Card>
+          );
+        }}
       </Query>
     </div>
   );
--- a/resources/assets/components/utilities/SchoolFinder/SchoolFinder.js
+++ b/resources/assets/components/utilities/SchoolFinder/SchoolFinder.js
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import { get } from 'lodash';
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
 
@@ -11,6 +12,12 @@ const USER_SCHOOL_QUERY = gql`
   query UserSchoolQuery($userId: String!) {
     user(id: $userId) {
       schoolId
+      school {
+        id
+        name
+        city
+        state
+      }
     }
   }
 `;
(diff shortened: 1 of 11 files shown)


aaronschachter:
This was working locally, but searching for schools in the review app (I've added it to the dev `pheonix` client) is triggering a 405 error. Will take a look on Monday with fresh 👀 🍃 

Screen Shot 2019-11-08 at 4 19 54 PM

resources/assets/components/utilities/SchoolFinder/SchoolSelect.js
```diff
@@ -1,17 +1,46 @@
...
+`;
+
+/**
+ * Ideally would call useApolloClient() but it triggers an error Invariant Violation
+ * No Apollo Client instance.
+ *
+ * @see https://stackoverflow.com/a/57743861
+ */
+const url = env('GRAPHQL_URL');
+console.log('debugging GraphQL URL:', url);
```
aaronschachter: Refs https://github.com/DoSomething/phoenix-next/pull/1694#issuecomment-552041185:

Localhost: `debugging GraphQL URL: https://graphql-dev.dosomething.org/graphql`

Review app: `debugging GraphQL URL: undefined`

DFurnes: That's so strange... I see `GRAPHQL_URL` set on that review app (and [in our `app.json`](https://github.com/DoSomething/phoenix-next/blob/7551491a1dbbbb9c5db2b9d118de31d67f05f054/app.json#L37-L39)), so it's weird that it's coming up `undefined`...

resources/assets/components/utilities/SchoolFinder/SchoolFinder.js
```diff
@@ -21,25 +28,42 @@ const SchoolFinder = ({ campaignId, userId }) => {
   if (campaignId !== '9001') {
     return null;
   }
-  const [currentSchool, setCurrentSchool] = useState(null);
+
+  const [schoolName, setSchoolName] = useState(null);
+  const [schoolCity, setSchoolCity] = useState(null);
+  const [schoolState, setSchoolState] = useState(null);
```
DFurnes: Is there a situation where these are set independently? If not, I wonder if it makes sense to merge into an object:

```js
const [school, setSchool] = useState({ name: null, city: null, state: null});

// and below:
if (get(res, 'user.school.id')) {
  const { name, city, state } = res.user.school;
  setSchool({name, city, state});
}

<UpdateSchool
  onSubmit={(name, city, state) => setSchool({ name, city, state });
```

resources/assets/components/utilities/SchoolFinder/SchoolFinder.js
```diff
@@ -21,25 +28,42 @@ const SchoolFinder = ({ campaignId, userId }) => {
...
 
   return (
     <div className="school-finder margin-bottom-lg margin-horizontal-md clear-both primary">
       <Query query={USER_SCHOOL_QUERY} variables={{ userId }}>
         {res => {
-          if (res.user && res.user.schoolId) {
-            setCurrentSchool(res.user.schoolId);
+          if (get(res, 'user.school.id')) {
+            const { name, city, state } = res.user.school;
```
DFurnes: This seems a little awkward to me (querying for the school from GraphQL, and then using that to set state on the component). We should be able to let Apollo automatically keep these in sync for us if we just use the query result here & then let the mutation result update the shared cache if someone updates their school.

resources/assets/components/utilities/SchoolFinder/SchoolSelect.js
DFurnes: Oh, I think this is coming up `undefined` because this is outside the component's render method (and so it runs immediately when the script is resolved, and possibly before `window.ENV` is defined in the DOM).

resources/assets/components/utilities/SchoolFinder/SchoolSelect.js
```diff
@@ -1,17 +1,46 @@
...
+const client = graphqlClient(url);
+
 const SchoolSelect = ({ filterByState, onChange }) => {
   // Debounce school search to query for schools after 250 ms typing pause.
   // @see https://github.com/JedWatson/react-select/issues/614#issuecomment-244006496
   const debouncedFetch = debounce((searchString, callback) => {
```
DFurnes: Instead of manually creating a client here, can we use [`useLazyQuery`](https://www.apollographql.com/docs/react/data/queries/#executing-queries-manually)?

aaronschachter: We'll need to add [`react-hooks`](https://www.apollographql.com/docs/react/api/react-hooks/), which I did try adding to get the [`useApolloClient`](https://www.apollographql.com/docs/react/api/react-hooks/#useapolloclient) hook working -- but ended up with a `Invariant Violation No Apollo Client instance` error.  Per this [SO thread](https://stackoverflow.com/questions/57743313/no-apollo-client-instance-can-be-found/57743861#57743861) -- it sounds like it could be resolved by updating a few other libraries, but didn't want to go too breaky if it could be avoided.

resources/assets/components/utilities/SchoolFinder/SchoolSelect.js
```diff
@@ -1,17 +1,46 @@
...
+      state
+    }
+  }
+`;
+
+/**
+ * Ideally would call useApolloClient() but it triggers an error Invariant Violation
+ * No Apollo Client instance.
+ *
+ * @see https://stackoverflow.com/a/57743861
```
aaronschachter: Refs https://github.com/DoSomething/phoenix-next/pull/1694#discussion_r344814721

resources/assets/components/utilities/SchoolFinder/SchoolSelect.js
aaronschachter: That was it, thanks. I've moved this into the render, which at least gets this functional... but do you think it's worth updating packages to get [`useApolloClient`](https://www.apollographql.com/docs/react/api/react-hooks/#useapolloclient) working?

resources/assets/components/utilities/SchoolFinder/SchoolFinder.js
aaronschachter: I've updated the branch making this change -- the user's `school_id` does get updated in Northstar, but the `SchoolFinder` component doesn't re-render with the `UpdateSchool` component upon the save... is there something I might be missing?

resources/assets/components/utilities/SchoolFinder/SchoolSelect.js
aaronschachter: Made this change in #1697

resources/assets/components/utilities/SchoolFinder/SchoolSelect.js
aaronschachter: I don't think  `useLazyQuery`  will place nicely with the async options in [React Select](https://react-select.com/async), which expects a Promise. The Apollo Client `query` does return a Promise, which is why this is working currently (per updates to latest `apollo-client` in #1697)

aaronschachter:
This is ready for review. I left TODOs for rabbit holes I went on, and ended up still not resolving:

* Getting the Apollo client's cache to update locally instead of using a `school` state in the `SchoolFinder`.  The [single entity cache](https://www.apollographql.com/docs/react/data/mutations/#updating-a-single-existing-entity) wasn't working as is, and couldn't get the update function working in the `getMutation` call. Would prefer to get these changes in first and investigate post-merge/deploy.

* Fixing the intermittent Beta Referral Page for valid user + no campaign Cypress test. It looks like the GraphQL request for the Embed hangs, so the `a` is never rendered. Instead of commenting out the cursed test (done in #1686), I've moved it to be called at the end of the file. It's maybe questionable to check for the existence of the query string at all, since we check for it in the 1st test... but you'll see by recent commits - both in this branch, in #1686, and in other PR's post merging #1682 -- that it's a bit of a mystery for when it decides to fail.

aaronschachter:
Tested on https://dosomething-phoenix-de-pr-1694.herokuapp.com/us/campaigns/test-teens-for-jeans/action and confirmed school is saved (for now, we need to clear our user's school ID in Aurora to re-test -- as we're not building out a way to do this via Phoenix/Northstar yet)

PR #1699, opened 2019-11-11
Cause Page - set grid columns for Content blocks

mendelB:
### What does this PR do?

This PR sets the `grid-column` styling for Content Blocks in Cause Page content. 
I couldn't figure out where this custom grid styling fit in to our existing scheme, so I scoped it to the `.cause-page` styling. I also couldn't think of a better class name than `content-block-grid` because I'm a creative 😢

Refs [Pivotal ID #169325037](https://www.pivotaltracker.com/story/show/169325037)
[Slack discussion](https://dosomething.slack.com/archives/C3ASB4204/p1573499042209300)
### Checklist

📸 
- [Large](https://user-images.githubusercontent.com/12417657/68619125-d139fd00-0498-11ea-928a-84ea49a7f2c5.png)
- [Medium](https://user-images.githubusercontent.com/12417657/68619160-e44ccd00-0498-11ea-97b5-b7ada45bb9d9.png)
- [Small](https://user-images.githubusercontent.com/12417657/68619177-eca50800-0498-11ea-863b-71d448dadfe3.png)

### DIFF
--- a/resources/assets/components/pages/CausePage/cause-page.scss
+++ b/resources/assets/components/pages/CausePage/cause-page.scss
@@ -27,4 +27,12 @@
       }
     }
   }
+
+  .content-block-grid {
+    grid-column: full-start / full-end;
+
+    @include media($large) {
+      grid-column: full-start / -5;
+    }
+  }
 }
--- a/resources/assets/components/pages/CausePage/CausePage.js
+++ b/resources/assets/components/pages/CausePage/CausePage.js
@@ -42,6 +42,7 @@ const CausePage = ({ coverImage, superTitle, title, description, content }) => {
           className="base-12-grid"
           classNameByEntry={{
             GalleryBlock: 'grid-full',
+            ContentBlock: 'content-block-grid',
           }}
         >
           {content}


Vmack:
Nice work! content-block-grid is fancy enough 😅

mendelB:
> Nice work! content-block-grid is fancy enough 😅

I'm nothing if not descriptive 🙃

resources/assets/components/pages/CausePage/cause-page.scss
```diff
@@ -27,4 +27,12 @@
...
     }
   }
+
+  .content-block-grid {
+    grid-column: full-start / full-end;
+
+    @include media($large) {
+      grid-column: full-start / -5;
+    }
+  }
```
weerd: I wouldn't recommend this name since it's so specific; if someone ends up wanting to use the same grid but in some gallery or other block, it'll be weird to use this class name when not a content block.

I'd recommend maybe finding a name that would fit with the others at the bottom of the [grid class name list](https://github.com/DoSomething/phoenix-next/blob/a24afa6b89306bb856ab34890654dfa61850cabe/resources/assets/scss/base.scss#L899-L961); something like `grid-wide-*`, which are meant to depict a grid layout based on where it should sit on wide devices, but has appropriate layout for smaller sizes.

Also, I'd recommend using `span X` over `-5` cause a negative column number  is less intuitive, rather than knowing where to start and how many columns to span :)

mendelB: > something like grid-wide-* with an appropriate fraction, which are meant to depict a grid layout based on where it should sit on wide devices, but has appropriate layout for smaller sizes.

Ahh, thanks this is helpful! I was confused about the naming scheme and just got scared and ran away -- I should have just pinged you!

Side note: do you think it's worth adding some comments atop those groupings at some point to clarify their structural meaning?

(Also going to address all of this in a follow up PR.)

weerd: Yeah, comments on the groupings sounds like a good idea. I think I meant to add some and forgot 😬

PR #1704, opened 2019-11-12
Starts on School Finder tests

aaronschachter:
### What does this PR do?

This PR branches from #1694 to start minimal test coverage for the School Finder component, mainly:

* Should only display when the campaign ID is 9001 (we'll need to additionally check for the production TFJ campaign ID once it exists)

* Should display "Find Your School" or "Your School" for campaign 9001 based on whether user has a `schoolId` set.

I've [downloaded the latest GraphQL schema](https://www.apollographql.com/docs/ios/downloading-schema/) and updated our `schema.json` in order to query by the newly added `schoolId` and `school` fields on a User.

### Any background context you want to provide?

More tests will include walking through the flow of selecting state, verifying school dropdown appears, verifying Submit button is enabled once a school is selected, etc.

### What are the relevant tickets/cards?

* https://www.pivotaltracker.com/story/show/169549385
* https://www.pivotaltracker.com/story/show/169549488

### Checklist

* [ ] Documentation added for new features/changed endpoints.
* [ ] Added appropriate feature/unit tests.

PR #1708, opened 2019-11-13
Fixes cursed Beta Referral Page test

aaronschachter:
### What does this PR do?

This PR comments out the test that works on StyleCI intermittently, which I thought I resolved in #1694 by moving it to the end of the file, but as you'll see by the merge of #1694 (and the non-merged #1704) - it did not. When the test fails, it's because the `Embed` component never fully loads with an `a` tag. I tried adding mocks for the GraphQL request in #1704 to resolve the issue, but it still would occur).

Refs:
* https://github.com/DoSomething/phoenix-next/pull/1682#issuecomment-551872581
* https://github.com/DoSomething/phoenix-next/pull/1691/commits/9febefe6af7bfbf344f6b8e168c1826c32f1b23e
* gif of panda having trouble with computer in an office cubicle

### Checklist

### DIFF
--- a/cypress/integration/beta-referral-page.js
+++ b/cypress/integration/beta-referral-page.js
@@ -39,8 +39,7 @@ describe('Beta Referral Page', () => {
   });
 
   /**
-   * When this test is executed after the 1st test, where we test with valid User + Campaign,
-   * the Embed request hangs with a mystery "Hello World", never rendering the Embed a tag.
+   * This Embed request hangs with a mystery "Hello World", never rendering the Embed a tag.
    * @see https://dashboard.cypress.io/#/projects/ayzqmy/runs/876
    */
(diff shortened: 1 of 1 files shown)


mendelB:
@aaronschachter did you try adding the `cy.mockGraphqlOps...` to actually mock the Embed request? (e.g. like what we do for [campaign submission galleries](https://github.com/DoSomething/phoenix-next/blob/a24afa6b89306bb856ab34890654dfa61850cabe/cypress/integration/campaign-post.js#L40-L49))

aaronschachter:
Ya, [I did](https://github.com/DoSomething/phoenix-next/pull/1704/files#r346034078)

mendelB:
Right, I saw that but nvm me! -- I was referring to a manually mocked query, forgetting that those defined mock operations would be run by default instead of the actual query

PR #1711, opened 2019-11-14
Quick fix for broken styles on Pitch Page.

weerd:
### What does this PR do?

This PR fixes a styling bug that appeared after the Tailwind activation and switching around of classes in #1695. Thanks for spotting it @mendelB 🙌 

Fixes the Pitch Page layout to add missing spacing at bottom of content before the call-to-action. It switches to also using padding instead of margins and adds appropriate clearfix, since the primary and secondary items are floated on larger sizes.

### DIFF
--- a/resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
+++ b/resources/assets/components/pages/LandingPage/templates/PitchTemplate.js
@@ -38,9 +38,9 @@ const PitchTemplate = ({
         showPartnerMsgOptIn={showPartnerMsgOptIn}
       />
 
-      <div className="clearfix bg-white">
-        <Enclosure className="md:w-3/4 mx-auto mt-6 px-3 pitch-landing-page">
-          <div className="campaign-page">
+      <div className="bg-white">
+        <Enclosure className="md:w-3/4 mx-auto py-6 px-3 pitch-landing-page">
+          <div className="campaign-page clearfix">
             <div className="primary">
               {displayAffiliateScholarshipBlock ? (
                 <AffiliateScholarshipBlockQuery


mendelB:
> hooray for tailwind! quick q, what does the clearfix class do precisely?

@lindsayrmaher It's a mixin defined in Forge here, used to help contain floated items:
https://github.com/DoSomething/forge/blob/3828463212a96e97b154b0afa007da30b869c5ab/scss/_utilities/_mixins.scss#L13-L21

W3schools has [a nice explanation](https://www.w3schools.com/howto/howto_css_clearfix.asp) of what it's accomplishing!

weerd:
> hooray for tailwind! quick q, what does the clearfix class do precisely?

@lindsayrmaher  Ohhhh boy! Fun little CSS thing... So when an item is floated, they are taken out of the context of the parent container, allowing any elements that follow to float around it. If you _only_ have floated elements (as is the case here with the `.primary` content floated left, and the `.secondary` floated right to get the content and sidebar layout working), then the container they sit inside of no longer has any context of the dimensions of these floated items, causing the containing parent element to collapse (to a computed height of `0`) as if there was nothing inside of it. So if you have a background color/image or padding/margins etc, nothing gets applied as expected and seemingly the layout is broken.

Thus the [`clearfix`, which is a tiny CSS hack](https://css-tricks.com/clearfix-a-lesson-in-web-development-evolution/), was developed a while back (and evolved a bit over time) to easily solve this problem. It works on the containing parent element and what it typically does is insert a pseudo element inside the parent element, `::after` the content it contains. Essentially, the element is invisible, but it has a `clear: both;` rule applied, which makes it clear the floated items and sit below them as it should if clearing properly. This forces the container to then extend and incorporate all the floated items in its computed height 💃 


I ❤️ floats by the way, but it wasn't until I really understood their behavior; once you grok it there's a lot of fun to be had with floating elements all over the place. Granted Floats are less needed nowadays thanks to Flex & Grid, but super helpful back in the day to create some pretty complicated layouts if planned properly! 👴

PR #1715, opened 2019-11-14
Campaign info header

lindsayrmaher:
### What does this PR do?

This PR fixes a small bug I noticed with the `Campaign Info` header in the marquee template!

Font size got updated with TW and the header was much larger than the original.

Before:
Screen Shot 2019-11-14 at 3 34 03 PM

After:
Screen Shot 2019-11-13 at 3 05 42 PM

### Checklist

### DIFF
--- a/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
+++ b/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
@@ -28,7 +28,7 @@ const CAMPAIGN_INFO_QUERY = gql`
 
 const CampaignInfoBlock = ({ campaignId, scholarshipAmount }) => (
   <Card className="bordered padded rounded campaign-info">
-    <h1 className="mb-4 text-m uppercase">Campaign Info</h1>
+    <h3 className="mb-4 text-m uppercase">Campaign Info</h3>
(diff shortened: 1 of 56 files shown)


resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
```diff
@@ -28,7 +28,7 @@ const CAMPAIGN_INFO_QUERY = gql`
 
 const CampaignInfoBlock = ({ campaignId, scholarshipAmount }) => (
   <Card className="bordered p-3 rounded campaign-info">
-    <h1 className="mb-4 text-m uppercase">Campaign Info</h1>
+    <h3 className="mb-4 text-m uppercase">Campaign Info</h3>
```
mendelB: Ah it looks like the `.text-m` was deprecated. Can we remove that class and use `.text-lg` instead to avoid switching tags?

lindsayrmaher: why yes! just updated

mendelB: 😱 It worked!

PR #1716, opened 2019-11-14
Cleaning up and removing manually added Tailwind margin classes. 🧼

weerd:
### What does this PR do?

This PR removes the manually added Tailwind-style margi classes added to `base.scss` and updates components to use the same Tailwind-style classes but with the updated pixel value reference.

For example, in our `base.scss` we had `m-4` defined as margin on all sides of `12px`. In proper Tailwind classes generated by the framework, `12px` all around is `m-3`, so I changed all references to `m-4` to `m-3`. Similar changes for lots of the other class changes you'll see in this PR!


weerd:
Fixed the merge conflict!

PR #1718, opened 2019-11-15
Replacing old margin utility classes with Tailwind classes.

weerd:
### What does this PR do?

This PR updates markup to use the appropriate new Tailwind generated margin classes instead of the utility classes we custom named when we first started trying out utility based classes.

🔍 👀 

### DIFF
--- a/resources/assets/scss/base.scss
+++ b/resources/assets/scss/base.scss
@@ -253,8 +253,6 @@ p {
   word-break: break-word;
 }
 
-///////////////////////////////////////////////////////////////
-
 // @TODO:forge-removal Required to override Forge; delete once Forge is removed.
 .mb-0 {
   margin-bottom: 0 !important;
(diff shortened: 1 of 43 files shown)


resources/assets/scss/base.scss
```diff
@@ -253,107 +253,14 @@ p {
...
-}
-
-.margin-lg {
-  margin: $base-spacing;
-}
-
-.margin-bottom-none {
+// @TODO:forge-removal Required to override Forge; delete once Forge is removed.
+.mb-0 {
   margin-bottom: 0 !important;
```
DFurnes: Do we want to enable the [`important`](https://tailwindcss.com/docs/configuration/) Tailwind option so we don't need to worry about specificity issues like this? (And then disable once Forge is outta here?)

weerd: I was debating that. I might just see how many more of these pop up where it actually is affected b/c of Forge. If only a handful, not a big deal, but if many more I say we turn it on.

PR #1668, opened 2019-10-31
Starts on School Finder

aaronschachter:
### What does this PR do?

This PR starts on the front-end for the School Finder component we'll be adding for Teens For Jeans, continuing the development with [React Select](https://react-select.com/) started in #1666.

Adds basic functionality, hardcoded to campaign 9001 for testing on dev:

* Display state select
* Display school select upon selecting state, querying our internal [lofi-schools](https://github.com/DoSomething/lofi-schools-flask) app
* Display completed state after selecting school

#### School search:

Screen Shot 2019-10-30 at 5 27 52 PM

#### School select:

Screen Shot 2019-10-30 at 5 28 11 PM

#### School submit:

Screen Shot 2019-10-30 at 5 33 51 PM

### Any background context you want to provide?

This isn't ready for prime-time (still plenty of placeholder copy), and plenty left to build:

* Saving selected school to Northstar user
* Displaying selected school upon visiting campaign if user previously selected school
* The "My school isn't available" option 

This felt like a good stopping point to keep things (hopefully) reviewable. I'll add the review app to our dev `phoenix` client so this can be manually tested.
 
Because this isn't ready for production, I won't include screenshots of all screens.

### What are the relevant tickets/cards?

R&D for https://www.pivotaltracker.com/n/projects/2401401/stories/169294661

### Checklist

* [ ] Documentation added for new features/changed endpoints.
* [ ] Added appropriate feature/unit tests.

PR #1719, opened 2019-11-15
Adds school-not-available option to School Finder

aaronschachter:
### What does this PR do?

This PR modifies the School Finder to allow users to select an option for "My school isn't available" if their school search doesn't return any results.

It also cleans things up a bit:
* Moves `CurrentSchool` component into the `SchoolFinder`, renames `UpdateSchool` as `SchoolFinderForm`
* Refactors check for when to display School Finder as a `shouldShowSchoolFinder` prop - refs https://github.com/DoSomething/phoenix-next/pull/1668#discussion_r341567911
* Adds a `school-finder` constants file, defining related hardcoded campaign ID's, school ID's and static text.
* Fixes TODOs for updating the `SchoolFinder` component via Apollo cache instead of maintaining state - refs https://github.com/DoSomething/phoenix-next/pull/1694#pullrequestreview-315002395

Eventually want to add Cypress tests to mock the steps for submitting the School Finder form, but will save for a future PR to help keep this reviewable (and begin work on the Rogue side soon)

Screen Shot 2019-11-15 at 11 24 57 AM
Screen Shot 2019-11-15 at 11 25 06 AM
Screen Shot 2019-11-15 at 11 25 14 AM

* https://www.pivotaltracker.com/story/show/169549547
* https://github.com/DoSomething/graphql/pull/160

### Checklist


resources/assets/components/pages/CampaignPage/CampaignPageContainer.js
```diff
@@ -30,6 +31,9 @@ const mapStateToProps = (state, ownProps) => {
     noun: get(state.campaign.additionalContent, 'noun'),
     pages: state.campaign.pages,
     shouldShowLandingPage: shouldShowLandingPage(state, entryContent),
+    shouldShowSchoolFinder: SCHOOL_FINDER_CAMPAIGN_IDS.includes(
+      state.campaign.campaignId,
+    ),
```
DFurnes: Instead of checking against a hard-coded list of IDs, would it make sense to make this an `additionalContent` field on the Contentful entry? That'd make it a little easier to test and toggle on-and-off.

```js
shouldShowSchoolFinder: get(state.campaign.additionalContent, 'shouldShowSchoolFinder', false),
```

resources/assets/components/utilities/SchoolFinder/SchoolSelect.js
```diff
@@ -29,15 +34,27 @@ const SchoolSelect = ({ filterByState, onChange }) => {
...
         },
       })
-      .then(res => callback(res.data.searchSchools))
+      .then(result => {
+        // Return search results, or School Not Available ID if no school was found.
+        const options = result.data.searchSchools.length
+          ? result.data.searchSchools
+          : [
+              {
+                id: SCHOOL_NOT_AVAILABLE_SCHOOL_ID,
```
DFurnes: As a `TODO`, I wonder if this makes sense to return from the query itself in GraphQL if no results exist? (Since this is something we then support in the resolver, so it doesn't necessarily seem app-specific).

resources/assets/components/pages/CampaignPage/CampaignPageContainer.js
aaronschachter: Will address in a future PR!

resources/assets/components/utilities/SchoolFinder/SchoolSelect.js
aaronschachter: I was definitely wondering about that over in https://github.com/DoSomething/graphql/pull/160... I felt it was a litle more app specific in the sense that if someone is searching for "zzzzzz", is their school really not available? *google searches sleep schools*

PR #1722, opened 2019-11-18
Moar updates and cleanup.

weerd:
### What does this PR do?

This PR does more cleanup removing manually added utility classes in favor of their Tailwind counterparts generated by the library. 

It also Tailwind-ifies the html for the search results page form and actually works a bit better now!

New Tailwind-ified Search

It provided an opportunity to create 2 new composed classes created using Tailwind's nifty [`@apply` method](https://tailwindcss.com/docs/extracting-components/#app), which allows us to use Tailwind utility classes, but within a composed class. This is a nice feature so that we stick to using the styles defined in the Tailwind config in these composed classes; if a value every changes, these composed classes get update too :) That being said, we don't want a lot of composed classes, but for buttons and input fields they kind of make sense™. Eventually, these composed classes should all live in new Forge!

### DIFF
--- a/resources/assets/components/AdminDashboard/AdminDashboard.js
+++ b/resources/assets/components/AdminDashboard/AdminDashboard.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const AdminDashboard = ({ enabled, children }) =>
-  enabled ? <div className="bg-light-gray -white">{children}</div> : null;
+  enabled ? <div className="bg-light-gray -white p-4">{children}</div> : null;
 
 AdminDashboard.propTypes = {
   children: PropTypes.node.isRequired,
--- a/resources/assets/components/AdminDashboard/CampaignDashboard/CampaignDashboard.js
+++ b/resources/assets/components/AdminDashboard/CampaignDashboard/CampaignDashboard.js
@@ -19,14 +19,14 @@ const CampaignDashboard = props => {
     <div>
       <button
         type="button"
-        className="button -secondary m-3"
+        className="button -secondary"
         onClick={clickedShowAffirmation}
       >
         Show Affirmation
       </button>
       <button
         type="button"
-        className="button -secondary m-3"
+        className="button -secondary ml-3"
         onClick={onSignUpClick}
       >
         {`Mock ${isSignedUp ? 'Un-sign Up' : 'Sign Up'}`}
--- a/resources/assets/components/LegacyQuiz/Answer.js
+++ b/resources/assets/components/LegacyQuiz/Answer.js
@@ -16,7 +16,7 @@ const Answer = props => {
     backgroundImage,
   } = props;
 
-  const classes = classnames('bordered rounded', {
+  const classes = classnames('bordered overflow-hidden rounded', {
     '-active': isActive,
     faded: isFaded,
   });
--- a/resources/assets/components/pages/AccountPage/Account.js
+++ b/resources/assets/components/pages/AccountPage/Account.js
@@ -14,7 +14,7 @@ const Account = props => (
       <div className="bg-white">
         <div className="base-12-grid">
           <div className="grid-wide pt-12">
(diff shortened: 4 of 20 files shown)


resources/assets/scss/base.scss
```diff
@@ -1,14 +1,55 @@
 // Tailwind CSS
 @tailwind base;
 @tailwind components;
-@tailwind utilities;
-
-@import 'next-toolbox';
 
 // Import Forge so we get our base styles.
 $forge-path: '../../../node_modules/@dosomething/forge/assets/';
 @import '~@dosomething/forge/scss/forge';
```
weerd: Forge was overriding some margins and other rules because before we were outputting Forge's CSS _AFTER_ the Tailwind Utility classes. So by shifting it after the Tailwind base & components, Forge can adjust some of the initial defaults Tailwind sets (until we can remove Forge), but then the Tailwind utilities classes can easily override rules that Forge has defined!

resources/assets/components/AdminDashboard/AdminDashboard.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const AdminDashboard = ({ enabled, children }) =>
-  enabled ? <div className="bg-light-gray -white">{children}</div> : null;
+  enabled ? <div className="bg-light-gray -white p-4">{children}</div> : null;
```
weerd: @mendelB fix for the admin buttons spacing issues.

resources/assets/components/AdminDashboard/CampaignDashboard/CampaignDashboard.js
```diff
@@ -19,14 +19,14 @@ const CampaignDashboard = props => {
...
-        className="button -secondary m-3"
+        className="button -secondary"
         onClick={clickedShowAffirmation}
       >
         Show Affirmation
       </button>
       <button
         type="button"
-        className="button -secondary m-3"
+        className="button -secondary ml-3"
```
weerd: @mendelB fix for the admin buttons spacing issues.

resources/assets/components/pages/AccountPage/Account.js
```diff
@@ -14,7 +14,7 @@ const Account = props => (
       <div className="bg-white">
         <div className="base-12-grid">
           <div className="grid-wide pt-12">
-            <h1 className="league-gothic -lg caps-lock">
+            <h1 className="font-league-gothic font-normal text-4xl md:text-5xl uppercase">
```
weerd: Add a little resizing of the header text from small screen to medium screen sizes!

resources/assets/scss/base.scss
```diff
@@ -1,14 +1,55 @@
...
+
+// @TODO:forge-removal Rename this class to "text-field".
+.text-input {
+  @apply bg-white p-3 border border-gray-200 border-solid rounded;
+}
+
+.text-input:focus {
+  @apply outline-none border border-solid border-blue-500;
+  box-shadow: 0 0 3px theme('colors.blue.500');
+}
```
weerd: These are the new [composed classes](https://tailwindcss.com/docs/extracting-components/#app)!

resources/assets/components/LegacyQuiz/Answer.js
```diff
@@ -16,7 +16,7 @@ const Answer = props => {
     backgroundImage,
   } = props;
 
-  const classes = classnames('bordered rounded', {
+  const classes = classnames('bordered overflow-hidden rounded', {
```
mendelB: Lol we really should 🔥 this component

resources/assets/scss/base.scss
mendelB: Oh snap! 🎉

resources/assets/components/LegacyQuiz/Answer.js
weerd: Haha! Ya, probs 😬

PR #1725, opened 2019-11-19
Page GraphQL Query Refactor

mendelB:
### What does this PR do?

This PR refactors the `CausePageQuery` component into a more general `PageQuery` component for wider page usage.

It's also generally refactored to use the [`useQuery`](https://www.apollographql.com/docs/react/data/queries/#executing-a-query) (shoutout @aaronschachter for getting the hooks rolling in https://github.com/DoSomething/phoenix-next/pull/1694) Apollo hook for a way leaner look & feel!

I also snuck in some validation scripting for cover images on Campaigns/Pages -- something I've already added to Contentful:
https://github.com/DoSomething/phoenix-next/commit/143166133e9087b92c9cd73f8043c132fa7e0306 to limit images to 20MBs re [this issue](https://dosomething.slack.com/archives/CP2D7UGAU/p1574115878193400?thread_ts=1574110151.183300&cid=CP2D7UGAU)
and https://github.com/DoSomething/phoenix-next/commit/1a83b2b85eb3949613d2a614290b7e1306e5ad79 to require cover image on Campaigns re [this one](https://dosomething.slack.com/archives/CP2D7UGAU/p1574111874187000?thread_ts=1574110151.183300&cid=CP2D7UGAU)

Getting all set to introduce Collection Pages, which will also make use of the `PageQuery`!

Gonna hopefully add some Cypress testing for this component via testing the Cause/Collection pages in a later PR!

Refs [Pivotal ID #169582403](https://www.pivotaltracker.com/story/show/169582403)

### Checklist


resources/assets/components/App.js
```diff
@@ -42,7 +43,14 @@ const App = ({ store, history }) => {
...
                 render={routeProps => (
-                  <CausePageQuery slug={routeProps.match.params.slug} />
+                  <PageQuery
+                    query={CAUSE_PAGE_QUERY}
+                    variables={{ slug: routeProps.match.params.slug }}
+                  >
+                    {({ causePageBySlug }) => (
+                      <CausePage {...causePageBySlug} />
+                    )}
+                  </PageQuery>
```
mendelB: This does introduce some extra rendering logic in the `App.js` which might not be optimal? We could still have this logic go in a `CausePageQuery` component to allow `App.js` to simply render it with the `slug` 🤔

DFurnes: Yeah, I agree this feels like a bit too much in `App`. I think I'd lean towards having `CausePage` itself make the query, since it's unlikely we'll ever want to render it with data provided from some other source.

resources/assets/components/pages/PageQuery.js
```diff
@@ -0,0 +1,41 @@
...
+
+  if (loading) {
+    return <Placeholder />;
+  }
+
+  if (error) {
+    return <ErrorPage />;
+  }
+
+  if (isEmpty(withoutNulls(data))) {
```
DFurnes: Can this just be `!data`? I think the server should return `null` if a page isn't found, right?

resources/assets/components/App.js
mendelB: Mmm, that's a good point and a good idea! Best of both worlds 🎉

resources/assets/components/pages/PageQuery.js
mendelB: It seems to always return the top-level query key with a null value -- so `{causePageBySlug: null}` when it's not found.

DFurnes: Ahh, right right. Would it make sense to alias this to `page: causePageBySlug(…) { … }`, and then we could check whether `data.page` is empty or not here (regardless of specific query used)?

mendelB: Clever! I couldn't for the life of me think of a great way to make this assumption cleaner without like a `dataKey` prop or something, but an alias makes so much sense!!

To confirm I understand -- so the `PageQuery` is just assuming a top-level `page` query, and whoever uses it just has the onus of setting that alias on the query?

DFurnes: Yep. I'm also mulling over the idea of a general `page(slug: "…")` query that could handle resolving which type of thing was routed (from, say, a `Routable` GraphQL interface), which could help clean up that assumption down the line... but that's for another day!

mendelB: That sounds super intriguing, looking forward to seeing that play out!

mendelB:
**Update**:
Moved query/rendering over to the component in https://github.com/DoSomething/phoenix-next/pull/1725/commits/3a7114953b5affd3a30ebb19dec9b02b2f73e404 re https://github.com/DoSomething/phoenix-next/pull/1725#discussion_r348003615
And added the alias in https://github.com/DoSomething/phoenix-next/pull/1725/commits/a15709155157d7941b7915c2c130947df5e5291a re https://github.com/DoSomething/phoenix-next/pull/1725#discussion_r348009101

resources/assets/components/pages/PageQuery.js
```diff
@@ -0,0 +1,40 @@
...
+
+  if (error) {
+    return <ErrorPage />;
+  }
+
+  if (!data.page) {
+    return <NotFoundPage />;
+  }
+
+  return children(data);
```
mendelB: Hmm now that we have the key, we should probs just yield `data.page` here directly? 🤔

DFurnes: Yeah, I think that makes sense. 👍

mendelB: Sweeeet - updated in https://github.com/DoSomething/phoenix-next/pull/1725/commits/d51b8a9320f9bee7d6c9ceeb0b60b75835cf04e4!

PR #1728, opened 2019-11-19
Remove deprecated PhotoSubmissionAction.additionalContent

aaronschachter:
### What does this PR do?

This PR fixes the TODO's of removing the `additionalContent.action` props. From what I can tell in Contentful, all recent `PhotoSubmissionAction` entries contain an `actionId` field value -- and this should be safe to remove.

It also seems safe to fix the [TODO of removing support of an `action` value in a Rogue `POST /v3/posts` request](https://github.com/DoSomething/rogue/blob/master/app/Http/Requests/PostRequest.php#L39), as both Phoenix and Gambit are sending over action ID's when creating posts (refs https://github.com/DoSomething/gambit/pull/480)

Prep for https://www.pivotaltracker.com/story/show/169549761
### Checklist

### DIFF
--- a/resources/assets/actions/post.js
+++ b/resources/assets/actions/post.js
@@ -70,7 +70,7 @@ export function resetPostSubmissionItem(id) {
  * @return {function}
  */
 export function storeCampaignPost(campaignId, data) {
-  const { action, actionId, blockId, body, pageId, type } = data;
+  const { actionId, blockId, body, pageId, type } = data;
 
   if (type === 'photo' && !(body instanceof FormData)) {
     throw Error(
@@ -105,7 +105,6 @@ export function storeCampaignPost(campaignId, data) {
         body,
         failure: POST_SUBMISSION_FAILED,
         meta: {
-          action,
           actionId,
           campaignId,
           id: blockId, // @TODO: rename property to blockId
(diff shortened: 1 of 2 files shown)


DFurnes:
> From what I can tell in Contentful, all recent `PhotoSubmissionAction` entries contain an `actionId` field value -- and this should be safe to remove.

Have we updated these for old database campaigns too?

aaronschachter:
After the first page of results for `PhotoSubmissionAction`, there are no Action ID's set, but spot checking quite a few of these entries and there is nothing in `additionalContent` either. 

Screen Shot 2019-11-19 at 11 29 07 AM

aaronschachter:
Tested on the [Trash Stash](https://www.dosomething.org/us/campaigns/trash-stash/) campaign, whose `PhotoSubmissionAction` doesn't have `additionalContent` or `actionId` fields set. These photo posts get created with a `default` action ID.

mendelB:
> It also seems safe to fix the TODO of removing support of an action value in a Rogue POST /v3/posts request, as both Phoenix and Gambit are sending over action ID's when creating posts (refs DoSomething/gambit#480)

@aaronschachter one thing to bear in mind if retiring `action` over on Rogue  -- we do have many Share Actions with `additionalContent.action` set. I think this was to differentiate Share Actions pushed to from SMS broadcasts, and one's showing up on other pages. 

I imagine at least most of them are out of use but just figured it was worth noting!

aaronschachter:
Oh that's a great call @mendelB - thanks for the context. Not looking to remove `action` on the Rogue anytime too soon (but sounds like a nice icebox card if it doesn't exist already...)

PR #1731, opened 2019-11-19
Hero Template for Action Page & opt out logic

lindsayrmaher:
### What does this PR do?

This PR updates our Lede Banner to support using the `MarqueeTemplate` by default once a member is signed up for a campaign. It also adds logic to handle opting out of using the Hero Template (NFL campaigns and Taco Bell will use this flag). 

I'd like to clean up our file structure now that the concept of `landing page` isn't really implemented. It would probably make sense for this to be done after the NFL and Taco Bell campaigns are complete.

We want to have a seamless experience for the users before and after they signup.

Refs [Pivotal ID # 169685104](https://www.pivotaltracker.com/story/show/169685104)


resources/assets/components/LedeBanner/LedeBanner.js
```diff
@@ -16,12 +17,21 @@ const LedeBanner = props => {
       return <JumboTemplate {...props} />;
```
DFurnes: Since [`template` is hard-coded to `mosaic`](https://github.com/DoSomething/phoenix-next/blob/bd3fc36bed215bf6a9bd29e768c08ee05a417521/app/Entities/Campaign.php#L132), should we just remove these options?

resources/assets/components/pages/LandingPage/LandingPageContainer.js
```diff
@@ -27,6 +27,11 @@ const mapStateToProps = (state, ownProps) => {
     coverImage: state.campaign.coverImage,
     endDate: state.campaign.endDate,
     scholarshipAmount: state.campaign.scholarshipAmount,
+    featureFlagUseLegacyTemplate: get(
+      state,
+      'campaign.additionalContent.featureFlagUseLegacyTemplate',
```
DFurnes: Nitpick: This is a mouthful! Would `useLegacyTemplate` work?

resources/assets/components/pages/LandingPage/LandingPageContainer.js
```diff
@@ -27,6 +27,11 @@ const mapStateToProps = (state, ownProps) => {
...
+      undefined,
```
DFurnes: Nitpick: We can omit this if we want it to default to `undefined`, since that's the default fallback!

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -127,6 +131,8 @@ MarqueeTemplate.propTypes = {
   campaignId: PropTypes.string,
   content: PropTypes.string.isRequired,
   coverImage: PropTypes.object.isRequired,
+  displaySignUp: PropTypes.bool,
```
DFurnes: Nitpick: I wonder if `displaySignupButton` might be clearer?

resources/assets/components/pages/LandingPage/LandingPageContainer.js
lindsayrmaher: ah yes! good call

resources/assets/components/pages/LandingPage/LandingPageContainer.js
weerd: Haha, I suggested this 😬 I felt it'd make it a lot easier to find AND immediately know this is just a feature flag. It could be nice to preface our feature flags with something easily searchable in code (rather than unique names per flag), so when we need to remove them, they're easy to find and deal with. Anyhoo, that was my thinking on it 😄

resources/assets/components/LedeBanner/LedeBanner.js
lindsayrmaher: the options for the Cover Template and the JumboTemplate?

lindsayrmaher: I'm not clear on when we're using them rn. ps where are we hard coding it?

DFurnes: I believe this is loaded from the `template` property on our Laravel API endpoint, linked above. Maybe there are other places it's being overridden, though?

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
lindsayrmaher: that does feel slightly more descriptive to me, i'm down to update.

resources/assets/components/pages/LandingPage/LandingPageContainer.js
DFurnes: Mmm, that's fair! I'm convinced.

lindsayrmaher: huzzah!

resources/assets/components/LedeBanner/LedeBanner.js
lindsayrmaher: @weerd @mendelB based on this ^^ are we missing something or are we not using these other templates anywhere else? I didn't see any reference to them with a quick search

mendelB: Nowhere that I can think of! They're just relics of other iterations/experiments for the `LedeBanner`

lindsayrmaher: thoughts on perhaps including this in a cleanup card? we could maybe also remove hard coding it at all on the API endpoint altogether, but that feels out of scope for the PR perhaps.

DFurnes: Yeah, that'd be a great cleanup card!

weerd: Ya, they were sixpack tests I did a while back. Safe to remove. Side note, the styling for them was fun 💃

resources/assets/components/pages/LandingPage/LandingPageContainer.js
mendelB: Lil memory lane trip -- we used to nest these feature flags under a `featureFlags` key! (though that requires its own nesting logic, but this made me think of this!)
https://github.com/DoSomething/phoenix-next/blob/95534d9d9c7c594782f1e1adcc52f642e5fd3d36/resources/assets/components/Campaign/Campaign.js#L60

PR #1736, opened 2019-11-20
Hero template campaign info updates

lindsayrmaher:
### What does this PR do?

This PR adds additional logic to display a report back action if there is no scholarship and report back action option

👀  

Refs [Pivotal ID # 169685104](https://www.pivotaltracker.com/story/show/169685104)


resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
```diff
@@ -26,62 +26,66 @@ const CAMPAIGN_INFO_QUERY = gql`
   }
 `;
 
-const CampaignInfoBlock = ({ campaignId, scholarshipAmount }) => (
-  <Card className="bordered p-3 rounded campaign-info">
-    <h1 className="mb-3 text-lg uppercase">Campaign Info</h1>
+const CampaignInfoBlock = ({ campaignId, scholarshipAmount }) => {
+  return (
```
mendelB: @lindsayrmaher should we revert this parenthesis to brackets change so the changes are more clear?

lindsayrmaher: oh yeah can do, i think i left it that way when i had my console log 😂

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
```diff
@@ -34,12 +34,12 @@ const CampaignInfoBlock = ({ campaignId, scholarshipAmount }) => (
...
-            ? res.campaign.actions.filter(
-                action => action.scholarshipEntry && action.reportback,
-              )
-            : [];
-          const action = actions.length ? actions[0] : null;
+          const actions = res.campaign.actions || [];
+          const scholarshipActions = actions.filter(
+            action => action.reportback && action.scholarshipEntry,
+          );
+          const reportbackActions = actions.filter(action => action.reportback);
```
DFurnes: Instead of using `.filter(…)` and grabbing the first element, let's use `.find(…)` (which short-circuits once it finds a match)! That seems to match what we're trying to do here a little closer, and saves us a wee bit of work too!

lindsayrmaher: ohhhhh! ok yes, i like that refactor. looking into eet 👀

lindsayrmaher:
@DFurnes ! Lemme know if this looks good to you, but seems to work as expected 🎉

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
```diff
@@ -34,12 +34,12 @@ const CampaignInfoBlock = ({ campaignId, scholarshipAmount }) => (
...
-                action => action.scholarshipEntry && action.reportback,
-              )
-            : [];
-          const action = actions.length ? actions[0] : null;
+          const actions = res.campaign.actions || [];
+
+          const actionItem =
+            actions.find(
+              action => action.reportback && action.scholarshipEntry,
+            ) || actions.find(action => action.reportback);
```
DFurnes: It feels like it could be easy to miss this if skimming this in the future. Let's split out into a second block?

```js
let actionItem = actions.find(
  action => action.reportback && action.scholarshipEntry
);

if (! actionItem) {
  actionItem = actions.find(action => action.reportback);
}

```

lindsayrmaher: sure sure, that feels more clear to me! i'll add that before i merge.

PR #1673, opened 2019-11-01
Adds Newsletter CTA to Scholarship Page - Part 1

Vmack:
### What does this PR do?

This PR adds the `CtaPopoverEmailForm` and `CtaPopoverButton` component to the `CtaPopover` directory, it adds styling to the email form using pre-existing CtaPopover styling, it tracks analytics, adds children to `CtaPopover` and adds `CtaPopover` to the `GeneralPage` component.

### Any background context you want to provide?
This feature will allow unauthenticated non-members to submit their email address to sign up for the newsletter when visiting the scholarship page. This popover will have similar design to the CtaPopover component. 

### What are the relevant tickets/cards?

Refs [Pivotal ID #168209412](https://www.pivotaltracker.com/story/show/168209412)

**small screen**
image

**medium screen**
image

**large screen**
image

### Checklist

* [ ] Documentation added for new features/changed endpoints.
* [ ] Added appropriate feature/unit tests.
* [x] Added screenshot to PR description of related front-end updates on **small** screens.
* [x] Added screenshot to PR description of related front-end updates on **medium** screens.
* [x] Added screenshot to PR description of related front-end updates on **large** screens.

PR #1740, opened 2019-11-20
Adds Dismissable Element to Cta Popover Email Form on Scholarship Page

Vmack:
### What does this PR do?

This PR adds delay functionality to `CtaPopoverEmailForm` so that the component mounts 3 seconds after a user visits the page 

Extension of [#1673](https://github.com/DoSomething/phoenix-next/pull/1673)

Refs [Pivotal ID #168209412](https://www.pivotaltracker.com/story/show/168209412)

### Checklist

### DIFF
--- a/resources/assets/components/pages/GeneralPage/GeneralPage.js
+++ b/resources/assets/components/pages/GeneralPage/GeneralPage.js
@@ -12,14 +12,14 @@ import ContentfulEntry from '../../ContentfulEntry';
 import { REGISTER_CTA_COPY } from '../../../constants';
 import AuthorBio from '../../utilities/Author/AuthorBio';
 import CtaBanner from '../../utilities/CtaBanner/CtaBanner';
-import CtaPopover from '../../utilities/CtaPopover/CtaPopover';
-import CtaPopoverEmailForm from '../../utilities/CtaPopover/CtaPopoverEmailForm';
+// import CtaPopover from '../../utilities/CtaPopover/CtaPopover';
+// import CtaPopoverEmailForm from '../../utilities/CtaPopover/CtaPopoverEmailForm';
 import TextContent from '../../utilities/TextContent/TextContent';
 import { contentfulImageUrl, withoutNulls } from '../../../helpers';
 import SocialShareTray from '../../utilities/SocialShareTray/SocialShareTray';
 import SiteNavigationContainer from '../../SiteNavigation/SiteNavigationContainer';
-import DelayedElement from '../../utilities/DelayedElement/DelayedElement';
-import DismissableElement from '../../utilities/DismissableElement/DismissableElement';
+// import DelayedElement from '../../utilities/DelayedElement/DelayedElement';
+// import DismissableElement from '../../utilities/DismissableElement/DismissableElement';
 
 import './general-page.scss';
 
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/GeneralPage/GeneralPage.js
```diff
@@ -145,15 +148,25 @@ const GeneralPage = props => {
...
+          render={(handleClose, handleComplete) => (
+            <DelayedElement delay={3}>
+              <Modal onClose={handleClose}>
+                <CtaPopover
+                  title="PAYS TO DO GOOD"
+                  content="Want to earn easy scholarships for volunteering?
+                  Subscribe to DoSomething's monthly scholarship email."
+                >
+                  <CtaPopoverEmailForm />
+                  <Modal onSubmit={handleComplete} />
```
mendelB: I don't think we'll need this new Modal here but can rather pass the `handleComplete` as a prop to the `CtaPopoverEmailForm` which it can invoke upon successful email submission so that once a user submits their email they never see the popup again!

resources/assets/components/pages/GeneralPage/GeneralPage.js
```diff
@@ -145,15 +148,25 @@ const GeneralPage = props => {
...
-        >
-          <CtaPopoverEmailForm />
-        </CtaPopover>
-      ) : null} */}
+      {slug === 'about/easy-scholarships' ? (
+        <DismissableElement
+          name="cta_popover_email"
+          render={(handleClose, handleComplete) => (
+            <DelayedElement delay={3}>
+              <Modal onClose={handleClose}>
```
mendelB: What's cool is we don't actually need to use a `Modal` here! Since the Popover functions as it's own sort of modal, all we need to do is pass through the `handleClose` as a prop to the `CtaPopover`, and it'll invoke it when a user clicks on the X button: https://github.com/DoSomething/phoenix-next/blob/5c865e345d5e577c8c7a3636e7a09f3cec59ff2e/resources/assets/components/utilities/CtaPopover/CtaPopover.js#L12

resources/assets/components/pages/GeneralPage/GeneralPage.js
Vmack: That's a great idea, I just added it. I'm assuming i'd have to import the 
`DissmissableElement` to the email form as well? Just so that the prop is defined.

mendelB: Sweet! And you shouldn't actually need to import it either, since the `handleClose` is just a function [being yielded](https://github.com/DoSomething/phoenix-next/blob/5c865e345d5e577c8c7a3636e7a09f3cec59ff2e/resources/assets/components/utilities/DismissableElement/DismissableElement.js#L39) by the `DismissableElement`, so as long as we have access to the function - which we will in the email form when we pass it through as a prop - we can just invoke it without needing to directly import the `DismissableElement` 🎉

resources/assets/components/pages/GeneralPage/GeneralPage.js
Vmack: So I tried this and it's not working. Not sure why? 🤔

resources/assets/components/pages/GeneralPage/GeneralPage.js
```diff
@@ -51,7 +53,7 @@ const GeneralPage = props => {
       <div className="main general-page base-12-grid">
         <Enclosure className="grid-narrow">
           <div className="general-page__heading text-center">
-            <h1 className="general-page__title uppercase">{title}</h1>
+            <h1 className="general-page__title caps-lock">{title}</h1>
```
mendelB: I think this needs to be `uppercase` to align with the new Tailwind implementation?

Vmack: good eye! Thanks @mendelB!

PR #1745, opened 2019-11-21
Tailwindifying the call to action component 🧼

weerd:
### What does this PR do?

This PR uses Tailwind to re-style the `CallToAction` component and the different variations it can show up as (light or dark). There is not really a lot of usage of this component throughout the site, apart from automatic output on certain pages, so for the most part felt very safe to revise this component with Tailwind 💃 

The component used to have variations for "sticky" and "transparent" but we have made the call to remove those for now.

The "sticky" version no longer worked as sticky and has not worked for some time; this was due to changes in parent containers that effectively rendered the sticky functionality to no longer work. 

The "transparent" version was only really used in a single location, specifically on campaign pages when the campaign is open and the member is not signed up (so I believe it was previously only really used on the community pages when they were unauthenticated). It was not worth keeping and by removing it allows further cleanup with the Tailwind efforts. I converted it to a "light" version and seems fine for the _one_ scenario where it can show up.

Here is how the variations for the `CallToAction` component look with Tailwind:

Light variation:
light variation

Light variation with impact info:
light variation with impact

Dark variation:
dark variation

Mosaic Pitch Landing Page version:
landing page variation

Refs [Pivotal ID #169850790](https://www.pivotaltracker.com/story/show/169850790)


tailwind.config.js
```diff
@@ -84,7 +84,7 @@ module.exports = {
         '200': '#ffe27a',
         '300': '#ffdd61',
         '400': '#ffd747',
-        '500': '#fcce2e',
+        '500': '#fcce2f',
```
weerd: Updating the main `500` value yellow to the approved new brand color yellow. The rest of the yellow's likely need to be updated with approved colors once ready. cc @lkpttn

PR #1748, opened 2019-11-21
Social Drive Action - Page Views toggle

mendelB:
### What does this PR do?

This PR adds a new Hide Page Views toggle field to the Social Drive Action on Contentful to allow toggling the display of the 'page views' info card.

My first pass explored using a 'Show Page Views' field, but that ended up making it slightly more complicated (dealing with the styling hack especially to set the column width based on this setting) without setting a default somewhere outside the component. And since the default _is_ for it to display, it seemed reasonable to have the editor toggle it off if desired.

Refs [Pivotal ID #169826562](https://www.pivotaltracker.com/story/show/169826562)

### Checklist


Vmack:
Awesome, this is great! Looks like there is a circleci fail, possibly from cypress test?

mendelB:
> Awesome, this is great! Looks like there is a circleci fail, possibly from cypress test?

Yeah I think so! I re-ran the build 🤞

Vmack:
Sweeeet!

PR #1751, opened 2019-11-21
Add Hello Bar script

katiecrane:
### What does this PR do?

This PR adds the Hello Bar script to our site! It can only show up on prod, but I'm seeing this message locally so I think I did it right:
image

We'll be testing this on prod on low traffic pages. For additional context, here is the PR for when we added this before (in Ashes!): https://github.com/DoSomething/phoenix/pull/7519

Refs [Pivotal ID #169608418](https://www.pivotaltracker.com/story/show/169608418)

### DIFF
--- a/resources/views/layouts/master.blade.php
+++ b/resources/views/layouts/master.blade.php
@@ -55,7 +55,7 @@
     @endisset
 
     @stack('scripts')
-    <script src="https://my.hellobar.com/043326c5f2334a35a23aff244adacb9d99149ca6.js" type="text/javascript" charset="utf-8" async="async"></script> 
+    <script src="https://my.hellobar.com/f70465b5f088ff5df39e838a358d27109b365641.js" type="text/javascript" charset="utf-8" async="async"></script> 
 </body>
 
 </html>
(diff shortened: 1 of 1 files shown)


DFurnes:
> I feeeeel like we discussed this in Planning but I can't recall -- did we want to add a QA HelloBar to be able to test things out? (Not that that would be a blocker for this since that'd also be configured via HelloBar, right?)

That feels like it could be helpful!

katiecrane:
@DFurnes @mendelB we'd need a separate snippet for each environment. The one we have now is locked to prod and can't be used on QA. As far as I can tell, in Ashes we only ever had on snippet? Should we have two?

DFurnes:
Oh, that's frustrating. If we've never needed it in the past, it's probably not worth the hassle then.

katiecrane:
@DFurnes right?? Seems like we should be able to whitelist it, but alas! Low traffic pages on prod to test it is

mendelB:
meme

katiecrane:
hahaha @mendelB you know I'd only agree to this if we had no choice!!

mendelB:
> hahaha @mendelB you know I'd only agree to this if we had no choice!!

lol but that adrenaline rush is to die for. I only test on _QA_ when they twist my arm.

katiecrane:
this is @mendelB:

free_solo_

PR #1758, opened 2019-11-22
Removes info bar on closed campaigns

lindsayrmaher:
### What does this PR do?

This PR add a check whether a campaign is closed to hide a second info bar from being visible on the Marquee template. It is duplicated at the bottom of the campaign page.

Flagged by design and Ben!

**Before:**
Before

**After:**
After

### Checklist

### DIFF
--- a/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
+++ b/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
@@ -26,6 +26,7 @@ const MarqueeTemplate = ({
   content,
   coverImage,
   displaySignupButton,
+  isCampaignClosed,
   isAffiliated,
   scholarshipAmount,
   scholarshipDeadline,
@@ -119,7 +120,7 @@ const MarqueeTemplate = ({
           </Enclosure>
         </div>
       </article>
-      {!isAffiliated ? <CampaignInfoBarContainer /> : null}
+      {!isAffiliated && isCampaignClosed ? <CampaignInfoBarContainer /> : null}
     </React.Fragment>
   );
 };
@@ -134,6 +135,7 @@ MarqueeTemplate.propTypes = {
   coverImage: PropTypes.object.isRequired,
   displaySignupButton: PropTypes.bool,
   isAffiliated: PropTypes.bool,
+  isCampaignClosed: PropTypes.bool,
   scholarshipAmount: PropTypes.number,
   scholarshipDeadline: PropTypes.string,
   subtitle: PropTypes.string.isRequired,
@@ -148,6 +150,7 @@ MarqueeTemplate.defaultProps = {
   campaignId: null,
   displaySignupButton: true,
   isAffiliated: false,
+  isCampaignClosed: false,
   scholarshipAmount: null,
   scholarshipDeadline: null,
 };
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -119,7 +120,7 @@ const MarqueeTemplate = ({
           </Enclosure>
         </div>
       </article>
-      {!isAffiliated ? <CampaignInfoBarContainer /> : null}
+      {!isAffiliated && isCampaignClosed ? <CampaignInfoBarContainer /> : null}
```
mendelB: Wouldn't this _show_ the extra campaign info bar atop the campaign page if the user is unaffiliated and the campaign is closed? Should it be:
`{!isAffiliated && !isCampaignClosed ? <CampaignInfoBarContainer /> : null}`

lindsayrmaher: that's what i thought as well, but when i set it that way it was still showing up on closed campaigns that you hadn't signed up for.

mendelB: iiiiinteresting lol I'm so confused

lindsayrmaher:
but now that you mention it, it's doing some wonky stuff in QA. i merged too soon 🤦‍♀

mendelB:
I'm gonna try writing down my understanding of when to render this to make sure I grok this correctly:
- if the user's unaffiliated OR affiliated with this campaign, and the campaign is closed == NO Info Bar (since we'll show the community page or a closed campaign component. ❌ 
- if the user's affiliated with the campaign, and the campaign is _not_ closed == NO info bar (since we'll show a campaign action page) ❌ 
- if the user's unaffiliated with this campaign, and the campaign is _not_ closed == Info bar! (since we won't be showing anything other than the hero template!) ✅ 

Does this make sense, or am I mucking this up?

lindsayrmaher:
Yah yah that's what we're going for. would it make sense to say `!isAffiliated || isAffiliated && !isCampaignClosed` or something along those lines?

mendelB:
> Yah yah that's what we're going for. would it make sense to say `!isAffiliated || isAffiliated && !isCampaignClosed` or something along those lines?

That's where I'm so confused 😖 I would've thought the logic would simply be `!isAffiliated && !isCampaignClosed` since those are the two conditions for when we'd render this. Maybe woth investigating why that _wouldnt_ simply work? 🤔

lindsayrmaher:
@mendelB I thiiiiink i've solved the mystery. After many circles of confusion with myself, I was just not passing down the right `isClosed` campaign to the closed campaign component, and thus it wasn't evaluating the way it should be. I'm putting in a new PR #1761 to put in the very similar but better fix. 😅

mendelB:
> @mendelB I thiiiiink i've solved the mystery. After many circles of confusion with myself, I was just not passing down the right `isClosed` campaign to the closed campaign component, and thus it wasn't evaluating the way it should be. I'm putting in a new PR #1761 to put in the very similar but better fix. 😅

🕵️‍♀️😌Glad we solved this mystery!

PR #1759, opened 2019-11-22
Add ability to load AffirmationBlock via GraphQL.

DFurnes:
### What does this PR do?

This pull request adds the ability to load `AffirmationBlock` via GraphQL. Here's the "normal" version, still provided by the PHP Content API, and the new version on `blocks/:id`:

Screen Shot 2019-11-22 at 1 56 44 PM

Screen Shot 2019-11-22 at 1 56 41 PM

I have not added the ability to specify an `onClose` yet, since I want to think if there's a better way we can do that without having to pass a "single-use" prop that doesn't apply to more components. Let me know if you have any ideas!

Refs [Pivotal ID #169216379](https://www.pivotaltracker.com/story/show/169216379).

### Checklist

### DIFF
--- a/resources/assets/components/Affirmation/Affirmation.js
+++ b/resources/assets/components/Affirmation/Affirmation.js
@@ -10,10 +10,27 @@ import Share from '../utilities/Share/Share';
 import Badge from '../pages/AccountPage/Badge';
 import Byline from '../utilities/Byline/Byline';
 import TextContent from '../utilities/TextContent/TextContent';
-import { contentfulImageUrl, withoutNulls } from '../../helpers';
+import { contentfulImageUrl } from '../../helpers';
 import CtaReferralPageBannerContainer from '../utilities/CtaReferralPageBanner/CtaReferralPageBannerContainer';
 
 import './affirmation.scss';
+
+export const AffirmationBlockFragment = gql`
+  fragment AffirmationBlockFragment on AffirmationBlock {
+    id
+    header
+    quote
+    callToActionHeader
+    callToActionDescription
+    author {
+      name
+      jobTitle
+      photo {
+        url
+      }
+    }
+  }
+`;
 
 const USER_QUERY = gql`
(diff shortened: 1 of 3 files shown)


resources/assets/components/Affirmation/Affirmation.js
```diff
@@ -42,65 +59,67 @@ const Affirmation = ({
...
-            >
-              <h4>1 Sign-Up</h4>
-              <p>
-                Congratulations! You signed up for your first campaign...and
-                earned your first badge. NICE. Ready to earn *another* badge?
-                Complete this campaign!
-              </p>
-            </Badge>
-          ) : null}
+    {userId ? (
```
katiecrane: Good call!

resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -81,10 +81,20 @@ class ContentfulEntry extends React.Component<Props, State> {
...
         );
 
+      // @TODO: Need to figure out 'onClose'...
+      case 'AffirmationBlock':
+        return (
+          <AffirmationContainer
+            userId={window.AUTH.id}
+            {...withoutNulls(json)}
+          />
+        );
```
katiecrane: Am I reading this right - this part is not in use yet?

DFurnes: Yep, right now this is only used to display the affirmation on the `blocks/:id` route.

DFurnes:
Thinking about this some more –– I'm not sure that the affirmation really should be a block at all, since we don't ever embed it within content, like other blocks. I think I'm going to go back and add an `affirmation` query to GraphQL so we can get at this directly.

DFurnes:
Oh, except sometimes we do allow multiple block types to be used as "affirmations", like the `affirmationBlock` reference field on the Share Block. Hmm...

PR #1761, opened 2019-11-22
More Closed Campaign Fixes

lindsayrmaher:
### What does this PR do?

This PR adds a cleaner and more correct fix for the info footer bar logic on closed campaigns.

Just bang operator confusion aaaaall over the place.

### DIFF
--- a/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
+++ b/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
@@ -77,7 +77,6 @@ const CampaignRoute = props => {
                   affiliatePartners={affiliatePartners}
                   affiliateSponsors={affiliateSponsors}
                   campaignLead={campaignLead}
-                  isClosed={isClosed}
                 />
               );
             }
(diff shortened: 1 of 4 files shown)


resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
```diff
@@ -77,6 +77,7 @@ const CampaignRoute = props => {
                   affiliatePartners={affiliatePartners}
                   affiliateSponsors={affiliateSponsors}
                   campaignLead={campaignLead}
+                  isClosed={isClosed}
```
DFurnes: What scenarios do we render a `CampaignClosedPage` where this would be false?

lindsayrmaher: I don't think there is one, but I needed to pass that check to the LedeBanner that's rendered in the CampaignClosedPage in order for it to evaluate displaying to info bar on the hero template.

DFurnes: If we don't ever pass `isClosed={false}` here, could we just use a boolean for the `LedeBannerContainer` prop below and avoid having to pass anything into `CampaignClosedPage`?

```jsx
<LedeBannerContainer isClosed={true} />
```

PR #1752, opened 2019-11-21
DRY TextSubmission and PetitionSubmission with ActionInformation, PostCreatedModal

aaronschachter:
### What does this PR do?

This PR swaps out repeated code with new components in #1741. As I started looking in each action component to see we'd want to include a `school_id` on the post, I figured I'd clean this up while I was in in each.

I created a new `PetitionSubmissionAction` over on the [Test School Finder Campaign](https://dev.dosomething.org/us/campaigns/test-school-finder-campaign/action) in order to manually test this worked (but haven't added school ID collection to these action components yet - can do in future PR)

### Any background context you want to provide?

.I made the swap in the [Social Share as well, but it broke jest tests](https://github.com/DoSomething/phoenix-next/commit/66c3794664722992200e2d94ade11e437cfec679)... which seemed very odd. Can save for a ☔️  

### What are the relevant tickets/cards?

Refs https://github.com/DoSomething/phoenix-next/pull/1732, https://github.com/DoSomething/phoenix-next/pull/1536#issuecomment-519957645

### Checklist

* [ ] Documentation added for new features/changed endpoints.
* [ ] Added appropriate feature/unit tests.

PR #1762, opened 2019-11-22
Post + School ID cleanup

aaronschachter:
### What does this PR do?

This PR moves the `getUserActionSchoolId` helper introduced in #1741 over to  a new `PostForm` class component, which all of our class components that create posts (and need to check whether to include the user's school ID) can inherit from. I've refactored the `PhotoSubmissionAction` to inherit from it, as well as `TextSubmissionAction` and `PetitionSubmissionAction`. 

The `PostForm` class component also has an `automatedTest` boolean prop that fixes the Invariant Violation we get when trying to create a new instance of ApolloClient within our MockedProvider in jest tests (refs https://github.com/DoSomething/phoenix-next/pull/1752#issuecomment-557584791, #1741)

I've been adding new actions to dev, and manually testing over on the Example School Finder Campaign I've been working against.

Last components left to refactor for School ID check will be for a future PR: SelectionAction, ShareAction, SoftEdgeWidgetAction

https://www.pivotaltracker.com/story/show/169549761

### Checklist

### DIFF
--- a/resources/assets/components/actions/PostForm.js
+++ b/resources/assets/components/actions/PostForm.js
@@ -27,7 +27,8 @@ class PostForm extends React.Component {
 
     /**
      * Needed to query GraphQL on form submit events while our Submission components are
-     * defined as classes -- and can't use React Hooks until we refactor as functional components.
+     * defined as classes and can't use React Hooks until they are as functional components.
+
      *
      * Because we're querying outside of our App, our tests fail with an Invariant Violation,
      * because MockedProvider can't find fetch for this new Apollo Client we're creating on the fly.
@@ -39,14 +40,20 @@ class PostForm extends React.Component {
 
   /**
    * Query GraphQL to determine whether to save user's current school when creating a post.
+   *
+   * @return {Promise}
    */
   async getUserActionSchoolId() {
-    const { actionId, userId } = this.props;
+    const { actionId, automatedTest, userId } = this.props;
 
-    if (!actionId || this.props.automatedTest) {
+    if (!actionId || automatedTest) {
       return Promise.resolve(null);
     }
 
+    /**
+     * We use ApolloClient.query here because React hooks fail wuith Invalid Hook Call warning.
+     * @see https://reactjs.org/warnings/invalid-hook-call-warning.html
+     */
     const result = await this.gqlClient.query({
       query: USER_ACTION_SCHOOL_ID_QUERY,
       variables: { userId, actionId },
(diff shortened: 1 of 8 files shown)


resources/assets/components/actions/PostForm.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    const result = await this.gqlClient.query({
+      query: USER_ACTION_SCHOOL_ID_QUERY,
+      variables: { userId, actionId },
+    });
+
+    return result.data.action.collectSchoolId
+      ? result.data.user.schoolId
+      : null;
```
mendelB: BTW, once we're making the call to grab the school ID anyways, would it pay to just collect the school ID on _any_ post since we'll have the info available, or is there a reason to only specifically collect it for certain posts?

resources/assets/components/actions/PostForm.js
```diff
@@ -0,0 +1,76 @@
...
+    super(props);
+
+    /**
+     * We need to create an ApolloClient instance to call query to make GraphQL requests within the
+     * Submission Action components because they are defined as classes, and can't use React Hooks.
+     * @see https://reactjs.org/warnings/invalid-hook-call-warning.html
+     *
+     * Because we're querying outside of our App, our tests fail with an Invariant Violation,
+     * because MockedProvider can't find fetch for this new Apollo Client we're creating on the fly.
+     * @see https://circleci.com/gh/DoSomething/phoenix-next/1398?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
```
mendelB: Ah, did you get the `Invariant violation` _even_ when wrapping the component in a `MockedProvider` in the Jest test?

aaronschachter: Yep, that's why I added `automatedTest` prop to this class, and set it in the Jest tests for the child components of this class.

resources/assets/components/actions/PostForm.js
aaronschachter: The reason we're not saving it across the board on all posts is because a user may not always be reporting back in affiliation with their school depending on what the campaign is. I believe the approach will be that we'll generally want to collect school ID  for actions when a school finder is available on the campaign page.

mendelB: Gotcha, so it's not really a passive identification of the school they're in, but a more specific school association for _this_ RB?

aaronschachter: Yup! That's why we need the query for the action's `collectSchoolId` property, newly added over in https://github.com/DoSomething/rogue/pull/954.

mendelB: Sweet, thanks for clarifying!

PR #1763, opened 2019-11-22
Configure JUnit output for test metadata.

DFurnes:
### What does this PR do?

This pull request starts aggregating [test metadata](https://circleci.com/docs/2.0/collect-test-data/#jest) for Jest & Cypress tests in this application, like we do in Northstar & Rogue. This will allow us to keep an eye on flaky and/or slow tests over time.

Here's the documentation for [enabling this for Jest](https://circleci.com/docs/2.0/collect-test-data/#jest) and [for Cypress](https://docs.cypress.io/guides/tooling/reporters.html#Spec-to-STDOUT-save-JUnit-XML-files).

N/A

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ stats.json
 /public/next/uploads
 /public/storage
 /resources/assets/fragmentTypes.json
+/reports
 /storage/debugbar
 /vendor
 /_book
--- a/.gitignore
+++ b/.gitignore
@@ -22,5 +22,6 @@ yarn-error.log
 .env
 .DS_Store
 
+junit.xml
 npm-debug.log
 /contentful/management-api-scripts/logs
--- a/package.json
+++ b/package.json
@@ -2,9 +2,9 @@
   "private": true,
   "scripts": {
     "test": "jest --watch",
-    "test:ci": "jest --max-Workers=2",
+    "test:ci": "jest --ci --runInBand --reporters=default --reporters=jest-junit",
     "cypress": "cypress open",
-    "cypress:ci": "cypress run -c baseUrl=http://127.0.0.1:8000 --record",
+    "cypress:ci": "cypress run -c baseUrl=http://127.0.0.1:8000 --record --reporter cypress-multi-reporters --reporter-options configFile=cypress-report-config.json",
     "lint": "eslint --ext .js resources/assets",
     "flow": "flow",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -2,9 +2,9 @@
   "private": true,
   "scripts": {
     "test": "jest --watch",
-    "test:ci": "jest --max-Workers=2",
+    "test:ci": "jest --ci --runInBand --reporters=default --reporters=jest-junit",
```
mendelB: Does the `runInBand` flag give us the safety to replace `--max-workers=2`? (It looks like from the linked doc that it'll also govern CPU allocation?)

.gitignore
```diff
@@ -8,6 +8,7 @@ stats.json
 /public/next/uploads
 /public/storage
 /resources/assets/fragmentTypes.json
+/reports
```
mendelB: If the results get git ignored, is there any central location we're storing the reports for the tests? Are we like inspecting these results locally, or per build?

package.json
DFurnes: Yeah, I switched to this since it seems to be what CircleCI recommended doing "officially" (and IIRC we found the `--max-Workers=2` solution from a user discussion).

Looking at the docs, it seems like [`--runInBand`](https://jestjs.io/docs/en/cli.html#--runinband) makes things single-threaded and [`--max-Workers=2`](https://jestjs.io/docs/en/cli.html#--maxworkersnumstring) would make 2 threads. From looking at our past few builds though, we seem to always get 2 CPUs so maybe we should keep `--max-Workers` for a lil' extra boost?

.gitignore
DFurnes: These are stored by CircleCI in the `store_test_results` step (so we can see overall metrics over all builds).

package.json
DFurnes: Although, the `--max-Workers` docs saying "this defaults to the number of the cores available on your machine minus one for the main thread" makes me think that if we have 2 CPUs available to us, we would only want these to run in a single-thread...

DFurnes: It doesn't look like the Jest tests ran appreciably slower with this setting (11 seconds vs. 10 seconds on the other builds I peeked at), so I think let's keep with `--runInBand` to play it safe.

mendelB: Ah cool, ok that makes good sense to me! 👍

PR #1764, opened 2019-11-25
One. More. Hero Template Update

lindsayrmaher:
### What does this PR do?

This PR passes down the isClosed prop from our `CampaignPage` so that the `campaignInfo` bar will also be suppressed correctly on closed campaigns that have a community as well as closed quiz campaigns.

🤞 this is the last update we need to cover all the bases with this.


lindsayrmaher:
@mendelB if you have time to put 👀 on this!

PR #1766, opened 2019-11-25
Cypress Tests for Hero Template

lindsayrmaher:
### What does this PR do?

This PR adds baseline tests for our new Hero Template page! Some of these are a copy of the OG campaign page signup, and there are also additional tests for the campaign info box!

Just baseline checks we'll need with Hero Template rendering

Refs [Pivotal ID #169604403](https://www.pivotaltracker.com/story/show/169604403)

### DIFF
--- a/cypress/integration/hero-campaign-signup.js
+++ b/cypress/integration/hero-campaign-signup.js
@@ -35,6 +35,60 @@ describe('Hero Template Landing Page', () => {
     cy.contains('30 minutes - 1 hour');
   });
 
+  it('Landing Page Loads As Expected with no scholarship action', () => {
+    cy.mockGraphqlOp('CampaignInfoQuery', {
+      campaign: (root, { campaignId }) => ({
+        id: campaignId,
+        actions: () => [
+          {
+            actionLabel: 'Share Something',
+            timeCommitmentLabel: '1 hour',
+            scholarshipEntry: false,
+            reportback: true,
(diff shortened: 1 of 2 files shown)


cypress/integration/hero-campaign-signup.js
```diff
@@ -0,0 +1,150 @@
...
+        ],
+      }),
+    });
+
+    // Visit the campaign pitch page:
+    cy.withState(exampleCampaign).visit('/us/campaigns/test-example-campaign');
+
+    cy.contains('Share Something');
+    cy.contains('1 hour');
+    cy.contains('$5,000');
```
katiecrane: Where is this coming from?

lindsayrmaher: It's getting pulled from the campaign page, not in the graphql query

lindsayrmaher: is that something we should explicitly set in the test so that we know for sure what the outcome should be?

katiecrane: ah so that is pulled from the example campaign?

lindsayrmaher: yah!

katiecrane: I think this is okay since we assert against attributes of the example campaign in other tests too!

PR #1770, opened 2019-11-25
Fix School Finder GraphQL error on auth redirect

aaronschachter:
### What does this PR do?

This PR adds an `id` to the School Finder's `UserSchoolQuery` GraphQL query, fixing an error that is thrown when visiting a ContentBlock with a SchoolFinder after user authentication redirect:

 ```
Store error: the application attempted to write an object with no provided id but the store already contains an id of User:[user id here] for this object.
```

Refs https://github.com/apollographql/apollo-client/issues/2510.

It also adds an optional `error` prop to the `ErrorBlock` that devs can manually alter while debugging. A TODO for a future PR is to set it via config.

This has got me wondering about the cleanest way to check that a `ContentBlock` only displays the `SchoolFinder` if we have a user ID.... or would need a view for "Login to update to your school" type of view if our ContentBlock that contains a SchoolFinder via the `additionalContent` config is used on an anonymous page.

https://dosomething.slack.com/archives/C3ASB4204/p1574712707454600?thread_ts=1574711377.449900&cid=C3ASB4204

### Checklist

### DIFF
--- a/resources/assets/components/Query.js
+++ b/resources/assets/components/Query.js
@@ -17,7 +17,7 @@ const Query = ({ query, variables, children, hideSpinner }) => (
       }
 
       if (result.error) {
-        return <ErrorBlock />;
+        return <ErrorBlock error={result.error} />;
       }
 
       return children(result.data);
--- a/resources/assets/components/Query.js
+++ b/resources/assets/components/Query.js
@@ -17,7 +17,8 @@ const Query = ({ query, variables, children, hideSpinner }) => (
       }
 
       if (result.error) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: resources/assets/components/Query.js [hunk 1]
```diff
@@ -17,6 +17,7 @@ const Query = ({ query, variables, children, hideSpinner }) => (
       }
 
       if (result.error) {
+        // If debugging, pass result.error as an error prop below:
         return <ErrorBlock />;
```
DFurnes: What if we conditionally included this prop if `process.env.NODE_ENV !== 'production'`?

PR #1768, opened 2019-11-25
Add GraphQL support for SoftEdgeBlock, SocialDriveBlock, and SelectionSubmissionBlock.

DFurnes:
# What does this PR do?

This adds support for loading `SoftEdgeBlock`, `SocialDriveBlock`, and `SelectionSubmissionBlock` via GraphQL (rather than our PHP Content API). This gets us closer to being able to load blocks via GraphQL across the board, and allows editors to preview these types via our `/us/blocks/:id` route immediately!

### Any background context you want to provide?

Here's some example blocks you can check out on the review app:

- **SocialDriveBlock:** [`/us/blocks/2T5ARr1AViKw2Kw0Q4S0so`](https://dosomething-phoenix-de-pr-1768.herokuapp.com/us/blocks/2T5ARr1AViKw2Kw0Q4S0so)
- **SelectionSubmissionBlock:** [`/us/blocks/jw6RaCInscML2rC7PorsT`](https://dosomething-phoenix-de-pr-1768.herokuapp.com/us/blocks/jw6RaCInscML2rC7PorsT)
- **SoftEdgeBlock:** [`/us/blocks/1MfeJxLR4b9CIvrzlYTNxB`](https://dosomething-phoenix-de-pr-1768.herokuapp.com/us/blocks/1MfeJxLR4b9CIvrzlYTNxB) (must be authenticated)

### What are the relevant tickets/cards?

Refs [Pivotal ID #169216379](https://www.pivotaltracker.com/story/show/169216379)

### Checklist

* [ ] Documentation added for new features/changed endpoints.
* [ ] Added appropriate feature/unit tests.
* [ ] Added screenshot to PR description of related front-end updates on **small** screens.
* [ ] Added screenshot to PR description of related front-end updates on **medium** screens.
* [ ] Added screenshot to PR description of related front-end updates on **large** screens.

PR #1771, opened 2019-11-25
Simplify SoftEdge block & fix error when rendering anonymously.

DFurnes:
### What does this PR do?

This pull request refactors `SoftEdgeBlock` into one file (from three) & fixes some edge cases – the form now renders properly if displayed on an anonymous page & would support multiple instances per page if we ever wanted (since we use refs, rather than hard-coded IDs).

We can wait until after tomorrow's deploy to merge this in! Zero rush.

Follow-up to #1768.

### Checklist


resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -7,6 +7,7 @@ import Loader from '../utilities/Loader';
 import StaticBlock from '../StaticBlock';
 import { ContentfulEntryJson } from '../../types';
 import PollLocator from '../PollLocator/PollLocator';
+import SoftEdgeBlock from '../actions/SoftEdgeBlock';
```
aaronschachter: 👍

DFurnes:
Made one more tiny change in 8132583 to use this [`useScript`](https://www.npmjs.com/package/@charlietango/use-script) hook ([546 bytes](https://bundlephobia.com/result?p=@charlietango/use-script@2.1.0)), which lets us delete some code & show a more graceful error if for some reason the third-party SDK fails.

resources/assets/components/actions/SoftEdgeBlock.js
```diff
@@ -0,0 +1,104 @@
...
+
+  // 2. If logged in, grab user fields to pre-fill form:
+  const { data, loading } = useQuery(SOFT_EDGE_AUTOFILL_QUERY, {
+    variables: { userId },
+    skip: !userId,
+  });
+
+  // 3. If we couldn't load CongressWeb SDK, show an error block:
+  if (scriptStatus === ScriptStatus.ERROR) {
+    return <ErrorBlock />;
```
aaronschachter: Is there something we can send over to the `error` prop (just added in https://github.com/DoSomething/phoenix-next/pull/1772) ?

resources/assets/components/actions/SoftEdgeBlock.js
```diff
@@ -0,0 +1,104 @@
...
+      address: addrStreet1
+      city: addrCity
+      state: addrState
+      zip: addrZip
+    }
+  }
+`;
+
+const SoftEdgeBlock = ({ actionId, softEdgeId, title }) => {
+  const userId = window.AUTH.id;
```
aaronschachter: Should components set `userId` this way now instead of the usual way?
```
  userId: getUserId(state),
});
```

resources/assets/components/actions/SoftEdgeBlock.js
```diff
@@ -0,0 +1,104 @@
...
+    user(id: $userId) {
+      memberId: id
+      firstName
+      lastName
+      email
+      mobile
+      address: addrStreet1
+      city: addrCity
+      state: addrState
+      zip: addrZip
```
aaronschachter: Should we ever want to save `school_id` on a user's action post - we'd need to check for their `schoolId` here (as well as check whether our given action should collect school ID), and then modify the Chompy import to expect a `school_id` property in the [SoftEdge API](https://github.com/DoSomething/chompy/blob/master/docs/endpoints/softedge.md)? Do they need to make changes on their end to send it over?

resources/assets/components/actions/SoftEdgeBlock.js
DFurnes: Great idea! Added in 4d24e16.

resources/assets/components/actions/SoftEdgeBlock.js
DFurnes: Both are fine for now, in my opinion. I think it'll make sense to provide these via context in the future (not necessarily via Redux), but haven't gotten around to setting that up yet!

resources/assets/components/actions/SoftEdgeBlock.js
DFurnes: Yeah, as far as I know that would require changes on SoftEdge's end to include a hidden form field we could use to populate that data.

aaronschachter: Since SoftEdge posts to our Chompy API to end up creating the post in Rogue, we could make the internal requests on our end when importing the post to see whether we should be saving the user's current `school_id` to the email post.

PR #1774, opened 2019-11-26
Refer A Friend - reward & campaign updates

mendelB:
### What does this PR do?

This PR updates the Refer a Friend feature to be tailored for the new scholarship entries reward, and limited to the Teens for Jeans Campaign.

Mostly just asset and copy updates. I'll note some notable commits:
- https://github.com/DoSomething/phoenix-next/commit/0483ed662ad628c3ee7c50c1c0221ae646f2333a initially, the new assets had different dimensions, which led me to need to finesse the styling to adjust for the new dimensions. We ended up getting new images with the same ole dimensions, but since I'd already spent a bunch of time finicking, it felt nice to apply it here as well so that the images match the OG mocks a bit closer for [large](https://user-images.githubusercontent.com/12417657/69657858-1b021600-1049-11ea-8edc-11f8114f6fa1.png), [small](https://user-images.githubusercontent.com/12417657/69657984-556bb300-1049-11ea-8d7d-ece1fb30b3eb.png), and [extra small](https://user-images.githubusercontent.com/12417657/69657985-556bb300-1049-11ea-8a8c-b360453a8381.png) screens.
- https://github.com/DoSomething/phoenix-next/commit/5acbe637ad5a1c3c0333aa61f4a482888a679d37 and https://github.com/DoSomething/phoenix-next/commit/2a4dbbd02fab088d72b2635b5244967e12234e8f filter for the Teens For Jeans campaign (and it's dev sandbox counterpart), since that's what we're limiting this RAF run to.

### Any background context you want to provide?
We're changing the referral reward to scholarship entries, and limiting the initiative to the upcoming Teens for Jeans campaign.

We're still missing one asset swap for the [share link metadata](https://github.com/DoSomething/phoenix-next/blob/fef37af48a294cd906f0e3c81779c4e96caed0f6/app/Http/Controllers/ReferralPageController.php#L41). I'm just [waiting on that asset](https://www.pivotaltracker.com/story/show/169376190/comments/209129642)

### What are the relevant tickets/cards?

Refs [Pivotal ID #169376190](https://www.pivotaltracker.com/story/show/169376190)

### Checklist

* [ ] Documentation added for new features/changed endpoints.
* [ ] Added appropriate feature/unit tests.
* [ ] Added screenshot to PR description of related front-end updates on **small** screens.
* [ ] Added screenshot to PR description of related front-end updates on **medium** screens.
* [ ] Added screenshot to PR description of related front-end updates on **large** screens.

[...]

PR #1780, opened 2019-11-26
Re-enable the bugging out test expectations for RAF test

mendelB:
### What does this PR do?

❗️This PR re-enables the previously buggy beta-referral tests

Sooo I think I finally solved the mystery here (while I was poking around for some updates for #1774)

#### The setup
Our `Embed` component [will output](https://github.com/DoSomething/phoenix-next/blob/c1c62a6de625c87c5ea4acd7fa2a9d79280cb58b/resources/assets/components/utilities/Embed/Embed.js#L69-L76) a `<div>` with embedded HTML for any provider returning an `iframe` or some such code (which happens for e.g. Youtube videos).

Now, our stubbed graphql response, was _not_ defining the `type` or `html` fields,  resulting in a stubbed `html: 'Hello World'` response, which resulted in that phantom [Hello World](https://files.slack.com/files-pri/T024GV2BW-FPYM5KA3U/screen_shot_2019-11-08_at_9.48.43_am.png) embed we were seeing ...which would cause the tests to intermittently fail.

#### Two oddities tho
Two oddities though:
1. Interesting that we'd get a stubbed response for a field _not_ defined in the graphql fixture. (you've thunk that the response would be _only_ those defined stubbed fields.)
2. Why were the tests passing locally?

#### A solution?!
As for #2: the `Embed` component, while waiting for the embed graphql query to load, outputs a skeletal `<a>` tag with loading state, but _containing_ the [`url`](https://github.com/DoSomething/phoenix-next/blob/c1c62a6de625c87c5ea4acd7fa2a9d79280cb58b/resources/assets/components/utilities/Embed/Embed.js#L78). So if Cypress didn't wait for the response to resolve, and ran its assertions, it'd find the `href` attribute with the expected URL! 

[Here's](https://dashboard.cypress.io/projects/ayzqmy/runs/1226/specs) a failure from when I accidentally stubbed `html: 'null'`as a String!
 
#### A half baked thought:
So perhaps our local machines ran _slower_ then CircleCI? With the response on Circle resolving quicker, causing the faulty HTML to embed and _not_ the `<a>` tag thus failing the tests on occasion? I'm not positive but probably something along those lines.

#### Ok, wat?!
Now the best part is, that I was hoping for a nice ❌ when I pushed my first commit to re-enable the tests, but alas, 4 re-builds and it's all ✅ ! Maybe something to do with the new `--run-in-band` flag which could [cause the builds to be a smidge slower](https://github.com/DoSomething/phoenix-next/pull/1763/files#r350267155)? 🤷‍♂ 

🆗  In any case, essay over. This should continue being green, and if this introduces more failures, I'll swallow my battered pride and revert this.

🤔 Something to mull over: How do we cause assertions to run against the _fully loaded_ embed component / graphql queries in general? Shouldn't this happen by default? What are we doing wrong or right here?

https://dosomething.slack.com/archives/C3ASB4204/p1573230848173700
#1708

### DIFF
--- a/cypress/fixtures/graphql.js
+++ b/cypress/fixtures/graphql.js
@@ -28,10 +28,12 @@ export const operations = {
   },
   EmbedQuery: {
     embed: {
+      type: 'LINK',
       title: faker.lorem.words(),
       description: faker.lorem.sentence(),
       providerName: faker.company.companyName(),
       thumbnailUrl: faker.image.imageUrl(),
+      html: null,
     },
   },
   ReferralPageCampaignQuery: {
--- a/cypress/integration/beta-referral-page.js
+++ b/cypress/integration/beta-referral-page.js
@@ -17,11 +17,10 @@ describe('Beta Referral Page', () => {
 
     cy.get('.referral-page-campaign').should('have.length', 1);
 
-    // @TODO: Our mock GraphQL doesn't seem to get called, this times out like below.
-    // cy.get('.referral-page-campaign > a')
-    //   .should('have.attr', 'href')
-    //   .and('include', `referrer_user_id=${userId}`)
-    //   .and('include', campaignPath);
+    cy.get('.referral-page-campaign > a')
+      .should('have.attr', 'href')
+      .and('include', `referrer_user_id=${userId}`)
+      .and('include', campaignPath);
   });
 
   it('Visit beta referral page, with invalid user ID', () => {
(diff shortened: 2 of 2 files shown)


DFurnes:
> Interesting that we'd get a stubbed response for a field not defined in the graphql fixture. (you've thunk that the response would be only those defined stubbed fields.)

By default, we mock fields based on what's allowed by the schema (so we'd return a random valid value for the `EmbedType` enum here, `Hello World` for a string, or a random `true` or `false` for Boolean fields, etc).

We only need to add overrides for things where the default mocks aren't representative of what real data would look like.

> How do we cause assertions to run against the fully loaded embed component / graphql queries in general? Shouldn't this happen by default? What are we doing wrong or right here?

Hmm, yeah - Cypress will keep re-checking until it finds the thing it's looking for, but it's tricky for things like Embed that include a placeholder while they're loading! It might make sense to add a `data-loading` attribute or something of the sort that we could assert against here.

mendelB:
@DFurnes Thanks for clarifying!

> It might make sense to add a data-loading attribute or something of the sort that we could assert against here.

That's smart! I feel like in this test, for now, we're OK since we're rendering links and stubbing links, but something like that will be useful for when we need to actually test more complex embeds?

PR #1781, opened 2019-11-27
Refer a Friend scholarship - edits and tweaks

mendelB:
### What does this PR do?

This PR follows up on #1774 with some minor copy updates we caught on QA (err, prod).

It also adds the Navigation bar to the RAF ~pages~ (updated to just alpha page https://github.com/DoSomething/phoenix-next/pull/1781#discussion_r351426449) in https://github.com/DoSomething/phoenix-next/pull/1781/commits/7ada61890641d08539dc4c0a853f9b6b7e11ff3b

Refs [Pivotal ID #169987137](https://www.pivotaltracker.com/story/show/169987137)

### DIFF
--- a/resources/assets/components/pages/ReferralPage/Beta/BetaPage.js
+++ b/resources/assets/components/pages/ReferralPage/Beta/BetaPage.js
@@ -6,7 +6,6 @@ import ErrorPage from '../../ErrorPage';
 import { query } from '../../../../helpers';
 import MoneyHandImage from './money-hand.svg';
 import CampaignLink from './BetaPageCampaignLink';
-import SiteNavigationContainer from '../../../SiteNavigation/SiteNavigationContainer';
 import {
   REFERRAL_CAMPAIGN_IDS,
   DEFAULT_REFERRAL_CAMPAIGN_ID,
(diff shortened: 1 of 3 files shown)


resources/assets/components/pages/ReferralPage/Beta/BetaPage.js
```diff
@@ -34,77 +35,84 @@ const BetaPage = () => {
     : DEFAULT_REFERRAL_CAMPAIGN_ID;
 
   return (
-    <Query query={REFERRAL_PAGE_USER} variables={{ id: userId }}>
-      {data => {
-        if (!data.user) {
-          return <ErrorPage />;
-        }
+    <>
+      <SiteNavigationContainer />
```
aaronschachter: It turned out to be good that the site navigation was no longer on Referral Pages, because we weren't sending over the `referral_user_id` if a user registered via site navigation (and not by clicking through one of the campaign links). Refs https://www.pivotaltracker.com/story/show/169205444

mendelB: Oh blast. Ya know, I had a sneaking suspicion there was a reason but got lazy about it. Ok, removed in https://github.com/DoSomething/phoenix-next/pull/1781/commits/6c1b9ff2d5b4403247e3e5a78240e4dd3592ee11!

mendelB: (It'd also be hella hot, if we could start formally persisting referral/utm details like this in localstorage or something, so we get better support of metadata like this on signups/reportbacks etc.)

aaronschachter: For sure. @ngjo added a story to finally PT that feature that's come up in a bit - https://www.pivotaltracker.com/n/projects/2328687/stories/169960864. Would you mind adding a comment to document why we're not sending? A`SiteNavigation` tweak that might be nice `@TODO` would be checking for existing query string parameters and append to all links it renders -- so we could keep the nav.

mendelB: Oh neat! and smart -- I'll throw a comment in. 
> A SiteNavigation tweak that might be nice @TODO would be checking for existing query string parameters and append to all links it renders -- so we could keep the nav

That's a really good idea. I realized that we actually _would_ support sending the current query params to NS if the `user_id` were `referrer_user_id` 
https://github.com/DoSomething/phoenix-next/blob/6c1b9ff2d5b4403247e3e5a78240e4dd3592ee11/resources/assets/selectors/index.js#L13

Any reason why it's not `referrer_user_id` here?
https://github.com/DoSomething/phoenix-next/blob/6c1b9ff2d5b4403247e3e5a78240e4dd3592ee11/resources/assets/components/pages/ReferralPage/Alpha/AlphaPage.js#L11

aaronschachter: Sounds like it should be! Great find.

mendelB: Sweet! I guess this is something we can tackle in a separate card if we indeed update? (I guess we'd have to update it while maintaining support for the current `user_id` param?)

PR #1784, opened 2019-12-02
Scholarship modal base information

lindsayrmaher:
### What does this PR do?

This PR pulls in the existing info that is used in the scholarship info card on a campaign page. 

Most of the fields we're using for the new scholarship modal are already existing props or queries, so this PR is just surfacing them. Some of these will be updated to be dynamic fields in Contentful in a following PR.

Refs [Pivotal ID #169919280](https://www.pivotaltracker.com/story/show/169919280)

### DIFF
--- a/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
+++ b/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
@@ -128,6 +128,7 @@ const MarqueeTemplate = ({
             onClose={() => setShowModal(!showModal)}
             scholarshipAmount={scholarshipAmount}
             scholarshipDeadline={scholarshipDeadline}
+            utmLabel={scholarshipAffiliateLabel.toLowerCase()}
           />
         ) : null}
       </article>
--- a/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
+++ b/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
@@ -124,6 +124,7 @@ const MarqueeTemplate = ({
         </div>
         {showModal ? (
           <ScholarshipModalQuery
+            affiliateSponsors={affiliateSponsors}
             campaignId={numCampaignId}
             onClose={() => setShowModal(!showModal)}
             scholarshipAmount={scholarshipAmount}
--- a/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
+++ b/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
@@ -7,9 +7,9 @@ import TextContent from '../../../utilities/TextContent/TextContent';
 import { SCHOLARSHIP_SIGNUP_BUTTON_TEXT } from '../../../../constants';
 import SignupButtonContainer from '../../../SignupButton/SignupButtonContainer';
 import CampaignInfoBlock from '../../../blocks/CampaignInfoBlock/CampaignInfoBlock';
-import ScholarshipModal from '../../../utilities/ScholarshipModal/ScholarshipModal';
 import CampaignInfoBarContainer from '../../../CampaignInfoBar/CampaignInfoBarContainer';
 import AffiliatePromotion from '../../../utilities/AffiliatePromotion/AffiliatePromotion';
+import ScholarshipModalQuery from '../../../utilities/ScholarshipModal/ScholarshipModalQuery';
 import AffiliateOptInToggleContainer from '../../../AffiliateOptInToggle/AffiliateOptInToggleContainer';
 import AffiliateScholarshipBlockQuery from '../../../blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery';
 import {
@@ -123,7 +123,12 @@ const MarqueeTemplate = ({
           </Enclosure>
         </div>
         {showModal ? (
-          <ScholarshipModal onClose={() => setShowModal(!showModal)} />
+          <ScholarshipModalQuery
+            campaignId={numCampaignId}
+            onClose={() => setShowModal(!showModal)}
+            scholarshipAmount={scholarshipAmount}
+            scholarshipDeadline={scholarshipDeadline}
+          />
         ) : null}
       </article>
       {!isAffiliated && !isClosed ? <CampaignInfoBarContainer /> : null}
--- a/resources/assets/components/utilities/ScholarshipModal/ScholarshipModal.js
+++ b/resources/assets/components/utilities/ScholarshipModal/ScholarshipModal.js
@@ -3,12 +3,39 @@ import PropTypes from 'prop-types';
 
 import Modal from '../Modal/Modal';
 import Card from '../Card/Card';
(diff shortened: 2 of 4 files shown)


weerd:
Just throwing this out there and feel free to ignore me since I've been out for a while... but I thought we were trying to move away from dedicated modal components like `ScholarshipModal` or `SomeOtherBlockModal` and instead just having Blocks that can sit inside a modal if they happen to be needed to showcase as a modal. So it'd just be a `ScholarshipBlock` and on the MarqueTemplate it just happens to show up inside a `<Modal />`.

I've been away for a week, so likely I'm wrong or there's other stuff involved. 😬 @mendelB maybe you recall from our prior efforts to remove modal specific blocks and aim to create a system where blocks can be passed inside of a modal component?

lindsayrmaher:
I didn't know there was a different pattern we wanted to follow! I can see the logic in making that change, and maybe put in a follow up PR where the components are adjusted to reflect that.

lindsayrmaher:
I do then have a follow up question for @DFurnes! If we transitioned this to be a block component, would we then want to create a graphql query to contentful for the info we'll need for the modal? Right now it's coming directly from the blob of campaign info.

I'm also wondering what would be most performant in this situation, using graphql or pulling it from php like we are now. i've noticed there is a couple second lag for the modal to load which may have to do with the graphql query it needs to make to fetch the action type info from rogue.

DFurnes:
> Just throwing this out there and feel free to ignore me since I've been out for a while... but I thought we were trying to move away from dedicated modal components like `ScholarshipModal` or `SomeOtherBlockModal` and instead just having Blocks that can sit inside a modal if they happen to be needed to showcase as a modal. So it'd just be a `ScholarshipBlock` and on the MarqueTemplate it just happens to show up inside a `<Modal />`.

Yeah, I like this idea (and it's been definitely paid off with things like the voter registration modal).

> I do then have a follow up question for @DFurnes! If we transitioned this to be a block component, would we then want to create a graphql query to contentful for the info we'll need for the modal? Right now it's coming directly from the blob of campaign info.

It's fine to keep querying via PHP for now to keep this work on-track, but the plan (that we're working towards this quarter on Platforms) is to get everything querying consistently via GraphQL.

> I'm also wondering what would be most performant in this situation, using graphql or pulling it from php like we are now. i've noticed there is a couple second lag for the modal to load which may have to do with the graphql query it needs to make to fetch the action type info from rogue.

I don't think PHP is faster necessarily, but it's performing the initial data-fetching before it returns markup for the page and so it may _feel_ faster since you don't see that intermediate loading state. The plan is to set that up for our GraphQL queries as well so we can get the best of both worlds.

resources/assets/components/utilities/ScholarshipModal/ScholarshipModalQuery.js
```diff
@@ -0,0 +1,71 @@
...
+import PropTypes from 'prop-types';
+
+import Query from '../../Query';
+import { env } from '../../../helpers';
+import ScholarshipModal from './ScholarshipModal';
+/**
+ * The GraphQL query to load data for this component.
+ */
+const AFFILIATE_QUERY = gql`
+  query AffiliateQuery(
```
DFurnes: It looks like we already have [another query named `AffiliateQuery`](https://github.com/DoSomething/phoenix-next/blob/a88d7113c75e515b9633c5b6615ffbf2cc6cc1ac/resources/assets/components/blocks/AffiliateScholarshipBlock/AffiliateScholarshipBlockQuery.js#L11-L34). Let's rename this to `ScholarshipModalQuery` so that this is uniquely named (which we need for dev tooling).

resources/assets/components/utilities/ScholarshipModal/ScholarshipModal.js
```diff
@@ -3,18 +3,81 @@ import PropTypes from 'prop-types';
...
+}) => (
   <Modal onClose={onClose}>
     <Card>
       <div>
         <h1 className="p-8">This is a test scholarship modal</h1>
+        {affiliateSponsors.length ? (
+          <div className="__image">
+            <img
+              src={affiliateSponsors[0].fields.logo.url}
+              alt={affiliateSponsors[0].fields.logo.title}
```
mendelB: I think this could potentially break for campaigns without an affiliate sponsor or partner? (I guess we don't tend to really run scholarships on those campaigns, but probably worth being safe.)

resources/assets/components/utilities/ScholarshipModal/ScholarshipModalQuery.js
```diff
@@ -0,0 +1,71 @@
...
+      preview: env('CONTENTFUL_USE_PREVIEW_API'),
+      campaignId: props.campaignId,
+    }}
+  >
+    {res => {
+      const title = res.affiliate.title;
+      const actions = get(res, 'actions', []).filter(
+        action => action.scholarshipEntry && action.reportback,
+      );
+      const action = actions[0];
```
mendelB: It could be nice to use `find` here as we do in the [info block](https://github.com/DoSomething/phoenix-next/blob/4437fab42866910968b7285d86093a3664183cd2/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js#L39-L41) to grab the first match automatically.

resources/assets/components/utilities/ScholarshipModal/ScholarshipModalQuery.js
lindsayrmaher: sounds good! will do with the PR to clean up the component structure 🔧

resources/assets/components/utilities/ScholarshipModal/ScholarshipModal.js
lindsayrmaher: Definitely! I think adding a check and a fallback makes sense 👍

mendelB: Lol omg ignore me, this is wrapped in a conditional

resources/assets/components/utilities/ScholarshipModal/ScholarshipModalQuery.js
lindsayrmaher: oooooo yah that's a good catch, it would be nice to keep it consistent

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -123,7 +123,14 @@ const MarqueeTemplate = ({
           </Enclosure>
         </div>
         {showModal ? (
-          <ScholarshipModal onClose={() => setShowModal(!showModal)} />
+          <ScholarshipModalQuery
+            affiliateSponsors={affiliateSponsors}
+            campaignId={numCampaignId}
+            onClose={() => setShowModal(!showModal)}
```
mendelB: Do you think it makes sense to just hardcode this to `false` for clarity since we're always closing the modal in this instance

lindsayrmaher: i was debating with that as well, i do think it might be slightly more clear 👍

PR #1787, opened 2019-12-03
Scholarship Modal Part 2: Adding New & Existing Content

lindsayrmaher:
### What's this PR do?

This pull request adds the new fields for our scholarship modal and does some component re-structuring/clean up as per comments from #1784! Here's a breakdown of changes:

- replaces `ScholarshipModal` with `ScholarshipInfoBlock`
- cleans up naming and small logic tid bits in `ScholarshipInfoBlock`
- adds `scholarshipCallToAction` & `scholarshipDescription` to contentful content type
- dynamically loads ^^ in the modal

Look over whether the naming conventions seem more inline with our expectations for components and queries, as well as making sure the migration update on the contentful campaign type is looking good!

👀 see above!

### Relevant tickets

References [Pivotal #169919280](https://www.pivotaltracker.com/story/show/169919280).

### Checklist


contentful/content-types/campaign.js
```diff
@@ -144,17 +144,8 @@ module.exports = function(migration) {
...
-        linkMimetypeGroup: ['image'],
-      },
-      {
-        assetFileSize: {
-          max: 20971520,
-        },
-      },
-    ])
+    .required(false)
+    .validations([])
```
mendelB: I think this was an update I made to qa & master manually re https://github.com/DoSomething/phoenix-next/pull/1725, but for some reason must have forgotten to add to `dev`! Hence it's being overwritten. I just added, so I think you can just re-generate the migration to get those changes back.

contentful/content-types/campaign.js
```diff
@@ -388,6 +398,25 @@ module.exports = function(migration) {
     .disabled(false)
     .omitted(false);
 
+  campaign
+    .createField('scholarshipDescription')
+    .name('Scholarship Description')
+    .type('Text')
```
mendelB: Could it make sense to use Rich Text here as the type, and disable using headings? Kinda like in the [Cause Page description](https://github.com/DoSomething/phoenix-next/blob/854d742577f57f08fd75110fa845c7fa06afbf33/contentful/content-types/cause-page.js#L103-#L120). (I _think_ we're trying to migrate generally to just use Rich Text fields when possible).

contentful/content-types/campaign.js
lindsayrmaher: should be reset!

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -123,14 +126,17 @@ const MarqueeTemplate = ({
...
         {showModal ? (
-          <ScholarshipModalQuery
-            affiliateSponsors={affiliateSponsors}
-            campaignId={numCampaignId}
-            onClose={() => setShowModal(!showModal)}
-            scholarshipAmount={scholarshipAmount}
-            scholarshipDeadline={scholarshipDeadline}
-            utmLabel={scholarshipAffiliateLabel.toLowerCase()}
-          />
+          <Modal onClose={() => setShowModal(showModal === false)}>
```
mendelB: Could this just be `setShowModal(false)`?

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlockQuery.js
```diff
@@ -5,12 +5,12 @@ import PropTypes from 'prop-types';
...
 import Query from '../../Query';
 import { env } from '../../../helpers';
-import ScholarshipModal from './ScholarshipModal';
+import ScholarshipInfoBlock from './ScholarshipInfoBlock';
 /**
  * The GraphQL query to load data for this component.
  */
-const AFFILIATE_QUERY = gql`
-  query AffiliateQuery(
+const SCHOLARSHIP_INFO_QUERY = gql`
```
mendelB: #naming idea -- maybe we call this the `ScholarshipAffiliateQuery` to still be clear that it's an 'affiliate' query?

contentful/content-types/campaign.js
```diff
@@ -378,6 +379,15 @@ module.exports = function(migration) {
     .validations([])
     .disabled(false)
     .omitted(false);
+  campaign
+    .createField('scholarshipCallToAction')
```
mendelB: What do you think about grouping the `scholarshipCallToAction` and `scholarshipDescription` fields together (atop each other?) for editor convenience?

lindsayrmaher: as in, moving the description in contentful so that it's next to the call to action?

contentful/content-types/campaign.js
lindsayrmaher: I think so, I was wondering about that as i set this up. would that just require a change in `dev` and then re-running the migration once more?

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
lindsayrmaher: yes! even cleaner

contentful/content-types/campaign.js
mendelB: Exactly. I _feel_ like these two fields kinda go hand in hand?

PR #1791, opened 2019-12-05
Utility class cleanup pt3 🧼

weerd:
### What's this PR do?

This PR updates a slew of component SCSS files to use the Tailwind `theme()` helper to pull the corresponding value from the config file for some fonts and colors. It also updates some old format utility classes to use Tailwind instead to further cleanup the `base.scss` file.

👀... It's A LOT of files, but most of it is pretty straightforward, tiny changes! I'm commenting on a couple items that might need some additional clarification.

### Relevant tickets

References [Pivotal #169578779](https://www.pivotaltracker.com/story/show/169578779).

### Checklist


resources/assets/components/utilities/SchoolFinder/SchoolFinder.js
```diff
@@ -30,7 +30,7 @@ const SchoolFinder = props => (
         return (
           <Card
             title={schoolId ? 'Your School' : 'Find Your School'}
-            className="rounded bordered overflow-visible"
+            className="rounded bordered"
```
weerd: @aaronschachter I was able to remove this from the `SchoolFinder` implementation of the `Card` component. I removed the `overflow: hidden` rule from the `.card` class definition, since it was only really being used to help ensure that the yellow background didn't overflow outside the rounded border of the card. It was probably a bit overkill to use that rule for that purpose, so I switched it up using some tailwind classes on the card header to handle the overflow better by just rounding the top edges there as well, which made the use of `overflow-visible` rule unnecessary. Just wanted to keep ya posted. `SchoolFinder` dropdown still works as expected 😄

resources/assets/components/pages/LandingPage/landing-page.scss
```diff
@@ -27,13 +27,13 @@
   .marquee-banner__headline-title {
     text-transform: uppercase;
     font-size: $font-large;
-    color: $purple-500;
+    color: theme('colors.blurple.700');
```
lkpttn: We can make this blurple 500 in the next update

resources/assets/components/blocks/ImagesBlock/ImagesBlock.js
```diff
@@ -15,10 +15,7 @@ export const ImagesBlockFragment = gql`
 `;
 
 const ImagesBlock = ({ className, images }) => (
-  <Gallery
-    type="triad"
-    className={classnames('expand-horizontal-md', className)}
-  >
+  <Gallery type="triad" className={classnames('-mx-3', className)}>
```
weerd: Random note, Tailwind supports negative margins! Just add `-` before the class, like `-mx-3` 🕺

PR #1792, opened 2019-12-06
Cause/Collection pages content UI fix

mendelB:
### What's this PR do?

This pull request adds a default class to the content on the Cause/Collection pages so that plain copy elements are assigned a grid column span, and won't just collapse to one column!

👁 
Is the `8/12` column-span reasonable? That's the span for the Content Blocks so it seemed like a sensible choice.

We didn't account for regular markdown in the Rich Text field since the designs called for Content Blocks & Gallery Blocks ideally, but we are technically allowing this in the field, and I guess that is the beauty of the Rich Text field so it'd be a shame to just disable embedded markdown.

#### Before
image
#### After
image

### DIFF
--- a/resources/assets/components/pages/CausePage/CausePage.js
+++ b/resources/assets/components/pages/CausePage/CausePage.js
@@ -66,6 +66,7 @@ const CausePageTemplate = ({
             GalleryBlock: 'grid-full',
             ContentBlock: 'grid-full-8/12',
           }}
+          classNameByEntryDefault="grid-full-8/12"
         >
           {content}
         </TextContent>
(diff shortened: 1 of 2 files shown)


weerd:
> My kingdom for some whitespace.

Unrelated, but we neeeeed more whitespace around these sections. The content is so compressed... let it breathe!

mendelB:
> > My kingdom for some whitespace.
> 
> Unrelated, but we neeeeed more whitespace around these sections. The content is so compressed... let it breathe!

LOL.

Which sections do you mean? Or is it just _everything_?

PR #1793, opened 2019-12-06
Updates UTM params for scholarship and removes scholarship card from Hero template

lindsayrmaher:
### What's this PR do?

This pull request does some clean up work in order to get the modal officially displaying for users coming from scholarship sites! I also removed the check to show the scholarship card on the Hero template as well as removing the template for our 'beta' `AffiliateScholarshipBlock` since we no longer need it! 

👀 

We're cleaning up unnecessary components and helper functions now that the scholarship information is separated from the Hero Template!

### Relevant tickets

References [Pivotal #169924504](https://www.pivotaltracker.com/story/show/169924504).

### Checklist


resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -38,7 +36,7 @@ const MarqueeTemplate = ({
   subtitle,
   title,
 }) => {
-  const [showModal, setShowModal] = useState(isTestReferral());
+  const [showModal, setShowModal] = useState(isScholarshipAffiliateReferral());
```
mendelB: I think this might be clearer if we renamed it to something like `showScholarshipModal`?

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -49,8 +47,6 @@ const MarqueeTemplate = ({
     small: contentfulImageUrl(coverImage.url, '360', '200', 'fill'),
   };
   const scholarshipAffiliateLabel = getScholarshipAffiliateLabel();
-  const displayAffiliateScholarshipBlock =
-    scholarshipAffiliateLabel && scholarshipAmount && scholarshipDeadline;
```
mendelB: Hmm I wonder if we still want to use this, or an enhanced series of conditions before displaying the modal. Like what happens in instances where the campaign doesn't have the proper scholarship description / amount / deadline, or the campaign doesn't have an action that's a scholarship entry or RB?

Or I guess you could argue that if we're being referred by a valid affiliate, then we'd expect all this information to be available? (But I guess someone can still share an outdated URL, or manipulate query params...)

What do you think?

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
lindsayrmaher: ➕ def agree!

resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
lindsayrmaher: i think it would be safer to account for the possibility of glitches or the outdated url etc! perhaps using this to set the `showScholarshipModal` instead of directly using the helper function.

mendelB: Yeah, I think that sounds reasonable! You can even have a `useEffect`, which would `setShowModal(true)` _if_ these conditions pass.

```js
useEffect(() => {
  if ([all necessary info exists]) {
    setShowModal(true);
  }
}, []);
```

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -52,10 +52,17 @@ const ScholarshipInfoBlock = ({
       </div>
       <div className="pt-6 pb-3">
         <strong className="text-lg">
-          Welcome to DoSomething.org
-          {affiliateTitle ? ` via ${affiliateTitle.toUpperCase()}` : null}!
+          Welcome from
+          {affiliateTitle ? ` ${affiliateTitle.toUpperCase()}` : null}!
```
mendelB: Hmm maybe just to be safe, in case we don't have a corresponding affiliate in Contentful, we do something like:

```js
Welcome 
{affiliateTitle ? ` from ${affiliateTitle.toUpperCase()}` : `to DoSomething.org`}!
```

PR #1794, opened 2019-12-06
Adds Affirmation Message on Form Submission

Vmack:
### What's this PR do?

This pull request allows an affirmation text to be displayed on the modal once the user submits an email. 

This one should be smooth, I would flag anything that looks out of place or redundant. For example, am I making good use of the affirmation hooks I added? Is adding the conditional before the form a good idea? Should handleComplete be added on the button or the form? Is it a good idea to make changes to the `DismissableElement` or is there another way?(more background below)

`setShowElement(false)` has been removed from the `handleCompletion()` method in the `DismissableElement` to prevent the modal from closing automatically once the email is submitted so that the affirmation text will display. Therefore, the modal will only be dismissed when the close button is clicked. We tested it with the survey and it works fine but if this may cause any issues with other modals, let me know. 

### Relevant tickets

References [Pivotal #169325168](https://www.pivotaltracker.com/story/show/169325168).

image

### Checklist

### DIFF
--- a/resources/assets/components/utilities/CtaPopover/CtaPopover.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopover.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import './cta-popover.scss';
 
-const CtaPopover = ({ content, handleClose, title, children }) => {
+const CtaPopover = ({ content, handleClose, title }) => {
   return (
     <div className="cta-popover p-3 bordered rounded">
       <button
@@ -17,7 +17,7 @@ const CtaPopover = ({ content, handleClose, title, children }) => {
         {title}
       </h3>
       <p className="text-white mt-3">{content}</p>
-      {children}
+      {/* {children} */}
     </div>
   );
 };
@@ -26,7 +26,7 @@ CtaPopover.propTypes = {
   content: PropTypes.string.isRequired,
   handleClose: PropTypes.func.isRequired,
   title: PropTypes.string.isRequired,
-  children: PropTypes.object.isRequired,
+  // children: PropTypes.object.isRequired,
 };
 
 export default CtaPopover;
--- a/resources/assets/components/utilities/CtaPopover/CtaPopover.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopover.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import './cta-popover.scss';
 
-const CtaPopover = ({ content, handleClose, title }) => {
+const CtaPopover = ({ content, handleClose, title, children }) => {
   return (
     <div className="cta-popover p-3 bordered rounded">
       <button
@@ -17,7 +17,7 @@ const CtaPopover = ({ content, handleClose, title }) => {
         {title}
       </h3>
       <p className="text-white mt-3">{content}</p>
-      {/* { children } */}
+      {children}
     </div>
   );
 };
@@ -26,7 +26,7 @@ CtaPopover.propTypes = {
   content: PropTypes.string.isRequired,
   handleClose: PropTypes.func.isRequired,
   title: PropTypes.string.isRequired,
-  // children: PropTypes.object.isRequired,
+  children: PropTypes.object.isRequired,
 };
 
 export default CtaPopover;
--- a/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
@@ -1,4 +1,5 @@
 import get from 'lodash/get';
+import PropTypes from 'prop-types';
 import React, { useState } from 'react';
 import { RestApiClient } from '@dosomething/gateway';
 
@@ -8,9 +9,10 @@ import { tabularLog } from '../../../helpers/api';
 
 import './cta-popover-email-form.scss';
 
-const CtaPopoverEmailForm = () => {
+const CtaPopoverEmailForm = ({ handleComplete }) => {
   const [emailValue, setEmailValue] = useState('');
   const [errorResponse, setErrorResponse] = useState(null);
+  const [showAffirmation, setShowAffirmation] = useState(false);
   const handleChange = event => setEmailValue(event.target.value);
 
   const handleSubmit = event => {
@@ -40,7 +42,7 @@ const CtaPopoverEmailForm = () => {
       });
   };
 
-  return (
+  return !showAffirmation ? (
     <div className="cta-popover-email-form pt-4">
       {errorResponse && errorResponse.fields.email ? (
         <div className="text-red-500">{errorResponse.fields.email[0]}</div>
@@ -56,7 +58,7 @@ const CtaPopoverEmailForm = () => {
         <Button
           className="email-form__button"
           type="submit"
-          onClick={handleSubmit}
+          onClick={handleComplete}
         >
           Sign Up
         </Button>
@@ -74,7 +76,13 @@ const CtaPopoverEmailForm = () => {
         a year for important announcements.
       </p>
     </div>
+  ) : (
+    setShowAffirmation(<div>Thank You For Submitting Your Email</div>, true)
   );
+};
+
+CtaPopoverEmailForm.propTypes = {
+  handleComplete: PropTypes.func.isRequired,
 };
 
 export default CtaPopoverEmailForm;
--- a/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
@@ -29,6 +29,7 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
       })
       .then(response => {
         setShowAffirmation(true);
+        handleComplete();
         tabularLog(get(response, 'data', null));
 
         return response;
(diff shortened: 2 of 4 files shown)


resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -74,7 +77,13 @@ const CtaPopoverEmailForm = () => {
         a year for important announcements.
       </p>
     </div>
+  ) : (
+    <div>Thank You For Submitting Your Email</div>
```
mendelB: Do we want any styling padding for this, or how does this look plain? (It could be neat to attach a screenshot to the PR description to see the visual change here).

Maybe a `text-green-700` class to set the 'success color' for this message?

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -56,7 +59,7 @@ const CtaPopoverEmailForm = () => {
         <Button
           className="email-form__button"
           type="submit"
-          onClick={handleSubmit}
+          onClick={handleComplete}
```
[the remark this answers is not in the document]
lindsayrmaher: I have a question here on why we need to `handleSubmit` on the form and on the button itself? But agree with @mendelB on where to invoke handlecomplete!

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
Vmack: Yes!! @mendelB I definitely noticed that I forgot to add some styling. Am working on that now. I like the success color 🤑

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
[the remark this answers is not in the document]
Vmack: @mendelB ok that makes sense! And yes, I have the same question as @lindsayrmaher , why do we need the `handleSubmit()` in both places?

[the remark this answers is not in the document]
lindsayrmaher: it looks like it's only a required prop for the button if it's not `type="submit"`, so could we just keep the onSubmit on the form itself and remove this altogether? Less confusing perhaps!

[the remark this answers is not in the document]
mendelB: Perfectly valid point! We can just get rid of this prop altogether ☺️ 
Good catch!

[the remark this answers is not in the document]
Vmack: Yup, very valid point @lindsayrmaher thanks! I removed it from the form.

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -48,7 +49,7 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
       {errorResponse && errorResponse.fields.email ? (
         <div className="text-red-500">{errorResponse.fields.email[0]}</div>
       ) : null}
-      <form className="email-form form pb-2" onSubmit={handleSubmit}>
+      <form className="email-form form pb-2">
```
mendelB: @Vmack lol sorry for the ♻️but we're actually going to want to keep _this_ `onSubmit` event while dropping the `onClick` for the button below. The rationale being that the form's `onSubmit` will capture both submitting the form with the ENTER key _and_ submitting the form by clicking on the submit button 🎉

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -78,7 +79,7 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
       </p>
     </div>
   ) : (
-    <div>Thank You For Submitting Your Email</div>
+    <div className="color-success">Thank You For Submitting Your Email</div>
```
mendelB: The `color-success` class is part of the deprecated classes we're 🗑 in favor of tailwind. I think you can just use the `text-green-700` here for the success color though!

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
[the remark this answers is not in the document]
lindsayrmaher: I think we want to keep the `onSubmit` on the form but we can remove the `onClick` from the button since the `onSubmit` will take care of form submission for us!

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
lindsayrmaher: oh lol i didn't see this before i replied to the thread, but yah @mendelB 🙈

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
Vmack: @mendelB Ohhhh I didn't notice that, thanks!

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
Vmack: that makes sense, thanks guys!

resources/assets/components/utilities/CtaPopover/CtaPopover.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import './cta-popover.scss';
 
-const CtaPopover = ({ content, handleClose, title, children }) => {
+const CtaPopover = ({ content, handleClose, title }) => {
```
mendelB: BTW, you don't need to comment this `children` stuff out here, as it's the de facto functionality for the `CtaPopover` and won't affect prod anyway, since it's not being rendered.

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -74,7 +74,13 @@ const CtaPopoverEmailForm = () => {
         a year for important announcements.
       </p>
     </div>
+  ) : (
+    <div className="text-green-700">Thank You For Submitting Your Email</div>
```
mendelB: @Vmack Mind just attaching a screenshot to the PR description for this affirmation state (It's always super helpful for visual changes like this!)

resources/assets/components/utilities/DismissableElement/DismissableElement.js
```diff
@@ -10,7 +10,6 @@ const DismissableElement = ({ name, render }) => {
   const handleCompletion = () => {
     // Mark the element as "hidden" in local storage.
     setStorage(`hide_${name}`, 'boolean', true);
-    setShowElement(false);
```
mendelB: Last thing we'll want to do is per [@katiecrane's advice](https://dosomething.slack.com/archives/C3ASB4204/p1575913282113900?thread_ts=1575587210.077100&cid=C3ASB4204), to only mark the element as dismissed below in the [`handleDismissal` callback](https://github.com/DoSomething/phoenix-next/pull/1794/files#diff-5ce9a6caedc44ead64abc9f5c0c191c9R15) _if_ the element hasn't been hidden. Something like:
```js
if (!getStorage(`hide_${name}`, 'boolean')) {
}
```

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
Vmack: ooh yes, that's what I forgot to do

resources/assets/components/utilities/CtaPopover/CtaPopover.js
Vmack: ok great

resources/assets/components/utilities/DismissableElement/DismissableElement.js
Vmack: sweet!

Vmack: Should I leave the comments or un-comment everything?

mendelB: I think you can leave the comments!

resources/assets/components/utilities/DismissableElement/DismissableElement.js
```diff
@@ -10,18 +10,20 @@ const DismissableElement = ({ name, render }) => {
...
   };
 
   const handleDismissal = () => {
-    // Mark the element as "dismissed" in local storage & hide it.
-    setStorage(`dismissed_${name}`, 'timestamp', Date.now());
-    setShowElement(false);
+    if (!getStorage(`hide_${name}`, 'boolean')) {
+      // Mark the element as "dismissed" in local storage & hide it.
+      setStorage(`dismissed_${name}`, 'timestamp', Date.now());
+      setShowElement(false);
```
mendelB: @Vmack One quick update we'll need to make before we deploy is to move the `setShowElement(false)` to outside this conditional since we'll always want to visually hide the element if the dismissal method is invoked. So it should be:
```js
const handleDismissal = () => {
  if (!getStorage(`hide_${name}`, 'boolean')) {
      // Mark the element as "dismissed" in local storage.
  }

  setShowElement(false);
}

Vmack: oohhh that actually works so much better, thanks!

PR #1796, opened 2019-12-06
Spinner component with Emotion JS 👩‍🎤

weerd:
### What's this PR do?

This pull request adds a new `Spinner` component to help remove some of the previous `.spinner` styles strewn about and makes it easier to use Tailwind with the `Spinner` component! The previous approach with a `.spinner` class that then pulled styles from [Forge](http://forge.dosomething.org/#spinner), utilized SVG [SMIL](https://caniuse.com/#search=SMIL) animation, and required providing GIF fallback if the browser did not support SMIL animation. With this update, we switch to only using CSS Animation which has [wider support](https://caniuse.com/#search=css%20animation), allowing us to just use the one spinner SVG animated with CSS.

It also implements the use of [EmotionJS](https://emotion.sh/) to allow component specific styles, CSS that can be defined within a React component and thus scoped to the component. We discussed potentially using EmotionJS in conjunction with Tailwind in the [Tailwind RFC](https://github.com/DoSomething/rfcs/pull/5/files#diff-1a47195afb536d3073e941ad1adb335bR97-R98), to help compliment Tailwind when a component needs some very specific styles that don't make sense as distinct utility classes.

👀 

Example usage:

```js
<Spinner size="22px" className="flex justify-center p-16" />
```

We could eventually update this new `Spinner` component to allow support for passing in a custom background & foreground color (or using better names) to customize the spinner depending on whether it is showing up on a light or dark background ✨ 

### Relevant tickets

References [Pivotal #]().

### Checklist


resources/assets/components/actions/SelectionSubmissionAction/SelectionSubmissionAction.js
```diff
@@ -146,7 +145,7 @@ class SelectionSubmissionAction extends PostForm {
             if (selection || this.state.submitted) {
               return (
                 <div className="pb-3 px-3">
-                  <p className="submission-text uppercase">
+                  <p className="submission-text font-league-gothic leading-none text-4xl uppercase">
```
weerd: I was able to remove the attached SCSS file for this component by just adding the styling as Tailwind classes for the element.

resources/assets/components/PollLocator/PollLocator.js
```diff
@@ -133,9 +134,9 @@ class PollLocator extends React.Component {
     return (
       <div id="_vit" className="mx-auto">
         {/** this spinner will be auto replaced by the vit content once it loads */}
-        <div
-          className="spinner -centered"
-          style={{ height: 350, width: 640 }}
+        <Spinner
+          className="flex items-center justify-center"
+          style={{ height: 350 }}
```
weerd: Checked and the width isn't really required since it spans the full width of the parent container.

resources/assets/components/artifacts/Spinner/Spinner.js
```diff
@@ -0,0 +1,52 @@
...
+const rotation = keyframes`
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(359deg);
+  }
+`;
+
+const Spinner = ({ className, size, style }) => (
```
weerd: I made a decision to include a `style` attribute for the component. In some scenarios (embeds for example), we needed to specify a height and it seemed like overkill to utilize emotion to output this instead of just passing along a `style` object which defined the height as an inline style.

I played around with potentially having this be passed as `containerHeight`, and using emotion to output the inline style so we limit the extent of what gets affected, but based on how it gets used in an embed, I needed to define the style with `!important` and seemed kind of messy. Switching to just passing the `style` object is much cleaner, but then leave the component open to getting a bunch of other styles applied. 🤷‍♂ 

As a separate question, should I rename `style` to be `containerStyle` to make it obvious that the `div` is getting this style applied and not the `svg`? 🤔

weerd:
Oh, forgot a quick last question I'm curious what y'all think. I initially intended for the `Spinner` component to be just the `<svg />` with no surrounding `<div />` and the context of where it should be used would dictate whether it needed a containing `<div />`. Thus, placing it in the `/artifacts` directory made sense. Essentially a little spinner icon you include where needed.

However, after a few uses it became apparent that it benefits from having the surrounding `<div />` since it's usually centrally located and easier to do if in a `<div />` that can be easily passed some classes, so I decided it best to bake the `<div />` into the component. I wonder if now that means it should sit in the `/utilities` directory instead.

Thoughts? 🤔

resources/assets/components/artifacts/Spinner/Spinner.js
DFurnes: I think it makes sense as `style`, since it's getting applied to the top-level "container" for this component. 👍

DFurnes:
> Oh, forgot a quick last question I'm curious what y'all think. I initially intended for the `Spinner` component to be just the `<svg />` with no surrounding `<div />` and the context of where it should be used would dictate whether it needed a containing `<div />`. Thus, placing it in the `/artifacts` directory made sense. Essentially a little spinner icon you include where needed.
> 
> However, after a few uses it became apparent that it benefits from having the surrounding `<div />` since it's usually centrally located and easier to do if in a `<div />` that can be easily passed some classes, so I decided it best to bake the `<div />` into the component. I wonder if now that means it should sit in the `/utilities` directory instead.
> 
> Thoughts? 🤔

Yeah, the `artifacts` vs. `utilities` divide feels sorta subjective. I feel like the most important thing is which directory I'd "look" for this component under, and I could see a good reason I'd check either one. (And if that's the case, I wonder if it's even a distinction worth making...)

I could also see an argument that centering something should happen in the parent (since layout is really the parent's concern), and that'd let us sidestep the "classification" problem for now.

weerd:
We're gonna leave it in `/artifacts` for now. The `CloseButton` component has a similar structure and we've considered that an "artifact". If our definition changes, we can always just easily move it to the `/utilities` folder and call it a day 🌆

resources/views/app.blade.php
```diff
@@ -3,6 +3,10 @@
 @section('content')
     <div id="app">
         <div class="placeholder">
+            {{--
+                @TODO:forge-removal Unless server rendering React, will need
+                a different solution for this spinner once Forge is removed.
+            --}}
```
mendelB: @weerd I think, because we removed the `placeholder.scss`, and this is not using the spinner, this one is no longer getting the full page + centered styling, and is just displaying like so:
image

I guess we can probs just use Tailwind classes here to get it back to snuff?

PR #1810, opened 2019-12-11
Update CampaignDashboard with GraphQL, Tailwind, and Emotion.

DFurnes:
### What's this PR do?

This pull request updates the "Campaign Dashboard" component to be able to be queried via GraphQL, and styles it with Tailwind and Emotion. This allows us to centralize nearly everything you could want to know or change about this component in one file, nice!

Here's screenshots of this component at [small](https://user-images.githubusercontent.com/583202/70660633-ade9a580-1c30-11ea-8621-4ea6c9ca684c.png), [medium](https://user-images.githubusercontent.com/583202/70660626-aa561e80-1c30-11ea-9ae3-0dab158d546d.png), and [large](https://user-images.githubusercontent.com/583202/70660644-b5a94a00-1c30-11ea-9450-85faf2f2e0f9.png) screen sizes. You can also play around with it by visiting [`/us/blocks/2An4B0YlYsKYe4SCsUMWsm`](http://dosomething-phoenix-de-pr-1810.herokuapp.com/us/blocks/2An4B0YlYsKYe4SCsUMWsm) on the review app.

Allowing this component to be rendered via GraphQL is a pre-requisite to updating the `SixpackExperiment` block to be able to be rendered via GraphQL (since this is one of the few things we run experiments on). Updating styles in this component was just a bonus!

### Relevant tickets

References [Pivotal #170053882](https://www.pivotaltracker.com/story/show/170053882).

### Checklist


resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -114,10 +114,11 @@ class ContentfulEntry extends React.Component {
...
         );
 
+      case 'CampaignDashboard':
+        return <CampaignDashboard {...withoutNulls(json)} />;
+
       case 'campaignDashboard':
-        return (
-          <CampaignDashboard id={json.id} {...withoutNulls(json.fields)} />
-        );
+        return <CampaignDashboard {...withoutNulls(json.fields)} />;
```
lindsayrmaher: quick q, why are there two cases for this? and as a follow up, why do we need to use `json.fields` vs `json` in the second case?

resources/assets/components/utilities/CampaignDashboard/CampaignDashboard.js
```diff
@@ -1,9 +1,21 @@
-import React from 'react';
+/** @jsx jsx */
+
+import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
+import { css, jsx } from '@emotion/core';
```
lindsayrmaher: i didn't read through the PR when Diego added this, what does this add for us??? Intrigue!

lindsayrmaher: ps are we using this `jsx` anywhere?

resources/assets/components/ContentfulEntry/ContentfulEntry.js
DFurnes: Good question! This is so that we can support rendering both the [PHP Content API response](https://github.com/DoSomething/phoenix-next/blob/3acff74328348d91207a9a21fdcaaaa9c54c09cb/app/Entities/CampaignDashboard.php), which has things under a nested `fields` object, and the GraphQL `CampaignDashboard` type, which doesn't.

Once we query these via GraphQL everywhere, we can remove the old `json.fields` ones.

resources/assets/components/utilities/CampaignDashboard/CampaignDashboard.js
DFurnes: I was confused by this at first too! The [JSX pragma](https://emotion.sh/docs/css-prop#jsx-pragma) up above this import tells the compiler to turn JSX elements (like `<div css={dashboardGrid}>...</div>`) into `jsx(…)` instead of `React.createElement(…)`). That's also why the `React` import is no longer needed here.

We'll probably want to use [Emotion's Babel plugin](https://emotion.sh/docs/@emotion/babel-preset-css-prop) to handle this automatically in the future.

weerd: Yep! Exactly what @DFurnes said 💃 Agreed on adding the Babel plugin to improve this as we start using it more!

resources/assets/components/utilities/CampaignDashboard/CampaignDashboard.js
```diff
@@ -15,9 +27,29 @@ const CampaignDashboard = props => {
...
+    @media (min-width: 960px) {
+      grid-template-columns: 1fr 1fr 1fr 1fr;
+    }
+  `;
+
+  const shareGridElement = css`
+    @media (min-width: 600px) {
+      grid-column: span 2;
+    }
+  `;
```
weerd: So cool that these super custom rules for this component can just reside in here 🎉

resources/assets/components/utilities/CampaignDashboard/CampaignDashboard.js
```diff
@@ -15,9 +27,29 @@ const CampaignDashboard = props => {
...
+
   return (
-    <div className="dashboard text-black p-3">
-      <div className="dashboard__segment px-3">
+    <div className="bg-white text-black p-3" css={dashboardGrid}>
```
weerd: Not vital, but I wonder if we add an ID `campaign-dashboard`. I find it's sometimes nice to see the components labeled when inspecting the DOM. Might also be useful for ghost-inspector type tests.

DFurnes: Hmm, I can see that. Maybe a class or a data-element, though, so it doesn't conflict if we had more than one on a page at once?

DFurnes: Made [#170268857](https://www.pivotaltracker.com/story/show/170268857) to revisit this!

weerd: Ya, I was originally thinking `class` and then was thinking for better selecting in 👻 inspector maybe an ID. Class is probs best 👍

PR #1811, opened 2019-12-12
Sign Up Button on Scholarship Modal

lindsayrmaher:
### What's this PR do?

This pull request adds the sign up button to the scholarship modal!

👀 

Shout out @weerd for showing me how to send this through as a child of the modal and significantly simplifying this work! 

### Relevant tickets

References [Pivotal #169929205](https://www.pivotaltracker.com/story/show/169929205).

Desktop:
https://user-images.githubusercontent.com/15236023/70726813-aaa3f780-1ccc-11ea-8ed5-0ca048f850fd.png

Mobile:
https://user-images.githubusercontent.com/15236023/70726851-b7285000-1ccc-11ea-8b12-18ac9b45e996.png

### Checklist

### DIFF
--- a/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
+++ b/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
@@ -135,8 +135,15 @@ const MarqueeTemplate = ({
               scholarshipDeadline={scholarshipDeadline}
               scholarshipDescription={scholarshipDescription}
               utmLabel={scholarshipAffiliateLabel.toLowerCase()}
-              closeModal={closeModal}
-            />
+            >
+              <SignupButtonContainer
+                text={
+                  isScholarshipAffiliateReferral()
+                    ? SCHOLARSHIP_SIGNUP_BUTTON_TEXT
+                    : undefined
+                }
+              />
+            </ScholarshipInfoBlockQuery>
           </Modal>
         ) : null}
       </article>
--- a/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
+++ b/resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import { get } from 'lodash';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/LandingPage/templates/MarqueeTemplate.js
```diff
@@ -131,7 +131,18 @@ const MarqueeTemplate = ({
...
-            />
+            >
+              <div className="pt-6 w-2/3 lg:w-1/3">
+                <SignupButtonContainer
+                  className="w-full"
+                  text={
+                    isScholarshipAffiliateReferral()
+                      ? SCHOLARSHIP_SIGNUP_BUTTON_TEXT
+                      : undefined
+                  }
```
mendelB: Quick thought -- if this scholarship info block is _always_ intended to have signup functionality, and only shows when this is indeed a scholarship seeker, could we just add the signup button directly to the block with the `SCHOLARSHIP_SIGNUP_BUTTON_TEXT` _always_ set to display?

Vmack: Good point!

lindsayrmaher: That is a fair point! My struggle was with how to handle then getting all of the campaign information needed to sign up from the campaign state without needing to pass a bunch of props. I do think we could probably remove the ternary for the text on the button since that'll always be true

lindsayrmaher: and/or needing to re-write logic that already exists in the SignUpContainer/Button

mendelB: Totaly! since that question's already answered implicitly since the info block is being rendered, we can just hardcode the `text= SCHOLARSHIP_SIGNUP_BUTTON_TEXT` 😌

mendelB: > and/or needing to re-write logic that already exists in the SignUpContainer/Button

And anything that's happening within the Signup button itself will just continue to work since that's independent of the parent renderer. So we can just render the signup button directly from the scholarship info block with
```js
<SignupButtonContainer className="..." text={SCHOLARSHIP_SIGNUP_BUTTON_TEXT} />

lindsayrmaher: i swear that this didn't work yesterday..... but now it seems that it works just fine. I'm going to check with design to see if we always expect a button to be on the `scholarshipBlock` and if the answer is yes then i def think we should add it there instead :)

mendelB: It's my magic charm ...though it usually works the other way with me causing the thing to break so it must be the Holiday spirit 🎄

lindsayrmaher: quick update on this @mendelB 

I think to future proof this a bit, I'm going to leave the button as a child of the component. It sounds like there's a possibility we'll want to render the modal sans button for future features.

PR #1815, opened 2019-12-12
Adds ContentBlock content type documentation

aaronschachter:
### What's this PR do?

This pull request adds ~~two~~ a new documentation page~~s~~ to the Development > Content Types section:

* `ContentBlock` - Wanted to have this available for linking to from the School Finder configuration documentation over in #1813 

~~* `Page` - Wanted to have this available for linking to from the ContentBlock documentation, for context to what a block is~~

👀 

It felt tedious to list each field for these types, especially when they're available via GraphQL -- but I do think there's value in being able to link together these various content type / features pages.

### Relevant tickets

References [Pivotal #169659230](https://www.pivotaltracker.com/n/projects/2401401/stories/169659230).

### Checklist


docs/development/content-types/content-block.md
```diff
@@ -0,0 +1,15 @@
+# Content Block
+
+## Overview
+
+A [Page](development/content-types/page.md) Block displaying headline, subtitle, supertitle, copy, and image. Commonly used to list steps in a campaign.
```
mendelB: I wonder if using the Contentful Page example specifically here is slightly misleading since this is technically available from other 'page's which aren't Contentful `Page`s such as Cause Pages, Story Pages, etc. so I guess it's more of an independent block? (It can also be used as a gallery block node for example, though that might fit in below where we call out the more unorthodox usage abilities)

aaronschachter: I agree. I'm removing the Page documentation now, since it'd make the most sense to at least all different page types (and the `Page` isn't a parent type)

PR #1816, opened 2019-12-13
Tailwind gray switcheroo

weerd:
### What's this PR do?

This pull request continues the Tailwind updates and in particular switches the currently neutral gray palette to the new cool gray palette (default Tailwind grays) on Phoenix. Along with changes to the new darker yellow.

👀 

Example design with old neutral grays:
Old Neutral Grays

Example design with new cool grays:
New Cool Grays

### Relevant tickets

References [Pivotal #169578779](https://www.pivotaltracker.com/story/show/169578779).

### Checklist


tailwind.config.js
```diff
@@ -106,6 +96,20 @@ module.exports = {
...
+        '70': 'rgba(0,0,0,0.7)',
+        '90': 'rgba(0,0,0,0.9)',
+      },
+      'transparent-white': {
+        '10': 'rgba(255,255,255,0.1)',
+        '30': 'rgba(255,255,255,0.3)',
+        '50': 'rgba(255,255,255,0.5)',
+        '70': 'rgba(255,255,255,0.7)',
+        '90': 'rgba(255,255,255,0.9)',
+      },
```
weerd: While we want to refrain from using transparent black and white, we needed access to some values. An example that popped up during this work related to the border on the `Card` component. Depending on the background the card appeared on, an opaque gray color for the border looked on with some background colors, so it made sense to allow the borders to be a transparent black and thus absorb a bit of the background color in them and blend better in the context.

resources/assets/components/Revealer/revealer.scss
```diff
@@ -1,5 +1,7 @@
 @import '../../scss/next-toolbox';
 
+$light-background-color: #f7fafc; // theme('colors.gray.100')
```
DFurnes: Is this comment meant to replace the Sass variable? Should we add a `TODO` explaining it?

weerd: I can add a TODO. The problem is that using the `theme()` function inside `rgba()` throws an error, so ended up leaving this as the variable and using the hex version of the gray Tailwind color, but placing in component. Not ideal, but I think this whole Revealer can likely be deprecated at some point or reworked.

DFurnes: Makes sense! Let's just make sure we don't forget that context!

weerd: Sounds good! Added the TODO! Twas a good call :)

PR #1825, opened 2020-01-02
Tailwindifying PostCard and associated components

weerd:
### What's this PR do?

This pull request updates the `PostCard` component and a few associated components that reside inside the `PostCard` to all utilize Tailwind.

I checked against all the different post types and variations within those (for example anonymous posts) and all seemed to look as they should. I didn't add screenshots because the before and after should just look the same, just have better structure and style :) 

👀 

I removed the use of `BaseFigure` in the `PostCard` component for the UI elements, since it it not well suited for such use. Simplified it all a bit thanks to Tailwind and some flexbox and should now be easier to edit in the future.

### Relevant tickets

References [Pivotal #169578798](https://www.pivotaltracker.com/story/show/169578798).

### Checklist


resources/assets/components/utilities/PostCard/PostCard.js
```diff
@@ -77,36 +74,45 @@ const PostCard = ({ post, hideCaption, hideQuantity, hideReactions }) => {
   }
 
   return (
-    <Card className={`rounded h-full post-ornament-${post.type}`} key={post.id}>
```
mendelB: `post-ornament-${post.type}` - are we getting rid of that border-left for text posts?

resources/assets/components/utilities/ReactionButton/ReactionButton.js
```diff
@@ -25,38 +26,50 @@ const TOGGLE_REACTION = gql`
   }
 `;
 
-const ReactionButton = ({ post }) => (
+// @TODO:accessibilty We should establish a "focus" style for when the reaction
+// button is focused by a keyboard only user.
```
mendelB: 🙌

resources/assets/components/utilities/PostCard/PostCard.js
```diff
@@ -77,36 +74,45 @@ const PostCard = ({ post, hideCaption, hideQuantity, hideReactions }) => {
...
-          <h4>
-            {authorLabel}
-            {isStaff() ? <ReviewLink url={post.permalink} /> : null}
-            <PostBadge status={post.status} tags={post.tags} />
-          </h4>
-
-          {post.impact && !hideQuantity ? (
-            <p className="text-sm text-gray-600">{post.impact}</p>
-          ) : null}
+    <Card className="rounded h-full overflow-hidden" key={post.id}>
```
DFurnes: I know this isn't new to this PR, but I think `key` can be removed here.

resources/assets/components/utilities/PostCard/PostCard.js
```diff
@@ -77,36 +74,45 @@ const PostCard = ({ post, hideCaption, hideQuantity, hideReactions }) => {
...
+      <div
+        className={classnames('post bg-white flex flex-col h-full relative')}
```
DFurnes: No need for `classnames()` wrapper around this (can just be a plain string).

resources/assets/components/utilities/PostCard/PostCard.js
weerd: Ya. After trying to make it work, consulted with @lkpttn and @DFurnes and we agreed it's not worth it. The border color was nice on story pages, but then weird on campaign pages and vice versa when changed.

weerd: Good question though. I should have mentioned something, but that was part of the work I did in 2019 on this PR and not 2020, so I didn't remember it so well, since 2019 was sooooo long ago. 😜

mendelB: [border](https://en.wikipedia.org/wiki/Break_the_Border)

resources/assets/components/utilities/PostCard/PostCard.js
```diff
@@ -77,36 +74,45 @@ const PostCard = ({ post, hideCaption, hideQuantity, hideReactions }) => {
...
+              ) : null}
+
+              <PostBadge
+                className="ml-2"
+                status={post.status}
+                tags={post.tags}
+              />
+            </div>
+
+            {reactionElement}
```
DFurnes: It might be nice to inline `reactionElement` here since it's only used in one place (just for easier reading).

resources/assets/components/utilities/ReactionButton/ReactionButton.js
```diff
@@ -1,11 +1,12 @@
...
 import gql from 'graphql-tag';
+import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { Mutation } from 'react-apollo';
 import { propType } from 'graphql-anywhere';
 
-import './reaction.scss';
-import { BaseFigure } from '../Figure/Figure';
+import heartEmptyIcon from './heart_empty_icon.svg';
+import heartFilledIcon from './heart_filled_icon.svg';
```
DFurnes: How do we usually format assets like this? I feel like I've seen `kebab-case` or `camelCase` elsewhere.

resources/assets/components/utilities/ReactionButton/ReactionButton.js
DFurnes: I really like this pattern of "namespaced" TODOs! I added a [backlog ticket](https://www.pivotaltracker.com/story/show/170511156) for these that we can hopefully tackle in Q1 or Q2.

resources/assets/components/utilities/ReactionButton/ReactionButton.js
weerd: I've been doing snake case for the file names to help distinguish these from class names when you see them in a file.

weerd:
I was trigger happy and merged before @DFurnes comments. Will follow up with fixes in new PR :)

resources/assets/components/utilities/ReactionButton/ReactionButton.js
DFurnes: Hmm, okay. It looks like we've got a couple different approaches in use:

Screen Shot 2020-01-03 at 10 02 50 AM

I think it'd be nice to standardize on something, and I'd personally lean towards whichever one is the most common pattern we're following in the app.

resources/assets/components/utilities/ReactionButton/ReactionButton.js
weerd: Ah, good call! Thanks @DFurnes 💃

resources/assets/components/utilities/PostCard/PostCard.js
weerd: Would it maybe be useful to switch this to an ID attribute? Might be helpful for testing purposes and targeting specific posts?

DFurnes: I'm a little wary of adding an ID since we might have the same post appear twice on the page (for example in a user's submission gallery & the public gallery below). But I defs see the value in having something targetable… maybe a data attribute?

resources/assets/components/utilities/ReactionButton/ReactionButton.js
weerd: Yeah, I've been trying to update what I can as I go through with these tailwind updates! Probably should define some rule somewhere 🤷‍♂

resources/assets/components/utilities/PostCard/PostCard.js
weerd: Ah, true. There could be more than one on page. Ok, I think data attribute work tho 👌

PR #1826, opened 2020-01-02
Add Collection Page Cypress Test

mendelB:
### What's this PR do?

This pull request adds a super basic Cypress test for Collection Pages, which serves more broadly to ensure that our `PageQuery` functionality (querying pages directly from GraphQL) is working properly.

This ended up being slightly involved due to some configuration issues and some other unexpected issues, so please see below for more context and to ensure I've made the correct decisions here!

**_Apologies, this is a thesis level essay.... I wanted to give context, but ended up giving a WHOLE LOTTA context. Bless your heart if you actually make it through this. And hopefully this commit by commit walkthrough ends up being helpful! There was just a lot that came up!_**
- https://github.com/DoSomething/phoenix-next/pull/1826/commits/279cc89a2dbbf7d36228cef64d58715c56246e02 The first issue was that our Cypress tests run on their own dedicated ENV variable sandbox, so we needed to config the `CONTENTFUL_USE_PREVIEW_API` for it specifically. 
- ...though, the above really shouldn't have been an issue since this param defaults to `false`. The problem was that if this variable wasn't set, we'd be manually setting the param to `undefined` which is not a Boolean. https://github.com/DoSomething/phoenix-next/pull/1826/commits/9db1898acd4d532714d415bd847ca79278624f10 updates our `env` helper to allow a default value (akin to the [Laravel env helper](https://laravel.com/docs/5.5/helpers#method-env)), and https://github.com/DoSomething/phoenix-next/pull/1826/commits/6623f923ad812c7f8d5d227c1c894e6e2c180d27 adds the `false` default to various places we assign this param across Phoenix. ([See Slack convo](https://dosomething.slack.com/archives/C3ASB4204/p1577988210009500?thread_ts=1577986770.007800&cid=C3ASB4204))
- https://github.com/DoSomething/phoenix-next/pull/1826/commits/9516f7b8763827dba2eafeda83fa3c7b5623dcdb Since the collection page [makes use of](https://github.com/DoSomething/graphql/blob/2df05a996d26c578363e5ad69dd5643fac34d282/src/schema/contentful/phoenix.js#L236-#L237) Rich Text fields which are `JSON` scalars, we need to define a mock.
- Problem though, in the [`TextContent` component](https://git.io/Jepmu) which renders these values, this empty object was not getting passed through to the `RichTextDocument` component since it's missing the `nodeType` field. However, the `StandardMarkdown` component combusts due to this being an object and not a String. However, even if we do pass this through to the `RichTextDocument`, it'll combust due to invalid formatting / missing essential properties. This actually presents an interesting question: How do we handle mocking Rich Text fields? I was able to think of 3 options:
1. Manually add a mock of a valid Rich Text formatted object for these Rich text fields in the graphql fixture. Downside is that we'd need to do so for each query & field we're running in our tests which seems excessive.
2. Update the `TextContent` component to more specifically handle rendering so that in case it's presented a value that won't work with Rich Text or standard Markdown, it'll just return `null` and the tests will pass too! 
3. If we want to actually address mocking Rich text fields for our tests instead of relying on an empty JSON object or manually defining a mock for each query & field, we _could_ introduce a new `RichText` scalar to GraphQL and add a mock within the graphql fixture with the expected contentful Rich Text format. I just don't know if this is necessary, and this is def more involved...

I opted for the simpler option # 2 in https://github.com/DoSomething/phoenix-next/pull/1826/commits/bab39dc944ec618de7116ed1f2ed916e87844859

- Finally, https://github.com/DoSomething/phoenix-next/pull/1826/commits/c61a8fa08a02470ac1560066d8b1192f9259fab5 adds a basic Cypress test covering a non-existent collection page, an erroring one, & a valid one. 

### Relevant tickets

References [Pivotal #169781947](https://www.pivotaltracker.com/story/show/169781947).

### Checklist

### DIFF
--- a/cypress/fixtures/graphql.js
+++ b/cypress/fixtures/graphql.js
@@ -55,6 +55,7 @@ export const mocks = {
   // Custom scalars:
   AbsoluteUrl: () => 'https://www.example.com/',
   DateTime: () => new Date().toISOString(),
+  JSON: () => ({}),
 
   // Types:
   User: () => ({
--- a/cypress/integration/collection-page.js
+++ b/cypress/integration/collection-page.js
@@ -1,0 +1,33 @@
+/// <reference types="Cypress" />
+import { GraphQLError } from 'graphql';
+
+describe('Collection Page', () => {
+  // Configure a new "mock" server before each test:
+  beforeEach(() => cy.configureMocks());
+
+  it('Visit a non existent collection page', () => {
+    cy.mockGraphqlOp('CollectionPageQuery', {
+      collectionPageBySlug: null,
+    });
+
+    cy.visit('/us/collections/not-found');
+
+    cy.contains('Not Found');
+  });
+
+  it('Visit a collection page with a triggered GraphQL error', () => {
+    cy.mockGraphqlOp('CollectionPageQuery', {
+      collectionPageBySlug: new GraphQLError('Oops!... I Did It Again'),
+    });
+
+    cy.visit('/us/collections/error');
+
+    cy.contains('Something went wrong');
+  });
+
+  it('Visit a working collection page', () => {
+    cy.visit('/us/collections/collection');
+
+    cy.contains('h1', 'Hello World');
+  });
+});
--- a/cypress/support/index.js
+++ b/cypress/support/index.js
@@ -17,6 +17,7 @@ Cypress.on('window:before:load', window => {
     VOTER_REG_MODAL_ENABLED: false,
     SIXPACK_ENABLED: false,
     SIXPACK_BASE_URL: 'http://sixpack.test', // Our Sixpack service will throw an error if this isn't set.
+    CONTENTFUL_USE_PREVIEW_API: false,
   };
 
   // Remove built-in 'fetch' support since it cannot yet be mocked by Cypress.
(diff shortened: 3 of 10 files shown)


weerd:
> I opted for the simpler option # 2

I think this makes sense to me. Maybe if we see we need more robustness we can go with option #3, but defs smart to go with the simpler option for now. As for #1 would it need to be a mock for ever field or can it just be single mock that applies to all RichText fields (I don't know enough about these mocks, so not actually sure)?

mendelB:
> > I opted for the simpler option # 2
> 
> I think this makes sense to me. Maybe if we see we need more robustness we can go with option #3, but defs smart to go with the simpler option for now. As for #1 would it need to be a mock for ever field or can it just be single mock that applies to all RichText fields (I don't know enough about these mocks, so not actually sure)?

Sweet! And yeah, what was annoying about option # 1 is that it seems that it'll indeed need to be added _per_ field. So while we can add one defined mock helper, it'll still need to be set for every field we want to use it in, e.g. in this case:

```js
const richTextMock = () => ......

export const operations = {
...
CollectionPageQuery: {
  description: richTextMock(),
  content: richTextMock(),
...

weerd:
Gotcha, thanks for clarifying! Defs doesn't seem ideal, but other options do seem better 💃

PR #1836, opened 2020-01-06
Adds migration for currentSchoolBlock content type

aaronschachter:
### What's this PR do?

This pull request adds a migration for a new `currentSchoolBlock` content type that I've created in `dev` (and mistakenly in `master` too - where I first created it). I ran the migration on `dev` to validate the content type was created as expected.

Following the steps in the Workflow documentation in #1832 -- this PR also updates the page migration to include the `currentSchoolBlock` type to the list of content types available in the multi-value `blocks` field. However, it seems to include quite a bit of other updates that I haven't made, that don't seem to have been manually updated when devs made these changes 😥 

👀 

I considered whether to just manually add the `currentSchoolBlock` into the `content-types/page.js` vs migrate, but shouldn't this be the proper workflow? 

Upon merge and deploy, I'll manually take the following steps:

* [ ] `qa` - Run the migration to add the `currentSchoolBlock` content type 
* [ ]  `qa` - Alter the `page.blocks` field to to make the `currentSchoolBlock` content type available to select
* [x] `master` - run the migration to add `currentSchoolBlock` - don't need to do this because it already exists
* [ ]  `master` - Alter the `page.blocks` field to to make the `currentSchoolBlock` content type available to select

### Relevant tickets

References [Pivotal #170510799](https://www.pivotaltracker.com/story/show/170510799).

### Checklist

### DIFF
--- a/contentful/content-types/currentSchoolBlock.js
+++ b/contentful/content-types/currentSchoolBlock.js
@@ -1,0 +1,129 @@
+module.exports = function(migration) {
+  const currentSchoolBlock = migration
+    .createContentType('currentSchoolBlock')
+    .name('Current School Block')
+    .description(
+      "Displays the user's current school, or allows them to select it if not set.",
+    )
+    .displayField('internalTitle');
+
+  currentSchoolBlock
+    .createField('internalTitle')
+    .name('Internal Title')
+    .type('Symbol')
+    .localized(false)
+    .required(true)
+    .validations([])
+    .disabled(false)
+    .omitted(false);
+  currentSchoolBlock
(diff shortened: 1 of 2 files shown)


mendelB:
> However, it seems to include quite a bit of other updates that I haven't made, that don't seem to have been manually updated when devs made these changes 😥

Yeah that's a bummer, we should definitely be more careful about that! I updated the Page content type on `dev` to align better with the current code, so if you rerun the migration generation it should be all set!

aaronschachter:
Thanks @mendelB - I re-ran the dev migration and pushed up the update.

PR #1840, opened 2020-01-07
Adds Analytics Tracking to Scholarship CTA 

Vmack:
### What's this PR do?

This pull request adds analytics tracking to the `CtaPopoverEmailForm` so that GA/snowplow tracks the following on the newsletter cta: 

Who clicks on the form to fill it out
Who submits their email in the form
Form validation errors (if it makes sense)

👀

[Newsletter CTA Tech Spec](https://docs.google.com/document/d/1ADzYNZegztHOD7xs3KPky0b06nxG6HZE3-iEfakSGHA/edit?ts=5e0f756d#)

### Relevant tickets

References [Pivotal #169330883](https://www.pivotaltracker.com/story/show/169330883).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/CtaPopover/CtaPopoverButton.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopoverButton.js
@@ -12,9 +12,9 @@ const CtaPopoverButton = ({ buttonText, link }) => {
         category: 'site_action',
         target: 'button',
         verb: 'clicked',
-        noun: 'newsletter_cta',
+        noun: 'call_to_action',
         adjective: 'popover',
-        label: 'newsletter_cta__popover',
+        label: 'call_to_action_popover',
       },
       context: {
         url: link,
--- a/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
@@ -18,7 +18,7 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
   const handleFocus = () => {
     trackAnalyticsEvent({
       metadata: {
-        label: 'call_to_action_popover_email',
+        label: 'call_to_action_popover',
         adjective: 'email',
         category: 'site_action',
         noun: 'call_to_action_popover',
@@ -71,10 +71,10 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         }
 
         trackAnalyticsEvent({
-          context: { error, errorMessage },
+          context: { cta_type: newsletter_scholarships, error, errorMessage },
           metadata: {
-            label: 'call_to_action_popover_email',
-            adjective: 'call_to_action_popover_email',
+            label: 'call_to_action_popover',
+            adjective: 'call_to_action_popover',
             category: 'submission_failed',
             noun: 'form_submission',
             target: 'form',
--- a/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
@@ -17,6 +17,7 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
 
   const handleFocus = () => {
     trackAnalyticsEvent({
+      context: { cta_type: 'newsletter_scholarships' },
       metadata: {
         adjective: 'email',
         category: 'site_action',
--- a/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
@@ -71,14 +71,14 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         }
 
         trackAnalyticsEvent({
-          context: { errorMessage },
+          context: { error, errorMessage },
           metadata: {
             label: 'call_to_action_popover_email',
-            adjective: 'email_error',
-            category: 'site-action',
-            noun: 'call_to_action_popover',
+            adjective: 'call_to_action_popover_email',
+            category: 'submission_failed',
+            noun: 'form_submission',
             target: 'form',
-            verb: 'error_triggered',
+            verb: 'failed',
           },
         });
       });
--- a/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
@@ -6,14 +6,27 @@ import { RestApiClient } from '@dosomething/gateway';
 import Button from '../Button/Button';
 import { env, report } from '../../../helpers/index';
 import { tabularLog } from '../../../helpers/api';
-
 import './cta-popover-email-form.scss';
+import { trackAnalyticsEvent } from '../../../helpers/analytics';
 
 const CtaPopoverEmailForm = ({ handleComplete }) => {
   const [emailValue, setEmailValue] = useState('');
   const [errorResponse, setErrorResponse] = useState(null);
   const [showAffirmation, setShowAffirmation] = useState(false);
   const handleChange = event => setEmailValue(event.target.value);
+
+  const handleFocus = () => {
+    trackAnalyticsEvent({
+      metadata: {
+        label: 'call_to_action_popover_email',
+        adjective: 'email',
+        category: 'site_action',
+        noun: 'call_to_action_popover',
+        target: 'field',
+        verb: 'focused',
+      },
+    });
+  };
 
   const handleSubmit = event => {
     event.preventDefault();
@@ -32,6 +45,17 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         handleComplete();
         tabularLog(get(response, 'data', null));
 
+        trackAnalyticsEvent({
+          metadata: {
+            label: 'call_to_action_popover_email',
+            adjective: 'email',
+            category: 'authentication',
+            noun: 'form_submited_popover',
+            target: 'button',
+            verb: 'clicked',
+          },
+        });
+
         return response;
       })
       .catch(error => {
@@ -45,6 +69,18 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         if (window.ENV.APP_ENV !== 'production') {
           console.log('🚫 failed response? caught the error!', error);
         }
+
+        trackAnalyticsEvent({
+          context: { errorMessage },
+          metadata: {
+            label: 'submit_subscriptions_email_error',
+            adjective: 'email_error',
+            category: 'onboarding',
+            noun: 'form_submitted_popover_email_error',
+            target: 'button',
+            verb: 'error_triggered',
+          },
+        });
       });
   };
 
@@ -60,6 +96,7 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
           value={emailValue}
           placeholder="Enter your email address"
           onChange={handleChange}
+          onFocus={handleFocus}
         />
         <Button className="email-form__button" type="submit">
           Sign Up
--- a/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
@@ -18,7 +18,6 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
   const handleFocus = () => {
     trackAnalyticsEvent({
       metadata: {
-        label: 'call_to_action_popover',
         adjective: 'email',
         category: 'site_action',
         noun: 'call_to_action_popover',
@@ -34,7 +33,6 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
     trackAnalyticsEvent({
       context: { cta_type: 'newsletter_scholarships' },
       metadata: {
-        label: 'call_to_action_popover',
         adjective: 'call_to_action_popover',
         category: 'site_action',
         noun: 'submission',
--- a/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
@@ -31,6 +31,18 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
   const handleSubmit = event => {
     event.preventDefault();
 
+    trackAnalyticsEvent({
+      context: { cta_type: 'newsletter_scholarships' },
+      metadata: {
+        label: 'call_to_action_popover',
+        adjective: 'call_to_action_popover',
+        category: 'site_action',
+        noun: 'submission',
+        target: 'form',
+        verb: 'submitted',
+      },
+    });
+
     const client = new RestApiClient(`${env('NORTHSTAR_URL')}`);
 
     client
@@ -44,18 +56,6 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         setShowAffirmation(true);
         handleComplete();
         tabularLog(get(response, 'data', null));
-
-        trackAnalyticsEvent({
-          context: { cta_type: 'newsletter_scholarships' },
-          metadata: {
-            label: 'call_to_action_popover',
-            adjective: 'call_to_action_popover',
-            category: 'site_action',
-            noun: 'submission',
-            target: 'form',
-            verb: 'submitted',
-          },
-        });
 
         return response;
       })
--- a/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
@@ -71,7 +71,7 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         }
 
         trackAnalyticsEvent({
-          context: { cta_type: newsletter_scholarships, error, errorMessage },
+          context: { cta_type: 'newsletter_scholarships', error, errorMessage },
           metadata: {
             label: 'call_to_action_popover',
             adjective: 'call_to_action_popover',
--- a/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
+++ b/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
@@ -46,11 +46,12 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         tabularLog(get(response, 'data', null));
 
         trackAnalyticsEvent({
+          context: { cta_type: 'newsletter_scholarships' },
           metadata: {
-            label: 'call_to_action_popover_email',
-            adjective: 'email',
+            label: 'call_to_action_popover',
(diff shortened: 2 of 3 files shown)


resources/assets/components/utilities/CtaPopover/CtaPopoverButton.js
```diff
@@ -12,9 +12,9 @@ const CtaPopoverButton = ({ buttonText, link }) => {
         category: 'site_action',
         target: 'button',
         verb: 'clicked',
-        noun: 'call_to_action',
+        noun: 'newsletter_cta',
         adjective: 'popover',
-        label: 'call_to_action_popover',
+        label: 'newsletter_cta__popover',
```
mendelB: Wouldn't this be updating the `CtaPopoverButton` which we use for signup CTA's? I figure we'd want to instead add a new event to the `CtaPopoverEmailForm` to add the analytics for that CTA?

weerd: Oh ya! Good catch @mendelB

weerd: So should this even live in the button then? Or does it need to pass the payload to the button from each type of component using the button?

Vmack: gotcha! I'll change that

mendelB: > So should this even live in the button then? Or does it need to pass the payload to the button from each type of component using the button?

@weerd I think this one can remain in the button since the URL comes externally and is just passed through to the `context.url`, so this will trigger specifically for button CTA's whichever link it may be (signup or otherwise).

Vmack:
@mendel and @weerd I made some changes!

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -32,6 +45,17 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         handleComplete();
         tabularLog(get(response, 'data', null));
 
+        trackAnalyticsEvent({
+          metadata: {
+            label: 'call_to_action_popover_email',
+            adjective: 'email',
+            category: 'authentication',
```
weerd: Not sure if this would be an `authentication` category. It just happens to lead to authentication if user is not logged in, but if they are it just submits the email for the newsletter signup right? Maybe this should just be `site_action` or possibly a new category... `subscription`? Not sure. Might be useful to ask data team? cc @DoSomething/storm

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -45,6 +69,18 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         if (window.ENV.APP_ENV !== 'production') {
           console.log('🚫 failed response? caught the error!', error);
         }
+
+        trackAnalyticsEvent({
+          context: { errorMessage },
+          metadata: {
+            label: 'submit_subscriptions_email_error',
+            adjective: 'email_error',
+            category: 'onboarding',
```
weerd: Is this considered `onboarding`?

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
mendelB: I agree that this ought to be categorized as a `site_action`. I _think_ the `noun` can also be consistent across these events as `call_to_action_popover `, with the `verb` perhaps being `'submitted'` since it's a typical form submission and the `target` being `'form'`?

Vmack: @mendelB @weerd ok! I wasn't sure. I used the GA/GTM Inventory as a guide but I will update those changes now. Thanks guys!

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -45,6 +69,18 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
...
+          context: { errorMessage },
+          metadata: {
+            label: 'call_to_action_popover_email',
+            adjective: 'email_error',
+            category: 'site-action',
+            noun: 'call_to_action_popover',
+            target: 'form',
+            verb: 'error_triggered',
+          },
+        });
```
mendelB: 🤔 in regards to this one, it looks like this was modeled it off of [the `northstar_triggered_error_submit_[form_name]`](https://docs.google.com/spreadsheets/d/1hNZdkcS7orp7xU0mOHSRKPodNoif0EP5UC2YwOnDmPI/edit#gid=1891031727&range=63:63) event in the sheet? Which makes good sense! 

Though, one interesting difference here is that this event in Northstar is triggered by the front end library preventing the actual error submission to the backend, by surfacing the validations straight away, where we then track this event. But here we're actually getting the errors from the API validation on Rogue. 

Would this be more akin to [the `phoenix_failed_post_request`](https://docs.google.com/spreadsheets/d/1hNZdkcS7orp7xU0mOHSRKPodNoif0EP5UC2YwOnDmPI/edit#gid=1891031727&range=52:52) we track for failed POST request on Phoenix broadly? (I think mostly for failed RB post submissions.)

Perhaps it's not the most helpful event for our purposes since it's so general...

Hmm so I'm thinking though, maybe a more specific version of the above event. Or, maybe we immitate what happens on the Northstar end when we fail validation on the _backend_, where we track the [`northstar_failed_validation` event](https://docs.google.com/spreadsheets/d/1hNZdkcS7orp7xU0mOHSRKPodNoif0EP5UC2YwOnDmPI/edit#gid=1891031727&range=59:59). (Or maybe even a more specific version of this, where we add the form type as an adjective -- `phoenix_failed_validation_call_to_action_popover`, with `context.ctaType: 'newsletter'` or something)

Ok, sorry this is a lot! That's my full train of thought here.

Agreed -- I think this could also be `'site_action'`, and I wonder if we can also borrow the general `phoenix_failed_post_request`https://docs.google.com/spreadsheets/d/1hNZdkcS7orp7xU0mOHSRKPodNoif0EP5UC2YwOnDmPI/edit#gid=1891031727&range=52:52

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -45,6 +69,18 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         if (window.ENV.APP_ENV !== 'production') {
           console.log('🚫 failed response? caught the error!', error);
         }
+
+        trackAnalyticsEvent({
+          context: { errorMessage },
+          metadata: {
+            label: 'call_to_action_popover_email',
+            adjective: 'email_error',
+            category: 'site-action',
```
weerd: Should be snake_case!

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
weerd: > Hmm so I'm thinking though, maybe a more specific version of the above event. Or, maybe we immitate what happens on the Northstar end when we fail validation on the backend, where we track the northstar_failed_validation event. (Or maybe even a more specific version of this, where we add the form type as an adjective -- phoenix_failed_validation_call_to_action_popover, with context.ctaType: 'newsletter' or something)

I think this makes sense to me. So this would be a failure with the submitted email right? Like if submitting `bobsmith@gmailcom`?

mendelB: Exactly. Though there is the possibility of failing due to request throttling, or some other issue... Maybe we should track separate events for those scenarios? Or a more general catch all instead of `validation`?

katiecrane: I'm wondering if the analytics tracking needs to be so fine grained as to track which sorts of errors people are getting? Seems to me like a general `error` bucket could be enough?

mendelB: That's a good point! Maybe a nice general new `phoenix_failed_form_submission_[form_type]` or something of that nature, which indicates a form submission to the backend, and an error response? A bit more specific than the `phoenix_failed_post_request`, but general enough for any errors a form submission yields?

katiecrane: Love that idea @mendelB ! I confirmed with Diego D that that is specific enough and we don't need to break it down more than that

Vmack: thanks guys! I'm still a bit confused...so the label and noun would be `phoenix_failed_form_submission_popover_email `?

katiecrane: @mendelB you meant for that to be the label right? should any of the other fields change?

mendelB: @katiecrane We ended up discussing this more in https://dosomething.slack.com/archives/CJU7WCQG4/p1578933216000900

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -45,6 +69,18 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
...
         }
+
+        trackAnalyticsEvent({
+          context: { cta_type: 'newsletter_scholarships', error, errorMessage },
+          metadata: {
+            label: 'call_to_action_popover',
+            adjective: 'call_to_action_popover',
+            category: 'submission_failed',
+            noun: 'form_submission',
+            target: 'form',
```
mendelB: Should the target here perhaps be `submission` or `form_submission` so that the GA action ends up being `Submission Failed` or `Form Submission Failed`?

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -45,6 +69,18 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         if (window.ENV.APP_ENV !== 'production') {
           console.log('🚫 failed response? caught the error!', error);
         }
+
+        trackAnalyticsEvent({
+          context: { cta_type: 'newsletter_scholarships', error, errorMessage },
+          metadata: {
+            label: 'call_to_action_popover',
+            adjective: 'call_to_action_popover',
+            category: 'submission_failed',
```
mendelB: Judging by our other categories in the list, I wonder if this might just also be a `site_action` 🤔 (Kinda like how [the `phoenix_failed_post_request`](https://docs.google.com/spreadsheets/d/1hNZdkcS7orp7xU0mOHSRKPodNoif0EP5UC2YwOnDmPI/edit#gid=1891031727&range=52:52) is a `campaign_action`)

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
```diff
@@ -32,6 +45,17 @@ const CtaPopoverEmailForm = ({ handleComplete }) => {
         handleComplete();
         tabularLog(get(response, 'data', null));
 
+        trackAnalyticsEvent({
+          metadata: {
+            label: 'call_to_action_popover_email',
```
mendelB: I _think_ based on `email` now just being _field_ descriptive and not CTA descriptive as discussed [in Slack](https://dosomething.slack.com/archives/CJU7WCQG4/p1578944679010300?thread_ts=1578933216.000900&cid=CJU7WCQG4) -- we should just have the label be `call_to_action_popover` like the other events?

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
weerd: I think `form_failed` is probs fine. We have `form` as the target in other locations, so I think it's fine to follow that. I think intention with `target` it the actual item performing "verb".

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
weerd: I think that makes the most sense.

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
mendelB: Ah, what's confusing to me with this event is the `noun` -- is it the `form` that `failed`, or is it the `form_submission` that `failed` as opposed to a `form` that was `submitted`. 😖

So like the `form_submitted` Action makes sense, but in this case wouldn't it be the `form_submission` or `submission` that `failed`?

omg I think my philosophy background is not holding me in good stead here 🙃

Vmack: I think it would be `form_submission` that fails because it specifies where the failure is coming from especially because the submit handler is attached to the form and not the submit button it just flows better.

weerd: There should probably be 3 events here:

1. the form button is clicked (or form is submitted)
2. the form submission data is accepted
3. the form submission data is rejected

The first one always fires. The second fires if the form data is successful, otherwise the third one fires since the form data is rejected.

Based on the above, maybe you are right and the second/third should be `form_submission` as the item being acted upon.

But we probably need the 3 events.

🤷‍♂

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
katiecrane: a-ha! thanks @mendelB !

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
mendelB: I think that makes good sense!

katiecrane: @weerd I think event 2 was not one of the ones specified on the pivotal card for this work so I think that one might be out of scope?

weerd: @katiecrane true, but the ticket didn't specify the naming of all the events etc anyhoo, which is what's got us into this mess of back-and-forth with naming 😖 

Event number 2 does exist though since it's being defined in the response _after_ having clicked and submitted the data, along with its counterpart in the `catch` if it's an error and data is rejected. So if anything I think event number 1 is the one that is missing.

katiecrane: @weerd so I'm thinking if we have 2 & 3 already, then we don't need 1 because we can infer than from 2 & 3 right?

weerd: Ya, potentially... in the past though we've wanted to know when something is clicked/submitted which is why I included it. But if not crucial can probs ignore 🤷‍♂ Not my call though! 🚫 ☎️

katiecrane: I've asked Diego D about this as well!

mendelB: BTW to clarify some context -- I think this event _was_ intended to be number 1 but would just need to be moved to the `handleSubmit` at the top instead of in the success callback.

Vmack: ok, that makes sense. I moved it.

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
Vmack: done!

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
katiecrane: ah perfect, I think that's resolved then?

weerd:
Ok, I'm super confused with all the back-and-forth and there's still some inconsistencies among the object data 😬 

Based on how we generate event names and data, and going off of the GA/GTM Inventory, this is what I think they should be:

```js
// handleFocus Event:
// eventName: phoenix_focused_call_to_action_popover_email
{
  context: { cta_type: 'newsletter_scholarships' },
  metadata: {
    adjective: 'email',
    category: 'site_action',
    noun: 'call_to_action_popover',
    target: 'field',
    verb: 'focused',
  },
}

// handleSubmit Event:
// eventName: phoenix_submitted_call_to_action_popover_email
{
  context: { cta_type: 'newsletter_scholarships' },
  metadata: {
    adjective: 'email',
    category: 'site_action',
    noun: 'call_to_action_popover',
    target: 'form',
    verb: 'submitted',
  },
}

// failed submission Event:
// eventName: phoenix_failed_call_to_action_popover_submission_email
{
  context: { cta_type: 'newsletter_scholarships', error, errorMessage },
  metadata: {
    adjective: 'email',
    category: 'site_action',
    noun: 'call_to_action_popover_submission',
    target: 'form',
    verb: 'failed',
  },
}
```

If it's deemed the `adjective` is unnecessary here, I'd recommend placing the information that it is an email field within the `context` so it's obvious this was an email field.

mendelB:
@weerd (as discussed a touch IRL just to document) - This seems spot on to me! And indeed, [we opted](https://dosomething.slack.com/archives/CJU7WCQG4/p1578937339008200?thread_ts=1578933216.000900&cid=CJU7WCQG4) to not use the `email` adjective, besides for in the `focused` event, where we need to actually describe the field itself. 

We're comfortable with `cta_type: 'newsletter_scholarship'` in context to denote the type of form being interacted with here (and it being an `email` field by default. Though if anyone feels strongly that a `field: email` would be helpful within context, we should add that as well!)

katiecrane:
So the final decision is to go with exactly @weerd wrote up above BUT do not include the `adjective` field? Is that correct?

mendelB:
> So the final decision is to go with exactly @weerd wrote up above BUT do not include the `adjective` field? Is that correct?

That sounds good to me! 
The one thing I'd point out is that the `failed` event is slightly different than the one we discussed in [this thread](https://dosomething.slack.com/archives/CJU7WCQG4/p1578934727002000?thread_ts=1578933216.000900&cid=CJU7WCQG4) but I feel good about this change personally.

Vmack:
@katiecrane correct but we'll only be using adjective for the `focus` event

resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js
ddonizeth: Apologies for chiming in late (was OOO yesterday). I agree that we don't need to have the 3 events set up (event 1 could be inferred by 2 and 3). So we can limit the scope of this work!

Vmack:
yesssssss! 💃🏽

PR #1841, opened 2020-01-07
Quick fix to button styles when loading.

weerd:
### What's this PR do?

This pull request fixes a quick bug @mendelB noticed when a button is in a "loading" state after a submission.

👀 

### Relevant tickets

🐛


Vmack:
Can show a photo?

weerd:
> Can you show a photo?

I didn't include one cause it's during a state transition, so not easiest to capture; plus just a pretty straight-forward text color fix :)

PR #1843, opened 2020-01-07
Site navigation link behavior updates

weerd:
### What's this PR do?

This pull request updates the behavior of the main `SiteNavigation` component, so that the Causes subnav is toggleable on small and medium devices (most common touch device sizes), but clickable on device sizes large+ and heads over to `/explore` campaigns.

The behavior was accomplished by using the `react-media` package, which conveniently allows rendering a different `<a>` element that uses a different method depending on the media query context; easy-peasy 😄 

Example test run:
image

👀 

This also removes the link from the subnav "Causes" title since it would be redundant and ultimately was meant to be temporary until we got the correct functionality in. 

### Relevant tickets

References [Pivotal #170319248](https://www.pivotaltracker.com/story/show/170319248).

### Checklist

### DIFF
--- a/cypress/integration/site-navigation.js
+++ b/cypress/integration/site-navigation.js
@@ -140,7 +140,7 @@ describe('Site Navigation', () => {
         '/us/facts/test-11-facts-about-testing',
       );
 
-      // Mouseover the Causes main nav item and assert it is visible:
+      // Mouseover the Causes main nav item and assert the subnav is visible:
       cy.get('#main-nav__causes')
         .trigger('mouseover')
         .should('be.visible');
--- a/cypress/integration/site-navigation.js
+++ b/cypress/integration/site-navigation.js
@@ -26,7 +26,7 @@ const allSizes = getViewportSizes();
 
 const smallAndMediumSizes = getViewportSizes(['xs', 'sm', 'md']);
 
-const largePlusSizes = getViewportSizes(['lg', 'xl', 'xxl']);
+const largeSizes = getViewportSizes(['lg', 'xl', 'xxl']);
 
 describe('Site Navigation', () => {
   // Configure a new "mock" server before each test:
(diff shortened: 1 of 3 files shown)


cypress/integration/site-navigation.js
```diff
@@ -127,4 +129,26 @@ describe('Site Navigation', () => {
...
+
+      // Go to an example site page:
+      cy.withState(exampleFactPage).visit(
+        '/us/facts/test-11-facts-about-testing',
+      );
+
+      // Mouseover the Causes main nav item and assert it is visible:
+      cy.get('#main-nav__causes')
+        .trigger('mouseover')
+        .should('be.visible');
```
mendelB: What is this case testing, wouldn't it be visible in all viewports automatically, or is this just a regular 'make sure it exists' assertion?

weerd: Well, I wanted to make sure that on large sizes and up, the subnav shows on "hover", since the behavior is changed now and on large sizes +, clicking on the main nav item goes to it's `href` but should still show the subnav on hover. Should I change the comment to make it clearer or maybe this isn't useful? 🤷‍♂

resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -168,36 +169,49 @@ class SiteNavigation extends React.Component {
...
-                id="main-nav__causes"
-                href="/"
-                onClick={e =>
-                  this.handleOnClickToggle(e, 'CausesSubNav', {
-                    label: 'causes',
-                  })
-                }
+              <Media
+                queries={{
+                  large: '(min-width: 960px)',
```
weerd: I think for a future update it would be nice to universalize the `getViewportSizes()` function somewhere as a helper, and use here to pull the Tailwind defined `lg` size from the config.

cypress/integration/site-navigation.js
mendelB: Ahhh gotcha, thanks for clarifying! That'd be nice -- maybe just including `assert the subnav is visible`? I think that'd be helpful for ole mendel

weerd: ya, makes sense!

PR #1844, opened 2020-01-08
Scholarship modal loading time updates

lindsayrmaher:
### What's this PR do?

This pull request updates the query logic to `unwrap` the entire scholarship modal from a Graphql `Query` component and use the `useQuery` hook instead. This let's us consolidate into a single file and load most of the data on the modal more quickly, while will utilizing the `react-apollo` library.

Check out the screenshot of the placeholder for the affiliate title space and action type, and make sure the logic for the updated scholarship block is looking good.

This initial [slack convo](https://www.pivotaltracker.com/story/show/169924504/comments/209498090), as well as a follow up [here](https://dosomething.slack.com/archives/C3ASB4204/p1578501099013700)!

Placeholder visuals

### Relevant tickets

References [Pivotal # 170295025](https://www.pivotaltracker.com/story/show/170295025).

### Checklist


resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -24,132 +50,160 @@ Header.defaultProps = {
...
+  const affiliateTitle = get(data, 'affiliate.title');
+  const actions = get(data, 'actions', []);
+  const actionItem = actions.find(
+    action => action.scholarshipEntry && action.reportback,
+  );
+  const actionType = get(actionItem, 'actionLabel', '');
+
+  if (error) {
+    return <ErrorBlock error={error} />;
+  }
```
mendelB: I wonder if we need to error out of the entire info block if the query fails, or, since it's just two fields, we can display some default or just nothing if we've got an error? 🤔 

(e.g. just display `Welcome to DoSomething.org!` for the affiliate title if the query fails)

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -24,132 +50,160 @@ Header.defaultProps = {
...
+                  </p>
+                </div>
+              ) : null}
+              {isLoaded ? (
+                <div className="lg:w-1/2 lg:float-right">
+                  <Header content="Action Type" />
+                  <p className="pb-2">{actionType}</p>
+                </div>
+              ) : (
+                <PlaceholderText size="medium" />
```
mendelB: Does this not need to be nested inside the `div` to get the `lg:w-1/2 lg:float-right` styling?

lindsayrmaher: it didn't look crazy out of place, but def better to put it inside to be safe!

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
lindsayrmaher: That's fair, I wasn't using an error block before but had noticed we had them in other Query Components. I have some checks in place right now for the affiliate title and the action type, so we should be good with those!

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -74,10 +73,6 @@ const ScholarshipInfoBlock = ({
   );
   const actionType = get(actionItem, 'actionLabel', '');
 
-  if (error) {
-    return <ErrorBlock error={error} />;
-  }
-
```
mendelB: Ok, now I'm just being annoying, but I wonder if it might still be worth logging and reporting this error like we do from the `ErrorBlock`
https://github.com/DoSomething/phoenix-next/blob/7837ca6ad174868a3ad9d75b9ae3276440166dd3/resources/assets/components/blocks/ErrorBlock/ErrorBlock.js#L13-L16
(maybe without the `[ErrorBlock]` bit in the console log)
This way the request can silently fail without erroring out the info block, but we'd still know that the error took place 🤔

lindsayrmaher: 😂 i like this idea! I'm going to add this in before merging

PR #1559, opened 2019-08-22
CompanyPage Component Styling & Quintet Gallery Option

jocoso:
### What does this PR do?

This PR introduces the CompanyPage component into Phoenix. CompanyPage is the template used on pages on the about '/about' pages on our site. 

### Any background context you want to provide?

Screen Shot 2019-07-29 at 4 06 14 PM

Screen Shot 2019-08-22 at 2 58 07 PM

Screen Shot 2019-08-22 at 2 58 18 PM

### What are the relevant tickets/cards?

Refs [#167297536](https://www.pivotaltracker.com/n/projects/2019313/stories/167297536)

### Checklist

* [ ] Documentation added for new features/changed endpoints.
* [ ] Added appropriate feature/unit tests.
* [ ] Added screenshot to PR description of related front-end updates on **small** screens.
* [ ] Added screenshot to PR description of related front-end updates on **medium** screens.
* [ ] Added screenshot to PR description of related front-end updates on **large** screens.

PR #1847, opened 2020-01-10
Properly determine 'refer a friend' default campaign ID

mendelB:
### What's this PR do?

This pull request actually fixes the bug attempted to be squashed here https://github.com/DoSomething/phoenix-next/pull/1559#issuecomment-572708917 to ensure we use a test campaign ID available on local & dev environments which should be using Rogue's dev environment, so that the referral link is valid, and doesn't break the embed.

👁 

Unfortunately, this means we need to move this out of the `/constants` file so that we have the `window.ENV` in scope to determine the `APP_ENV` config variable.

### Relevant tickets
https://dosomething.slack.com/archives/C3ASB4204/p1578679794042200

### Checklist

### DIFF
--- a/resources/assets/components/pages/ReferralPage/Beta/BetaPage.js
+++ b/resources/assets/components/pages/ReferralPage/Beta/BetaPage.js
@@ -6,7 +6,6 @@ import ErrorPage from '../../ErrorPage';
 import MoneyHandImage from './money-hand.svg';
 import { query, env } from '../../../../helpers';
 import CampaignLink from './BetaPageCampaignLink';
-import { REFERRAL_CAMPAIGN_IDS } from '../../../../constants';
 
 const REFERRAL_PAGE_USER = gql`
   query ReferralPageUserQuery($id: String!) {
@@ -24,16 +23,11 @@ const BetaPage = () => {
     return <ErrorPage />;
   }
 
-  // The Referral Campaign ID defaults to the Teens for Jeans campaign,
-  // or a test campaign for development environments.
-  const useDevCampaign = ['local', 'development'].includes(env('APP_ENV'));
-  const DEFAULT_REFERRAL_CAMPAIGN_ID = useDevCampaign ? '9001' : '9037';
-
-  const queryCampaignId = query('campaign_id');
-
-  const campaignId = REFERRAL_CAMPAIGN_IDS.includes(queryCampaignId)
-    ? queryCampaignId
-    : DEFAULT_REFERRAL_CAMPAIGN_ID;
+  // The Refer a Friend feature is limited to the Teens for Jeans campaign,
+  // or a test campaign for development environments. (hardcoded by design https://git.io/JvJ4H).
+  const campaignId = ['local', 'development'].includes(env('APP_ENV'))
+    ? '9001'
+    : '9037';
 
   return (
     // We *do not* render a SiteNavigationContainer here to avoid losing the referral metadata (see https://git.io/JeX2A).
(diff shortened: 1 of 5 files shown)


resources/assets/components/pages/ReferralPage/Beta/BetaPage.js
```diff
@@ -3,13 +3,10 @@ import gql from 'graphql-tag';
...
 import ErrorPage from '../../ErrorPage';
-import { query } from '../../../../helpers';
 import MoneyHandImage from './money-hand.svg';
+import { query, env } from '../../../../helpers';
 import CampaignLink from './BetaPageCampaignLink';
-import {
-  REFERRAL_CAMPAIGN_IDS,
-  DEFAULT_REFERRAL_CAMPAIGN_ID,
-} from '../../../../constants';
+import { REFERRAL_CAMPAIGN_IDS } from '../../../../constants';
```
aaronschachter: I get that that this is looking to stay backwards compatible with the previous iteration of Refer A Friend, where we had multiple campaign ID's to display the `ReferralPageBanner` for within the signup Affirmation.... but I'm wondering if this should all be simplified by just changing line 34 to
```
const campaignId =  ['local', 'development'].includes(env('APP_ENV')) ? '9001' : '9037';
```
and removing the `REFERRAL_CAMPAIGN_IDS` from `constants`, as well as checks for a query string campaign ID, which don't seem applicable anymore.

mendelB: That would definitely simplify this! I remember that we also maintained this query param functionality since @ngjo had mentioned probably opening this feature back up to more campaigns in the future, which would just mean adding more campaign ID's to the list instead of building that functionality back. 

But I can understand wanting to just keep things simpler by removing that functionality until that's indeed the case if you think that's wiser?

weerd: Just noting here that I started a joint [RFC](https://github.com/DoSomething/rfcs/pull/8) with @mendelB to hopefully address the pain points with having to hard-code campaign IDs for different environments:

aaronschachter: I'd be ok with leaving it in if we commented what the query `campaign_id` is for -- but I lean towards removing the functionality, because as is, I think the code is very confusing.  

If we go back to supporting multiple hand-picked campaigns for referral pages, I think our best bet would be to add a new boolean property like `display_referral_page` to a Rogue campaign, and expose in GraphQL. If the beta page contains a `campaign_id` query, we'd check the campaign ID in GraphQL to see whether its a valid referral page campaign (we could also use this in the Campaign page to determine whether to display the Referral Page banner in the Affirmation)

mendelB: Smart! Forces ourselves to properly formalize this feature, as opposed to continuing to hardcode / hack it along. I'm convinced! I'll remove this functionality and implement your suggestion above.

mendelB:
https://github.com/DoSomething/phoenix-next/pull/1847/commits/5114d81455271d3bae948fcda341756d17ac4295 ... https://github.com/DoSomething/phoenix-next/pull/1847/commits/f56539baeddadc1f9a159636152d638edeb9261c Updates as per https://github.com/DoSomething/phoenix-next/pull/1847#discussion_r365417973 to remove the `campaign_id` query param functionality completely from the referral feature and simplify the hardcoded campaign ID.

PR #1842, opened 2020-01-07
Adds currentSchoolBlock content type support

aaronschachter:
### What's this PR do?

This pull request renames the `utilities/SchoolFinder` component as `blocks/CurrentSchoolBlock` and adds support for including  a School Finder by creating a Current School Block entry in Contentful (added in #1836), in addition to the current `ContentBlock.additionalContent` method currently in use on production.

### How should this be reviewed?

I've added a Current School Block entry to the end of the blocks in the [Example School Finder Campaign](https://dev.dosomething.org/us/campaigns/test-school-finder-campaign/action), which we can use to test this branch in a review app.  

### Any background context you want to provide?

Holding off on making changes to the Cypress fixtures for now to help keep this PR reviewable. The current method of rendering School Finders via `ContentBlock.additionalContent` is still the only method used on production, so ideally in a future PR we'll test fixtures that use both ways (and then can remove support for `additionalContent` once we migrate the current production School Finders into Current School Block entries.

### Relevant tickets

References [Pivotal #170510799](https://www.pivotaltracker.com/story/show/170510799).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #1848, opened 2020-01-10
Fixes School Impact in CurrentSchoolBlock

aaronschachter:
### What's this PR do?

This pull request fixes a bug from #1842:  I missed adding the `actionId` Contentful field in the `app/Entities/CurrentSchoolBlock`, which was causing the School Impact to no longer display.

This PR also updates Cypress tests accordingly:

* Renames `school-finder` to `current-school-block`
* Adds tests for Current School Blocks with and without an Action ID field set

👀 

A big thank you to @ngjo for testing and finding this bug! 🙇 

### Relevant tickets

References [Pivotal #170510799](https://www.pivotaltracker.com/n/projects/2401401/stories/170510799/comments/210297744).

### Checklist


cypress/integration/current-school-block.js
```diff
@@ -25,12 +26,12 @@ const exampleAction = {
 };
 
 const schoolFinderConfig =
-  exampleSchoolFinderCampaign.campaign.pages[0].fields.blocks[0].fields;
+  exampleCurrentSchoolBlockWithAction.campaign.pages[0].fields.blocks[0].fields;
```
mendelB: This makes me wonder if at some point it would be cleaner for us to test blocks like these in isolation using the `/blocks/:id` route (though we'd then have to configure the Graphql mocks instead of the Contentful fixtures), or is it still preferable for us to test actions within the context of a campaign page?

And maybe a moot point once we move everything to GraphQL and it's easier to mock specific things like this?

mendelB:
Re the build failure: https://github.com/DoSomething/phoenix-next/pull/1847 should have the Cypress tests working again once that's merged!

PR #1849, opened 2020-01-10
'View more' drawer scholarship - Part One

lindsayrmaher:
### What's this PR do?

This pull request is adding an improvement for the mobile experience of the users seeing scholarship modals. On mobile, the non-essential details of the scholarship card will be hidden under a 'view more' drawer, and when clicked the drawer will slide down to show the rest. 

This should be one of two PRs on this feature add. As of right now, the 'Show Details' options should be visible only on mobile. If clicked, it should 'show' the `Action Type` and `Requirements` and update to say 'Hide Details'. If 'Hide Details' is clicked, they should be hidden again. 

The modal should look as it normally does on medium and large screens. 

Preview App: https://dosomething-phoenix-de-pr-1849.herokuapp.com/us/campaigns/test-teens-for-jeans?utm_source=scholarship&utm_campaign=fastweb

I went back and forth on a few different options for implementation here, so if there is a "cleaner" way you have encountered in the past let me know!

### Relevant tickets

References [Pivotal # 169929291](https://www.pivotaltracker.com/story/show/169929291).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipActionType.js
+++ b/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipActionType.js
@@ -1,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Header from './InfoHeader';
+
+const ScholarshipActionType = ({ actionLabel }) => (
+  <div>
+    <Header content="Action Type" />
+    <p className="pb-2">{actionLabel}</p>
+  </div>
+);
+
+export default ScholarshipActionType;
+
+ScholarshipActionType.propTypes = {
+  actionLabel: PropTypes.string.isRequired,
+};
--- a/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipActionType.js
+++ b/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipActionType.js
@@ -2,16 +2,24 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import Header from './InfoHeader';
+import PlaceholderText from '../../utilities/PlaceholderText/PlaceholderText';
 
-const ScholarshipActionType = ({ actionLabel }) => (
-  <div className="lg:w-1/2 lg:float-right">
-    <Header content="Action Type" />
-    <p className="pb-2">{actionLabel}</p>
-  </div>
+const ScholarshipActionType = ({ actionLabel, isLoaded }) => (
+  <>
+    {isLoaded && actionLabel ? (
+      <div className="lg:w-1/2 lg:float-right">
+        <Header content="Action Type" />
+        <p className="pb-2">{actionLabel}</p>
+      </div>
+    ) : (
+      <PlaceholderText size="medium" />
+    )}
+  </>
 );
 
 ScholarshipActionType.propTypes = {
   actionLabel: PropTypes.string.isRequired,
+  isLoaded: PropTypes.bool.isRequired,
 };
 
 export default ScholarshipActionType;
--- a/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipActionType.js
+++ b/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipActionType.js
@@ -4,14 +4,14 @@ import PropTypes from 'prop-types';
 import Header from './InfoHeader';
 
 const ScholarshipActionType = ({ actionLabel }) => (
-  <div>
+  <div className="lg:w-1/2 lg:float-right">
     <Header content="Action Type" />
     <p className="pb-2">{actionLabel}</p>
   </div>
 );
 
-export default ScholarshipActionType;
-
 ScholarshipActionType.propTypes = {
   actionLabel: PropTypes.string.isRequired,
 };
+
+export default ScholarshipActionType;
--- a/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
+++ b/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
@@ -86,6 +86,10 @@ const ScholarshipInfoBlock = ({
   const toggleHiddenInfo = () => {
     setDrawerOpen(!drawerOpen);
     setDetailsLabel('Less');
+
+    if (drawerOpen && detailsLabel === 'Less') {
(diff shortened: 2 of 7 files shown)


resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -166,25 +193,69 @@ const ScholarshipInfoBlock = ({
...
+                        className="lg:w-1/2 lg:float-right"
+                      >
+                        {isLoaded && actionType ? (
+                          <>
+                            <Header content="Action Type" />
+                            <p className="pb-2">{actionType}</p>
+                          </>
+                        ) : (
+                          <PlaceholderText size="medium" />
+                        )}
```
aaronschachter: What do you think about extracting this into a separate `ScholarshipInfoBlock/ScholarshipActionType` component, so we don't have to repeat this markup below?

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -166,25 +193,69 @@ const ScholarshipInfoBlock = ({
...
+                    </div>
+                  ) : (
+                    <div>
+                      <Header content="Requirements" />
+                      <ul className="mt-2 pb-2 list-disc list-inside">
+                        <li>Under 26 years old</li>
+                        <li>No minimum GPA</li>
+                        <li>No essay</li>
+                      </ul>
+                    </div>
```
aaronschachter: Similarly, what do you think about extracting this into a `ScholarshipInfoBlock/ScholarshipRequirements` component to DRY?

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -1,13 +1,18 @@
-import React from 'react';
+/** @jsx jsx */
+
+// eslint-disable-next-line no-unused-vars
```
aaronschachter: I'm not familiar with what's going on here -- is this so we can use Emotion?

lindsayrmaher: Yeah, it was throwing an eslint error if I kept React defined because of bringing `jsx` on the line before, but then the component couldn't render without React being defined.

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
lindsayrmaher: I like! I'll work on this for the `requirements` and the `action type`

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -166,25 +183,51 @@ const ScholarshipInfoBlock = ({
...
+                {matches => (
                   <>
-                    <Header content="Action Type" />
-                    <p className="pb-2">{actionType}</p>
+                    {matches.small ? (
+                      <div
+                        css={!drawerOpen ? isVisible : null}
+                        className="lg:w-1/2 lg:float-right"
+                      >
+                        {isLoaded && actionType ? (
```
aaronschachter: I was still hoping this logic wouldn't get duplicated, and was thinking that the `isLoaded`prop would be passed to the `ScholarshipActionType` component as well... but I could see an argument for not returning the `Placeholder` within it  Maybe a different way to clean this up could be:
```
const scholarshipActionType = isLoaded && actionType ? (
    <div>
      <Header content="Action Type" />
      <p className="pb-2">{actionLabel}</p>
    </div>
  )
  :  <PlaceholderText size="medium" />;
```
What do you think of that approach?

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipActionType.js
```diff
@@ -0,0 +1,17 @@
...
+import Header from './InfoHeader';
+
+const ScholarshipActionType = ({ actionLabel }) => (
+  <div>
+    <Header content="Action Type" />
+    <p className="pb-2">{actionLabel}</p>
+  </div>
+);
+
+export default ScholarshipActionType;
```
aaronschachter: Usually this shows up at the end of the file (although it make sense to define in `render` per the other comment)

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
lindsayrmaher: oh i think i understand more what you're thinking with this. I think having placeholder inside would make sense since we would want to account for the loading state. i'll push up more updates!

lindsayrmaher:
@aaronschachter any other thoughts on this pr?

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -65,6 +60,30 @@ const ScholarshipInfoBlock = ({
...
+    overflow: hidden;
+    clip: rect(1px, 1px, 1px, 1px);
+    white-space: nowrap; /* added line */
+  `;
+
+  const toggleHiddenInfo = () => {
+    setDrawerOpen(!drawerOpen);
+    setDetailsLabel('Hide');
+
+    if (drawerOpen && detailsLabel === 'Hide') {
```
aaronschachter: I forgot to ask this in the last PR. Won't `detailsLabel` always be set to `Hide` because of line 80?
Does this do the same thing?
```
  const toggleHiddenInfo = () => {
    setDrawerOpen(!drawerOpen);
    setDetailsLabel(drawerOpen ? 'Show' : `Hide`);
  }
```

lindsayrmaher: It does! I had tried to use a ternary before I added the if statement but I didn't try it inside the setDetailsLabel func. This feels like a cleaner way to do this. I'm going to add it with my next PR for this ticket.

lindsayrmaher:
@aaronschachter thanks so much for the feedback, i think this is a much cleaner PR after separating some of that code out 🙌

PR #1850, opened 2020-01-13
Gridifying site chrome

weerd:
### What's this PR do?

This pull request updates the site chrome to use CSS Grid.

Before this update, pages that had minimal content did not span the full vertical height and created an awkward visual with the footer:

image

With these updates, the central area always spans the full vertical height while still accounting for the navigation up top and the footer at the bottom:

image

👀 

Moving forward based on this new site chrome structure and with how the CSS grid works within the site chrome there is a specific structure to keep in mind when creating new page _types_ in the future:

```html
<body>
  <div class="chrome">
    <div id="app" class="app">

      <nav id="nav" class="nav"></nav>

      <main>

        <article class="">
          <!-- page content goes here -->
        </article>

        <!--
          popups or other page modals should go here, 
          after the <article> but contained within <main>
        -->
      </main>

      <footer class="site-footer"></footer>

    </div>
  </div>
</body>
```

### Relevant tickets

References [Pivotal #170761789](https://www.pivotaltracker.com/story/show/170761789).

### Checklist


resources/assets/components/utilities/Modal/modal.scss
```diff
@@ -21,7 +21,7 @@
     display: flex;
     height: 100%;
     justify-content: center;
-    overflow-y: scroll;
+    overflow-y: auto;
```
weerd: I noticed this happens for all our environments, but it adds scrolls bars to the modal portal div and wrapper even if the modal actually fits within the window, which seems unnecessary. So this just updates it to only add the vertical scroll bars when modal content is actually longer than the browser window.

I'm just noting here in case maybe there's something I'm missing and we do want the scroll bars always showing? I can't remember if this was a very deliberate style rule 🤔 I seems to otherwise work fine...

DFurnes: Hmm, I can't remember any reason we'd need this to be `scroll` off-hand. As long as it works with short & overflowing modal content, then I think this seems good to me!

weerd: Okie dokie! Thanks for the input. And yep it seems to work fine :)

resources/views/campaigns/index.blade.php
```diff
@@ -1,27 +1,33 @@
...
-    </main>
+                    @if($campaigns->count() >= $count)
+                        <a class="button -tertiary float-right clear-none" href="?page={{$nextPage}}">Next</a>
+                    @endif
+                </div>
+
+            </article>
+        </main>
+
+        @include('partials.footer')
```
weerd: @katiecrane something to note here. This is one place I found made it easier to keep the `footer.blade.php` file around, but I know that the `/campaigns` page is likely getting a React-ification soon, so once that happens we should be able to remove this `/campaigns/index.blade.php` file and the `footer.blade.php`. Just wanted to note that there's a reason it is still sticking around for the time being! 😄

katiecrane: Oh rad! That's awesome that we'll be able to clean up a whole bunch when we start that.

resources/views/search/results.blade.php
```diff
@@ -1,37 +1,43 @@
...
+
+                @if(! empty($campaigns))
+                    @include('partials.campaign_gallery', ['campaigns' => $campaigns])
+                @endif
+
+            </article>
+
+        </main>
+
+        @include('partials.footer')
```
weerd: Ah! Another spot we use the `footer.blade.php`! Just noting for future reference!

resources/assets/components/pages/GeneralPage/GeneralPage.js
```diff
@@ -48,10 +48,11 @@ const GeneralPage = props => {
   const ctaCopy = REGISTER_CTA_COPY[pageCategory];
 
   return (
-    <div>
+    <>
       <SiteNavigationContainer />
-      <div className="main general-page base-12-grid">
-        <Enclosure className="grid-narrow">
+
+      <main role="main" className="base-12-grid bg-white">
```
weerd: @mendelB I previously had the `relative` Tailwind class on this `<main>` element since the prior `.main` class had a rule for `position: relative`, but not sure if it was actually needed. The pop-up is `position: fixed` so relative to the `<body>` but maybe I'm forgetting some other element on GeneralPage that would require the positioning context. 

I was also debating just adding it to all `<main>` tags, but seemed like overkill 🤔 cc @DFurnes

mendelB: Hmm, no specific reason that I can recall personally 🤔

resources/assets/components/pages/ErrorPage.js
```diff
@@ -1,13 +1,14 @@
...
+import SiteFooter from '../utilities/SiteFooter/SiteFooter';
 import SiteNavigationContainer from '../SiteNavigation/SiteNavigationContainer';
 
 const ErrorPage = () => (
   <>
     <SiteNavigationContainer />
 
-    <main role="main" className="py-20">
-      <article className="max-w-xl mx-auto text-center px-4 leading-relaxed mb-8">
+    <main role="main" className="py-20 relative">
```
mendelB: What's the reason for the `relative` positioning here? (I've seen this in a few other places, so assuming there's something specific in certain pages to warrant this?)

resources/assets/scss/base.scss
```diff
@@ -59,13 +59,22 @@ html {
...
 body {
   @apply bg-gray-300 text-gray-900;
 }
 
+.app {
+  display: grid;
+  grid-template-columns: 100%;
+  grid-template-rows: auto 1fr auto;
+  max-width: 1440px;
+}
```
mendelB: Ooh this is interesting! Are we displaying the whole `.app` as one columned grid?

resources/assets/components/AdminDashboard/AdminDashboard.js
```diff
@@ -2,7 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const AdminDashboard = ({ enabled, children }) =>
-  enabled ? <div className="bg-gray-200 p-4">{children}</div> : null;
+  enabled ? (
+    <div className="legacy-admin-dashboard bg-gray-200 p-4">{children}</div>
```
DFurnes: Is this `legacy-admin-dashboard` class being used? If it's just for targeting (like for automated tests and whatnot), I wonder if a data attribute might be clearer to use a data-attribute?

resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -150,7 +150,7 @@ class SiteNavigation extends React.Component {
 
   render() {
     return (
-      <nav role="navigation" id="nav" className="nav">
+      <nav role="navigation" id="nav" className="nav site-navigation">
```
DFurnes: Same thought as above, I wonder if this is clearer as a `data-` attribute?

resources/assets/components/pages/AccountPage/AccountNavigation.js
```diff
@@ -3,7 +3,7 @@ import { NavLink } from 'react-router-dom';
 import PropTypes from 'prop-types';
 
 const AccountNavigation = props => (
-  <div className="page-navigation -no-fade">
+  <nav className="page-navigation -no-fade">
```
DFurnes: Wooo, semantic tags! ♿️

resources/assets/components/pages/BlockPage/BlockPage.js
```diff
@@ -14,11 +14,13 @@ const BlockPage = ({ match }) => (
   <>
     <SiteNavigationContainer />
 
-    <div className="main clearfix">
-      <Enclosure className="md:w-3/4 mx-auto my-16 px-3">
+    <main role="main" className="clearfix">
```
DFurnes: I think we can omit the `role="main"` since this is a `main` tag, right?

resources/assets/components/pages/CausePage/CausePage.js
```diff
@@ -49,60 +50,71 @@ const CausePageTemplate = ({
...
-                  </p>
-                  <a
-                    className="text-white hover:text-white font-normal underline cursor-pointer"
-                    href={stat.link.url}
-                  >
-                    {stat.link.text}
-                  </a>
-                </div>
-              ))}
+      <main role="main">
```
DFurnes: I think we can remove `role="main"` here too.

resources/assets/components/utilities/ModalRoute/ModalRoute.js
```diff
@@ -53,12 +53,12 @@ class ModalRoute extends React.Component {
         />
         {isModal(location) ? (
           <StaticRouter location={withoutModal(location)} context={{}}>
-            <div>{children}</div>
+            {children}
           </StaticRouter>
         ) : (
-          <div>{children}</div>
+          <>{children}</>
```
DFurnes: Does this need a fragment around it? I believe this could just be:

```suggestion
          children
```

resources/assets/components/utilities/PostCard/post.scss
```diff
@@ -6,7 +6,7 @@
 }
 
 // HACK: Ensure text post cards look okay on "classic" campaign pages.
-.campaign-page .post-ornament-text {
+.campaign-page__content .post-ornament-text {
```
DFurnes: Did we end up removing these "ornaments" when refactoring these to use Tailwind?

resources/assets/components/utilities/SiteFooter/SiteFooter.js
```diff
@@ -0,0 +1,161 @@
...
+        </ul>
+      </div>
+      <div className="footer__column -links">
+        <h4>Who We Are</h4>
+        <ul>
+          <li>
+            <a href="https://join.dosomething.org/">What is DoSomething.org?</a>
+          </li>
+          <li>
+            <a href={`${PHOENIX_URL}/us/about/our-people`}>Our Team</a>
```
DFurnes: I know this isn't new, but why do we need `PHOENIX_URL` here?

Could we just do `href="/us/about/our-people"`

resources/views/partials/footer.blade.php
```diff
@@ -1,4 +1,4 @@
-<footer class="footer pb-32 md:pb-3">
+<footer role="contentinfo" class="footer pb-32 md:pb-3">
```
DFurnes: It looks like [`<footer>` implies `role="contentinfo"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Contentinfo_role):

> Using the `<footer>` element will automatically communicate a section has a role of `contentinfo`. Developers should always prefer using the correct semantic HTML element over using ARIA, making sure to test for known issues in VoiceOver.

resources/assets/components/AdminDashboard/AdminDashboard.js
weerd: I added it as a helpful class, to easily see the div in the markup when inspecting and not need to wonder what that div is. I just couldn't name it `admin-dashboard` since that exists for the new Admin Dashboard. If this isn't helpful I can also just remove it :)

resources/assets/components/SiteNavigation/SiteNavigation.js
weerd: Oh I think I meant to update and remove the `nav` class and switch to using `site-navigation` it'd be more specific for the styles and I liked that it matched the class name for the footer of `site-footer`.

resources/assets/scss/base.scss
weerd: Yep yep! And with this single column we have the row with the `<nav>` as `auto` (takes up as much room as it needs), the main content area row as `1fr` to span most of the height, and the row with the `<footer>` as `auto` as well; essentially, fits the full layout and accomplishes a "sticky footer" 😄

resources/assets/components/pages/ErrorPage.js
weerd: I went back and forth on this one. I think I meant to remove it here, but overall, any page that has a class of `main`, had a stylesheet declaration of `position: relative` and I was afraid to not have that after removing the `main` class.

weerd:
> One observation -- so now we're manually rendering the footer on each page, which makes sense to ensure things follow the correct layout and hierarchy within main etc. Though I wonder if we can eventually add some sort of wrapper component for convenience since we probably want the footer everywhere?

@DFurnes had the same thought, but I think it's clearer to have the `<SiteFooter>` component and the `<SiteNavigation>` showing up in the same page structure. That, and as soon as we try and create a wrapper component for convenience, we'll inevitably get a request that we want a rapid response, or custom art directed page with no footer. 😏

weerd:
> One note -- re the .campaign-page -> .campaign-page__content update. Once you rebase/merge master to resolve the conflicts. I think we'll need to also update the LandingPage as well since that class was added there recently in #1855

Ah! Good note! Will do 💃

resources/assets/components/utilities/SiteFooter/SiteFooter.js
weerd: 🤔 I don't see why not...

resources/assets/components/AdminDashboard/AdminDashboard.js
mendelB: I was wondering about the best practice for these 'targeting' or 'descriptive' classes. I think we have a bunch of 'em throughout our app

resources/assets/components/utilities/PostCard/post.scss
weerd: Yep, [I believe so](https://github.com/DoSomething/phoenix-next/pull/1825/files#diff-3f17090eebf2444f3d8679e75c7bfb5eR77). I think I just forgot to remove the styles. Good 👀!

resources/assets/components/AdminDashboard/AdminDashboard.js
weerd: @mendelB we're going to try our using `data-ref` as the way to target elements for testing, etc instead of using a class. Ideally this won't conflict with React's use of the `ref` attribute and since it's recommended to refrain from using it in React for the most part we think it should be ok. 💃

PR #1851, opened 2020-01-13
Zendesk Ticket API Endpoint

mendelB:
### What's this PR do?

This pull request adds a new API endpoint and controller to handle creating Zendesk tickets from Phoenix via the https://github.com/huddledigital/zendesk-laravel package.

(This replaces our initial approach of using the Anonymous API as explained [here](https://www.pivotaltracker.com/story/show/167352209/comments/210254929))

Make sure the basic setup & logic makes sense. There's a bunch of refactoring to do which I'll list below.

A bunch of this is inspired by the setup of our preexisting Phoenix Ashes functionality which did more or less precisely what we're trying to do here.

[Pheonix Ashes Zendesk module](https://github.com/DoSomething/phoenix/blob/11cb8b73af0c1edbe7cb697ff82c835dbb594436/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module)
[Phoenix Ashes documentation](https://github.com/DoSomething/phoenix/wiki/DoSomething-Zendesk)

We're notably no longer setting Group ID's which on Ashes were based on the campaign's cause. We're satisfied with the Campaign name in the ticket subject which uses [Zendesk triggers](https://support.zendesk.com/hc/en-us/articles/203662246-About-triggers-and-how-they-work) to assign the correct campaign lead/agent on Zendesk.

I refrained from assigning the campaign lead explicitly using the Contentful campaign's assigned campaign lead email for now for the reasons listed [here in the Pivotal ticket](https://www.pivotaltracker.com/story/show/167352209/comments/210365437)

TODO:
- I think this ticket creation logic needs to be placed inside a Job, where we can then queue ticket creation events from the API so the whole app isn't blocked for these relatively expensive processes.
- We might want some logging here? I guess we can log when we're creating/finding a user and when we're creating the ticket
- Retries? The client by default will retry in [certain scenarios](https://www.pivotaltracker.com/story/show/167352209/comments/210254929), and then maybe we can just rely on the queuing. [@aaronschachter flagged this here](https://dosomething.slack.com/archives/C2BPA7M8F/p1563212550008900)
- updating the `ZendeskForm` to make requests to this endpoint.

### Relevant tickets

References [Pivotal #167352209](https://www.pivotaltracker.com/story/show/167352209).

### Checklist


app/Http/Controllers/Api/ZendeskTicketsController.php
```diff
@@ -0,0 +1,72 @@
...
+        ]);
+
+        $campaignName = $request->campaign_name;
+        $question = $request->question;
+
+        $user = gateway('northstar')->withToken(token())->get('v1/profile');
+
+        $northstarId = data_get($user, 'data.id');
+        $userEmail = data_get($user, 'data.email');
+        $userName = data_get($user, 'data.first_name').' '.data_get($user, 'data.last_name');
```
DFurnes: If we can, let's use `display_name`, e.g. "Mendel B.", here so we don't pull PII that we don't need.

mendelB: Ah wise! Updated in https://github.com/DoSomething/phoenix-next/pull/1851/commits/625140df9468343c465172584ce023cf0178cde1

mendelB: Oops, (I _think_ since we're using the legacy API endpoint,) the `display_name` attribute isn't available within the user data, so I just updated to assign it manually in https://github.com/DoSomething/phoenix-next/pull/1851/commits/43b7944e02e9fdd3face6acf827e7cb7e4643620

PR #1853, opened 2020-01-14
Adds Analytics to Dismissable Element

Vmack:
### What's this PR do?

This pull request builds on [PR #1840 ](https://github.com/DoSomething/phoenix-next/pull/1840/files) adding analytics to the `DismissableElement` so that we can track the number of users who close the newsletter cta. 

👀

N/A

### Relevant tickets

References [Pivotal #169330883](https://www.pivotaltracker.com/story/show/169330883).

### Checklist

### DIFF
--- a/resources/assets/components/pages/GeneralPage/GeneralPage.js
+++ b/resources/assets/components/pages/GeneralPage/GeneralPage.js
@@ -152,6 +152,7 @@ const GeneralPage = props => {
         {slug === 'about/easy-scholarships' ? (
           <DismissableElement
             name="cta_popover_scholarship_email"
+            context={{ ContextSource: 'newsletter_scholarships' }}
             render={(handleClose, handleComplete) => (
               <DelayedElement delay={3}>
                 <CtaPopover
--- a/resources/assets/components/pages/GeneralPage/GeneralPage.js
+++ b/resources/assets/components/pages/GeneralPage/GeneralPage.js
@@ -152,7 +152,7 @@ const GeneralPage = props => {
         {slug === 'about/easy-scholarships' ? (
           <DismissableElement
             name="cta_popover_scholarship_email"
-            context={{ ContextSource: 'newsletter_scholarships' }}
+            context={{ contextSource: 'newsletter_scholarships' }}
             render={(handleClose, handleComplete) => (
               <DelayedElement delay={3}>
                 <CtaPopover
--- a/resources/assets/components/utilities/DismissableElement/DismissableElement.js
+++ b/resources/assets/components/utilities/DismissableElement/DismissableElement.js
@@ -3,8 +3,9 @@ import { useState, useEffect } from 'react';
 
 import { isTimestampValid, query } from '../../../helpers';
 import { get as getStorage, set as setStorage } from '../../../helpers/storage';
+import { trackAnalyticsEvent } from '../../../helpers/analytics';
 
-const DismissableElement = ({ name, render }) => {
+const DismissableElement = ({ name, render, context }) => {
   const [showElement, setShowElement] = useState(true);
 
   const handleCompletion = () => {
@@ -16,6 +17,16 @@ const DismissableElement = ({ name, render }) => {
     if (!getStorage(`hide_${name}`, 'boolean')) {
       // Mark the element as "dismissed" in local storage & hide it.
       setStorage(`dismissed_${name}`, 'timestamp', Date.now());
+
+      trackAnalyticsEvent({
+        context,
+        metadata: {
+          category: 'site_action',
+          noun: name,
+          target: 'dismissable_element',
+          verb: 'dismissed',
+        },
+      });
     }
     setShowElement(false);
   };
@@ -42,8 +53,13 @@ const DismissableElement = ({ name, render }) => {
 };
 
 DismissableElement.propTypes = {
+  context: PropTypes.object,
   name: PropTypes.string.isRequired,
   render: PropTypes.func.isRequired,
+};
+
+DismissableElement.defaultProps = {
+  context: {},
 };
 
 export default DismissableElement;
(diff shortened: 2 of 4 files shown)


resources/assets/components/utilities/DismissableElement/DismissableElement.js
```diff
@@ -16,6 +17,16 @@ const DismissableElement = ({ name, render }) => {
...
+
+      trackAnalyticsEvent({
+        metadata: {
+          label: 'call_to_action_popover',
+          category: 'site_action',
+          noun: 'call_to_action_popover',
+          target: 'form',
+          verb: 'abandoned',
+        },
+      });
```
mendelB: Since this event would trigger for _any_ `DismissableElement`, I wonder if this can be a general `dismissed` event which can trigger for whichever element is being dismissed, using the name as the adjective like so:
`phoenix_dismissed_[name]` (e.g. `phoenix_dismissed_call_to_action_popover`)
```js
metadata: {
  // no label, it'll default to the noun.
  category: 'site_action',
  noun: name,
  target: 'dismissable_element',
  verb: 'dismissed',
}
```

Alternatively, the `noun` can be `'dismissable_element'` and the `adjective` can be the `name`. (e.g. `phoenix_dismissed_dismissable_element_call_to_action_popover`...)

Then, in order to get the `cta_type` for the context. We could allow passing through a `context` or  `analyticsContext` or something prop, which could be an object containing `cta_type: 'newsletter_scholarships'`, which we could set as the `context`.

Vmack: Hey @weerd @mendelB and I were discussing best approaches to this: wondering if we should add analytics to the `call_to_action_popover` instead and just add a prop and a new context(which can also be done in the `DismissableElement` ) but if we decide to keep it in the `DismissableElement` should we use the `name` as the noun/adjective or dismissable_element as the noun/adjective?

resources/assets/components/utilities/DismissableElement/DismissableElement.js
```diff
@@ -16,6 +17,16 @@ const DismissableElement = ({ name, render }) => {
     if (!getStorage(`hide_${name}`, 'boolean')) {
       // Mark the element as "dismissed" in local storage & hide it.
       setStorage(`dismissed_${name}`, 'timestamp', Date.now());
+
+      trackAnalyticsEvent({
+        context: { ContextSource: 'newsletter_scholarships' },
```
katiecrane: This `newsletter_scholarships` is gonna get added to all the `DismissableElement` events since it's hardcoded in here. I think what we want and what @mendelB was suggesting [here](https://github.com/DoSomething/phoenix-next/pull/1853#discussion_r367472189) was adding a `context` prop to this component (so then it would have `name`, `render`, and `context`) and then filling in the `context` here with whatever `context` we received via the prop. You'll also need to [pass the new prop over here](https://github.com/DoSomething/phoenix-next/blob/23a36944edaefb57b068b01a43bf54edbf002ec4/resources/assets/components/pages/GeneralPage/GeneralPage.js#L151-L153) where we create the popover. Does that make sense?

weerd: What @katiecrane and @mendelB are suggesting seems like the way to go :)

resources/assets/components/pages/GeneralPage/GeneralPage.js
```diff
@@ -152,6 +152,7 @@ const GeneralPage = props => {
         {slug === 'about/easy-scholarships' ? (
           <DismissableElement
             name="cta_popover_scholarship_email"
+            context={{ ContextSource: 'newsletter_scholarships' }}
```
lindsayrmaher: Small thing, but `contextSource` should be camelCase!

Vmack: Oh great catch!

resources/assets/components/utilities/DismissableElement/DismissableElement.js
```diff
@@ -16,6 +17,16 @@ const DismissableElement = ({ name, render }) => {
     if (!getStorage(`hide_${name}`, 'boolean')) {
       // Mark the element as "dismissed" in local storage & hide it.
       setStorage(`dismissed_${name}`, 'timestamp', Date.now());
+
+      trackAnalyticsEvent({
+        context,
```
katiecrane: What does this payload look like if we don't pass through any `context`? Are we passing an extra empty object here? If so, are we okay with that? (cc: @mendelB on my if so question - I would think we only want to include `context` here at all if it's not empty? but my thinking there is general and not analytics specific so there could be something I'm missing)

mendelB: We should be fine passing through an empty object since that's the default value either way!
We strips any 'valueless' properties away, and always end up adding some additional UTM context data from within the analytics helper, so we should be fine in any case ☺️ 

https://github.com/DoSomething/phoenix-next/blob/d9e180baf0a3b1a1f218e764784b2b7e2f7e18ea/resources/assets/helpers/analytics.js#L253-L256

katiecrane: Right! I wasn't sure if we needed to conditionally include it based on if we actually passed `context` with some content in it or not. A-ha! Stripping away "valueless" properties - this is the kind of analytics specific thing I wasn't sure about! Thanks Mendel!!

mendelB: I do feel bad calling them valueless though 😢 I'm sure they have their own special value regardless of our cruel 'truthiness' assessments

mendelB:
@Vmack the Jest test is failing for this component because we're now triggering an analytics event which uses some ENV variables we don't have set in the testing environment.

I think we can just mock the analytics in `DismissableElement.test.js` kinda like how we do it in this test:
https://github.com/DoSomething/phoenix-next/blob/4288a6529a3bb2c5c375b0eaea53b8b753f6c802/resources/assets/components/actions/ShareAction/ShareAction.test.js#L10

Vmack:
@mendelB ahhhh ok! Thank you! I was wondering why the tests kept failing so that makes sense. I'll try that.

PR #1854, opened 2020-01-14
Refactor Quiz block to work anywhere.

DFurnes:
### What's this PR do?

This pull request refactors the Quiz to work as a standard block, so it doesn't rely on being on a special `/quiz/:slug` route and swapping out URLs to function.

I've also updated `ScrollConcierge` to include a "trigger" prop that allows it to be re-triggered without any sort of futzing with timers, and a `useGate` hook that allows us to "flash" state client-side when we leave for the authentication flow.

~The last step is figuring out a clean way to "queue" the signup after the user is on a result & authenticated, like we do currently with the `requiresAuthentication` redux middleware...~

I was able to do this by running a mutation once `QuizResult` is shown... not sure if that's the best we can do or not. In particular, thinking about how we'd do this for the pitch page...

🍡

### Relevant tickets

References [Pivotal #170053835](https://www.pivotaltracker.com/story/show/170053835).

### Checklist


DFurnes:
Ugh, running into more trouble than I expected on writing a Cypress test here – it seems like the "mock" GraphQL server doesn't support inline fragments on interfaces? Continuing to investigate, but this is otherwise ready for review.

DFurnes:
@mendelB I'd love your eyes on this as well, since you're the quiz master! 🙏

DFurnes:
Editorially, this introduces the following changes (cc: @lkpttn):
 - All quiz results _must_ be blocks (plain-text results are no longer supported).
 - As blocks are now loaded on-demand by GraphQL, there is no need to link result blocks or follow-up quiz steps in the `resultBlocks` reference field.
- Quizzes must have a `additionalContent.campaignId` attached. We should add a `campaignId` field to this content type with the [extension](https://github.com/DoSomething/contentful-campaign-extension) to make this a nicer editor UX.

DFurnes:
This is blocked on DoSomething/graphql#187 before we can test on a review app or deploy.

resources/assets/components/Quiz/QuizBlock.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  ${QuizBlockFragment}
+`;
+
+const QuizBlock = props => {
+  // If we're returning from an authentication redirect, go to "intended"
+  // block. Otherwise, start at the beginning of this quiz:
+  const [flash, authenticate] = useGate(`QuizBlock:${props.id}`);
```
DFurnes: Note: Having the parameter for `useGate` being the "unique key" for this gate, rather than the "initial state" for the flashed data is a little confusing (especially side-by-side with a `useState` where the parameter _is_ initial state).

Is there a better way we can keep track of which gate we're storing/reading flash data for?

resources/assets/components/Quiz/QuizBlock.js
```diff
@@ -0,0 +1,66 @@
...
+  if (showingResult && !isAuthenticated()) {
+    authenticate({ id });
+
+    return <Spinner className="mx-auto my-6" />;
+  }
+
+  return (
+    <div className="mx-3">
+      {hasNavigated ? <ScrollConcierge trigger={id} /> : null}
+      {data.block.__typename === 'QuizBlock' ? (
```
DFurnes: @mendelB said "let's use `showingResult` here!"

DFurnes:
Merging this in since we're trying to get out a quiz broadcast this morning! I'll follow-up on the Cypress tests and some of the improvements we discussed in follow-up PRs.

PR #1856, opened 2020-01-16
Scholarship Signup Button Tracking

lindsayrmaher:
### What's this PR do?

This pull request adds a `modalType` flag to the signup button `context` section of `scholarship_modal` if a user signs up via the modal vs the campaign page. 

Does the general logic of the implementation make sense? Also do you see any strong reasons for or against using the `modalType` naming vs something different? 

This is actually very similar to how we are doing tracking for all modals. see #1838 
Check in on the conversation in the pivotal ticket for the reasoning behind naming etc. @mendelB and I were thinking it would make more sense from the dev perspective to choose a name related to `source`, but I do also want to take consideration from the data team on this since they will be pulling this info in the end. 

### Relevant tickets

References [Pivotal # 170053904](https://www.pivotaltracker.com/story/show/170053904).

### Checklist

### DIFF
--- a/resources/assets/components/LedeBanner/templates/HeroTemplate.js
+++ b/resources/assets/components/LedeBanner/templates/HeroTemplate.js
@@ -106,6 +106,7 @@ const HeroTemplate = ({
                         ? SCHOLARSHIP_SIGNUP_BUTTON_TEXT
                         : undefined
                     }
+                    contextSource="campaign_landing_page"
                   />
                   {affiliateOptInContent ? (
                     <AffiliateOptInToggleContainer
@@ -141,7 +142,7 @@ const HeroTemplate = ({
                 <SignupButtonContainer
                   className="w-full md:px-2"
                   text={SCHOLARSHIP_SIGNUP_BUTTON_TEXT}
-                  trackingId="scholarship_modal"
+                  contextSource="scholarship_modal"
                 />
               </div>
             </ScholarshipInfoBlock>
--- a/resources/assets/components/LedeBanner/templates/HeroTemplate.js
+++ b/resources/assets/components/LedeBanner/templates/HeroTemplate.js
@@ -141,7 +141,7 @@ const HeroTemplate = ({
                 <SignupButtonContainer
                   className="w-full md:px-2"
                   text={SCHOLARSHIP_SIGNUP_BUTTON_TEXT}
-                  modal_type="scholarship_modal"
+                  trackingId="scholarship_modal"
                 />
               </div>
             </ScholarshipInfoBlock>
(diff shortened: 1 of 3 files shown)


resources/assets/components/LedeBanner/templates/HeroTemplate.js
```diff
@@ -141,6 +141,7 @@ const HeroTemplate = ({
                 <SignupButtonContainer
                   className="w-full md:px-2"
                   text={SCHOLARSHIP_SIGNUP_BUTTON_TEXT}
+                  trackingId="scholarship_modal"
```
katiecrane: If this gets filled in as the `modalType`, why do we call it `trackingId`? I see that on modals we use `trackingId` to define what sort of modal they are, is the idea to use the same language here?

mendelB: This does tie to why we felt uncomfortable about using `modalType` here to begin with. Since, to my understanding, what we're tracking here isn't _which modal_ we're using, but _where_ the signup came from. So in this case, the signup is either a Scholarship Info Block in a modal, or the Hero Template Lede Banner. But in the abstract, it could be from a Call to action in a community page, where `modalType` would seem like an ill-fitting descriptor.

If we feel like this oddity is balanced out by the Data tech effort to pull a _new_ variable into Looker I get it, but just wanted to advocate this point here in case we think it's convincing enough for a switcharoo.

lindsayrmaher: yeah 2 reasons for that i guess. One that I wanted to name it something that made sense as a dev as far as what it actually is, a way to ID that particular tracking event. And then two because I figured maybe it would at least make sense to follow a similar pattern as we do in other places. 

in general it feels a bit hacky so i'm not sold in keep it this way, but based on feedback from the data team I just went for it lol

katiecrane: My confusion is if this is _always_ pulled in as `modalType` in the analytics, I'm wondering why it is ever called `trackingId` instead of `modalType`, so was just wondering the reasoning behind that.

lindsayrmaher: Fair question! I don't know why we named the prop `trackingId` in the modal tracking originally, but I could see the argument for the fact that we could use it as the "trackingId" for more than one piece of analytics in the future. 🤷‍♀

Do we think it makes sense to move forward with `modalType` in this particular case?

katiecrane: I think I prefer `modalType`? Do you have strong opinions on one making more sense?

lindsayrmaher:
@katiecrane updated per the epic thread convo!

lindsayrmaher:
i ❤️ threads!!!

PR #1862, opened 2020-01-21
Closed Campaign Scholarship Display Updates

lindsayrmaher:
### What's this PR do?

This pull request adds a check to the scholarship amount display to only display is the campaign is still open. 

👀 

Want to have a clearer user experience for users looking at closed campaigns and what is available in that campaign.

### Relevant tickets

References [Pivotal #170001678](https://www.pivotaltracker.com/story/show/170001678).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
+++ b/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 
 import Query from '../../Query';
 import Card from '../../utilities/Card/Card';
-import { getHumanFriendlyDate, isCampaignClosed } from '../../../helpers';
+import { getHumanFriendlyDate } from '../../../helpers';
 
 import './campaign-info-block.scss';
 
--- a/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
+++ b/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
@@ -27,11 +27,7 @@ const CAMPAIGN_INFO_QUERY = gql`
   }
 `;
 
-const CampaignInfoBlock = ({
-  campaignId,
-  scholarshipAmount,
-  scholarshipDeadline,
-}) => (
+const CampaignInfoBlock = ({ campaignId, scholarshipAmount }) => (
   <Card className="bordered p-3 rounded campaign-info">
     <h1 className="mb-3 text-lg uppercase">Campaign Info</h1>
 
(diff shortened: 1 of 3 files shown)


mendelB:
@lindsayrmaher Do you think it might make sense to _not_ display the scholarship info block at all if the campaign is closed? And just run this `isCampaignClosed` check right from the Hero Template along with:
https://github.com/DoSomething/phoenix-next/blob/e01788be7f4ae90ff29db2a91d0485c81c44fb5e/resources/assets/components/LedeBanner/templates/HeroTemplate.js#L130

lindsayrmaher:
I think we'll need to do that in addition to this change, but Diego D made a separate ticket for it! I think for this PR we do want to see the general campaign info card, but remove the line with the amount they could win from the scholarship.

mendelB:
> I think we'll need to do that in addition to this change, but Diego D made a separate ticket for it! I think for this PR we do want to see the general campaign info card, but remove the line with the amount they could win from the scholarship.

🤦‍♂ I totally missed that this was the Campaign Info Block and not the scholarship info block! 😵

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
```diff
@@ -64,7 +64,7 @@ const CampaignInfoBlock = ({ campaignId, scholarshipAmount }) => (
                   <dd>{actionItem.actionLabel}</dd>
                 </React.Fragment>
               ) : null}
-              {scholarshipAmount ? (
+              {scholarshipAmount && !isCampaignClosed(endDate) ? (
```
mendelB: Would it at all make sense to use the `scholarshipDeadline` here? 
https://github.com/DoSomething/phoenix-next/blob/33e4f03827134565e7bf3c010782ae4451a4164c/resources/assets/components/LedeBanner/templates/HeroTemplate.js#L37

That way if the scholarship ends before the campaign we still wouldn't display this. (And maybe we even check both, since the scholarship deadline isn't a required field 🤔 )

lindsayrmaher: I think checking for both might be a good coverall in case the dates don't align!

lindsayrmaher: Coming back to adding the scholarship deadline check, I talked to Luke this morning and there doesn't seem to be a case where the scholarship deadline would be before the campaign closing. So i'm not going to add an edge case check for that right now. But definitely a good thing to keep in mind if that workflow changes!

mendelB: Ah, interesting! Hmm then I wonder if we should even just remove that field entirely from Contentful and use the campaign end date... but maybe this ties into @lkpttn's work on Scholarship updates in general!

PR #1867, opened 2020-01-22
Contentful migration for Campaign blurb field

aaronschachter:
### What's this PR do?

This pull request adds a migration that updates all `campaign` entries, setting the `blurb` field to its Landing Page's `content` field if it doesn't use the legacy template. 

It also modifies the `campaign` content type migration to temporarily allow `page` entries to be saved to the campaign's `landingPage` reference field.  This change will need to be manually made in each environment, and then changed back once the migration as ended. See background context.

👀 

When this script attempts to update campaigns that have a `page` set for `landingPage`, we're unable to save our changes because of validation errors -- we can't save the `page` entry to the `landingPage` field. It's possible we could manually create `LandingPage` entries for the 10 campaign entries on production to avoid this, but we'd also need to copy/paste all the `content` fields for each. Let's create those entries manually once we no longer need their `content` field values anymore, because this migration will have run.

### Relevant tickets

References [Pivotal #170735335](https://www.pivotaltracker.com/story/show/170735335).

### Checklist

### DIFF
--- a/contentful/content-types/campaign.js
+++ b/contentful/content-types/campaign.js
@@ -325,7 +325,7 @@ module.exports = function(migration) {
     .required(false)
     .validations([
       {
-        linkContentType: ['landingPage', 'sixpackExperiment'],
+        linkContentType: ['landingPage', 'page', 'sixpackExperiment'],
       },
     ])
     .disabled(false)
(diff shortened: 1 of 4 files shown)


contentful/content-types/campaign.js
```diff
@@ -325,7 +325,7 @@ module.exports = function(migration) {
     .required(false)
     .validations([
       {
-        linkContentType: ['landingPage', 'sixpackExperiment'],
+        linkContentType: ['landingPage', 'page', 'sixpackExperiment'],
```
aaronschachter: This is the only change I meant to add when generating this migration (from `dev` space). It seems all other changes are around `builtin` support, but I'm assuming they should be ok to commit?

DFurnes: Weird! Maybe Contentful just updated this API at some point.

mendelB: Yeah, this happened to me as well -- it's totally an updated API format!

contentful/management-api-scripts/2020_01_17_import_campaign_blurb.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  // Determine if the campaign entry was already published to prevent publishing a draft entry.
+  const wasCampaignEntryPublished = await campaignEntry.isPublished();
+  // Initalize an object to write save to the Campaign blurb field.
+  const newBlurbValue = {};
+
+  // Update the Campaign blurb with its Landing Page content.
+  newBlurbValue[LOCALE] = landingPageContent;
+  campaignEntry.fields.blurb = newBlurbValue;
```
mendelB: We actually have a lil helper for this! (not that you need this here it's a small update but just for ref!)
https://github.com/DoSomething/phoenix-next/blob/a09c764ecb1b1d8fe8b53d9e6802700efaeaa16a/contentful/management-api-scripts/helpers.js#L112-#L121

PR #1869, opened 2020-01-22
Display Campaign blurb as LedeBanner content

aaronschachter:
### What's this PR do?

This pull request modifies the LedeBanner container to set the `campaign.blurb` for the `content` component prop, and adds checks to the Campaign Signup tests to verify the blurb content appears on the campaign page for both not-signed-up and signed-up views.

I've modified the blurb on Campaign entry for /us/campaigns/missing-history to test that the blurb field is displayed for both not-signed-up and signed-up views.

I was hoping that this change would allow us to delete the Landing Page `content` field -- but it's still used in the Mosaic Template. Next steps for that are [being discussed in Slack](https://dosomething.slack.com/archives/C3ASB4204/p1579731855016400).

### Relevant tickets

References [Pivotal #170735335](https://www.pivotaltracker.com/n/projects/2401401/stories/170735335) - this should be the last PR for that story.

### Checklist

### DIFF
--- a/cypress/integration/campaign-signup.js
+++ b/cypress/integration/campaign-signup.js
@@ -6,6 +6,8 @@ import { emptyResponse, newSignup } from '../fixtures/signups';
 import exampleCampaign from '../fixtures/contentful/exampleCampaign';
 
 const API = `/api/v2/campaigns/${campaignId}`;
+// Text included in the campaign blurb.
+const exampleBlurb = `Did you know that the world's oldest cat`;
 
 describe('Campaign Signup', () => {
   // Configure a new "mock" server before each test:
@@ -19,6 +21,7 @@ describe('Campaign Signup', () => {
 
     cy.contains('Example Campaign');
     cy.contains('This is an example campaign for automated testing.');
+    cy.contains(exampleBlurb);
 
     // Mock the responses we'll be expecting once we hit "Join Now":
     cy.route(`${API}/signups?filter[northstar_id]=${user.id}`, emptyResponse);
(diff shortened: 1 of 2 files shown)


cypress/integration/campaign-signup.js
```diff
@@ -6,6 +6,8 @@ import { emptyResponse, newSignup } from '../fixtures/signups';
 import exampleCampaign from '../fixtures/contentful/exampleCampaign';
 
 const API = `/api/v2/campaigns/${campaignId}`;
+// Text included in the campaign blurb.
+const exampleBlurb = `Did you know that the world's oldest cat`;
```
mendelB: Can we not check against the field itself because the markdown formatting in the fixture would prevent a direct match?

resources/assets/components/LedeBanner/LedeBannerContainer.js
```diff
@@ -25,10 +25,11 @@ const mapStateToProps = (state, props) => ({
   ),
   affiliateOptInContent: state.campaign.affiliateOptInContent,
   affiliateSponsors: state.campaign.affiliateSponsors,
+  // @TODO: We won't need this blurb property once MosaicTemplate is retired.
```
mendelB: 🙏

resources/assets/components/LedeBanner/LedeBannerContainer.js
```diff
@@ -25,10 +25,11 @@ const mapStateToProps = (state, props) => ({
...
   blurb: get(props, 'blurb', state.campaign.blurb),
   campaignId: state.campaign.campaignId,
   coverImage: get(props, 'coverImage', state.campaign.coverImage),
-  content: get(state, 'campaign.landingPage.fields.content'),
+  content: get(state, 'campaign.blurb'),
```
mendelB: I imagine we don't need `get` here since it's a direct campaign fields like the fields below, but I suppose a little dash of Lodash never hurt anyone...

cypress/integration/campaign-signup.js
aaronschachter: Yes, the tests were failing, expecting to find the `**` characters.

PR #1870, opened 2020-01-22
Contentful Script to Create Company Pages

mendelB:
### What's this PR do?

This pull request adds a new Contentful management API script to derive new Company Page entries from any About Page entry (any Page entry with a `slug` field prefixed with `about/`).

👁 

It's mostly par for the course. The trick is to convert the Long Text markdown `content` field and the Entry Reference `blocks` field into a singular Rich Text `content` field format.

For the above I followed along with the epic [Rich Text migration documentation](https://www.contentful.com/developers/docs/tutorials/general/migrate-to-rich-text/) from Contentful, utilizing the [@contentful/rich-text-from-markdown](https://www.npmjs.com/package/@contentful/rich-text-from-markdown) package to convert the markdown to Rich Text format, and employing their specified formatting for parsing the block references.

(We may be able to extract this functionality into a helper for future Markdown -> Rich Text migrations with the caveat being that embedded markdown image links pose a unique challenge and require a bunch more work (see [their step #9 here](https://www.contentful.com/developers/docs/tutorials/general/migrate-to-rich-text/#9-convert-the-markdown-text-with-images)). I was able to run a quick script to verify that none of the about pages contain embedded images (any `...` content))

### Relevant tickets

References [Pivotal #167297536.

### Checklist

### DIFF
--- a/contentful/management-api-scripts/2020_01_22_create_company_pages_from_about_pages.js
+++ b/contentful/management-api-scripts/2020_01_22_create_company_pages_from_about_pages.js
@@ -1,0 +1,87 @@
+const { flatten } = require('lodash');
+const { contentManagementClient } = require('./contentManagementClient');
+const { richTextFromMarkdown } = require('@contentful/rich-text-from-markdown');
+const {
+  processEntries,
+  getField,
+  withFields,
+  createLogger,
+  attempt,
+} = require('./helpers');
+
+const logger = createLogger('create_company_pages_from_about_pages');
+
+const createCompanyPageFromAboutPage = async (environment, pageEntry) => {
+  const pageEntryId = pageEntry.sys.id;
+  const pageInternalTitle = getField(pageEntry, 'internalTitle');
+  const pageSlug = getField(pageEntry, 'slug', '');
+
+  // Skip over non 'about' pages.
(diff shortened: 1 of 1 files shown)


contentful/management-api-scripts/2020_01_22_create_company_pages_from_about_pages.js
```diff
@@ -0,0 +1,87 @@
...
+      },
+    };
+  });
+
+  // Format Rich Text value from converted markdown and embedded blocks.
+  const richTextContent = {
+    nodeType: 'document',
+    content: flatten([
+      richTextFormattedContent.content,
+      richTextFormattedBlocks,
```
katiecrane: Does order matter here?

mendelB: Yes indeed! In the case of the Pages we're deriving from, we had the `content` Markdown field display _on top_ of the linked reference blocks on the page, so this would preserve that ordering on the Company Pages as well.

katiecrane: Oh perfect! That makes it easy!

PR #1871, opened 2020-01-23
Company Pages Feature Flag Route

mendelB:
### What's this PR do?

This pull request adds a feature flagged route for the new Company Pages (`/about/:slug`), so we can render them via GraphQL from our React app.

I've added a new `config/features` file inspired by [Rogue](https://github.com/DoSomething/rogue/blob/287908265922464776f6050d6080b524e23027b8/config/features.php) & [Northstar](https://github.com/DoSomething/northstar/blob/5ae12b19efe855449e99ce82a51fc180f66b74dc/config/features.php) so we can set these feature flags with ENV variables.

Working off of [the RFC](https://github.com/DoSomething/rfcs/pull/8) for more dynamic ENV variables, this adds a new `featureFlag` helper in https://github.com/DoSomething/phoenix-next/commit/5f80ab1ff11281a9b1d2666a85996901229ef188 so that we can more conveniently check feature flags in JS using an umbrella `FEATURE_FLAGS` variable https://github.com/DoSomething/phoenix-next/commit/bd1b2bcc6f9e9dd15a595a85f6eaac64a13bb0db.

Finally, it converts our existing feature flags to use this new format as well.

How's the naming of these variables?
How do we feel about this approach and this new helper?

Documentation and Cypress testing in a follow-up PR!

### Relevant tickets

### Checklist


mendelB:
> My only question is why we have the routing set up in both PHP and React.
Fair question!
The PHP side will affect the actual app entry routing, process things through the Laravel backedn/middleware etc. And render the front end app. From there, we have our top-level React app routing to render the proper component.

I think it's a bit more confusing because we're juggling things on the front & back end simultaneously, but as we get everything GraphQL-ified, I imagine that we'll be able to remove most of that backend routing!

> So if in the future we want to add more feature flags, we'll need to update the cypress index as well as the features config?

At least anytime we want things configured specifically for our test suite (and overriding whatever might be set by the env running the test) we'd need to explicitly set it here. (more context here https://github.com/DoSomething/phoenix-next/pull/1550)

PR #1872, opened 2020-01-23
View details Carat toggle

lindsayrmaher:
### What's this PR do?

This pull request adds a carat to the view details dropdown menu and starts work to animate it's rotation when clicked.

Lil Gif Action - sorry it's so gigantic 😂 
Kapture 2020-01-24 at 14 11 47

This is a start after getting a new SVG from Luke and talking with Diego, I'm feeling close to solving when to add the rotation but need an extra pair of eyes. I'm thinking I'll need to set off the actual transform from the `ScholarshipInfoBlock` rather than in the menu carat itself, and perhaps doing that by passing it as a class to the `MenuCarat` component? Please let me know thoughts!

### Relevant tickets

References [Pivotal #169929291](https://www.pivotaltracker.com/story/show/169929291).

### Checklist


lindsayrmaher:
@weerd when you have a chance to 👀

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipActionType.js
```diff
@@ -1,14 +1,13 @@
...
 import PropTypes from 'prop-types';
 
-import Header from './InfoHeader';
 import PlaceholderText from '../../utilities/PlaceholderText/PlaceholderText';
 
 const ScholarshipActionType = ({ actionLabel, isLoaded }) => (
   <>
     {isLoaded && actionLabel ? (
       <div className="lg:w-1/2 lg:float-right">
-        <Header content="Action Type" />
```
mendelB: What was the impetus for removing this `<Header/>`, did it just feel like overkill?

resources/assets/components/artifacts/MenuCarat/MenuCarat.js
```diff
@@ -0,0 +1,38 @@
+/** @jsx jsx */
+
+import PropTypes from 'prop-types';
+import { jsx, css } from '@emotion/core';
+
+const MenuCarat = ({ color, width, height }) => (
```
weerd: Nit-pick, switch these in alphabetical order :)

resources/assets/components/artifacts/MenuCarat/MenuCarat.js
```diff
@@ -0,0 +1,38 @@
...
+  <div className="menu-carat">
+    <svg
+      css={css`
+        transition: all 0.2s ease-in-out;
+        height: ${height};
+        width: ${width};
```
weerd: May want to add `pointerEvents: 'none',` to keep from events firing on this when clicked... similar to how the `CloseButton` is setup, we don't want the SVG firing events, but the actual button that the close icon sits inside of.

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -166,10 +162,9 @@ const ScholarshipInfoBlock = ({
         <div className="bg-white mx-2 my-6 md:mx-6 md:my-10 p-6 pb-2 rounded">
           {scholarshipAmount ? (
             <div>
-              <Header
-                content={scholarshipCallToAction}
-                textColor="text-purple-600"
-              />
+              <div className="font-bold uppercase text-purple-600">
+                {scholarshipCallToAction}
+              </div>
```
weerd: Would it make sense to make these some level of heading tag? `<h3>` or `<h4>` or something that sounds reasonable?

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipActionType.js
lindsayrmaher: Yeah, now that i've separated out some of the other fields being displayed it didn't feel as necessary to also split out the Header

resources/assets/components/artifacts/MenuCarat/MenuCarat.js
```diff
@@ -0,0 +1,39 @@
...
+
+const MenuCarat = ({ color, width, height, ...props }) => (
+  <div className="menu-carat">
+    <svg
+      css={css`
+        pointerevents: 'none';
+        height: ${height};
+        width: ${width};
+      `}
+      {...props}
```
mendelB: I'm not completely familiar with emotion yet, but would the `css` prop we pass in from the `ScholarshipInfoBlock` overwrite the `css` prop we're defining here above? or can you assign multiple `css` props at once?

lindsayrmaher: That's a good q, based on [their limited docs referencing this](https://emotion.sh/docs/with-props), any css passed in as props will take precedence over what's in the the component. But in this case the css we're passing in doesn't override any pre existing properties so i think it should still work as expected.

mendelB: Ah, makes sense, thanks for clarifying! Maybe worth defining a `css` prop explicitly for this component, or do we anticipate various other sorts of props to need this to be dynamic?

lindsayrmaher: I think it might get confusing since we're already importing `css` in the upper scope of the component from emotion itself, so I feel ok with leaving it dynamic. Maybe a note in the component would be helpful so that future users can see why we're using the props there in the first place

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -219,9 +217,14 @@ const ScholarshipInfoBlock = ({
...
-            <button type="button" onClick={toggleHiddenInfo}>
-              <p className="text-sm font-bold">{`${detailsLabel} Details`}</p>
+          <div className="sm:hidden text-center align-bottom flex justify-center">
+            <button
+              type="button"
+              className="flex items-center focus:outline-none"
+              onClick={toggleHiddenInfo}
+            >
+              <p className="text-sm font-bold pr-2">{`${detailsLabel} Details`}</p>
+              <MenuCarat css={drawerOpen ? caratRotate : null} />
```
weerd: We probs don't need the variable if only used one place... likely can just pass the `css` directly here 🤔

resources/assets/components/artifacts/MenuCarat/MenuCarat.js
weerd: I do prefer the idea of defining a specific prop... maybe `styles` or `cssStyles` so it doesn't confuse w/ the imported `css`? Just feels more straightforward and less likely to accidentally break if other props are passed to `<MenuCarat />` that aren't CSS but then try to get added because they're spread into it.

lindsayrmaher:
Looking into @weerd & @mendelB suggestions for the specific prop naming & directly passing the css rather than defining `caratRotate`. 

As far as giving it a custom name, I'm not sure emotion would know what do with it since it's looking for that specific prop (at least that's my understanding of it). Are you thinking more that we would append the style inside of the existing css block instead?

It might be helpful to talk through this IRL! The emotion docs are limited on these particular use cases.

mendelB:
> Looking into @weerd & @mendelB suggestions for the specific prop naming & directly passing the css rather than defining `caratRotate`.
> 
> As far as giving it a custom name, I'm not sure emotion would know what do with it since it's looking for that specific prop (at least that's my understanding of it). Are you thinking more that we would append the style inside of the existing css block instead?
> 
> It might be helpful to talk through this IRL! The emotion docs are limited on these particular use cases.

What if we embedded the prop inside the template literal so it just gets concatenated to the one big `css` string? So e.g. if we called it `cssStyles`
```js
css={css`
  pointerevents: 'none';
  height: ${height};
  width: ${width};
  ${cssStyles}
`}
```

lindsayrmaher:
ahhhhh yes Mendel you articulated that so much better than i could lolz, that works beautifully 😅

mendelB:
> ahhhhh yes Mendel you articulated that so much better than i could lolz, that works beautifully 😅

ohhh lol no you literally said what I said I just didn't realize 🤦‍♂! Awesome that it worked 🎉

PR #1873, opened 2020-01-23
Disable LandingPage content field, add docs

aaronschachter:
### What's this PR do?

This pull request updates the Landing Page content type migration to disable the `content` long text field, as it's no longer used in the hero template. 

This PR also starts on documentation for Campaign and Landing Page content types.

This branch is available to read on [Gitbook](https://dosomething.gitbook.io/phoenix-documentation/v/docs%252Flanding-page/development/content-types/campaign).

An upcoming PR will hardcode the `content` field values we need to keep for the mosaic template, so we can eventually delete the `content` field from the Landing Page and recreate it as a Rich Text field.

### Relevant tickets

References [Pivotal #170735335](https://www.pivotaltracker.com/n/projects/2401401/stories/170735335).

### Checklist


docs/development/content-types/landing-page.md
```diff
@@ -0,0 +1,11 @@
+# Landing Page
+
+## Overview
+
+The landing page displayed on a [campaign](development/content-types/campaign.md) when the current user is not signed up.
+
+## Content Type Fields
+
+- **Internal Title**: This is for our internal Contentful organization and will be how the entry shows up in search results, etc.
```
DFurnes: It might be worth specifically mentioning that these should not be displayed to end users. 🚷

DFurnes:
It'd be link to the [GraphQL documentation](http://graphql.dosomething.org) for these items (and could save us from documenting in two places), but I can't find a way to link directly to the documentation viewer... 🤔

aaronschachter:
I've thought about linking out to the GraphQL docs too but also couldn't figure out a way. I forgot to add documentation for legacy template Landing Page fields like `sidebar` but will add in an upcoming PR once the `content` field gets recreated as Rich Text.

docs/development/content-types/landing-page.md
aaronschachter: I copied this from other docs... I wonder about just removing this field completely - I'd bet its leftovers from when the Gitbook was intended for the campaigns team to use as well. I do think it's useful to document what fields are used where when it's not obvious (e.g. a Campaign blurb)

PR #1865, opened 2020-01-21
Contentful Migrations updates for Company Page & Gallery Block

mendelB:
### What's this PR do?

This pull request updates the Contentful migrations for Company Page & Gallery Block content types based on changes in https://github.com/DoSomething/phoenix-next/pull/1559

### How should this be reviewed?
👁 

### Any background context you want to provide?
I opted to remove the `/category` prefix as we did with [Collection pages](https://github.com/DoSomething/phoenix-next/blob/81e70c03e3128fd1c2cce5e0f2431ed41312f078/contentful/content-types/collection-page.js#L35-L43) since this will always be `/about/:slug`.

### Relevant tickets

References [Pivotal #167297536](https://www.pivotaltracker.com/story/show/167297536).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #1877, opened 2020-01-24
Fix Company Page migration

mendelB:
### What's this PR do?

I'm a bit of a [Schlemiel](https://en.wikipedia.org/wiki/Schlemiel), and added the embedded entry block validation in the wrong way.

This fixes that, plus, updates the ordering of the Cover Image field to follow the way it's displayed on the page.

Also, while testing the #1870 script on QA, I bumped into validation errors because we weren't allowing numerical values in slugs, but some about pages have 'em! I think we do mean to support alphanumeric values (as per the validation message), so updated in https://github.com/DoSomething/phoenix-next/pull/1877/commits/c3c7a0cbddfc3dbe01f2f16b02dc714362a6d431 

While I was at it and re @aaronschachter's pro tip https://github.com/DoSomething/phoenix-next/pull/1865#pullrequestreview-346773991 I added some documentation for the company page. (@aaronschachter how do you link to the Gitbooks version from your PR's?)

👀 

### Relevant tickets

### Checklist

### DIFF
--- a/contentful/content-types/companyPage.js
+++ b/contentful/content-types/companyPage.js
@@ -33,7 +33,7 @@ module.exports = function(migration) {
       },
       {
         regexp: {
-          pattern: '^[a-z\\-]+$',
+          pattern: '^[a-z[0-9]\\-]+$',
         },
 
         message:
--- a/contentful/content-types/companyPage.js
+++ b/contentful/content-types/companyPage.js
@@ -59,25 +59,6 @@ module.exports = function(migration) {
     .linkType('Entry');
 
   companyPage
-    .createField('title')
-    .name('Title')
-    .type('Symbol')
-    .localized(true)
-    .required(true)
-    .validations([])
-    .disabled(false)
-    .omitted(false);
-  companyPage
-    .createField('subTitle')
-    .name('Subtitle')
-    .type('Symbol')
-    .localized(true)
-    .required(false)
-    .validations([])
-    .disabled(false)
-    .omitted(false);
-
-  companyPage
     .createField('coverImage')
     .name('Cover Image')
     .type('Link')
@@ -98,6 +79,25 @@ module.exports = function(migration) {
     .linkType('Asset');
 
   companyPage
+    .createField('title')
+    .name('Title')
+    .type('Symbol')
+    .localized(true)
+    .required(true)
+    .validations([])
+    .disabled(false)
+    .omitted(false);
+  companyPage
+    .createField('subTitle')
+    .name('Subtitle')
+    .type('Symbol')
+    .localized(true)
+    .required(false)
+    .validations([])
+    .disabled(false)
+    .omitted(false);
+
+  companyPage
     .createField('content')
     .name('Content')
     .type('RichText')
@@ -106,7 +106,7 @@ module.exports = function(migration) {
     .validations([
       {
         nodes: {
-          'embedded-entry-inline': [
+          'embedded-entry-block': [
             {
               linkContentType: [
                 'contentBlock',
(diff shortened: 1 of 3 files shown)


contentful/content-types/companyPage.js
```diff
@@ -106,7 +106,7 @@ module.exports = function(migration) {
     .validations([
       {
         nodes: {
-          'embedded-entry-inline': [
+          'embedded-entry-block': [
```
katiecrane: What's the difference between `inline` and `block` here?

docs/development/content-types/company-page.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Technical Notes
+
+#### GraphQL
+
+The Company Page is available as a `CompanyPage` in GraphQL.
+
+#### Feature Flag
+
+Currently, `/about/:slug` pages pull from regular Page entries in Contentful, and _not_ from Company Pages. To toggle on the new Company Pages, `DS_ENABLE_COMPANY_PAGES=true` must be set in the `.env` file.
```
katiecrane: Is this feature flag to facilitate launching this or are we always gonna need it? (just a side question here - these docs look great!)

contentful/content-types/companyPage.js
mendelB: Honestly, I find it a bit confusing. (Well clearly from this PR lol).

It seems to me that an `inline` entry shows up embedded within the line of text, almost like a link (see [here](https://www.contentful.com/developers/docs/concepts/rich-text/#embedded-and-linked-entries))
image

And `block` entries are more traditional embedded blocks within the content.

But there doesn't seem to be much practical difference:
https://www.contentful.com/faq/rich-text/#what-is-the-difference-between-embed-block-entry-and-embed-inline-entry

Though we've only set up to process `block` embeds:
https://github.com/DoSomething/phoenix-next/blob/cea5c84ca916b3cfda436550758691dd8c114f2e/resources/assets/helpers/text.js#L165-L172

docs/development/content-types/company-page.md
mendelB: Just to facilitate the launch. I figured we'd be able to safely test on QA without blocking deploys, and pull the lever when we're comfortable!

katiecrane: love it!

contentful/content-types/companyPage.js
katiecrane: Wow that is super confusing! Even their docs are unclear. Thanks for explaining!

mendelB: Lol def! Reading through that link another 3 times. It does seem clearer that we are indeed after embedded `block` entries in our content since we're after the more robust 'widgets' - or 'blocks' in our terminology - embedded on the page, as opposed to annotation or 'microcopy' i.e. more copy format-ey type of inline embedding

katiecrane: Yeah that was the bit that was the most helpful!

contentful/content-types/companyPage.js
```diff
@@ -33,7 +33,7 @@ module.exports = function(migration) {
       },
       {
         regexp: {
-          pattern: '^[a-z\\-]+$',
+          pattern: '^[a-z[0-9]\\-]+$',
```
weerd: 🔢 ✔️

contentful/content-types/companyPage.js
weerd: I think we'll rarely end up using `inline` and most often opting for `block` for the most part. But yeah, it is confusing!

weerd:
Post-merge 👍

PR #1876, opened 2020-01-24
Rich Text Landing Page content

aaronschachter:
### What's this PR do?

Now that #1875 has been deployed, I've deleted the `content` long text field and recreated it as a Rich Text field. This PR adds the migration for that change, and displays the field in the Landing Page for non legacy template campaigns.

### How should this be reviewed?

I've created a test landing page with Rich Text content over on https://dosomething-phoenix-de-pr-1876.herokuapp.com/us/campaigns/aaron-test-campaign.

### Any background context you want to provide?

I've significantly edited this PR description now that all the things are working and it's no longer a WIP.

### Relevant tickets

References [Pivotal #170606520](https://www.pivotaltracker.com/n/projects/2401401/stories/170606520).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Documentation added for new features/changed endpoints.
- [x] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #1882, opened 2020-01-24
Remove unused title and subtitle fields from Landing Page

aaronschachter:
### What's this PR do?

This pull request modifies the migration for the `landingPage` content type to remove the `title` and `subtitle` fields, which are [unused](https://github.com/DoSomething/phoenix-next/pull/1876#discussion_r370809474).

Also adds a test for #1876.

👀 

Seems like we can safely delete the disabled `blocks` multi-value reference field from the `landingPage` content type as well? cc @mendelB 

### Relevant tickets

References [Pivotal #170606520](https://www.pivotaltracker.com/story/show/170606520).

### Checklist

### DIFF
--- a/app/Entities/LandingPage.php
+++ b/app/Entities/LandingPage.php
@@ -18,8 +18,6 @@ class LandingPage extends Entity implements JsonSerializable
             'type' => $this->getContentType(),
             'fields' => [
                 'internalTitle' => $this->internalTitle,
-                'title' => $this->title,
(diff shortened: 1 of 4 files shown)


mendelB:
> Seems like we can safely delete the disabled blocks multi-value reference field from the landingPage content type as well? cc @mendelB

Yes I don't think we ever ended up using these. RIP

PR #1883, opened 2020-01-24
🐛 🚨 patch the quiz loader

mendelB:
### What's this PR do?

This pull request fixes the quiz loader via `ContentfulEntryLoader`.

It seems we may have accidentally reverted the `id` prop due to a git history mismatch https://github.com/DoSomething/phoenix-next/blame/1e656cc6fdf6c51add92c7571cb9f25a6df6de2b/resources/assets/components/pages/CampaignPage/CampaignPage.js#L39

https://dosomething.slack.com/archives/C09ANFQLA/p1579903379013300

### DIFF
--- a/resources/assets/components/pages/CampaignPage/CampaignPage.js
+++ b/resources/assets/components/pages/CampaignPage/CampaignPage.js
@@ -36,7 +36,7 @@ const CampaignPage = props => {
             {!entryContent ? (
               <CampaignPageContent {...props} />
             ) : (
-              <ContentfulEntryLoader json={entryContent} />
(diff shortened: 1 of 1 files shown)


weerd:
Dang. I thought I caught all the changes. I had a substantial rebase I had to deal with and I remember that file being tricky 😫

mendelB:
> Dang. I thought I caught all the changes. I had a substantial rebase I had to deal with and I remember that file being tricky 😫

Totally, I missed this too!! I think it was just a really tough diff to catch subtleties because of the wider layout structure changing

PR #1884, opened 2020-01-27
Persisting UTM parameters across user journey.

weerd:
### What's this PR do?

When a user comes to the DoSomething site from a partner's site, they typically have UTMs attached to the URL as query parameters that help with our analytics to let us know where the user was coming from. However, these UTMs only exist on the first DoSomething site page the user lands on. We would like the UTMs to persist throughout the user's journey so our analytics reflects the full context of that user's journey.

This pull request allows UTMs attached to the URL from a partner's site to persist throughout the users session on the web, even after logging in or signing up for a campaign. The UTMs are cleared after the user's session ends, so when they close the browser or tab they are using.

Example:
Persisting UTM Parameters

👀 

### Any background context you want to 

We're using [`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) to handle storing the UTMs for the session.

### Relevant tickets

References [Pivotal #](https://www.pivotaltracker.com/story/show/169960864).

### Checklist

### DIFF
--- a/resources/assets/helpers/utm.js
+++ b/resources/assets/helpers/utm.js
@@ -52,8 +52,8 @@ export function persistUtms() {
   // Get any current query param utms:
   const utms = withoutValueless(getUtmParameters());
 
-  // If no current query param utms, check to see if there are any in session storage:
-  let sessionUtms = getSessionUtms(UTM_SESSION_KEY);
+  // Check to see if there are any in session storage:
+  const sessionUtms = getSessionUtms(UTM_SESSION_KEY);
 
   // If no current query param utms and no utms in session storage, return.
   if (isEmpty(utms) && isEmpty(sessionUtms)) {
@@ -62,7 +62,7 @@ export function persistUtms() {
 
   // If no session utms, then store the current query param utms:
   if (isEmpty(sessionUtms)) {
-    sessionUtms = setSessionUtms(UTM_SESSION_KEY, utms);
+    setSessionUtms(UTM_SESSION_KEY, utms);
 
     return;
   }
@@ -72,5 +72,7 @@ export function persistUtms() {
     const url = appendToQuery(sessionUtms);
 
     history.replaceState(null, '', url.href);
+
+    return;
   }
 }
--- a/resources/assets/helpers/utm.js
+++ b/resources/assets/helpers/utm.js
@@ -55,11 +55,6 @@ export function persistUtms() {
   // Check to see if there are any in session storage:
   const sessionUtms = getSessionUtms(UTM_SESSION_KEY);
 
-  // // If no current query param utms and no utms in session storage, return.
-  // if (isEmpty(utms) && isEmpty(sessionUtms)) {
-  //   return;
-  // }
-
   // If no session utms, then store the current query param utms:
   if (isEmpty(sessionUtms) && !isEmpty(utms)) {
     setSessionUtms(UTM_SESSION_KEY, utms);
--- a/resources/assets/helpers/utm.js
+++ b/resources/assets/helpers/utm.js
@@ -67,7 +67,5 @@ export function persistUtms() {
     const url = appendToQuery(sessionUtms);
 
     history.replaceState(null, '', url.href);
-
-    return;
   }
 }
--- a/resources/assets/helpers/utm.js
+++ b/resources/assets/helpers/utm.js
@@ -1,4 +1,34 @@
-import { query } from '.';
+/* global history */
+/* eslint-disable no-restricted-globals */
+
+import { isEmpty } from 'lodash';
+
+import { appendToQuery, query, withoutValueless } from '.';
+
+const UTM_SESSION_KEY = 'ds_utm_params';
+
+/**
+ * Get UTM parameters from session storage.
+ *
+ * @param  {String} key
+ * @return {Object}
+ */
+export function getSessionUtms(key) {
+  return JSON.parse(sessionStorage.getItem(key));
+}
+
+/**
+ * Store UTM parameters in session storage.
+ *
+ * @param  {String} key
+ * @param  {Object} utms
+ * @return {Object}
+ */
+export function setSessionUtms(key, utms) {
+  sessionStorage.setItem(key, JSON.stringify(utms));
+
+  return utms;
+}
 
 /**
  * Get UTM parameters.
@@ -13,4 +43,34 @@ export function getUtmParameters() {
   };
 }
 
-export default null;
+/**
+ * Persist UTM parameters in session storage.
+ *
+ * @return void
+ */
+export function persistUtms() {
+  // Get any current query param utms:
+  const utms = withoutValueless(getUtmParameters());
+
+  // If no current query param utms, check to see if there are any in session storage:
+  let sessionUtms = getSessionUtms(UTM_SESSION_KEY);
+
+  // If no current query param utms and no utms in session storage, return.
+  if (isEmpty(utms) && isEmpty(sessionUtms)) {
+    return;
+  }
+
+  // If no session utms, then store the current query param utms:
+  if (isEmpty(sessionUtms)) {
+    sessionUtms = setSessionUtms(UTM_SESSION_KEY, utms);
+
+    return;
+  }
+
+  // If no current query param utms, but found session utms, add them to the URL:
+  if (isEmpty(utms) && sessionUtms) {
+    const url = appendToQuery(sessionUtms);
+
+    history.replaceState(null, '', url.href);
+  }
+}
--- a/resources/assets/helpers/utm.js
+++ b/resources/assets/helpers/utm.js
@@ -55,20 +55,20 @@ export function persistUtms() {
   // Check to see if there are any in session storage:
   const sessionUtms = getSessionUtms(UTM_SESSION_KEY);
 
-  // If no current query param utms and no utms in session storage, return.
-  if (isEmpty(utms) && isEmpty(sessionUtms)) {
-    return;
-  }
+  // // If no current query param utms and no utms in session storage, return.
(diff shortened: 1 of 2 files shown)


resources/assets/helpers/utm.js
```diff
@@ -13,4 +43,34 @@ export function getUtmParameters() {
...
+
+    return;
+  }
+
+  // If no current query param utms, but found session utms, add them to the URL:
+  if (isEmpty(utms) && sessionUtms) {
+    const url = appendToQuery(sessionUtms);
+
+    history.replaceState(null, '', url.href);
+  }
```
weerd: I added comments for the different logical steps in the process; I thought it would help with understanding better, but if it's overkill, lemme know and I can remove :)

resources/assets/helpers/utm.js
```diff
@@ -13,4 +43,34 @@ export function getUtmParameters() {
...
+    sessionUtms = setSessionUtms(UTM_SESSION_KEY, utms);
+
+    return;
+  }
+
+  // If no current query param utms, but found session utms, add them to the URL:
+  if (isEmpty(utms) && sessionUtms) {
+    const url = appendToQuery(sessionUtms);
+
+    history.replaceState(null, '', url.href);
```
weerd: Not sure why ESLint was complaining about using `history` even after I added it as a `global` in the code comment above; ended up also needing to add the `eslint-disable` comment too. 🤷‍♂

resources/assets/helpers/utm.js
mendelB: I thought they were helpful!

resources/assets/helpers/utm.js
```diff
@@ -13,4 +43,34 @@ export function getUtmParameters() {
...
+  let sessionUtms = getSessionUtms(UTM_SESSION_KEY);
+
+  // If no current query param utms and no utms in session storage, return.
+  if (isEmpty(utms) && isEmpty(sessionUtms)) {
+    return;
+  }
+
+  // If no session utms, then store the current query param utms:
+  if (isEmpty(sessionUtms)) {
+    sessionUtms = setSessionUtms(UTM_SESSION_KEY, utms);
```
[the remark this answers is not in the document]
mendelB: Unrelated, is the scenario where we _do_ have UTMs in session storage, but the URL params have changed impossible because we wouldn't link internally to a URL with UTM params, to begin with, and thus, we'd never need to update the existing session UTMs with changed ones in the URL?

resources/assets/helpers/utm.js
```diff
@@ -13,4 +43,34 @@ export function getUtmParameters() {
...
+/**
+ * Persist UTM parameters in session storage.
+ *
+ * @return void
+ */
+export function persistUtms() {
+  // Get any current query param utms:
+  const utms = withoutValueless(getUtmParameters());
+
+  // If no current query param utms, check to see if there are any in session storage:
```
mendelB: > If no current query param utms

We'd be running this check in _either_ case, no?

resources/assets/helpers/utm.js
```diff
@@ -13,4 +43,34 @@ export function getUtmParameters() {
...
+  let sessionUtms = getSessionUtms(UTM_SESSION_KEY);
+
+  // If no current query param utms and no utms in session storage, return.
+  if (isEmpty(utms) && isEmpty(sessionUtms)) {
+    return;
+  }
```
mendelB: 🤔 I wonder, if we reverse the ordering of our conditionals here. Could we do away with this case? (Since we're nesting our logic within these conditionals anyways, I think it _might_ be helpful if we avoided this extra check for emptiness?)

resources/assets/helpers/utm.js
mendelB: I think I remember bumping into this before, and it wanted it to be explicitly called on the window `window.history...`

resources/assets/helpers/utm.js
weerd: Ah yep. Initial setup stuff; wording is confusing, good 👁

resources/assets/helpers/utm.js
weerd: I think I get what you mean and it was more of an effort to avoid nesting logic.

So, we only really want to fully exit out of the `persistUtms()` logic if there are both no UTMs in URL query params and none in the `sessionStorage`. If that's the case the user didn't come from a partner site and we don't need to worry about UTMs, and the method can exit out early.

With this conditional removed, it ends up setting a session variable with an empty object, and on every subsequent page load it will continue to get the data from the session storage, which just seemed like wasteful processing, ultimately grabbing an empty object that doesn't get used.

image

Alternatively, I could remove this conditional and instead do a check within the subsequent conditionals and exit early if the data is empty, but as a nested `if` conditional.

resources/assets/helpers/utm.js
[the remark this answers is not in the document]
weerd: Ah, true... I think that's some lingering code from when I was `console.log()`-ing.

>Unrelated, is the scenario where we do have UTMs in session storage, but the URL params have changed impossible because we wouldn't link internally to a URL with UTM params, to begin with, and thus, we'd never need to update the existing session UTMs with changed ones in the URL?

And yep, I think for the most part we are ideally assuming we're not using internal UTMs in links which is a big "no no".

resources/assets/helpers/utm.js
weerd: ah! ok! Will change it to that and see if it's happier 🥇

weerd: Nope, still not happy 😢

resources/assets/helpers/utm.js
mendelB: Ah yeah I think I was assuming your alternative suggestion, but I think we can do it without nesting?
```js
  // If no session utms, then store the current query param utms:
  if (isEmpty(sessionUtms) && !isEmpty(utms)) {
    sessionUtms = setSessionUtms(UTM_SESSION_KEY, utms);
  }

  // If no current query param utms, but found session utms, add them to the URL:
  if (isEmpty(utms) && !isEmpty(sessionUtms)) {
    const url = appendToQuery(sessionUtms);

    history.replaceState(null, '', url.href);
  }
```

Do you think this reads clearer or is just more confusing?

resources/assets/helpers/utm.js
```diff
@@ -13,4 +43,34 @@ export function getUtmParameters() {
...
+
+  // If no session utms, then store the current query param utms:
+  if (isEmpty(sessionUtms)) {
+    sessionUtms = setSessionUtms(UTM_SESSION_KEY, utms);
+
+    return;
+  }
+
+  // If no current query param utms, but found session utms, add them to the URL:
+  if (isEmpty(utms) && sessionUtms) {
```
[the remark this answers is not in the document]
weerd: Yep, more for clarity. It was confusing thinking through the different logical scenarios and the clarity helped! 😅

resources/assets/helpers/utm.js
weerd: 🤔 I think I usually opt for exiting out of functions sooner, in a simpler conditional. That being said, I think your suggestion still reads well. I'll give it a shot 👍

resources/assets/helpers/utm.js
```diff
@@ -13,4 +43,29 @@ export function getUtmParameters() {
...
+    setSessionUtms(UTM_SESSION_KEY, utms);
+
+    return;
+  }
+
+  // If no current query param utms, but found session utms, add them to the URL:
+  if (isEmpty(utms) && !isEmpty(sessionUtms)) {
+    const url = appendToQuery(sessionUtms);
+
+    history.replaceState(null, '', url.href);
```
DFurnes: Will this trigger two pageviews on GA or [this Snowplow listener](https://github.com/DoSomething/phoenix-next/blob/7d8dddc376705c30a7f68dc834701bc9d41aa466/resources/assets/helpers/analytics.js#L204-L228) (for the initial UTM-less load & then the `replaceState`)?

DFurnes: I'm curious if it might be enough to persist it with this helper, and then just load from the `sessionStorage` in `getUtmParameters` without doing any `history.pushState`.

mendelB: It didn't seem like it was based on an initial test (though that's anecdotal) it seemed that we only landed here _after_ the URL had been tampered with. (I don't know if that's necessarily how things are ordered, or if it had to do with my load time or something).

One thing I did notice - and maybe this is by design? - the `persistUtms` would not take place between single app pages since we're only calling it in `init`. Which I guess means, if I click a link to a campaign I'm signed up for, switch to say `/faqs`, and then hit up a different campaign & sign up, we'd still lose the UTM referral data?

DFurnes: > One thing I did notice - and maybe this is by design? - the `persistUtms` would not take place between single app pages since we're only calling it in `init`. Which I guess means, if I click a link to a campaign I'm signed up for, switch to say `/faqs`, and then hit up a different campaign & sign up, we'd still lose the UTM referral data?

Oh, good point – that's definitely not how things should work (since any "hard refresh" routing is something that we'll be able to gradually phase out over time).

weerd: Oh hmmmm 🤔 I hadn't considered the single app pages issue. I went with the `history.pushState` based on how the process was described in the PivotalTicket comments. But if we want to go without using `history.pushState` and it just kind of handles it behind the scenes, I can do an update.

weerd: @mendelB I tried the scenario you described (if I understood it correctly), and while the `/faq` page doesn't get the UTMs, if I go to another campaign (using the same browser tab session) the UTMs do get added to the URL.

PR #1885, opened 2020-01-27
Render Company Pages

mendelB:
### What's this PR do?

This pull request allows the new Company Pages to be rendered via GraphQL on all `/about/:slug` routes.

See #1145 for screenshots/styling of this new page.

Once the review app deploys, I'll flip on the `company_pages` feature flag so we can test that the about pages load. (I've already run the migration script (#1870) on `dev` so we can visit these pages for testing)

- https://github.com/DoSomething/phoenix-next/commit/50a9e5ee5024d64e2740cf24b4f9c11062220311 Updated the `schema.json` so we have this new `companyPageBySlug` query for testing.
- https://github.com/DoSomething/phoenix-next/commit/e71d8296851b3993bb818c66c9f8a56f7bf10934 added a Cypress command inspired from our [`withState` command](https://github.com/DoSomething/phoenix-next/blob/01e19d1fd1a886b663b94821943ed80054db3f30/cypress/support/commands.js#L108) to allow assigning feature flag ENV variables per test for the `window.ENV`. (I thought would be helpfully distinct from a command to assign regular env variables since these are nested under `FEATURE_FLAGS`)

### Relevant tickets

### Checklist

### DIFF
--- a/resources/assets/components/pages/CompanyPage/CompanyPage.js
+++ b/resources/assets/components/pages/CompanyPage/CompanyPage.js
@@ -1,13 +1,29 @@
 import React from 'react';
+import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
 
+import PageQuery from '../PageQuery';
 import LazyImage from '../../utilities/LazyImage';
-import { contentfulImageUrl } from '../../../helpers';
 import SiteFooter from '../../utilities/SiteFooter/SiteFooter';
 import TextContent from '../../utilities/TextContent/TextContent';
+import { contentfulImageUrl, withoutNulls } from '../../../helpers';
 import SiteNavigationContainer from '../../SiteNavigation/SiteNavigationContainer';
 
-const CompanyPage = props => {
+export const COMPANY_PAGE_QUERY = gql`
+  query CollectionPageQuery($slug: String!, $preview: Boolean!) {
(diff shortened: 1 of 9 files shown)


resources/assets/components/pages/CompanyPage/CompanyPage.js
```diff
@@ -50,10 +91,20 @@ CompanyPage.propTypes = {
...
-CompanyPage.defaultProps = {
+CompanyPageTemplate.defaultProps = {
   coverImage: {},
   content: null,
   subTitle: null,
 };
 
+const CompanyPage = ({ slug }) => (
+  <PageQuery query={COMPANY_PAGE_QUERY} variables={{ slug }}>
+    {page => <CompanyPageTemplate {...withoutNulls(page)} />}
```
lindsayrmaher: Are we just adding withoutNulls here to avoid anything breaking down the line with null props? Just wondering since most of the props in the Template aren't required.

mendelB: > Are we just adding withoutNulls here to avoid anything breaking down the line with null props

👌 exactly 

e.g. for the `coverImage` prop, if it's not set, it'll explicitely be passed through as `null`  overriding the `{}` default which would cause breakage when we query for the fields.

PR #1886, opened 2020-01-27
Clean up from details toggle ticket

lindsayrmaher:
### What's this PR do?

This pull request does some clean up based on the conversation from #1872 !

👀 

Wanted to follow through on the suggestions to more explicitly define what props we could pass to `MenuCarat` and get rid of unnecessary variables.

### Relevant tickets

References [Pivotal #169929291](https://www.pivotaltracker.com/story/show/169929291).

### Checklist

### DIFF
--- a/resources/assets/components/artifacts/MenuCarat/MenuCarat.js
+++ b/resources/assets/components/artifacts/MenuCarat/MenuCarat.js
@@ -28,7 +28,7 @@ const MenuCarat = ({ color, cssStyles, height, width }) => (
 
 MenuCarat.propTypes = {
   color: PropTypes.string,
-  cssStyles: PropTypes.string,
+  cssStyles: PropTypes.object,
   width: PropTypes.string,
   height: PropTypes.string,
 };
--- a/resources/assets/components/artifacts/MenuCarat/MenuCarat.js
+++ b/resources/assets/components/artifacts/MenuCarat/MenuCarat.js
@@ -3,17 +3,17 @@
 import PropTypes from 'prop-types';
 import { jsx, css } from '@emotion/core';
 
-const MenuCarat = ({ color, height, width, ...props }) => (
+const MenuCarat = ({ color, cssStyles, height, width }) => (
   <div className="menu-carat">
     <svg
       css={css`
         pointerevents: 'none';
         height: ${height};
         width: ${width};
+        ${cssStyles};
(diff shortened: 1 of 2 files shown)


resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -224,7 +220,9 @@ const ScholarshipInfoBlock = ({
               onClick={toggleHiddenInfo}
             >
               <p className="text-sm font-bold pr-2">{`${detailsLabel} Details`}</p>
-              <MenuCarat css={drawerOpen ? caratRotate : null} />
+              <MenuCarat
+                cssStyles={drawerOpen ? `transform: rotate(180deg);` : null}
```
DFurnes: Would this make sense to provide as a `` css`…` `` so we get syntax highlighting and such? That seems to be suggested in Emotion's [Composition docs](https://emotion.sh/docs/composition).

resources/assets/components/artifacts/MenuCarat/MenuCarat.js
```diff
@@ -28,12 +28,14 @@ const MenuCarat = ({ color, height, width, ...props }) => (
 
 MenuCarat.propTypes = {
   color: PropTypes.string,
+  cssStyles: PropTypes.string,
```
DFurnes: If we do send a `css` template string, there might be better (maybe `instanceOf`?) prop-type to use here!

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
lindsayrmaher: Hmmmmm like this! even more clear that it's emotion specific.

resources/assets/components/artifacts/MenuCarat/MenuCarat.js
lindsayrmaher: well if we do use the css template string it gets interpreted as an obj, so maybe better to update to an obj?

DFurnes: Ah yeah, if it's just a plain object then that'll do it! 👍

PR #1889, opened 2020-01-28
Revising approach to persist UTMs for analytics.

weerd:
### What's this PR do?

After some discussion with the crew regarding the approach for persisting UTMs in #1884, we decided it best to revise the approach a bit. While we continue persisting the UTMs in session storage as planned, instead of using `history.pushState` to re-write the URL and append the UTM query params for every page in a user's journey, we would alternatively just pull the values from session storage (or URL query if no items in session) when an analytics event is triggered.

This should avoid any edge-case scenarios like when a page doesn't complete a full reload (going from campaign `/action` to `/faq` for example) and allow the URL to behave as expected w/ the UTMs disappearing after the initial arrival to a page from a third-party partner.

👀 and maybe using ye ole `window.DS.Debug.enableLogs(['google', 'snowplow'])` to confirm context data includes UTMs when testing!

Apart from discussion in #1884 there's also some additional discussion in a [Slack thread](https://dosomething.slack.com/archives/CJ340DEEL/p1580235818005600).

### Relevant tickets

References [Pivotal #169960864](https://www.pivotaltracker.com/story/show/169960864).

### Checklist

### DIFF
--- a/resources/assets/helpers/utm.js
+++ b/resources/assets/helpers/utm.js
@@ -1,9 +1,6 @@
-/* global history */
-/* eslint-disable no-restricted-globals */
-
 import { isEmpty } from 'lodash';
 
-import { appendToQuery, query, withoutValueless } from '.';
+import { query, withoutValueless } from '.';
 
 const UTM_SESSION_KEY = 'ds_utm_params';
 
@@ -31,16 +28,20 @@ export function setSessionUtms(key, utms) {
 }
 
 /**
- * Get UTM parameters.
+ * Get UTMs from session storage or query parameters.
  *
  * @return {Object}
  */
 export function getUtmParameters() {
-  return {
-    utm_source: query('utm_source'),
-    utm_medium: query('utm_medium'),
-    utm_campaign: query('utm_campaign'),
-  };
+  const sessionUtms = getSessionUtms(UTM_SESSION_KEY);
+
+  return !isEmpty(sessionUtms)
+    ? sessionUtms
+    : withoutValueless({
+        utm_source: query('utm_source'),
+        utm_medium: query('utm_medium'),
+        utm_campaign: query('utm_campaign'),
+      });
 }
 
 /**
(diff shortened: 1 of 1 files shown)


resources/assets/helpers/utm.js
```diff
@@ -49,23 +50,19 @@ export function getUtmParameters() {
...
-  // Check to see if there are any in session storage:
+  // Check to see if there are any utms in session storage:
   const sessionUtms = getSessionUtms(UTM_SESSION_KEY);
 
-  // If no session utms, then store the current query param utms:
-  if (isEmpty(sessionUtms) && !isEmpty(utms)) {
-    setSessionUtms(UTM_SESSION_KEY, utms);
-
+  // If utms in session storage, no need to store them again, so exit out.
+  if (!isEmpty(sessionUtms)) {
```
DFurnes: This might be a data team question, but I'm curious if we'd want to override existing UTMs if a user follows another link with UTMs (like, say if I follow a FastWeb link to one campaign and exit out, but later that session follow another FastWeb link to a different campaign – should it be attributed to that second UTM'd link)?

resources/assets/helpers/utm.js
```diff
@@ -31,16 +28,20 @@ export function setSessionUtms(key, utms) {
...
  *
  * @return {Object}
  */
 export function getUtmParameters() {
-  return {
-    utm_source: query('utm_source'),
-    utm_medium: query('utm_medium'),
-    utm_campaign: query('utm_campaign'),
-  };
+  const sessionUtms = getSessionUtms(UTM_SESSION_KEY);
```
DFurnes: Would it make sense to reference `UTM_SESSION_KEY` inside `getSessionUtms` and `setSessionUtms`, rather than passing it as the `key` argument to the function? (Assumption: we're not passing different session keys in different contexts).

weerd: Oh since it's accessible from either of those without need to pass them in. Sure, I can update that.

resources/assets/helpers/utm.js
weerd: Ya. I think maybe this might be something to see over time, but wasn't sure and likely something to bring up w/ Data Team.

weerd: [Slack discussion thread](https://dosomething.slack.com/archives/CJ340DEEL/p1580247106016300)

mendelB: One reason in favor of indeed allowing overwriting the session if UTM's change would be for more parity with [Google's definition](https://support.google.com/analytics/answer/2731565?hl=en) of a session, where new referral data _would_ overwrite the existing session. That being said, since we're only persisting per local _session_ and not _local storage_, I think the odds of UTM's changing within a single tab session are highly unlikely?

PR #1894, opened 2020-01-29
Add support for querying CurrentSchoolBlock via GraphQL.

DFurnes:
### What's this PR do?

This pull request adds the ability to load `CurrentSchoolBlock` via GraphQL.

👀

This is a blocker for updating Phoenix to query all blocks via GraphQL, and is dependent on the changes made in DoSomething/graphql#189 (where I added `CurrentSchoolBlock` to our schema).

I made the `ContentfulEntryLoader` fragments more verbose in 9412638, because I'd run into some bugs with our in-memory GraphQL mock server (for Cypress) on the `blocks/:id` route otherwise... they should be functionally identical, but for some reason this works when the short-syntax doesn't.

🤷‍♂

### Relevant tickets

References [Pivotal #169216496](https://www.pivotaltracker.com/story/show/169216496).

### Checklist

### DIFF
--- a/resources/assets/components/ContentfulEntry/ContentfulEntry.js
+++ b/resources/assets/components/ContentfulEntry/ContentfulEntry.js
@@ -167,6 +167,9 @@ class ContentfulEntry extends React.Component {
       case 'currentSchoolBlock':
         return <CurrentSchoolBlockContainer {...withoutNulls(json.fields)} />;
 
+      case 'CurrentSchoolBlock':
+        return <CurrentSchoolBlockContainer {...withoutNulls(json)} />;
+
       case 'embed':
         return (
           <EmbedBlockContainer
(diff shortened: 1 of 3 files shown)


DFurnes:
> Oooh nice! Would this also affect the field collision issue?

Hmm, not sure but worth re-trying. That'd be nice if it does solve that for us!

mendelB:
> > Oooh nice! Would this also affect the field collision issue?
> 
> Hmm, not sure but worth re-trying. That'd be nice if it does solve that for us!

But alas, it does not 😞 

This might just be a fundamental misunderstanding of GraphQL, but it seems strange to me that these fields conflict if they're both nested under separate `... on` statements.

Could a `ContentBlock` ever also be a `CampaignUpdateBlock`?
Or is there some sort of field merging at a higher level that causes these fields to conflict?

[The spec](http://spec.graphql.org/June2018/#sec-Field-Selection-Merging) _seems_ to hold the answer, but my brain can't quite compute right now.

DFurnes:
Bummer!

I _think_ the reason this fails is that since you can also use fragments on interfaces, you could have a scenario where something is, say, a `Block` and a `Showcasable` and have a field with conflicting types on each interface (like if returning a `ContentBlock` that implements both).

mendelB:
> Bummer!
> 
> I _think_ the reason this fails is that since you can also use fragments on interfaces, you could have a scenario where something is, say, a `Block` and a `Showcasable` and have a field with conflicting types on each interface (like if returning a `ContentBlock` that implements both).

Ohhhh right, yeah that makes sense! Thanks for clarifying!

PR #1897, opened 2020-01-31
Analytics refactoring

weerd:
### What's this PR do?

This pull request refactors our approach to analytics in Phoenix.

We wanted to refactor how we track analytics events on DoSomething.org since the system now is a little convoluted because of the need to maintain support for legacy event naming.

The system currently handles dynamically generating the event name based on provided `verb`, `noun`, and `adjective` properties. However, after our recent reconfiguring of analytics data platforms and structure of events, we included `category`, `action`, and `label` which complicated how we set up our custom events.

We've decided it will be easier if we specify the whole event name when triggering an event, instead of dynamically generating it, and thus removing the need for `verb`, `noun`, and `adjective`. We now end up with just `category`, `action`, and `label`, along with the previous `context` object for any extra event data, which matches how our analytics platforms (Google Analytics and Snowplow) register our events.

👀 

In an attempt to minimize the number of changes per PR, I've renamed the old analytics event tracking helper function to `legacyTrackAnalyticsEvent()`. Now when `trackAnalyticsEvent()` is called, it checks to see if the first argument is an object and if it is, it means it contains event data in our old format, so I send it off to `legacyTrackAnalyticsEvent()`, otherwise if it's a string, it's using the new format and can continue on its way.

This will allow us to update the slew of event triggers across our system over smaller PRs instead of all in one giant, and intimidating fell swoop of a PR 💃 

### Relevant tickets

References [Pivotal #](https://www.pivotaltracker.com/story/show/170710129).

### Checklist

### DIFF
--- a/resources/assets/components/actions/ShareAction/ShareAction.js
+++ b/resources/assets/components/actions/ShareAction/ShareAction.js
@@ -39,8 +39,16 @@ export const ShareBlockFragment = gql`
   }
 `;
 
+const EVENT_CATEGORY = 'campaign_action';
+
 class ShareAction extends PostForm {
   state = { showModal: false };
+
+  contextData = {
+    blockId: this.props.id,
+    campaignId: this.props.campaignId,
+    pageId: this.props.pageId,
+  };
 
   componentDidMount() {
     // If this is a Facebook share action, make sure we load SDK.
@@ -79,40 +87,17 @@ class ShareAction extends PostForm {
     });
   };
 
-  /**
-   * Component helper method for tracking analytics events.
-   *
-   * @param  {String} service
-   * @param  {String} target
-   * @param  {String} verb
-   * @param  {Object} data
-   * @return {Void}
-   */
-  trackEvent = (service, target, verb, data = {}) => {
-    trackAnalyticsEvent({
-      context: {
-        ...data,
-        blockId: this.props.id,
-        campaignId: this.props.campaignId,
-        pageId: this.props.pageId,
-      },
-      metadata: {
-        adjective: service,
-        category: 'campaign_action',
-        label: service,
-        noun: 'share_action',
-        target,
-        verb,
-      },
-    });
-  };
-
   handleFacebookClick = url => {
     const { link, userId } = this.props;
 
     let trackingData = { url: link };
 
-    this.trackEvent('facebook', 'button', 'clicked', trackingData);
+    trackAnalyticsEvent('clicked_share_action_facebook', {
+      action: 'button_clicked',
+      category: EVENT_CATEGORY,
+      label: 'facebook',
+      context: { ...this.contextData, ...trackingData },
+    });
 
     showFacebookShareDialog(url)
       .then(() => {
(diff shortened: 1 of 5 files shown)


resources/assets/components/actions/ShareAction/ShareAction.js
```diff
@@ -39,9 +39,17 @@ export const ShareBlockFragment = gql`
...
+const EVENT_CATEGORY = 'campaign_action';
+
 class ShareAction extends PostForm {
   state = { showModal: false };
 
+  contextData = {
+    blockId: this.props.id,
+    campaignId: this.props.campaignId,
+    pageId: this.props.pageId,
+  };
```
DFurnes: If these props change, will this update as well? (Or would it be safer to have this be a function?)

weerd: Props change how? Might be confused what you mean 🤔

DFurnes: Like, if the component rendering `ShareAction` were to change these props (say, changing `campaignId` or w/e)... would this static property on the component instance update? I'm not sure, since I don't think we'd be calling the constructor again in that case...

weerd: 🤔 that's fair. Not sure if it'd update the values, but also feels like these specific props wouldn't really be changing unless it's a different page. But I do agree we should be safer about it and a function makes sense! I can update 💃

resources/assets/helpers/analytics.js
```diff
@@ -257,3 +266,52 @@ export function trackAnalyticsEvent({ metadata, context = {}, service }) {
...
+
+    return;
+  }
+
+  // @REMOVE: We will switch back to destructuring the variables in the function signature,
+  // but while we support the legacyTrackAnalyticsEvent(), we need to destruct after
+  // checking against whether name is a string or object or will error out.
+  const { action, category, label, context = {}, service } = metadata;
+
+  if (!name && !isString(name)) {
```
mendelB: Which scenario are we protecting against with this conditional? Would an empty array or some other non-falsey but invalid value pass this test?

resources/assets/components/actions/ShareAction/ShareAction.js
```diff
@@ -79,40 +87,17 @@ class ShareAction extends PostForm {
...
-        adjective: service,
-        category: 'campaign_action',
-        label: service,
-        noun: 'share_action',
-        target,
-        verb,
-      },
-    });
-  };
-
```
mendelB: I quite like, that while we're adding repetition by removing this helper, it's a whole lot easier to grok at first glance!

DFurnes: Agreed!!

resources/assets/helpers/analytics.js
weerd: I figured it would be good to make sure that a string is the only thing that is passed as the first item, so we always at least have an event name string.

resources/assets/components/actions/ShareAction/ShareAction.js
weerd: Yep yep, me too. Felt like a better balance! ⚖️

resources/assets/helpers/analytics.js
weerd: I may be misunderstanding what you mean though...

mendelB: Ah ok, yeah that makes sense! I guess my other point is that technically this would allow an array or some other param as well since the `isString` check will only run _if_ the `name` is a falsey value, but not if it's e.g. an array

weerd: Ah, I think I get what you mean... and indeed, with an array or number, it pops out of the conditional and continues on its way. I tried just checking agains `!isString(name)` and that actually worked a lot better. Did a bunch more testing and behave as expected now. Also, while digging in it highlighted that I was using the wrong Lodash method in the first check to see if data needs to be sent to legacy function; instead of `isObject()` needed `isPlainObject()`! 🚥

weerd:
One note, I tackled these triggers in Shares and ShareAction related events because they seemed a bit extra convoluted. So if this approached worked well for those, then the rest of the analytics triggers will hopefully be a breeze! ⛵️

resources/assets/helpers/analytics.js
mendelB: Iiiinteresting! I forgot about the ole 'everything is an object' in JS 💫 

I liked this SO answer on clarification https://stackoverflow.com/a/34185403/5657278 the ending is concise: 
> I believe the use case for isPlainObject is to answer the question "is this just data?" or "is this code?"

PR #1898, opened 2020-01-31
Refer a Friend V1 Refactor

mendelB:
### What's this PR do?

This pull request refactors the Refer a Friend feature to once again utilize a `campaign_id` query param to allow multiple campaigns to run this feature.

It uses Feature flags and env configuration to allow for more feasible and configurable campaign selection & defaulting 

How's the new setup? Does this logic make sense? Do the tests feel good?

We retired this feature in favor of hardcoding the campaign in #1847 

We were hoping to only re-introduce this feature once we formalized a Rogue campaign setting for refer-a-friend participation https://github.com/DoSomething/phoenix-next/pull/1847#discussion_r365935116

We struck a compromise - to allow adding the new Mental Health campaign to this feature - by adding this feature flag configurable functionality for now, with [a ticket](https://www.pivotaltracker.com/story/show/171011197) set up to refactor in the future.

(See the Slack discussion linked in the [Pivotal ticket](https://www.pivotaltracker.com/story/show/171011155), and [this comment](https://www.pivotaltracker.com/story/show/171011155/comments/210982758) for more background, which campaign ID's we're configuring, and which defaults we're setting.)

I also added a new test to ensure the CTA referral banner shows up in the signup affirmation for configured campaigns!

### Relevant tickets

References [Pivotal #171011155](https://www.pivotaltracker.com/story/show/171011155).

### Checklist


mendelB:
I've added the feature flag env variables to the review app (9001 & 9002. plus 9001 as the default), and hooked it up for authentication, so you should be able to test the flow from
https://dosomething-phoenix-de-pr-1898.herokuapp.com/us/campaigns/do-something-about-bugs

(You can also test that it defaults to `9001`, by appending a wrong campaign ID to the beta or alpha page e.g. https://dosomething-phoenix-de-pr-1898.herokuapp.com/us/join?user_id=5bbd00271f1c570099415d02&campaign_id=9005)

cypress/integration/beta-referral-page.js
```diff
@@ -2,15 +2,22 @@
 
 import { userFactory } from '../fixtures/user';
 import { userId } from '../fixtures/constants';
+import { campaignId } from '../fixtures/constants';
 
 describe('Beta Referral Page', () => {
   // Configure a new "mock" server before each test:
   beforeEach(() => cy.configureMocks());
+  const referralFeatureFlags = {
+    refferal_campaign_ids: [campaignId],
```
katiecrane: should be `referral` 😬

mendelB: ohhhhh snap lollllll imma update. Good eye! 🙌

mendelB: I guess the tests passed because I had it set in my local 🤔

mendelB: For a sec I thought I'd spelled it wrong all over the app and was like yikes

katiecrane: that was the longshot nightmare with `recommendation` and `reccomendation` scattered throughout 😂

katiecrane: would you mind unsetting it locally and making sure the test fails? and then fix the spelling here and make sure it passes? I want to make sure we can really trust this test!

mendelB: lol there should legit be some machine learning smart-ass spell check plugin or something on builds, how cool would that be!

katiecrane: I am having trouble imagining a world where that does not do more harm than good 😂

mendelB: Wait also, while that loads, I realize it for sure passed because the `default_referral_campaign_id` _was_ spelled correctly so it always used that one. But double-checking!

And hmm, maybe worth using a different ID, to ensure it explicitly loads when the default is different?

cypress/integration/campaign-signup.js
```diff
@@ -55,6 +55,31 @@ describe('Campaign Signup', () => {
     cy.get('.card.affirmation').contains('Thanks for joining us!');
   });
 
+  it('Display Referral Page Banner CTA in affirmation for configured campaign & feature flagged user', () => {
```
katiecrane: Do you think it would be worth having the opposite of this test? Like making sure it doesn't show up when we don't want it to?

cypress/integration/beta-referral-page.js
katiecrane: oooooo interesting - I think that sounds like a good call

mendelB: Hmm tricky, it's hard to test against the embed link because that's a graphql query which is stubbed anyway. This will be easier to explicitly test on an Alpha page integration test where we can explicitly test against the `campaign_id` within the link we generate.

For now, I made sure the test would fail if the default wasn't set when the spelling was off. And
- fixed the spelling https://github.com/DoSomething/phoenix-next/pull/1898/commits/5b033fe3461491e5d31168a0c7a23216e432f7d2
- and ensure we test correctly against the proper URL! https://github.com/DoSomething/phoenix-next/pull/1898/commits/6b53da9dd756995bfd5f9a5343582e00ccaefaeb (also double-checked that the tests fail in the absence of a default for this one.)
- finally, I updated to ensure we only set the feature flags necessary per test so we have aren't implicitly passing the tests in https://github.com/DoSomething/phoenix-next/pull/1898/commits/6a4f43515ef957e0538cc35f3507d307b64da05e

cypress/integration/campaign-signup.js
mendelB: This is a good call, I'll add a case!

cypress/integration/beta-referral-page.js
katiecrane: love that last bullet point!

cypress/integration/campaign-signup.js
mendelB: Added in https://github.com/DoSomething/phoenix-next/pull/1898/commits/fcce7eae0010e6f95eb872fedc98ef9599ff9836

PR #1901, opened 2020-01-31
Quick Phoenix patches

mendelB:
### What's this PR do?

This pull request patches two small things I spotted in the latest deploy:
- https://github.com/DoSomething/phoenix-next/commit/7a4c812b20754aab8a14ffbbc87ea91f7f91061c a typo in a `context` prop 
- https://github.com/DoSomething/phoenix-next/commit/7e7cb7b29e0a729a2148c057ae0d58c19e84e2cb spreading `trackingData`
- oop! And one more thing. https://github.com/DoSomething/phoenix-next/pull/1901/commits/5c36774a8055aa8c24af32b92942a3772463eae7 reverts the homepage nps survey traffic distribution to its original 5 percent. I think we must have accidentally swapped it while testing at some point 🙃 (We should probably let the data team know, in case this caused an awkward spike or dip in our web NPS?)
- and another! https://github.com/DoSomething/phoenix-next/pull/1901/commits/eb46eca6c36bb4a8a1f06a0a048f57129b7e2dad fixes a typo in the `ContentBlock` graphql fragment for the `imageAlignment` field alias https://dosomething.slack.com/archives/CP2D7UGAU/p1580569340000400

https://github.com/DoSomething/phoenix-next/commit/7e7cb7b29e0a729a2148c057ae0d58c19e84e2cb I'm not completely sure about -- it seems that we've had it this way for a while, but it originally was the top-level context for this event, but we eventually started nesting it _inside_ `context` without spreading the object.
At this point, I'm not sure if it would do more harm then good to start spreading the object since we've been tracking it otherwise for a while? Or if it'll be helpful for when we actually want to start using this data?

### DIFF
--- a/resources/assets/components/blocks/ContentBlock/ContentBlock.js
+++ b/resources/assets/components/blocks/ContentBlock/ContentBlock.js
@@ -22,7 +22,7 @@ export const ContentBlockFragment = gql`
       description
     }
     # Aliasing to avoid conflicting with *required* imageAlignment field in GalleryBlockFragment.
-    contentBlockimageAlignment: imageAlignment
+    contentBlockImageAlignment: imageAlignment
   }
 `;
 
--- a/resources/assets/components/pages/HomePage/HomePage.js
+++ b/resources/assets/components/pages/HomePage/HomePage.js
@@ -102,7 +102,7 @@ class HomePage extends React.Component {
           </article>
 
           {featureFlag('nps_survey') ? (
-            <TrafficDistribution percentage={100} feature="nps_survey">
+            <TrafficDistribution percentage={5} feature="nps_survey">
               <DismissableElement
                 name="nps_survey"
                 render={(handleClose, handleComplete) => (
--- a/resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
+++ b/resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
@@ -64,7 +64,7 @@ class SocialShareTray extends React.Component {
       action: 'button_clicked',
       category: EVENT_CATEGORY,
       label: 'facebook_messenger',
-      contetxt: { url: trackLink },
+      context: { url: trackLink },
     });
 
     if (getFormattedScreenSize() === 'large') {
--- a/resources/assets/helpers/index.js
+++ b/resources/assets/helpers/index.js
@@ -762,7 +762,7 @@ export function handleFacebookShareClick(href, trackingData) {
     action: 'button_clicked',
     category: 'social_share',
     label: 'facebook',
-    context: { trackingData, url: href },
+    context: { ...trackingData, url: href },
   });
 
   // @todo 12/13/2018: Use the showFacebookShareDialog to track
@@ -782,7 +782,7 @@ export function handleTwitterShareClick(href, trackingData, quote = '') {
     action: 'button_clicked',
     category: 'social_share',
     label: 'twitter',
-    context: { trackingData, url: href },
+    context: { ...trackingData, url: href },
   });
 
   showTwitterSharePrompt(href, quote);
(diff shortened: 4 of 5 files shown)


resources/assets/components/pages/HomePage/HomePage.js
```diff
@@ -102,7 +102,7 @@ class HomePage extends React.Component {
           </article>
 
           {featureFlag('nps_survey') ? (
-            <TrafficDistribution percentage={100} feature="nps_survey">
+            <TrafficDistribution percentage={5} feature="nps_survey">
```
weerd: Oh crappola. That was me when working on site grid 😫 I thought I reset it but must have missed pushing that commit up 🤦‍♂️Twas too many things to keep track of, blarg.

weerd:
> At this point, I'm not sure if it would do more harm then good to start spreading the object since we've been tracking it otherwise for a while? Or if it'll be helpful for when we actually want to start using this data?

I’d prefer to fix it so we are doing it correctly moving forward. We can let the Data Team know, but best to correct it, otherwise we end up questioning again in the future why it’s different in this one scenario.

resources/assets/components/pages/HomePage/HomePage.js
mendelB: Gives us more survey data is all! 😈
I'll let the data team know about this one as well as the other analytics thing.

weerd: Yeah, I was about to flag this to @ngjo as well and see who else might need to know.

ngjo: Does anyone remember the context as to why we were only prompting 5% of homepage traffic?

mendelB: All's I could find is this https://www.pivotaltracker.com/story/show/154484254 (which applied to the campaign pages, but I assume was the impetus for the homepage as well)

weerd: I do not remember well. Maybe it was to help curb a user experiencing the NPS modal too often in case they were going to the site on different devices? At 100%, if the user see the NPS modal on web and complete/close it, it shouldn't appear again (unless they clear their local browser storage), and if they go again to website on phone, they'll definitely see it again. At 5% it's much less likely they see it every time on different devices. I _doubt_ that's the reason though and something more related to what the Data Team wanted.

mendelB:
Update: https://github.com/DoSomething/phoenix-next/pull/1901/commits/901f370a8888466730deafb7f84a295da45bd52a adds some extra timeout time to the initial beta-referral-page test which seems to have been timing out a lot lately https://circleci.com/build-insights/gh/DoSomething/phoenix-next/master

My uneducated theory is that since this is the initial page load in the suite, it somehow takes longer (caching or whatever initial setup might take place in the first-page load??) I figure it's worth a try!

resources/assets/components/blocks/ContentBlock/ContentBlock.js
```diff
@@ -22,7 +22,7 @@ export const ContentBlockFragment = gql`
       description
     }
     # Aliasing to avoid conflicting with *required* imageAlignment field in GalleryBlockFragment.
-    contentBlockimageAlignment: imageAlignment
+    contentBlockImageAlignment: imageAlignment
```
weerd: camel case! 🐫

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
```diff
@@ -64,7 +64,7 @@ class SocialShareTray extends React.Component {
       action: 'button_clicked',
       category: EVENT_CATEGORY,
       label: 'facebook_messenger',
-      contetxt: { url: trackLink },
+      context: { url: trackLink },
```
weerd: 🍝 typo culprit!

resources/assets/helpers/index.js
```diff
@@ -762,7 +762,7 @@ export function handleFacebookShareClick(href, trackingData) {
     action: 'button_clicked',
     category: 'social_share',
     label: 'facebook',
-    context: { trackingData, url: href },
+    context: { ...trackingData, url: href },
```
weerd: We should holla at the Data Team about this too! They're gonna have a fun day with us today, haha 😬

PR #1902, opened 2020-02-03
Newsletter CTA Cypress Tests  

Vmack:
### What's this PR do?

This pull request adds cypress tests to the newsletter cta

👀

N/A

### Relevant tickets

References [Pivotal #169966143](https://www.pivotaltracker.com/story/show/169966143).

### Checklist


cypress/integration/company-page.js
```diff
@@ -9,5 +9,10 @@ describe('Company Page', () => {
     cy.visit(`/us/${exampleAboutPage.page.fields.slug}`);
 
     cy.contains('h1', 'Hello World');
+    it('Renders a Newsletter Cta', () => {
```
katiecrane: so this test is currently inside the first test `it('Renders a Company Page', () => {` so make sure to move it outside of the closing `});` from this first test

cypress/integration/company-page.js
```diff
@@ -10,4 +10,10 @@ describe('Company Page', () => {
 
     cy.contains('h1', 'Hello World');
   });
+
+  it('Renders a Newsletter Cta', () => {
+    cy.withFeatureFlags({ company_pages: true }).visit(
```
katiecrane: now you can remove this feature flag since we don't need it anymore and update this to be just `cy.visit`

cypress/integration/company-page.js
```diff
@@ -10,4 +10,33 @@ describe('Company Page', () => {
...
+
+    cy.get('.email-form__input').type('vmack@dosomething.org');
+    cy.get('.email-form__button').click();
+
+    cy.contains('Thank You For Submitting Your Email');
+
+    cy.get('.email-form__input').type('hsjadcusdcg');
+    cy.get('.email-form__button').click();
+
+    cy.contains('The email must be a valid email address.');
```
katiecrane: These should be broken up into 3 separate tests in this file:
- rendering the popover
- successful submission
- failed submission

cypress/integration/scholarship-page.js
```diff
@@ -0,0 +1,14 @@
...
+
+describe('Pages', () => {
+  // Configure a new "mock" server before each test:
+  beforeEach(() => cy.configureMocks());
+
+  it('Does this thing', () => {
+    cy.withState(exampleScholarshipPage).visit(
+      '/us/facts/test-11-facts-about-testing',
+    );
+  });
```
katiecrane: This isn't actually testing anything right now! It's just visiting the page. This line needs to be added `    cy.contains('Pays To Do Good');` so we're actually making sure the popover appears.

cypress/integration/scholarship-page.js
```diff
@@ -0,0 +1,14 @@
+/// <reference types="Cypress" />
+
+import exampleScholarshipPage from '../fixtures/contentful/exampleScholarshipPage';
+
+describe('Pages', () => {
+  // Configure a new "mock" server before each test:
+  beforeEach(() => cy.configureMocks());
+
+  it('Does this thing', () => {
```
katiecrane: Let's change this to be about what the test is actually doing!

resources/assets/components/pages/CompanyPage/CompanyPage.js
```diff
@@ -38,11 +38,13 @@ const CompanyPageTemplate = props => {
       <main className="wrapper base-12-grid company-page">
         <article className="grid-wide card rounded border border-solid border-gray-300">
           {coverImage.url ? (
-            <LazyImage
-              className="w-full"
-              alt={coverImage.description || 'Page Cover Image'}
-              src={contentfulImageUrl(coverImage.url, 1440, 620)}
-            />
+            <div style={{ maxHeight: '400px' }}>
```
katiecrane: @DFurnes I'm not super familliar with the company pages styling - when we did this during our jam did you mean for it to be a permanent change or was it just to make the testing easier for us?

resources/assets/components/pages/CompanyPage/CompanyPage.js
```diff
@@ -62,7 +64,7 @@ const CompanyPageTemplate = props => {
             render={(handleClose, handleComplete) => (
               <DelayedElement delay={3}>
                 <CtaPopover
-                  title="PAYS TO DO GOOD"
+                  title="Pays To Do Good"
```
katiecrane: This is a great change, so much easier to look at!

cypress/integration/scholarship-page.js
Vmack: so instead of . visiting the url it should look for 'Pays To Do Good' ?

katiecrane: We still have to visit the url in order to render the page and the popover, but just visiting the page doesn't really test anything so we have to both visit the page, and check that the popover is there, which is what `cy.contains('Pays To Do Good');` will take care of. Does that make sense?

Vmack: @katiecrane ah ok got it!

cypress/integration/company-page.js
```diff
@@ -19,4 +19,33 @@ describe('Company Page', () => {
...
+
+  cy.get('.email-form__input').type('vmack@dosomething.org');
+  cy.get('.email-form__button').click();
+
+  cy.contains('Thank You For Submitting Your Email');
+
+  cy.get('.email-form__input').type('hsjadcusdcg');
+  cy.get('.email-form__button').click();
+
+  cy.contains('The email must be a valid email address.');
```
katiecrane: Copying this over since it disappeared but is still relevant!

These should be broken up into 3 separate tests in this file:
- rendering the popover
- successful submission
- failed submission

cypress/integration/company-page.js
```diff
@@ -19,4 +19,65 @@ describe('Company Page', () => {
...
+    cy.mockGraphqlOp('CompanyPageQuery', {
+      companyPageBySlug: {
+        title: 'Easy Scholarships',
+        slug: 'easy-scholarships',
+      },
+    });
+
+    cy.visit(`/us/about/easy-scholarships`);
+
+    cy.contains('Pays To Do Good');
```
katiecrane: We don't need to check for this here - we only care about testing for a successful submission in this test

cypress/integration/company-page.js
```diff
@@ -19,4 +19,65 @@ describe('Company Page', () => {
...
```
katiecrane: We don't need this piece here either!

cypress/integration/company-page.js
```diff
@@ -19,4 +19,65 @@ describe('Company Page', () => {
...
+  });
+
+  it('Successful Submission', () => {
```
katiecrane: I think it'd be helpful if this was a little more descriptive, something like `Handles a successful newsletter CTA submission` and something similar about a failed submission for the other one

resources/assets/components/pages/CompanyPage/CompanyPage.js
DFurnes: Ooooh, right! Yeah, we should see what the proper height would be to fit [this cover image aspect ratio](https://github.com/DoSomething/phoenix-next/blob/ef7d63291265b70a45528fcfe62abb924e2fd449/resources/assets/components/pages/CompanyPage/CompanyPage.js#L44) (since I don't think these ever actually reach the full 1440px width).

I wonder if this is something we could build in to [`LazyImage`](https://github.com/DoSomething/phoenix-next/blob/cdcb01a6cf6128a6d75b25e477b7a126ee3c194b/resources/assets/components/utilities/LazyImage.js) (like we provide an image aspect ratio and it makes sure that the image is fit to that size)?

DFurnes: Oh, actually it looks like some of these cover images are taller than those provided dimensions (like [here](https://www.dosomething.org/us/about/our-people), which I'm seeing as 627px tall... that's strange!)

katiecrane: Hmm I don't really know about aspect ratios but based on [this line](https://github.com/DoSomething/phoenix-next/blob/ef7d63291265b70a45528fcfe62abb924e2fd449/resources/assets/components/pages/CompanyPage/CompanyPage.js#L44) wouldn't the max height be `620`? Is there more to it than that? Seems like there might be!

mendelB: I think the `w-full` forced the image to stretch full width which in turn causes the image to bump up in height! In a plain ``, it seems to indeed respect the specified API dimensions 
image

mendelB: I suspect there might be company pages - which, having been ported over from the legacy pages - might not have proper dimensions and would perhaps be awkwardly verticle... which _might_ be why we used the `w-full` here? cc @lkpttn

katiecrane: Maybe we just toss these changes because they aren't really relevant to this PR? We stumbled upon doing this because it made the testing a bit easier because we could see the page better

mendelB: Maybe worth a bug or chore ticket to dig into this! I've noticed that massively tall placeholder lazy load sometimes when my connection was slower

katiecrane: @mendelB Is [this card](https://www.pivotaltracker.com/story/show/171161258) fair for this work?

Vmack: great, I removed it

mendelB: > @mendelB Is [this card](https://www.pivotaltracker.com/story/show/171161258) fair for this work?

Yeah! 🙌 and maybe also worth linking here for the context on
1. what the deal is with these images and our theory on why we wanted full width here.
2. @DFurnes idea  https://github.com/DoSomething/phoenix-next/pull/1902#discussion_r376464904 about embedding aspect ratio into the LazyImage which would be neat!

katiecrane: oh great call! done!

PR #1905, opened 2020-02-04
Campaign page design qa

lkpttn:
### What's this PR do?
A couple small visual updates to the campaign page, page navigation and campaign info components.

- Changes CampaignPage background to white
- Changes PageNavigation highlight colors and adds top/bottom borders
- Ups the size of the Campaign Info component to 18px, and changes the color of item titles

Check out a couple campaign pages

### Relevant tickets

References [Pivotal #171080775](https://www.pivotaltracker.com/story/show/171080775).

Screen Shot 2020-02-04 at 3 49 50 PM

//

Screen Shot 2020-02-04 at 3 50 07 PM

### DIFF
--- a/resources/assets/components/utilities/PageNavigation/PageNavigation.js
+++ b/resources/assets/components/utilities/PageNavigation/PageNavigation.js
@@ -63,7 +63,7 @@ class PageNavigation extends React.Component {
       <div
         ref={node => (this.node = node)}
         id="page-navigation"
-        className={classnames('page-navigation mt-3 -mb-3', {
+        className={classnames('page-navigation mb-8', {
           'is-stuck': this.state.isStuck,
         })}
       >
--- a/resources/assets/components/blocks/CampaignInfoBlock/campaign-info-block.scss
+++ b/resources/assets/components/blocks/CampaignInfoBlock/campaign-info-block.scss
@@ -5,6 +5,7 @@
 .campaign-info {
   margin: 0 auto;
   max-width: 450px;
+  font-size: theme('fontSize.base');
 
   .affiliate-logo {
     max-height: 50px;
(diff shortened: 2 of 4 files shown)


resources/assets/components/utilities/PageNavigation/PageNavigation.js
```diff
@@ -63,7 +63,7 @@ class PageNavigation extends React.Component {
       <div
         ref={node => (this.node = node)}
         id="page-navigation"
-        className={classnames('page-navigation mt-3 -mb-3', {
+        className={classnames('page-navigation mb-8', {
```
weerd: You sure this works correctly... I remember this needing to be a negative margin... 🤔

weerd:
> (I do see a lot of stylesheet rules though -- maybe worth Tailwind-ifying some of em if possible?)

Agreed. I had a similar thought but wasn't sure if out of scope for these intended fixes 🤷‍♂

resources/assets/components/utilities/PageNavigation/PageNavigation.js
lkpttn: I didn't see anything visually wrong in my testing, do you remember what the negative margin provided?

lkpttn:
> > (I do see a lot of stylesheet rules though -- maybe worth Tailwind-ifying some of em if possible?)
> 
> Agreed. I had a similar thought but wasn't sure if out of scope for these intended fixes 🤷‍♂

Yeah it would be fun to come back to this but I have just a day or so to get a couple quick things updated before Harry's gets super big.

resources/assets/components/utilities/PageNavigation/PageNavigation.js
weerd: I think it may have been with different device screen sizes. If you've tested this change on the different devices sizes and all seems 👌, then I'm good. Just wanted to confirm it's all been checked!

PR #1906, opened 2020-02-04
Updating trackAnalyticsEvent() calls in numerous components (Part 3)

weerd:
### What's this PR do?

This pull request updates the information passed to `trackAnalyticsEvent()` in numerous components to follow the updates established in #1897. 

👀 

### Relevant tickets

### Checklist

### DIFF
--- a/resources/assets/components/utilities/DismissableElement/DismissableElement.js
+++ b/resources/assets/components/utilities/DismissableElement/DismissableElement.js
@@ -18,14 +18,16 @@ const DismissableElement = ({ name, render, context }) => {
       // Mark the element as "dismissed" in local storage & hide it.
       setStorage(`dismissed_${name}`, 'timestamp', Date.now());
 
-      trackAnalyticsEvent({
+      trackAnalyticsEvent(`dismissed_${name}`, {
+        // @Question: maybe this should just be element_dismissed or even button_clicked?
+        // Or maybe it could be passed in and dependent on the element in question, since
+        // sometimes it is a modal and other times a delayed element...
+        // For context, there already is a modal_opened action, so for modals, a modal_closed
+        // would make sense to be passed in.
+        action: 'dismissable_element_dismissed',
+        category: 'site_action',
+        label: name,
         context,
-        metadata: {
-          category: 'site_action',
-          noun: name,
-          target: 'dismissable_element',
-          verb: 'dismissed',
-        },
       });
     }
     setShowElement(false);
--- a/resources/assets/components/utilities/DismissableElement/DismissableElement.js
+++ b/resources/assets/components/utilities/DismissableElement/DismissableElement.js
@@ -19,8 +19,8 @@ const DismissableElement = ({ name, render, context }) => {
       setStorage(`dismissed_${name}`, 'timestamp', Date.now());
 
       trackAnalyticsEvent(`dismissed_${name}`, {
-        // @see discussion regarding reducing this event action name to simpler version
-        // https://github.com/DoSomething/phoenix-next/pull/1906/files#r374962200
+        // @TODO: will discuss with Data Team possibility of reducing
+        // the action to just "element_dismissed".
         action: 'dismissable_element_dismissed',
         category: 'site_action',
         label: name,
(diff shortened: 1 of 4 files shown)


resources/assets/components/utilities/DismissableElement/DismissableElement.js
```diff
@@ -18,14 +18,16 @@ const DismissableElement = ({ name, render, context }) => {
       // Mark the element as "dismissed" in local storage & hide it.
       setStorage(`dismissed_${name}`, 'timestamp', Date.now());
 
-      trackAnalyticsEvent({
+      trackAnalyticsEvent(`dismissed_${name}`, {
+        // @Question: maybe this should just be element_dismissed or even button_clicked?
+        // Or maybe it could be passed in and dependent on the element in question, since
+        // sometimes it is a modal and other times a delayed element...
+        // For context, there already is a modal_opened action, so for modals, a modal_closed
+        // would make sense to be passed in.
```
weerd: > Maybe this should just be `element_dismissed` or even `button_clicked`? Or maybe it could be passed in and dependent on the element in question, since sometimes it is a modal and other times a delayed element... For context, there is already a `modal_opened` action, so for modals, a `modal_closed` would make sense to be passed in.

@mendelB this is the question I posed to you in Slack. I can remove from code post a discussion.

resources/assets/components/utilities/Embed/Embed.js
```diff
@@ -35,17 +35,13 @@ const Embed = props => {
...
   /*
   const onClick = () => {
-    trackAnalyticsEvent({
+    trackAnalyticsEvent('clicked_link_embed', {
+      action: 'link_clicked',
+      category: 'site_action',
+      label: 'embed',
       context: {
         url,
       },
```
weerd: This event was commented out with explanation above it in comments. I updated it regardless 🤷‍♂

resources/assets/components/utilities/DismissableElement/DismissableElement.js
mendelB: Totally agreed that this can be confusing coupled with a modal event for embedded modals!

The one catch about this event is that it targets the 'dismissal' of the feature _before_ completion. Meaning it tracks when we actually mark this element as 'dismissed' to be hiddne for 30 days via local storage, and _not_ when the element was marked as 'completed'. 

So there's a bit of special analytical data we can deduce from this in addition to the regular modal closed event which is tracked whenever the modal is closed regardless of if the embedded feature was 'completed' or not. 

Maybe worth renaming though for more clarity? Though we have been tracking this (specifically intended) for the CTA Popover for a while now, so we'd have to deal with the data ramification of changing it. 

Or do you think that this is _still_ unnecessary in the case of a modal?

weerd: 🤔 Yeah... maybe just renaming it to `element_dismissed` to go with the flow of the other simple action names in the system.

mendelB: lol yeah agreed, I'm pro that update!

weerd: I just wish these analytics platforms were more malleable to our whims when we change our minds 😅

mendelB: And do you think that distinction makes sense? The idea that there's a modal close, but additionally, there's the 'element' or 'feature' itself being 'dismissed' _as opposed_ to complete (since we wouldn't track this if it were complete)?

weerd: Ya, I think so. The clarification you provided makes sense to me 👍

weerd: I note this with the Data Team and see what they say.

resources/assets/components/utilities/DismissableElement/DismissableElement.js
```diff
@@ -18,14 +18,13 @@ const DismissableElement = ({ name, render, context }) => {
       // Mark the element as "dismissed" in local storage & hide it.
       setStorage(`dismissed_${name}`, 'timestamp', Date.now());
 
-      trackAnalyticsEvent({
+      trackAnalyticsEvent(`dismissed_${name}`, {
+        // @see discussion regarding reducing this event action name to simpler version
+        // https://github.com/DoSomething/phoenix-next/pull/1906/files#r374962200
```
weerd: @mendelB 💃

PR #1907, opened 2020-02-05
Campaigns from laravel to react

lindsayrmaher:
### What's this PR do?

This pull request lays some major groundwork for transitioning our explore campaigns page from laravel to React & Graphql! There is some Graphql work that needs to get pulled in here before we can fully realize the pagination etc, but this is establishing the base design and setting up a feature flag so that our routes are unaffected until we're ready to launch.

We are now setup with our graphql query and styling to mimic the current page, so this ticket is covering everything we need with the exception of pagination.

Current Page on QA: 
Screen Shot 2020-02-10 at 1 36 17 PM

Page in Local using QA data:
Screen Shot 2020-02-13 at 12 01 29 PM

Screen Shot 2020-02-13 at 12 01 08 PM

Screen Shot 2020-02-13 at 11 48 43 AM

We are updating these to lay the groundwork on the campaigns page for filtering!

### Relevant tickets

References [Pivotal #171069006](https://www.pivotaltracker.com/story/show/171069006).

### Checklist


config/features.php
```diff
@@ -16,4 +16,5 @@
     'voter_reg_modal' => env('DS_ENABLE_VOTER_REG_MODAL', false),
     'referral_campaign_ids' => explode(',', env('DS_REFERRAL_CAMPAIGN_IDS')),
     'default_referral_campaign_id' => env('DS_DEFAULT_REFERRAL_CAMPAIGN_ID'),
+    'dynamic_explore_campaigns' => env('DS_DYNAMIC_CAMPAIGNS_PAGE_ENABLED', false),
```
mendelB: nitpicky -- but perhaps just for consistencies sake we maintain the `DS_ENABLE...` format?

docs/development/features/affiliate-scholarship-block.md
```diff
@@ -17,7 +17,7 @@ There are a few separate components controlling what is displayed in the card, a
 The URL for the current user will need to contain the following UTM parameters:
 
 - `utm_source` with a value containing the word `scholarship`. e.g. `utm_source=scholarship_listing`
-- `utm_campaign` with the value of the referring affiliate's UTM label e.g. `utm_campaign=fastweb` (explained more below). If the value is a snake_cased string, the value should be the _first_ single value in the string e.g. `utm_campaign=fastweb_2019_07`.
+- `utm_campaign` with the value of the referring affiliate's UTM label e.g. `utm_campaign=fastweb` (explained more below). If the value is a snake*cased string, the value should be the \_first* single value in the string e.g. `utm_campaign=fastweb_2019_07`.
```
mendelB: I imagine this was an auto code editor formatting change, but I think it got tripped up on the `snake_cased` to `snake*cased` since it isn't a markdown format but rather specifically an underscore? (Not sure about the `\_first*` either)

resources/assets/components/pages/CampaignsPage/CampaignsPage.js
```diff
@@ -0,0 +1,75 @@
...
+        cursor
+        node {
+          id
+          startDate
+          campaignWebsite {
+            url
+            coverImage {
+              url
+              description
+            }
```
mendelB: I think it'd be neat to utilize the `showcaseImage` here!

resources/assets/components/pages/CampaignsPage/CampaignsPage.js
```diff
@@ -0,0 +1,75 @@
...
+
+const CampaignsPageTemplate = () => {
+  const { error, loading, data } = useQuery(CAMPAIGNS_QUERY, {
+    notifyOnNetworkStatusChange: true,
+  });
+
+  // const { endCursor, hasNextPage } = get(data, 'campaigns.pageInfo', {});
+
+  if (error) {
+    return <ErrorBlock error={error} />;
```
mendelB: Since this is a full-page, maybe worth using the [`ErrorPage`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/pages/ErrorPage.js) here?

resources/assets/components/pages/CampaignsPage/CampaignsPage.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  return (
+    <>
+      <SiteNavigationContainer />
+      <main className="md:w-3/4 mx-auto">
+        <h1 className="w-full my-6 pl-3">Campaigns For All Causes</h1>
+        {!loading ? (
+          <CampaignCards campaigns={data.campaigns.edges} />
```
mendelB: Wondering if we can use the [`GalleryBlock`](https://github.com/DoSomething/phoenix-next/blob/3f01a957318661e4a213876594b2849f53183c28/resources/assets/components/blocks/GalleryBlock/GalleryBlock.js) here which would render [`CampaignGalleryItem`s](https://github.com/DoSomething/phoenix-next/blob/3f01a957318661e4a213876594b2849f53183c28/resources/assets/components/utilities/Gallery/templates/CampaignGalleryItem/CampaignGalleryItem.js) instead of creating our own custom campaign gallery? (The formatting might be _slightly_ different than the current explore campaigns page, but probably in a good way, since it's closer to @lkpttn envisioned gallery nodes formatting)

lindsayrmaher: @mendelB That's a good point! I was struggling with the level of nesting with the data, but I would be nice to be consistent with design across the site. I reached out to Leah to see if there's a preference to stick with the original styling or try the new gallery.

mendelB: Totes! If there's a straightforward way we could map the campaign website data so that it's the proper format for the gallery then it'd be oh so convenient ☺️ 

But if it gets too hairy, or if we're not cool with the slight design tweak then custom gallery it must be!

docs/development/features/affiliate-scholarship-block.md
DFurnes: Yeah, something seems off here! 👻

weerd:
@lindsayrmaher could we get some different device size screenshots as well 📸

docs/development/features/affiliate-scholarship-block.md
weerd: Ya... peculiar. Should we have Prettier handle the docs files or ignore them? I've seen it get a little messed up in the past as well, but having them formatted is a nice-to-have, as long as it does it correctly 🤔

resources/assets/components/pages/CampaignsPage/CampaignCard.js
```diff
@@ -0,0 +1,55 @@
...
+      </a>
+    </li>
+  );
+};
+
+Campaign.propTypes = {
+  campaign: PropTypes.object.isRequired,
+};
+
+const CampaignCards = ({ campaigns }) => (
```
DFurnes: It might be cleaner to keep this (the `CampaignCards` gallery & mapped items) in the page itself, and just have this be a reusable `CampaignCard` element that we could place anywhere it makes sense...

resources/assets/components/pages/CampaignsPage/CampaignsPage.js
```diff
@@ -0,0 +1,75 @@
...
+        )}
+      </main>
+      <SiteFooter />
+    </>
+  );
+};
+
+CampaignsPageTemplate.propTypes = {};
+
+export default CampaignsPageTemplate;
```
weerd: I think the the export should be named the same as the file. That being said, I wonder if this should be `CampaignsIndexPage` instead of `CampaignsPage` since it might be confused with a regular single campaign page instead of the listing... 🤔

resources/assets/components/pages/CampaignsPage/CampaignCard.js
```diff
@@ -0,0 +1,55 @@
...
+            key={`contentful-campaign${campaign.cursor}`}
+          />
+        );
+      })}
+    </ul>
+  </>
+);
+
+CampaignCards.propTypes = {
+  campaigns: PropTypes.arrayOf(PropTypes.object).isRequired,
```
DFurnes: Let's make a fragment with this component's data needs ([example](https://github.com/DoSomething/phoenix-next/blob/3f01a957318661e4a213876594b2849f53183c28/resources/assets/components/utilities/PostCard/PostCard.js#L15-L37)) that we can export and use in our "Explore Campaigns" query, and then we can enforce those specific fields using [the `propType` helper](https://github.com/DoSomething/phoenix-next/blob/3f01a957318661e4a213876594b2849f53183c28/resources/assets/components/utilities/PostCard/PostCard.js#L121).

resources/assets/components/pages/CampaignsPage/CampaignCard.js
```diff
@@ -0,0 +1,55 @@
...
+    <li className="gallery-item">
+      <a className="wrapper" href={campaign.url}>
+        <img
+          src={appendToQuery(
+            {
+              width: 400,
+              height: 400,
+              fit: 'fill',
+            },
+            campaign.coverImage.url,
```
DFurnes: Let's use [`contentfulImageUrl`](https://github.com/DoSomething/phoenix-next/blob/3f01a957318661e4a213876594b2849f53183c28/resources/assets/helpers/index.js#L119-L145) here.

resources/assets/components/pages/CampaignsPage/CampaignsPage.js
```diff
@@ -0,0 +1,75 @@
...
+import { useQuery } from '@apollo/react-hooks';
+
+import CampaignCards from './CampaignCard';
+import Spinner from '../../artifacts/Spinner/Spinner';
+import ErrorBlock from '../../blocks/ErrorBlock/ErrorBlock';
+import SiteFooter from '../../utilities/SiteFooter/SiteFooter';
+import SiteNavigationContainer from '../../SiteNavigation/SiteNavigationContainer';
+
+const CAMPAIGNS_QUERY = gql`
+  query CampaignsQuery($isOpen: Boolean, $cursor: String) {
```
DFurnes: Just for clarity, let's name this `CampaignsPageQuery` (so it's clearer that's the campaigns we're querying).

resources/assets/components/pages/CampaignsPage/CampaignCard.js
```diff
@@ -0,0 +1,55 @@
...
+      })}
+    </ul>
+  </>
+);
+
+CampaignCards.propTypes = {
+  campaigns: PropTypes.arrayOf(PropTypes.object).isRequired,
+};
+
+export default CampaignCards;
```
weerd: I wonder if this should live outside the `CampaignsPage` since it feels like it'll be used in more places... I could see the Homepage campaign cards being the new design for this and thus used here and on the Homepage, etc.

resources/assets/components/pages/CampaignsPage/CampaignCard.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const CampaignCards = ({ campaigns }) => (
+  <>
+    <ul className="gallery-grid gallery-grid-quartet">
+      {campaigns.map(campaign => {
+        return (
+          <Campaign
+            campaign={campaign.node.campaignWebsite}
+            key={`contentful-campaign${campaign.cursor}`}
```
DFurnes: This key can just be the `campaign.cursor`.

resources/assets/components/pages/CampaignsPage/CampaignCard.js
weerd: Ha! Dave made the [same comment earlier, but I missed it](https://github.com/DoSomething/phoenix-next/pull/1907/files#r377769415).

resources/assets/components/pages/CampaignsPage/CampaignsPage.js
lindsayrmaher: That's a fair point, my only thought was whether it made more sense to show the `errorBlock` since they would like run into problems with the filters/sorting and it would be helpful to still see that part of the Page.

lindsayrmaher:
@DFurnes @weerd & @mendelB I made some updates based on your feedback as well as updating this to utilize the Paginated Campaigns Query that Mendel and I worked on from #1925 . I also added some updated screenshots for all sizes and I can rebase off master once the other PR is approved! I'll do one more push to remove the campaign card component since we no longer need it!

docs/development/features/affiliate-scholarship-block.md
DFurnes: I believe the problem is the `_` in `snake_case` here is being interpreted as Markdown syntax. I think we can fix that  by either escaping the first `_` (so `snake\_cased`) or placing it inside an inline code block so the underscore isn't parsed (so `` `snake_cased` ``).

lindsayrmaher: Good call! fixed in [2d9fa16](https://github.com/DoSomething/phoenix-next/pull/1907/commits/2d9fa16af05998d2241c917c3bd2d066dc820b79)

PR #1900, opened 2020-01-31
Load blocks via GraphQL everywhere!

DFurnes:
### What's this PR do?

This pull request updates the campaign, article and affirmation to load blocks via GraphQL, rather than our PHP Content API. This means we won't have to account for two different ways data might be loaded for these components, and sets us up for easier server-rendering this quarter.

### How should this be reviewed?

I've left comments to explain some of the more interesting bits! 👁‍🗨 

### Any background context you want to provide?

~This is ready for review, but can't be merged until we update the remaining campaigns that use [these "Custom Block" galleries](https://app.contentful.com/spaces/81iqaqpfd8fy/entries?contentTypeHidden=false&id=odKYc1MfB7PVKKwY&contentTypeId=customBlock&order.fieldId=updatedAt&order.direction=descending&displayedFieldIds=updatedAt&displayedFieldIds=author&login=1&filters.0.key=fields.type&filters.0.val=gallery) and sort out [the closed campaigns that use custom blocks on their community pages](https://dosomething.slack.com/archives/C03V8B1HA/p1580421709079400?thread_ts=1580404756.074900&cid=C03V8B1HA).~ And ready to merge and deploy!

### Relevant tickets

References [Pivotal #169216496](https://www.pivotaltracker.com/story/show/169216496).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #1908, opened 2020-02-06
Quick bug fix for Content Blocks image alignment

mendelB:
### What's this PR do?

This pull request patches a quick oversight on my part for GraphQL loaded Content Blocks https://github.com/DoSomething/phoenix-next/pull/1692 where I failed to ensure the _optional_ image alignment field, where we're resolving a field alias and it thus outside our `withoutNulls` safety call, doesn't end up being `null` and overriding the default value (which we process).

👀 

This was causing breakage on the[ 👻 test](https://app.ghostinspector.com/tests/5c4b8976638e692a2327b467) for a content block with no image alignment set on Contentful.

We've swapped over to query page blocks via graphql in #1900

### DIFF
--- a/resources/assets/components/ContentfulEntry/ContentfulEntry.js
+++ b/resources/assets/components/ContentfulEntry/ContentfulEntry.js
@@ -159,7 +159,7 @@ class ContentfulEntry extends React.Component {
             className={className}
             // Resolves the aliases used in the ContentBlockFragment.
             content={json.contentBlockContent}
-            imageAlignment={json.contentBlockImageAlignment}
+            imageAlignment={json.contentBlockImageAlignment || undefined}
             {...withoutNulls(json)}
           />
         );


resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -159,7 +159,7 @@ class ContentfulEntry extends React.Component {
             className={className}
             // Resolves the aliases used in the ContentBlockFragment.
             content={json.contentBlockContent}
-            imageAlignment={json.contentBlockImageAlignment}
+            imageAlignment={json.contentBlockImageAlignment || undefined}
```
weerd: Do we want to just also wrap it in `withoutNulls()` so it shows it's technically being treated similarly as the item below?

mendelB: Ohhhh blast I just merged this 🤦‍♂ that makes good sense to do though!

weerd: haha, post-merge PR 💃

PR #1911, opened 2020-02-06
Updating trackAnalyticsEvent() calls in numerous components (Part 4)

weerd:
### What's this PR do?

This pull request updates the information passed to `trackAnalyticsEvent()` in numerous components to follow the updates established in #1897.

The `SiteNavigation` component and the API middleware turned out having the more complex analytics event triggers to deal with updating, but I tried to find a reasonable compromise in each situation. For the API middleware, as we move to using more GraphQL calls within our components, I could see this middleware being deprecated, so the event triggering could get even simpler in the future (without the need to be so dynamic in the middleware).

👀 

### Relevant tickets

### Checklist

### DIFF
--- a/resources/assets/components/SiteNavigation/SiteNavigation.js
+++ b/resources/assets/components/SiteNavigation/SiteNavigation.js
@@ -31,34 +31,21 @@ class SiteNavigation extends React.Component {
     };
   }
 
-  analyzeEvent = (event, analytics = {}) => {
-    const targets = {
-      A: 'link',
-      BUTTON: 'button',
-    };
+  analyzeEvent = (name, data = {}) => {
+    console.log('💩 event: ', name);
+    console.log('🧻 analytics: ', data);
 
-    const inferredTarget = get(targets, event.target.tagName, 'element');
+    const { action, category, label, context = {} } = data;
 
-    const target = get(analytics, 'target', inferredTarget);
-
-    const context = get(analytics, 'context', {});
-
-    const label = get(analytics, 'label', null);
-
-    trackAnalyticsEvent({
+    trackAnalyticsEvent(name, {
+      action,
+      category,
+      label,
       context: {
         ...getPageContext(),
         ...getUtmContext(),
         referrer: document.referrer,
         ...context,
-      },
-      metadata: {
-        adjective: get(analytics, 'adjective', label),
-        category: get(analytics, 'category', 'navigation'),
-        label,
-        noun: get(analytics, 'noun', 'nav_link'),
-        target,
-        verb: get(analytics, 'verb', 'clicked'),
       },
     });
   };
@@ -94,7 +81,9 @@ class SiteNavigation extends React.Component {
   };
 
   handleOnClickLink = (event, analytics = {}) => {
-    this.analyzeEvent(event, analytics);
+    console.log('🐞 analytics: ', analytics);
+
+    this.analyzeEvent(analytics.name, analytics);
 
     this.setState({
       activeSubNav: null,
@@ -155,7 +144,14 @@ class SiteNavigation extends React.Component {
           <div className="logo-nav">
             <a
               href="/"
-              onClick={e => this.handleOnClickLink(e, { label: 'homepage' })}
+              onClick={e =>
+                this.handleOnClickLink(e, {
+                  name: 'clicked_nav_link_homepage',
+                  action: 'link_clicked',
+                  category: 'navigation',
+                  label: 'homepage',
+                })
+              }
             >
               <DoSomethingLogo />
             </a>
@@ -182,6 +178,9 @@ class SiteNavigation extends React.Component {
                         href="/campaigns"
                         onClick={e =>
                           this.handleOnClickLink(e, {
+                            name: 'clicked_nav_link_causes',
+                            action: 'link_clicked',
+                            category: 'navigation',
                             label: 'causes',
                           })
                         }
@@ -218,7 +217,9 @@ class SiteNavigation extends React.Component {
                             href="/us/causes/education"
                             onClick={e => {
                               this.handleOnClickLink(e, {
-                                noun: 'subnav_link',
+                                name: 'clicked_subnav_link_causes_education',
+                                action: 'link_clicked',
+                                category: 'navigation',
                                 label: 'causes_education',
                               });
                             }}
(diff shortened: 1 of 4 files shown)


resources/assets/middleware/api.js
```diff
@@ -93,22 +93,20 @@ const postRequest = (payload, dispatch, getState) => {
         },
       };
 
-      trackAnalyticsEvent({
+      const verb =
+        statusCode === 200 && postType === 'signup' ? 'found' : 'completed';
+
+      trackAnalyticsEvent(`${verb}_${formatEventNoun(postType)}`, {
+        action: `${postType}_${verb}`,
+        category: 'campaign_action_test',
```
mendelB: 👾 *_test

resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -364,8 +436,10 @@ class SiteNavigation extends React.Component {
                 })}
                 onClick={e =>
                   this.handleOnClickToggle(e, 'SearchSubNav', {
+                    name: 'clicked_nav_button_search_form_toggle',
+                    action: 'link_clicked',
+                    category: 'navigation',
```
mendelB: I don't know just how useful this would be, but maybe worth defining this as a constant above since we pretty much only use this same category across these events?

resources/assets/middleware/api.js
```diff
@@ -93,22 +93,20 @@ const postRequest = (payload, dispatch, getState) => {
...
+        label: campaignId || formatEventNoun(postType), // @TODO: make this the campaign title if available; but also may need to get passed in as an argument.
```
mendelB: Ooh this will be helpful! I guess we could try and read campaign title from `state` since we do have access? But that's not relevant to this work per se

resources/assets/middleware/api.js
weerd: whoop! nice 👀

resources/assets/components/SiteNavigation/SiteNavigation.js
weerd: Oh ya, I was considering that and had meant to discuss but forgot.

I was also thinking what if we defined an object style constant... something like:

```js
const analyticsCategories = {
  authentication: 'authentication',
  campaignAction: 'campaign_action',
  navigation: 'navigation',
  search: 'search',
  signup: 'signup',
};
```

Then when calling them, we'd likely have our editors auto suggested from the established categories:

```js
import { analyticsCategories } from '../some/constants/location';
...
  category: analyticsCategories.navigation,
...
```

The benefit is that our editors could auto-suggest as we type them and the benefit of not causing any typos in the resulting string.

On a similar note, @DFurnes had suggested potentially doing this via a function:

```js
useAnalyticsCategory('Navigation');
```

One reason I like the object is that it could easily auto-suggest from the list of properties, but maybe a function like Dave suggests would be more robust if we throw an error if the category is not found in some array list of specified event categories... 

Thoughts? 🤔

resources/assets/middleware/api.js
weerd: Yep, my thoughts too. Just seemed out of scope for this PR!

resources/assets/components/SiteNavigation/SiteNavigation.js
mendelB: Oooh I think formalizing this is a _great_ idea!

I like the object idea for the pro's you list, but also see the benefit of the error handling 🤔 

Maybe as a sort of combo, we could use the object notation, and have the analytics helper run the validation to make sure it's a valid supported property?

weerd: Oh interesting idea! Yeah, I can try that 💃

resources/assets/helpers/analytics.js
```diff
@@ -20,6 +20,21 @@ import { debug, stringifyNestedObjects, withoutValueless } from '.';
...
+  focusedField: 'focused_field',
+  modal: 'modal',
+  navigation: 'navigation',
+  onboarding: 'onboarding',
+  search: 'search',
+  signup: 'signup',
+  siteAction: 'site_action',
+  socialShare: 'social_share',
+  waypoint: 'waypoint',
+};
```
weerd: @mendelB new constant referencing an object with all the event categories 🎉

resources/assets/helpers/analytics.js
```diff
@@ -295,6 +310,14 @@ export function trackAnalyticsEvent(name, metadata = {}) {
   // checking against whether name is a string or object or will error out.
   const { action, category, label, context = {}, service } = metadata;
 
+  // @TODO: uncomment out this code once all the events have had their categories
+  // updated to use the new EVENT_CATEGORIES object.
+  // if (!EVENT_CATEGORIES[category]) {
+  //   console.error('The event category specified is not valid!');
```
weerd: Will uncomment this in upcoming cleanup PR; commented out for now so it wouldn't get angry for all the events not yet using the new `EVENT_CATEGORIES` object.

weerd:
image

@mendelB here's how it looks like in an editor (VSCode) as it makes suggestions for what category to choose 💃

resources/assets/helpers/analytics.js
mendelB: Could we even activate this now, since the other categories should still be within the list since they'll _both_ be the same string value?

weerd: The properties (so the indexes for when grabbing them via array syntax) are camelCase 😉

mendelB: Ah right, I think we'd have to switch this logic around then? Since we're passing through the values themselves from the event calls, so this would _only_ be valid for single words in both cases. Maybe we'd need to do something like `Object.values(EVENT_CATEGORIES).includes(category)`?

weerd: oh shiza! 🤦‍♂ Yeah, I did that quick since this didn't need to be part of this PR. I'll remove for now and handle it properly in upcoming cleanup PR when I switch to using the new constant!

mendelB: 🙅‍♂ Diego, you can only code in R on the LIRR! so it is written.

weerd: @mendelB I think another reason I like the conditional check inside of the `trackAnalyticsEvent()` function, is that if someone ends up not grabbing the category from `EVENT_CATEGORIES`, and just manually adds the string to the event payload, it still gets checked here against the approved list of categories 💃

PR #1914, opened 2020-02-07
Social Drive Action should constrain its own size if hiding page views.

DFurnes:
### What's this PR do?

This pull request fixes the layout of the Social Drive block if page views are hidden:

Screen Shot 2020-02-07 at 4 30 13 PM

Screen Shot 2020-02-07 at 4 30 28 PM

👀

There may be a better long-term solution for handling variable-width blocks like this, but this works nicely in the meantime!

### Relevant tickets

References [Pivotal #171161967](https://www.pivotaltracker.com/story/show/171161967).

### Checklist

### DIFF
--- a/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
+++ b/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
@@ -63,6 +63,7 @@ class SocialDriveAction extends React.Component {
   render() {
     const {
       link,
+      fullWidth,
       shareCardDescription,
       shareCardTitle,
       hidePageViews,
@@ -73,7 +74,11 @@ class SocialDriveAction extends React.Component {
       <div
         className={classNames('clearfix pb-6', { 'lg:flex': !hidePageViews })}
       >
-        <div className="social-drive-action lg:w-2/3 lg:pr-3">
+        <div
+          className={classNames('social-drive-action', {
+            'lg:w-2/3 lg:pr-3': !fullWidth,
+          })}
+        >
           <Card title={shareCardTitle} className="rounded bordered">
             {shareCardDescription ? (
               <div className="p-3">
@@ -155,6 +160,7 @@ class SocialDriveAction extends React.Component {
 
 SocialDriveAction.propTypes = {
   campaignId: PropTypes.string,
+  fullWidth: PropTypes.bool,
   link: PropTypes.string.isRequired,
   pageId: PropTypes.string,
   shareCardTitle: PropTypes.string,
@@ -166,6 +172,7 @@ SocialDriveAction.propTypes = {
 
 SocialDriveAction.defaultProps = {
   campaignId: null,
+  fullWidth: false,
   shareCardDescription: null,
   shareCardTitle: 'Your Online Drive',
   pageId: null,
--- a/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
+++ b/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
@@ -73,11 +73,7 @@ class SocialDriveAction extends React.Component {
       <div
         className={classNames('clearfix pb-6', { 'lg:flex': !hidePageViews })}
       >
-        <div
-          className={classNames('social-drive-action', {
-            'lg:w-2/3 lg:pr-3': !hidePageViews,
-          })}
-        >
+        <div className="social-drive-action lg:w-2/3 lg:pr-3">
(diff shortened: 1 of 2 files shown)


resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -73,11 +73,7 @@ class SocialDriveAction extends React.Component {
       <div
         className={classNames('clearfix pb-6', { 'lg:flex': !hidePageViews })}
       >
-        <div
-          className={classNames('social-drive-action', {
-            'lg:w-2/3 lg:pr-3': !hidePageViews,
-          })}
-        >
+        <div className="social-drive-action lg:w-2/3 lg:pr-3">
```
mendelB: The issue here is that we don't want this limited to two-thirds width on the [refer-friends](https://dosomething.org/us/refer-friends) page re this [Pivotal story](https://www.pivotaltracker.com/n/projects/2401401/stories/169826562)

mendelB: It's such an awkward balancing act of supporting this on pages we're we _want_ it to expand full-width _without_ the page views info bar, but want the exact opposite on campaign pages 😖

DFurnes: Ahhhh, that's frustrating.

weerd: Now I'm confused... why was the original approach no longer working? Is it because loading via GraphQL now we don't know if the block has the `hidePageViews` set to `true` or `false`?

DFurnes: Yep, exactly. Before [we could reach into each block](https://github.com/DoSomething/phoenix-next/blob/e01788be7f4ae90ff29db2a91d0485c81c44fb5e/resources/assets/components/pages/CampaignPage/CampaignPageContent.js#L42-L50) to inspect whether it needed to be full-width or not. Since we now query blocks via GraphQL, they aren't loaded by the time the campaign page is making its layout decisions.

weerd: Is there other info we know about the page? Maybe we should add a class along with the `general-page` one to be `referral-page`... or maybe we just check against the `campaign-page` class to give us context what kind of page it's sitting in? 🤔

DFurnes: I'd prefer not to introduce implicit dependencies like that (since there's no way to tell at a glance that the component would render differently in different contexts).

DFurnes: I added an optional `fullWidth` prop to the Social Drive Block in ea89fc5 which allows us to force the "main" block to fill the width of the container.

resources/assets/components/pages/ReferralPage/Alpha/AlphaPage.js
```diff
@@ -25,6 +25,7 @@ const AlphaPage = ({ userId }) => (
             shareCardTitle="Refer A Friend"
             link={`${PHOENIX_URL}/us/join?user_id=${userId}&campaign_id=${getReferralCampaignId()}`}
             hidePageViews
+            fullWidth
```
weerd: Ah, nice 👌

weerd: Oh wait... is there a possibility to just pass in the width related classes via `className` prop from here, instead of having a unique prop just to decide it?

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
DFurnes: From what I can tell, this works well in both cases now (thank you for catching that @mendelB!). It's a little late for a Friday deploy, but we can get this out first thing Monday.

resources/assets/components/pages/ReferralPage/Alpha/AlphaPage.js
DFurnes: Mmm, that'd be nice. It might be a little awkward since this is on one of two sibling child elements (and I think usually if we have a `className` prop, it applies to the parent element).

weerd: Ah, ok. Yeah, I didn't delve into the component so I wasn't 100% sure.

PR #1917, opened 2020-02-10
Test Fix - stubbed PostGalleryBlock to ensure reactions enabled

mendelB:
### What's this PR do?

This pull request should hopefully fix the buggy `campaign-gallery` Cypress test we saw failing a lot, where we'd fail to find the `.reaction` button.

It mocks the `PostGalleryBlock` so that the `hideReactions` field is always `false`. I _believe_ this was sporadically failing because this would be auto-stubbed to `true` leading the `.reaction` button to obviously not display!

👀 
If we want to test galleries with hidden reactions, I guess we can just specifically mock this `ContentfulBlockQuery` within the test?

### Relevant tickets
https://dosomething.slack.com/archives/C3ASB4204/p1581008569015400


DFurnes:
> If we want to test galleries with hidden reactions, I guess we can just specifically mock this `ContentfulBlockQuery` within the test?

Yep, I think that's exactly what we'd want to do!

PR #1918, opened 2020-02-10
Remove unused PHP "block" entitites.

DFurnes:
### What's this PR do?

This pull request removes logic used to load blocks via our PHP Content API, since these are all being fetched via GraphQL now. This allows us to use the same mental model everywhere we embed blocks into the site (and eventually for data fetching across the board)!

🔥

🍃

### Relevant tickets

References [Pivotal #169216496](https://www.pivotaltracker.com/story/show/169216496).

### Checklist


app/Entities/Entity.php
```diff
@@ -60,74 +60,26 @@ public function parseBlock($block)
...
-                return new SoftEdgeWidgetAction($block->entry);
             case 'storyPage':
                 return new TruncatedStoryPage($block->entry);
-            case 'textSubmissionAction':
-                return new TextSubmissionAction($block->entry);
-            case 'voterRegistrationAction':
-                return new VoterRegistrationAction($block->entry);
             default:
-                return $block->entry;
+                return ['id' => $block->entry->getId()];
```
weerd: Why'd this get switched to just return the ID?

DFurnes: We only need to return the IDs here, since these are just used as hooks for querying the relevant content from GraphQL. Swapping this saves us from sending a bunch of unused information over the wire.

weerd: Were there any default entries that would otherwise hit this `default` condition in the past that aren't handled via GraphQL? I'm guessing not, but just saying it out loud 🗣 to make sure :)

DFurnes: I don't believe so, but it's certainly worth some additional testing to be sure!

PR #1921, opened 2020-02-10
Analytics event docs

weerd:
### What's this PR do?

This pull request adds documentation for the new event analytics approach in Phoenix to make it easier for developers on the team to implement event triggers moving forward!

Like this, with maybe a nice mulled wine by the fireplace:
image

### Relevant tickets

### Checklist


docs/data-and-performance/analytics.md
```diff
@@ -0,0 +1,73 @@
...
+Import the function, along with the `EVENT_CATEGORIES` object, which contains a list of approved event categories that can be used to designate a category for the event.
+
+```js
+import { EVENT_CATEGORIES, trackAnalyticsEvent } from '../../helpers/analytics';
+
+// Within React component or other JS file:
+trackAnalyticsEvent('clicked_signup', {
+  action: 'button_clicked',
+  category: EVENT_CATEGORIES.signup,
+  label: campaignTitle,
```
mendelB: It's funny, but this particular event diverges slightly from the described function of the Label above, where instead of describing the action, it merely provides the campaign title context for where the signup took place. (e.g. intuitively based on that explanation, I'd expect the `label` to be `signup`, but in this case we glean that from the category I guess?)

docs/data-and-performance/analytics.md
```diff
@@ -0,0 +1,73 @@
...
+- `context`
+- `service` _(optional)_
+
+{% hint style="info" %}
+If no `context` information is specified in the ticket, please confirm with the ticket requester that the event trigger does not require any additional context data supplied.
+
+We also recommend reviewing the location where the event trigger is supposed to fire, and considering if there is any potential context data that is available to supply to the `context` object and mention it in the ticket. It could reveal that there is useful information that could help paint a full picture of the event that would be ideal to add to the `context` for analytics.
+{% endhint %}
+
+After adding an event trigger, you can test it in any of our environments by using the **DS Debug** console tool, which can be called in the browser console using the following command:
```
mendelB: It'd be neat to have separate documentation from the Debug service that we can link to from here! (But lol add that to the list 😅)

docs/data-and-performance/analytics.md
```diff
@@ -0,0 +1,73 @@
+# Analytics
+
+The structure of an analytics event is comprised of a specified event name and an object providing additional event metadata. The metadata is based on the data structure that our third-party analytics platforms require, which includes the following properties:
+
+- **category**: referenced from a list of approved categories (it should not be created on the fly).
+- **action**: defines a specified item/element and describes the action that was taken upon it (e.g. "Button Clicked").
+- **label**: additional meta data about the action to help distinguish what it was (e.g. "Register", "Twitter" or "Email", to describe the "Button Clicked" action).
```
mendelB: See my comment below re the label functionality on the signup click example. Maybe worth calling out that this will sometimes be essentially the metadata for which page/campaign the event took place? (Unrelated to this doc, but In hindsight, I wonder if it was a mistake for us to pull that info into the label, as that can be gleaned from the URL anyways... )

docs/data-and-performance/analytics.md
weerd: 😅

docs/data-and-performance/analytics.md
weerd: Ya... I think maybe this event is just not ideal 🤔 I think the description stands, and the `campaignTitle` should be in the `context` but this was maybe a specific request to be able to see which campaign easily in GA 🤷‍♂

Maybe I can add a note that there are sometimes unique situations?

I'll switch out the example so it's more straightforward with the descriptions.

docs/data-and-performance/analytics.md
weerd: Alright @mendelB I updated the examples to all use the `clicked_share_facebook` event so it's all consistent and also to avoid the unique scenario (that we should change) with the signup button click!

PR #1925, opened 2020-02-12
Paginated Campaigns Gallery

mendelB:
### What's this PR do?

This pull request is a collaboration with @lindsayrmaher which adds a new `PaginatedCampaignGallery` component which...surprise! Displays campaigns in a paginated gallery! 

We heavily modeled this after the [`CampaignsTable`](https://github.com/DoSomething/rogue/blob/7aca5b68c993da35c260df987aea2b28d1fdeef9/resources/assets/components/CampaignsTable.js) in Rogue which provides similar functionality from a query / pagination perspective.

- How's the component naming? 
- Do you agree that this belongs in `/utilities`?
- We opted not to add a general 'PaginatedQuery' component, but that can be abstracted away from this if necessary in the future to model the current [`PaginatedQuery`](https://github.com/DoSomething/phoenix-next/blob/3f01a957318661e4a213876594b2849f53183c28/resources/assets/components/PaginatedQuery.js) which uses the 'page' pagination format.
- We're rendering a wrapping div for error, loading, and the main case to allow e.g. grid formatting. Should we clean this up and have one rendering div which accepts the diverse set of children?
- We hardcoded the `hasWebsite` filter for this query, since this is presumably the only use case for this gallery!

We'll be using this for two impending features: 
1. Auto pulling cause-related campaigns into cause pages ([Pivotal ID#170606523](https://www.pivotaltracker.com/n/projects/2401401/stories/170606523))
2. The new and improved Explore Campaigns page #1907 

I'll add a Cypress test for this once we add this to the cause-page, to test it in that context! (And documentation for this utility once this is approved!)

https://github.com/DoSomething/phoenix-next/pull/1925/commits/538603582eff33e61e039d038507c07593468167 Also as a bonus re Leah's request to alighn the gallery items 'article' whitespace to align with the rest of the items on the row cc @lkpttn 

Before
image

After
image

### Relevant tickets

References [Pivotal #170606523](https://www.pivotaltracker.com/n/projects/2401401/stories/170606523).
References [Pivotal #170823582](https://www.pivotaltracker.com/story/show/170823582).

Kapture 2020-02-12 at 14 54 43

### Checklist

### DIFF
--- a/resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
+++ b/resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
@@ -1,0 +1,135 @@
+import React from 'react';
+import gql from 'graphql-tag';
+import PropTypes from 'prop-types';
+import { useQuery } from '@apollo/react-hooks';
+import { get, isArray, mergeWith } from 'lodash';
+
+import Button from '../Button/Button';
+import Spinner from '../../artifacts/Spinner/Spinner';
+import ErrorBlock from '../../blocks/ErrorBlock/ErrorBlock';
+import GalleryBlock from '../../blocks/GalleryBlock/GalleryBlock';
+
+const PAGINATED_CAMPAIGNS_QUERY = gql`
+  query campaignsByCauseQuery(
+    $causes: [String]
+    $cursor: String
+    $first: Int
+    $hasWebsite: Boolean
+    $isOpen: Boolean
+    $orderBy: String
+  ) {
+    campaigns: paginatedCampaigns(
+      after: $cursor
+      causes: $causes
+      first: $first
+      hasWebsite: $hasWebsite
+      isOpen: $isOpen
+      orderBy: $orderBy
+    ) {
+      edges {
+        cursor
+        node {
+          id
+          campaignWebsite {
+            id
+            slug
+            showcaseTitle
+            showcaseDescription
+            showcaseImage {
+              url
+              description
+            }
+          }
+        }
+      }
+      pageInfo {
+        endCursor
+        hasNextPage
+      }
+    }
+  }
+`;
+
+const PaginatedCampaignGallery = ({ className, itemsPerRow, variables }) => {
+  const { error, loading, data, fetchMore } = useQuery(
+    PAGINATED_CAMPAIGNS_QUERY,
+    {
+      variables: { ...variables, hasWebsite: true },
+      notifyOnNetworkStatusChange: true,
+    },
+  );
+
+  const { endCursor, hasNextPage } = get(data, 'campaigns.pageInfo', {});
+
+  const handleViewMore = () => {
+    fetchMore({
+      variables: { cursor: endCursor },
+      updateQuery: (previous, { fetchMoreResult }) =>
+        mergeWith({}, previous, fetchMoreResult, (dest, src) =>
+          // By default, Lodash's `merge` would try to merge *each* array
+          // item (e.g. `edges[0]` with then next page's `edges[0]`).
+          isArray(dest) ? [...dest, ...src] : undefined,
+        ),
+    });
+  };
+
+  // Parse out the nested campaign website nodes so we can pass them to the Gallery Block.
+  const campaigns = get(data, 'campaigns.edges', []).map(edge =>
+    get(edge, 'node.campaignWebsite'),
+  );
+
+  if (error) {
+    return (
+      <div className={className}>
+        <ErrorBlock error={error} />
+      </div>
+    );
+  }
+
+  if (loading && !data) {
+    return (
+      <div className={className}>
+        <Spinner className="flex justify-center" />
+      </div>
+    );
+  }
+
+  return (
+    <div className={className}>
+      <GalleryBlock
+        blocks={campaigns}
+        itemsPerRow={itemsPerRow}
+        imageAlignment="TOP"
+      />
+      {hasNextPage ? (
+        <div className="p-6 text-center">
+          <Button
+            className="-tertiary"
+            onClick={handleViewMore}
+            disabled={loading}
+          >
+            view more...
+          </Button>
+        </div>
+      ) : null}
+    </div>
+  );
+};
+
+PaginatedCampaignGallery.propTypes = {
+  className: PropTypes.string,
+  itemsPerRow: PropTypes.oneOf([2, 3, 4, 5]).isRequired,
+  variables: PropTypes.shape({
+    causes: PropTypes.arrayOf(PropTypes.string),
+    isOpen: PropTypes.bool,
+    orderBy: PropTypes.string,
+    first: PropTypes.number,
+  }),
+};
+
+PaginatedCampaignGallery.defaultProps = {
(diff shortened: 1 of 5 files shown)


resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
```diff
@@ -0,0 +1,135 @@
...
+
+  const { endCursor, hasNextPage } = get(data, 'campaigns.pageInfo', {});
+
+  const handleViewMore = () => {
+    fetchMore({
+      variables: { cursor: endCursor },
+      updateQuery: (previous, { fetchMoreResult }) =>
+        mergeWith({}, previous, fetchMoreResult, (dest, src) =>
+          // By default, Lodash's `merge` would try to merge *each* array
+          // item (e.g. `edges[0]` with then next page's `edges[0]`).
```
katiecrane: Is this the reasoning behind using `mergeWith`?

katiecrane:
> How's the component naming?

I love it!

> Do you agree that this belongs in /utilities?

Yes!
    
> We opted not to add a general 'PaginatedQuery' component, but that can be abstracted away from this if necessary in the future to model the current PaginatedQuery which uses the 'page' pagination format.

Was that component not used here because we're using cursor pagination?

> We're rendering a wrapping div for error, loading, and the main case to allow e.g. grid formatting. Should we clean this up and have one rendering div which accepts the diverse set of children?

If we have on rendering div can we still allow the grid formatting? How do the 3 separate divs make grid formatting easier?

> We hardcoded the hasWebsite filter for this query, since this is presumably the only use case for this gallery!

I think this is smart, we definitely don't want to try to display campaigns that have nothing to display!

resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
```diff
@@ -0,0 +1,135 @@
...
+import { useQuery } from '@apollo/react-hooks';
+import { get, isArray, mergeWith } from 'lodash';
+
+import Button from '../Button/Button';
+import Spinner from '../../artifacts/Spinner/Spinner';
+import ErrorBlock from '../../blocks/ErrorBlock/ErrorBlock';
+import GalleryBlock from '../../blocks/GalleryBlock/GalleryBlock';
+
+const PAGINATED_CAMPAIGNS_QUERY = gql`
+  query campaignsByCauseQuery(
```
DFurnes: I think we've been using `TitleCase` for query names (not a big deal, but nice to keep consistent).

resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
mendelB: Indeed! @DFurnes gave a nice explanation over in Rogue (which we're totally just ripping off here) https://github.com/DoSomething/rogue/pull/927#discussion_r342165648

resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
```diff
@@ -0,0 +1,135 @@
...
+    $first: Int
+    $hasWebsite: Boolean
+    $isOpen: Boolean
+    $orderBy: String
+  ) {
+    campaigns: paginatedCampaigns(
+      after: $cursor
+      causes: $causes
+      first: $first
+      hasWebsite: $hasWebsite
```
DFurnes: You can just hard-code this as `hasWebsite: true` (since as you said in the PR, this is likely constant).

resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
```diff
@@ -0,0 +1,135 @@
...
+      notifyOnNetworkStatusChange: true,
+    },
+  );
+
+  const { endCursor, hasNextPage } = get(data, 'campaigns.pageInfo', {});
+
+  const handleViewMore = () => {
+    fetchMore({
+      variables: { cursor: endCursor },
+      updateQuery: (previous, { fetchMoreResult }) =>
```
DFurnes: Let's pull this out into a helper like we do in Rogue, since it's be useful in any Relay-style paginated component!

resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
mendelB: Smart! Done in https://github.com/DoSomething/phoenix-next/pull/1925/commits/157b528ce1ecebad86a425ff7850135795a97f26

resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
mendelB: ahhh consistency, who needs it. (done in https://github.com/DoSomething/phoenix-next/pull/1925/commits/35797c491c34664432ef335a30c4b638839cd0ed!)

resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
mendelB: Good call! done in https://github.com/DoSomething/phoenix-next/pull/1925/commits/9b1734aea5915fa581b1ab67f34e28eccfcd9995

mendelB:
@katiecrane thanks for the blow-by-blow! 🙌 

> Was that component not used here because we're using cursor pagination?

Exactly 👌 this is our first implementation in Phoenix, and we realized we'd probably need this cursor-paginated-campaigns gallery as its own component, so it made sense to package up!

> If we have on rendering div can we still allow the grid formatting? How do the 3 separate divs make grid formatting easier?

Yeah, I was thinking as an alternative we'd define the rendering child in a variable which would either be the error block or the spinner or the gallery, and then render it out within the containing `div` which has the assigned `className` instead of doing it three times separately?

katiecrane:
> Yeah, I was thinking as an alternative we'd define the rendering child in a variable which would either be the error block or the spinner or the gallery, and then render it out within the containing div which has the assigned className instead of doing it three times separately?

Hmm I don't have a strong opinion on this piece - I think the code reads fine the way it is but sounds like it could be _slightly_ DRY-er, but no huge pieces are being repeated as is 🤷‍♀

mendelB:
> > Yeah, I was thinking as an alternative we'd define the rendering child in a variable which would either be the error block or the spinner or the gallery, and then render it out within the containing div which has the assigned className instead of doing it three times separately?
> 
> Hmm I don't have a strong opinion on this piece - I think the code reads fine the way it is but sounds like it could be _slightly_ DRY-er, but no huge pieces are being repeated as is 🤷‍♀

Gotcha. In that case, my vote would be to keep as-is since Lindsay and I have features depending on this. And we can always refactor if necessary! Does that sound good?

katiecrane:
@mendelB Definitely!

PR #1937, opened 2020-02-20
Remove usage of Lorempixel in tests to fix timeouts.

DFurnes:
### What's this PR do?

This pull request swaps this usage of `faker.image.imageUrl` (which gives a random [lorempixel](http://lorempixel.com) URL) to a static image on our local web server. When this third-party service was down (like it is now), it could cause any tests that include an Embed component to time out.

Tests should pass without timing out!

This seems to be a [known recurring issue](https://git.io/JvRvW) with this service.

### Relevant tickets

References [Pivotal #171185760](https://www.pivotaltracker.com/story/show/171185760).

### Checklist

### DIFF
--- a/cypress/fixtures/graphql.js
+++ b/cypress/fixtures/graphql.js
@@ -51,7 +51,7 @@ export const operations = {
       title: faker.lorem.words(),
       description: faker.lorem.sentence(),
       providerName: faker.company.companyName(),
-      thumbnailUrl: faker.image.imageUrl(),
+      thumbnailUrl: '/apple-touch-icon.png',
       html: null,
     },
   },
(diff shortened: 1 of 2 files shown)


cypress/fixtures/graphql.js
```diff
@@ -51,7 +51,7 @@ export const operations = {
       title: faker.lorem.words(),
       description: faker.lorem.sentence(),
       providerName: faker.company.companyName(),
-      thumbnailUrl: faker.image.imageUrl(),
+      thumbnailUrl: '/apple-touch-icon.png',
```
DFurnes: We could also consider using `faker.image.dataUri` like we do for posts, but this seems to work decently enough.

weerd: Maybe we eventually add a series of images for testing in the `cypress` directory?

DFurnes: Mmm, that could be handy. We'd have to put them in `public/` so they're in the document root for the web server.

weerd:
Removing the `userFactory()` calls in the beta-referral-page tests was also intentional?

DFurnes:
Ah sorry, didn't call that out in the PR description. Yes, that variable was unused.

PR #1939, opened 2020-02-20
Fix issue rendering campaign's `/modal/:id` route.

DFurnes:
### What's this PR do?

This pull request fixes [an issue](https://www.pivotaltracker.com/story/show/171369910) rendering the campaign's `/modal/:id` route.

Screen Shot 2020-02-20 at 1 29 55 PM

Here's how this looks on [medium devices](https://user-images.githubusercontent.com/583202/74966785-c40ca280-53e5-11ea-9d25-9010fc43e921.png) and [small devices](https://user-images.githubusercontent.com/583202/74966795-c8d15680-53e5-11ea-8359-9bc8fca47f49.png).

👀

I'd like to give more thought to how we handle this in the future (such as what to do about things like `ContentBlock` that have transparent backgrounds or uneven sizes), but this fixes up the immediate problem!

### Relevant tickets

References [Pivotal #171369910](https://www.pivotaltracker.com/story/show/171369910).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/ModalRoute/ModalRoute.js
+++ b/resources/assets/components/utilities/ModalRoute/ModalRoute.js
@@ -7,7 +7,7 @@ import { StaticRouter } from 'react-router';
 import ReactRouterPropTypes from 'react-router-prop-types';
 
 import Modal from '../Modal/Modal';
-import ContentfulEntryContainer from '../../ContentfulEntry/ContentfulEntryContainer';
+import ContentfulEntryLoader from '../ContentfulEntryLoader/ContentfulEntryLoader';
 
 // Helpers:
 const isModal = location => location.pathname.includes('/modal/');
@@ -47,7 +47,9 @@ class ModalRoute extends React.Component {
           path={`${match.url}/modal/:id`}
           render={routingProps => (
             <Modal onClose={() => history.push(this.previousLocation)}>
-              <ContentfulEntryContainer id={routingProps.match.params.id} />
+              <ContentfulEntryLoader
+                id={routingProps.match.params.id}
+              />
             </Modal>
           )}
         />
--- a/resources/assets/components/utilities/ModalRoute/ModalRoute.js
+++ b/resources/assets/components/utilities/ModalRoute/ModalRoute.js
@@ -49,6 +49,7 @@ class ModalRoute extends React.Component {
             <Modal onClose={() => history.push(this.previousLocation)}>
               <ContentfulEntryLoader
                 id={routingProps.match.params.id}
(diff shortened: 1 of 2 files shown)


resources/assets/components/utilities/ModalRoute/ModalRoute.js
```diff
@@ -47,7 +47,10 @@ class ModalRoute extends React.Component {
           path={`${match.url}/modal/:id`}
           render={routingProps => (
             <Modal onClose={() => history.push(this.previousLocation)}>
-              <ContentfulEntryContainer id={routingProps.match.params.id} />
+              <ContentfulEntryLoader
+                id={routingProps.match.params.id}
+                classNameByEntry={{ ContentBlock: 'bg-white rounded p-6' }}
```
lindsayrmaher: I've seen this in a few places (the `classNameByEntry`) and I just wanted some clarity on how it works exactly. And also why we made the switch from the Container to the Loader here

DFurnes: This lets us conditionally apply class names based on the type of block that's returned by the GraphQL query. So in this case, we'd only apply `bg-white rounded p-6` if it's a ContentBlock (which otherwise has a transparent background that [doesn't quite cut it](https://user-images.githubusercontent.com/583202/74971009-5d8b8280-53ed-11ea-9537-f0036d74a6aa.png)). Elsewhere, we use it to do things like customize the width of different blocks on the page (so some are full-width and others are narrow).

We switched from `ContentfulEntryContainer` to `ContentfulEntryLoader` here so that we can load this based on the provided `id` prop (whereas `ContentfulEntryContainer` only has a `json` prop for already loaded responses). So just a straight-up typo before!

weerd: @lindsayrmaher here's a good example in use on the `CampaignPageContent.js` that further expands on Dave's explanation:

```jsx
<div className="base-12-grid clear-both">
  {blocks.map(block => (
    <ContentfulEntryLoader
      key={block.id}
      id={block.id}
      className="mb-6 clear-both"
      classNameByEntryDefault="grid-wide-7/10"
      classNameByEntry={{
        ContentBlock: 'grid-wide',
        ImagesBlock: 'grid-wide',
        PostGalleryBlock: 'grid-wide',
        PhotoSubmissionBlock: 'grid-wide',
        SocialDriveBlock: 'grid-wide',
      }}
    />
  ))}
</div>
```
Where when it loads a Contentful Entry, for all block types it will by default add a `grid-wide-7/10` class to outputted component. BUT for some specific types of entries, they get a specific class instead of using the default.

PR #1941, opened 2020-02-24
Homepage content type updates

weerd:
### What's this PR do?

This pull request updates the Contentful `HomePage` content type to include two new reference fields; one for campaigns and one for articles. This is for the upcoming Home Page updates.

Instead of repurposing the `blocks` fields, we felt it would be better to just add the two new fields so that the HomePage can still behave as expected for the old design and easily use feature flags to handle the new design and the supplied content in these new fields.

### Relevant tickets

References [Pivotal #171056257](https://www.pivotaltracker.com/story/show/171056257).

### Checklist

### DIFF
--- a/contentful/content-types/homePage.js
+++ b/contentful/content-types/homePage.js
@@ -4,48 +4,122 @@ module.exports = function(migration) {
     .name('Home Page')
     .description('The Home Page for DoSomething.org.')
     .displayField('internalTitle');
-
   homePage
     .createField('internalTitle')
     .name('Internal Title')
     .type('Symbol')
+    .localized(false)
     .required(true)
-    .localized(false);
-
+    .validations([])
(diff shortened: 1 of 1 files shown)


contentful/content-types/homePage.js
```diff
@@ -4,48 +4,125 @@ module.exports = function(migration) {
...
+    .required(false)
+    .validations([])
+    .disabled(false)
+    .omitted(false)
+    .items({
+      type: 'Link',
+
+      validations: [
+        {
+          linkContentType: ['page'],
```
mendelB: Just confirming 🙅‍♂we don't want company page articles linkable from the homepage?

weerd: I don't believe so. If it turns out we do we can always update the validations.

PR #1943, opened 2020-02-24
Add Filters (Causes filter first)

lindsayrmaher:
### What's this PR do?

This pull request adds the cause filter button to our explore campaigns page and the responsive menu that allows users to select the causes to select by. It's missing a some styling polish and functionality (unchecking when they clear their cause choices etc). 

Does the file structure seem reasonable?
Could it be DRYer?
👀 

This will be our MVP for the filters feature, and the basis for future filters in upcoming sprints.

Large Screen:
Screen Shot 2020-03-04 at 11 31 06 AM

Screen Shot 2020-03-04 at 11 31 49 AM

Screen Shot 2020-03-04 at 11 31 32 AM

Medium Screen:
Screen Shot 2020-03-04 at 11 33 15 AM

Screen Shot 2020-03-04 at 11 32 10 AM

Small Screen:
Screen Shot 2020-03-04 at 11 32 55 AM

Screen Shot 2020-03-04 at 11 32 29 AM

### Relevant tickets

References [Pivotal #](https://www.pivotaltracker.com/story/show/170824165).

### Checklist


lindsayrmaher:
@weerd This is definitely a hefty PR so if there's anything I can talk through more let me know!!

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
```diff
@@ -0,0 +1,68 @@
...
+        type="button"
+      >
+        <p className="font-bold text-blue-500">clear</p>
+      </button>
+      <Button onClick={handleMenuToggle}>Show Campaigns</Button>
+    </div>
+  </form>
+);
+
+CauseFilter.propTypes = {
```
weerd: You can probably name this something more generic like we discussed 😄 Especially since you're importing the `causes`. You could name this `FiltersForm.js` and instead of importing the `CauseVariables` in this file do it where this component is called and pass the "causes" as an `options` prop: 

```js
<FiltersForm {...fields} options={causes} />
```

weerd: Oh wait, I'm forgetting that before we said not all the filters would be checkboxes 🤔 So maybe you create a component for each potential type of filter? So maybe a `<CheckboxFiltersForm />`? To differentiate between the different types... or do we know if they're all so different that generalizing it wouldn't be worth the effort?

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
```diff
@@ -0,0 +1,68 @@
...
+        name={causeValue}
+        id={causeValue}
+        type="checkbox"
+        value={causeValue}
+        onClick={handleSelect}
+        checked={causeChecked}
+      />
+      <span className="pl-4">{causeName}</span>
+    </label>
+  </>
```
DFurnes: You can get rid of this wrapping fragment, since it's just a single `label` element inside it:

```suggestion
  <label className="flex justify-start pb-2" htmlFor={causeName}>
    <input
      name={causeValue}
      id={causeValue}
      type="checkbox"
      value={causeValue}
      onClick={handleSelect}
      checked={causeChecked}
    />
    <span className="pl-4">{causeName}</span>
  </label>
```

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
lindsayrmaher: From the looks of things theres either `checkboxes`, `radio buttons` or `toggles` !

weerd: Ok! Your call, if it seems like those are the lone 3 types and don't end up getting combined maybe they could be generalized into those 3 components `<CheckboxFiltersForm />`, `<RadioFiltersForm />` and `<ToggleFiltersForm />`. But also could be a later thing as the other filters reveal themselves!

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -0,0 +1,106 @@
...
+    staticCauses[event.target.value].checked = !staticCauses[event.target.value]
+      .checked;
+    if (causes.includes(event.target.value)) {
+      const newCauses = causes.filter(cause => {
+        return cause !== event.target.value;
+      });
+      setCauses([...newCauses]);
+    } else {
+      setCauses([...causes, event.target.value]);
+    }
```
DFurnes: I think it might be cleaner (and easier) to only store the "machine names" for these causes in state & then just reference the labels from a dictionary... that way the only things in state are actual changeable states:

Thinking something like this:

```js
// from `./CauseVariables`;
const causeLabels = { 'animal-welfare': 'Animal Welfare', bullying: 'Bullying', /* ... */};

// here, in FilterNavigation.js:
const [causes, setCauses] = useState([]);

const handleCauseSelect = event => setCauses(xor(causes, [event.target.value]));
const clearAllSelected = () => setCauses([]);

// and in CauseFilter.js:
<span className="pl-4">{causeLabels[causeValue]}</span>

```

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterSubNav.js
```diff
@@ -0,0 +1,46 @@
...
+import { withoutNulls } from '../../../../helpers';
+import CauseFilter from '../CampaignFilters/CauseFilter/CauseFilter';
+
+const isVisible = css`
+  height: 1px;
+  width: 1px;
+  overflow: hidden;
+  clip: rect(1px, 1px, 1px, 1px);
+  white-space: nowrap; /* added line */
+`;
```
DFurnes: Should this be a helper class in `base.scss`?

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import causes from './CauseVariables';
+import Button from '../../../../utilities/Button/Button';
+
+const CauseInputs = ({ causeChecked, causeName, causeValue, handleSelect }) => (
```
weerd: Nit-pick. Make this singular (`CauseInput`) since only outputting a single input.

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterSubNav.js
```diff
@@ -0,0 +1,46 @@
...
+    case 'Causes':
+      return <CauseFilter {...fields} />;
+    default:
+      return null;
+  }
+};
+
+const FilterSubNav = props => (
+  <div
+    css={!props.showFilterMenu ? isVisible : null}
```
DFurnes: Actually, follow-up question – would this be better handled with [`aria-expanded`](https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/expanded/) & traditionally hidden (e.g. `display:none`) content? (I'm thinking that we don't want a screen reader to have to read through all these filters unless a user wants them, right?)

resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js
```diff
@@ -1,22 +1,35 @@
...
+
+const CampaignsIndexPage = () => {
+  const [causes, setCauses] = useState([]);
+
+  return (
+    <>
+      <SiteNavigationContainer />
+      <main className="campaigns-page md:w-3/4 mx-auto">
+        <h1 className="w-full my-6 pl-6 md:pl-0">Explore Campaigns</h1>
+        <FilterNavigation causes={causes} setCauses={setCauses} />
```
weerd: I wonder if given how each type of filter is kind of its own little separate filter navigation subnav, as more are added it could complicate things... `causes` would need to be changed to `filters` and the passed an array of filter types `{[causes, time]}` etc. Or what if Filter Navigation is just a wrapper and each filter subnav is passed as a child?

```js
<FilterNavigation>
  <CausesFilter />
  <TimeFilter />
</FilterNavigation>
```
Would something like this work at all? Could simplify popping in more filters in the future 🤔 I think we _may_ have discussed something like this IRL but don't remember well 😬

resources/assets/components/pages/CampaignsPage/campaigns-page.scss
```diff
@@ -0,0 +1,9 @@
+@import '../../../scss/next-toolbox';
+
+.campaigns-page {
+  .cause-filter {
+    @include media($largest) {
+      max-height: 250px;
```
weerd: Is this `max-height` necessary? Not keen on stuff like this having a set height or height it can't go past...

lindsayrmaher: Yeah it doesn't feel great, but it was the cleanest way i could think of to get the inputs to wrap correctly into two columns without splitting up the data. Open to other options!

resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js
lindsayrmaher: Yeah! We did talk through something like this. I think it would make sense to perhaps have FilterNav accept children. and then perhaps we could pass the correct props down per filter directly?

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterSubNav.js
lindsayrmaher: Ohh that's a fair point, I'll see if I can simplify things a bit.

resources/assets/components/pages/CampaignsPage/campaigns-page.scss
weerd: Sounds like a job for CSS Grids!

weerd: Actually, you could maybe try this: <https://css-tricks.com/guide-responsive-friendly-css-columns/>

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
lindsayrmaher: Yeah i pulled them out to keep the file as small as possible, but it does make sense to keep them coupled with the `[causes, setCauses]`. The only thing that's tricky is how to best handle updating the `checked` attribute for each input. So i might leave the causes as objects..

DFurnes: For the checkbox, could we do `checked={causes.includes(cause)}`?

lindsayrmaher: OOOOOOOOOOOOOOOOOOO i'll try it. there's a React warning about using the checked field without handling it properly with an `onChange` but that would be so simple and lovely.

resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js
```diff
@@ -1,22 +1,45 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { get } from 'lodash';
```
weerd: nit-pick! switch the ordering, shortest to longest line length!

resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js
```diff
@@ -1,22 +1,45 @@
...
+  return (
+    <>
+      <SiteNavigationContainer />
+
+      <div className="bg-white border-b border-gray-300 border-solid ">
+        <div className="md:w-3/4 mx-auto pt-6 md:pt-10 pl-6 md:pl-0">
+          <h2 className="w-full mt-6">Explore Campaigns</h2>
+        </div>
+        <FilterNavigation filters={filters} setFilters={setFilters} />
+      </div>
```
weerd: Oh I don't think I noticed this before... why is `<main>` sitting outside of the div containing the filters? There should only be 3 elements with everything else inside of them:

- `<SiteNavigation />`
- `<main />`
- `<SiteFooter />`

Mentioned it when I [re-gridified the chrome](https://github.com/DoSomething/phoenix-next/pull/1850) but easy to lose track of 😬 

Whatever the classes that existed before for `<main>` that may not be ideal in this new scenario can change, but the structure of the 3 elements shouldn't!

resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js
```diff
@@ -1,22 +1,45 @@
...
+const CampaignsIndexPage = () => {
+  const [filters, setFilters] = useState({ causes: [] });
+
+  return (
+    <>
+      <SiteNavigationContainer />
+
+      <div className="bg-white border-b border-gray-300 border-solid ">
+        <div className="md:w-3/4 mx-auto pt-6 md:pt-10 pl-6 md:pl-0">
+          <h2 className="w-full mt-6">Explore Campaigns</h2>
```
weerd: Also wondering why this `<h2>` needs to be wrapped in `<div>`s ... and why it became an `<h2>` instead of keeping it an `<h1>`... if it's size changed, that's just a Tailwind class, but should stay an `<h1>` 🤔

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -0,0 +1,58 @@
...
+    } else {
+      setChosenFilter(filterName);
+    }
+    setShowFilterMenu(!showFilterMenu);
+  };
+
+  return (
+    <div className="md:w-full z-50 bg-white">
+      <div className="base-12-grid items-center filter-nav md:w-3/4 mx-auto pl-6 md:pl-0">
+        <h6 className="color-gray-200">FILTERS</h6>
```
weerd: Same with the `<h6>` here. I don't think this needs to be hierarchically that low of a heading, but it may have been selected because `h6` happens to be a smaller heading. Best to select appropriate level of heading based on hierarchy of content and then just style it to be smaller (or larger) if it's too big.

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -0,0 +1,58 @@
...
+    if (chosenFilter) {
+      setChosenFilter('');
+    } else {
+      setChosenFilter(filterName);
+    }
+    setShowFilterMenu(!showFilterMenu);
+  };
+
+  return (
+    <div className="md:w-full z-50 bg-white">
```
weerd: The filter menu is no longer absolutely positioned right? Probs can get rid of `z-50`.

weerd:
Would it be possible to add the Feature Flag to the app.json file so when checking the review app we can see the new Causes filter, etc?

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -0,0 +1,58 @@
...
+    } else {
+      setChosenFilter(filterName);
+    }
+    setShowFilterMenu(!showFilterMenu);
+  };
+
+  return (
+    <div className="md:w-full bg-white">
+      <div className="flex items-center filter-nav md:w-3/4 mx-auto pl-6 md:pl-0 pt-2 md:pt-0 pb-6">
+        <h1 className="color-gray-200 pr-4 text-base">FILTERS</h1>
```
weerd: I think an `h2` makes sense here since `Explore Campaigns` is already the `h1` for the page/section (there's really only one section on the page). An easy way to see how the hierarchy looks is to use a tool like the [*Web Developer*](https://chrome.google.com/webstore/detail/web-developer/bfbameneiokkgbdmiekhjnmfkcnldhhm?hl=en-US) extension and just click to disable all the CSS, then you can get an idea of page hierarchy...

image

You can now see that "Explore Campaigns" and "Filters" are at the same level... also "FILTERS" shouldn't be all caps, you can just use the `uppercase` Tailwind class for that.

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -0,0 +1,58 @@
...
+    if (chosenFilter) {
+      setChosenFilter('');
+    } else {
+      setChosenFilter(filterName);
+    }
+    setShowFilterMenu(!showFilterMenu);
+  };
+
+  return (
+    <div className="md:w-full bg-white">
```
weerd: You don't need the second div after this one. Just combine this outer one with the one below, and update the classes (some margins may need to be switched to padding) so it's just:

```html
<div>
  <h1>Explore Campaigns</h1>
  <div>
    // all the filter stuff
  </div>
</div>
```

resources/assets/components/utilities/MenuButton/MenuButton.js
```diff
@@ -0,0 +1,58 @@
...
+}) => (
+  <button
+    onClick={() => onClick(title)}
+    type="button"
+    className={classnames(
+      'items-center px-2 py-2 border-solid border border-black-600 rounded-lg focus:outline-none',
+      className,
+    )}
+  >
+    <p className={classnames('font-bold', titleColor)}>{title}</p>
```
weerd: Eep! Not ideal to use `<p>` inside the button. If it's just cause you need a hook to add classes, then use a `<span>`. That being said, these seem like classes you can just apply to the `<button>` directly (along with the others you're adding), no need to wrap the text just to add classes to it unless you need to override.

resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js
```diff
@@ -1,22 +1,46 @@
...
+const CampaignsIndexPage = () => {
+  const [filters, setFilters] = useState({ causes: [] });
+
+  return (
+    <>
+      <SiteNavigationContainer />
+
+      <main>
+        <div className="bg-white border-b border-gray-300 border-solid mb-6">
+          <div className="md:w-3/4 mx-auto pt-6 md:pt-10 pl-6 md:pl-0">
```
weerd: You don't need the second div after this one. Just combine this outer one with the one below, and update the classes (some margins may need to be switched to padding) so it's just:

```html
<div>
  <h1>Explore Campaigns</h1>
  <div>
    // all the filter stuff
  </div>
</div>
```

resources/assets/components/utilities/MenuButton/MenuButton.js
weerd: You should also use the [`btn` component class](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/scss/base.scss#L18-L30) that I mentioned before for buttons. While it defaults to the lighter blue buttons of the old branding, you can add classes to override to use the blurple (`btn bg-blurple-500`, etc) as well as for focus and hover :)

resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js
```diff
@@ -1,22 +1,46 @@
...
+const CampaignsIndexPage = () => {
+  const [filters, setFilters] = useState({ causes: [] });
+
+  return (
+    <>
+      <SiteNavigationContainer />
+
+      <main>
+        <div className="bg-white border-b border-gray-300 border-solid mb-6 pt-6 md:pt-10">
+          <h1 className="md:w-3/4 mx-auto w-full mt-6 md:mb-2 text-xl pl-6 md:pl-0">
```
weerd: Small nit-pick. You can probably get rid of `w-full`. Heading tags are block level by default and expand full width unless you style them specifically to not do so!

resources/assets/components/utilities/MenuButton/MenuButton.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import { css } from '@emotion/core';
+
+import MenuCarat from '../../artifacts/MenuCarat/MenuCarat';
```
weerd: Is this `MenuCarat` still needed in this component? It doesn't seem to get used.

resources/assets/components/pages/CampaignsPage/campaigns-page.scss
```diff
@@ -0,0 +1,37 @@
...
+.filter-nav {
+  h1,
+  h2 {
+    color: theme('colors.gray.600');
+  }
+}
+
+.filter-nav-button {
+  color: theme('colors.blurple.500');
+}
```
weerd: Could these not be added as classnames to the respective elements?

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterSubNav.js
```diff
@@ -0,0 +1,37 @@
...
+import classNames from 'classnames';
+
+import { withoutNulls } from '../../../../helpers';
+import CauseFilter from '../CampaignFilters/CauseFilter/CauseFilter';
+
+const renderedFilterMenu = props => {
+  const fields = withoutNulls(props);
+  switch (fields.chosenFilter) {
+    case 'Cause':
+      return <CauseFilter {...fields} />;
```
weerd: I wonder if it would be better to pass `fields.filters` and `fields.setFilters`, instead of spreading all the fields into `CauseFilter` since it'll pass through the rest of them including `className` from `FilterSubNav` which could bite us later if we end up needing to pass in a `className` prop to `CauseFilter` 🤔

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
```diff
@@ -0,0 +1,98 @@
+import React from 'react';
+import { get } from 'lodash';
+import PropTypes from 'prop-types';
+
+const CauseInput = ({ causeChecked, causeName, causeValue, handleSelect }) => (
+  <label className="flex justify-start pb-2" htmlFor={causeName}>
```
weerd: Right now, clicking on the text label for the checkbox, does not toggle the checkbox. But if you change `htmlFor={causeName}` to `htmlFor={causeValue}` (so it matches what's used for the `id` on the `<input />`, it'll allow clicking the text of the label to also toggle the checkbox, which is a much better user experience, especially for small screens! mobile first! ☎️

lindsayrmaher: ooooooo good catch! we love to see a good mobile experience.

resources/assets/components/pages/CampaignsPage/campaigns-page.scss
lindsayrmaher: It wasn't picking up when i directly added them on the elements, there's probably a reason but this felt like an easy way to get around it.

resources/assets/components/utilities/MenuButton/MenuButton.js
lindsayrmaher: 🤔 that is a very fair point! I left it in if we use this button in a different place that needs a carat. There's not a use case i can think of right now so I would be ok with getting rid of it.

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterSubNav.js
lindsayrmaher: Yeah plus one to this!

resources/assets/components/pages/CampaignsPage/campaigns-page.scss
weerd: 🤔 ok, mind just adding a `@TODO` to investigate at a later date :)

resources/assets/components/utilities/MenuButton/MenuButton.js
weerd: Ya, if we decide to use it in the future we can always just add it back in.

resources/assets/components/pages/CampaignsPage/campaigns-page.scss
```diff
@@ -0,0 +1,37 @@
...
+    grid-column-gap: theme('spacing.6');
+  }
+}
+
+.filter-nav {
+  h1,
+  h2 {
+    color: theme('colors.gray.600');
+  }
+}
```
weerd: image

Hmmm, I just tried `text-gray-600` on the `h2` and seems to work fine...  also there's no `h1` in the `.filter-nav` anymore so can at least remove that one. But can probs use the tailwind class for the `h2` and remove this entirely :)

resources/assets/components/pages/CampaignsPage/campaigns-page.scss
weerd: image

Same same for this one. Maybe you were using `font-blurple-500` instead of `text-blurple-500` (I've done this before!), but it seems to work if using the class.

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -0,0 +1,58 @@
...
+      document.getElementById(filterName).focus();
+      setChosenFilter(filterName);
+    }
+    setShowFilterMenu(!showFilterMenu);
+  };
+
+  return (
+    <div className="md:w-full bg-white">
+      <div className="flex items-center filter-nav md:w-3/4 mx-auto pt-2 md:pt-0 pb-6 pl-6 md:pl-0">
+        <h2 className="gray-filter-nav pr-4 text-base uppercase">Filters</h2>
```
weerd: Can probably remove this `.gray-filter-nav` class right?

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -0,0 +1,58 @@
...
+        <MenuButton
+          title="Cause"
+          onClick={handleMenuToggle}
+          className={
+            chosenFilter
+              ? 'bg-blurple-500 color-white border border-solid border-blurple-500'
```
weerd: Ok! I see what the culprit is... `color-white` is not a Tailwind class... it's a vestige of our old utility classes... which is why `color-blurple` probs didn't work for the unselected state... cause it doesn't exist! Switch this to `text-white` and then you can remove this:

```css
.filter-nav-button {
  color: theme('colors.blurple.500');
}
```
from the stylesheet and just use the `text-blurple-500` Tailwind class!

weerd: I'd recommend installing one of the Tailwind VSCode plugins which autocompletes/shows your tailwind config values to choose from to help ensure you're using Tailwind classes based on the project's configuration :)

PR #1944, opened 2020-02-26
Update Campaign Info Bar Help Link

mendelB:
### What's this PR do?

This pull request updates the help link in our `CampaignInfoBar` component to show a link to the Help Center for unauthenticated users (instead of the mailto link to the campaign lead), and the Zendesk Form link for authenticated users (instead of only _signed up_ users.)

UPDATE: We're keeping the mailto link to campaigns leads, but appending a Subject Line with the campaign title so we can better categorize these messages on Zendesk.

We've also added a new default email help address.

https://github.com/DoSomething/phoenix-next/commit/a2b8b5738a436c8f93e6a8dfd6918992b6230d00 updates the function to find the campaign's FAQs page to also look for `.../faq` (without the trailing S) since we have instances of both (sigh). ([See example from Slack](https://dosomething.slack.com/archives/C03T8SDLY/p1581112677007400?thread_ts=1580920284.003800&cid=C03T8SDLY))

### How should this be reviewed?
👁 
How's the new copy?

### Any background context you want to provide?
As per feedback from Hannah for what the best and most helpful flow is! This will ~prevent an influx of avoidable questions while increasing~ increase the flow of categorized questions via our Zendesk API.

### Relevant tickets

References [Pivotal #171202296](https://www.pivotaltracker.com/story/show/171202296).
https://dosomething.slack.com/archives/C03T8SDLY/p1581112677007400?thread_ts=1580920284.003800&cid=C03T8SDLY

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Documentation added for new features/changed endpoints.
- [x] Added screenshots of front-end changes on large screens.
- [x] Added appropriate feature/unit tests.

image
image

PR #1945, opened 2020-02-26
Deprecate Legacy Generated Campaign Lead

mendelB:
### What's this PR do?

This pull request removes some legacy logic we used to generate a Campaign Lead for campaigns pulling from additional content.

There were two campaigns still doing this, and I just updated them to use the new Campaign Lead entry. We also no longer rely on the Campaign necessarily having a referenced campaign lead entry so this is safe to remove.

👀 
Any edge case or context I'm missing?

Bumped into this opportunity in #1944 where I noticed we were hardcoding the campaign lead email for any campaign without a defined campaign lead.

### Relevant tickets

https://github.com/DoSomething/phoenix-next/pull/1944#discussion_r384808882

### DIFF
--- a/app/Entities/Campaign.php
+++ b/app/Entities/Campaign.php
@@ -42,38 +42,18 @@ class Campaign extends Entity implements JsonSerializable
     }
 
     /**
-     * Parse the campaign lead from other
+     * Parse the campaign lead
      *
      * @param  Entry $campaignlead
-     * @param  array $additionalContent
      * @return array
      */
-    public function parseCampaignLead($campaignlead, $additionalContent)
+    public function parseCampaignLead($campaignlead)
     {
         if ($campaignlead) {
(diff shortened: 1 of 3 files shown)


app/Entities/Campaign.php
```diff
@@ -42,38 +42,18 @@ public function parseCampaignDashboard($campaignDashboard)
...
-    public function parseCampaignLead($campaignlead, $additionalContent)
+    public function parseCampaignLead($campaignlead)
     {
         if ($campaignlead) {
             // @TODO (2018-09-13): We should make the CampaignLead field required and thus
             // no longer need a conditional check here.
             return  new Person($campaignlead->entry);
         }
-
-        // @TODO (2018-08-29): we should do away with this additional content item.
```
weerd: Yay! The date-stamped `@TODO` approach to see when we noticed this :D

PR #1946, opened 2020-02-27
Add self-service account deletion page.

DFurnes:
### What's this PR do?

This pull request adds a self-service interface for requesting account deletion, wrapping up the [Self-Service Account Deletion](https://docs.google.com/document/d/1vj_AcfZ_I92dtBH6y83RhSFYJCshQy2wAXhpecAWlWo/edit#) product build. They'll see the following form when hitting "Delete My Account" in the sidebar of their "My Account" page:

Screen Shot 2020-02-27 at 1 05 37 PM

If a user changes their mind before the end of the 14-day "grace period", they can always undo:

Screen Shot 2020-02-27 at 1 05 49 PM

This is pending some additional work in Northstar so that account deletion requests automatically unsubscribe users from all further messaging (since that's a common theme in these requests).

👀

🔥

### Relevant tickets

References [Pivotal #170953839](https://www.pivotaltracker.com/story/show/170953839).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/DeleteAccountTab.js
+++ b/resources/assets/components/pages/AccountPage/DeleteAccountTab.js
@@ -18,6 +18,7 @@ const DeleteAccountMutation = gql`
     requestDeletion(id: $id) {
       id
       deletionRequestedAt
+      emailSubscriptionTopics
     }
   }
 `;
--- a/resources/assets/components/pages/AccountPage/DeleteAccountTab.js
+++ b/resources/assets/components/pages/AccountPage/DeleteAccountTab.js
@@ -13,6 +13,9 @@ const DeleteAccountQuery = gql`
   }
 `;
 
+// We're including 'emailSubscriptionTopics' in this query so that
+// if the user clicks over to their subscriptions tab, this update
+// will be displayed (without a full page refresh).
 const DeleteAccountMutation = gql`
   mutation DeleteAccountMutation($id: String!) {
     requestDeletion(id: $id) {
--- a/resources/assets/components/pages/AccountPage/DeleteAccountTab.js
+++ b/resources/assets/components/pages/AccountPage/DeleteAccountTab.js
@@ -1,0 +1,129 @@
+import React from 'react';
+import gql from 'graphql-tag';
+import classnames from 'classnames';
+import { format } from 'date-fns';
+import { useQuery, useMutation } from '@apollo/react-hooks';
+
+const DeleteAccountQuery = gql`
+  query DeleteAccountQuery($id: String!) {
+    user(id: $id) {
+      id
+      deletionRequestedAt
+    }
+  }
+`;
+
+const DeleteAccountMutation = gql`
+  mutation DeleteAccountMutation($id: String!) {
+    requestDeletion(id: $id) {
+      id
+      deletionRequestedAt
+    }
+  }
+`;
+
+const UndoAccountDeletionMutation = gql`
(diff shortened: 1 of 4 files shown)


resources/assets/scss/base.scss
```diff
@@ -90,6 +90,23 @@ button {
   }
 }
 
+// Utility class for treating buttons like links:
+.link-button {
```
mendelB: Trivial - but maybe worth using `@apply` here to tailwindify these rules?

resources/assets/components/pages/AccountPage/DeleteAccountTab.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+`;
+
+const DeleteAccountMutation = gql`
+  mutation DeleteAccountMutation($id: String!) {
+    requestDeletion(id: $id) {
+      id
+      deletionRequestedAt
+      emailSubscriptionTopics
```
mendelB: I don't think we use this value anywhere?

resources/assets/components/pages/AccountPage/DeleteAccountTab.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+`;
+
+const DeleteAccountForm = () => {
+  const options = { variables: { id: window.AUTH.id } };
+
+  const { data, loading, error } = useQuery(DeleteAccountQuery, options);
+
+  const [undo] = useMutation(UndoAccountDeletionMutation, options);
+  const [deleteUser, { loading: modifying }] = useMutation(
```
mendelB: Oof that deconstructed object alias syntax tripped me up for a sec! Is it just me, or would that read better in the opposite direction? 🤔

resources/assets/components/pages/AccountPage/DeleteAccountTab.js
```diff
@@ -0,0 +1,130 @@
...
+        You&apos;ve been unsubscribed from all emails and text messages, and
+        your account will be permanently deleted in 14 days.
+      </p>
+      <p>
+        We&apos;ll send you one final reminder message before that happens. In
+        the meantime, you can always change your mind:
+      </p>
+      <p>
+        <button type="button" className="link-button" onClick={undo}>
+          I changed my mind! Don&apos;t delete my account.
```
mendelB: 🙏

resources/assets/components/pages/AccountPage/DeleteAccountTab.js
DFurnes: Yeah, it's not great. I had to look up the syntax to remember how to do it. 😖

resources/assets/components/pages/AccountPage/DeleteAccountTab.js
DFurnes: I added this so that we'll update the local cache for this user, so that if the user clicks over to the subscriptions tab, they'll see that they were actually unsubscribed.

(We could also fix this by moving that query into the `EmailSubscriptions` component, rather than in the top-level `AccountQuery` wrapper but this seemed a lil' quicker of a fix!)

DFurnes: Added a comment to clarify this in c5e5c2a5!

resources/assets/scss/base.scss
DFurnes: Good idea! Just did that b6fbb27.

PR #1952, opened 2020-03-02
Zendesk Ticket Grouping

mendelB:
### What's this PR do?

This pull request uses the current campaign's first assigned Cause name to assign a Zendesk group ID to new Zendesk tickets created via our Zendesk API. (Say Zendesk one more time).

You can commit-by-commit this one! Here's some info on the notable commits:

- https://github.com/DoSomething/phoenix-next/commit/82852fd8e01f84ebe55030a28097262e734f439b upgrades to using the Northstar V2 API for fetching the user. I hadn't realized during the first iteration of this that we could. This gives us the `display_name` so we don't have to compute that ourselves, plus forces us to explicitly request the sensitive email field. Neato!
- https://github.com/DoSomething/phoenix-next/commit/8a9e9d61437ba52fb0a9043ceb2102df5a84626d adds a `RogueCampaignRepository` which uses our Gateway `RestApiClient` to fetch campaigns from Rogue by ID so we can parse out the first cause name. I wasn't sure if this was overkill, but it felt like a good approach here. (Should we just use Guzzle or something for this, or is it cool to utilize our own library, though there's no auth or anything?)
- https://github.com/DoSomething/phoenix-next/commit/8332f459afc382352e4bbafcb6ab6de2041a6f10, https://github.com/DoSomething/phoenix-next/commit/0ecf55d1030e08150ce0e627f797705426990733 (update), https://github.com/DoSomething/phoenix-next/commit/0ecf55d1030e08150ce0e627f797705426990733 (patch) is where the fun happens. We add a `campaign_id` param to this request which we use to fetch the campaign from Rogue and parse out its first Cause name. We then ~ask Zendesk for its index of Groups~ search Zendesk for groups fuzzy matching this cause name (though the documentation implies it, [they don't](https://support.zendesk.com/hc/en-us/articles/203663226/comments/204720757) seem to allow _exact_ phrase matching which is frustrating so we have to) then filter out the group with an _exact_ matching name so we can use its ID as our tickets assigned `group_id`. Zendesk doesn't seem to have a `findByName` or some other filter method for the groups API. lame!
- finally, https://github.com/DoSomething/phoenix-next/commit/7d820a9652ba420e7e031f7788df765eb4c889f2 passes through the new `campaign_id` param from our `ZendeskForm` in React land.

#### Questions:
Is the logic to assign the group_id too gnarly? It felt like the best way to do this for now, and since our ticket volume is quite low, the extra non-cached API request to fetch the groups shouldn't be a big deal. I figure worse comes to worst we have two improvement options:
1. Start caching these group requests for a while since we wouldn't anticipate these groups to change much
2. Store a mapping of cause names to group ID's locally, either via config variable or in some JSON file in storage.

#### Refactor!
This controller action is feeling crowded! I'd love to follow up on this PR with some general cleanup. Including, pulling most of this logic into a `ZendeskRepository`!

Check out the Pivotal Story, but this is essentially a simple way for us to start grouping Zendesk Tickets for Agent assignment, which should help preserve Hannah's sanity and declutter the Zendesk inbox!

### Relevant tickets

References [Pivotal #171191450](https://www.pivotaltracker.com/story/show/171191450).

### Checklist

### DIFF
--- a/app/Services/Zendesk.php
+++ b/app/Services/Zendesk.php
@@ -134,8 +134,10 @@ class Zendesk
         $firstCampaignCause = $this->getFirstCampaignCauseName($campaignId);
 
         $zendeskGroups = optional(
-            // Zendesk Search API: http://bit.ly/2TihwCC
-            ZendeskClient::search()->find('type:group name:".'.$firstCampaignCause.'"')
+            remember('zendesk_groups', 15, function () use ($firstCampaignCause) {
+                // Zendesk Search API: http://bit.ly/2TihwCC
+                return ZendeskClient::search()->find('type:group name:".'.$firstCampaignCause.'"');
+            })
         )->results;
 
         // Filter out the *exact* matching group. (The search API includes a more generous matching logic).
--- a/app/Services/Zendesk.php
+++ b/app/Services/Zendesk.php
@@ -64,9 +64,7 @@ class Zendesk
             // ],
         ];
 
-        $zendeskTicket = ZendeskClient::tickets()->create($zendeskTicketData);
-
-        return [$zendeskTicket, $zendeskUser];
+        return ZendeskClient::tickets()->create($zendeskTicketData);
     }
 
     /**
--- a/app/Services/Zendesk.php
+++ b/app/Services/Zendesk.php
@@ -1,0 +1,148 @@
+<?php
+
+namespace App\Services;
+
+use Illuminate\Support\Facades\Log;
+use App\Repositories\RogueCampaignRepository;
+use Huddle\Zendesk\Facades\Zendesk as ZendeskClient;
+
+class Zendesk
+{
+    /**
+     * Rogue Campaign repository instance.
+     * @var RogueCampaignRepository
+     */
+    private $rogueCampaignRepository;
+
+    /**
+     * Zendesk constructor.
+     *
+     * @param RogueCampaignRepository $rogueCampaignRepository
+     */
+    public function __construct(RogueCampaignRepository $rogueCampaignRepository)
+    {
+        $this->rogueCampaignRepository = $rogueCampaignRepository;
+    }
+
+    /**
+     * Create/Update a Zendesk User for this Northstar user and open a new
+     * ticket for them with their submitted question.
+     *
+     * @param string  $campaignId
+     * @param string  $campaignName
+     * @param string  $question
+     * @return string
+     */
+    public function createTicket($campaignId, $campaignName, $question)
+    {
+        $northstarId = auth()->id();
+
+        Log::debug('[Phoenix] ZendeskTicketsController@store: Creating Zendesk ticket:', [
+            'northstar_id' => $northstarId,
+            'question' => str_limit($question, 5000, '(...)'),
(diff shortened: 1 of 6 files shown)


app/Services/Zendesk.php
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * Fetch user email and display name from Northstar.
+     *
+     * @param string $northstarId
+     * @return array
+     */
+    protected function getUserDataFromNorthstar($northstarId)
+    {
+        $user = gateway('northstar')->withToken(token())->get('v2/users/'.$northstarId, [
```
katiecrane: Are we guaranteed to get a `$user` back here?

katiecrane: I guess if we have a valid Northstar ID we in theory should!

app/Services/Zendesk.php
```diff
@@ -0,0 +1,146 @@
...
+    {
+        $firstCampaignCause = $this->getFirstCampaignCauseName($campaignId);
+
+        $zendeskGroups = optional(
+            // Zendesk Search API: http://bit.ly/2TihwCC
+            ZendeskClient::search()->find('type:group name:".'.$firstCampaignCause.'"')
+        )->results;
+
+        // Filter out the *exact* matching group. (The search API includes a more generous matching logic).
+        $zendeskGroup = collect($zendeskGroups)->firstWhere('name', $firstCampaignCause);
```
katiecrane: Gah such a bummer that we need the group id and can't just use the group name!

katiecrane:
> This controller action is feeling crowded! I'd love to follow up on this PR with some general cleanup. Including, pulling most of this logic into a ZendeskRepository!

I feel like the `Zendesk` service took care of this!

app/Services/Zendesk.php
mendelB: Yeah, I figured since the endpoint is authenticated we _should_ always have a valid Northstar ID and thus `$user`

app/Services/Zendesk.php
mendelB: Totally 😖it certainly made this delightfully more complicated...

mendelB:
> I think this looks good! I like your idea of caching the group ids since agreed those probably won't change - wish there was an easier way to match a group name to a group id but doesn't seem like there is!

@katiecrane I added a 15-minute cache in https://github.com/DoSomething/phoenix-next/pull/1952/commits/2987f7d49bfad051f5401e974fea1015a6ce0ae3 do you think that's good for now, or should we extend that?

katiecrane:
@mendelB I'm almost wondering if we could `rememberForever` because I can't imagine something's ID changing but I don't know if we can really count on that. 15 minutes sure feels safe though, from the view of having the correct data!

mendelB:
> @mendelB I'm almost wondering if we could `rememberForever` because I can't imagine something's ID changing but I don't know if we can really count on that. 15 minutes sure feels safe though, from the view of having the correct data!

Yeah that's a good point. Our question traffic is pretty darn low for now, so maybe something to bear in mind if we see that increasing!

PR #1963, opened 2020-03-10
Add Analytics to campaigns filters

lindsayrmaher:
### What's this PR do?

This pull request adds analytics tracking to our new filters on the campaigns page and the `view more` button.

👀 

I'd like feedback on whether the way I implemented adding the url to context is the best, and whether the approach of interpolating the cause name into the label and event name makes sense. It felt like it fit the best since we'll want to track this kind of option choosing on other filters but the analytics tracking will be very similar. 

I'm basing this off of previous tracking we've added as well as a clarifying convo with Diego L on slack. 
https://dosomething.slack.com/archives/CUQMU4Q6B/p1583787779011300

### Relevant tickets

References [Pivotal # 171441057](https://www.pivotaltracker.com/story/show/171441057).

### Checklist

### DIFF
--- a/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
+++ b/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
@@ -69,6 +69,12 @@ const CauseFilter = ({ filters, setFilters, link }) => {
   };
 
   const clearAllSelected = () => {
+    trackAnalyticsEvent('filter_clear_options_cause', {
+      action: 'link_clicked',
+      category: EVENT_CATEGORIES.filter,
+      label: 'cause',
+      context: { url: link },
+    });
     if (causes) {
       setFilters({ causes: [] });
     }
--- a/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
+++ b/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
@@ -2,6 +2,11 @@ import React from 'react';
 import { get } from 'lodash';
 import PropTypes from 'prop-types';
 
+import {
+  EVENT_CATEGORIES,
+  trackAnalyticsEvent,
+} from '../../../../../helpers/analytics';
+
 const CauseInput = ({ causeChecked, causeName, causeValue, handleSelect }) => (
   <label className="flex justify-start pb-2" htmlFor={causeValue}>
     <input
@@ -46,6 +51,13 @@ const CauseFilter = ({ filters, setFilters }) => {
   };
 
   const handleCauseSelect = event => {
+    trackAnalyticsEvent('clicked_filter_options_cause', {
+      action: 'option_clicked',
+      category: EVENT_CATEGORIES.filter,
+      label: 'cause',
+      context: { value: event.target.value },
+    });
+
     if (causes.includes(event.target.value)) {
       const newCauses = causes.filter(cause => {
         return cause !== event.target.value;
--- a/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
+++ b/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
@@ -32,7 +32,7 @@ CauseInput.defaultProps = {
   causeChecked: false,
 };
 
-const CauseFilter = ({ filters, setFilters }) => {
+const CauseFilter = ({ filters, setFilters, link }) => {
   const causes = get(filters, 'causes', []);
 
   const causeLabels = {
@@ -52,10 +52,10 @@ const CauseFilter = ({ filters, setFilters }) => {
 
   const handleCauseSelect = event => {
     trackAnalyticsEvent('clicked_filter_options_cause', {
-      action: 'option_clicked',
+      action: 'button_clicked',
       category: EVENT_CATEGORIES.filter,
-      label: 'cause',
-      context: { value: event.target.value },
+      label: event.target.name,
+      context: { value: event.target.value, url: link },
     });
 
     if (causes.includes(event.target.value)) {
@@ -104,7 +104,12 @@ const CauseFilter = ({ filters, setFilters }) => {
 
 CauseFilter.propTypes = {
   filters: PropTypes.object.isRequired,
+  link: PropTypes.string,
   setFilters: PropTypes.func.isRequired,
+};
+
+CauseFilter.defaultProps = {
+  link: null,
 };
 
 export default CauseFilter;
--- a/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
+++ b/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
@@ -54,7 +54,7 @@ const CauseFilter = ({ filters, setFilters, link }) => {
     trackAnalyticsEvent('clicked_filter_options_cause', {
       action: 'button_clicked',
       category: EVENT_CATEGORIES.filter,
-      label: event.target.name,
+      label: event.target.value,
       context: { value: event.target.value, url: link },
     });
 
--- a/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
+++ b/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
@@ -69,7 +69,7 @@ const CauseFilter = ({ filters, setFilters, link }) => {
   };
 
   const clearAllSelected = () => {
-    trackAnalyticsEvent('filter_clear_options_cause', {
+    trackAnalyticsEvent('clicked_filter_clear_options_cause', {
       action: 'link_clicked',
       category: EVENT_CATEGORIES.filter,
       label: 'cause',
--- a/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
+++ b/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
@@ -32,7 +32,7 @@ CauseInput.defaultProps = {
   causeChecked: false,
 };
 
-const CauseFilter = ({ filters, setFilters, link }) => {
+const CauseFilter = ({ filters, setFilters }) => {
   const causes = get(filters, 'causes', []);
 
   const causeLabels = {
@@ -55,7 +55,7 @@ const CauseFilter = ({ filters, setFilters, link }) => {
       action: 'button_clicked',
       category: EVENT_CATEGORIES.filter,
       label: event.target.value,
-      context: { value: event.target.value, url: link },
+      context: { value: event.target.value },
     });
 
     if (causes.includes(event.target.value)) {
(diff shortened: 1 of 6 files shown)


resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
```diff
@@ -46,6 +51,13 @@ const CauseFilter = ({ filters, setFilters }) => {
   };
 
   const handleCauseSelect = event => {
+    trackAnalyticsEvent('clicked_filter_options_cause', {
+      action: 'button_clicked',
+      category: EVENT_CATEGORIES.filter,
+      label: event.target.name,
```
mendelB: Why do we use `event.target.name` here and `event.target.value` in the `context.value`?

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
```diff
@@ -57,6 +69,12 @@ const CauseFilter = ({ filters, setFilters }) => {
   };
 
   const clearAllSelected = () => {
+    trackAnalyticsEvent('filter_clear_options_cause', {
```
mendelB: Is this meant to be `clicked_filter_clear_options_cause`?

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
```diff
@@ -57,6 +69,12 @@ const CauseFilter = ({ filters, setFilters }) => {
...
+      action: 'link_clicked',
+      category: EVENT_CATEGORIES.filter,
+      label: 'cause',
+      context: { url: link },
```
mendelB: I don't think we need this `link` here since we'll always have the current URL included in any tracked event.

resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
```diff
@@ -67,6 +72,12 @@ const PaginatedCampaignGallery = ({
   const { endCursor, hasNextPage } = get(data, 'campaigns.pageInfo', {});
 
   const handleViewMore = () => {
+    trackAnalyticsEvent('clicked_link_view_more_campaigns', {
+      action: 'link_clicked',
+      category: EVENT_CATEGORIES.siteAction,
+      label: 'view_more',
+      context: { url: link },
```
mendelB: Same as the above `link`s, this will be tracked automatically by the event, so I don't think we need to explicitly include this.

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
lindsayrmaher: ahh yes, i misread the google sheet. thanks for 🎏

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
lindsayrmaher: I think I mostly was trying to match the `field_name` we call out on the analytics spreadsheet! Do you think it makes for sense for them to be the same value?

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
lindsayrmaher: Ah ok, so should the URL show up in the context object when I trigger events and log them in my console? I was under the impression that the context we always pass through was UTMs not the URL

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
lindsayrmaher: updated on `0da4a13`

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
mendelB: Hmm it looks like the `name` and `value` attributes are always going to be [the same](https://github.com/DoSomething/phoenix-next/blob/ec6e9e73aeb7cd2a00a8087c198d60eecd56bfd1/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js#L8-L11) on this `input` tag, so maybe worth just sticking with one of 'em for consistency?

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
mendelB: Sorry I should have clarified! This would be tracked by the analytics libraries themselves, so Snowplow will always include the URL in the events metadata (which I believe is filterable and discoverable in Looker via the `event.href` or something), and GTM/GA will do the same. So it _shouldn't_ actually be something we'd need to worry about or debug! 

But if you want to be sure, you can inspect the network request for the analytics event (e.g. for Snowplow you can filter for `fivetran` in the Network tab, and check the `url` param):
image

weerd: We do toss in `url` in other analytics events and maybe just something we need to clear up across our events cause I could see this coming up again at a later date. While it technically doesn't hurt to include it in context, we may want to also callout that it's always a part of every triggered event payload (similar to how we call it out w/ `utms` in the context).

weerd: Unless maybe there's a very specific reason (I don't remember well) that we added `url` to other triggered events in context that maybe the Data Team knows about 🤷‍♂

mendelB: I believe we use `url` in context for events where we're either sharing or clicking on a distinct link, so the `url` is important aside from the current page URL the user is on.

(e.g. in the `CtaPopoverButton`, where you're clicking a button which is its own link, which is useful to track, or the [`ShareAction`](https://github.com/DoSomething/phoenix-next/blob/ec6e9e73aeb7cd2a00a8087c198d60eecd56bfd1/resources/assets/components/actions/ShareAction/ShareAction.js#L94) where we want to know which link you're sharing)

But here we're literally just using this to track the current page URL, which seems superfluous?

weerd: Counter argument, I don't think it does any _harm_ if we do add it, and to paraphrase Señor Heydari, "more/extra data is better than missing less/missing data"... 🤔 

cc @sheyd

sheyd: I think there's some nuance the meaning of `url`. To Mendel's point, if we're using it for other actions to get the url of a link related to an event itself vs current page url, then either leaving it out _or_ using a different name would be useful. However, if I'm misunderstanding and there is no ambiguity, then at this point (especially as Param is getting onboarded), I'd opt to include it vs exclude it.

In general I'm not a fan of data bloat, but Param coming to this with fresh eyes I think will help us determine what to keep/discard in the long run.

mendelB: Def agreed that it can't hurt to leave in an extra piece of context. I just feel like in this case:
1. We're literally repeating a piece of metadata tracked with each event. But also -- a piece of metadata that we already _use_ to filter events in Looker dashboards or GA filters (e.g. [here's a Look](https://dsdata.looker.com/explore/blade/phoenix_next_events?toggle=fil&qid=tH0yH8EPc0WswacUOLYJsj) Tej linked to the other day which he was using to analyze quiz activity, with the Events Path (the page url) as a filter). The page URL or event path is like an essential bit of metadata which is why this might be confusing.
2. To your point Sena -- we've been using the `url` to mean specifically 'the url of the link related to the event itself vs current page url', so in this case, it could be misleading!

weerd: Ah, all good points @mendelB @sheyd! I initially suspected it seemed a bit superfluous but I didn't want to discount it, since not sure to what degree it needed to be used or why it was being considered added. I could see if we end up allowing to add filters directly on the list of campaigns on the new homepage or some other location outside of just the explore campaigns, I could maybe see `url` being useful in the context 🤷‍♂ 

I'm fine with whatever the recommendation is 😄

lindsayrmaher: it sounds like there's a case for removing it, so i shall if everyone wants to give me their 👍

weerd: > we've been using the url to mean specifically 'the url of the link related to the event itself vs current page url', so in this case, it could be misleading!

I think this is an important point to note for Product to when their deciding what data to include in context... cc @ngjo (tried to tag DiegoD, but not sure if he has a GH account)

lindsayrmaher:
all unnecessary references to the url in context were removed in [869b2b3](https://github.com/DoSomething/phoenix-next/pull/1963/commits/869b2b38c7085834c948c5a7a31e60e7033800e7)

cc @sheyd @mendelB @weerd

weerd:
"The Great URL Context Purge" of 2020 🗑

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -9,6 +13,13 @@ const FilterNavigation = ({ filters, setFilters }) => {
   const [showFilterMenu, setShowFilterMenu] = useState(false);
 
   const handleMenuToggle = filterName => {
+    console.log(filterName);
```
weerd: Lingering `console.log()`! 👾

weerd: I recommend always using emoji in console logs!

1. You can have some fun with them when logging to the console (I sometimes try to console log multiple things and keep a theme... (🌂, ☂️, ☔️ ), or default to 💩
2. They stand out in the console amongst a slew of other debug messages. 
3. Both in code and in the GH diff view they're IMMEDIATELY obvious that you forgot to remove a console log 😉 

🎉

lindsayrmaher: drat these darn console logs! brb 🏃‍♀

lindsayrmaher:
@weerd just pushed updates based on the disco for the view more event. would you mind just 👀 before I merge?

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -9,6 +13,12 @@ const FilterNavigation = ({ filters, setFilters }) => {
   const [showFilterMenu, setShowFilterMenu] = useState(false);
 
   const handleMenuToggle = filterName => {
+    trackAnalyticsEvent(`clicked_filter_button_${filterName}`, {
+      action: 'button_clicked',
+      category: EVENT_CATEGORIES.filter,
+      label: `${filterName.toLowerCase()}_filter_menu`,
```
weerd: You don't necessarily need to worry about lowercasing the value. We handle this in `helpers/analtyics.js`, `snakeCase()` for Snowplow and `startCase()` for GA :)

PR #1964, opened 2020-03-10
Additional Content Patching to Correct Bug

Vmack:
### What's this PR do?

This pull request updates the `LedeBannerContainer` and the `HeroTemplate` to fix additional content bug and allow for `numberOfScholarships` to be added manually. See [thread](https://dosomething.slack.com/archives/CUQMU4Q6B/p1583773079003600)

👀

...

### Relevant tickets

References [Pivotal #171086615](https://www.pivotaltracker.com/story/show/171086615).

### Checklist

### DIFF
--- a/resources/assets/components/LedeBanner/LedeBannerContainer.js
+++ b/resources/assets/components/LedeBanner/LedeBannerContainer.js
@@ -26,7 +26,7 @@ const mapStateToProps = (state, props) => ({
   numberOfScholarships: get(
     state,
     'campaign.additionalContent.numberOfScholarships',
-    null,
+    undefined,
   ),
 
   affiliateOptInContent: state.campaign.affiliateOptInContent,
--- a/resources/assets/components/LedeBanner/LedeBannerContainer.js
+++ b/resources/assets/components/LedeBanner/LedeBannerContainer.js
@@ -23,6 +23,7 @@ const mapStateToProps = (state, props) => ({
     'campaign.additionalContent.affiliateCreditText',
     undefined,
   ),
+  numberOfScholarships: get(state, 'campaign.additionalContent', null),
   affiliateOptInContent: state.campaign.affiliateOptInContent,
   affiliateSponsors: state.campaign.affiliateSponsors,
   // @TODO: We won't need this blurb property once MosaicTemplate is retired.
@@ -39,6 +40,7 @@ const mapStateToProps = (state, props) => ({
   isAffiliated: isSignedUp(state),
   scholarshipAmount: state.campaign.scholarshipAmount,
   scholarshipDeadline: state.campaign.scholarshipDeadline,
+  numberOfScholarships: state.campaign.additionalContent,
   subtitle: get(props, 'subtitle', state.campaign.callToAction),
   title: get(props, 'title', state.campaign.title),
 });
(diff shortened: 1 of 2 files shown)


resources/assets/components/LedeBanner/LedeBannerContainer.js
```diff
@@ -23,6 +23,12 @@ const mapStateToProps = (state, props) => ({
     'campaign.additionalContent.affiliateCreditText',
     undefined,
   ),
+  numberOfScholarships: get(
+    state,
+    'campaign.additionalContent.numberOfScholarships',
+    1,
```
weerd: Should this default to `1`? Won't this incorrectly show that a campaign without scholarships force it to assume it has one?

weerd: Before this code existed inside of the conditional that depended on `showScholarshipModal` being true. I think it could potentially be confusing if it now defaults to `1` instead of `null`.

katiecrane: I think here it should be `null`, but later in the modal when we are displaying the scholarship information, if we do not have a `numberOfScholarships` value we should display 1

Vmack: Ok, makes sense. So I'll set it to null in the `LedeBanner` container. Thanks!

resources/assets/components/LedeBanner/templates/HeroTemplate.js
```diff
@@ -139,11 +140,7 @@ const HeroTemplate = ({
               scholarshipCallToAction={scholarshipCallToAction || undefined}
               scholarshipDeadline={scholarshipDeadline}
               scholarshipDescription={scholarshipDescription}
-              numberOfScholarships={get(
-                additionalContent,
-                'numberOfScholarships',
-                1,
-              )}
+              numberOfScholarships={numberOfScholarships}
```
katiecrane: If this ends up being `null`, will we still display 1 from the `ScholarshipInfoBlock`?

Vmack: That's a good question, so far it's working fine but what's another valid campaign I can test?

katiecrane: Did you test on a campaign that does not have `numberOfScholarships` in its `additionalContent`?

katiecrane: ^I think if we test that way and it still looks good then we've got our bases all covered!

weerd: Looks like the `ScholarshipInfoBlock` indicates that `numberOfScholarships` is required, so `null` there would likely error out.

You could probably set the value in the `LedeBannerContainer` to default to `undefined` and then in this `HeroTemplate` set the default prop to `numberOfScholarships: 1` 🤔

weerd: Default props in React don't trigger if the incoming prop is set to `null`. It needs to be `undefined` for it to trigger... thus why we made the `withoutNulls()` helper function to pass an object and remove any props with `null` values so we can have the default props on the component trigger 💃

katiecrane: > You could probably set the value in the LedeBannerContainer to default to undefined instead of null and then in this HeroTemplate set the default prop to numberOfScholarships: 1

This seems like a clean solution to me!

PR #1966, opened 2020-03-13
Pull scholarship CTA and description into LedeBanner

katiecrane:
### What's this PR do?

This pull request fixes a bug where the scholarship CTA and description were not pulled in from Contentful, so we always displayed the defaults for these fields.

Data in Contentful:
image

and that data now showing up as expected in the modal!
image

Is the content being pulled in as expected now?

This was a flagged earlier this week.

I'm bumping Cypress tests to a future PR because the content is coming straight from Contentful, not GraphQL so I can't just easily mock (is that true??) it and I want to get this fix out!

### Relevant tickets

References [Pivotal #171726725](https://www.pivotaltracker.com/story/show/171726725).

### Checklist

### DIFF
--- a/resources/assets/components/LedeBanner/LedeBannerContainer.js
+++ b/resources/assets/components/LedeBanner/LedeBannerContainer.js
@@ -44,7 +44,9 @@ const mapStateToProps = (state, props) => ({
   ),
   isAffiliated: isSignedUp(state),
   scholarshipAmount: state.campaign.scholarshipAmount,
+  scholarshipCallToAction: state.campaign.scholarshipCallToAction,
   scholarshipDeadline: state.campaign.scholarshipDeadline,
+  scholarshipDescription: state.campaign.scholarshipDescription,
   subtitle: get(props, 'subtitle', state.campaign.callToAction),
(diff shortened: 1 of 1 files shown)


DFurnes:
> I'm bumping Cypress tests to a future PR because the content is coming straight from Contentful, not GraphQL so I can't just easily mock (is that true??) it and I want to get this fix out!

We can mock the `window.STATE` variable (using our [`withState`](https://github.com/DoSomething/phoenix-next/blob/170c333f3f58a08122003c9d6cc26c9272e57f15/cypress/support/commands.js#L103-L113) helper & something like [this fixture](https://github.com/DoSomething/phoenix-next/blob/master/cypress/fixtures/contentful/exampleCampaign.js)) but it's a little awkward to do, so I'd probably recommend against it for now!

katiecrane:
@DFurnes yeah seemed like it was gonna be awkward...would you recommend against testing it at all right now? (even in a future PR?)

DFurnes:
Yeah, personally I think customizing those Contentful mocks per test-case is more trouble than it's worth right now.

PR #1971, opened 2020-03-17
Homepage graphql queries

weerd:
### What's this PR do?

This pull request further progresses the new Homepage design. It incorporates the GraphQL work in https://github.com/DoSomething/graphql/pull/197 and continues on the styling. There is still a bit more work to be done.

New Home Page progress with Campaigns

👀 and maybe some clickity-clicks on the review app.

A couple things to note. The top header is not finalized just yet, and the first "featured" item in the campaign gallery is not completely styled/behaving as it should based on the mocks. More on that to come, but I didn't want the PR to get too complex for review.

### Relevant tickets

References [Pivotal #171056375](https://www.pivotaltracker.com/story/show/171056375).

### Checklist

- [x] Added GIF of front-end changes on small, medium, and large screens.

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.3.0",
     "sixpack-client": "^2.1.0",
+    "tailwindcss": "^1.2.0",
     "twin.macro": "^1.0.0-alpha.7",
     "usa-states": "0.0.5"
   },
@@ -169,7 +170,6 @@
     "redux-devtools": "^3.5.0",
     "redux-devtools-dock-monitor": "^1.1.3",
     "redux-devtools-log-monitor": "^1.4.0",
-    "tailwindcss": "^1.2.0",
     "turndown": "^5.0.3",
     "webpack": "~4.28.0",
     "webpack-bundle-analyzer": "^3.6.0",
--- a/resources/assets/components/pages/HomePage/HomePageCampaignGallery.js
+++ b/resources/assets/components/pages/HomePage/HomePageCampaignGallery.js
@@ -1,27 +1,73 @@
 import React from 'react';
-import { first } from 'lodash-es';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/core';
+import resolveConfig from 'tailwindcss/resolveConfig';
+
+import tailwindConfig from '../../../../../tailwind.config';
+import CampaignGalleryItem from '../../utilities/Gallery/templates/CampaignGalleryItem/CampaignGalleryItemV2';
+import CampaignGalleryFeaturedItem from '../../utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem';
 
 const HomePageCampaignGallery = ({ campaigns }) => {
-  console.log('💩', campaigns);
+  const tailwindSpacing = resolveConfig(tailwindConfig).theme.spacing;
+  const tailwindScreens = resolveConfig(tailwindConfig).theme.screens;
 
   return (
-    <ul className="campaign-gallery">
-      {campaigns.map((campaign, index) => {
-        console.log('😎', index, campaign);
+    <ul
+      className="campaign-gallery mt-0"
+      css={css`
+        display: grid;
+        grid-template-columns: 1fr;
+        grid-row-gap: ${tailwindSpacing['8']};
 
+        @media (min-width: ${tailwindScreens.md}) {
+          grid-template-columns: 1fr 1fr;
+          grid-column-gap: ${tailwindSpacing['8']};
+        }
+
+        @media (min-width: ${tailwindScreens.xxl}) {
+          grid-template-columns: 1fr 1fr 1fr;
+        }
+      `}
+    >
+      {campaigns.map((campaign, index) => {
         const firstItem = index === 0;
 
         return (
-          <li key={campaign.id}>
-            <article>
-              {firstItem ? <span>Featured</span> : null}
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -131,6 +131,7 @@
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.3.0",
     "sixpack-client": "^2.1.0",
+    "tailwindcss": "^1.2.0",
```
weerd: Moving this into the dependencies list allows us to pull in the `resolveConfig` helper which makes it A LOT easier to dynamically insert Tailwind config values into our JS rendered output when needed!

resources/assets/components/pages/HomePage/HomePageCampaignGallery.js
```diff
@@ -0,0 +1,73 @@
...
+import { css } from '@emotion/core';
+import resolveConfig from 'tailwindcss/resolveConfig';
+
+import tailwindConfig from '../../../../../tailwind.config';
+import CampaignGalleryItem from '../../utilities/Gallery/templates/CampaignGalleryItem/CampaignGalleryItemV2';
+import CampaignGalleryFeaturedItem from '../../utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem';
+
+const HomePageCampaignGallery = ({ campaigns }) => {
+  const tailwindSpacing = resolveConfig(tailwindConfig).theme.spacing;
+  const tailwindScreens = resolveConfig(tailwindConfig).theme.screens;
```
weerd: This is super helpful. I'll be moving the work I did to separate out the Tailwind Screens values in a separate file, back into the main `tailwind.config.js` file and using this method instead to grab them moving forward!

tailwind.config.js
```diff
@@ -1,15 +1,23 @@
...
+        100: '#f7fafc',
+        200: '#edf2f7',
+        300: '#e2e8f0',
+        400: '#cbd5e0',
+        500: '#a0aec0',
+        600: '#718096',
+        700: '#4a5568',
+        800: '#2d3748',
+        900: '#1a202c',
+      },
```
weerd: I was chatting with @lkpttn and we agreed that having the gray values in the config was easier for when needing to look up and reference values in the mocks, as opposed to needing to head into the default Tailwind config in the NPM package. To that end, it made sense to just add the other 3 items we were importing the `colors` variable for and just remove the `require()` statement. Easy peasy.

Donezo

resources/assets/components/pages/HomePage/NewHomePage.js
```diff
@@ -1,15 +1,58 @@
...
+        ... on Showcasable {
+          showcaseTitle
+          showcaseDescription
+          showcaseImage {
+            url
+          }
+        }
+        ... on CampaignWebsite {
+          id
+          url
```
weerd: So I think I need to go back and expose `staffPick` from Contentful to properly do the "Feature" badge on the campaigns within the gallery that aren't the first "featured" item in the gallery. 🤔

resources/assets/components/pages/HomePage/HomePageCampaignGallery.js
DFurnes: I wonder if we should make a `tailwind.js` that does this for us, so we don't have to import `resolveConfig` and re-process the config files each time we want to access a variable?

resources/assets/components/pages/HomePage/HomePageCampaignGallery.js
```diff
@@ -0,0 +1,73 @@
...
+
+  return (
+    <ul
+      className="campaign-gallery mt-0"
+      css={css`
+        display: grid;
+        grid-template-columns: 1fr;
+        grid-row-gap: ${tailwindSpacing['8']};
```
DFurnes: This should also be possible with Tailwind's grid helpers, if you think that reads any nicer:

```js
css={tw`mt-0 grid gap-8 grid-cols-1 md:grid-cols-2 xxl:grid-cols-3`}
```

resources/assets/components/pages/HomePage/NewHomePage.js
```diff
@@ -1,15 +1,58 @@
...
 
-const NewHomePage = () => {
+const HOME_PAGE_QUERY = gql`
+  query HomePageQuery($preview: Boolean!) {
+    page: homePage(preview: $preview) {
+      id
+      title
+      subTitle
+      campaigns {
+        __typename
```
DFurnes: The `__typename` field should be automatically included, I believe.

resources/assets/components/pages/HomePage/HomePageCampaignGallery.js
weerd: Do you mean maybe a helper function in `/helpers` instead or better as standalone, importable file?

resources/assets/components/pages/HomePage/HomePageCampaignGallery.js
weerd: Ah nice! I just moved Tailwind package to main dependencies but didn't realize you had also updated to the latest Tailwind version which includes grid classes in your prior PR! I'll review and then update for upcoming PR 💃

resources/assets/components/pages/HomePage/HomePageCampaignGallery.js
DFurnes: Oh, a helper could also be all we need! I'm good either way. 😸

weerd: Ok! I'll revise in upcoming PR 💃

PR #1972, opened 2020-03-18
Use aliases for conflicting LinkBlock fragment fields

mendelB:
### What's this PR do?

This pull request adds [aliases](https://graphql.org/learn/queries/#aliases) to two conflicting fields in the `LinkBlockFragment`. The `title` & `link` conflict with other `block` fields of the same name which are _not_ required.

We've updated them to required in https://github.com/DoSomething/graphql/commit/53806590a533f2863a096c4e9b0c0dccfa1d9083 which broke our epic block query [here](https://github.com/DoSomething/phoenix-next/blob/92040861169ab80d1d927c82d3077e793d89a683/resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js#L35)

👀 

You can check out the PR description here for more context on these aliases https://github.com/DoSomething/phoenix-next/pull/1681

### Relevant tickets

References [Pivotal #171824364](https://www.pivotaltracker.com/story/show/171824364).
https://dosomething.slack.com/archives/C02BBP0CU/p1584540907002700?thread_ts=1584540771.002300&cid=C02BBP0CU
### Checklist

### DIFF
--- a/resources/assets/components/ContentfulEntry/ContentfulEntry.js
+++ b/resources/assets/components/ContentfulEntry/ContentfulEntry.js
@@ -141,7 +141,16 @@ class ContentfulEntry extends React.Component {
         return <LandingPageContainer {...json.fields} />;
 
       case 'LinkBlock':
-        return <LinkActionContainer {...withoutNulls(json)} />;
+        return (
+          <LinkActionContainer
+            {...withoutNulls({
+              ...json,
+              // Resolves the aliases used in the LinkBlockFragment.
+              title: json.linkBlockTitle,
+              link: json.linkActionLink,
+            })}
+          />
+        );
 
       case 'PetitionSubmissionBlock':
         return (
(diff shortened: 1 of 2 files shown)


DFurnes:
Just as an aside, I've been thinking about trying to adjust how we query blocks to get rid of the need for this monster "loader" component... since it causes some awkwardness with naming collisions (like here), and it also sets us up for trouble with non-block loaders.

I'm going to set aside some time to dig into a better solution (via a prototype or at least a better write-up of my thoughts) later this week or next, but just wanted to say I know it's a pain point!

mendelB:
> Just as an aside, I've been thinking about trying to adjust how we query blocks to get rid of the need for this monster "loader" component... since it causes some awkwardness with naming collisions (like here), and it also sets us up for trouble with non-block loaders.
> 
> I'm going to set aside some time to dig into a better solution (via a prototype or at least a better write-up of my thoughts) later this week or next, but just wanted to say I know it's a pain point!

🙌 That sounds awesome!! Something I guess very relevant that we could also discuss in our Landing Page jam!

PR #1976, opened 2020-03-18
Changes image crop parameter to common 16:9 ratios

lkpttn:
### What's this PR do?

This pull request updates the Hero Template of the Lede Banner component to use proper (and common) 16:9 crops, to better match with assets we're receiving from Creative/Marketing/Campaigns. We'll now serve the following sizes for cropped images:

- Extra-Large is 2880x1620
- Large is 1440 x 810
- Medium is 1024×576
- Small is 640x360

👀 

When we swapped over from the mosaic template to the hero template, we wanted to lower the amount of differences displaying creative assets so we decided that all campaign images in the lede banner, galleries, articles, etc would be in the common 16:9 format. There was some miscommunication in the lede banner which would lead to parts of people's faces or other important image parts being cut off, so we're correcting the cropped image aspect ratios.

### Relevant tickets

References [Pivotal #171859459](https://www.pivotaltracker.com/story/show/171859459).

**Before**
poorcrop

**After**
propercrop

### DIFF
--- a/resources/assets/components/LedeBanner/templates/HeroTemplate.js
+++ b/resources/assets/components/LedeBanner/templates/HeroTemplate.js
@@ -51,7 +51,7 @@ const HeroTemplate = ({
   // @TODO: If this experiment is successful we should turn generating the series urls for
   // the cover image photo at different sizes into a helper function!
   const coverImageUrls = {
-    extraLarge: contentfulImageUrl(coverImage.url, '1920', '1080', 'fill'),
+    extraLarge: contentfulImageUrl(coverImage.url, '2880', '1620', 'fill'),
     large: contentfulImageUrl(coverImage.url, '1440', '810', 'fill'),
     medium: contentfulImageUrl(coverImage.url, '1024', '576', 'fill'),
     small: contentfulImageUrl(coverImage.url, '640', '360', 'fill'),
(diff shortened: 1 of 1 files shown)


resources/assets/components/LedeBanner/templates/HeroTemplate.js
```diff
@@ -51,10 +51,10 @@ const HeroTemplate = ({
   // @TODO: If this experiment is successful we should turn generating the series urls for
   // the cover image photo at different sizes into a helper function!
   const coverImageUrls = {
-    extraLarge: contentfulImageUrl(coverImage.url, '2232', '1000', 'fill'),
-    large: contentfulImageUrl(coverImage.url, '1116', '500', 'fill'),
-    medium: contentfulImageUrl(coverImage.url, '720', '350', 'fill'),
-    small: contentfulImageUrl(coverImage.url, '360', '200', 'fill'),
+    extraLarge: contentfulImageUrl(coverImage.url, '1920', '1080', 'fill'),
```
weerd: This should be double `1440` so `2880`, similar to how `2232` was double the `1116`. I believe this is to address issues before when there were complaints that on retina screens the images were grainy/blurry.

lkpttn: The proper ratio would be `2800` x `1575`

```
2800 / 16 = 175
1575 / 9  = 175
```

weerd:
This also requires updating the values defined in the `srcset` attribute that indicates the images intrinsic widths; the `360w`, etc bits.

weerd:
The values you changed is just changing the different sized image URL requests to Contentful's Image API.

weerd:
> Extra-Large is 1920x1080

Update your description 😄

PR #1977, opened 2020-03-18
View Details Hyperlink 

Vmack:
### What's this PR do?

This pull request adds a hyperlink to the `scholarshipInfoBlock` that allows the user to view additional scholarship details

👀

**Large**
Screen Shot 2020-03-19 at 12 53 41 PM

**Medium**
Screen Shot 2020-03-19 at 12 57 46 PM

**Small**
Screen Shot 2020-03-19 at 12 54 14 PM

When view scholarship details is clicked [this is what it looks like](https://github.com/DoSomething/phoenix-next/pull/1961)

### Relevant tickets

References [Pivotal #171060593](https://www.pivotaltracker.com/story/show/171060593).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
+++ b/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
@@ -65,8 +65,12 @@ const CampaignInfoBlock = ({
                   <dt>Next Deadline</dt>
                   <dd>{getHumanFriendlyDate(scholarshipDeadline)}</dd>
                   <div>
-                    <button type="button" onClick={showModal}>
-                      <p className="view__details">View Scholarship Details</p>
+                    <button
+                      className="text-blue-500"
+                      type="button"
+                      onClick={showModal}
+                    >
+                      View Scholarship Details
                     </button>
                   </div>
                   <hr className="clear-both pb-3 border-gray-500" />
--- a/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
+++ b/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
@@ -66,7 +66,7 @@ const CampaignInfoBlock = ({
                   <dd>{getHumanFriendlyDate(scholarshipDeadline)}</dd>
                   <div>
                     <button type="button" onClick={showModal}>
-                      <p>View Scholarship Details</p>
+                      <p className="view__details">View Scholarship Details</p>
                     </button>
                   </div>
                   <hr className="clear-both pb-3 border-gray-500" />
--- a/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
+++ b/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
@@ -31,6 +31,7 @@ const CampaignInfoBlock = ({
   campaignId,
   scholarshipAmount,
   scholarshipDeadline,
+  showModal,
 }) => (
   <Card className="bordered p-3 rounded campaign-info">
     <Query query={CAMPAIGN_INFO_QUERY} variables={{ campaignId }}>
@@ -63,6 +64,11 @@ const CampaignInfoBlock = ({
 
                   <dt>Next Deadline</dt>
                   <dd>{getHumanFriendlyDate(scholarshipDeadline)}</dd>
+                  <div>
+                    <button type="button" onClick={showModal}>
+                      <p>View Scholarship Details</p>
+                    </button>
+                  </div>
                   <hr className="clear-both pb-3 border-gray-500" />
                 </React.Fragment>
               ) : null}
@@ -97,11 +103,13 @@ CampaignInfoBlock.propTypes = {
   campaignId: PropTypes.number.isRequired,
   scholarshipAmount: PropTypes.number,
   scholarshipDeadline: PropTypes.number,
+  showModal: PropTypes.func,
 };
 
 CampaignInfoBlock.defaultProps = {
   scholarshipAmount: null,
   scholarshipDeadline: null,
+  showModal: null,
 };
 
 export default CampaignInfoBlock;
--- a/resources/assets/components/blocks/CampaignInfoBlock/campaign-info-block.scss
+++ b/resources/assets/components/blocks/CampaignInfoBlock/campaign-info-block.scss
@@ -52,7 +52,4 @@
     font-size: theme('fontSize.3xl');
     line-height: 1;
   }
-  p.view__details {
-    color: #23b7fb;
-  }
 }
--- a/resources/assets/components/blocks/CampaignInfoBlock/campaign-info-block.scss
+++ b/resources/assets/components/blocks/CampaignInfoBlock/campaign-info-block.scss
@@ -52,4 +52,7 @@
     font-size: theme('fontSize.3xl');
     line-height: 1;
   }
+  p.view__details {
+    color: #23b7fb;
+  }
 }
--- a/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
+++ b/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
@@ -106,7 +106,7 @@ const ScholarshipInfoBlock = ({
   if (error || scholarshipError) {
     console.error(`[ErrorBlock] ${error}`);
     report(error);
-    return <ErrorBlock />;
+    return <ErrorBlock error={error || scholarshipError} />;
   }
 
   return (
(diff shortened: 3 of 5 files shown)


resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
```diff
@@ -63,6 +64,11 @@ const CampaignInfoBlock = ({
 
                   <dt>Next Deadline</dt>
                   <dd>{getHumanFriendlyDate(scholarshipDeadline)}</dd>
+                  <div>
+                    <button type="button" onClick={showModal}>
+                      <p className="view__details">View Scholarship Details</p>
```
weerd: You don't really need the `<p>` tag here. You can just leave the text inside of the `<button>` without needing to add the `<p>` inside it. Might also remove the need to add that single style rule.

resources/assets/components/blocks/CampaignInfoBlock/campaign-info-block.scss
```diff
@@ -52,4 +52,7 @@
     font-size: theme('fontSize.3xl');
     line-height: 1;
   }
+  p.view__details {
+    color: #23b7fb;
+  }
```
weerd: See [comment](https://github.com/DoSomething/phoenix-next/pull/1977/files#r395197825), but even if you need to style it, you should use Tailwind classes instead.

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -45,6 +45,21 @@ const SCHOLARSHIP_AFFILIATE_QUERY = gql`
   }
 `;
 
+const SCHOLARSHIP_INFO_QUERY = gql`
+  query ScholarshipAffiliateQuery($campaignId: Int!) {
```
weerd: This should be renamed right?

weerd: I think I'm a bit confused what's going on here. Why is there a second query with very similar information retrieved? 🤔

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
Vmack: you're right, made the adjustments.

resources/assets/components/blocks/CampaignInfoBlock/campaign-info-block.scss
Vmack: thank you thank you

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
lindsayrmaher: hello! I have some ideas we could try, it does feel like we could avoid duplication.

I'm guessing the second query should be used in the case that the user isn't coming directly from an affiliate site etc. (@Vmack let me know if that's not true!) So instead of duplicating the entire query, we could _hopefully_ just make some adjustments to the existing one so that it meets our needs either way!

1. Maybe we use the `SCHOLARSHIP_INFO_QUERY` name instead of including Affiliate in the name
2. Could we remove the requirement to send over the `utmLabel` and `preview`? Or remove the `isRequired` flag in propTypes from `utmLabel` so that it we can define a default value for it? 

As long as the query can still run and return the key pieces of info, hopefully tweaking the existing query would do the trick!

DFurnes: You might also consider breaking this into two (completely separate) queries – one with `affiliate` (used if we need it) and one with `actions, campaign`. Since these get batched up into a single request, there should be no performance difference between querying three things in one query or two.

Vmack: @DFurnes your suggestion worked, thanks! 🙌🏽 Let me know if this is what you were describing?

DFurnes: Yeah, the two split queries look perfect! I think you'll still want to update your component code to query both (via two separate `useQuery()` hooks) so that you can still see campaign/action info on scholarship info blocks, though.

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -56,12 +60,22 @@ const ScholarshipInfoBlock = ({
...
+    ({ loading, error, data } = useQuery(SCHOLARSHIP_AFFILIATE_QUERY, {
+      variables: {
+        utmLabel,
+        preview: env('CONTENTFUL_USE_PREVIEW_API'),
+        campaignId,
+      },
+    }));
+  }
+
+  const { loading2, error2, data2 } = useQuery(SCHOLARSHIP_INFO_QUERY, {
```
DFurnes: This is [destructuring an object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) that the `useQuery` hook responds – so it's actually looking for properties named `loading2`, `error2`, and `data2` on the object (and likely not finding them!). Here's what [that response object](https://www.apollographql.com/docs/react/api/react-hooks/#result) looks like.

You probably want to rename these instead, like this:

```suggestion
  const { loading: scholarshipLoading, error: scholarshipError, data: scholarshipData } = useQuery(SCHOLARSHIP_INFO_QUERY, {
```

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -56,12 +60,22 @@ const ScholarshipInfoBlock = ({
...
+
+  if (utmLabel) {
+    ({ loading, error, data } = useQuery(SCHOLARSHIP_AFFILIATE_QUERY, {
+      variables: {
+        utmLabel,
+        preview: env('CONTENTFUL_USE_PREVIEW_API'),
+        campaignId,
+      },
+    }));
+  }
```
DFurnes: This might read a little more cleanly using the [`skip` parameter](https://www.apollographql.com/docs/react/api/react-hooks/#params) for the `useQuery` hook:

```suggestion
  let loading;
  let error;
  let data;

const { loading, error, data } = useQuery(SCHOLARSHIP_AFFILIATE_QUERY, {
  skip: utmLabel === null,
  variables: {
    utmLabel,
    preview: env('CONTENTFUL_USE_PREVIEW_API'),
    campaignId,
  },
}));
```

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
Vmack: @DFurnes You are a genius! That works, thanks! And that makes total sense, it couldn't find the value of the prop because I renamed the props instead of just giving the props an alternate value.

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -81,16 +94,15 @@ const ScholarshipInfoBlock = ({
...
+  const endDate = get(scholarshipData, 'campaign.endDate');
+  const actions = get(scholarshipData, 'actions', []);
   const actionItem = actions.find(
     action => action.scholarshipEntry && action.reportback,
   );
   const actionType = get(actionItem, 'actionLabel', '');
-
-  if (error) {
+  if (error || scholarshipError) {
     console.error(`${error}`);
```
DFurnes: Should this return an alternative error interface (since I think if we error'ed out but continued down to "main" render function, it'd probably look a little broken). Maybe we should return an [ErrorBlock](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/blocks/ErrorBlock/ErrorBlock.js) here?

Vmack: @DFurnes Great idea, added! Let me know if the changes are good to go!

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -81,17 +95,17 @@ const ScholarshipInfoBlock = ({
...
+  const actions = get(scholarshipData, 'actions', []);
   const actionItem = actions.find(
     action => action.scholarshipEntry && action.reportback,
   );
   const actionType = get(actionItem, 'actionLabel', '');
-
-  if (error) {
-    console.error(`${error}`);
+  if (error || scholarshipError) {
+    <ErrorBlock />;
```
DFurnes: To render this component, you need to add a `return` to the beginning of this line. Otherwise, it's just a "statement" that gets discarded. (Some good news – you can pass an `error={error || scholarshipError}` prop to this component, which will automatically log and report that error when displaying this error block!)

PR #1979, opened 2020-03-18
Add hacky hardcoded metadata fallback.

DFurnes:
### What's this PR do?

This pull request adds a nice and hacky `$HARDCODED_METADATA` object that we can use to override metadata on content types (like campaign collections) that don't support it.

Please be nice! This is rushed together to help with a COVID-19 rapid response going out today.

This is necessary because this page is fully client-rendered, and while Google is able to crawl and parse client-rendered pages, Facebook and Twitter are not.

### Relevant tickets

I'll write a ticket later today! References [this Slack conversation](https://dosomething.slack.com/archives/CJ340DEEL/p1584550963018100)

### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -418,8 +418,7 @@ function get_metadata($entry)
  * A hacky function that allows us to fall back to hard-coded metadata for special
  * pages that don't support editor-specified metadata (e.g. campaign collections).
  *
- * @param array $metadata
- * @param string $path
+ * @return array
  */
 function metadata_fallback()
 {
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -415,6 +415,34 @@ function get_metadata($entry)
 }
 
 /**
+ * A hacky function that allows us to fall back to hard-coded metadata for special
+ * pages that don't support editor-specified metadata (e.g. campaign collections).
+ *
+ * @param array $metadata
+ * @param string $path
+ */
+function metadata_fallback()
+{
+    $HARDCODED_METADATA = [
+        'us/collections/corona-virus-campaigns' => [
+            'type' => 'article',
+            'url' => request()->url(),
+            'title' => 'Don\'t let COVID-19 stop you from changing the world.',
+            'description' => 'Join millions of DoSomething members using our resources to stay healthy, fight anxiety, and make a difference on the causes we care about, from mental health to climate change and beyond. Let’s Do This.',
+            'image' => [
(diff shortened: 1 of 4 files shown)


weerd:
> Please be nice! This is rushed together to help with a COVID-19 rapid response going out today.

😈

resources/views/partials/metadata.blade.php
```diff
@@ -1,26 +1,36 @@
-@if ($metadata['description'])
-    <meta name="description" content="{{ $metadata['description'] }}">
-@endif
+@php
+    // Hacky! This lets us hard-coded metadata for client-rendered pages.
+    // We'll be able to handle this way more gracefully in the future.
+    if (! isset($metadata)) {
+        $metadata = metadata_fallback();
+    }
+@endphp
```
weerd: Instead of wrapping it in `@php` couldn't we just add a `@else` below with this fallback?

app/helpers.php
```diff
@@ -414,6 +414,34 @@ function get_metadata($entry)
...
+ * A hacky function that allows us to fall back to hard-coded metadata for special
+ * pages that don't support editor-specified metadata (e.g. campaign collections).
+ *
+ * @param array $metadata
+ * @param string $path
+ */
+function metadata_fallback()
+{
+    $HARDCODED_METADATA = [
+        'us/collections/corona-virus-campaigns' => [
```
mendelB: I'd love it if we could in a follow-up add a default generic fallback for any other URL's, since currently, the homepage, and other collection/cause pages would still be a blank embed and it'd be cool for them to at least have DS logo, blanket description! re https://www.pivotaltracker.com/n/projects/2401401/stories/170287909

app/helpers.php
```diff
@@ -414,6 +414,34 @@ function get_metadata($entry)
     ];
 }
 
+/**
+ * A hacky function that allows us to fall back to hard-coded metadata for special
+ * pages that don't support editor-specified metadata (e.g. campaign collections).
+ *
+ * @param array $metadata
+ * @param string $path
```
mendelB: Oh I guess these params aren't used since we're running that logic within the function itself?

resources/views/partials/metadata.blade.php
DFurnes: Would we have to duplicate all the `<meta />` tags in that case? Sorry, having trouble visualizing.

app/helpers.php
DFurnes: Sure, totally!

app/helpers.php
DFurnes: Oh oops! Yep, this was from a prior attempt that got stymied by the undefined `$metadata` causing an exception. I'd assumed it would just let me pass it as `null`.

resources/views/partials/metadata.blade.php
weerd: Oh sorry! I misread it and was thinking of something else. Ignore me 😬

PR #1983, opened 2020-03-19
Tailwind helper

weerd:
### What's this PR do?

This pull request updates how we pull Tailwind configuration values within JS to use a new, simpler approach.

👀


resources/assets/helpers/index.js
```diff
@@ -918,6 +920,15 @@ export function getScholarshipAffiliateLabel() {
   return isScholarshipAffiliateReferral() ? utmLabel : null;
 }
 
+/**
+ * Get specified theme setting from the resolved Tailwind configuration object.
+ *
+ * @param {String} themeSetting
+ */
+export function tailwind(themeSetting) {
+  return get(resolveConfig(tailwindConfig).theme, themeSetting, null);
```
weerd: I was debating if `null` is the ideal default here 🤔 Or would an empty object or even throwing an error so it's obvious in the console? Lemme know your thoughts!

DFurnes: Mmm, that's an interesting thought. Yeah, maybe we should `console.error`?

DFurnes: What this all really makes me want is to have TypeScript types for these so we could trigger a build error if we tried to read a value that didn't exist... but that's for another day! 🤓

mendelB: bring. back. flow.

┻━┻︵ \(°□°)/ ︵ ┻━┻

resources/assets/helpers/index.js
```diff
@@ -918,6 +921,29 @@ export function getScholarshipAffiliateLabel() {
...
+
+  const setting = get(resolveConfig(tailwindConfig).theme, themeSetting, null);
+
+  if (!setting) {
+    console.error(
+      `The ${themeSetting} setting specified was not found in the Tailwind theme configuration.`,
+    );
+  }
+
+  return setting;
```
weerd: Ok! Here are the updates to help with errors. First the function will throw an error if the provided `themeSetting` is not a string (throws an error if not a string or if left empty).

Once it tries to retrieve the value from the config, if a specified setting was incorrect and/or not found it is defined as `null` and provides an error in the console, but still returns `null` as the value so that then in the code where it's used, `null` causes other problems and is specified in the console to make it easier to find which component went awry.

Does that seem cool with y'all @DFurnes @mendelB?

DFurnes: Yeah, looks great! 👍

mendelB: ➕

resources/assets/helpers/index.js
```diff
@@ -378,15 +381,7 @@ export function makeHash(string) {
...
-export function makeShareLink(
-  resource,
-  options: {
-    domain: string,
-    slug?: string,
-    key: string,
-    type: 'blocks' | 'modal',
-  },
-) {
+export function makeShareLink(resource, options = {}) {
```
weerd: Cypress got upset and we needed to get rid of the FB Flow syntax 😈 Speak of the devil eh, @mendelB

resources/assets/helpers/index.js
```diff
@@ -378,15 +381,7 @@ export function makeHash(string) {
  * @return {String}
  * @flow
```
DFurnes: @weerd You can remove the `@flow` annotation here now too.

weerd: Oh right! 

image

PR #1986, opened 2020-03-20
Cause/Interests Page!!

lindsayrmaher:
### What's this PR do?

This pull request adds the cause preferences section to our Interests tab in a users' profile! There is a small bit of styling love that needs to happen on the buttons and spacing for mobile, but since this is behind a feature flag I wanted to get the meatier portion of this into a PR while I work that out. 

Let me know if you have any specific thoughts about structure. I took a slightly different approach to where I'm querying for the user's cause array since I didn't feel that 12 loader spinner on the page was a great user experience. So i did that in the outer cause preference component.

Again, i know there are some CSS bugs to work out (see screenshots lol) BUT i think this is ready for some 👀 

Large Screen:
Screen Shot 2020-03-20 at 3 40 05 PM

Medium Screen:
Screen Shot 2020-03-20 at 3 43 36 PM

Small:
Screen Shot 2020-03-20 at 3 42 59 PM

### Relevant tickets

References [Pivotal # 171439769](https://www.pivotaltracker.com/story/show/171439769).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Interests/CausePreferenceItem.js
+++ b/resources/assets/components/pages/AccountPage/Interests/CausePreferenceItem.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
-// import classNames from 'classnames';
+import classNames from 'classnames';
 import { useMutation } from '@apollo/react-hooks';
 
 const CAUSE_PREFERENCE_MUTATION = gql`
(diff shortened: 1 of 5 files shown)


resources/assets/components/pages/AccountPage/Interests/CausePreferences.js
```diff
@@ -1,9 +1,103 @@
...
+
+  if (error) {
+    return <p>Something went wrong!</p>;
+  }
+
+  if (loading) {
+    return <div className="spinner align-center" />;
+  }
+
+  const { causes } = data.user;
```
DFurnes: My personal preference would be to move this into `CausePreferenceItem`, since it keeps the data dependency and where it's used closer (and therefore a little easier to reason about). Apollo is smart enough to realize all of these buttons will be looking for the same data & only make one network request.

resources/assets/components/pages/AccountPage/Interests/CausePreferenceItem.js
```diff
@@ -0,0 +1,70 @@
...
+        className={classNames(
+          'btn mx-4 mb-6 md:my-4 border border-solid border-blurple-500 focus:outline-none align-middle',
+          !causes.includes(cause)
+            ? 'bg-blurple-500 text-white hover:bg-blurple-300 focus:bg-blurple-500 focus:text-white'
+            : 'bg-white border text-blurple-500 border-solid hover:border-blurple-300 hover:text-blurple-200 focus:bg-white focus:text-blurple-500',
+        )}
+        onClick={() =>
+          updateInterest({
+            variables: {
+              cause,
```
DFurnes: Since `cause` is always consistent, it can be moved up alongside `userId`:

```js
const options = { variables: { cause, userId: window.AUTH.id } };
const [updateInterest] = useMutation(CAUSE_PREFERENCE_MUTATION, options);

// ...

<button
  onClick={() => updateInterest({ variables: { interested: !causes.includes(cause) }})}

```

resources/assets/components/pages/AccountPage/Interests/CausePreferences.js
lindsayrmaher: So i did originally have this in `CausePreferenceItem` but i don't want each item to show a load spinner when it's loading and was thinking this allows us to just show one as the user query is made initially. Do you have any thoughts as a work around with that? I'm all for keeping things concise, but i just couldn't think of a pretty way to handle that on each item.

DFurnes: In my opinion, it's also a nicer experience to show the user the full interface immediately (since we technically can, and that lets them start reading topics and descriptions right away) and only make them wait on whether or not the buttons are filled in for their user.

If showing a loading spinner on each button is too much, another idea might be to use (a variant of) the [PlaceholderText](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/utilities/PlaceholderText/PlaceholderText.js) component for the button while these are loading.

cc: @lkpttn, who might have other loading ideas!

lindsayrmaher: oooooOOoooOoo, I like that.

PR #1989, opened 2020-03-23
Homepage newsletter section

weerd:
### What's this PR do?

This pull request wraps up the new Homepage Newsletter section. I ended up extracting a new `NewsletterItem` component within the HomePage component so it would be easier to universally style the four items by just adjusting styles in a single component.

Responsive example:
Homepage Newsletter Section

👀 

### Relevant tickets

References [Pivotal #171844576](https://www.pivotaltracker.com/story/show/171844576).

### Checklist


resources/assets/components/pages/HomePage/NewHomePage.js
```diff
@@ -9,6 +9,10 @@ import SiteFooter from '../../utilities/SiteFooter/SiteFooter';
 import { contentfulImageUrl, tailwind } from '../../../helpers';
 import HomePageCampaignGallery from './HomePageCampaignGallery';
 import SiteNavigationContainer from '../../SiteNavigation/SiteNavigationContainer';
+import newsletterNewsImage from './Images/newsletter_news.svg';
+import newsletterCommunityImage from './Images/newsletter_community.svg';
+import newsletterLifestyleImage from './Images/newsletter_lifestyle.svg';
+import newsletterScholarshipsImage from './Images/newsletter_scholarships.svg';
```
weerd: I'll re-arrange these in line-length order, but originally I tried to keep them as a separate clump of imports since they were all SVGs. However, ESLint complained about the carriage return separating the imports. I could re-separate them and just have it ignore that rule in this file 🤷‍♂, or keep them mixed in with the other imports based on line-length. Thoughts 🤔 

I also understand this is small potatoes 🥔, but it's a crazy world out there right now and I need some semblance of order. 😅

mendelB: Maybe worth trying something similar to what we do with clumping together [the newsletter](https://github.com/DoSomething/phoenix-next/blob/355515da4ffe8f236cc116152c2a1492b3872a22/resources/assets/components/pages/AccountPage/Subscriptions/EmailSubscriptions.js#L4) and badges images? It feels like a nice way to keep things organized!

resources/assets/components/pages/HomePage/NewHomePage.js
```diff
@@ -46,6 +50,32 @@ const HOME_PAGE_QUERY = gql`
...
+    <h3 className="mb-2 text-white">{title}</h3>
+    <p className="mb-4 flex-grow text-white">{children}</p>
+    <a href={link.url} className="font-normal text-white underline">
+      {link.copy}
+    </a>
+  </div>
+);
+
+NewsletterItem.propTypes = {
+  children: PropTypes.string.isRequired,
```
mendelB: Could it be worth using an explicit e.g. `description` prop since this is just a string? Or is the idea to keep it open in case we need more actual nodes in the `children`?

resources/assets/components/pages/HomePage/NewHomePage.js
```diff
@@ -46,6 +50,32 @@ const HOME_PAGE_QUERY = gql`
...
+  image: PropTypes.string.isRequired,
+  link: PropTypes.object.isRequired,
```
mendelB: Maybe worth `PropTypes.shape`-ing this, since there's an explicitly required structure?

resources/assets/components/pages/HomePage/NewHomePage.js
```diff
@@ -178,22 +208,63 @@ const NewHomePageTemplate = ({ campaigns, title }) => {
...
                   Get Active.
                 </span>
               </h2>
-
               <p className="text-lg text-white">
                 Sign up for one of our newsletters.
               </p>
+
+              <ul className="grid grid-cols-1 md:grid-cols-2 lg:grid-cols-4 md:gap-16 mt-8">
+                <li className="text-white mb-8 lg:mb-0">
```
mendelB: I'm having trouble grokking the logic behind the disparate margins on these `li`s -- I imagine it's due to the responsive grid, and them appearing in multiple rows at times?

resources/assets/components/pages/HomePage/NewHomePage.js
DFurnes: It's totally fine in each of these scenarios, but I'd recommend against [this approach](https://github.com/DoSomething/phoenix-next/blob/355515da4ffe8f236cc116152c2a1492b3872a22/resources/assets/components/pages/AccountPage/Subscriptions/NewsletterImages.js#L6-L13) since it means these are "all or nothing" (so if we needed just one of the images in the future, we'd not be able to separate it from all the others when we build bundles for the page).

Instead, let's use named exports if we want to "group" these up:

```js
// NewsletterImages.js
export News from './newsletter_news.svg';
export Community from './newsletter_community.svg';
export Lifestyle from './newsletter_lifestyle.svg';
export Scholarship from './newsletter_scholarship.svg';

// NewHomePage.js
import NewsletterImages from './NewsletterImages';

<NewsletterItem title="News" image={NewsletterImages.News} />
<NewsletterItem title="Community" image={NewsletterImages.Community} />
<NewsletterItem title="Scholarships" image={NewsletterImages.Scholarship} />
```

resources/assets/components/pages/HomePage/NewHomePage.js
DFurnes: +1! I'd either define the shape, or separate these into separate `url` and `copy` props.

resources/assets/components/pages/HomePage/NewHomePage.js
DFurnes: Same question! Could we use CSS grid gap for this spacing?

resources/assets/components/pages/HomePage/NewHomePage.js
mendelB: Ohhh yes way smart!

weerd: Ah nice. Ok, I'll give that a whirl. Fine if I keep the `NewsletterImages.js` file within the HomePage component directory I take it? We can always move out if needing to use it elsewhere.

resources/assets/components/pages/HomePage/NewHomePage.js
weerd: Oh I meant to ask with this one. It seemed silly to me to define _everything_ as a prop for the component and this could leave it open, but I do have the `<p>` tag defined in the component w/ classes... so maybe a prop is a better route.

resources/assets/components/pages/HomePage/NewHomePage.js
weerd: Yep! My only explanation is brain fart 🤦‍♂ Defs a problem for CSS grid.

resources/assets/components/pages/HomePage/NewHomePage.js
weerd: I went ahead and changed it to a prop of `content`. Needing to have a conversation on it implied it was awkward, and highlighted it would work better as a prop 👌

resources/assets/components/pages/HomePage/NewHomePage.js
DFurnes: Sometimes it just takes a fresh set of eyes! It's easy to miss something when you've been deep in the code for a while (especially when it's went through a few iterations). 😸

resources/assets/components/pages/HomePage/NewHomePage.js
DFurnes: Yeah, that makes sense to me!

PR #1991, opened 2020-03-24
Updates to wrap up the homepage campaign featured item.

weerd:
### What's this PR do?

This pull request wraps up the new `CampaignGalleryFeaturedItem` component for showcasing a featured item in the campaign gallery on the new Homepage. It also does some additional cleanup and other small updates.

Homepage Campaign Featured Item

👀 on the generated [review app](https://dosomething-phoenix-de-pr-1991.herokuapp.com/us)!

There's some trickery going on with the image in this component when it hits the `xxl` size. Since the image ratio isn't the most ideal size for responsive adjustments as the viewport increases in size, to fit better when the layout changes to a side-by-side style, I ended up visually hiding the `` and using a `background-image` on the parent `<div>` to execute on the desired effect.

### Relevant tickets

References [Pivotal #171056375](https://www.pivotaltracker.com/story/show/171056375).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem.js
+++ b/resources/assets/components/utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem.js
@@ -39,6 +39,7 @@ const CampaignGalleryFeaturedItem = ({
               '826',
               'fill',
             )}');
+            background-position: center;
             background-repeat: no-repeat;
             background-size: cover;
           }
--- a/resources/assets/components/utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem.js
+++ b/resources/assets/components/utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem.js
@@ -1,7 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { css } from '@emotion/core';
 
-import { contentfulImageUrl } from '../../../../../helpers';
+import {
+  contentfulImageUrl,
+  contentfulImageSrcset,
+  tailwind,
+} from '../../../../../helpers';
 
 const CampaignGalleryFeaturedItem = ({
   showcaseDescription,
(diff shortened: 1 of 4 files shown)


resources/assets/components/utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem.js
```diff
@@ -1,35 +1,70 @@
...
+          }
+        `}
+      >
+        <img
+          alt={showcaseImage.description || `Cover photo for ${showcaseTitle}`}
+          className="xxl:invisible"
+          srcSet={srcset}
+          src={contentfulImageUrl(showcaseImage.url, '365', '205', 'fill')}
+        />
+      </div>
```
DFurnes: Could we handle this by using [`LazyImage`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/utilities/LazyImage.js) and a [`Media`](https://github.com/ReactTraining/react-media) tag to swap between "normal" and "background" modes? Thinking something like this (which also makes me think we'll want to add `srcSet` support to LazyImage...):

```jsx
<Media queries={{ xxl: … }}>
    <LazyImage
      src={contentfulImageUrl(showcaseImage.url, '365', '205', 'fill')}
      srcSet={srcset}
      asBackground={matches.xxl}
    />
</Media>
```

DFurnes: Anyways, not something that I think makes sense to tackle now but let's mull on it!

resources/assets/components/utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem.js
```diff
@@ -1,35 +1,70 @@
...
+        css={css`
+          @media (min-width: ${tailwindScreens.xxl}) {
+            background-image: url('${contentfulImageUrl(
+              showcaseImage.url,
+              '1468',
+              '826',
+              'fill',
+            )}');
+            background-repeat: no-repeat;
+            background-size: cover;
```
DFurnes: Let's add a `background-position: center` here so these don't end up off-center?

resources/assets/components/utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem.js
weerd: 🤔 potentially, but might be worth holding off until we see this pattern become a need in multiple places. Is the concern regarding having it all be collected in a single component rather than defined a bit separated within the `` and `<div>`?

DFurnes: This is something that the `LazyImage` and `Media` components should already able to handle (sans `srcSet` support), so it'd be just be nice to not have to build it from scratch here!

PR #1995, opened 2020-03-26
Feature add more to social share toolkit  voter registration

Dinnall:
### What's this PR do?

This pull request adds social shares buttons to the social share toolkit for Online Voter Registration Drives

1. A feature flag was added to this PR. Turn `DS_ENABLE_SNAPCHAT` in Heroku to true. 
2. Your Snapchat username will need to be added to the Snapchat Dev kit.
3. Open the link provided in Heroku on your phone. 
4. You will be redirected to Snapchat, there you will see the URL for the user running the voter reg drive.
5. You can send it to your contacts or attach it to your story.

Important: The Snapchat button can only we viewed on a mobile device.
...

### Relevant tickets

References [Pivotal #171938213](https://www.pivotaltracker.com/n/projects/2417735/stories/171938213).

### Checklist

- [ x] This PR has been added to the relevant Pivotal card.
- [ x] Added one social media objects to tailwinds.
- [ X] Added screenshots of front-end changes on small, medium, and large screens.
Screen Shot 2020-03-25 at 8 18 52 PM
Screen Shot 2020-04-22 at 5 42 47 PM

image
image


resources/assets/helpers/index.js
```diff
@@ -777,6 +807,41 @@ export function handleFacebookShareClick(href, trackingData) {
...
+ * @param {Object} trackingData
+ * @param {String} quote
+ */
+export function handleInstagramShareClick(href, trackingData, quote = '') {
+  trackAnalyticsEvent('clicked_share_twitter', {
+    action: 'button_clicked',
+    category: EVENT_CATEGORIES.socialShare,
+    label: 'twitter',
+    context: { ...trackingData, url: href },
+  });
```
katiecrane: The arguments here need to be updated to be instagram instead of twitter!

resources/assets/helpers/index.js
```diff
@@ -777,6 +807,41 @@ export function handleFacebookShareClick(href, trackingData) {
...
+ * @param {Object} trackingData
+ * @param {String} quote
+ */
+export function handleSnapchatShareClick(href, trackingData, quote = '') {
+  trackAnalyticsEvent('clicked_share_twitter', {
+    action: 'button_clicked',
+    category: EVENT_CATEGORIES.socialShare,
+    label: 'twitter',
+    context: { ...trackingData, url: href },
+  });
```
katiecrane: And this one should be snapchat instead of twitter

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
```diff
@@ -123,7 +127,6 @@ class SocialShareTray extends React.Component {
   render() {
     const { shareLink, platforms, responsive, title } = this.props;
     const trackLink = this.props.trackLink || this.props.shareLink;
-
```
aaronschachter: Was this intentional? We often leave a blank line before returning the component - let's add this back in.

resources/assets/helpers/analytics.js
```diff
@@ -251,7 +251,6 @@ export function trackAnalyticsPageView(history) {
  */
 export function trackAnalyticsEvent(name, metadata = {}) {
   const { action, category, label, context = {}, service } = metadata;
-
```
aaronschachter: We also tend to leave an empty line after defining a `const` - let's add this back in too.

aaronschachter:
I think we'll be talking through this in our design meeting today.. not sure if we have designs yet but the desktop view looks odd with 4 icons in the first row and 2 in the second row.

aaronschachter:
> the desktop view looks odd with 4 icons in the first row and 2 in the second row.

My mistake, didn't realize is the intended design!

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
Dinnall: Added the space back before returning the component

resources/assets/helpers/analytics.js
Dinnall: Added a line after declaring metadata const

resources/assets/helpers/index.js
Dinnall: updated this change and removed Instagram

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
```diff
@@ -140,7 +157,23 @@ class SocialShareTray extends React.Component {
               text="Share"
             />
           ) : null}
-
+          {platforms.includes('snapchat') &&
+          /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
+            navigator.userAgent,
```
aaronschachter, line 162: This logic should live in a helper vs inline. I'm wondering if we could use Tailwind to display the button for certian breakpoints and hide from others, to avoid needing this check altogether?

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
```diff
@@ -29,6 +32,19 @@ class SocialShareTray extends React.Component {
     loadFacebookSDK();
   }
 
+  /**
+   * Check to see if the sharedLink is present within
+   * a bertly link. This is done because the snapchatSDK needs a link to provide to the user that opened the link
+   * if no link is provided it will pass the current page URL as our attachmenturl
+   *
+   */
+
```
aaronschachter, line 41: let's remove this blank line.

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
```diff
@@ -29,6 +32,19 @@ class SocialShareTray extends React.Component {
     loadFacebookSDK();
   }
 
+  /**
+   * Check to see if the sharedLink is present within
+   * a bertly link. This is done because the snapchatSDK needs a link to provide to the user that opened the link
+   * if no link is provided it will pass the current page URL as our attachmenturl
```
aaronschachter, line 38: I'm having trouble following this. Is this correct? If we previously didn't a link and now we have it, we load the SDK.

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
```diff
@@ -191,7 +224,14 @@ SocialShareTray.propTypes = {
 SocialShareTray.defaultProps = {
   shareLink: null,
   trackLink: null,
-  platforms: ['facebook', 'twitter', 'messenger', 'email'],
+  platforms: [
+    'facebook',
+    'instagram',
```
aaronschachter, line 229: Remove `instagram`?

resources/assets/helpers/index.js
```diff
@@ -776,6 +795,25 @@ export function handleFacebookShareClick(href, trackingData) {
...
+/**
+ * Handle click event from a Snapchat share button
+ *
+ * @param {String} href
+ * @param {Object} trackingData
+ *
+ */
+export function handleSnapchatShareClick(href, trackingData) {
+  console.log('href:', href);
+  console.log('trackingData:', trackingData);
```
aaronschachter, line 807: Let's remove this console logging. There's analytics logging available -- [docs here](https://dosomething.gitbook.io/phoenix-documentation/data-+-performance/analytics#day-to-day).

tailwind.config.js
```diff
@@ -2,7 +2,165 @@ const variables = require('./tailwind.variables');
 
 module.exports = {
   theme: {
-    ...variables,
```
aaronschachter, line 5: What is the context for all of these changes? Why are we removing `variables` here?

weerd: @Dinnall should rebase this PR with `master`. @DFurnes extracted these in #2049 because of a bug in Edge 18.

DFurnes: It looks like this is rebased, I think, since I see the `require('./tailwind.variables')` and extracted file on this branch. We can just remove the duplicated configuration lines below & keep `...variables` intact!

Dinnall: I did include the `require('./tailwind.variables')`, but I fixed the changes manually on Github, then pushed more changes. I should have fixed them locally instead.

resources/assets/helpers/index.js
Dinnall: Missed these. Removed now.

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
Dinnall: This was from the initial PR before the priorities of the ticket were reworded.
Removed it now

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
Dinnall: removed

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
Dinnall: I moved this logic into a helper function and wrapped the display in a featureflag last night. 
I think a helper function is good, it does seem like something we can use in the codebase.

aaronschachter: @weerd Is there anywhere in the codebase where we only conditionally show an element for mobile devices? Am I correct in thinking Tailwind would be a way to avoid checking user agent (or should we extract this into a helper if need it)?

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
mendelB: This one's tricky, but the `loadSnapchatSDK();` crawls the DOM for buttons with the Snapchat class, and uses their `data-share-url` as the share URL, but defaults to the current URL if there is no `data-share-url` assigned, so we need to wait for the `shareUrl` prop which loads asynchronously via the Phoenix API in the `SocialDriveAction` to be passed in for this to work. 

I don't think this is the very best way we can do this, but at least here, we'll be checking for the presumed 'second' render, when the previous shareLink was the default `''`, and it's now populated with the Bertly/default URL.

But we should also add a `platforms.includes('snapchat')` conditional so we don't run this for naught.

One potential improvement is to imitate @DFurnes refactor to [the `SoftEdgeBlock`](https://github.com/DoSomething/phoenix-next/blob/0c3d8aebbd52e1330d1f93827ac480fad7dd1e22/resources/assets/components/actions/SoftEdgeBlock.js#L45-L46) where we use a specialized hook to load the script?

I also wonder if we could manipulate [the logic](https://github.com/DoSomething/phoenix-next/pull/1995/files#diff-927b208f806cb78d14c0f0587928cb8bR625) we assign to fetch the Snapchat button in the script to somehow work better with our flow here 🤔

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
mendelB: An instance I can think of is where we use the `Media` component [in `IframeEmbed`s](https://github.com/DoSomething/phoenix-next/blob/0c3d8aebbd52e1330d1f93827ac480fad7dd1e22/resources/assets/components/utilities/IframeEmbed/IframeEmbed.js#L19-L34) to render a preview image instead of the iframe embed for mobile views.

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
Dinnall: I've played around with `useScript ` before, it could work here, might be able to erase the `componentDidUpdate` altogether.

mendelB: @Dinnall That would be neat!

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
aaronschachter: Do you need to push up your changes? This is still listing `instagram` as an array item. Same question around the ask for removing the console logging.

Dinnall: I just pushed the changes, but I'm making more changes to refactor `componentDidUpdate ` to use `useScript `

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
aaronschachter: Thanks @mendelB , this looks like it'd be sweet! I suppose not necessary for merge, but I do think it'd preferable to re-use the `Media` query to keep things consistent vs adding the new code for the  helper function...

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
Dinnall: @aaronschachter I pushed new changes. I included the `Media` component to display the button only on mobile. `useScript `wasn't included because of a custom `snapKitInit ` function they require in custom buttons. If it was just the src it would be much easier. Let me know your thoughts. I rebase the tailwind file that was causing the error as well

resources/assets/helpers/analytics.js
aaronschachter: This removed line seems to have re-appeared..

resources/assets/components/utilities/SocialShareTray/SocialShareTray.js
aaronschachter: This removal also seems to have re-appeared

resources/assets/helpers/index.js
```diff
@@ -611,6 +611,24 @@ export function loadFacebookSDK() {
...
+ */
+export function loadSnapchatSDK() {
+  return new Promise(resolve => {
+    const script = document.createElement('script');
+    script.id = 'snapkit-creative-kit-sdk';
+    script.src = 'https://sdk.snapkit.com/js/v1/create.js';
+    document.head.append(script);
+
+    window.snapKitInit = function() {
+      snap.creativekit.initalizeShareButtons(
```
aaronschachter, line 625: One workaround to get our StyleCi tests to pass could be to disable the `no-undef` linting rule here:
```
snap.creativekit.initalizeShareButtons( // eslint-disable-line no-undef
```

resources/assets/helpers/index.js
```diff
@@ -611,6 +611,24 @@ export function loadFacebookSDK() {
   });
 }
 
+/**
+ * Load and return the Snapchat SDK.
+ */
+export function loadSnapchatSDK() {
+  return new Promise(resolve => {
```
aaronschachter, line 618: Per [StyleCi failure](https://circleci.com/gh/DoSomething/phoenix-next/2896?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) this ` no-unused-vars` error can be fixed be 
```
return new Promise(() => {
 ```

File: resources/assets/helpers/index.js [hunk 13]
```diff
@@ -1012,3 +1046,19 @@ export function buildVoterRegUrl(source = 'web', sourceDetails, url) {
 
   return `${baseUrl}?r=${userId}source:${source},source_details:${sourceDetails}`;
 }
+
+/**
+ * Check to see if user is on desktop or mobile.
+ *
+ * @return {Boolean}
+ */
+export function isUserOnDesktop() {
```
aaronschachter, line 1055: We don't this function anymore now that we're using a `Media` component to display the button, right?

aaronschachter:
Continued over in #2058.

PR #1978, opened 2020-03-18
Feature total voter registration

Dinnall:
### What's this PR do?

This pull request...
Updated pull request 
### How should this be reviewed?
added feature flagging 
...

### Any background context you want to provide?
Added feature flagging with the suggestion from the previous comments 
...

### Relevant tickets

References [Pivotal #171707739](https://www.pivotaltracker.com/n/projects/2417735/stories/171707739/comments/212250543).

PR #1998, opened 2020-03-26
Query GraphQL for count of voter registration referrals

aaronschachter:
### What's this PR do?

This pull request adds a GraphQL query to find the total number of voter registration referrals for the logged in user, replacing the placeholder value added in #1978.

The only trouble is that we need the posts count query to filter by status, which it currently doesn't support. We only want to count the total number of completed voter registration posts -- which is why this still remains feature flagged and should be committed with a big ol' `@TODO`

Our review app should be created with `DS_ENABLE_VOTER_REG_DRIVE_TOTAL` set to true, so the total will be visible -- but it will be 0 across the board unless we manually edit the `referrer_user_id` in posts to your logged in user... which, is not ideal. I'll aim to add Cypress tests for this once GraphQL / Rogue is updated to support filtering by status - as we're hoping to launch this Monday.

Created a `VoterRegistrationStats` component to remove all of this voter registration specific logic out of the `SocialActionDrive` component.

### Relevant tickets

References [Pivotal #170775705](https://www.pivotaltracker.com/story/show/170775705).

### Checklist

### DIFF
--- a/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
+++ b/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
@@ -9,6 +9,7 @@ import linkIcon from './linkIcon.svg';
 import Card from '../../utilities/Card/Card';
 import Embed from '../../utilities/Embed/Embed';
 import { postRequest } from '../../../helpers/api';
+import VoterRegistrationStats from './VoterRegistrationStats';
 import { dynamicString, withoutTokens, featureFlag } from '../../../helpers';
 import SocialShareTray from '../../utilities/SocialShareTray/SocialShareTray';
 import {
@@ -69,6 +70,7 @@ class SocialDriveAction extends React.Component {
       shareCardDescription,
       shareCardTitle,
       hidePageViews,
+      userId,
     } = this.props;
     const { shortenedLink } = this.state;
     return (
@@ -127,42 +129,10 @@ class SocialDriveAction extends React.Component {
         </div>
 
         {!hidePageViews ? (
-          <div
-            className={classNames('social-drive-information mt-6', {
-              'lg:w-1/3 lg:pl-3 lg:mt-0': !hidePageViews,
-            })}
-          >
-            <Card className="bordered rounded" title="More info">
-              <div className="link-info p-3">
-                <p className="info__title">What happens next?</p>
-
-                <p className="info__text">
-                  As you share your voter registration page, we&#39;ll keep
-                  track of how many people you bring in. Check back often and
-                  try to get as many views as possible!
-                </p>
-              </div>
-
-              <div className="p-3 page-views">
-                <span className="page-views__text uppercase">
-                  total page views
-                </span>
-                <h1 className="page-views__amount">
-                  {shortenedLink ? this.state.count : '?'}
-                </h1>
-              </div>
-              {featureFlag('voter_reg_drive_total') ? (
-                <div className="p-3 voter-registrations">
-                  <span className="voter-registrations__text uppercase">
-                    total voter registrations
-                  </span>
-                  <h1 className="voter-registrations__amount">
-                    {shortenedLink ? this.state.count : '?'}
-                  </h1>
-                </div>
-              ) : null}
-            </Card>
-          </div>
+          <VoterRegistrationStats
+            userId={userId}
+            pageViewsCount={this.state.count}
+          />
         ) : null}
       </div>
     );
--- a/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
+++ b/resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
@@ -10,7 +10,7 @@ import Card from '../../utilities/Card/Card';
 import Embed from '../../utilities/Embed/Embed';
 import { postRequest } from '../../../helpers/api';
 import VoterRegistrationStats from './VoterRegistrationStats';
-import { dynamicString, withoutTokens, featureFlag } from '../../../helpers';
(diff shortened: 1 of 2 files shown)


resources/assets/components/actions/SocialDriveAction/VoterRegistrationStats.js
```diff
@@ -0,0 +1,64 @@
...
+
+import { featureFlag } from '../../../helpers';
+import Query from '../../Query';
+import Card from '../../utilities/Card/Card';
+
+// @TODO: This query needs to filter by completed voter registrations before we go live.
+
+const USER_VOTER_REGISTRATION_REFERRAL_COUNT_QUERY = gql`
+  query UserVoterRegistrationReferralCount($userId: String!) {
+    postsCount(referrerUserId: $userId, type: "voter-reg", limit: 50)
```
mendelB: Would the `limit: 50` be limiting the returned referral total and displaying a capped referral count?

aaronschachter: It would.. I wasn't sure what exactly I should set here. If we don't send a limit over, [it defaults to 20 ](https://github.com/DoSomething/graphql/blob/9ac29efbee35bb5d8dd4bc3a16e7fdd1334f6578/src/repositories/rogue.js#L715)- which isn't ideal. I'm assuming this in place for performance reasons...

@DFurnes do you an idea for the value we should send here? 

And while I'm summoning you, in order to query for the status -- I'd guess we should add a new compound index in `referrer_user_id`, `type` and `status` to the Rogue `posts` table? I added `referrer_user_id` and `type`... not sure if it'd be ideal to drop that index and create a new one, or just the additional one...

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -127,42 +129,10 @@ class SocialDriveAction extends React.Component {
...
-                  <h1 className="voter-registrations__amount">
-                    {shortenedLink ? this.state.count : '?'}
-                  </h1>
-                </div>
-              ) : null}
-            </Card>
-          </div>
+          <VoterRegistrationStats
+            userId={userId}
+            pageViewsCount={this.state.count}
```
mendelB: I think we'd be losing our 'loading state' be only passing through the `this.state.count` which would initially be `undefined`. Maybe worth initializing this prop in the constructor as `?` (it's prior loading state which I suppose accounts for both loading and error)? Or some improved UI?

aaronschachter: I think this is handled by the `VoterRegistrationStatus.defaultProps`.

mendelB: Oh snap, my bad -- I missed that!

resources/assets/components/actions/SocialDriveAction/VoterRegistrationStats.js
mendelB: If we're maxed out at 50 until we figure things out on the Rogue end, perhaps, for now, we get away with `50+` in that scenario? (Not sure that it's worth our trouble, but I'd venture to say if we had a referral whale, they'd deserve that effort 😅)

aaronschachter: Ya I had the same thought -- I figured no matter what the cap is, even if we use something like 500 -- we'd still want to add that + sign. So maybe our limit should be 51, or 501 -- whatever our sweet spot is, in order to add the +

aaronschachter: I'll add that + in once I come back to add the filter for `status` - going to merge this for now. Thanks for reviewing so quickly!

mendelB: I think that makes sense! Then we hold our breath and wait for our members to prove us wrong and surpass that cap 🙏

DFurnes: Ahh yeah – we never got around to improving this did we! Since the way we "count" these things is by literally loading the whole response from Rogue and getting the `.length` of the returned array, we have a hard cap of 100. (And even that may get a little slow, so agreed that 50 is smart for now!)

If we did want to have a "real" count here, we'd want to add a new `post-counts` endpoint in Rogue that does a `SELECT COUNT(*)` query for the given filters and just returns that number for this query.

DFurnes: > And while I'm summoning you, in order to modify this query to include the post status -- I'd guess we should add a new compound index in referrer_user_id, type and status to the Rogue posts table? I added referrer_user_id and type... not sure if it'd be ideal to drop that index and create a new one, or just the additional one...

Yeah, that makes sense to me. Are there cases where we'd make this same query _without_ status? If so, we might be able to get away with the current query (since chopping it down to the particular referrer & type is probably the "heaviest" lifting filter-wise)...

aaronschachter: > Are there cases where we'd make this same query without status?

I could imagine a vLater where you could see the progress of each of your referrals, to know that X number of friends are still in Step 1 status, etc.

I think for v1 - let's keep the cap at 51, and show "50+" if our count is 51. I'll check with product whether we need to build support for the real count, but let's plan to launch without it for now.

DFurnes: Makes perfect sense to me! 👍

PR #1975, opened 2020-03-18
total voter registration display

Dinnall:
Added another card component to display total voter registration.

### What's this PR do?
This pull request further progresses on the voter registration campaign.
Added total voter registration for campaign alphas.

### How should this be reviewed?
👀
...

### Any background context you want to provide?
One note, the component isn't complete. Currently, the number being displayed isn't accurate yet.
This is just the placeholder of a bigger ticket. Will circle back to once another thicket is complete to pull data from.
...

### Relevant tickets

References [Pivotal #171707739](https://www.pivotaltracker.com/story/show/171707739).

### Checklist

- [X ] This PR has been added to the relevant Pivotal card.

Screen Shot 2020-03-12 at 2 59 27 PM

PR #2000, opened 2020-03-27
Filter voter registration referrals by status + documentation

aaronschachter:
### What's this PR do?

This pull request fixes the TODO in #1998 to filter the referrals count by completed statuses, and removes the `voter_reg_drive_total` feature flag added in #1975 - as this is ready to go live.

Also adds test coverage and documentation.

👀 

* Yes, background context on why we're limiting the count results to 51 can be found over in https://github.com/DoSomething/phoenix-next/pull/1998#discussion_r398831782

* I had to manually edit the GraphQL `schema.json` to include changes in to test the voter registration counts -- left a note in the docs about how we need to remove the duplicate `UserPostsQuery` in order update it via `apollo client:download-schema`, which would be preferred.

### Relevant tickets

Finally closes [Pivotal #170775705](https://www.pivotaltracker.com/story/show/170775705)!

### Checklist


aaronschachter:
Oh, great question! I should have added that for background context too. This was @DFurnes 's awesome idea from a [Slack thread yesterday](https://dosomething.slack.com/archives/CUQMU4Q6B/p1585254450024700?thread_ts=1585253703.024000&cid=CUQMU4Q6B).

Because this PR only requires testing the behavior of a `SocialActionDrive` block with its `hidePageViews` is set to false , we can create said block in Contentful and visit its BlockPage at `us/blocks/:id`. This avoids needing to create a Contentful campaign that references the block and copying its `window.STATE` into  `cypress/fixtures` - which is what I did for the CurrentSchoolBlock tests:
https://github.com/DoSomething/phoenix-next/blob/7b0931ad2d79e1f96b37feb0488e95a826fc1e75/cypress/fixtures/contentful/exampleCurrentSchoolBlockWithAction.js#L62-L67

in order to visit the campaign and run tests:
https://github.com/DoSomething/phoenix-next/blob/7b0931ad2d79e1f96b37feb0488e95a826fc1e75/cypress/integration/current-school-block.js#L45-L53

So a future PR for the Current School Block can eliminate those fixtures by just testing the Current School Block's permalink at `us/blocks/6Kj1Av7MnvNcCXLYHLGee0` instead of the campaign action page URL (`us/campaigns/test-school-finder-campaign/action`)

PR #2003, opened 2020-03-30
Homepage staff pick feature

weerd:
### What's this PR do?

This pull request does some last cleanup to utilize the new fields accessible after the GraphQL updates in https://github.com/DoSomething/graphql/pull/211. Campaigns that are designated as `staffPick` in Contentful will now properly show a "featured" badge in the top-right corner when presented in the new home page campaign gallery (note, this does not apply to `StoryPage` style campaigns, since they do not have a field for it; we've created a [ticket](https://www.pivotaltracker.com/story/show/172031799) on the Platforms team to address this during a cleanup sprint).

This also updates the new home page to use the new `coverImage` field in the `homePage` content type on Contentful, so the impact header background can be changed by content managers instead of having it be hardcoded.

Lastly, this PR also adds some nice class name hooks for use in our upcoming updates to our Ghost Inspector test for the home page.

👀 

### Relevant tickets

References [Pivotal #171336897](https://www.pivotaltracker.com/story/show/171336897).

### Checklist

### DIFF
--- a/resources/assets/components/pages/HomePage/NewHomePage.js
+++ b/resources/assets/components/pages/HomePage/NewHomePage.js
@@ -114,7 +114,7 @@ const NewHomePageTemplate = ({ articles, campaigns, coverImage, title }) => {
     }
   `;
 
-  const headerBackgroundStyles = !coverImage
+  const headerBackgroundStyles = coverImage
     ? css`
         background-image: url(${contentfulImageUrl(
           coverImage.url,
--- a/resources/assets/components/pages/HomePage/NewHomePage.js
+++ b/resources/assets/components/pages/HomePage/NewHomePage.js
@@ -407,11 +407,12 @@ NewHomePageTemplate.propTypes = {
   campaigns: PropTypes.arrayOf(PropTypes.object).isRequired,
   coverImage: PropTypes.shape({
     url: PropTypes.string.isRequired,
-  }).isRequired,
+  }),
   title: PropTypes.string,
 };
 
 NewHomePageTemplate.defaultProps = {
+  coverImage: null,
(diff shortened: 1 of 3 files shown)


resources/assets/components/pages/HomePage/NewHomePage.js
```diff
@@ -110,30 +114,56 @@ const NewHomePageTemplate = ({ articles, campaigns, title }) => {
...
+    : null;
+
   return (
     <Fragment>
       {/* @TODO: Once EmotionJS supports shorthand syntax for React.Fragment, switch <Fragment> out for <> syntax! */}
       <SiteNavigationContainer />
 
       <main>
-        <article>
+        <article className="home-page">
```
DFurnes: Can we use something like `data-test=home-page` to make it clear this is used for finding this element in test suites (and not applying any styling or behavior)?

resources/assets/components/pages/HomePage/NewHomePage.js
```diff
@@ -28,6 +31,7 @@ const HOME_PAGE_QUERY = gql`
         }
         ... on CampaignWebsite {
           id
+          staffPick
           url
         }
         ... on StoryPageWebsite {
```
DFurnes: For later – let's pull this out into a fragment that's inline with the actual component where these props are used in `HomePageCampaignGallery` (so there's not that extra level of indirection between the query & it's usage).

resources/assets/components/pages/HomePage/NewHomePage.js
weerd: oh right! I forgot we decided on that 👍

PR #2008, opened 2020-04-01
Updates to the base grid vertical padding.

weerd:
### What's this PR do?

This pull request removes the vertical padding on the `base-12-grid` class to allow containers to define their own vertical padding based on mocks.

Prior to this update, overriding the predefined vertical padding on containers with the `base-12-grid` class required  having to add a dedicated stylesheet with more specific declarations since Tailwind classes do not have high enough specificity to override the vertical padding styles.

👀 

### Relevant tickets

References [Pivotal #172090293](https://www.pivotaltracker.com/story/show/172090293).

### Checklist

### DIFF
--- a/resources/assets/components/pages/CampaignPage/CampaignPageContent.js
+++ b/resources/assets/components/pages/CampaignPage/CampaignPageContent.js
@@ -28,7 +28,7 @@ const CampaignPageContent = props => {
     <div className="campaign-page__content" id={subPage.id}>
       <ScrollConcierge />
       {content ? (
-        <div className="base-12-grid">
+        <div className="base-12-grid py-3 md:py-6">
           <div className="grid-wide-7/10">
             <TextContent className="mx-3">{content}</TextContent>
           </div>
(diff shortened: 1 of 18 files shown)


resources/assets/components/pages/CampaignPage/CampaignPageContent.js
```diff
@@ -44,23 +44,25 @@ const CampaignPageContent = props => {
...
+                ContentBlock: 'grid-wide',
+                ImagesBlock: 'grid-wide',
+                PostGalleryBlock: 'grid-wide',
+                PhotoSubmissionBlock: 'grid-wide',
+                SocialDriveBlock: 'grid-wide',
+              }}
+            />
+          ))}
+        </div>
+      ) : null}
```
weerd, line 65: Added this condition cause it was outputting an empty `<div>` that took up space when `blocks` was an empty array.

weerd:
`py-6` partay! 💃 🕺

PR #2009, opened 2020-04-01
Re-triggering scroll after campaign signup

lindsayrmaher:
### What's this PR do?

This pull request fixes a bug on our campaign pages where the `Scroll Concierge` was no longer working as expected. After users signed up and saw a confirmation modal, they were not pulled down to the action page.

Does this seem the most efficient? Also wondering if there's anywhere else this needs to be passed in on the `CampaignRoute` page. 

Kapture 2020-04-01 at 14 45 33

### Relevant tickets

References [Pivotal # 170979245](https://www.pivotaltracker.com/story/show/170979245).

### Checklist

### DIFF
--- a/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
+++ b/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
@@ -106,7 +106,12 @@ const CampaignRoute = props => {
 
             if (isClosed) {
               if (slug === 'community' && hasCommunityPage) {
-                return <CampaignPageContainer {...routeProps} />;
+                return (
+                  <CampaignPageContainer
+                    {...routeProps}
+                    shouldShowAffirmation={shouldShowAffirmation}
+                  />
+                );
               }
 
               return (
@@ -130,7 +135,12 @@ const CampaignRoute = props => {
               );
             }
 
-            return <CampaignPageContainer {...routeProps} />;
+            return (
+              <CampaignPageContainer
+                {...routeProps}
+                shouldShowAffirmation={shouldShowAffirmation}
+              />
+            );
           }}
         />
       </Switch>
--- a/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
+++ b/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
@@ -106,12 +106,7 @@ const CampaignRoute = props => {
 
             if (isClosed) {
               if (slug === 'community' && hasCommunityPage) {
-                return (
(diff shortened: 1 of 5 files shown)


resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
```diff
@@ -106,7 +106,12 @@ const CampaignRoute = props => {
 
             if (isClosed) {
               if (slug === 'community' && hasCommunityPage) {
-                return <CampaignPageContainer {...routeProps} />;
+                return (
+                  <CampaignPageContainer
+                    {...routeProps}
+                    shouldShowAffirmation={shouldShowAffirmation}
```
mendelB, line 112: Since the `shouldShowAffirmation` value comes from our redux store via [the container](https://github.com/DoSomething/phoenix-next/blob/b849a6e472ed7a322227ea740a71c05d2510e363/resources/assets/components/Campaign/CampaignRoute/CampaignRouteContainer.js#L21) anyways, it might be _slightly_ neater to fetch this directly again from the `CampaignPageContainer` itself instead of manually passing it down the chain.

resources/assets/components/pages/CampaignPage/CampaignPageContent.js
```diff
@@ -26,7 +26,7 @@ const CampaignPageContent = props => {
 
   return (
     <div className="campaign-page__content" id={subPage.id}>
-      <ScrollConcierge />
+      <ScrollConcierge trigger={!shouldShowAffirmation} />
```
mendelB, line 29: Do we need to use the `!` here? I think the `trigger` will just affect if the value changes, regardless of what it is initially. (But please do confirm!)

lindsayrmaher: oh that's a fair point, i think i did that so it read more clearly to me lol. as in, when the modal is closed, this should happen.

mendelB: Yeah that's a really good point! That is generally the logical flow for when it would _actually_ take effect. My other brain's rejoinder to that was 'oh but maybe that sends the wrong message as to the _functionality_ of the trigger?' 

I guess it'd make sense to keep it at whichever way you feel better about!

resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
lindsayrmaher: I like this better! Updated on dc775eb

PR #2010, opened 2020-04-01
Homepage tweaks

weerd:
### What's this PR do?

This pull request tackles a few design QA tweaks for the new Homepage design.

👀 

### Relevant tickets

References [Pivotal #172084017](https://www.pivotaltracker.com/story/show/172084017).

### Checklist


lkpttn:
There's a small issue with button focus states still taking the old blue color instead of blurple, but we might get the go ahead to change the global button style instead of having to override it in every individual component.

lkpttn:
I think we're also gonna see a push for card and secondary button styles to be 2px instead of 1, I'll update the mocks

PR #2012, opened 2020-04-02
Design & behavior related tweaks.

weerd:
### What's this PR do?

This pull request has a few updates to help fix up some design/behavior tweaks to the new Home Page.

Notably, it handles adding links to card images and titles universally and only showing the Join CTA banner when the user is anonymous.

👀 

### Relevant tickets

References [Pivotal #172127924](https://www.pivotaltracker.com/story/show/172127924).
References [Pivotal #172127969](https://www.pivotaltracker.com/story/show/172127969).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem.js
+++ b/resources/assets/components/utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem.js
@@ -45,12 +45,16 @@ const CampaignGalleryFeaturedItem = ({
           }
         `}
       >
-        <img
-          alt={showcaseImage.description || `Cover photo for ${showcaseTitle}`}
-          className="xxl:invisible"
-          srcSet={srcset}
-          src={contentfulImageUrl(showcaseImage.url, '365', '205', 'fill')}
-        />
+        <a className="block" href={url}>
+          <img
+            alt={
+              showcaseImage.description || `Cover photo for ${showcaseTitle}`
+            }
+            className="xxl:invisible"
+            srcSet={srcset}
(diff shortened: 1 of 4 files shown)


resources/assets/components/utilities/Gallery/templates/CampaignGalleryFeaturedItem/CampaignGalleryFeaturedItem.js
```diff
@@ -45,12 +45,16 @@ const CampaignGalleryFeaturedItem = ({
           }
         `}
       >
-        <img
-          alt={showcaseImage.description || `Cover photo for ${showcaseTitle}`}
-          className="xxl:invisible"
-          srcSet={srcset}
-          src={contentfulImageUrl(showcaseImage.url, '365', '205', 'fill')}
-        />
+        <a className="block" href={url}>
```
mendelB, line 48: Why do we need to set the `display: block` to these?

weerd: `<a>` are by default inline elements. It's a recommended preference I have to specify them as blocks to avoid weird little bugs from past experience along with if you add any other styles that affect layout to the `<a>`.

If you inspect the element and hover over it, it also helps to show what the `<a>` is surrounding and actually shows it versus if not specified as a block:

No block class and hovering over element in inspector:
Anchor with no block class

Block class and hovering over element in inspector:
Anchor with block class

mendelB: Ahhh, good to know, thanks for clarifying!!

PR #2014, opened 2020-04-03
Volunteer Credits Query and Data Parsing

mendelB:
### What's this PR do?

This pull request adds a new component to query the Volunteer Credit specific posts via GraphQL, and parse out the necessary data per row for the Volunteer Credits table.

You can commit-by-commit this one. The meat of the thing is in https://github.com/DoSomething/phoenix-next/commit/83cd6176aca4fa09f7f5dc6e81e86ba9f93faa41 whence we introduce the beefy GraphQL query to obtain the specific posts we need and then run a series of step to parse out the specific post data per certificate row. I tried to break it up logically and clearly, but please do 👀 that part more carefully and call out anything gnarly!

Is the GraphQL query _too_ much? We're pulling some pretty nested bits of data here (post->action->rogue campaign->contentful campaign). I guess we could hypothetically embed some of the bits in their respective components (e.g. the campaign website from the [`CampaignPreview`](https://github.com/DoSomething/phoenix-next/blob/b1538a3b30d8de31ef5cbd7d2e79bd755cd1e239/resources/assets/components/pages/AccountPage/Credits/CampaignPreview.js) directly?)

Perhaps, in some ideal world, we might be able to filter Actions per user from Rogue, so we can have pre-grouped actions and posts per user. 🤔 

Check out the [Tech Spec here](https://docs.google.com/document/d/1fUGIdIXLJ0f9LwirG1AgJglCF2xYl78YV-X_BnnFysI/edit?ts=5e667213#heading=h.dyav5g7zz4s9) for some context on which pieces of data we're pulling out for each row from the larger group of user posts.

### Relevant tickets

References [Pivotal #171729105](https://www.pivotaltracker.com/story/show/171729105).

### Checklist
Tests/Documentation pending!

You can see how things look in the [review app](https://dosomething-phoenix-de-pr-2014.herokuapp.com/us/account/profile/credits). If you've reported back on some of the test campaigns in the dev sandbox, there's a mighty chance you'll have a pending row at least.

image

### DIFF
--- a/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.js
+++ b/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.js
@@ -1,0 +1,101 @@
+import React from 'react';
+import gql from 'graphql-tag';
+import { groupBy, last } from 'lodash';
+import { useQuery } from '@apollo/react-hooks';
+
+import { getUserId } from '../../../../helpers/auth';
+import Spinner from '../../../artifacts/Spinner/Spinner';
+import { getHumanFriendlyDate } from '../../../../helpers';
+import VolunteerCreditsTable from './VolunteerCreditsTable';
+import ErrorBlock from '../../../blocks/ErrorBlock/ErrorBlock';
+
+export const VOLUNTEER_CREDIT_POSTS_QUERY = gql`
+  query volunteerCreditPostsQuery($userId: String!) {
+    paginatedPosts(
+      userId: $userId
+      type: "photo"
+      volunteerCredit: true
+      status: [ACCEPTED, PENDING]
+      first: 100
+    ) {
+      edges {
+        node {
+          id
+          createdAt
+          quantity
+          status
+          actionDetails {
+            id
+            actionLabel
+            timeCommitmentLabel
+          }
+          campaign {
+            campaignWebsite {
+              showcaseImage {
+                url
+                description
+              }
+              showcaseTitle
+              showcaseDescription
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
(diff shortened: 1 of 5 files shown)


resources/assets/helpers/auth.js
```diff
@@ -80,6 +80,13 @@ export function buildAuthRedirectUrl(options = null, actionId = null) {
  */
 export const isAuthenticated = () => window.AUTH.isAuthenticated;
 
+/**
+ * Get the user id.
+ *
+ * @return {String}
+ */
+export const getUserId = () => window.AUTH.id;
+
```
mendelB, line 89: We've been fetching the user data from the window directly to stave off of Redux state in a bunch of places now so it felt like a good idea to encapsulate this!

weerd: Makes sense to me!

resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.js
```diff
@@ -0,0 +1,101 @@
...
+              }
+              showcaseTitle
+              showcaseDescription
+            }
+          }
+        }
+      }
+    }
+  }
+`;
```
weerd, line 46: There was some chatting I did with @DFurnes regarding some of the queries used for the new Home Page and he suggested breaking some stuff into Fragments in their respective components. So we might be able to combine some of this after, for example importing a fragment for the `CampaignGalleryItem` (v2), like you mention (I'd refrain from the `CampaignPreview` since I think it'll likely get deprecated for the newer style I made for the homepage.

mendelB: Ohhh nice! Looking forward to that.

PR #2015, opened 2020-04-03
Adds Analytics Tracking to View More Hyperlink

Vmack:
### What's this PR do?

This pull request adds analytics tracking to the view more scholarship details hyperlink. 

Is the syntax correctly formatted? Does the implementation make sense? Am I missing anything? 

### Relevant tickets

References [Pivotal #171704246](https://www.pivotaltracker.com/story/show/171704246).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
+++ b/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
@@ -5,6 +5,10 @@ import PropTypes from 'prop-types';
 import Query from '../../Query';
 import Card from '../../utilities/Card/Card';
 import { getHumanFriendlyDate } from '../../../helpers';
+import {
+  EVENT_CATEGORIES,
+  trackAnalyticsEvent,
+} from '../../../helpers/analytics';
 
 import './campaign-info-block.scss';
 
@@ -40,6 +44,15 @@ const CampaignInfoBlock = ({
         const actions = res.campaign.actions || [];
         const isOpen = res.campaign.isOpen;
 
+        const handleViewMorelinkSelect = () => {
+          trackAnalyticsEvent('phoenix_clicked_view_more_link_scholarships', {
+            action: 'link_clicked',
+            category: EVENT_CATEGORIES.modal,
+            label: 'SCHOLARSHIP_MODAL',
+            context: { campaignId },
+          });
+        };
+
         let actionItem = actions.find(
           action => action.reportback && action.scholarshipEntry,
         );
@@ -68,7 +81,7 @@ const CampaignInfoBlock = ({
                     <button
                       className="text-blue-500 pb-4"
                       type="button"
-                      onClick={showModal}
+                      onClick={showModal && handleViewMorelinkSelect}
                     >
                       View Scholarship Details
                     </button>
--- a/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
+++ b/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
@@ -44,7 +44,8 @@ const CampaignInfoBlock = ({
         const actions = res.campaign.actions || [];
         const isOpen = res.campaign.isOpen;
 
-        const handleViewMorelinkSelect = () => {
+        const handleViewMoreLinkSelect = () => {
+          showModal();
           trackAnalyticsEvent('phoenix_clicked_view_more_link_scholarships', {
             action: 'link_clicked',
             category: EVENT_CATEGORIES.modal,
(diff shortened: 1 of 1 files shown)


resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
```diff
@@ -40,6 +44,15 @@ const CampaignInfoBlock = ({
         const actions = res.campaign.actions || [];
         const isOpen = res.campaign.isOpen;
 
+        const handleViewMorelinkSelect = () => {
+          trackAnalyticsEvent('phoenix_clicked_view_more_link_scholarships', {
+            action: 'link_clicked',
+            category: EVENT_CATEGORIES.modal,
+            label: 'SCHOLARSHIP_MODAL',
+            context: { campaignId },
```
lindsayrmaher, line 52: The analytics event looks good! It looks like in the ticket the context also includes the `pageId`, so I think that would be a quick add to the context obj. The utms should be added if there are any to add automagically ✨

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
```diff
@@ -68,7 +81,7 @@ const CampaignInfoBlock = ({
                     <button
                       className="text-blue-500 pb-4"
                       type="button"
-                      onClick={showModal}
+                      onClick={showModal && handleViewMorelinkSelect}
```
lindsayrmaher, line 84: I'm trying to decide whether it would be better to add the [tracking here](https://github.com/DoSomething/phoenix-next/blob/3f33d7d39f17224d7ea9291ee175288aceca374a/resources/assets/components/LedeBanner/templates/HeroTemplate.js#L129-L129) rather than invoking both functions. SInce the scholarship modal does technically live in the HeroTemplate component. But I can also see how handling it within the CampaignInfoBlock might be a bit clearer since the link itself lives there. 

What do you think? Also very nitpicky but the "L" in `handleViewMorelinkSelect` should be capitalized!

mendelB: One other important thing -- since we're evaluating both functions using the `&&` logical operator, the functions themselves won't actually be invoked! (It's kind of a neat shortcut that we can just pass the single `showModal` to the `onClick` and it will automatically invoke it when the click event occurs, but we can't do that with two functions)

What we might be able to do though, is invoke the `showModal` directly from the `handleViewMorelinkSelect ` itself! So here, we'd go
```js
onClick={handleViewMorelinkSelect}
```

And then above, in the `handleViewMorelinkSelect`:
```js
const handleViewMorelinkSelect = () => {
  showModal();
  trackAnalyticsEvent('phoenix_clicked_view_more_link_scholarships', {
  ....
}
```

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
mendelB: Since this is a Campaign, there won't actually be any `pageId` to track! Generally speaking, if it's a campaign page -- we'll have a `window.STATE.campaign.id`, and if it's a page (e.g. articles, story pages, etc.) we'll have a `window.page.id`.

We've also got a nifty helper method we can invoke which takes care of grabbing the both of 'em to just account for both cases!
https://github.com/DoSomething/phoenix-next/blob/ca0ff6b49e8a619cea93416a1192293f9af7b8b1/resources/assets/helpers/analytics.js#L181-L193

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
lindsayrmaher: OOOO mendel coming in with door number 3! I like that. That feels very clear and handles everything in the component we're concerned with

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
lindsayrmaher: Ah gotcha, I couldn't remember if the pageId got pulled with a helper or not!

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
mendelB: lolol 

memes

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
mendelB: It was a good call though - because this one doesn't get invoked automagically on events so we'd have to manually import and invoke it!

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
Vmack: This is wonderful, thanks guys!!

PR #2016, opened 2020-04-03
Adding the ability to close the modal by Clicking outside!

lindsayrmaher:
### What's this PR do?

This pull request restores the ability to close a modal on our site by clicking outside of the modal as well as using the `close` button on the modal window. 

👀 

Large Screen
fullsize

Small Screen
mobile

Restoring this [capability that we had](https://github.com/DoSomething/phoenix-next/blob/3bde7a29ad636f7529d54d4b877659da936fba42/resources/assets/components/Modal/Modal.js) before the Modal was updated.

### Relevant tickets

References [Pivotal # 169928815](https://www.pivotaltracker.com/story/show/169928815).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/Modal/Modal.js
+++ b/resources/assets/components/utilities/Modal/Modal.js
@@ -23,7 +23,8 @@ class Modal extends React.Component {
     this.el.className = 'wrapper';
     this.scrollOffset = window.scrollY;
 
-    this.handleClick = this.handleClick.bind(this);
+    this.handleModalCloseClick = this.handleModalCloseClick.bind(this);
+    this.handleModalCloseEsc = this.handleModalCloseEsc.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


resources/assets/components/utilities/Modal/Modal.js
```diff
@@ -32,6 +35,8 @@ class Modal extends React.Component {
     );
     this.modalPortal.classList.add('is-active');
     this.modalPortal.appendChild(this.el);
+    document.addEventListener('click', this.handleModalCloseClick, false);
```
weerd: By adding it to the document won't this mean that ANY click on anything activates the close? I see that the `handleModalCloseClick()` checks against the the `this.el` but couldn't you also just add it to the `this.el.addEventListener(...)`? I could also be misinterpreting something 🤔

weerd: Ya, I think it's clearer and less chance to cause effects if more specific :)

PR #2018, opened 2020-04-03
Test the Volunteer Credits query and data parsing

mendelB:
### What's this PR do?

This pull request adds a test to ensure the somewhat complex data-parsing `VolunteerCreditsTableQuery` component passes the correctly parsed list of posts to the `VolunteerCreditsTable`

👀 

I spent most of my time wrangling with the [`MockedProvider`](https://www.apollographql.com/docs/react/development-testing/testing/#mockedprovider) which is lame. This isn't exhaustive, but I thought it can at least provide us with some level of reassurance, especially as we iterate on this processing logic. 

A potential refactor you might argue would be prudent -- would be to abstract away the parsing logic from the `VolunteerCreditsTableQuery` as helper methods, and add more compartmentalized unit tests on those! This could be a nice refactor once we add a Cypress integration test following our server-rendering updates in Q2(?)

### Relevant tickets

References [Pivotal #172150775](https://www.pivotaltracker.com/story/show/172150775).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Credits/volunteer-credits-mock-data.js
+++ b/resources/assets/components/pages/AccountPage/Credits/volunteer-credits-mock-data.js
@@ -1,0 +1,138 @@
+// Mock Action Details:
+const ACTION_ONE = {
+  id: 1,
+  actionLabel: 'Contact a Decision-Maker',
+  timeCommitmentLabel: '3+ hours',
+};
+
+const ACTION_TWO = {
+  id: 2,
+  actionLabel: 'Attend an Event',
+  timeCommitmentLabel: '4+ hours',
+};
+
+const ACTION_THREE = {
+  id: 3,
+  actionLabel: 'Collect Something',
+  timeCommitmentLabel: '30 minutes - 1 hour',
+};
+
+// Mock Campaign Website Showcase:
+const CAMPAIGN_ONE = {
+  showcaseImage: {
+    url: 'https://images.api.info/1',
+    description: 'Cover Photo 1',
+  },
+  showcaseTitle: 'Example Campaign 1',
+  showcaseDescription: 'Example Description 1',
+};
+
+const CAMPAIGN_TWO = {
+  showcaseImage: {
+    url: 'https://images.api.info/2',
+    description: 'Cover Photo 2',
+  },
+  showcaseTitle: 'Example Campaign 2',
+  showcaseDescription: 'Example Description 2',
+};
+
+const CAMPAIGN_THREE = {
+  showcaseImage: {
+    url: 'https://images.api.info/3',
+    description: 'Cover Photo 3',
+  },
+  showcaseTitle: 'Example Campaign 3',
+  showcaseDescription: 'Example Description 3',
+};
+
+// Mock reponse paginatedPost data from GraphQL:
+export const mockPostsResponse = [
+  {
+    node: {
+      id: 1,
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/AccountPage/Credits/volunteer-credits-mock-data.js
```diff
@@ -0,0 +1,138 @@
...
+];
+
+// Mocked paginatedPost data should be parsed into the following result:
+export const mockParsedPostsData = [
+  {
+    id: 1,
+    campaignWebsite: CAMPAIGN_ONE,
+    actionLabel: ACTION_ONE.actionLabel,
+    dateCompleted: 'April 1st, 2020',
+    volunteerHours: ACTION_ONE.timeCommitmentLabel,
```
katiecrane, line 119: What is the relation here with `timeCommitmentLabel` and `volunteerHours`? I would have expected this to go with the `volunteer_credit` field on an action?

resources/assets/components/pages/AccountPage/Credits/volunteer-credits-mock-data.js
```diff
@@ -0,0 +1,138 @@
...
+    pending: true,
+  },
+  {
+    id: 4,
+    campaignWebsite: CAMPAIGN_TWO,
+    actionLabel: ACTION_TWO.actionLabel,
+    dateCompleted: 'December 10th, 2019',
```
katiecrane, line 126: Excellent idea to check the date here!

resources/assets/components/pages/AccountPage/Credits/volunteer-credits-mock-data.js
mendelB: We return this data as a `volunteerHours` prop [to the table here](https://github.com/DoSomething/phoenix-next/blob/6926f746fcef5cfb079399e5e8f4e3241901605c/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.js#L93) since that's the [title of the column](https://github.com/DoSomething/phoenix-next/blob/6926f746fcef5cfb079399e5e8f4e3241901605c/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsTable.js#L24). I honestly wasn't positive about mapping this over to a new name as opposed to just passing over the `timeCommitmentLabel` directly, but I guess it felt like a better name here in terms of what we're displaying -- the (estimated) volunteer hours they spent on this action. But do you think it'd be less confusing as just `timeCommitmentLabel`?

katiecrane: OH I see now, I was thinking this was the boolean that held whether or not this action qualifies for volunteer hours. But in the query we're already pulling only those actions right?

mendelB: Yes exactly! [Right here](https://github.com/DoSomething/phoenix-next/blob/5946ea289d1edd0ca283ca64371937509e1d384a/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.js#L17) we've filtered for posts who's parent action has the `volunteer_credit` field set to `true`.

katiecrane: Ah and it's `volunteer_credit` not hours. Perfect. Thanks for explaining!

PR #2024, opened 2020-04-08
Allow PhotoSubmissionAction on story pages

katiecrane:
### What's this PR do?

This pull request makes some changes in order to:
1. Let the `PhotoSubmissionAction` work on story pages for authenticated users (I copied how we do this for `TextSubmissionAction`) - which is basically submitting to Rogue using action id rather than campaign id.

2. If a `PhotoSubmissionAction` is on a story page and a user is unauthenticated, they can click an "Add Photo" button (which is basically the same as the "Join Now" button) which will take them through the registration/login flow and bring them back to the story page. They will NOT be scrolled back down to the `PhotoSubmissionAction` when they return to the page NOR will they be signed up for any campaign upon registering. There is no campaign, this is a story page!

BIGGEST CHANGE:
In the `PhotoSubmissionAction`, if we have an action id we now use `storePost` and only fall back on `storeCampaignPost` if we do not have an action id. A cypress test has been updated to reflect this.

Here's what it looks like if you're unauthenticated:
image

Upon submission, users still get the photo submission affirmation - is that okay?

You can test it out [in the review app on this story page](https://dosomething-phoenix-de-pr-2024.herokuapp.com/us/stories/test-for-photo-uploader). UPDATE: this isn't actually gonna redirect you correctly and I think that is a review app thing, but you'll still be able to log in and submit a photo.

We couldn't do this the same way as text posts because we need you to actually be logged in first!

### Relevant tickets

References [Pivotal #171943340](https://www.pivotaltracker.com/story/show/171943340).

### Checklist

### DIFF
--- a/cypress/integration/campaign-post.js
+++ b/cypress/integration/campaign-post.js
@@ -104,7 +104,7 @@ describe('Campaign Post', () => {
 
       // Mock the backend response:
       const response = newPhotoPost(campaignId, user);
-      cy.route('POST', CAMPAIGN_POSTS_API, response).as('submitPost');
+      cy.route('POST', POSTS_API, response).as('submitPost');
 
       // Submit the form, and assert we made the API request:
       cy.contains('Submit a new photo').click();
--- a/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
+++ b/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
@@ -216,15 +216,29 @@ class PhotoSubmissionAction extends PostForm {
       school_id: await this.getUserActionSchoolId(),
     });
 
-    // Send request to store the campaign photo submission post.
-    this.props.storeCampaignPost(this.props.campaignId, {
+    const data = {
       action,
       actionId: this.props.actionId,
       blockId: this.props.id,
       body: formatPostPayload(formFields),
       pageId: this.props.pageId,
+      campaignId: this.props.campaignId,
       type,
-    });
+    };
+
+    // Send request to store the campaign photo submission post.
+    if (this.props.actionId) {
+      this.props.storePost(data);
+    } else {
+      this.props.storeCampaignPost(this.props.campaignId, {
+        action,
+        actionId: this.props.actionId,
+        blockId: this.props.id,
+        body: formatPostPayload(formFields),
+        pageId: this.props.pageId,
+        type,
+      });
+    }
   };
 
   /**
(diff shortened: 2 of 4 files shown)


lindsayrmaher:
@katiecrane so when i click on the `Add Photo` should it take me to the log in page if i'm unauthenticated? It keeps redirecting me to the campaigns page.

lindsayrmaher:
Screen Shot 2020-04-08 at 4 08 05 PM

I think we just need to turn off decoration on hover/click, and maybe set the background to a slightly lighter blue on hover? I think the `button` class handles it a bit differently than the `btn` class.

katiecrane:
@lindsayrmaher hmm I used `btn` because that one was in the tailwind section so I thought it was the preferred one. We didn't get designs for this so I'll see what Leah thinks about how it looks.

Yeah I think the redirect piece won't work on the review app because they use that funky url but I'll make sure!

katiecrane:
okay yeah the `APP_URL` of the review app does not match the actual url of the review app so I believe that means we won't be able to test the redirecting portion from the review app (but I swear it works locally!)

katiecrane:
also switched from `btn` to `button` and it looks a lot better and more similar to the "submit your tip" button that can appear with a `TextSubmissionAction`

lindsayrmaher:
> hmm I used btn because that one was in the tailwind section so I thought it was the preferred one. 

yah i think @weerd made a `btn` class as well so it might've been pulling from that instead of TW. The update looks good!

mendelB:
> also switched from `btn` to `button` and it looks a lot better and more similar to the "submit your tip" button that can appear with a `TextSubmissionAction`

@weerd do you think it'd make sense to add default `btn:hover` rules along with [the others](https://github.com/DoSomething/phoenix-next/blob/fd7125a7d08e270937f8d593ae11927ded144915/resources/assets/scss/base.scss#L19-L30)? I had a similar issue when adding to the [Volunteer Credits table](https://github.com/DoSomething/phoenix-next/blob/fd7125a7d08e270937f8d593ae11927ded144915/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsTableRow.js#L60-L61) where I had to manually add hover state.

mendelB:
@katiecrane out of curiosity -- have we considered querying the campaign ID via the existing action ID for the photo uploader and then actually signing them up for the campaign behind the scenes when they authenticate, kinda like how we do it [for gated quizzes](https://github.com/DoSomething/phoenix-next/blob/0623ddc2a5c1f4982d30d7ecec1010b23e38877f/resources/assets/components/Quiz/QuizResult.js#L22-L28)? _if_ there's some value add from that, e.g. capturing intent from users who don't actually end up reporting back, perhaps that's a good idea! But not sure if that flow makes sense without campaign context and the explicit user intent.

katiecrane:
oooooo interesting idea @mendelB ! as far as I know no, no discussions about creating a signup upon registration, only upon photo submission. @ngjo what do you think about that ^ idea?

PR #2029, opened 2020-04-09
Hello bar in-house - part 1/2 (possibly 3)

lindsayrmaher:
### What's this PR do?

This pull request is to recreate the functionality of helloBar, the 3rd party service we're using to run sitewide CTA banners. Because of past bugs we've encountered with the way hellobar interacts with our in-house functionality, we're making our own!

This PR adds base functionality & styling and puts this under a feature flag until we're ready to deploy.

Does the implementation make sense? most of the styling is there, but I'll need to possibly make updates in a separate PR once the functionality is working.

All the functionality to track dismissal is in place and working as expected, so next steps are getting this to render as expected.

I'll be submitted a follow up PR to this in order to get the rendering of the bar to work as expected. (using React Portal)

Screen Shot 2020-04-09 at 12 19 36 PM

[Product Brief](https://docs.google.com/document/d/1xaLxDVwXWjYXMqNI_yF-Sf4AEfE1WO3Td2AwxXhhyWA/edit#heading=h.k5cdtc3qvsql)

### Relevant tickets

References [Pivotal # 172147577](https://www.pivotaltracker.com/story/show/172147577).

### Checklist

### DIFF
--- a/resources/assets/components/App.js
+++ b/resources/assets/components/App.js
@@ -18,6 +18,8 @@ import CollectionPage from './pages/CollectionPage/CollectionPage';
 import CampaignsIndexPage from './pages/CampaignsPage/CampaignsIndexPage';
 import AccountContainer from './pages/AccountPage/Account/AccountContainer';
 import PageDispatcherContainer from './PageDispatcher/PageDispatcherContainer';
+import SitewideCtaBanner from './utilities/SitewideCtaBanner/SitewideCtaBanner';
+import DismissableElement from './utilities/DismissableElement/DismissableElement';
 import AlphaReferralPageContainer from './pages/ReferralPage/Alpha/AlphaPageContainer';
 
 const App = ({ store, history }) => {
@@ -25,6 +27,21 @@ const App = ({ store, history }) => {
 
   return (
     <Provider store={store}>
+      {featureFlag('sitewide_cta_banner') ? (
+        <DismissableElement
+          name="sitewide_banner_voter_registration"
+          context={{ contextSource: 'voter_registration' }}
+          render={(handleClose, handleComplete) => (
+            <SitewideCtaBanner
+              cta="Get Started"
+              description="Make your voice heard. Register to vote in less than 2 minutes."
+              handleClose={handleClose}
+              handleComplete={handleComplete}
+              link="https://vote.dosomething.org/"
+            />
+          )}
+        />
+      ) : null}
       <ApolloProvider client={graphqlClient(env('GRAPHQL_URL'))}>
         <Router history={history}>
           <Switch>
--- a/resources/assets/components/App.js
+++ b/resources/assets/components/App.js
@@ -25,6 +25,22 @@ const App = ({ store, history }) => {
 
   return (
     <Provider store={store}>
+      <div className="w-full flex justify-center bg-yellow-500 p-4 fixed z-50">
+        <button type="button" className="modal__close">
+          &times;
+        </button>
+        <p className="pb-2 md:pr-4 md:pb-0 align-middle">
+          Make your voice heard. Register to vote in less than 2 minutes.
+        </p>
+        <a
+          className="py-2 px-4 border border-solid-blurple rounded-md bg-blurple-500 text-white uppercase"
+          href="https://vote.dosomething.org/"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          Get Started
+        </a>
+      </div>
       <ApolloProvider client={graphqlClient(env('GRAPHQL_URL'))}>
         <Router history={history}>
           <Switch>
--- a/resources/assets/components/App.js
+++ b/resources/assets/components/App.js
@@ -25,22 +25,6 @@ const App = ({ store, history }) => {
 
   return (
     <Provider store={store}>
-      <div className="w-full flex justify-center bg-yellow-500 p-4 fixed z-50">
-        <button type="button" className="modal__close">
-          &times;
-        </button>
-        <p className="pb-2 md:pr-4 md:pb-0 align-middle">
-          Make your voice heard. Register to vote in less than 2 minutes.
-        </p>
-        <a
-          className="py-2 px-4 border border-solid-blurple rounded-md bg-blurple-500 text-white uppercase"
-          href="https://vote.dosomething.org/"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Get Started
-        </a>
-      </div>
       <ApolloProvider client={graphqlClient(env('GRAPHQL_URL'))}>
         <Router history={history}>
           <Switch>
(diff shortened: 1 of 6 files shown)


aaronschachter:
All the changes make it a little tough to get a feel for what's going on here, especially as someone relatively green to the `SiteNavigation` component. It seems we're wrapping it all in a React Fragment and adding in a `DismissableElement`. Could we feature flag the DismissableElement to get this in a place where we could merge this as is?

Also out of scope for this PR, but I'm curious about extracing some of these navigation `<ul>` 's into separate components,e.g. `SiteNavigation/CausesList` to make the overall `SiteNavigation` component a bit easier to make these big changes to... (cc @weerd )

lindsayrmaher:
> All the changes make it a little tough to get a feel for what's going on here, especially as someone relatively green to the `SiteNavigation` component. It seems we're wrapping it all in a React Fragment and adding in a `DismissableElement`. Could we feature flag the DismissableElement to get this in a place where we could merge this as is?
> 
> Also out of scope for this PR, but I'm curious about extracing some of these navigation `<ul>` 's into separate components,e.g. `SiteNavigation/CausesList` to make the overall `SiteNavigation` component a bit easier to make these big changes to... (cc @weerd )

Definitely will add a feature flag. I'm wondering whether rendering it in the `SiteNavigation` component makes the most sense vs `App.js` since it is sitewide. But I'm not sure it will matter too much once it's using a portal. 

Also agree about pulling out the pieces of the nav into smaller components! That could be a good clean up item perhaps? 🤔

aaronschachter:
> whether rendering it in the SiteNavigation component makes the most sense vs App.js since it is sitewide. But I'm not sure it will matter too much once it's using a portal.

Could rendering it `App.js` for now cut down on the number of changes this PR introduces? It's really tough to see what's going on there. One advantage of rendering it in App.js would be that the banner will display on pages that don't display the SiteNavigation, like the Beta Referral Page.

Should we choose to move into SiteNavigation instead, we could potentially move some of those lists into smaller components to make the changes more reviewable. But you mention it might not make a difference if we use a React portal -- why wouldn't it make a difference?

lindsayrmaher:
> Could rendering it App.js for now cut down on the number of changes this PR introduces? It's really tough to see what's going on there.

Yeah it definitely does make the PR much easier to digest! (updated on 4b2f36b) And it doesn't effect how it currently renders which is nice.

> But you mention it might not make a difference if we use a React portal -- why wouldn't it make a difference?

So the way I'm hoping to handle what parent container this renders inside directly in the `SitewideCtaBanner` component using the React Portal! It should be something similar to this in the [Modal component](https://github.com/DoSomething/phoenix-next/blob/dd4555a41bc8d3d98b81461085bbe16bed439311/resources/assets/components/utilities/Modal/Modal.js#L37-L37) So my thought was that whether we render the `SitewideCtaBanner` inside of `SiteNavigation` or `App` it will show up in the DOM where we want it to be. 

But given your point about it rendering on pages without the `SiteNavigation`, I think it makes more sense for this to live in `App`

lindsayrmaher:
@aaronschachter and @katiecrane i think this is in a spot to be reviewed! I'll update the description to reflect what should be coming in future PRs.

resources/assets/helpers/index.js
```diff
@@ -987,3 +987,16 @@ export function updateQuery(previous, { fetchMoreResult }) {
     isArray(dest) ? [...dest, ...src] : undefined,
   );
 }
+
+/**
+ * Formula to convert time to milliseconds.
+ *
+ * @param {Number} days
+ */
+export function getTimeinMilliseconds(days) {
```
katiecrane, line 996: I'm wondering if it would be clearer to call this `getDaysInMilliseconds` or `getMillisecondsFromDays` or something so you can see from the name of the function that it is expecting you to give it a number of days?

lindsayrmaher: Yeah I like that! updated on c94334f

resources/assets/helpers/index.js
```diff
@@ -987,3 +987,16 @@ export function updateQuery(previous, { fetchMoreResult }) {
...
   );
 }
+
+/**
+ * Formula to convert time to milliseconds.
+ *
+ * @param {Number} days
+ */
+export function getMillisecondsFromDays(days) {
+  // default to 30 days if no value is provided
```
katiecrane, line 997: Sorry another Q on this helper - do we need a default here? Why would we call this if we don't give it a number of days?

resources/assets/components/utilities/DismissableElement/DismissableElement.js
```diff
@@ -46,7 +50,10 @@ const DismissableElement = ({ name, render, context }) => {
 
     // Was the element dismissed less than 30 days ago?
     const dismissalTime = getStorage(`dismissed_${name}`, 'timestamp');
-    const isDismissed = isTimestampValid(dismissalTime, 30 * 1440 * 60 * 1000);
+    const isDismissed = isTimestampValid(
+      dismissalTime,
+      getMillisecondsFromDays(7),
```
katiecrane, line 55: Why do we want to change this for all `<DismissableElement>`s from 30 to 7? Looks like 30 is that standard we expect?

lindsayrmaher:
> Sorry another Q on this helper - do we need a default here? Why would we call this if we don't give it a number of days?

> Why do we want to change this for all <DismissableElement>s from 30 to 7? Looks like 30 is that standard we expect?

I totally meant to set the `DismissableElement` with a default and not the helper and my brain said, NOPE. Updated on 7cbc4d4

@katiecrane

resources/assets/components/utilities/DismissableElement/DismissableElement.js
```diff
@@ -56,12 +63,14 @@ const DismissableElement = ({ name, render, context }) => {
 
 DismissableElement.propTypes = {
   context: PropTypes.object,
+  daysToReRender: PropTypes.number,
```
aaronschachter, line 66: What do you think about naming this to something like `daysToWaitUntilDisplayAgain`?  I think it might be a little more descriptive.

aaronschachter: Nothing needed for now, but wondering about re-purposing this property, and we could potentially want to wait hours vs days for other dismissable elements.

resources/assets/components/App.js
```diff
@@ -18,13 +18,31 @@ import CollectionPage from './pages/CollectionPage/CollectionPage';
...
+            <SitewideCtaBanner
+              cta="Get Started"
+              description="Make your voice heard. Register to vote in less than 2 minutes."
+              handleClose={handleClose}
+              handleComplete={handleComplete}
+              link="https://vote.dosomething.org/"
+            />
+          )}
+        />
+      ) : null}
```
aaronschachter, line 45: Thanks for moving this into `App` vs `SiteNavigation` ! Makes it way easier to see what's going on.

resources/assets/components/utilities/DismissableElement/DismissableElement.js
lindsayrmaher: Yeah! I like making it a bit more descriptive. It should probably be `timeToWaitUntilDisplayAgain` but since it's only set up rn to accept a days amount i'll leave it! I'll make a note in the helper as a `TODO`

PR #2030, opened 2020-04-09
Enabling Snowplow link tracking

weerd:
### What's this PR do?

This pull request enables tracking link clicks in Snowplow using a manual method that is called automatically whenever any link is clicked on a rendered page.

👀 

Ooof, here we go...

We initially saw that you could enable link tracking in Snowplow by calling the [`snowplow('enablelinkclicktracking')` method](https://github.com/snowplow/snowplow/wiki/2-Specific-event-tracking-with-the-Javascript-tracker-v2.5#391-enablelinkclicktracking), but that depends on links being present on the page for it to "active" them to track clicks.

However, since our page is rendered with React on the client side, by the time this method is called, there are no links on the page yet. I tried seeing if I could call this method from inside the `NewHomePage` component using a `useEffect()` but didn't seem to work. I then found that there's an additional method that can be called if links are added to a page after the fact, called [`snowplow('refreshLinkClickTracking')`](https://github.com/snowplow/snowplow/wiki/2-Specific-event-tracking-with-the-Javascript-tracker-v2.5#392-refreshlinkclicktracking), and the docs indicate:

> `enableLinkClickTracking` only tracks clicks on links which exist when the page has loaded. If new links can be added to the page after then which you wish to track, just use `refreshLinkClickTracking`. This will add Snowplow click listeners to all links which do not already have them...

This seemed hopeful, but again I tried running it in a `useEffect()` on the component and since these functions run in the background by the Snowplow script, it was hard to tell if they were actually working. Even checking the Network tab in the browser tools wasn't super helpful since links result in full page refreshes and clears out the prior Network tab info.

Lastly, after chatting it over with @DFurnes, we concluded that maybe the best approach would be to handle it ourselves, using a listener on the `body` and relying on _event bubbling_ to find when a link is clicked and then trigger the [`snowplow('trackLinkClick')` method](https://github.com/snowplow/snowplow/wiki/2-Specific-event-tracking-with-the-Javascript-tracker-v2.5#393-tracklinkclick) manually. An added bonus to this approach is that since we can trigger using our `analyze()` wrapper function, we can display the event in the console along with our other analytic events!

This all seemed great until I ran into a couple of frustrating things that stalled my progress for a bit. As it turns out, it was the crappy documentation for the older Snowplow v2.5! So as to not make mistakes, I copy-pasted the example calls from the docs, here:

image

The first bug, was that while the method signature shows the first parameter is `targetUrl` (the only required argument), the example entirely misses this, which was a bit confusing.

Along with this, the example also has a typo for the method name; it is `trackLinkCLick ` instead of `trackLinkClick `, which was especially frustrating given the cryptic error that resulted in the console.

To add to the frustration, these mistakes in the documentation were fixed in the [later versions of the docs](https://github.com/snowplow/snowplow/wiki/2-Specific-event-tracking-with-the-Javascript-tracker#393-tracklinkclick), but were never retroactively fixed 😡

### Relevant tickets

References [Pivotal #172202644](https://www.pivotaltracker.com/story/show/172202644).

### Checklist

### DIFF
--- a/resources/assets/init.js
+++ b/resources/assets/init.js
@@ -43,7 +43,11 @@ import { init as historyInit } from './history';
 import { bindTokenRefreshEvent } from './helpers/auth';
 import { bindFlashMessageEvents } from './helpers/flash-message';
 import { bindAdminDashboardEvents } from './helpers/admin-dashboard';
-import { analyze, trackAnalyticsPageView } from './helpers/analytics';
+import {
+  analyze,
+  trackAnalyticsLinkClick,
+  trackAnalyticsPageView,
+} from './helpers/analytics';
 
 ready(() => {
   // Enable Debug tools.
@@ -98,4 +102,13 @@ ready(() => {
   if (navAppElement) {
     ReactDom.render(<NavApp store={store} history={history} />, navAppElement);
   }
+
+  // Track link clicks for Snowplow analytics.
+  document.body.addEventListener('click', clickEvent => {
+    if (clickEvent.target.tagName.toLowerCase() === 'a') {
+      clickEvent.preventDefault();
+
+      trackAnalyticsLinkClick(clickEvent.target);
+    }
+  });
 });
--- a/resources/assets/init.js
+++ b/resources/assets/init.js
@@ -106,8 +106,6 @@ ready(() => {
   // Track link clicks for Snowplow analytics.
   document.body.addEventListener('click', clickEvent => {
     if (clickEvent.target.tagName.toLowerCase() === 'a') {
-      clickEvent.preventDefault();
-
       trackAnalyticsLinkClick(clickEvent.target);
     }
   });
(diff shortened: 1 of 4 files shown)


resources/assets/init.js
```diff
@@ -98,4 +102,11 @@ ready(() => {
   if (navAppElement) {
     ReactDom.render(<NavApp store={store} history={history} />, navAppElement);
   }
+
+  // Track link clicks for Snowplow analytics.
+  document.body.addEventListener('click', clickEvent => {
+    if (clickEvent.target.tagName.toLowerCase() === 'a') {
+      trackAnalyticsLinkClick(clickEvent.target);
+    }
+  });
```
DFurnes, line 111: Since this is binding an delegated event to the body & no longer relies on the page being fully rendered to function, let's move this up above with the other event listeners (~line 93).

weerd:
@sheyd meant to ask, any idea if the Fivetran limitations that kept us on an older version of the Snowplow JavaScript script have been resolved so we could update to the newer versions with the more up to date docs, haha 😉

sheyd:
@weerd Sadly no. 😡 I just checked on things and the version is the same 🤷 .

PR #2034, opened 2020-04-10
Hello Bar In-house 2/2

lindsayrmaher:
### What's this PR do?

This pull request is wrapping up making this helloBar visible via react portals and finishing up styling that was started in #2029 

I'd like some feedback on if all the naming makes sense. And also whether the CSS i added is the most efficient etc

**Large Screen**
Screen Shot 2020-04-10 at 5 02 07 PM

**Medium Screen**
Screen Shot 2020-04-10 at 5 02 25 PM

**Small Screen**
Screen Shot 2020-04-10 at 5 02 41 PM

### Relevant tickets

References [Pivotal # 172147577](https://www.pivotaltracker.com/story/show/172147577).

### Checklist


resources/views/layouts/master.blade.php
```diff
@@ -37,6 +37,8 @@
         @yield('content')
     </div>
 
+    <div id="sitewide-cta-banner" class="sitewide-cta-banner" role="presentation"></div>
```
weerd, line 40: Could we be more generic on the naming and maybe call it something like `banner-portal` to also follow suffix pattern similar to the `modal-portal`?

resources/assets/components/utilities/SitewideCtaBanner/sitewidectabanner.scss
```diff
@@ -0,0 +1,10 @@
+@import '../../../scss/next-toolbox.scss';
+
+.sitewide-cta-banner {
+  display: block;
+  width: 100%;
+  left: 0;
+  top: 0;
+  z-index: 501;
+  position: absolute;
+}
```
weerd, line 10: Can we do these as Tailwind classes? I think we could probs add a `z-500` class in the Tailwind config and we can use that on the Site Navigation and for this. I _think_ it should be fine to use the same z-index for both since this banner appears later in the markup and thus if using the same `z-index` the item that renders later in the stack renders on top :)

resources/views/layouts/master.blade.php
weerd: If we can't get away with just simple `banner-portal` and `Banner` & `BannerContent`... then `sitewide-banner-portal` and respective `SitewideBanner` & `SitewideBannerContent` also works. I think the CTA part of it, should be a component that is passed into the Banner Content... as could some notification or something else. Doesn't always have to be a CTA.

lindsayrmaher:
@weerd thanks! your comments were super helpful. at this point i'm just trying to get the banner to display _above_ the site nav rather than over the top of it since it basically makes it unusable otherwise. 

I've tried rendering the `banner-portal` from within the `chrome` div so that they're direct siblings, but I'm still getting the same overlay as before. Any thoughts on an approach that would work for this? Similar to [this page](https://storycorps.org/participate/great-questions/).

weerd:
> at this point i'm just trying to get the banner to display above the site nav rather than over the top of it since it basically makes it unusable otherwise.

Ah! Haha, sorry I got confused cause I thought we were replicating the less than ideal, original HelloBar experience which didn't care and went on top of everything, lol. Ok, will take another look shortly. Might be a bit more complicated 🤔

resources/views/layouts/master.blade.php
```diff
@@ -37,6 +37,8 @@
         @yield('content')
     </div>
 
+    <div id="banner-portal" class="top-0 left-0 w-full block absolute z-500" role="presentation"></div>
```
mendelB, line 40: what if we placed the portal _inside_ the app chrome and did away with the `absolute` and `fixed` positioning, so that it just displays as a regular child in the chrome content and fits naturally in the page? (apologies if this was considered and/or makes no sense!)

weerd: Yeah, I was considering this but might be a moot point if it all still needs to be absolutely positioned. This would help if the end result is going with `sticky` positioning; but as we've seen that can cause problems on its own.

mendelB: ugh totally... I wonder if `sticky` would work these days now that you've cleaned up the content hierarchy a bit

weerd: Playing around with it now in the inspector and not as straightforward as you would expect. On homepage at least it sticks for a while and then seems to arbitrarily decide to stop sticking at some point further down the page 🤷‍♂️ Messing with it...

lindsayrmaher: yah i just noticed that as well.... with the z-index up a bit higher than site nav and sticky it's _almost_ working as expected. but seems like halfway down it just stops sticking

weerd: Looks like it's the `body`'s height limiting it. Essentially it stops sticking when the stuff at the bottom of the viewport window reaches the top, the sticky-ness ends. Haha

weerd: But also playing with the Site Nav, testing interference with the nav drop-down and the `fixed` positioned white underlay 😓

mendelB: oy, lol, this really illustrates the problem we had with just popping in an external dependency that fixes itself within your content

lindsayrmaher: it also looks like the point at which the sticky wears off is a lot sooner on mobile.

weerd: Yep, expected, cause the "body" only stretches as far as the screen height.

@lindsayrmaher Are you adding position sticky to element sitting in portal _outside_ of App or inside of it? Just curious to compare with how I'm doing it. I suspect if inside of App like Mendel mentioned sticky would work better, but I fear other complications.

lindsayrmaher: this is what i have inside my master.blade.php:
`<div id="chrome" class="chrome h-full w-full">
        <div id="banner-portal" class="top-0 left-0 w-full block sticky z-800" role="presentation"></div>
        @yield('content')
    </div>`

lindsayrmaher: and these classes are on the `SitewideBannerContent` So App > SitewideBanner > SitewidebannerContent

`<div className="w-full flex justify-center bg-yellow-500 p-4 z-50">`

lindsayrmaher: i guess tldr, i moved it back inside of "chrome" and am seeing the same behavior you were talking about diego

mendelB: hmmm I think we might need to put it inside the `#app` itself. perhaps by pretending the content directly from the `SitewideBanner` without a preexisting portal? (since the contents of the app are replaced with the react app on the page load)

This way it's a direct sibling to the other app content (site nav, page content, etc.)

mendelB: the catch i think would be that the campaign page navigation would overtake it since you scroll down since it's also sticky 
https://github.com/DoSomething/phoenix-next/blob/fd7125a7d08e270937f8d593ae11927ded144915/resources/assets/components/utilities/PageNavigation/page-navigation.scss#L14

weerd: image

Ok, I got it sticky for the whole page, but b/c it sits inside the `#chrome` > `#app`, but then can't go full site width. I'm trying something else that might allow us to put it still as a portal sitting outside the chrome... hold please...

weerd:
Banner Portal Example

Ok! Played around with it a bit and found a solution.

image

First off, lets move the `#banner-portal` up above the `#chrome` div, so that it naturally takes up as much space as it needs (more or less what @mendelB was mentioning when suggesting to put inside the_chrome_) and keeps us from absolute positioning it and then dealing with how much vertical space it takes up and matching that same amount as padding above the Site Navigation which would have been a headache.

Second, apply `position: sticky` to the `#banner-portal` and it needs a new `z-index`. Let's keep the `z-500` custom class (I'll eventually use this on the Site Nav), but lets also add a `z-1000` class and use that on this banner.

Third (and I checked this a bit, but I recommend checking a few other pages on your local @lindsayrmaher), we need to remove the [`height: 100%`](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/scss/base.scss#L77) on the `html` and `body`. And instead _just_ apply it to the `html`. By setting it on both, it's forcing the height of the `html` to only span the height of the viewport, which is fine, but by having it also on the `body` it forces it to then only go 100% of the `html` height which is the viewport. Instead we don't want to force this and allow the `body` to take up as much space as the page content, which it can naturally do (hover over the `<body>` in the inspector and you'll see a blue highlight in the viewport showing how far the `<body>` stretches) when no `height` declaration is specified.

Keeping the `height: 100%` only on the `html` also helps to ensure that pages that have minimal content  and don't stretch the full viewport height to still allow the container to span the full height.

Lemme know if anything doesn't make sense 😬

lindsayrmaher:
@weerd this was perfection!!!!! I just pushed up those changes and everything is working as expected for me on all screens 💃

weerd:
> I just pushed up those changes and everything is working as expected for me on all screens 💃

Yay! I'm glad it's working! Goodbye "HelloBar"! 👋

resources/assets/components/App.js
```diff
@@ -18,7 +18,7 @@ import CollectionPage from './pages/CollectionPage/CollectionPage';
 import CampaignsIndexPage from './pages/CampaignsPage/CampaignsIndexPage';
 import AccountContainer from './pages/AccountPage/Account/AccountContainer';
 import PageDispatcherContainer from './PageDispatcher/PageDispatcherContainer';
-import SitewideCtaBanner from './utilities/SitewideCtaBanner/SitewideCtaBanner';
+import SitewideBanner from './utilities/SitewideBanner/SitewideBanner';
```
weerd, line 21: Nit-pick! Stacking order, shortest to longest lines 😄

resources/assets/components/utilities/SitewideBanner/SitewideBannerContent.js
```diff
@@ -0,0 +1,42 @@
...
+  description,
+  handleClose,
+  handleComplete,
+  link,
+}) => {
+  return (
+    <div className="w-full flex justify-center bg-yellow-500 p-4 z-50">
+      <button type="button" className="modal__close" onClick={handleClose}>
+        &times;
+      </button>
```
weerd, line 15: Let's switch to using the established [close button component](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/artifacts/CloseButton/CloseButton.js)

resources/assets/components/utilities/SitewideBanner/SitewideBannerContent.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SitewideBannerContent = ({
```
weerd, line 4: Not necessary for this PR, but moreso a future refactor, I think it would be useful for this to have a similar structure as the Modal Portal, etc, where the the `SitewideBannerContent` receives a component or markup as `children` to render out.

Could be a nice feature, say in the future where we want to get a message out to all users for some sort of rapid response and we pass it a different notification component to showcase sitewide across the banner 🤔

lindsayrmaher: agreed, agreed. i might note this as a good item for the clean up sprint in may! or we could include it as a part of the follow up work where we are adding user targeting as well.

weerd:
Once merged, I recommend checking a variety of pages on Dev/QA, triggering some modals etc to make sure that `height: 100%` rule change within the html/body is alright. I only tested a couple pages I thought would definitely be affected! 😅

PR #2035, opened 2020-04-13
Adds Cross-Site NPS Survey 

Vmack:
### What's this PR do?
This pull request adds the NPS survey across the entire site except the homepage. The homepage has a separate nps survey.

Screen Shot 2020-04-13 at 2 38 52 PM

Do we want the full page url or just the path? Some users may have different UTM's.

**UPDATE**
So I updated the conditional to check against the `"/us" `path using the history variable as per Diego L's [advice](https://dosomething.slack.com/archives/CUQMU4Q6B/p1587415405012300?thread_ts=1586980300.079400&cid=CUQMU4Q6B) however, because the history only stores the path after visiting the page at least once, you'll notice that the survey will not render on the initial visit. So good news is it works! 🙌🏽 but not the initial time. So not sure if this is sufficient but just a heads up. Also I will wrap in a feature flag and change the traffic distribution percentage and delay once this PR is ready to be merged.

Uses similar survey to the one on the new `Homepage` and the one in `Campaign.js` except this survey will render across the entire site and capture the url and northstar Id.

### Relevant tickets

References [Pivotal #171759043](https://www.pivotaltracker.com/story/show/171759043).

### Checklist


lindsayrmaher:
Since App.js is the root component I'm not sure how any `props` would be getting passed to it. You might need to get whether they are authenticated off of `state` instead.

resources/assets/components/App.js
```diff
@@ -8,24 +9,50 @@ import graphqlClient from '../graphql';
...
 import AlphaReferralPageContainer from './pages/ReferralPage/Alpha/AlphaPageContainer';
 
-const App = ({ store, history }) => {
+const App = ({ store, history, props }) => {
   initializeStore(store);
 
   return (
     <Provider store={store}>
       <ApolloProvider client={graphqlClient(env('GRAPHQL_URL'))}>
+        {props.isAuthenticated ? (
```
mendelB, line 35: we'll want to render the survey to _all_ users regardless of authentication status. so to @lindsayrmaher's point https://github.com/DoSomething/phoenix-next/pull/2035#issuecomment-612942763 if we just remove this condition, along with the `props` we should be all set!

Vmack: Aha! Awesome, thanks! So how do I trigger the survey, just to test if it works? 🤔

mendelB: one way for testing is to set the `TrafficDistribution`'s `percentage` to `100` (or locally within your [application storage](https://developers.google.com/web/tools/chrome-devtools/storage/localstorage), you can update the `TrafficDistribution_nps_survey-object` value to `{"showFeature":true}`), and then you can set the `DelayedElement`'s `delay` prop to `0` or `1` so it pops up right away!

Vmack: UPDATE
So I updated the conditional to check against the HomePage using the history variable as per Diego L's advice however, because the history only stores the path after visiting the page at least once, you'll notice that the survey will not render on the initial visit. So good news is it works! 🙌🏽 but not the initial time. So not sure if this is sufficient but just a heads up. Also I will wrap in a feature flag and change the traffic distribution percentage and delay once this PR is ready to be merged.

mendelB: @Vmack looks like the survey's launching on the initial visit! (at least for me, is it not on your end? 🤔 )

Vmack: @mendelB oh yea that's great! It wasn't the case before but certainly good to hear. Wonder why though

resources/assets/components/App.js
```diff
@@ -45,6 +50,27 @@ const App = ({ store, history }) => {
...
+            <DismissableElement
+              name="nps_survey"
+              render={(handleClose, handleComplete) => (
+                <DelayedElement delay={60}>
+                  <Modal onClose={handleClose} trackingId="SURVEY_MODAL">
+                    <TypeFormEmbed
+                      displayType="modal"
+                      typeformUrl="https://dosomething.typeform.com/to/Bvcwvm"
+                      queryParameters={{
+                        northstar_id: get(window.AUTH, 'id', null),
```
mendelB, line 64: Do we intend to also provide the `url` query parameter?

resources/assets/components/App.js
```diff
@@ -45,6 +50,27 @@ const App = ({ store, history }) => {
         />
       ) : null}
       <ApolloProvider client={graphqlClient(env('GRAPHQL_URL'))}>
+        {history.location.pathname !== '/us' ? (
+          <TrafficDistribution percentage={5} feature="nps_survey">
+            <DismissableElement
+              name="nps_survey"
+              render={(handleClose, handleComplete) => (
+                <DelayedElement delay={60}>
```
mendelB, line 58: I noticed in the Pivotal story we wanted the delay to be `30` seconds?

resources/assets/components/App.js
```diff
@@ -45,6 +50,27 @@ const App = ({ store, history }) => {
...
+                        northstar_id: get(window.AUTH, 'id', null),
+                      }}
+                      onSubmit={handleComplete}
+                    />
+                  </Modal>
+                </DelayedElement>
+              )}
+            />
+          </TrafficDistribution>
+        ) : null}
```
mendelB, line 73: Do we want to wrap this in a feature flag?

resources/assets/components/App.js
Vmack: ah! good eye!

resources/assets/components/App.js
Vmack: yea I was actually just checking with @ddonizeth regarding that

resources/assets/components/App.js
Vmack: Yes! totally forgot about that. So assuming it would look something like this: 
`url: get(window.location),` ? @mendelB

mendelB: Exactly! though I think we can grab the pathname the same way we do for the conditional above: `url: history.location.pathname,` 
(though I wonder if in _both_ the conditional above and here, we should just go with the `window.location.pathname`. Not sure why we need to use `history` here at all?)

resources/assets/components/App.js
```diff
@@ -45,6 +50,29 @@ const App = ({ store, history }) => {
         />
       ) : null}
       <ApolloProvider client={graphqlClient(env('GRAPHQL_URL'))}>
+        {!featureFlag('sitewide_nps_survey') &&
```
mendelB, line 53: This could just be flipped to 
```suggestion
        {featureFlag('sitewide_nps_survey') &&
```

Vmack: yea but that doesn't work either. The survey doesn't appear when I do this.

resources/assets/components/App.js
Vmack: great point! pushing up those changes now. I used history because I thought it was the only way to get access to the pathname.  But using window actually is a better access point

resources/assets/components/App.js
mendelB: Hmm 🤔  do you have the `DS_ENABLE_SITEWIDE_NPS_SURVEY=true` set in your `.env`?

Vmack: you're right that was the one thing that I forgot to do 😓thanks so much @mendelB 🙌🏽alrighty it's working now and ready to be merged 😃

PR #2036, opened 2020-04-13
Make sure we have quantity before displaying it in galleries

katiecrane:
### What's this PR do?

This pull request grabs `quantity` in addition to `impact` in `<PostCard>` so we can check if we have `quantity` AND `impact` (which is like "3 jeans collected") before displaying it so we can avoid displaying things like "null things done."

Before:
Screen Shot 2020-04-13 at 10 22 34 AM

After:
Screen Shot 2020-04-13 at 11 16 19 AM

Will this ever prevent us from displaying the impact when we might actually want to? Anywhere else I need to add this change? Does this prevent us from displaying `null` and `0`?

We used to just check if we have `impact` but that doesn't save us if the `quantity` is `null`!

### Relevant tickets

References [Pivotal #171901122](https://www.pivotaltracker.com/story/show/171901122).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/PostCard/PostCard.js
+++ b/resources/assets/components/utilities/PostCard/PostCard.js
@@ -21,6 +21,7 @@ export const postCardFragment = gql`
     text
     tags
     impact
+    quantity
     location(format: HUMAN_FORMAT)
     signupId
     createdAt
@@ -104,7 +105,7 @@ const PostCard = ({ post, hideCaption, hideQuantity, hideReactions }) => {
             </p>
           ) : null}
 
-          {post.impact && !hideQuantity ? (
+          {post.quantity && post.impact && !hideQuantity ? (
             <p className="mt-1 text-gray-600 text-sm">{post.impact}</p>
           ) : null}
 


resources/assets/components/utilities/PostCard/PostCard.js
```diff
@@ -104,7 +105,7 @@ const PostCard = ({ post, hideCaption, hideQuantity, hideReactions }) => {
             </p>
           ) : null}
 
-          {post.impact && !hideQuantity ? (
+          {post.quantity && post.impact && !hideQuantity ? (
```
weerd, line 108: Should we use that JS function that helps deal with pluralization to handle when it's "1 Jeans Collected" or whatever? @DFurnes do you recall the function/library... I think you were involved in using it in the past 🤔

DFurnes:
Instead of pulling another field into Phoenix to sidestep this bug, could we fix the [`impact` resolver](https://github.com/DoSomething/graphql/blob/master/src/repositories/rogue.js#L470-L481) in GraphQL to return `null` if quantity is null instead of `"null things done"`?

resources/assets/components/utilities/PostCard/PostCard.js
DFurnes: Pluralization is handled in the [`impact` resolver](https://github.com/DoSomething/graphql/blob/master/src/repositories/rogue.js#L470-L481) in GraphQL!

weerd: oh! then I'm confused because the impact resolver seems to also include the quantity... 🤔

katiecrane:
@DFurnes would we be okay with also returning `null` if the `quantity` is `0` from GraphQL for `impact`? We don't want to display anything for `null` or `0` in Phoenix. If that's cool then yeah I think that is a better fix

resources/assets/components/utilities/PostCard/PostCard.js
katiecrane: `impact` gives you the whole statement like "3 jeans collected" - so it's quantity noun and verb

weerd: ok, I might just be misreading this all 😬 

So I'm confused why we need the `quantity` then? If `impact` comes back w/ the full expected phrase, shouldn't the resolver return `null` (or whatever it defaults to if no impact) or the appropriately stated phrase? Why do we need the `quantity` check as well?

weerd: Is this because sometimes we have "impact" that's no quantity based? If so I still kind of feel like GraphQL should just deal with that scenario better.

katiecrane: @weerd that's what Dave is saying down below! but I'm wondering about when the quantity is `0` because in Phoenix we don't want to display that but I'm not sure if there are cases where we would actually want to return "0 jeans collected" - but otherwise yeah I think it'd be cool to be able to handle this in GraphQL instead of here

weerd: Oh sorry, I was in the "Files changed" page and didn't see @DFurnes's comment! 😬

weerd:
@katiecrane I think it makes sense to me to return `null` if the `quantity` is `0` 👍

DFurnes:
That makes good sense to me! 👍 I can't think of a scenario where we'd need those.

resources/assets/components/utilities/PostCard/PostCard.js
katiecrane: no worries!!

katiecrane:
thanks crew! I'll close this one out and make the change in GraphQL instead - we'll return `null` if for `impact` if the `quantity` is `0` or `null` 👍

PR #2037, opened 2020-04-13
Volunteer Credit Certificate Data

mendelB:
### What's this PR do?

This pull request queries and parses out the remaining bits of data needed for the Volunteer Credits Table certificates.

This might look like a lot, but it should all be pretty straightforward! I'd recommend a commit-by-commit approach for a digestible review.

You'll notice that we're essentially replicating some logic we've set up [in GraphQL](https://github.com/DoSomething/graphql/blob/ba64e6594745686aae8479962f447ada2bc991e5/src/repositories/rogue.js#L470-L481) to generate the impact label per post. This is because we're in a bit of a unique situation: we want to generate this label based on the _total_ quantity of post impact for _this_ action. 
- Each post contains an impact label specific to that post's own quantity
- The campaign signup contains a total of quantity for _all_ campaign actions' posts.

This could be something we eventually add to GraphQL (maybe a field on the action) but would require more thinking around specifics (particularly how to handle pagination et al. (this overall has similar qualities to [this issue](https://github.com/DoSomething/phoenix-next/pull/1998#discussion_r398806993))) Things we'll hopefully be talking about in later iterations!

Another notable change here, was to add a central prop types definition for this beefy `post` object, so we can pass that around to the various table rows and utility components more seamlessly.

Finally, there's the big open question of what to do about posts with no quantity or actions where we generally don't collect quantity info (which relates to [Pivotal #171901122](https://www.pivotaltracker.com/story/show/171901122)). I left a TODO for now, and will raise this question in Pivotal.

### Relevant tickets

References [Pivotal #172150720](https://www.pivotaltracker.com/story/show/172150720).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.js
+++ b/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.js
@@ -1,8 +1,8 @@
 import React from 'react';
 import gql from 'graphql-tag';
 import pluralize from 'pluralize';
+import { get, groupBy, last } from 'lodash';
 import { useQuery } from '@apollo/react-hooks';
-import { get, groupBy, last, findLast } from 'lodash';
 
 import { getUserId } from '../../../../helpers/auth';
 import Spinner from '../../../artifacts/Spinner/Spinner';
--- a/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.js
+++ b/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import gql from 'graphql-tag';
-import { groupBy, last } from 'lodash';
 import { useQuery } from '@apollo/react-hooks';
+import { get, groupBy, last, findLast } from 'lodash';
 
 import { getUserId } from '../../../../helpers/auth';
 import Spinner from '../../../artifacts/Spinner/Spinner';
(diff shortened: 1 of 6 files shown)


resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.js
```diff
@@ -111,6 +111,7 @@ const VolunteerCreditsQuery = () => {
     const acceptedPosts = posts.filter(post => post.status === 'ACCEPTED');
 
     // Calculate total quantity of accepted posts.
+    // @TODO: How do we handle 'null' quantity on a post? Or generally, actions not collecting quantity?
```
Vmack, line 114: Is it possible to maybe give it a label of "no action data" or something of that sort? And quantify how many posts have no action or quantity?  Of course there's more to be explored with this question but I was just wondering.

mendelB: That's a good point! We do have a few ways to figure out if the generated 'post' here is a non-quantity collecting one (`null` quantity on any of the action's posts, or `0` total quantity for the action's accepted posts). I think we'll just have to decide on what we want to present in the certificate for such cases if we allow these types of actions to collect volunteer credit.

I'll follow up on this question in the Pivotal story!

mendelB: Closing the loop here -- looks like we'll be ensuring no impact label is generated in these cases https://www.pivotaltracker.com/story/show/172150720/comments/213541203.

PR #2044, opened 2020-04-17
Put the negative margin in the wrong spot

lkpttn:
### What's this PR do?

Fixes an issue with negative margin on the account page navigation

👀 

ugh

Before:
Screen Shot 2020-04-17 at 11 19 19 AM

After:
Screen Shot 2020-04-17 at 11 21 24 AM


lkpttn:
The navigation (and gallery-grids) have a built in x-margin that pushes them away from the grid lines, so you get things misaligned. I think you ended up doing something similar on the Hub pages, just using negative margin to offset the built in padding and go back to the grid.

mendelB:
> The navigation (and gallery-grids) have a built in x-margin that pushes them away from the grid lines, so you get things misaligned. I think you ended up doing something similar on the Hub pages, just using negative margin to offset the built in padding and go back to the grid.

Ahhh right right, those pesky margins! Thanks for clarifying!

PR #2048, opened 2020-04-20
Blurple buttons foreva part deux

weerd:
### What's this PR do?

This pull request renames the button components to follow a better "PrefixButton" naming pattern rather than "ButtonSuffix" pattern and begins updating to use the new button components in a few areas.

To allow switching to using the new button components in the `FilterNavigation` component, it required doing some refactoring of that component, along with updates to the `PrimaryButton` and the `SecondaryButton` components. The updates allow for having a permanent "active" state on these buttons when clicked.

👀 

There are a few more components that need button updates but wanted to limit the scope of changes in this PR. Upcoming button changes include adding a new `ToggleButton` for use in some of the profile pages, with a new visual look that @lkpttn and I have been mulling over.

### Relevant tickets

### Checklist


resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
```diff
@@ -15,7 +15,7 @@ const CauseInput = ({ causeChecked, causeName, causeValue, handleSelect }) => (
       type="checkbox"
       value={causeValue}
       onClick={handleSelect}
-      checked={causeChecked}
+      defaultChecked={causeChecked}
```
weerd, line 18: We were getting an error in the browser console about how this was being set.

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -1,61 +1,67 @@
...
-      document.getElementById(filterName).blur();
-    } else {
-      document.getElementById(filterName).focus();
-      setChosenFilter(filterName);
+    if (activeFilter === selectedFilter) {
+      return setActiveFilter(null);
     }
-    setShowFilterMenu(!showFilterMenu);
+
+    return setActiveFilter(selectedFilter);
```
weerd, line 33: I originally had this as a ternary but ESLint kept yelling at me about it... I guess this might be a bit more readable anyhoo 🤷‍♂️ 

```js
// error  Expected an assignment or function call and instead saw an expression  no-unused-expressions

activeFilter === selectedFilter
      ? setActiveFilter(null)
      : setActiveFilter(selectedFilter);
```

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterSubNav.js
```diff
@@ -7,11 +7,12 @@ import CauseFilter from '../CampaignFilters/CauseFilter/CauseFilter';
 
 const renderedFilterMenu = props => {
   const fields = withoutNulls(props);
-  switch (fields.chosenFilter) {
-    case 'Cause':
+  switch (fields.activeFilter) {
+    case 'cause':
```
weerd, line 11: I find it better to rely on lowercases across the board and only use the capitalized case in the html visual output! The values of all the potential `activeFilter` options should now all be lowercase based on the filters set in the [CampaignIndexpage](https://github.com/DoSomething/phoenix-next/blob/f4dd374d76661c6e90cb36d33f38d4283fc451b2/resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js#L12) :)

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -1,61 +1,67 @@
...
 const FilterNavigation = ({ filters, setFilters }) => {
-  const [chosenFilter, setChosenFilter] = useState('');
-  const [showFilterMenu, setShowFilterMenu] = useState(false);
+  const [activeFilter, setActiveFilter] = useState(null);
 
-  const handleMenuToggle = filterName => {
-    trackAnalyticsEvent(`clicked_filter_button_${filterName}`, {
+  const filterCategoryNames = Object.keys(filters).map(filter =>
+    pluralize.singular(filter),
+  );
```
weerd, line 18: This essentially grabs all the lowercase keys defined in the `filters` in the [CampaignIndexpage](https://github.com/DoSomething/phoenix-next/blob/f4dd374d76661c6e90cb36d33f38d4283fc451b2/resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js#L12)! No need to continue hardcoding the value and now will automagically update when more filters are added to this object ✨

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -1,61 +1,67 @@
...
-            className={!chosenFilter ? 'hidden' : ''}
+
+        {filterCategoryNames.map(name => (
+          <SecondaryButton
+            key={`${name}_button`}
+            className="mr-8"
+            data={{ 'data-filter': name }}
+            onClick={handleMenuToggle}
+            text={startCase(name)}
+            isActive={activeFilter === name}
```
weerd, line 48: The filter buttons will now inherit all the state behavior of secondary buttons, like hover, focus, active, etc!

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
weerd: It also seems like we visually want to display the filter as a singular word, but when it comes from `filters` it is plural; thus the use of the nifty `pluralize` package since we already include it in the repo 😄

weerd:
@lindsayrmaher pinging you, since it includes a bit of refactoring of the `FilterNavigation` you worked on!

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
lindsayrmaher: 😍 I'm so excited about this!

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
lindsayrmaher: I think I had the exact same problem when I was working on this originally 😭

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterSubNav.js
lindsayrmaher: This is much cleaner! It makes more sense for this not to feel so hardcoded to match the text in the buttons themselves.

PR #2049, opened 2020-04-20
Extract Tailwind variables into their own file.

DFurnes:
### What's this PR do?

This pull request fixes an issue where the site would not render in Edge 18 (~2% of total traffic). This was occurring because we import Tailwind's `resolveConfig` method to be able to access variables from our own `tailwind` helper (e.g. `tailwind('colors.gray.200')`).

Since Tailwind is [compiled for Node.js](https://github.com/tailwindcss/tailwindcss/blob/7ef2e0a61311d0c65cbb258e7b032dd9faa77fcb/package.json#L69-L71) (and not browsers, like Edge 18, that don't support the same baseline language features as Node.js 8.x), this was erroring out. The quick fix is to extract those variables we access from this helper (screen sizes, colors, and spacing) into their own object that we can read from both environments independently.

It's easy to forget nowadays, but this is the curse of the web platform! What works perfectly well in one browser may completely explode in another.

I generated `tailwind.variables.js` by reading each relevant section from the result of the real `resolveConfig` helper, to make sure that this didn't result in any changes in the end result.

In the future, I'd love to revisit whether we can continue to read these from the Tailwind config using a [compile-time macro](https://github.com/kentcdodds/babel-plugin-macros) (like how [Twin](https://github.com/ben-rogerson/twin.macro) works!). This would save us from having to evaluate the config at runtime in the browser & let us send less bytes over the wire!

### Relevant tickets

References [Pivotal #172411856](https://www.pivotaltracker.com/story/show/172411856).

### Checklist


tailwind.variables.js
```diff
@@ -0,0 +1,176 @@
...
+    '20': '5rem',
+    '24': '6rem',
+    '32': '8rem',
+    '40': '10rem',
+    '48': '12rem',
+    '56': '14rem',
+    '64': '16rem',
+    px: '1px',
+  },
+};
```
weerd, line 176: Ah yeah, I was gonna mentioned we'd need to bring in the spacing defaults. 👍

PR #2050, opened 2020-04-21
Utility Function for adding VR query params

lindsayrmaher:
### What's this PR do?

This pull request creates a utility function that can be used in VR work throughout phoenix to build the correct url with tracking params. I'm using it on this PR with the "helloBar". 

👀 

Because we want to send similar if not identical information along in multiple places throughout Phoenix, hopefully this will help us to streamline it! 

### Relevant tickets

References [Pivotal # 172411428](https://www.pivotaltracker.com/story/show/172411428).

### Checklist

### DIFF
--- a/resources/assets/components/App.js
+++ b/resources/assets/components/App.js
@@ -5,7 +5,6 @@ import { ApolloProvider } from '@apollo/react-common';
 import { Router, Route, Switch } from 'react-router-dom';
 
 import graphqlClient from '../graphql';
-import { env, featureFlag } from '../helpers';
 import { initializeStore } from '../store/store';
 import HomePage from './pages/HomePage/HomePage';
 import BlockPage from './pages/BlockPage/BlockPage';
@@ -13,6 +12,7 @@ import CausePage from './pages/CausePage/CausePage';
 import NewHomePage from './pages/HomePage/NewHomePage';
 import CompanyPage from './pages/CompanyPage/CompanyPage';
 import CampaignContainer from './Campaign/CampaignContainer';
+import { env, featureFlag, buildVoterRegUrl } from '../helpers';
 import BetaReferralPage from './pages/ReferralPage/Beta/BetaPage';
 import CollectionPage from './pages/CollectionPage/CollectionPage';
 import SitewideBanner from './utilities/SitewideBanner/SitewideBanner';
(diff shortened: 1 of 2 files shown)


resources/assets/helpers/index.js
```diff
@@ -998,3 +998,22 @@ export function getMillisecondsFromDays(days) {
...
+ * @param {String} baseUrl
+ * @param {String} source
+ * @param {String} sourceDetails
+ */
+
+export function buildVoterRegUrl(baseUrl, source = 'web', sourceDetails) {
+  const isUserId =
+    window.AUTH.isAuthenticated && window.AUTH.id
+      ? `user:${window.AUTH.id}`
+      : '';
```
mendelB, line 1014: Should we maybe just call this variable `userId`? Also, maybe worth throwing the trailing comma here following `user:${window.AUTH.id}` instead of having a potentially extra comma before `source:...` if there's no user ID?

resources/assets/helpers/index.js
```diff
@@ -998,3 +998,22 @@ export function getMillisecondsFromDays(days) {
...
+ * Build URL with UTMs for Voter Registration Content
+ *
+ * @param {String} baseUrl
+ * @param {String} source
+ * @param {String} sourceDetails
+ */
+
+export function buildVoterRegUrl(baseUrl, source = 'web', sourceDetails) {
+  const isUserId =
+    window.AUTH.isAuthenticated && window.AUTH.id
```
mendelB, line 1012: Curious about the `window.AUTH.isAuthenticated`, are we concerned about potentially having an assigned ID on the window without explicit authentication? AFAIK the scenario where that'd happen (beyond a curious hack to append the query param explicitly), would be web links in SMS broadcasts, where we occasionally used the `user_id` query param as 'soft auth'. though not sure if we even do this anymore 🤔 

https://github.com/DoSomething/phoenix-next/blob/ee80282895c55feec4090859773bab803035e06d/app/Providers/AppServiceProvider.php#L58

resources/assets/components/App.js
```diff
@@ -39,7 +39,11 @@ const App = ({ store, history }) => {
               description="Make your voice heard. Register to vote in less than 2 minutes."
               handleClose={handleClose}
               handleComplete={handleComplete}
-              link="https://vote.dosomething.org/?r=source:web,source_details:hellobar"
+              link={buildVoterRegUrl(
+                'https://vote.dosomething.org/',
```
mendelB, line 43: Is the base URL ever anything other than `https://vote.dosomething.org/`? could we perhaps bake that into the helper for now?

PR #2053, opened 2020-04-22
Blurple buttons foreva part trois

weerd:
### What's this PR do?

This pull request continues the button updates on Phoenix. 

Specifically it adds a new `ToggleButton` that visually looks like a combination of the `PrimaryButton` and `SecondaryButton` (styling is subject to change) but behaves differently; it is meant to be entirely a toggle that "activate" or "deactivates" a setting with a very distinct visual style and language to represent "on" and "off". Example usage would be subscribe/unsubscribe from newsletters or follow/unfollow cause interests.

GIF example for email subscriptions:
ToggleButton Email Subscriptions

Image example with cause preferences:
image

The above sections now both utilize the `ToggleButton` and it contains all the expected :hover, :focus, active styles including the different styles based on whether the button is toggled "on" or "off".

👀 

### Relevant tickets

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Interests/CausePreferenceItem.js
+++ b/resources/assets/components/pages/AccountPage/Interests/CausePreferenceItem.js
@@ -2,8 +2,10 @@ import React from 'react';
 import { get } from 'lodash';
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
-import classNames from 'classnames';
 import { useMutation, useQuery } from '@apollo/react-hooks';
+
+import Spinner from '../../../artifacts/Spinner/Spinner';
+import ToggleButton from '../../../utilities/Button/ToggleButton';
 
 const CAUSE_PREFERENCE_QUERY = gql`
   query CausePreferenceQuery($userId: String!) {
@@ -45,35 +47,32 @@ const CausePreferenceItem = ({ cause, description, title }) => {
   const causes = get(data, 'user.causes', []);
 
   return (
-    <div className="border border-solid border-gray-300 p-4 rounded-md flex justify-between">
+    <div className="border border-solid border-gray-300 p-4 rounded-md flex">
       <div className="w-2/3 lg:w-3/4 pr-4">
         <h1 className="text-blurple-500 text-base text-bold">{title}</h1>
+
         <p className="text-sm text-gray-500">{description}</p>
       </div>
+
       <div className="w-1/3 lg:w-1/4">
-        <button
-          type="button"
-          className={classNames(
-            'btn w-full border border-solid border-blurple-500 focus:outline-none',
-            !causes.includes(cause)
-              ? 'bg-blurple-500 text-white hover:bg-blurple-300 focus:bg-blurple-500 focus:text-white'
-              : 'bg-white border text-blurple-500 border-solid hover:border-blurple-300 hover:text-blurple-200 focus:bg-white focus:text-blurple-500',
-          )}
-          onClick={() =>
-            updateInterest({
-              variables: {
-                cause,
-                interested: !causes.includes(cause),
-              },
-            })
-          }
-        >
-          {!loading ? (
-            <>{causes.includes(cause) ? 'Unfollow' : ' Follow '}</>
-          ) : (
-            <div className="spinner" />
-          )}
-        </button>
+        {loading ? (
+          <Spinner className="flex justify-center p-2" />
+        ) : (
+          <ToggleButton
+            activateText="Follow"
+            className="w-full"
+            deactivateText="Unfollow"
+            isToggled={causes.includes(cause)}
+            onClick={() =>
+              updateInterest({
+                variables: {
+                  cause,
+                  interested: !causes.includes(cause),
+                },
+              })
+            }
+          />
+        )}
       </div>
     </div>
   );
--- a/resources/assets/components/pages/AccountPage/Interests/CausePreferenceItem.js
+++ b/resources/assets/components/pages/AccountPage/Interests/CausePreferenceItem.js
@@ -38,7 +38,10 @@ const CausePreferenceItem = ({ cause, description, title }) => {
 
   // Make the initial query to get the user's subscriptions
   const { data, loading, error } = useQuery(CAUSE_PREFERENCE_QUERY, options);
-  const [updateInterest] = useMutation(CAUSE_PREFERENCE_MUTATION, options);
+  const [updateInterest, { loading: modifying }] = useMutation(
+    CAUSE_PREFERENCE_MUTATION,
+    options,
+  );
 
   if (error) {
     return <p>Something went wrong!</p>;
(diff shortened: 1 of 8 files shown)


resources/assets/components/utilities/Button/ToggleButton.js
```diff
@@ -0,0 +1,67 @@
...
+    className,
+  );
+
+  const deactivateClasses = classnames(
+    'bg-white active:bg-gray-200',
+    'border-2 border-blurple-500 active:border-blurple-700 hover:border-blurple-300 border-solid focus:rounded-none',
+    'focus:outline-2 focus:outline-blurple-100 focus:outline-solid',
+    'text-base text-blurple-500 active:text-blurple-700 hover:text-blurple-300',
+    className,
+  );
```
weerd, line 33: These styles mimic those of both the `PrimaryButton` and `SecondaryButton`. I debated just having the `ToggleButton` resort to rendering either one depending on if it was toggled or not, but currently there is no guarantee the visual design of the toggle button won't change. I figured it'd be best to just mimic the styles and allow for better control of what toggle buttons look like by having it all be in the component instead of calling other styled components to render.

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -35,8 +35,8 @@ const FilterNavigation = ({ filters, setFilters }) => {
 
   return (
     <div className="md:w-full bg-white">
-      <div className="flex items-center md:w-3/4 mx-auto pt-2 md:pt-0 pb-6 pl-6 md:pl-0">
-        <h2 className="text-gray-600 pr-4 text-base uppercase">Filters</h2>
+      <div className="flex items-center md:w-3/4 mx-auto pt-2 pb-6 pl-6 md:pl-0">
+        <h2 className="mb-0 pr-4 text-gray-600 text-base uppercase">Filters</h2>
```
weerd, line 39: Quick fix to help align button to title.

resources/assets/components/pages/AccountPage/Interests/CausePreferenceItem.js
```diff
@@ -45,35 +47,32 @@ const CausePreferenceItem = ({ cause, description, title }) => {
...
+            onClick={() =>
+              updateInterest({
+                variables: {
+                  cause,
+                  interested: !causes.includes(cause),
+                },
+              })
+            }
+          />
+        )}
```
weerd, line 75: I was debating on something. Previously the spinner was inside the button but it only really showed up on initial load, because the `loading` variable is only for the initial fetch query for the user's preferences. So I moved it out so it spins and then the buttons load. Not sure if the intention was for the spinner to also show after toggling to follow/unfollow... and if that's the case then we'd need a `loading` variable associated with the mutation query. If we can do this I would argue that we just disable the button when it's mutating the query rather than show a spinner 🤷‍♂️  Thoughts?

DFurnes: Yeah, that's tricky! I do think moving this outside the button looks a tiny bit nicer.

I definitely think showing both "loading" states makes sense (and disabling is a fine "loading" state for the mutation), since otherwise the interface seems a little laggy while it waits for the mutation to complete.

Another option would be to use an optimistic response (like [this](https://github.com/DoSomething/rogue/blob/46dd50a30a490dd1c2f6bb0645d68c5ce5b0fad8/resources/assets/components/ReviewButton.js#L29-L37)) to make things appear to take effect immediately. (And if it, for some reason, were to error out it'd just "pop back" to whatever we get in the real mutation result.)

resources/assets/components/pages/AccountPage/Subscriptions/EmailSubscriptionItem.js
```diff
@@ -51,32 +53,28 @@ const EmailSubscriptionItem = ({ topic, name, image, description }) => {
...
-              variables: {
-                topic,
-                subscribed: !topics.includes(topic),
-              },
-            })
-          }
-        >
-          {topics.includes(topic) ? 'Unsubscribe' : 'Subscribe'}
-        </button>
+        <img style={{ width: '100%' }} src={image} alt="newsletter" />
```
DFurnes, line 56: I wonder if a more appropriate `alt` tag for this is "The Breakdown logo"? I feel like "newsletter" doesn't help me much as a vision-impaired user. (I know this isn't new, but thought of it while skimming this section!)

resources/assets/components/pages/AccountPage/Subscriptions/EmailSubscriptionItem.js
```diff
@@ -51,32 +53,28 @@ const EmailSubscriptionItem = ({ topic, name, image, description }) => {
...
+
+          <p className="flex-grow">{description}</p>
+
+          <ToggleButton
+            activateText="Subscribe"
+            deactivateText="Unsubscribe"
+            isToggled={topics.includes(topic)}
+            className="mt-4"
+            onClick={() =>
+              updateSubscription({
```
DFurnes, line 69: Another place where we should show a mutation loading state (or optimistic response)!

resources/assets/components/pages/AccountPage/Subscriptions/EmailSubscriptionItem.js
weerd: 🤔 ya, that's a good call.

resources/assets/components/pages/AccountPage/Subscriptions/EmailSubscriptionItem.js
weerd: Yeah, I was thinking that too. This and the newsletter topics seem similar but I was mostly just trying to revise with what was already there. Defs should be similar behavior!

resources/assets/components/pages/AccountPage/Interests/CausePreferenceItem.js
DFurnes: Yeah, it's a bummer that they conflict by default. I've [renamed to `modifying`](https://github.com/DoSomething/phoenix-next/blob/764468a4f7f5a76f0b8284e61a7d1efb84a712fd/resources/assets/components/pages/AccountPage/Profile/DeleteAccountTab.js#L41-L44) in the past.

weerd: Oh nice, that's helpful. I was wondering how best to go about renaming it when pulling the variable 👌

weerd:
Disabled When Toggling

@DFurnes made update to use `modifying` to disable button for now 🕺 Works well! Also made the same style updates on the email subscription so they work the same way.

DFurnes:
Love it! One quick thought – could we add a `isLoading` variant to the button that applies the same gray styles but uses [`.cursor-wait`](https://tailwindcss.com/docs/cursor/#wait) for the pointer styling? The 🚫 sort-of makes me think something has went wrong.

DFurnes:
Or perhaps just remove the cursor override for now? Your call!

weerd:
`isLoading` makes sense to me! Seems like a worthwhile update to have. Good call ☎️

weerd:
Disabled When Toggling With Wait Cursor

Ok @DFurnes! Updated with waiting cursor. Lemme know what you think of the approach 🤔

PR #2054, opened 2020-04-23
Volunteer Credit Certificate Template Implementation

mendelB:
### What's this PR do?

This pull request implements the Volunteer Credit certificate template (using stubbed data).

I'll try and link to relevant documentation where applicable. But mostly, this is good ole markup! Think of `<View>`s as `<div>`s, `<Text>`s as `<p>`s, and `Image`s... well you take a good guess 😁 

Main thing to 👀 would be
- the font registration implementation (I did my best, but I'm a little fuzzy on whether this was implemented perfectly)
- the markup formatting and structuring
- the styling in general (both in the general `Stylesheet`, and respective inline `style={{}}` objects)

I added a temporary view so you can easily check out the PDF in the review app [here](https://dosomething-phoenix-de-pr-2054.herokuapp.com/pdf). I'll delete this before merging.

I'll try and comment where helpful but https://react-pdf.org/ has some quickly digestible documentation if any of this is confusing! You'll notice stubbed data, once this is approved and merged, I'll go ahead and implement the real shizzazle.

### Relevant tickets

- References [Pivotal #172150619](https://www.pivotaltracker.com/story/show/172150619).
- Design mocks [Pivotal 171960189](https://www.pivotaltracker.com/n/projects/2438772/stories/171960189)
- [Formal mock](https://drive.google.com/drive/folders/1GEk9Io9pmUvS_Wqrxbluhr1OfE6PsN2e)
- [Colors and fonts annotation](https://user-images.githubusercontent.com/12417657/80120247-b47e1700-8558-11ea-8211-013fe57eb604.png)

- ~[Here's the rendered PDF!](https://github.com/DoSomething/phoenix-next/files/4523524/4e022b6f-5b9d-4fc2-893b-c357f77461e5.pdf)~
[Updated Version](https://github.com/DoSomething/phoenix-next/files/4525660/e86c14a5-47ce-4b92-879a-26c522cab589.pdf) per https://github.com/DoSomething/phoenix-next/pull/2054#discussion_r414167580


resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
```diff
@@ -19,19 +29,225 @@ export const certificatePostType = PropTypes.shape({
...
 const styles = StyleSheet.create({
-  page: { backgroundColor: 'tomato' },
-  section: { color: 'white', textAlign: 'center', margin: 30 },
+  flex: { display: 'flex', flexDirection: 'row' },
+  postDetailsTitle: {
+    fontWeight: 'bold',
+    color: colors.purple,
+    textTransform: 'uppercase',
+  },
 });
```
mendelB, line 70: See https://react-pdf.org/styling. We're able to define some base styles here. Most styling though felt better inline. 

I should note, that we can technically use [Styled Components](https://react-pdf.org/styling#styled-components) here, but that would necessitate installing a separate library, and didn't seem to me like a big enough win as of now. Though good to keep an eye on if we iterate on or add to PDF rendering!

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
```diff
@@ -1,11 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Document, Page, Text, View, StyleSheet } from '@react-pdf/renderer';
+import {
+  Document,
+  Page,
+  Text,
+  View,
+  Image,
```
mendelB, line 8: See https://react-pdf.org/components

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
```diff
@@ -19,19 +29,225 @@ export const certificatePostType = PropTypes.shape({
...
+    },
+  ],
+});
+
+Font.register({
+  family: 'League Gothic',
+  src: '/fonts/leaguegothic-regular.woff',
+  fontWeight: 400,
+  fontStyle: 'normal',
+});
```
mendelB, line 53: See https://react-pdf.org/fonts. I coped this over from our [local CSS `@font-face` 'registration'](https://github.com/DoSomething/phoenix-next/blob/3db4f95fac9505e8aadab2b4a97cb63392622ffd/resources/assets/scss/fonts.scss#L1). 

We needed these to be publicly accessible, hence the copy over to `/public/fonts`.

There doesn't seem to be a way to specify the `format` here.

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
```diff
@@ -19,19 +29,225 @@ export const certificatePostType = PropTypes.shape({
...
+          height: '100%',
+          paddingHorizontal: 100,
+        }}
+      >
+        <View
+          style={[styles.flex, { justifyContent: 'center', marginTop: 30 }]}
+        >
+          <View>
+            <Image
+              src="/images/dosomething-logo.jpg"
```
mendelB, line 92: I added this and the `volunteer-credit-certificate-title.jpg` to the `public/images` dir. SVG's [aren't yet supported](https://github.com/diegomura/react-pdf/issues/29) by this library.

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
```diff
@@ -19,19 +29,225 @@ export const certificatePostType = PropTypes.shape({
...
+              </Text>
+              <Text style={{ fontStyle: 'italic' }}>
+                Community Associate, DoSomething.org
+              </Text>
+            </View>
+
+            <View style={{ width: '50%' }}>
+              <Text style={{ marginTop: 10, textAlign: 'right' }}>
+                For questions and verification issues, please reach out to Maddy
+                {'\n'} at{' '}
```
mendelB, line 226: You could argue that this is a bit much, but I did this manual newline in one or two places to exactly copy the PDF mock.

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
```diff
@@ -19,19 +29,225 @@ export const certificatePostType = PropTypes.shape({
...
+  src: '/fonts/leaguegothic-regular.woff',
+  fontWeight: 400,
+  fontStyle: 'normal',
+});
+
+// PDF colors config.
+const colors = {
+  lightBlue: '#5fc6c8',
+  darkBlue: '#3b3d98',
+  purple: '#864f9f',
```
DFurnes, line 59: Could we perhaps read these from our `tailwind()` helper? (Not sure if they line up 1:1 with web colors.)

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
```diff
@@ -19,19 +29,225 @@ export const certificatePostType = PropTypes.shape({
...
+// We're still stubbing the data for now so:
+// eslint-disable-next-line no-unused-vars
 const CertificateTemplate = ({ certificatePost }) => (
   <Document>
-    <Page size="A4" style={styles.page}>
-      <View style={styles.section}>
-        <Text>Certificate of Credit.</Text>
-        <Text>{certificatePost.id}</Text>
+    <Page
+      orientation="landscape"
```
DFurnes, line 77: Do we also need to specify `size="LETTER"` here?

It looks like the [default](https://react-pdf.org/components#page) is `A4` (which, TIL, is the default page size that the rest of the world has decided on).

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
DFurnes: Bummer that they don't support Emotion! I agree this is the right trade-off for now. 👍

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
DFurnes: I'm curious if we could reference the files via an `import`... Webpack _should_ provide an absolute URL to the file path as the "return value" for that import, which we could then use here. (These could then use the same compiled `063baf84a53cb42775b59f6fea7bc57c.woff` or whatnot that the user already has cached in their browser).

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
mendelB: Oof! This is a stellar catch! You know what, I made the dumb assumption that we wanted `A4`, I had this notion that that was the standard in the US, but it seems that it's `LETTER`! 

I'd made a note in one of the Pivotal cards that this output seemed different than the mock PDF, failing to realize it's because it's a different size 😖 

I'll make this update, along with some spacing updates to account for the smaller width. I'll also note this in Pivotal to confirm.

mendelB: maybe this is too much for launch, but do you think it's worth noting for the v2 to decide the `size` here based on country code via fastly so we accommodate the hashtag global `A4` standard?

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
mendelB: These, unfortunately, are indeed unique to our web colors 🙁

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
mendelB: Ah interesting! I'll test this out!

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
mendelB: Thanks again for this flag. Confirmed this [in Pivotal](https://www.pivotaltracker.com/story/show/172150619/comments/213722517) updated in https://github.com/DoSomething/phoenix-next/pull/2054/commits/255f022ca6e288d6eb9668e2212b842756338785 and finessed the styling a bit in https://github.com/DoSomething/phoenix-next/pull/2054/commits/b8536b71cb22cea87e4bd6ba99ab23b5843ba810 (diverting a bit from the font-size annotation, so @lkpttn please let me know what you think), but this is looking pretty damn close to the original now!!

resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
mendelB: 🙌 This worked like a charm in https://github.com/DoSomething/phoenix-next/pull/2054/commits/6f2cf12e913c84f0b767f7d1fe2aa92bb89889d6 for the image assets as well!

lkpttn:
Yeah I think this is looking good as a PDF. I don't think we allow CTAs that would be long enough, but does the RB container flex if it needs to be multiple lines?

mendelB:
> Yeah I think this is looking good as a PDF. I don't think we allow CTAs that would be long enough, but does the RB container flex if it needs to be multiple lines?

Thanks for the 👀! 

That's a really good flag. It won't currently because of it's fixed size, so in the follow up to use dynamic data instead of the stub, I'll account for the `256` max on the CTA. (It might entail pushing down the font-size a tad, but I'll try to avoid that if possible) and I'll tag you on the result!

mendelB:
> > Yeah I think this is looking good as a PDF. I don't think we allow CTAs that would be long enough, but does the RB container flex if it needs to be multiple lines?
> 
> Thanks for the 👀!
> 
> That's a really good flag. It won't currently because of it's fixed size, so in the follow up to use dynamic data instead of the stub, I'll account for the `256` max on the CTA. (It might entail pushing down the font-size a tad, but I'll try to avoid that if possible) and I'll tag you on the result!

@lkpttn I actually went ahead and adjusted things in https://github.com/DoSomething/phoenix-next/pull/2054/commits/d07fe7fbef87eeb594d307998236a14479797e56 
Removing the image container allowed removing the fixed-height. Once we're pulling in the dynamic data, we can dynamically adjust the font size for the CTA to minimize it per char count which I think is the best way to responsively size this within our static confines.

PR #2055, opened 2020-04-23
A migration for the new CTA type

lkpttn:
### What's this PR do?
Migrates the Call to Action content type to hold the new fields for the updated Call to Action component

👀 

User facing changes coming after!

### Relevant tickets

References [Pivotal #172203389](https://www.pivotaltracker.com/story/show/172203389).


DFurnes:
Does that work for updating an existing content type? It looks like the documentation suggests that we make those changes manually (which we can do, but I'd worry we'd make a mistake)!

mendelB:
We spoke IRL -- we're happy with a migration here since the changes are rather extensive! We'll generate a new `content-types` file once we've finished updating this Content type

PR #2060, opened 2020-04-23
Fix for filter navigation clear button.

weerd:
### What's this PR do?

This pull request resolves a bug that was introduced in [#2048](https://github.com/DoSomething/phoenix-next/pull/2048/files#diff-02a27f20fa3f7ba1788718a9c7ebed90R18), which cause the React component to no longer handle the state changes and caused problems with not allowing to clear the checked items when clicking the "clear" button.

In that PR the `checked` attribute was changed to `defaultChecked` to deal with a console error that was popping up. While it resolved the error, that change converted the form from a controlled component to an uncontrolled one.

Resolving this bug simply involved switching back to using `checked` and changing the `onClick` handler to `onChange` which is the handle form inputs are meant to use.

Also took the opportunity for further cleanup and changing the "clear" button to use `ElementButton`!

👀 

### Relevant tickets

### Checklist

### DIFF
--- a/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
+++ b/resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
@@ -2,53 +2,65 @@ import React from 'react';
 import { get } from 'lodash';
 import PropTypes from 'prop-types';
 
+import ElementButton from '../../../../utilities/Button/ElementButton';
 import {
   EVENT_CATEGORIES,
   trackAnalyticsEvent,
 } from '../../../../../helpers/analytics';
 
-const CauseInput = ({ causeChecked, causeName, causeValue, handleSelect }) => (
-  <label className="flex justify-start pb-2" htmlFor={causeValue}>
+const causeLabels = {
+  'animal-welfare': 'Animal Welfare',
+  bullying: 'Bullying',
+  education: 'Education',
+  environment: 'Environment',
+  'gender-rights': 'Gender Rights & Equality',
+  'homelessness-and-poverty': 'Homelessness & Poverty',
+  immigration: 'Immigration & Refugees',
+  'lgbtq-rights': 'LGBTQ+ Rights & Equality',
+  'mental-health': 'Mental Health',
+  'physical-health': 'Physical Health',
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
```diff
@@ -2,54 +2,66 @@ import React from 'react';
...
+  environment: 'Environment',
+  'gender-rights': 'Gender Rights & Equality',
+  'homelessness-and-poverty': 'Homelessness & Poverty',
+  immigration: 'Immigration & Refugees',
+  'lgbtq-rights': 'LGBTQ+ Rights & Equality',
+  'mental-health': 'Mental Health',
+  'physical-health': 'Physical Health',
+  'racial-justice': 'Racial Justice & Equity',
+  'sexual-harassment': 'Sexual Harassment & Assault',
+};
```
weerd, line 24: I moved this outside the component so the component code is easier to read :)

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
```diff
@@ -2,54 +2,66 @@ import React from 'react';
...
+
+/**
+ * Checkbox input component.
+ *
+ * @param {Object}
+ */
```
weerd, line 30: When there's multiple components in a single component file, I like adding docblocks so it's easier to visually register them when scrolling :)

resources/assets/components/pages/CampaignsPage/CampaignFilters/CauseFilter/CauseFilter.js
```diff
@@ -2,54 +2,66 @@ import React from 'react';
...
-      onClick={handleSelect}
-      defaultChecked={causeChecked}
     />
     <span className="pl-4">{causeName}</span>
   </label>
 );
 
 CauseInput.propTypes = {
-  causeChecked: PropTypes.bool,
+  isChecked: PropTypes.bool,
```
weerd, line 47: I renamed this prop, since the convention in-house is to prefix with verbs like `is*` and `has*` for variables that are booleans.

PR #2063, opened 2020-04-24
POC - Volunteer Credit Certificate Download Imperative flow

mendelB:
### What's this PR do?

This pull request utilizes the `react-pdf`  library's [`pdf` function](https://react-pdf.org/advanced#access-blob-data) to imperatively download the PDF certificate upon button click.

Do you agree with the logic outlined below, and think this approach makes sense? Point 1 feels stronger to me. But I'm honestly not completely convinced and open to counter thoughts and ditching this approach.

How do you feel about this implementation? Is the code clear and readable?

There are two issues with the current (more straightforward) implementation of the library's [`PDFDownloadLink`](https://react-pdf.org/advanced#download-link) component:
1. It automatically generates the PDF upon initial render as opposed to waiting for you to click the download button. This could cause a VC table with multiple certificates to all to generate PDFs immediately. Not a crisis, but it does seem like an expensive strain just for a page view.
2. Since the flow is so declarative, we cede some control over error handling (we would be yielded an `error` object and could run our reporting before rendering some error state, but this would presumably be before the user even clicks download which feels less logical to me. We would now have a more logical - a. user clicked download b. pdf generation failed - flow)

The downside is that we force ourselves to do some busy work to download the certificate via the phantom `<a>` tag et al. 

But using the `pdf` to imperatively handle the render/download flow allows us to only run this render once the user intends to download the certificate, and _arguably_, handle the reporting & error state more cleanly.

A counter-argument I can think of is that perhaps we'd _rather_ know about any failed PDF generation via analytics, even if it wasn't prompted by a user clicking download. The strain on the browser to 'eager-generate' PDF's is perhaps both worthwhile and not expensive enough to worry about for now.

### Relevant tickets

References [Pivotal #172439408](https://www.pivotaltracker.com/story/show/172439408).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Credits/CertificateDownloadButton.js
+++ b/resources/assets/components/pages/AccountPage/Credits/CertificateDownloadButton.js
@@ -13,15 +13,19 @@ const CertificateDownloadButton = ({ certificatePost }) => {
     if (!isPdfGenerated) {
       setIsPdfGenerated(true);
 
-      pdf(<CertificateTemplate certificatePost={certificatePost} />)
-        .toBlob()
-        .then(blob => {
-          pdfLink.current.href = URL.createObjectURL(blob);
-          pdfLink.current.click();
-        })
-        .catch(error => {
-          console.error(error);
-        });
+      try {
+        pdf(<CertificateTemplate certificatePost={certificatePost} />)
+          .toBlob()
+          .then(blob => {
+            pdfLink.current.href = URL.createObjectURL(blob);
+            pdfLink.current.click();
+          })
+          .catch(error => {
+            console.error(error);
+          });
+      } catch (error) {
+        console.error(error);
+      }
     }
   };
 
--- a/resources/assets/components/pages/AccountPage/Credits/CertificateDownloadButton.js
+++ b/resources/assets/components/pages/AccountPage/Credits/CertificateDownloadButton.js
@@ -6,19 +6,19 @@ import CertificateTemplate, {
 } from './CertificateTemplate';
 
 const CertificateDownloadButton = ({ certificatePost }) => {
-  const [isPdfGenerated, setIsPdfGenerated] = useState(false);
+  const [hasGeneratedPdf, setHasgeneratedPdf] = useState(false);
   const pdfLink = useRef(null);
 
   const handleClick = () => {
-    if (!isPdfGenerated) {
-      setIsPdfGenerated(true);
-
+    if (!hasGeneratedPdf) {
       try {
         pdf(<CertificateTemplate certificatePost={certificatePost} />)
           .toBlob()
           .then(blob => {
             pdfLink.current.href = URL.createObjectURL(blob);
             pdfLink.current.click();
+
+            setHasgeneratedPdf(true);
           })
           .catch(error => {
             console.error(error);
@@ -46,7 +46,7 @@ const CertificateDownloadButton = ({ certificatePost }) => {
       download="certa.pdf"
       ref={pdfLink}
       className="btn w-full py-4 text-lg bg-blue-500 hover:bg-blue-300"
-      onClick={() => handleClick()}
+      onClick={handleClick}
     >
       Download
     </a>
--- a/resources/assets/components/pages/AccountPage/Credits/CertificateDownloadButton.js
+++ b/resources/assets/components/pages/AccountPage/Credits/CertificateDownloadButton.js
@@ -1,30 +1,48 @@
 import React from 'react';
-import classNames from 'classnames';
-import { PDFDownloadLink } from '@react-pdf/renderer';
+import { pdf } from '@react-pdf/renderer';
 
 import CertificateTemplate, {
   certificatePostType,
 } from './CertificateTemplate';
 
-// The certificate PDF Download button with pending/ready state.
-const buttonClassNames = 'btn w-full py-4 text-lg';
-const CertificateDownloadButton = ({ certificatePost }) =>
-  certificatePost.pending ? (
-    <button type="button" disabled className={buttonClassNames}>
-      Pending
-    </button>
-  ) : (
-    <PDFDownloadLink
-      document={<CertificateTemplate certificatePost={certificatePost} />}
-      fileName="dosomething-volunteer-credit-certificate.pdf"
-      className={classNames(
-        buttonClassNames,
-        'bg-blue-500 hover:bg-blue-300 hover:no-underline hover:text-white',
-      )}
+const CertificateDownloadButton = ({ certificatePost }) => {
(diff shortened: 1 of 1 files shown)


katiecrane:
Oooo interesting trade-offs here. I agree that creating all the PDFs and potentially getting errors from them seems unnecessary if no one is necessarily actively trying to download them. But the phantom stuff also seems not ideal - is the only way to download the PDF to click a link? Is there anyway we can bypass that part?

mendelB:
> Oooo interesting trade-offs here. I agree that creating all the PDFs and potentially getting errors from them seems unnecessary if no one is necessarily actively trying to download them. But the phantom stuff also seems not ideal - is the only way to download the PDF to click a link? Is there anyway we can bypass that part?

As far as I can tell the phantom link seems to be the best way to _download_ the pdf. We can use `window.open` to open it in a new tab, and there seem to be various hacky approaches to triggering the download (via phantom `iframe`s and whatnot). But this def seemed like the relatively cleanest approach!

weerd:
_Disclaimer: I know very little about how all this works and haven't been directly involved in convos about it..._

The above being said, I wonder if we can do something like how we used to treat some components we didn't necessarily want to load immediately, and instead lazy loaded the component when needed? (I forget the specific example, maybe with the quiz or something).

What if instead of having the PDF Download Link in the table of certificates to instead have a "Generate PDF To Download" button. Clicking that button then dynamically grabs the PDF generating component (all while a spinner shows), and when it's ready it replaces the button with a "PDF Download Link". Then clicking on that takes the user to the generated PDF for download. I think this would keep the page snappy, since it's never generating a PDF unless it's triggered by the user.

@mendelB This is a thought 🤔, but I may have it all wrong how the pdf library works 😬

mendelB:
> _Disclaimer: I know very little about how all this works and haven't been directly involved in convos about this volunteer credits PDF generation..._
> 
> The above being said, I wonder if we can do something like how we used to treat some components we didn't necessarily want to load immediately, and instead lazy loaded the component when needed? (I forget the specific example, maybe with the quiz or something).
> 
> What if instead of having the PDF Download Link in the table of certificates to instead have a "Generate PDF To Download" button. Clicking that button then dynamically grabs the PDF generating component (all while a spinner shows), and when it's ready it replaces the button with a "PDF Download Link". Then clicking on that takes the user to the generated PDF for download. I think this would keep the page snappy, since it's never generating a PDF unless it's triggered by the user.
> 
> @mendelB This is a thought 🤔, but I may have it all wrong how the pdf library works 😬

Hmm, I think the issue here would be that we're introducing a second click in the UX to download the certificate. Do you think that would be worth the tradeoff of avoiding the phantom `<a>` tag? 

But building off of your train of thought here and @katiecrane's comment above, I wonder if we might make this cleaner by combining the two approaches:
1. only render the `PDFDownloadLink` _once_ the user clicks 'download'
2. use a [`ref`](https://reactjs.org/docs/refs-and-the-dom.html) to trigger a click event on this component to download the PDF (and introduce loading/error states)

This way, we avoid generating the PDF immediately, but also avoid introducing our own logic to generate the PDF and 'click' a phantom `<a>` tag.

If this makes sense, I can try to see if this is feasible, but I think it is! And the only custom logic we'd be introducing would be auto-clicking the rendered PDF link.

weerd:
> If this makes sense, I can try to see if this is feasible, but I think it is! And the only custom logic we'd be introducing would be auto-clicking the rendered PDF link.

Ya, I think this is worth a shot! And if it's not feasible... the fallback could be the slightly less ideal UX of clicking the link twice? I don't think it's the end of the world, especially further highlighting how "custom" the generated file is for the user 😉

mendelB:
I implemented per https://github.com/DoSomething/phoenix-next/pull/2063#issuecomment-619169371 in https://github.com/DoSomething/phoenix-next/pull/2063/commits/7e29d7f92f9822429e6ea5261abf9d6094038eab.

The advantage here is that we're no longer creating a phantom link but rather manipulating the existing one in the dom and retroactively assigning the certificate blob as it's href and triggering the click. 

LMK watcha think!

katiecrane:
Current solution seems like a best of both worlds approach to me! 👍

resources/assets/components/pages/AccountPage/Credits/CertificateDownloadButton.js
```diff
@@ -1,30 +1,57 @@
...
+      role="link"
+      tabIndex="0"
+      download="certa.pdf"
+      ref={pdfLink}
+      className="btn w-full py-4 text-lg bg-blue-500 hover:bg-blue-300"
+      onClick={() => handleClick()}
     >
       Download
-    </PDFDownloadLink>
+    </a>
```
weerd, line 52: Should this not just be a `<button>`?

weerd: Oh wait, cause you're setting the `href` in the `handleClick`?

resources/assets/components/pages/AccountPage/Credits/CertificateDownloadButton.js
```diff
@@ -1,30 +1,57 @@
...
-        buttonClassNames,
-        'bg-blue-500 hover:bg-blue-300 hover:no-underline hover:text-white',
-      )}
+const CertificateDownloadButton = ({ certificatePost }) => {
+  const [isPdfGenerated, setIsPdfGenerated] = useState(false);
+  const pdfLink = useRef(null);
+
+  const handleClick = () => {
+    if (!isPdfGenerated) {
+      setIsPdfGenerated(true);
```
weerd, line 14: I wonder if this should be named `hasGeneratedPdf`... setting it to `true` here reads weird to me, since technically nothing is generated yet. If named `hasGeneratedPdf` then you can set it to `true` if the `try` is successful.

resources/assets/components/pages/AccountPage/Credits/CertificateDownloadButton.js
mendelB: Exactly! And this way we avoid the phantom link. So there is still manipulation happening but hopefully `ref`ing an existing node makes this a lot more straightforward?

resources/assets/components/pages/AccountPage/Credits/CertificateDownloadButton.js
mendelB: Yeah, that makes better sense!

mendelB: Updated in https://github.com/DoSomething/phoenix-next/pull/2063/commits/6ab8c7b72a9c9a04d11d9dc9f466f20c21d99079

PR #2066, opened 2020-04-25
Starts on Alpha VoterRegistrationDrivePage

aaronschachter:
### What's this PR do?

This pull request starts to give a little more structure to the new OVRD effort, refactoring the current `resources/assets/components/pages/VoterRegistrationDrive` directory into `Alpha` and `Beta` directories, mimicking our general [ReferralPage](https://github.com/DoSomething/phoenix-next/tree/master/resources/assets/components/pages/ReferralPage) structure.

This PR also adds a new feature flag, `voter_reg_alpha_page`, that will override action page blocks of the Ready Set Vote campaign with the `AlphaPage` component that this PR also introduces.

Take a peek at the code, and/or pull this branch down and manually test.

* Without the feature flag set, the OVRD action page (`/us/campaigns/online-registration-drive/action`) should display its Contentful blocks, as per current behavior

* With the feature flag set, the OVRD action page  (`/us/campaigns/online-registration-drive/action`) should display the `pages/VoterRegistrationDrive/AlphaPageContainer` component.

We're hardcoding the alpha page to easily support few one-off features we need:

* Displaying a list of the alpha's voter registration referrals, to be built out further

* Modifying a `SocialDriveAction` to:
    * Allow user to select a list of "causes", which will change the query string parameter of the share action URL
    * Display a "preview" link instead of the `Embed` UI with a thumbnail, title, and description 

It'll be quicker to launch by hardcoding instead of adding new Contentful fields/GraphQL  for functionality we may choose to remove completely from the site after the elections. This should hopefully err on the side of less tech debt -- should we want to roll out some of these features to be re-usable, we can do so after we launch a v1 (hopefully this sprint!)

Future PR's will include documentation and tests for all of this, but opening as-is to hopefully help keep this reviewable. This isn't going live anytime soon, so not including screenshots for UI.

### Relevant tickets

References [Pivotal #172417306](https://www.pivotaltracker.com/story/show/172417306).

### Checklist


resources/assets/components/pages/CampaignPage/CampaignPageContent.js
```diff
@@ -79,6 +91,9 @@ const CampaignPageContent = props => {
 
 CampaignPageContent.propTypes = {
   campaignEndDate: PropTypes.string,
+  location: PropTypes.shape({
```
Dinnall, line 94: Hey @aaronschachter curious to know why you listed `location` as propTypes and defaultProps. 
Shouldn't it just be included as propTypes?

aaronschachter: We need `location` defined in `defaultProps` as well as `PropTypes` because it's not being added as required here. Without it added into `defaultProps`, our linter will throw an error:
```
 propType "location" is not required, but has no corresponding defaultProps declaration  react/require-default-props
```

Dinnall: Gotcha, thanks for sharing

PR #2052, opened 2020-04-22
Allow editors to override the action displayed in CampaignInfoBlock

katiecrane:
# What's this PR do?

This pull request allows editors to override the action displayed in `CampaignInfoBlock`. Context on why we want to do that is in the Pivotal card.

Since this is more of a one-off, we are using `additionalContent`. An editor can override the action displayed by including something like: `"actionToDisplay": 3,` in the Additional Content field of a Campaign. Then, the information for that action will be displayed in the `CampaignInfoBlock`.

Before:
Screen Shot 2020-04-22 at 12 15 53 PM

After with `"actionToDisplay": 3,`:
Screen Shot 2020-04-22 at 3 06 28 PM

### How should this be reviewed?

Did I cover all the cases?

### Any background context you want to provide?

All in the card!

### Relevant tickets

References [Pivotal #172316942](https://www.pivotaltracker.com/story/show/172316942).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2077, opened 2020-04-29
Have action id override apply to scholarship modal

katiecrane:
### What's this PR do?

This pull request builds on the work from https://github.com/DoSomething/phoenix-next/pull/2052 so that if an editor has added an action id override to be displayed in the `CampaignInfoBlock`, it will carry over to the `ScholarshipInfoBlock` as well!

Before this change (note that action type is DIFFERENT in the `CampaignInfoBlock` and `ScholarshipInfoBlock`):
before

After the change (note that action type is THE SAME in the `CampaignInfoBlock` and `ScholarshipInfoBlock`): 
after

Does the logic around choosing the action still make sense?

See the pivotal card! Basically, this is our workaround to better support multiple actions in this area.

### Relevant tickets

References [Pivotal #172316942](https://www.pivotaltracker.com/story/show/172316942).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
+++ b/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
@@ -48,7 +48,7 @@ const SCHOLARSHIP_INFO_QUERY = gql`
 `;
 
 const ScholarshipInfoBlock = ({
-  actionToDisplay,
+  actionIdToDisplay,
   affiliateSponsors,
   campaignId,
   children,
@@ -101,8 +101,8 @@ const ScholarshipInfoBlock = ({
   // Decide which action to display
   let actionItem;
 
-  if (actionToDisplay) {
-    actionItem = actions.find(action => action.id === actionToDisplay);
+  if (actionIdToDisplay) {
+    actionItem = actions.find(action => action.id === actionIdToDisplay);
   } else {
     actionItem = actions.find(
       action => action.scholarshipEntry && action.reportback,
@@ -285,7 +285,7 @@ const ScholarshipInfoBlock = ({
 };
 
 ScholarshipInfoBlock.propTypes = {
-  actionToDisplay: PropTypes.number,
+  actionIdToDisplay: PropTypes.number,
   affiliateSponsors: PropTypes.arrayOf(PropTypes.object),
   campaignId: PropTypes.number,
   children: PropTypes.object,
@@ -298,7 +298,7 @@ ScholarshipInfoBlock.propTypes = {
 };
 
 ScholarshipInfoBlock.defaultProps = {
-  actionToDisplay: null,
+  actionIdToDisplay: null,
   affiliateSponsors: [],
   campaignId: null,
   children: null,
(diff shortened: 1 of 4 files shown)


resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -95,9 +97,18 @@ const ScholarshipInfoBlock = ({
...
+  let actionItem;
+
+  if (actionToDisplay) {
+    actionItem = actions.find(action => action.id === actionToDisplay);
+  } else {
+    actionItem = actions.find(
+      action => action.scholarshipEntry && action.reportback,
+    );
+  }
+
```
DFurnes, line 111: Since we assign `actionItem = actions.find(…)` in both places, I wonder if it's clearer to just swap our method?

```js
const findActionStrategy = actionToDisplay
  ? (action) => action.id === actionToDisplay
  : (action) => action.scholarshipEntry && action.reportback;

const actionItem = actions.find(findActionStrategy);
```

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -47,6 +48,7 @@ const SCHOLARSHIP_INFO_QUERY = gql`
 `;
 
 const ScholarshipInfoBlock = ({
+  actionToDisplay,
```
DFurnes, line 51: Do you think `actionId` might be a clearer prop (indicating that it's a specific ID, and not an object or bool)?

katiecrane: Like `actionIdToDisplay`? I thought about that! Do you like that better?

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
katiecrane: Oh yeah we could do that here since there are only two options! Ended up with it [this way](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js#L59-L71) in `<CampaignInfoBlock>` where we have 3 possible routes: 

```
    // Decide which action to display
        let actionItem;

        if (actionToDisplay) {
          actionItem = actions.find(action => action.id === actionToDisplay);
        } else {
          actionItem = actions.find(
            action => action.reportback && action.scholarshipEntry,
          );
        }
        if (!actionItem) {
          actionItem = actions.find(action => action.reportback);
        }

```

DFurnes: Ah, got it! I'm good keeping the same format since these are such similar components. (And really, makes me think that we need to extract this filtering logic into Rogue/GraphQL).

katiecrane: oooooooooooo I like that thought!

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
katiecrane: I like the idea of keeping it with the same name as the field we're using in `additionalContent` (seems we do that for most things) and for that having just `actionId` isn't specific enough I don't think - I can update to `actionIdToDisplay`!

katiecrane: udpated in https://github.com/DoSomething/phoenix-next/pull/2077/commits/5cbddd74776ba1080759865978b0db816b2a82e0 !

PR #2062, opened 2020-04-24
Ovrd Herotemplating

lindsayrmaher:
### What's this PR do?

This pull request adds the Hero Image to the OVRD page for the Beta users!

### How should this be reviewed?

👀 

### Any background context you want to provide?

Starting the work to build out this template.

paired with @Dinnall !!

### Relevant tickets

References [Pivotal # 172439286](https://www.pivotaltracker.com/story/show/172439286).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2028, opened 2020-04-08
Starts on Voter Registration Drive Page

aaronschachter:
### What's this PR do?

This pull request starts on a new member online voter registration drive (OVRD) page,  which will eventually deprecate the version we currently host on Instapage (e.g. here's a link to my [Instapage-hosted OVRD](https://vote.dosomething.org/member-drive?userId=5547be89469c64ec7d8b518d&r=user:5547be89469c64ec7d8b518d,campaignID:7059,campaignRunID:8128,source:web,source_details:onlinedrivereferral,referral=true), and then here's a link to the [in-house Phoenix version this PR introduces](https://dosomething-phoenix-de-pr-2028.herokuapp.com/us/members/55767606a59dbf3c7a8b4571/voter-registration-drive)).

It's building from [this design](https://app.abstract.com/projects/fadf0790-f1f9-11e6-9a30-b54ba7e8e705/branches/a02a32be-978a-44ec-a73f-006d65a9744f/commits/2cc30a2724a419eb02f10ed26a706c11853af3fc/files/05f5faae-3c13-4947-b624-f9e0ea97822c/layers/D2B292D9-A0F6-4522-87D0-6108286A40E0?collectionId=1db495fa-8566-4250-b453-da4713ad90bb&collectionLayerId=c0c72a51-942f-4685-9694-4427cc45d593), and isn't ready for public use yet -- but opening this PR now to get the groundwork in.

### How should this be reviewed?

Not adding screenshots as these designs aren't final yet. Can take a 👀  and make sure code all looks good so far (keeping in mind this isn't ready to go live, or to accidentally navigate to)

👀 
https://dosomething-phoenix-de-pr-2028.herokuapp.com/us/members/55767606a59dbf3c7a8b4571/voter-registration-drive

OVRD page

### Any background context you want to provide?

Creating a separate page component that borrows from campaign styles felt like the best approach to go with, as this isn't exactly a campaign page. We eventually want to add a form with email + zip text fields into the Step 1 to send over to the Rock the Vote registration form, so I think we're better off hardcoding all of this vs trying to create blocks in Contentful that would only be used for this page. 

### Relevant tickets

References [OVRD Product Brief](https://docs.google.com/document/d/1_fklMzqD09uiGqu__bOVyOWz3zRqu4SEpwvxPOwmLYQ/edit) -- we don't have Pivotal tickets yet as this was to prototype to get a feel for whether we can migrate off Instapage -- which per design review today is a "Yes, we can!"

### Checklist

[...]

PR #2080, opened 2020-04-30
Fixes Beta OVRD page on dev

aaronschachter:
### What's this PR do?

The dev beta OVRD page is currently displaying an ErrorPage with  `TypeError: Cannot read property 'campaignWebsite' of null` per changes in #2062 

* https://dev.dosomething.org/us/my-voter-registration-drive?referrer_user_id=55767606a59dbf3c7a8b4571

This pull request fixes a mistake I made in #2028, where I was inspecting `process.env.NODE_ENV` to determine whether we're using a dev environment. I was following by example over [here](https://github.com/DoSomething/phoenix-next/blob/0b1534db613cdfd65d702eec2097a7bb5a531c10/resources/assets/store/store.js#L24). 

I see a few other examples where we're checking for the `window.ENV.APP_ENV`, which could have been the fix for the [broken dev beta page](https://github.com/DoSomething/phoenix-next/pull/2062#issuecomment-620860305)- but thought it'd be useful to add a new helper, `isDevEnvironment` to determine whether we're in a dev environment, where all of our Contentful, Action, User, Everything ID's don't match QA / production.

Then, in looking to fix the broken beta OVRD page, I noticed a happy coincidence that the Contentful ID is the same in dev and non-dev environments for our OVRD campaign because it's such an old school entry. And figured this was a good PR to additionally start some documentation on all things old and new OVRD.

Verify the review app's beta page loads correctly:

* https://dosomething-phoenix-de-pr-2080.herokuapp.com/us/my-voter-registration-drive?referrer_user_id=55767606a59dbf3c7a8b4571

I created a new dev Rogue campaign with ID 9035 called Test Voter Registration Campaign, so it'll be much easier to create our own test actions and not worry about reusing a campaign that any other dev Contentful entries might be referencing. I changed our dev OVRD campaign to use this campaign ID instead of 9004 (which is used by multiple dev Contentful entries)

### Relevant tickets

References [Pivotal #172439286](https://www.pivotaltracker.com/story/show/172439286).

### Checklist

### DIFF
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
@@ -61,7 +61,7 @@ const BetaVoterRegistrationDrivePage = () => {
   }
 
   if (error) {
-    return <ErrorPage />;
+    return <ErrorPage error={error} />;
   }
 
   if (!data.user) {
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
@@ -4,7 +4,7 @@ import { useQuery } from 'react-apollo';
 
 import Faq from './Faq';
 import ErrorPage from '../../ErrorPage';
-import { query } from '../../../../helpers';
+import { isDevEnvironment, query } from '../../../../helpers';
 import NotFoundPage from '../../NotFoundPage';
 import Placeholder from '../../../utilities/Placeholder';
 import ButtonLink from '../../../utilities/ButtonLink/ButtonLink';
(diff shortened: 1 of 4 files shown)


resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
```diff
@@ -61,26 +61,25 @@ const BetaVoterRegistrationDrivePage = () => {
...
+   * TODO: Add campaignId, scholarshipAmount, and scholarshipDeadline as available properties
+   * to our CampaignWebsite type in GraphQL to avoid hardcoding.
    *
-   * 04/28/20 - keeping the scholarship info hard coded for v1 and hiding details link
+   * For now, we're hiding the scholarship details.
    */
   const campaignInfoBlock = (
     <CampaignInfoBlock
-      campaignId={voterRegistrationDriveCampaignId}
+      campaignId={isDevEnvironment ? 9035 : 9054}
```
aaronschachter, line 82: Feels worth noting that I'd specifically like to avoid the `featureFlag` helper, because this is really more of a site configuration than a feature flag. I'd warm up to setting these via config variables if we could rename the `featureFlag` helper to something like `environmentVar` or `siteConfig`...

aaronschachter: Also, we won’t need to hardcore these ids at all if https://github.com/DoSomething/graphql/pull/218 is merged.

resources/assets/helpers/index.js
```diff
@@ -205,6 +205,15 @@ export function isAuthenticated() {
...
 
+/**
+ * Check to see if running in a development enviroment.
+ *
+ * @return {Boolean}
+ */
+export function isDevEnvironment() {
+  return ['local', 'development'].includes(get(window.ENV, 'APP_ENV', 'local'));
+}
+
```
mendelB, line 216: 👏 handy!

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
mendelB: Agreed! And I feel like it might make sense to have both for clarity? The `featureFlag` helper for config variables under the `env.FEATURE_FLAGS` umbrella configured in `/config/features` for feature flags, and an additional `environmentVar/siteCofig` for non-feature-flag configuration variables?

lindsayrmaher:
LOVE! Love the helper, love the change to just directly query the campaignWebsite info. 👏

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
Dinnall: I love this suggestion!

PR #2084, opened 2020-05-01
Adds Contentful ID's to Beta OVRD 

aaronschachter:
### What's this PR do?

This pull request continues #2083 by adding a config of dev and production Contentful ID's, which I've manually created, to render `ContentBlock` entries on the beta OVRD page, to avoid hardcoding content. Also adds some additional changes:

* Fixes a bug I missed in https://github.com/DoSomething/phoenix-next/pull/2083#discussion_r418319760

* Adds an `id` prop in a `NotFound` component and displays it. I was confused about the many `NotFound` components I saw once #2083 was deployed and I ran a Contentful QA refresh -- but it was because the QA GraphQL cache didn't contain the new production Contentful ID's. This should help troubleshoot gotchas, a la #2068

### How should this be reviewed?

👀 

Not adding screenshots for all screens as this uses existing components.

### Any background context you want to provide?

🐶 🥩 

### Relevant tickets

References [Pivotal #172440078](https://www.pivotaltracker.com/story/show/172440078).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2083, opened 2020-04-30
Adds total scholarship entries + ContentBlocks to Alpha OVRD

aaronschachter:
### What's this PR do?

This pull request gets the new alpha OVRD page in pretty good shape for a launch, making a few changes per the [design](
https://app.abstract.com/projects/fadf0790-f1f9-11e6-9a30-b54ba7e8e705/branches/a02a32be-978a-44ec-a73f-006d65a9744f/collections/72c40c1f-06db-4544-8724-69c62d9224e3):

* Modifies the `SocialDriveAction` sidebar to display total scholarship entries instead of existing page views/referrals count. 

* Hardcodes the id's of `ContentBlock` entries to use for the "Share with Friends" and "FAQs" sections - I've manually created these ID's in dev and production (and will run a Contentful QA refresh to populate the production ID's on QA for stakeholder review tomorrow )

### How should this be reviewed?

👀 

This uses existing components, so not going to add screenshots for all screen sizes.

### Any background context you want to provide?

🧶 🐈 

### Relevant tickets

References [Pivotal #172423982](https://www.pivotaltracker.com/story/show/172423982).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Documentation added for new features/changed endpoints.
- [x] Added screenshots of front-end changes on small, medium, and large screens.
- [x] Added appropriate feature/unit tests.

PR #2085, opened 2020-05-01
Adds comments and list of voting reasons to OVRD config

aaronschachter:
### What's this PR do?

An upcoming feature for the OVRD pages is allowing the alpha to customize their beta page with the causes they care about, so that they can better explain their motivation for voting, and why their friends should vote too (plagiarizing Diana's elegant description in [#172087475](https://www.pivotaltracker.com/story/show/172087475)).

This PR sets up the groundwork for adding implementation details into our Pivotal stories, adding:

* The voting reason options to display on the alpha's `SocialDriveAction` to customize their share URL, indexed by the value to include in a `voting-reasons` query parameter.

* The description for each `voting-reason` query value to display within the quote on the beta's page 

* Clarifying comments for each item in the OVRD configs that were added in #2083 and #2084 

It also changes our beta's `HeroSection` to display the default quote when no causes have been selected, per [#172087475 comment](https://www.pivotaltracker.com/story/show/172087475/comments/213970701).

👀 

* We've been calling them causes a lot in the voter reg team when discussing this feature, but explicitly not calling them causes in code here, as they don't relate to our `CausePage` world.

* We could use a single `votingReasons` config and change the case of leading characters on the beta side, but because COVID should remain capitalized -- seems straight forward enough to keep them separate in case we'd ever want to further customize the text of each reason on the different alpha/beta views.

### Relevant tickets

References [Pivotal #172087475](https://www.pivotaltracker.com/story/show/172087475).

### Checklist

### DIFF
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/config.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/config.js
@@ -21,7 +21,10 @@ export const voterRegistrationDriveCampaignLink = {
     : '30rCn63G1rnpzojCXC9PmF',
 };
 
-// Map of query string parameters and the voting reasons to display in the hero section quote.
+/**
+ * Map of the 'voting-reasons` query values, and the corresponding voting reason to display.
+ * e.g. /my-voter-registration-page?voting-reasons=gun-violence,immigration-reform&referrer_user_id=583d...
+ */
 export const votingReasons = {
   'climate-change': 'climate change',
   'gun-violence': 'gun violence',
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/config.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/config.js
@@ -1,19 +1,33 @@
 import { isDevEnvironment } from '../../../../helpers';
 
 export const faq = {
+  // ContentBlock used for FAQ section.
   contentBlockId: isDevEnvironment()
     ? '3cXc0RPMVNeE4surEqFujL'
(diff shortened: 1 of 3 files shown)


resources/assets/components/pages/VoterRegistrationDrivePage/Beta/config.js
```diff
@@ -1,19 +1,36 @@
...
   contentBlockId: isDevEnvironment()
     ? '3p2qz2JPCvgVitgRVBoMFz'
     : '30rCn63G1rnpzojCXC9PmF',
 };
+
+/**
+ * Map of the 'voting-reasons` query values, and the corresponding voting reason to display.
+ * e.g. /my-voter-registration-page?voting-reasons=gun-violence,immigration-reform&referrer_user_id=583d...
+ */
+export const votingReasons = {
```
lindsayrmaher, line 28: do we need to have this in the Beta config as well? since we could grab them from the same object we are using for the alpha page. or did you put them here to not end up using the alpha config for the beta page.

aaronschachter: I mentioned this in "background context" section - it is intentional to keep separate. We could use the alpha configs and manipulate the string values to change the case of each word... but it does seem like an extra complexity for only 7 values. I liked that this approach gives us the flexibility to change either alpha or beta text independently should we need to.

aaronschachter: We can always revisit if we want a single definition -- if we did, i'd prefer a root level `pages/VoterRegistrationDrive/config.js` vs the beta referencing the alpha config.. which is also why it felt cleaner to just define each separately.

PR #2086, opened 2020-05-01
Updates and cleanup for layout.

weerd:
### What's this PR do?

This pull request does a little bit of cleanup and fixes an oversight in my recommendation in [#2034](https://github.com/DoSomething/phoenix-next/pull/2034#issuecomment-613025051).

I tested to make sure this doesn't affect the new SitewideBanner component and it functions as intended, all while still allowing pages (like the 404 page) that have minimal content to span the full height of the viewport.

Along with that, I include a couple other tweaks as well.

👀 

### Relevant tickets

🌵

### DIFF
--- a/resources/assets/components/App.js
+++ b/resources/assets/components/App.js
@@ -52,6 +52,7 @@ const App = ({ store, history }) => {
             )}
           />
         ) : null}
+
         <ApolloProvider client={graphqlClient(env('GRAPHQL_URL'))}>
           {featureFlag('sitewide_nps_survey') &&
           window.location.pathname !== '/us' ? (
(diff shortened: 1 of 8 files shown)


resources/assets/components/App.js
```diff
@@ -76,50 +77,62 @@ const App = ({ store, history }) => {
...
               <Route
                 path="/us/my-voter-registration-drive"
                 component={BetaVoterRegistrationDrivePage}
               />
+
               <Route
                 path="/us/refer-friends"
                 component={AlphaReferralPageContainer}
               />
+
```
weerd, line 135: This poor file needed some breathing room between elements 😌

PR #2087, opened 2020-05-01
Adding all feature flag env variables

weerd:
### What's this PR do?

This pull request simply updates the "Feature Flags" section of the `.env.example` file to account for all the existing feature flags so we can easily reference to add any missing ones into our local `.env` files.

👀 

### Relevant tickets

🌵

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -71,13 +71,17 @@ ZENDESK_SUBDOMAIN=
 ZENDESK_USERNAME=
 ZENDESK_TOKEN=
 
-#
+# ==============================================================================
 # Feature Flags
-#
-DS_ENABLE_NPS_SURVEY=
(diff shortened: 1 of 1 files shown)


.env.example
```diff
@@ -71,13 +71,17 @@ ZENDESK_SUBDOMAIN=
 ZENDESK_USERNAME=
 ZENDESK_TOKEN=
 
-#
+# ==============================================================================
 # Feature Flags
-#
-DS_ENABLE_NPS_SURVEY=
-DS_ENABLE_VOTER_REG_MODAL=
+# ==============================================================================
```
weerd, line 76: If we're adding a visual separation for the "Feature Flags header, lets REALLY go for it! 👀

PR #2089, opened 2020-05-01
Button updates and sitewide banner update

weerd:
### What's this PR do?

This pull request further updates buttons and our new components.

It renames the `data` prop to `attributes` to indicate that other attributes apart from `data-*` attributes can be passed to our new button components. This was necessary to allow passing other attributes like the `target` and `rel` attributes in link buttons, without needing to resort to creating new props for each possible attribute, which would be daunting.

This PR also updates the button in the `SitewideBannerContent` component. While updating it, I also took the liberty to address the close button size and tweak a bit of the styles.

Does this `attributes` prop make sense?

### Relevant tickets

References [Pivotal #172413748](https://www.pivotaltracker.com/story/show/172413748).

### Checklist

### DIFF
--- a/resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
+++ b/resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
@@ -40,12 +40,12 @@ const FilterNavigation = ({ filters, setFilters }) => {
 
         {filterCategoryNames.map(name => (
           <SecondaryButton
-            key={`${name}_button`}
+            attributes={{ 'data-filter': name }}
             className="mr-8"
-            data={{ 'data-filter': name }}
+            isActive={activeFilter === name}
+            key={`${name}_button`}
             onClick={handleMenuToggle}
             text={startCase(name)}
-            isActive={activeFilter === name}
           />
         ))}
       </div>
--- a/resources/assets/components/utilities/SitewideBanner/SitewideBannerContent.js
+++ b/resources/assets/components/utilities/SitewideBanner/SitewideBannerContent.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import PrimaryButton from '../Button/PrimaryButton';
 import CloseButton from '../../artifacts/CloseButton/CloseButton';
 import {
   EVENT_CATEGORIES,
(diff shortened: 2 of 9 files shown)


resources/assets/components/utilities/SitewideBanner/SitewideBannerContent.js
```diff
@@ -25,25 +26,24 @@ const SitewideBannerContent = ({
...
+          className="py-2 px-4"
           href={link}
-          target="_blank"
-          rel="noopener noreferrer"
           onClick={handleCompleteWithTracking}
-        >
-          {cta}
-        </a>
+          text={cta}
+        />
```
weerd, line 46: @lindsay I wanted to mention my reasoning behind some of these updates with the styles.

I removed the `flex-wrap` class. While this can allow for some pseudo-responsive behavior it's not ideal since it can't be controlled very well. I think it's best used when you have a series of items, that are all the same size in some sort of gallery that can be variable. But in this context it didn't seem like the ideal approach and better to just use media-queries to handle different layout changes.

Apart from that, the button padding changes also led to some padding changes in the container to help with the prior issues of the button overlapping onto the content in different sizes:

resources/views/search/results.blade.php
```diff
@@ -26,7 +26,7 @@
                         ])
                         <input class="border-none leading-none pl-10 text-input w-full" name="query" type="text" value="{{ $query }}">
                     </div>
-                    <input class="btn" type="submit" value="Search">
+                    <input class="btn bg-blurple-500 hover:bg-blurple-400" type="submit" value="Search">
```
weerd, line 29: Missed this button in my initial updates.

resources/assets/components/utilities/SitewideBanner/SitewideBannerContent.js
lindsayrmaher: These changes make sense! Also I like the addition of the attributes prop, makes it a bit more clear what's going where when the button actually renders

resources/assets/components/pages/CampaignsPage/FilterNavigation/FilterNavigation.js
```diff
@@ -40,12 +40,12 @@ const FilterNavigation = ({ filters, setFilters }) => {
 
         {filterCategoryNames.map(name => (
           <SecondaryButton
-            key={`${name}_button`}
+            attributes={{ 'data-filter': name }}
```
DFurnes, line 43: Ahh much clearer! I'd mistakenly though this was some built-in React magic before when I saw this used.

PR #2097, opened 2020-05-04
Updating more buttons and adding documentation.

weerd:
### What's this PR do?

This pull request updates a few more buttons in components to utilize the new button components.

It also starts to add some documentation for Buttons for our Phoenix docs to be further expanded upon in upcoming PRs.

👀 

While updating and replacing more of the buttons in Phoenix with the new buttons, it has provided opportunities to showcase where the new buttons need added functionality. 

This happened in particular with the share button that used the old `ButtonLink` component, which would check if the url for the `href` provided was external or not and add appropriate attributes to launch the link in a new tab. I was able to replicate the same idea in the new `LinkButton` component, but with the added bonus of now allowing it to be overriding if necessary! More in this [comment](https://github.com/DoSomething/phoenix-next/pull/2097#discussion_r420903641).

Some screenshots:

Share Action with `PrimaryButton`:

ShareAction

Share Action with `PrimaryButton` and Embed:

image

Text Submission Action with `PrimaryButton`:

image

Link Action with `PrimaryButton` & default template:

image

Another Link Action with `PrimaryButton` & default template:

image

### Relevant tickets

### Checklist


resources/assets/components/utilities/Button/LinkButton.js
```diff
@@ -2,22 +2,32 @@ import React from 'react';
...
-    {...attributes}
-  >
-    {text}
-  </a>
-);
+const LinkButton = ({ attributes, className, href, onClick, text }) => {
+  const linkAttributes = {
+    rel: isExternal(href) ? 'noopener noreferrer' : null,
+    target: isExternal(href) ? '_blank' : null,
+  };
```
weerd, line 17: While updating components that used the reversed name `ButtonLink` component, I found that that old button component would check whether a link was external and add appropriate `target` and `rel` attributes.

I included this same support in this `LinkButton` component but in a slightly different way, and the nice side-effect is that by default, it'll add appropriate `target` and `rel` attributes, but a developer can just as easily override it by passing those specific attributes in the `attributes` prop, which makes this button much more versatile!

resources/assets/components/utilities/Embed/Embed.js
```diff
@@ -77,7 +77,7 @@ const Embed = props => {
           ) : (
             <a
               href={url}
-              className="no-underline hover:no-underline"
+              className="block no-underline hover:no-underline"
```
weerd, line 80: When an `<a>` is utilized as a button or a giant clickable box (like with this embed), I find it helps avoid layout issues in the future if you specify a rule to declare it display as a `block` level item (i.e. `display: block`). 

Otherwise, `<a>`'s by default are `inline`; this can cause issues if the content inside of the _inline by-default_ `<a>` happens to be block level like a `<div>`, which is precisely the case here. Better safe than sorry!

The More You Know

weerd:
Need to fix up the Jest tests but should be ready for 👀 otherwise!

PR #2101, opened 2020-05-05
Beta step one

Dinnall:
### What's this PR do?

This pull request...
* Created a new component - `pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm`
* Use a Card component to render the yellow title bar
* Replaced our `ButtonLink` with `StartVoterRegistrationForm`

...

...

### Relevant tickets

References [https://www.pivotaltracker.com/n/projects/2417735/stories/172439461).

### Checklist

### DIFF
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
@@ -106,7 +106,6 @@ const BetaVoterRegistrationDrivePage = () => {
             <div className="pb-6">
               <StartVoterRegistrationForm
                 referrerUserId={referrerUserId}
-                blockId={registerToVote.contentBlockId}
                 campaignId={campaignId}
               />
             </div>
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
@@ -104,7 +104,11 @@ const BetaVoterRegistrationDrivePage = () => {
               className="grid-wide clearfix wrapper pb-3"
             />
             <div className="pb-6">
-              <StartVoterRegistrationForm referrerUserId={referrerUserId} />
+              <StartVoterRegistrationForm
+                referrerUserId={referrerUserId}
+                blockId={registerToVote.contentBlockId}
+                campaignId={campaignId}
+              />
             </div>
             <ContentfulEntryLoader
               id={faq.contentBlockId}
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
@@ -15,6 +15,7 @@ import Modal from '../../../utilities/Modal/Modal';
 import Placeholder from '../../../utilities/Placeholder';
 import ButtonLink from '../../../utilities/ButtonLink/ButtonLink';
 import SiteFooter from '../../../utilities/SiteFooter/SiteFooter';
+import StartVoterRegistrationForm from './StartVoterRegistrationForm';
 import SiteNavigationContainer from '../../../SiteNavigation/SiteNavigationContainer';
 import ScholarshipInfoBlock from '../../../blocks/ScholarshipInfoBlock/ScholarshipInfoBlock';
 import ContentfulEntryLoader from '../../../utilities/ContentfulEntryLoader/ContentfulEntryLoader';
@@ -103,12 +104,7 @@ const BetaVoterRegistrationDrivePage = () => {
               className="grid-wide clearfix wrapper pb-3"
             />
             <div className="pb-6">
-              {/* We will eventually want to add form fields for email and zip, and send as query parameters to Rock the Vote */}
-              <ButtonLink
-                link={`https://register.rockthevote.com/registrants/new?partner=37187&source=user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true`}
-              >
-                Register To Vote
-              </ButtonLink>
+              <StartVoterRegistrationForm referrerUserId={referrerUserId} />
             </div>
             <ContentfulEntryLoader
               id={faq.contentBlockId}
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
@@ -1,51 +1,27 @@
 import PropTypes from 'prop-types';
-import React, { useState, useRef } from 'react';
+import React, { useReducer } from 'react';
 
 import {
   EVENT_CATEGORIES,
   trackAnalyticsEvent,
 } from '../../../../helpers/analytics';
 import Card from '../../../utilities/Card/Card';
-import { buildVoterRegUrl } from '../../../../helpers/index';
 import PrimaryButton from '../../../utilities/Button/PrimaryButton';
 
-const StartVoterRegistrationForm = ({ blockId, campaignId }) => {
-  const [emailValue, setEmailValue] = useState('');
-  const [zipcodeValue, setZipcodeValue] = useState('');
-  const [emailLabel, setEmailLabel] = useState('email');
-  const [zipCodeLabel, setZipCodeLabel] = useState('zipcode');
-  const zipCodeLength = useRef(0);
-  const url = buildVoterRegUrl(
-    'web',
-    'onlinedrivereferral,referral=true',
-    'https://register.rockthevote.com/registrants/new?partner=37187&',
-    { email_address: emailValue, home_zip_code: zipcodeValue },
+const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
+  const [userInput, setUserInput] = useReducer(
+    (state, newState) => ({ ...state, ...newState }),
+    {
+      email: '',
+      zipcode: '',
+    },
   );
-  /**
-   * should we be passing the userID or referrerId ?
-   * */
 
-  // `https://register.rockthevote.com/registrants/new?partner=37187&source=user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true`;
-
-  const isDisabled = !zipcodeValue || !emailValue;
-
-  const redirectToRockTheVote = () => {
-    window.location = url;
-  };
+  const isDisabled = !userInput.zipcode || !userInput.email;
 
   const handleChange = event => {
-    if (event.target.name === 'email') {
-      () => {
-        setEmailValue(event.target.value);
-        setEmailLabel(event.target.name);
-      };
-    }
-
-    if (event.target.name === 'zipcode') {
-      () => {
-        setZipcodeValue(event.target.value);
-      };
-    }
+    const { name, value } = event.target;
+    setUserInput({ [name]: value });
   };
 
   const handleClick = () => {
@@ -55,62 +31,59 @@ const StartVoterRegistrationForm = ({ blockId, campaignId }) => {
         category: EVENT_CATEGORIES.campaignAction,
         label: 'voter_registration',
         context: {
-          blockId,
           campaignId,
-          url,
         },
       });
-
-      redirectToRockTheVote();
+      window.location = `https://register.rockthevote.com/registrants/new?partner=37187&source=user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true,email_address:${userInput.email},home_zip_code:${userInput.zipcode}`;
+      console.log('Window:', window.location);
     }
   };
 
   return (
     <>
-      <Card className="bordered rounded" title="Register Online to vote">
-        <form className="form pb-2">
-          <div>
-            <label htmlFor="zipcode">
-              {zipCodeLabel}
-              <input
-                ref={zipCodeLength}
-                className="text-field"
-                type="text"
-                name="zipcode"
-                value={zipcodeValue}
-                onChange={handleChange}
-                required
-              />
-            </label>
-          </div>
-
-          <div>
+      <Card className="bordered rounded" title="Register online to vote">
+        <form className="form p-3">
+          <div className="form-item stretched">
             <label htmlFor="email">
-              {emailLabel}
+              email
               <input
                 className="text-field"
                 required
                 type="text"
                 name="email"
+                value={userInput.email}
                 onChange={handleChange}
-                value={emailValue}
               />
             </label>
           </div>
+
+          <div className="form-item stretched">
+            <label htmlFor="zipcode">
+              zip code
+              <input
+                className="text-field"
+                type="text"
+                name="zipcode"
+                value={userInput.zipcode}
+                onChange={handleChange}
+                required
+              />
+            </label>
+          </div>
+          <PrimaryButton
+            onClick={handleClick}
+            isDisabled={isDisabled}
+            text="Register To Vote"
+          />
         </form>
-        <PrimaryButton
-          onClick={handleClick}
-          isDisabled={isDisabled}
-          text="Register To Vote"
-        />
       </Card>
     </>
   );
 };
 
 StartVoterRegistrationForm.propTypes = {
-  blockId: PropTypes.string.isRequired,
   campaignId: PropTypes.number.isRequired,
+  referrerUserId: PropTypes.string.isRequired,
 };
 
 export default StartVoterRegistrationForm;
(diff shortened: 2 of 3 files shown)


resources/assets/components/pages/VoterRegistrationDrivePage/Beta/BetaPage.js
```diff
@@ -103,12 +104,11 @@ const BetaVoterRegistrationDrivePage = () => {
...
             <div className="pb-6">
-              {/* We will eventually want to add form fields for email and zip, and send as query parameters to Rock the Vote */}
-              <ButtonLink
-                link={`https://register.rockthevote.com/registrants/new?partner=37187&source=user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true`}
-              >
-                Register To Vote
-              </ButtonLink>
+              <StartVoterRegistrationForm
+                referrerUserId={referrerUserId}
+                blockId={registerToVote.contentBlockId}
```
aaronschachter, line 109: I don't think we should send along a `blockId` here -- this `contentBlockId` doesn't contain the URL that the user clicks on.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -0,0 +1,116 @@
...
+  const zipCodeLength = useRef(0);
+  const url = buildVoterRegUrl(
+    'web',
+    'onlinedrivereferral,referral=true',
+    'https://register.rockthevote.com/registrants/new?partner=37187&',
+    { email_address: emailValue, home_zip_code: zipcodeValue },
+  );
+  /**
+   * should we be passing the userID or referrerId ?
+   * */
```
aaronschachter, line 26: We need to send the `referrerUserId` as the `user` parameter in the tracking source. I think trying to repurpose `buildVoterRegUrl` makes this way more complicated than it's worth -- let's please keep the URL defined as per commented out below, append the `email` and `home_zip_code` query variables to keep it very simple to find where we're setting this URL (that we're not linking to anywhere else).

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -0,0 +1,116 @@
...
+
+import {
+  EVENT_CATEGORIES,
+  trackAnalyticsEvent,
+} from '../../../../helpers/analytics';
+import Card from '../../../utilities/Card/Card';
+import { buildVoterRegUrl } from '../../../../helpers/index';
+import PrimaryButton from '../../../utilities/Button/PrimaryButton';
+
+const StartVoterRegistrationForm = ({ blockId, campaignId }) => {
```
aaronschachter, line 12: Let's remove `blockId` per earlier comment.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  const handleClick = () => {
+    if (!isDisabled) {
+      trackAnalyticsEvent('click_voter_registration_action', {
+        action: 'button_clicked',
+        category: EVENT_CATEGORIES.campaignAction,
+        label: 'voter_registration',
+        context: {
+          blockId,
```
aaronschachter, line 58: Again - let's remove `blockId`

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -0,0 +1,116 @@
...
+          campaignId,
+          url,
+        },
+      });
+
+      redirectToRockTheVote();
```
aaronschachter, line 64: We don't need this to be a function, since the only place in the file we're redirecting from is here. let's specifically just set `window.location` here:
```
window.location = `https://register.rockthevote.com/registrants/new?partner=37187&source=user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true&email_address...`;
```

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -0,0 +1,116 @@
...
+    }
+  };
+
+  return (
+    <>
+      <Card className="bordered rounded" title="Register Online to vote">
```
aaronschachter, line 70: Inconsistent capitalization, let's just make this "Register online to vote"

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -0,0 +1,116 @@
...
+          isDisabled={isDisabled}
+          text="Register To Vote"
+        />
+      </Card>
+    </>
+  );
+};
+
+StartVoterRegistrationForm.propTypes = {
+  blockId: PropTypes.string.isRequired,
```
aaronschachter, line 112: Remove per earlier comments.

resources/assets/helpers/index.js
```diff
@@ -1060,11 +1060,24 @@ export function getMillisecondsFromDays(days) {
...
  * @param {String} url
+ * @param {Object} userDetails
  */
 
-export function buildVoterRegUrl(source = 'web', sourceDetails, url) {
+export function buildVoterRegUrl(
+  source = 'web',
+  sourceDetails,
+  url,
+  userDetails,
```
aaronschachter, line 1070: Let's remove these changes altogether. Out of scope for this PR, but I missed this earlier -- I don't think we want a `url` parameter if we aren't using it anywhere.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -1,51 +1,27 @@
...
-    'onlinedrivereferral,referral=true',
-    'https://register.rockthevote.com/registrants/new?partner=37187&',
-    { email_address: emailValue, home_zip_code: zipcodeValue },
+const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
+  const [userInput, setUserInput] = useReducer(
+    (state, newState) => ({ ...state, ...newState }),
+    {
+      email: '',
+      zipcode: '',
+    },
```
aaronschachter, line 17: Is there a reason you're using `useReducer` here vs `useState`? Per [the docs](https://reactjs.org/docs/hooks-reference.html#usereducer): 

> useReducer is usually preferable to useState when you have complex state logic that involves multiple sub-values or when the next state depends on the previous one

I think `useState` would help readbility here, and help with code consistency (we don't use it anywhere in the repo, but [use state hooks quite a bit](https://github.com/DoSomething/phoenix-next/search?q=useState&unscoped_q=useState)). Let's also use `zip` instead of `zipcode` to keep consistency in codebase too.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -0,0 +1,89 @@
...
+      trackAnalyticsEvent('click_voter_registration_action', {
+        action: 'button_clicked',
+        category: EVENT_CATEGORIES.campaignAction,
+        label: 'voter_registration',
+        context: {
+          campaignId,
+        },
+      });
+      window.location = `https://register.rockthevote.com/registrants/new?partner=37187&source=user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true,email_address:${userInput.email},home_zip_code:${userInput.zipcode}`;
+      console.log('Window:', window.location);
```
aaronschachter, line 38: Let's remove this console logging.

resources/assets/helpers/index.js
```diff
@@ -1059,11 +1059,10 @@ export function getMillisecondsFromDays(days) {
...
- * @param {String} url
  */
 
-export function buildVoterRegUrl(source = 'web', sourceDetails, url) {
-  const baseUrl = url || 'https://vote.dosomething.org/';
+export function buildVoterRegUrl(source = 'web', sourceDetails) {
+  const baseUrl = 'https://vote.dosomething.org/';
   const userId = window.AUTH.id ? `user:${window.AUTH.id},` : '';
 
   return `${baseUrl}?r=${userId}source:${source},source_details:${sourceDetails}`;
```
aaronschachter, line 1068: Not necessary for merge, but it'd be nice to return this all inline since we only use `baseUrl` and `userId` one time:

```
return 'https://vote.dosomething.org?r=${window.AUTH.id ? `user:${window.AUTH.id},` : ''}source:${source},source_details:${sourceDetails}'`;
```

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -0,0 +1,89 @@
...
+    if (!isDisabled) {
+      trackAnalyticsEvent('click_voter_registration_action', {
+        action: 'button_clicked',
+        category: EVENT_CATEGORIES.campaignAction,
+        label: 'voter_registration',
+        context: {
+          campaignId,
+        },
+      });
+      window.location = `https://register.rockthevote.com/registrants/new?partner=37187&source=user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true,email_address:${userInput.email},home_zip_code:${userInput.zipcode}`;
```
aaronschachter, line 37: I tested on the review app with my dev user ID: https://dosomething-phoenix-de-pr-2101.herokuapp.com/us/my-voter-registration-drive?referrer_user_id=55767606a59dbf3c7a8b4571

This URL does not re-populate my email and zip -- the end should be
```
&email_address-${userInput.email}&home_zip_code=${userInput.zipcode}
```

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -0,0 +1,89 @@
...
+      console.log('Window:', window.location);
+    }
+  };
+
+  return (
+    <>
```
aaronschachter, line 43: This needs some tailwind -- Per [the designs](https://app.abstract.com/projects/fadf0790-f1f9-11e6-9a30-b54ba7e8e705/branches/a02a32be-978a-44ec-a73f-006d65a9744f/collections/1db495fa-8566-4250-b453-da4713ad90bb) this form should not stretch across the full width of the page on a large screen. You should be able to reference some of the classes we're using in the `BetaPage` or `AlphaPage` within `pages/VoterRegistrationDrivePage`- I think the 3/4 class may do the trick.

aaronschachter: This would also be an ideal opportunity to add a `data-test` attribute and add to our Cypress test (#2096) to confirm the element appears when we have a valid user.  Per our [docs](https://dosomething.gitbook.io/phoenix-documentation/installation-+-usage/usage#cypress), you can run Cypress tests by running `npm run cypress` from the command line in your local Phoenix dir.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
lindsayrmaher: @aaronschachter - We ran into issues around using `useState` and the form inputs being controlled vs uncontrolled. When the component was first rendering and we used `useState` instead of `userReducer`, we weren't able to add any input to the fields because they weren't defaulting to the empty string as expected. 

`useReducer` seemed to solve this problem for us. If you think we should stick with `useState` it might be helpful to talk through it on a call and figure out why it's not working as expected.

`
const [userInput, setUserInput] = useState({
email: ' ',
zip: ' ',
})
`

aaronschachter: I pulled this branch down as was able to get `useState` working by declaring different constants:

```
  const [email, setEmail] = useState('');
  const [zip, setZip] = useState('');
  const isDisabled = !email || !zip;
```
I do think this helps improve readability, and reducecomplexity. Could we make this change? Then we should be good to :shipit: this first pass.

aaronschachter: Later in `handleClick` we can check in `handleClick` with something like
```
(name === 'email') ? setEmail(value) : setZip(value);
```

Dinnall: It was easier for us to read and we declared one variable instead of two.
I went ahead and pushed your suggestion.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -0,0 +1,86 @@
...
+  const [email, setEmail] = useState('');
+  const [zip, setZip] = useState('');
+
+  const isDisabled = !zip || !email;
+  const handleChange = event => {
+    const { name, value } = event.target;
+    return name === 'email' ? setEmail(value) : setZip(value);
+  };
+
+  const handleClick = () => {
```
mendelB, line 21: @Dinnall 
BTW quick thought -- if we turned this into an `onSubmit` event and attached it as the `onSubmit` to the form, and added `type="submit"` to the submit button (which we should probably do anyway to capture someone hitting ENTER with the keyboard while the form is in focus), then we could take advantage of the native HTML `type="email"` to validate the email field, and maybe a simple `pattern` attribute to validate the `zip` field.` (We could event use remove the `if (!isDisabled)` condition below, and add this URL we're redirecting to as the forms `action`.)

I coullllld be wrong about some or all of this, but this _seems_ to me like it'd work and be a super easy win!

aaronschachter: This is all great suggestions -- thanks @mendelB. We discussed relying on Rock The Vote for validation over in [Slack](https://dosomething.slack.com/archives/CTVPG6L4R/p1588790880370900) -- but I'd love to see if we could get at least native email validation by making these simple changes, and agree :100: with

> added type="submit" to the submit button (which we should probably do anyway to capture someone hitting ENTER with the keyboard while the form is in focus)

@Dinnall @lindsayrmaher could we try this in the upcoming PR for Cypress tests?

lindsayrmaher: yupp! i just basically said that exact thing on slack 😂 the synergy!

lindsayrmaher: So I was testing this out. I'm running into an issue with using the traditional submit and adding the URL to redirect to as the action on the form. 

Basically @Dinnall and I chose not to use the onSubmit since this isn't really functioning in the way other traditional forms would. Since basically all we're doing is appending what the user inputs to the URL and redirecting them there. 

Some of the source UTM params get stripped off the URL and all the DoSomething styling is gone.

Screen Shot 2020-05-06 at 4 20 48 PM

mendelB: Ah totally, using the form `action` would indeed be kinda cheating. I think it's fine to keep the manual redirect in place - we get the main wins regardless!

lindsayrmaher: @aaronschachter @mendelB - Unless we re think the way the form is implemented completely, I don't think we can change the way we handle the URL redirect or form "submission".

mendelB: Hmm if the submit button is switched to `type="submit"` and the `onClick` is _only_ used as the `<form  onSubmit=...` callback, I would've expected everything to just work 🤔 The URL redirect can stay in that handle function since that's where we deviate from standard for submission in any case. 

But also, I certainly do not mean to derail this! I feel like either way these changes could be a nice refactor ticket since we should optimally run the form as natively as we can?

weerd: Just FYI, you can change the `type` for a button as @mendelB when using a `PrimaryButton`... if it renders as a `<button>` (because no `href` is passed), then it defaults to a `type` of `button` but you can override it by passing a `type` property of `submit` :)

weerd: In regards with what @mendelB mentions, I would also maybe recommend `e.preventDefault()` to stop the form from trying to submit on itself. I believe it would still try to do that even if missing the `action` attribute, but I might be wrong... just something to note!

lindsayrmaher: @weerd Ahhhh ok yeah we'll have to try that tomorrow!

PR #2106, opened 2020-05-06
New script for updating LinkActions to CTA Blocks

lkpttn:
### What's this PR do?

This is a script that will convert all Link Action entries with template type of `CTA` to a new CTA block, moving over all relevant information and links.

> "Give it a look see, i don't know" - David Furnes

This is the second half of using the new CTA blocks, not just on story pages, but everywhere on site. See [this PR](https://github.com/DoSomething/phoenix-next/pull/2104). For information about contentful-management see [documentation](https://contentful.github.io/contentful-management.js/contentful-management/1.3.0/typedef/index.html#static-typedef-Entry).

### Relevant tickets

References [Pivotal #169758942](https://www.pivotaltracker.com/n/projects/2328687/stories/169758942).

### DIFF
--- a/contentful/management-api-scripts/2020_05_06_migrate_cta_blocks.js
+++ b/contentful/management-api-scripts/2020_05_06_migrate_cta_blocks.js
@@ -1,0 +1,112 @@
+/* eslint-disable no-restricted-syntax, no-await-in-loop, no-continue */
+
+const { contentManagementClient } = require('./contentManagementClient');
+const {
+  createLogger,
+  getField,
+  withFields,
+  linkReference,
+  constants,
+  sleep,
+} = require('./helpers');
+
+const { LOCALE } = constants;
+
+const logger = createLogger('migrate_cta_blocks');
+
+contentManagementClient.init(async environment => {
+  logger.info(
+    `Running 'migrate_cta_blocks', using Contentful's '${environment.sys.id}' environment.`,
+  );
+  logger.info('Kicking things off in 5 seconds...');
+
+  await sleep(5000);
+
+  const blocks = await environment.getEntries({
+    content_type: 'linkAction',
+    'fields.template': 'cta',
+  });
+
+  for (let i = 0; i < blocks.items.length; i += 1) {
+    const linkAction = blocks.items[i];
+
+    // Find all Entries which link to the CTA-type LinkAction.
+    const linkedEntries = await environment.getEntries({
+      links_to_entry: linkAction.sys.id,
+    });
+
+    const internalTitle = getField(linkAction, 'internalTitle');
+    logger.info(
+      `\n\nProcessing ${linkedEntries.items.length} links to "${internalTitle}".`,
+    );
+
+    // Archive things that aren't linked
+    if (linkedEntries.items.length === 0) {
+      logger.info(`⇢ Archiving '${internalTitle}', not linked anywhere.`);
+      linkAction.archive();
+      continue;
+    }
+
+    // Make the new CTA block
+    const ctaBlock = await environment.createEntry(
+      'callToAction',
+      withFields({
+        internalTitle,
+        content: getField(linkAction, 'content'),
+        linkText: getField(linkAction, 'buttonText'),
+        link: getField(linkAction, 'link'),
+        template: 'Purple',
+        alignment: 'Center',
+      }),
+    );
+
+    // Publish new block
+    await ctaBlock.publish();
+
+    // Loop through and get all places where the old block is linked
+    for (let j = 0; j < linkedEntries.items.length; j += 1) {
+      const page = linkedEntries.items[j];
+      const pageTitle = getField(page, 'internalTitle');
+      const id = page.sys.id;
+
+      // Skip archived pages
+      const isPageArchived = await page.isArchived();
+      if (isPageArchived) {
+        logger.info(`⇢ Skipping '${id}', archived.`);
+        continue;
+      }
+
+      // Replace old entity with new one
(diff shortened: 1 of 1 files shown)


contentful/management-api-scripts/2020_05_06_migrate_cta_blocks.js
```diff
@@ -0,0 +1,112 @@
...
+      if (!wasPublished) {
+        logger.warn(`→ Needs review: ${updatedPage.sys.id}`);
+        continue;
+      }
+
+      const publishedPage = await updatedPage.publish();
+
+      if (publishedPage) {
+        logger.info(
+          `✔ Published "${pageTitle}": https://qa.dosomething.org/us/campaigns/${publishedPage.fields.slug[LOCALE]}`,
```
mendelB, line 104: Maybe worth grabbing the `APP_URL` or something from the `.env` for the base URL here instead of hardcoding to `qa.dosomething.org`?

mendelB: Also, I assume most pages that link to the CTA Link Actions would be article pages and not campaign pages. (We could derive this by checking the `slug` for any `/` since campaign pages slugs would be prefixed by the campaign slug. But also this is just a log after all so 🤷‍♂️)

contentful/management-api-scripts/2020_05_06_migrate_cta_blocks.js
```diff
@@ -0,0 +1,112 @@
...
+
+    const internalTitle = getField(linkAction, 'internalTitle');
+    logger.info(
+      `\n\nProcessing ${linkedEntries.items.length} links to "${internalTitle}".`,
+    );
+
+    // Archive things that aren't linked
+    if (linkedEntries.items.length === 0) {
+      logger.info(`⇢ Archiving '${internalTitle}', not linked anywhere.`);
+      linkAction.archive();
```
mendelB, line 46: Not that this holds anything back, but maybe worth `await`ing this in case it errors out asynchronously and mucks up the script flow?

contentful/management-api-scripts/2020_05_06_migrate_cta_blocks.js
```diff
@@ -0,0 +1,112 @@
...
+
+      if (publishedPage) {
+        logger.info(
+          `✔ Published "${pageTitle}": https://qa.dosomething.org/us/campaigns/${publishedPage.fields.slug[LOCALE]}`,
+        );
+      }
+    }
+
+    // Archive old LinkActions
+    linkAction.archive();
```
mendelB, line 110: I feel like we've tended to `sleep` for a sec or two after each iteration to avoid being API throttled by Contentful, but no biggie either way, I don't think.

contentful/management-api-scripts/2020_05_06_migrate_cta_blocks.js
```diff
@@ -25,6 +25,7 @@ contentManagementClient.init(async environment => {
   const blocks = await environment.getEntries({
     content_type: 'linkAction',
     'fields.template': 'cta',
+    limit: 1000,
```
Vmack, line 28: why is the limit 1000?

mendelB:
BTW, do we have any CTA `linkActions` linked in Story Page `SectionBlocks` or other Rich Text fields? (I don't recall seeing any myself, but my memory...)
https://github.com/DoSomething/phoenix-next/blob/3c5572cc32f20ef9aa44c78b108a51eb853e4330/contentful/content-types/sectionBlock.js#L95

aaronschachter:
I was noticing that our quiz uses `LinkAction` entries in the results. Will this script change those? And if it does, can you confirm it won't break anything?

lkpttn:
@aaronschachter @mendelB This script is only targeting `LinkAction`s that are using the CTA template, almost exclusively made by Jackie and Ben for use in article pages like The Breakdown. 

I did a survey and didn't find any in Story Pages and all the Link Actions for Quiz Results use the default template and won't be changed.

contentful/management-api-scripts/2020_05_06_migrate_cta_blocks.js
```diff
@@ -0,0 +1,132 @@
...
+      `\n\nProcessing ${linkedEntries.items.length} links to "${internalTitle}".`,
+    );
+
+    // Archive things that aren't linked
+    if (linkedEntries.items.length === 0) {
+      logger.info(`⇢ Archiving '${internalTitle}', not linked anywhere...`);
+      if ((await linkAction.isPublished) && (await !linkAction.isArchived)) {
+        await linkAction.unpublish();
+        logger.info(`⇢ Unpublished LinkAction...`);
+      }
```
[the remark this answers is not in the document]
lkpttn: Yeah you have to unpublish before you can archive programmatically. This ended up being more trouble than it was worth since it is so easy to archive them with the GUI.

contentful/management-api-scripts/2020_05_06_migrate_cta_blocks.js
lkpttn: 1000 is the maximum number of entries you can process according to their documentation. We only have 322 Link Actions using this template style, so upping the limit seems more effective than grabbing 100 at a time and paginating.

mendelB: Our pagination helper for the record grabs 'em [1000 at a time](https://github.com/DoSomething/phoenix-next/blob/d52505c2c52151d447d8212cc8b6f6fec6c2e524/contentful/management-api-scripts/helpers.js#L70)! I believe things like Content Blocks exist in 1000+ bulk

PR #2108, opened 2020-05-06
Removes Alpha OVRD feature flag + deprecated VoterRegistrationDriveInfo component

aaronschachter:
### What's this PR do?

This pull request is my favorite kind, one of the ones where we delete a bunch of code we no longer use 🚒 

Once #2105 is deployed to production and we set the `DS_ENABLE_VOTER_REG_ALPHA_PAGE` to `true`, we're free to remove the `VoterRegistrationDriveInfo` component entirely from the `SocialActionDrive`. It was only used in one place -- the OVRD campaign action page's blocks, and upon setting the feature flag to true - we will no longer render those blocks on the action page (we'll instead display the `pages/VoterRegistrationDrivePage/Alpha/AlphaPage` component).

This means we can remove:

* related tests and css for the `VoterRegistrationDriveInfo` component

* the `hidePageViews` field from the GraphQL query / `SocialActionDrive` prop -- this was what was used to toggle whether to display the `VoterRegistrationDriveInfo`

* the `count` variable in our `SocialActionDrive` state -- we no longer are displaying the count of page views returned from a Bertly link-shortening request

* the check for the `voter_reg_alpha_page` feature flag, as well as the feature flag itself 

I'll add the review app to Aurora and include links to dev campaigns that I've spot checked that use a `SocialDriveAction`. 👀 and confirm no breakage.

Also removes already-removed voter reg feature flags from `.env.example` that snuck back in per #2087.

### Relevant tickets

This is related to [Pivotal #172419329](https://www.pivotaltracker.com/story/show/172419329) - where we want to add new functionality to a `SocialDriveAction`. Let's get it cleaned up first, now that we're about to go live with the new OVRD Alpha Page.

### Checklist


aaronschachter:
Added the review app to make this testable:



* https://dosomething-phoenix-de-pr-2108.herokuapp.com/us/campaigns/test-school-finder-campaign/action - Standard usage of `SocialDriveAction` component, without sidebar (where the default `true` value was set for `hidePageViews` prop that this PR now removes)

* https://dosomething-phoenix-de-pr-2108.herokuapp.com/us/campaigns/online-registration-drive/action -- Our OVRD alpha page is the only instance where we're displaying the sidebar `TotalAcceptedQuantity` component (we used to instead display the now removed `VoterRegistrationDriveInfo` component that this PR removes)

mendelB:
BTW I was seeing a Not Found block in the Alpha page -- is that just because it's a review app?
image

aaronschachter:
Yup, I plan to remove the field from GraphQL once this is merged in (which I can't do until we deploy today and I enable the feature flag).

For the Not Found blocks... they sure have been buggy. Usually a hard-refresh solves -- it has to do with how we're using `window.ENV.APP_ENV` to determine whether to display our dev or production entry ID's. #2107 makes the change to display production ID's by default if  `window.ENV.APP_ENV` is not found -- which would mean on our review apps, we'll get NotFound blocks since our review apps build from dev. If you still see the NotFound blocks, can you inspect what `console.log(window.ENV)` returns in the Chrome debugger?

aaronschachter:
I see the NotFound blocks too when I visit the page, but `window.ENV.APP_DEV` looks like `development` as suspected.. which is frustrating. Sometimes hard refresh doesn't clear it (just tried) and I end up clearing cache, application data -- which is really weak sauce. Even if we set only the dev ID's via config variable, I wonder if we'd run into the same problem.

aaronschachter:
AHH I think that might be it! totally not an incoherent ramble. so checking for `window.ENV.APP_ENV` within the render ya?

mendelB:
> AHH I think that might be it! totally not an incoherent ramble. so checking for `window.ENV.APP_ENV` within the render ya?

Exactly!! lol ok I sure am glad that made sense 😅

PR #2109, opened 2020-05-07
Beta step one validation

lindsayrmaher:
### What's this PR do?

This pull request adds a cypress test to look for the register to vote form on the Beta page!! We also added some HTML5 validation on the fields, and did some minor updates based on design feedback

👀  
should we add any more testing? does the pattern matching make sense?

## Disabled Button
Screen Shot 2020-05-07 at 11 00 34 AM

## Filled In Form
Screen Shot 2020-05-07 at 11 00 47 AM

## Zip Code Validation
Screen Shot 2020-05-07 at 11 01 11 AM

## Email Validation
Screen Shot 2020-05-07 at 11 01 26 AM

### Relevant tickets

References [Pivotal #](https://www.pivotaltracker.com/story/show/172439461).

### Checklist

### DIFF
--- a/cypress/integration/alpha-voter-registration-drive-page.js
+++ b/cypress/integration/alpha-voter-registration-drive-page.js
@@ -1,6 +1,7 @@
 /// <reference types="Cypress" />
 
 import faker from 'faker';
+
 import { userFactory } from '../fixtures/user';
 import exampleVoterRegistrationDriveCampaign from '../fixtures/contentful/exampleVoterRegistrationDriveCampaign';
 
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -126,7 +126,7 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     );
   });
 
-  it('Beta OVRD Step One Register to Vote Section Displays as Expected', () => {
+  it('Beta OVRD Step One register to vote section displays as expected', () => {
     const user = userFactory();
 
     cy.mockGraphqlOp('BetaVoterRegistrationDrivePageQuery', {
@@ -142,7 +142,7 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     );
   });
 
-  it('Beta OVRD Step One Register to Vote Section Button Is Disabled When Form is Empty', () => {
+  it('Beta OVRD Step One register to vote section button is disabled when form is empty', () => {
     const user = userFactory();
 
     cy.mockGraphqlOp('BetaVoterRegistrationDrivePageQuery', {
@@ -157,7 +157,7 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     );
   });
 
-  it('Beta OVRD Step One Register to Vote Section Button Is Enabled When Form is Filled In', () => {
+  it('Beta OVRD Step One register to vote section button is enabled when form is filled in', () => {
     const user = userFactory();
 
     cy.mockGraphqlOp('BetaVoterRegistrationDrivePageQuery', {
@@ -167,11 +167,8 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
 
     cy.visit(getBetaPagePathForUser(user));
 
-    const email = 'text@test.com';
-    const zip = '12345';
-
-    cy.get('[data-id=voter-registration-email-field]').type(email);
-    cy.get('[data-id=voter-registration-zip-field]').type(zip);
+    cy.get('[data-id=voter-registration-email-field]').type('text@test.com');
+    cy.get('[data-id=voter-registration-zip-field]').type('12345');
 
     cy.get('[data-test=voter-registration-submit-button]').should('be.enabled');
   });
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -136,7 +136,10 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
 
     cy.visit(getBetaPagePathForUser(user));
 
-    cy.get('[data-test=card-title]').contains(`Register online to vote`);
+    cy.get('[data-test=voter-registration-form-card]').should('have.length', 1);
+    cy.get('[data-test=voter-registration-form-card] > header > h1').contains(
+      'Register online to vote',
+    );
   });
 
   it('Beta OVRD Step One Register to Vote Section Button Is Disabled When Form is Empty', () => {
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -124,4 +124,19 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
       `250,000+ young people have registered to vote via DoSomething (it takes less than 2 minutes!). After you register, share with your friends to enter to win a $1,500 scholarship!`,
     );
   });
+
+  it('Beta OVRD Step One Register to Vote Section Displays as Expected', () => {
+    const user = userFactory();
+
+    cy.mockGraphqlOp('BetaVoterRegistrationDrivePageQuery', {
+      user,
+      campaignWebsite,
+    });
+
+    cy.visit(getBetaPagePathForUser(user));
+
+    cy.get('.beta-page-registration-form > header > h1').contains(
+      `Register online to vote`,
+    );
+  });
 });
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -152,7 +152,9 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
 
     cy.visit(getBetaPagePathForUser(user));
 
-    cy.get('[data-test=element-button-component]').should('be.disabled');
+    cy.get('[data-test=voter-registration-submit-button]').should(
+      'be.disabled',
+    );
   });
 
   it('Beta OVRD Step One Register to Vote Section Button Is Enabled When Form is Filled In', () => {
@@ -171,6 +173,6 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     cy.get('[data-id=voter-registration-email-field]').type(email);
     cy.get('[data-id=voter-registration-zip-field]').type(zip);
 
-    cy.get('[data-test=element-button-component]').should('be.enabled');
+    cy.get('[data-test=voter-registration-submit-button]').should('be.enabled');
   });
 });
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -1,6 +1,7 @@
 /// <reference types="Cypress" />
 
 import faker from 'faker';
+
 import { userFactory } from '../fixtures/user';
 
 const betaPagePath = '/us/my-voter-registration-drive';
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -137,4 +137,36 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
 
     cy.get('[data-test=card-title]').contains(`Register online to vote`);
   });
+
+  it('Beta OVRD Step One Register to Vote Section Button Is Disabled When Form is Empty', () => {
+    const user = userFactory();
+
+    cy.mockGraphqlOp('BetaVoterRegistrationDrivePageQuery', {
+      user,
+      campaignWebsite,
+    });
+
+    cy.visit(getBetaPagePathForUser(user));
+
+    cy.get('[data-test=element-button-component]').should('be.disabled');
+  });
+
+  it('Beta OVRD Step One Register to Vote Section Button Is Enabled When Form is Filled In', () => {
+    const user = userFactory();
+
+    cy.mockGraphqlOp('BetaVoterRegistrationDrivePageQuery', {
+      user,
+      campaignWebsite,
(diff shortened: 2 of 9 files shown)


resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -18,59 +18,61 @@ const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
...
+      },
+    });
+    window.location = `https://register.rockthevote.com/registrants/new?partner=37187&source=user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true&email_address=${email}&home_zip_code=${zip}`;
   };
 
   return (
     <>
       <Card
-        className="md:w-3/5 bordered rounded"
+        className="md:w-3/5 bordered rounded beta-page-registration-form"
```
aaronschachter, line 37: We should be using a `data-test` attribute instead of a class, when the class doesn't have any styles associated with it and is only used for testing. Recently added this documentation into https://github.com/DoSomething/communal-docs/tree/master/Contributing#data-attributes.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -18,59 +18,61 @@ const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
...
+              Email
               <input
                 className="text-field"
                 required
-                type="text"
+                type="email"
                 name="email"
                 value={email}
                 onChange={handleChange}
+                placeholder="member@dosomething.org"
```
aaronschachter, line 51: I don't see these placeholders in the designs. Are these values approved?

cypress/integration/alpha-voter-registration-drive-page.js
```diff
@@ -1,6 +1,7 @@
 /// <reference types="Cypress" />
 
 import faker from 'faker';
+
```
aaronschachter, line 4: Good catch! 👀

resources/assets/components/utilities/Card/Card.js
```diff
@@ -4,7 +4,10 @@ import classnames from 'classnames';
 
 const renderHeader = (title, link, onClose) => (
   <header className="text-gray-900 bg-yellow-500 p-3 rounded-t clip-padding-box">
-    <h1 className="inline font-source-sans text-base uppercase">
+    <h1
+      data-test="card-title"
```
aaronschachter, line 8: I'm curious to know what the rest of the team thinks about setting `data-test` within utility components like our `Card` and `ElementButton`, vs in a wrapper around the elements we're testing. My first reaction was to prefer the `data-test` attribute to be added somewhere within our `StartVoterRegistrationForm` component that we're testing... but I do think it could be a useful addition for standardizing the names we use within tests.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -18,18 +18,17 @@ const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
...
-        label: 'voter_registration',
-        context: {
-          campaignId,
-        },
-      });
-      window.location = `https://register.rockthevote.com/registrants/new?partner=37187&source=user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true&email_address=${email}&home_zip_code=${zip}`;
-    }
+  const handleSubmit = event => {
+    event.preventDefault();
+    trackAnalyticsEvent('click_voter_registration_action', {
```
aaronschachter, line 23: Nit - I think new lines before and after this `trackAnalyticsEvent` function call would improve readability a smidge.

resources/assets/components/utilities/Card/Card.js
weerd: A potential concern is based on how the item is being targeted in the test. If there are multiple Card components on a test page, and you aim to grab `[data-test="card-title"]` I could see complications with this since there's likely more than one on a page if it has multiple cards.

So then potentially the ideal would be targeting the Card specifically and then chaining the `h1`: `[data-test="card-type-name-hook"] h1` or whatever. 

One update I did do to the new buttons fairly recently is the ability to easily pass any extra attributes (like `data-*` attributes) that would then get output. It could be something we add to the `Card` component to make it easier to give them hooks for data attribute testing purposes.

Here's an [example](https://github.com/DoSomething/phoenix-next/blob/15583dccdbee7c42621832bb04e933066cf8a1c0/resources/assets/components/pages/HomePage/NewHomePage.js#L309-L314) from the new home page component and a `PrimaryButton`.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
weerd: Agreed! I ❤️ carriage returns and room to breeeeathe.

resources/assets/components/utilities/Card/Card.js
mendelB: Ooh, yeah I do kinda like that idea of yielding the `data-*` attributes. Feels like the best of both worlds to me! I've found myself wishing that React components supported native attributes as props built-in...

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
mendelB: Should this event name be past-tense?
```suggestion
    trackAnalyticsEvent('clicked_voter_registration_action', {
```

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -18,18 +18,19 @@ const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
...
+  const handleSubmit = event => {
+    event.preventDefault();
+
+    trackAnalyticsEvent('click_voter_registration_action', {
+      action: 'button_clicked',
+      category: EVENT_CATEGORIES.campaignAction,
+      label: 'voter_registration',
+      context: {
+        campaignId,
+      },
```
mendelB, line 30: Would it be helpful to include the URL we're redirecting to within the `context`, like we do in the `VoterRegistrationAction`?
https://github.com/DoSomething/phoenix-next/blob/15583dccdbee7c42621832bb04e933066cf8a1c0/resources/assets/components/actions/VoterRegistrationAction/VoterRegistrationAction.js#L48

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
aaronschachter: @mendelB 👀 🦅 !! You are right:

https://github.com/DoSomething/phoenix-next/blob/15583dccdbee7c42621832bb04e933066cf8a1c0/resources/assets/components/actions/VoterRegistrationAction/VoterRegistrationAction.js#L40

We ended up creating a separate component vs adding this into the `VoterRegistrationAction` like we had talked about in a past megagroom - because it was too much overhead to try to break the existing component `VoterRegistrationAction` into templates for our one-time use.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
mendelB: Oop, there's PII in it!

mendelB: (though maybe worth including the `contextSource` or even the unparsed URL (with placeholder data or something) so we have an easy way of knowing this is from _this_ version of the voter reg redirect per https://github.com/DoSomething/phoenix-next/pull/2109#discussion_r421754699 (but also hmm, the page URL would tell us that implicitly...))

resources/assets/components/utilities/Card/Card.js
```diff
@@ -37,6 +38,7 @@ const Card = ({
 }) => (
   <article
     className={classnames('card bg-white w-full clip-padding-box', className)}
+    {...attributes}
```
weerd, line 41: woot woot!

resources/assets/components/blocks/CampaignUpdate/__snapshots__/CampaignUpdate.test.js.snap
```diff
@@ -32,6 +33,7 @@ exports[`it generates a campaign update in affiliate mode snapshot 1`] = `
 
 exports[`it generates a campaign update snapshot 1`] = `
 <Card
+  attributes={Object {}}
```
weerd, line 36: Are the tests getting mad if you don't have this prop here? I would've expected the default props stuff to kick in? Not a huge deal if not, just kinda curious 🤔

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
aaronschachter: Not sure how to advise here. Question for data team? Don't think this should block merging as we still have a bit to go before we start directing betas to this page.

lindsayrmaher: > Oop, there's PII in it!

What does this mean? But also yeah I'm not sure what the best choice is. Perhaps I can post in the help-data channel and ask if there are any thoughts?

cypress/integration/beta-voter-registration-drive-page.js
```diff
@@ -124,4 +125,54 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
...
+      campaignWebsite,
+    });
+
+    cy.visit(getBetaPagePathForUser(user));
+
+    const email = 'text@test.com';
+    const zip = '12345';
+
+    cy.get('[data-id=voter-registration-email-field]').type(email);
+    cy.get('[data-id=voter-registration-zip-field]').type(zip);
```
aaronschachter, line 174: Nit that I missed at first pass - it'd be nice to just pass the email and zip values directly here since the constants are only reference once:
```
cy.get('[data-id=voter-registration-email-field]').type( 'text@test.com');
```

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
mendelB: Ah sorry! Yes totally not a blocker! Totally me being confusing.

PII - personally identifiable information (zip and email) which we'd rather not track with analytics. But the stripped URL might be nice? Totes a data team Q!

cypress/integration/beta-voter-registration-drive-page.js
```diff
@@ -124,4 +125,54 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
       `250,000+ young people have registered to vote via DoSomething (it takes less than 2 minutes!). After you register, share with your friends to enter to win a $1,500 scholarship!`,
     );
   });
+
+  it('Beta OVRD Step One Register to Vote Section Displays as Expected', () => {
```
aaronschachter, line 129: This also a small nit - but the capitalization of these tests Don't Follow The Rest of How The Other Tests are Capitalized (usually they're cased like regular sentences)

resources/assets/components/blocks/CampaignUpdate/__snapshots__/CampaignUpdate.test.js.snap
lindsayrmaher: Yeah when I ran the update failing snapshots command these were included. I guess I had assumed this was with the default props

weerd: Gotcha, ok!

weerd:
Reviewed the `attributes` prop work and that looks good to me 👯

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -18,59 +18,65 @@ const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
...
                 value={zip}
                 onChange={handleChange}
                 required
+                pattern="^\s*?\d{5}(?:[-\s]\d{4})?\s*?$"
+                data-id="voter-registration-zip-field"
               />
             </label>
           </div>
           <PrimaryButton
+            attributes={{ 'data-test': 'voter-registration-submit-button' }}
```
aaronschachter, line 75: Adding a comment here to reference in #2116

cypress/integration/beta-voter-registration-drive-page.js
```diff
@@ -124,4 +125,51 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
...
+      campaignWebsite,
+    });
+
+    cy.visit(getBetaPagePathForUser(user));
+
+    cy.get('[data-test=voter-registration-form-card]').should('have.length', 1);
+    cy.get('[data-test=voter-registration-form-card] > header > h1').contains(
+      'Register online to vote',
+    );
+  });
```
aaronschachter, line 143: Where did this test go? It's not on [`master`. ](https://github.com/DoSomething/phoenix-next/blob/master/cypress/integration/beta-voter-registration-drive-page.js)

PR #2112, opened 2020-05-07
Moar button updates and docs

weerd:
### What's this PR do?

This pull request fixes old component buttons that were styled as `-tertiary` buttons. It also updates the `PetitionSubmissionAction` and `VoterRegistrationAction` to use the new button components as well as further fleshing out the "Buttons" documentation.

👀 

This PR also removes the `CtaPopoverButton` component which was not actually used anywhere to called by any other component. I removed it for now since it would have otherwise also needed a button update and not worth it if not being used. We can always restore it and update it if we need to.

It also removes the `LoadMore` component. The functionality was replaced by the new buttons and is no longer needed.

### Relevant tickets

### Checklist


resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js
```diff
@@ -155,52 +155,54 @@ class TextSubmissionAction extends PostForm {
...
-                  this.state.textValue.length > CHARACTER_LIMIT
-                }
-                isLoading={submissionItem && submissionItem.isPending}
-                text={this.props.buttonText}
-                type="submit"
-              />
-
-              <PrivacyLanguage className="mb-1 mt-5" />
-            </form>
+            <div className="p-3">
```
weerd, line 158: Simply updated the markup to match other actions to make it easier to deal with; so moving the `p-3` from the `<form>` to a `<div>` inside the `Card`.

docs/development/features/buttons.md
```diff
@@ -2,24 +2,130 @@
...
+- [ElementButton](./elementbutton) \(renders a basic button element `<button>`\)
 
 Some of the specialized buttons that extend from the above core button components are:
 
-- `PrimaryButton`
-- `SecondaryButton`
-- `ToggleButton`
+- [PrimaryButton](./primarybutton)
+- [SecondaryButton](./secondarybutton)
+- [ToggleButton](./togglebutton)
```
weerd, line 16: We'll see if these links work... not sure how GitBooks will generate the headers for the items below... either lowercasing the title case or switching to hyphens. I guessed lowercase, we'll see if that's right!

weerd:
I need to update some tests 🚥, but this is otherwise ready for 👀

docs/development/features/buttons.md
```diff
@@ -2,24 +2,130 @@
...
+>
+```
+
+Example Rendered output:
+
+![LinkButton Example]('../../.gitbook/assets/link-button-example.png)
+
+#### Important Notes
+
+Anchor links in the browser do not support the `disabled` attribute. If added to an `<a>` it does absolutely nothing and the link can still be clicked, thus the `LinkButton` does not support the `isDisabled` property.
```
lindsayrmaher, line 70: a v helpful note, as I can attest from personal experience 😅

PR #2114, opened 2020-05-07
Adds voting reasons query selector to OVRD Alpha Page (Part 1/2)

aaronschachter:
### What's this PR do?

This pull request adds a new `beta_voter_reg_drive_page` feature flag. When it's enabled, a checkbox will appear on the `SocialDriveAction` component on the OVRD Alpha Page, allowing an alpha to customize the query string of their beta page URL. The beta page won't change at all yet - that will be handled in a different PR.

It also holds off on adding styles (and hiding the `Embed`) per the [design](https://app.abstract.com/projects/fadf0790-f1f9-11e6-9a30-b54ba7e8e705/branches/a02a32be-978a-44ec-a73f-006d65a9744f/collections/72c40c1f-06db-4544-8724-69c62d9224e3), saving it for a Part 2 PR to help keep this reviewable.

Screen Shot 2020-05-07 at 4 38 58 PM

I'll add this review app to Aurora for testing out.

Also holding off on adding tests and debouncing the toggle  `POST /links` request for another PR.

### Relevant tickets

References [Pivotal #172419329](https://www.pivotaltracker.com/story/show/172419329).

### Checklist


aaronschachter:
Confirmed the cause selector works as expected on https://dosomething-phoenix-de-pr-2114.herokuapp.com/us/campaigns/online-registration-drive/action.

And regular `SocialDriveAction` works as expected on https://dosomething-phoenix-de-pr-2114.herokuapp.com/us/campaigns/test-school-finder-campaign/action.

I tried to add Bertly credentials to the review app to verify that the link shortening requests are working as expected, but the Phoenix `POST /v2/links` proxy restricts requests to a few DS domains:
https://github.com/DoSomething/phoenix-next/blob/6eafc7c741a18af94179bc0eaa4cfd6793453565/app/Http/Controllers/Api/LinkController.php#L35-L42

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -92,8 +138,17 @@ class SocialDriveAction extends React.Component {
...
+                  onSelect: queryStr =>
+                    this.setState({
+                      longUrl: this.getLongUrl(queryStr),
+                    }),
+                })
+              : null}
+
             <div className="p-3">
-              <Embed url={link} />
+              <Embed url={this.state.longUrl} />
```
aaronschachter, line 151: There's a decent chance this could account for the `referrer_user_id` values of `{userId}` we sometimes see over in [#172465010](https://www.pivotaltracker.com/story/show/172465010). The `url` prop being passed to the `Embed` here (before these changes) doesn't get `dynamicString` called -- which is what replaces all the `{userId}` token instances with the logged in user's ID.

.env.example
```diff
@@ -82,3 +82,4 @@ DS_ENABLE_SITEWIDE_CTA=
 DS_ENABLE_SITEWIDE_NPS_SURVEY=
 DS_ENABLE_SNAPCHAT=
 DS_ENABLE_VOLUNTEER_CREDITS=
+DS_ENABLE_VOTER_REG_BETA_PAGE=
```
weerd, line 85: 🎉

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -30,19 +30,41 @@ class SocialDriveAction extends React.Component {
...
+   * @return String
+   */
+  getLongUrl(queryStr) {
+    let result = dynamicString(this.props.link, { userId: this.props.userId });
+
+    if (queryStr) {
+      // Append the queryStr to our current result.
+      result = `${result}${
+        this.props.link.includes('?') ? '&' : '?'
+      }${queryStr}`;
```
weerd, line 64: You could possibly use the [`appendToQuery()`](https://github.com/DoSomething/phoenix-next/blob/293a43ad4d578428c36f479045e4f9af76e9b3bd/resources/assets/helpers/index.js#L38-L52) helper function here. It uses built in methods to deal with mashing a URL together with numerous query params and not duplicating the use of `?`.

weerd: Might be worth a shot, only caveat is it expects an href string, not just a query string 🤔 If not, at least defs gank the approach to handling mashing the query string together. It handles a lot of edge cases with weird things in strings.

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -62,17 +84,43 @@ class SocialDriveAction extends React.Component {
     });
   };
 
+  /**
+   * Executes API request to shorten given longUrl, and saves result to state.shortUrl.
+   *
+   * @param String longUrl
+   */
+  shortenLink(longUrl) {
```
weerd, line 92: Thoughts on maybe following similar method naming scheme as `getLongUrl()`, so `getShortUrl()`?

weerd: Or mabye `setShortUrl()` since it's setting it into state? 🤷‍♂️

aaronschachter: I like!

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
aaronschachter: I'll take a look but might save this one for the Part 2, because for now we're only using this in the OVRD pages -- where we shouldn't have to handle any weird things (editors aren't adding in their own URLs)

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
aaronschachter: The thing I liked about the original `shortenLink` function name is that it helps to indicate there's an API request involved... `setShortUrl` feels like it doesn't indicate that. the `getShortUrl` I could _get_ behind :wocka:

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
weerd: Gotcha. Ya, it's just never a good practice in general for us to manually construct URLs like this. It's something @DFurnes has pointed out to me in the past and best to typically let libs handle for us :) Fine for now, but recommend updating for Part 2!

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
weerd: We need to request that GitHub include support for :wocka:

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/AlphaPage.js
```diff
@@ -22,13 +22,7 @@ const AlphaPage = ({ userId }) => {
...
         className="grid-wide clearfix wrapper pb-3"
       />
-      <div className="grid-wide">
-        <SocialDriveActionContainer
-          link={`https://vote.dosomething.org/member-drive?userId=${userId}&r=user:${userId},source:web,source_details:onlinedrivereferral,referral=true`}
-          actionId={config.shareLink.actionId}
-          hidePageViews
-        />
-      </div>
+      <ShareLink referrerUserId={userId} actionId={config.shareLink.actionId} />
```
weerd, line 25: I'm a little confused why you broke this out into a `ShareLink` component instead of revising a bit the props on the `SocialDriveActionContainer`? I kept seeing `ShareLink` and thinking this is _just_ the component for the link part of the Social Drive Action.

I could be missing something but just feels like an extra, unnecessary layer. Couldn't this `AlphaPage` just have the feature flag and the ternaries for the props to the `SocialDriveActionContainer`?

aaronschachter: Yeah, it could be. I moved it into its own dir to live along side the `VotingReasonsQueryOptions` component... but don't feel too strongly about keeping it there. We'd just move the `VotingReasonsQueryOptions` into the `Alpha` dir too. While we're actively developing these components, it's been helpful to have smaller components to avoid constant merge conflicts with other devs - but we're basically done with this template.

aaronschachter: I don't mind moving it out, I can make that change.

weerd: Sounds good! I think it helps a bit with clarity for someone not as deep into the weeds with OVRD stuff! 😅

aaronschachter: Agreed! and upated

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -92,8 +140,17 @@ class SocialDriveAction extends React.Component {
...
             ) : null}
 
+            {queryOptions
+              ? React.cloneElement(queryOptions, {
+                  onChange: queryStr =>
+                    this.setState({
+                      longUrl: this.getLongUrl(queryStr),
+                    }),
+                })
+              : null}
```
weerd, line 150: I'm not very familiar with `React.cloneElement`; don't think we use it elsewhere... but I'm wondering if just having the feature flag here and outputting the `VotingReasonsQueryOptions` based on the flag would also work and be more straightforward and also setting the `onChange` as a prop there?

```js
{betaPageEnabled ? <VotingReasonsQueryOptions onChange={...anon func to set state...} /> : null}
```
Unless there's a specific reason the above wouldn't work? 🤔

aaronschachter: This was future thinking, for being able to pass any component into the `queryOptions` prop to customize the query string of the link given to `SocialDriveAction` - and keep the `VotingReasonsQueryOptions` defined within the `pages/VoterRegistrationDrivePage/Alpha` dir. In discussing this feature, it had come up as a wishlist item before (have no idea for which campaign).

aaronschachter: I would prefer to keep this as-is, and only check the beta page feature flag from one place (the alpha page component, where it is now).

aaronschachter: I'm also remembering that the beta page feature flag will go away, once this is all ready to go live. So... we cannot check for it within the `SocialDriveAction`

weerd: You could pass a boolean prop alternatively from the Alpha Page while passing other props.

Ok, anyhoo, I don't know enough about the future plans for this just noting it's not a pattern we've used before, but defs doesn't mean we can't moving forward! I'll leave it up to the OVRD team then :)

PR #2115, opened 2020-05-08
Adds markdown-it-collapsible

aaronschachter:
### What's this PR do?

This pull request adds the [`markdown-it-collapsible`](https://www.npmjs.com/package/markdown-it-collapsible) plugin to render `<summary>` and `<details>` HTML5 tags using syntax:
```
+++ Click me!
Hidden text
+++
```

Here's some example content:

```
+++ Does my vote _actually_ matter?

Yes, it definitely does! There are countless elections that have been decided by only a few votes. For example, a state election in 2017 was tied, so the winner was selected by drawing names out of a hat. Literally. In a 2014 election in Louisiana, one race was decided by just 1%. Who pushed the winner over the top? Young people, who supported that candidate more than any other group. 

+++

+++I'm not 18, but I will be by Election Day. Can I still register to vote?

It depends on your state. In most states, YES. In all but a few states, you can register to vote if you’ll be 18 by Election Day. In the following states, you can only register if your 18th birthday is within a certain number of days or months before Election Day.

+++

+++ When are my elections?

There are hundreds of local elections in each state. Luckily, our friends at Rock the Vote can send you reminders about when elections are coming up in your state or city.

+++
```

And its rendered result, with the third item expanded:

Screen Shot 2020-05-08 at 9 06 31 AM

With a grain of 🧂 and ideally some guidance on the best way to style these new `<summary>` and `<details>` tags and the additional background context I want to provide.

Where should the styles go? Also how should we handle the custom carat image our [designs](https://share.goabstract.com/99a2866c-437b-4d78-9799-d74cc91ba4e9) require? 

Closed:
Screen Shot 2020-05-08 at 9 11 12 AM

Open:
Screen Shot 2020-05-08 at 9 09 07 AM

### Relevant tickets

References [Pivotal #172087924](https://www.pivotaltracker.com/story/show/172087924).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/TextContent/markdown.scss
+++ b/resources/assets/components/utilities/TextContent/markdown.scss
@@ -154,5 +154,5 @@ details > *:not(summary) {
 
 summary {
   @apply font-bold;
-  font-size: 18px;
+  font-size: inherit;
 }
--- a/resources/assets/components/utilities/TextContent/markdown.scss
+++ b/resources/assets/components/utilities/TextContent/markdown.scss
@@ -143,3 +143,16 @@
     overflow-wrap: break-word;
   }
 }
+
+details {
+  @apply pb-3;
+}
+
+details > *:not(summary) {
+  @apply pt-5;
+}
+
+summary {
+  @apply font-bold;
+  font-size: 18px;
+}
--- a/resources/assets/components/utilities/TextContent/markdown.scss
+++ b/resources/assets/components/utilities/TextContent/markdown.scss
@@ -133,6 +133,19 @@
     margin: 60px auto;
     width: 33.333333333%;
   }
+
+  details {
+    @apply pb-3;
+  }
+
+  details > *:not(summary) {
+    @apply pt-5;
+  }
+
+  summary {
+    @apply font-bold;
+    font-size: inherit;
+  }
 }
 
 .footnotes {
@@ -142,17 +155,4 @@
     color: theme('colors.gray.600');
     overflow-wrap: break-word;
   }
-}
-
-details {
-  @apply pb-3;
-}
-
-details > *:not(summary) {
-  @apply pt-5;
-}
-
-summary {
-  @apply font-bold;
-  font-size: inherit;
 }
--- a/resources/assets/scss/base.scss
+++ b/resources/assets/scss/base.scss
@@ -106,6 +106,18 @@ a > * {
   border-radius: 0;
 }
 
+details {
+  @apply pb-3;
+}
+
+summary {
+  @apply font-bold;
+}
+
+details > *:not(summary) {
+  padding-top: 1.25em;
+}
+
 /*
 |-------------------------------------------------------------------------------
 | START: DEPRECATION WARNING!!
--- a/resources/assets/scss/base.scss
+++ b/resources/assets/scss/base.scss
@@ -106,18 +106,6 @@ a > * {
   border-radius: 0;
 }
 
-details {
-  @apply pb-3;
-}
-
-summary {
-  @apply font-bold;
-}
-
-details > *:not(summary) {
-  padding-top: 1.25em;
-}
-
 /*
 |-------------------------------------------------------------------------------
 | START: DEPRECATION WARNING!!
--- a/package-lock.json
+++ b/package-lock.json
@@ -11783,6 +11783,11 @@
       "resolved": "https://registry.npmjs.org/markdown-it-anchor/-/markdown-it-anchor-5.2.5.tgz",
       "integrity": "sha512-xLIjLQmtym3QpoY9llBgApknl7pxAcN3WDRc2d3rwpl+/YvDZHPmKscGs+L6E05xf2KrCXPBvosWt7MZukwSpQ=="
     },
+    "markdown-it-collapsible": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/markdown-it-collapsible/-/markdown-it-collapsible-1.0.0.tgz",
+      "integrity": "sha512-WRkgDH66GNxkdkZGEyW/0VnsEtOzxZuQCAY1e5FjVYACAhMtDNleZs1Y7NZSj5Pn3yksQ6SrgPkEwQW2IFGAQw=="
+    },
     "markdown-it-footnote": {
       "version": "3.0.2",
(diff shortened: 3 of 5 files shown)


aaronschachter:
I was testing with the dev FAQ block on our beta OVRD page: https://dosomething-phoenix-de-pr-2115.herokuapp.com/us/my-voter-registration-drive?referrer_user_id=55767606a59dbf3c7a8b4571

katiecrane:
@aaronschachter hmm as for the carats are the current ones automatically rendered by the package? is there a way to override?

aaronschachter:
The current carats are native to the [`details`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) HTML element that this plugin renders when we use its syntax.

It's possible to [override them via CSS ](https://github.com/bioruebe/markdown-it-collapsible#example-css)-- I just don't know where in the codebase we should be styling these (beyond where I've guessed, in the base.scss). Looking at our `/resources/assets/scss` dir, seems like that'd be a good spot for the carat png -- I've just never(knowingly?) added a global style / element to the repo before. 

> Sorry I'm not sure the answer to where the styles should live 

Paging doctors @weerd and @DFurnes 🚑

resources/assets/scss/base.scss
```diff
@@ -106,6 +106,18 @@ a > * {
...
+}
+
+summary {
+  @apply font-bold;
+}
+
+details > *:not(summary) {
+  padding-top: 1.25em;
+}
+
```
weerd, line 120: You should take a look at how we style other markdown related content. Typically they fall within a container div with a class of `markdown` and we use that to help avoid collisions, so wrapping these with `.markdown` would be ideal. I don't recall offhand, but I would recommend moving these styles out of `base.scss` which we want to aim to reduce the rules it contains and move to where we have other markdown related styling rules.

resources/assets/scss/base.scss
```diff
@@ -106,6 +106,18 @@ a > * {
...
+details {
+  @apply pb-3;
+}
+
+summary {
+  @apply font-bold;
+}
+
+details > *:not(summary) {
+  padding-top: 1.25em;
```
weerd, line 118: There is a specific size for 1.25 rem in [Tailwind](https://tailwindcss.com/docs/customizing-spacing/#default-spacing-scale). You can use `pt-5`.

resources/assets/scss/base.scss
aaronschachter: Thanks @weerd -- looks like https://github.com/DoSomething/phoenix-next/blob/35659a69a406d6f15233f0fac97ef549a492460d/resources/assets/components/utilities/TextContent/markdown.scss#L3 should do the trick?

weerd: Ah, yep. That's the spot 👍

resources/assets/scss/base.scss
weerd: Oh wait, @aaronschachter in this case you'd use:

```css
padding-top: theme('spacing.5');
```

Since this is in the stylesheet and not in the html!

DFurnes:
Just was this float by in #notify-applications... so good!!! Absolutely love this addition. 👏

resources/assets/components/utilities/TextContent/markdown.scss
```diff
@@ -143,3 +143,16 @@
...
+}
+
+details > *:not(summary) {
+  @apply pt-5;
+}
+
+summary {
+  @apply font-bold;
+  font-size: 18px;
+}
```
weerd, line 158: I'm not sure why the `footnotes` are sitting outside the `.markdown` class, but I recommend placing these _inside_ the `.markdown` class. After the `hr` rule, before the closing `}` for the `.markdown` rule :)

resources/assets/scss/base.scss
aaronschachter: Yeah I was having trouble getting `font-size-base` to work and your comment makes me totes go ‼️

resources/assets/components/utilities/TextContent/markdown.scss
```diff
@@ -143,3 +143,16 @@
...
+  @apply pb-3;
+}
+
+details > *:not(summary) {
+  @apply pt-5;
+}
+
+summary {
+  @apply font-bold;
+  font-size: 18px;
```
weerd: you an use the Tailwind `text-base` instead of the `font-size` declaration.

```
@apply font-bold text-base;
```

aaronschachter: Would we want to inherit the `summary` anyway though?

resources/assets/scss/base.scss
aaronschachter: But the `pt-5` is working here.  It should change per your comment just becuase that's preferred for stylesheet vs the html inline classes?

resources/assets/components/utilities/TextContent/markdown.scss
weerd: I saw you changed it to `inherit`, so that makes sense. Comment applied moreso to defining the specific size in px instead of using Tailwind :)

resources/assets/scss/base.scss
weerd: So the change you made using my suggested `pt-5` works fine!

The main difference is... you can use Tailwind classes when using the `@apply` helper method.

BUT if you define your own rule, like `padding-top` _instead of utilizing `@apply` THEN you need to use the `theme()` helper.

With your update you popped the `pt-5` into the `@apply` so that works gang-busters.

If you did `padding-top: pt-5` it would fail cause CSS has no idea what `pt-5` means, but the `@apply` knows how to interpret that. Hopefully that helps clarify!

weerd: In general, using `@apply` is the preferred way to go in a stylesheet since it'll help us all get used to the Tailwind class names, but in some scenarios you can't as easily use it, and need to resort to the `theme()` helper.

resources/assets/components/utilities/TextContent/markdown.scss
aaronschachter: Great! Is this good to go? I'm blocked from merge until you ✅ per requesting changes.

weerd:
👍 on the CSS related updates 💃 🕺

PR #2118, opened 2020-05-08
Hero Section Cypress Tests

lindsayrmaher:
### What's this PR do?

This pull request adds more cypress tests for the beta page hero section and adds a `contextSource` to the analytics for the register to vote button. 

👀 

Retroactively adding some testing for this section now that some baseline test files were added.

### Relevant tickets

References [Pivotal # 172681512](https://www.pivotaltracker.com/story/show/172681512).

### Checklist

### DIFF
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -16,8 +16,8 @@ const campaignWebsite = {
   },
   scholarshipAmount: '1500',
   scholarshipDeadline: '2022-04-25T00:00-08:00',
-  url: mockUrl,
   additionalContent: null,
+  url: mockUrl,
 };
 
 /**
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -82,7 +82,10 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
 
     cy.visit(getBetaPagePathForUser(user));
 
-    cy.get('[data-test=beta-ovrd-page-cover-image]').should('have.length', 1);
+    cy.get('[data-test=beta-voter-registration-drive-page-cover-image]').should(
+      'have.length',
+      1,
+    );
   });
 
   it('Beta OVRD HeroSection displays referrer first name if referrer user found', () => {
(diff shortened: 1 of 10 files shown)


cypress/integration/beta-voter-registration-drive-page.js
```diff
@@ -89,6 +104,53 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
...
+    });
+
+    cy.visit(getBetaPagePathForUser(user));
+
+    cy.get('[data-test=beta-ovrd-campaign-info-block]').should(
+      'have.length',
+      1,
+    );
+    cy.contains('button', 'View Scholarship Details').click();
+    cy.get('[data-test=beta-page-scholarship-details]').should(
```
aaronschachter, line 143: These `data-test` names are pretty inconsistent in this PR, and also breaks from the `beta-voter-registration-drive-page-*` naming I started off with in #2096. I know it's long, and I don't really mind `beta-ovrd-page-*` -- but we should really change it in all the components if that's what we go with. I chose `beta-voter-registration-drive-*`page-*` because it's essentially unknown what `ovrd` means in the repo - it's more of a business acronym than anything we use in code (thus far, at least).

cypress/integration/beta-voter-registration-drive-page.js
```diff
@@ -89,6 +104,53 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
...
+      'have.length',
+      1,
+    );
+    cy.get('.modal-portal > .wrapper.modal-container').click('topRight');
+    cy.get('[data-test=beta-page-scholarship-details]').should(
+      'have.length',
+      0,
+    );
+  });
```
aaronschachter, line 152: This test is definitely useful, but really feels like it belongs in a separate test file as this scholarship modal is a shared component. What do you think about moving it a new file called `campaign-info-block.js`, following our [`campaign-info-bar.js`](https://github.com/DoSomething/phoenix-next/blob/master/cypress/integration/campaign-info-bar.js) test by example?  Speaking of examples, a great example of how these names can sometimes make your 👀 turn into 🍝

aaronschachter: It doesn't have to be a full test of all things in the `CampaignInfoBlock`, but the modal opening/closing in particular is one that feels like it should go in there (it'd be fine even if the file only had just the one test)

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -117,7 +118,7 @@ const ScholarshipInfoBlock = ({
   }
 
   return (
-    <Card className="flex flex-col-reverse md:flex-row">
+    <Card attributes={attributes} className="flex flex-col-reverse md:flex-row">
```
aaronschachter, line 121: Digging adding the `attributes` prop into all of these usual suspects components!

cypress/integration/beta-voter-registration-drive-page.js
```diff
@@ -89,6 +104,53 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     );
   });
 
+  it('Beta OVRD HeroSection displays scholarhsip info if it is provided', () => {
```
aaronschachter, line 107: Typo 🥼

cypress/integration/beta-voter-registration-drive-page.js
lindsayrmaher: Keeping it consistent makes sense! Updated on 9beb2f1

cypress/integration/beta-voter-registration-drive-page.js
lindsayrmaher: Sounds good re: moving it to a new file. I had rummaged through a couple of other test files to see if it already existed or not. we're doing some `campaign-info-block` specific tests in other places like the `hero-campaign-signup` so it might be nice to keep them all in one place

PR #2121, opened 2020-05-11
Suppressing sitewide banner from Beta OVRD page

lindsayrmaher:
### What's this PR do?

This pull request adds a check to the siteWide banner to make sure we are suppressing this from the Beta OVRD pages!

👀 

N/A

### Relevant tickets

References [Pivotal # 172440373](https://www.pivotaltracker.com/story/show/172440373).

### Checklist


resources/assets/components/App.js
```diff
@@ -36,7 +36,8 @@ const App = ({ store, history }) => {
   return (
     <ReduxProvider store={store}>
       <ErrorBoundary FallbackComponent={ErrorPage}>
-        {featureFlag('sitewide_cta_banner') ? (
+        {featureFlag('sitewide_cta_banner') &&
+        window.location.pathname !== '/us/my-voter-registration-drive' ? (
```
aaronschachter, line 40: Let's create a new Cypress test for the sitewide banner where we can verify it won't appear on the beta OVRD URL, but will appear on any other random page (like maybe visiting our `exampleCampaign` fixture). #2108 is an example PR that used the `withFeatureFlags` cypress command to mock enabling a feature (maybe not in but can we remove the feature now that its live?)

resources/assets/components/App.js
```diff
@@ -36,8 +36,7 @@ const App = ({ store, history }) => {
   return (
     <ReduxProvider store={store}>
       <ErrorBoundary FallbackComponent={ErrorPage}>
-        {featureFlag('sitewide_cta_banner') &&
-        window.location.pathname !== '/us/my-voter-registration-drive' ? (
+        {featureFlag('sitewide_cta_banner') ? (
```
aaronschachter: Is there a reason to keep this anymore? I can't imagine we're ever going to turn this off now -- Seems simple enough to remove while we're in here?

cypress/integration/sitewide-banner.js
```diff
@@ -0,0 +1,75 @@
...
+      1,
+    );
+  });
+
+  it('The Site Wide Banner is not displayed on the beta voter registration (OVRD) drive page', () => {
+    const user = userFactory();
+
+    cy.mockGraphqlOp('BetaVoterRegistrationDrivePageQuery', {
+      user,
+      campaignWebsite,
```
aaronschachter, line 65: Let's move our `campaignWebsite` const here, as we're only using it in one place.

cypress/integration/sitewide-banner.js
```diff
@@ -0,0 +1,75 @@
+/// <reference types="Cypress" />
+
+import faker from 'faker';
+
+import { userFactory } from '../fixtures/user';
+import exampleCampaign from '../fixtures/contentful/exampleCampaign';
+
+const betaPagePath = '/us/my-voter-registration-drive';
+const mockUrl = faker.image.imageUrl();
```
aaronschachter, line 9: We don't need this `mockUrl` since we're not testing for it at all.

resources/assets/components/utilities/SitewideBanner/SitewideBannerContent.js
```diff
@@ -26,7 +27,10 @@ const SitewideBannerContent = ({
     });
   };
   return (
-    <div className="w-full md:flex md:justify-center bg-yellow-500 p-8 pb-4 md:pt-4 sm:px-10 z-50">
+    <div
+      className="w-full md:flex md:justify-center bg-yellow-500 p-8 pb-4 md:pt-4 sm:px-10 z-50"
+      {...attributes}
```
aaronschachter, line 32: This isn't a biggie, but I think we'd be fine to just set a `data-test` attribute directly here to reduce some complexity. The only component that will ever be rendering a `SitewideBanner` is our `App`.

cypress/integration/sitewide-banner.js
```diff
@@ -0,0 +1,75 @@
...
+
+  it('The Site Wide Banner is not displayed on the beta voter registration (OVRD) drive page', () => {
+    const user = userFactory();
+
+    cy.mockGraphqlOp('BetaVoterRegistrationDrivePageQuery', {
+      user,
+      campaignWebsite,
+    });
+
+    cy.visit(getBetaPagePathForUser(user));
```
aaronschachter, line 68: Similarly, this can just be ` cy.visit(`'/us/my-voter-registration-drive?referrer_user_id=${user.id}`))` and we can remove all the copy/paste variable/functions that we only need for this single test.

resources/assets/components/App.js
lindsayrmaher: Katie had pointed out that it's useful to use to toggle the banner on and off in our slack thread. https://dosomething.slack.com/archives/CUQMTP0RM/p1589221570003500?thread_ts=1589209254.001900&cid=CUQMTP0RM

Which I kind of liked the idea of, rather than deleting and re-adding the sitewide banner to `App.js` when we want to use it.

katiecrane: Sorry I might have been confusing in slack - I meant if we have no plans to turn this off or see any need to toggle on and off easily or anything like that, then we can remove the flag. We only need the flag if we need a quick way to turn on and off, and I'm not sure what the plans around that are

lindsayrmaher: ah ok ok, i guess we are definitely keeping this feature long term so it would make sense to clean this up now. i can remove it!

cypress/integration/sitewide-banner.js
```diff
@@ -0,0 +1,61 @@
...
+        superTitle: faker.lorem.words(),
+        title: faker.company.bsBuzz(),
+        subTitle: faker.lorem.words(),
+        content: faker.lorem.sentence(),
+      },
+    });
+  });
+
+  it('The Site Wide Banner is displayed on campaign pages', () => {
+    cy.withState(exampleCampaign).visit('/us/campaigns/test-example-campaign');
```
aaronschachter, line 24: There's a fun helper for this!
```suggestion
    cy.anonVisitCampaign(exampleCampaign);
```

lindsayrmaher: yeeeeees! i shall use!

PR #2125, opened 2020-05-11
Starts on styles and tests for OVRD alpha page voting reasons

aaronschachter:
### What's this PR do?

This pull request begins on styling the `VotingReasonsQueryOptions` component introduced in #2114, and starts on some test coverage as well. 

Style changes:

* Display two cells of voting causes on medium+ screens
* Display a preview link instead of an `Embed` if beta feature is enabled
* Hide the "Share your link" element if beta feature is enabled

Also takes care of a few cleanup tasks from old PR's 

* Fixes missing card title when beta feature disabled (see [Slack thread](https://dosomething.slack.com/archives/CTVPG6L4R/p1589222654499800)) - refs #2114 
* Adds a `null` default for the `VotingReasonsQueryOptions.onChange` prop - refs #2114
* Adds relevant tests removed in #2108 as a standalone `social-drive-action` test

* [Small](https://user-images.githubusercontent.com/1236811/81605494-62295c80-9386-11ea-8b7e-cbea3dccd24c.png)
* [Medium](https://user-images.githubusercontent.com/1236811/81605552-7bcaa400-9386-11ea-9d78-377847665a5c.png)
* [Large](https://user-images.githubusercontent.com/1236811/81605592-8be28380-9386-11ea-8102-e17a8212c79b.png)

The styles aren't quite done per [designs](https://app.abstract.com/share/8cbe354f-edc9-4b04-9308-2dd810b31f62) yet -- the "Copy Link" button will need to be blurple and the `SocialShareTray` will need to be left aligned. Saving that for another PR to help keep this reviewiable.

### Relevant tickets

References [Pivotal #172419329](https://www.pivotaltracker.com/story/show/172419329).

### Checklist


resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -206,6 +221,10 @@ SocialDriveAction.propTypes = {
   fullWidth: PropTypes.bool,
   link: PropTypes.string.isRequired,
   pageId: PropTypes.string,
+  /**
+   * This prop expects a React component to allow user customization of link prop's query string.
+   * @see /resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/AlphaPage
+   */
   queryOptions: PropTypes.object,
```
mendelB, line 228: I _think_ a `PropTypes.element` would denote a React component here.
```suggestion
  queryOptions: PropTypes.element,
```

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/AlphaPage.js
```diff
@@ -39,7 +39,9 @@ const AlphaPage = ({ userId }) => {
               ? 'Urge your friend to vote based on the causes you care about most.'
               : null
           }
-          shareCardTitle={betaPageEnabled ? 'Share with your friends' : null}
+          shareCardTitle={
+            betaPageEnabled ? 'Share with your friends' : 'Your online drive'
+          }
```
mendelB, line 44: Unless you think this is _worse_, maybe worth using `undefined` to force the default prop, so we can avoid redefining that value here?
```suggestion
          shareCardTitle={ betaPageEnabled ? 'Share with your friends' : undefined }
```

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/VotingReasonsQueryOptions.js
```diff
@@ -15,37 +15,54 @@ const VotingReasonsQueryOptions = ({ onChange }) => {
...
+            .map(checkbox => checkbox)}
+        </div>
+      </div>
     </div>
   );
 };
 
 VotingReasonsQueryOptions.propTypes = {
-  onChange: PropTypes.func.isRequired,
+  onChange: PropTypes.func,
```
mendelB, line 61: Is there a case where we wouldn't expect this prop? (also if it's ever `null`, I think our invocation of it in the `useEffect` would throw an error? We could default it to an empty function I suppose.)

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/VotingReasonsQueryOptions.js
```diff
@@ -15,37 +15,54 @@ const VotingReasonsQueryOptions = ({ onChange }) => {
...
+        <div className="flex-1">
+          {checkboxes.slice(0, numCheckboxes / 2 + 1).map(checkbox => checkbox)}
         </div>
-      ))}
+        <div className="flex-1">
+          {checkboxes
+            .slice(numCheckboxes / 2 + 1, numCheckboxes)
+            .map(checkbox => checkbox)}
+        </div>
+      </div>
```
mendelB, line 55: Hmmm I wonder if we could simplify this by assigning the `flex` (or even `flex-basis`) to assign two items per row on medium screens and 1 per row by default? (I don't have the concrete implementation for that, but I feel like that _should_ be very possible!)

cypress/integration/social-drive-action.js
```diff
@@ -0,0 +1,49 @@
...
+    cy.mockGraphqlOp('ContentfulBlockQuery', contentfulBlockQueryResult);
+
+    cy.route('POST', linksApiUrl, { url: shortenedLink });
+
+    cy.authVisitBlockPermalink(user, blockId);
+
+    cy.get('.link-bar input').should('contain.value', shortenedLink);
+  });
+
+  it('Displays unshortened link and N/A page views if links API request fails', () => {
```
mendelB, line 40: Should we mock a failed API response from the `linksApiUrl` explicitly to avoid hitting the real world API in the test?

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/AlphaPage.js
aaronschachter: It looks like the test still fails. This feature won't be on for longer than this week, so I think it'll be safe to change it back to providing the duplicate default value for now.

mendelB: I think it fails because the capitalization is inconsistent between the test and the default prop, but agreed that it's no biggie for the short term in any case!

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/VotingReasonsQueryOptions.js
aaronschachter: Yeah, we don't set it in the `AlphaPage` itself when passing it into the `SocialDriveAction`. Agreed that defaulting to empty function seems cleaner, I made that change.

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/VotingReasonsQueryOptions.js
aaronschachter: I'm a little stuck on this one. Wouldn't this require adding custom styles (vs using tailwind)?

mendelB: I think it would indeed, which we could do using emotion. I do feel like this case is a good contender for custom styling though, no? Since using classes purely necessitates some involved finessing here?

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/VotingReasonsQueryOptions.js
mendelB: I wonder if we can just pass in the component type itself without rendering it from the Alpha page, or do we specifically need to render it / clone it?
```
// alpha page
<SocialDrive queryOptions={... ? VotingReasonsQueryOptions : ...}

// social drive 
queryOptions ? <queryOptions onChange=... 
```

aaronschachter: We specifically need to clone in order to send the function prop to the child. We discussed this piece in https://github.com/DoSomething/phoenix-next/pull/2114/files#r422382939 a bit -- I'd prefer to keep this reusable (as it is) should we ever want to customize a `SocialDriveAction` query again.

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/VotingReasonsQueryOptions.js
aaronschachter: Sounds right, added a new scss file.

mendelB: Apologies, I should have clarified -- I think we'd want to use an Emotion `css` prop here (re this discussion for the Landing page here https://github.com/DoSomething/phoenix-next/pull/2094#discussion_r419725849)

mendelB: This is some niche knowledge we've incorporated into Phoenix (that this would be a blocker for server-side rendering, plus our general usage of Emotion `css`, `styled`, and `tw` macros...) that I think is a prime contender for some documentation re our dev roundtable discussion last week!

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/VotingReasonsQueryOptions.js
mendelB: Ah gotcha, thanks for that context! I don't think classically rendering this from within this component would make this any more specific though, no? If we're expecting a component _type_, and just rendering it out from here and passing it an `onChange` prop, it wouldn't matter what the component actually was and would work with our `VotingReasonsQueryOptions`, but just as well with any other component we'd make in the future... I _think_ 🤔

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/VotingReasonsQueryOptions.js
aaronschachter: Is it possible to use the `@include media($medium)` within that `css` prop? This PR still isn't even ready to go live yet, there are still additional styles - so I'd love to save this for part 3 if we can live with the scss file.

mendelB: Any classic scss works! [Media queries included](https://emotion.sh/docs/media-queries). We just wouldn't be able to use the Forge `$medium` variable but it could be something like `@media (min-width: ${tailwind('screens.md')})`

(but what's better is that we can use Tailwind classes themselves with the `tw` macro [like here](https://github.com/DoSomething/phoenix-next/blob/3c5572cc32f20ef9aa44c78b108a51eb853e4330/resources/assets/components/pages/LandingPage/LandingPage.js#L18)).

aaronschachter: Got it working, thanks for the guidance!

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/VotingReasonsQueryOptions.js
```diff
@@ -15,37 +16,64 @@ const VotingReasonsQueryOptions = ({ onChange }) => {
...
+        width: 100%;
+      }
+
+      @include media($medium) {
+        div {
+          flex: 0 50%;
+        }
+      }
+    }
+  `;
```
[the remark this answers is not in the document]
aaronschachter: I tried that initially, couldn't get working.

Screen Shot 2020-05-12 at 7 20 41 AM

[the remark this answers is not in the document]
mendelB: Ah gotcha sorry, hmm not sure why that wouldn't work but 🤷‍♀️ (I'm also surprised that `$medium` worked, but I imagine something to do with how it's built and the Forge rules being in scope 🤔)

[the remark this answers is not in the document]
weerd: I'm confused why you can't use Tailwind classes in the html for what you're doing with Emotion css 🤔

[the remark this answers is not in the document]
weerd: Ya, and I agree with @mendelB... I'm surprised an SCSS variable of `$medium` and associated SCSS `media()` helper function is working here... 🤔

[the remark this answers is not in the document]
aaronschachter: Only I was able to get Tailwind classes to work was by splitting the checkboxes into 2 div's, which is how this PR was initially opened: https://github.com/DoSomething/phoenix-next/pull/2125/commits/69d44e56629b2d8c042def6b123bd07a434e4030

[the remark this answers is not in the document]
weerd: Ah apologies, I didn't know the full history. I saw this pop up in the Slack notifications and the current approach threw me because it doesn't seem ideal. Is there a review app link where this component can be visualized? And just to be sure, these styles are for getting the 2 column effect from medium size and up right?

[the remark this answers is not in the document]
aaronschachter: Oops, the SCSS variables weren't working, sorry about that. I've reverted the `css` prop commit to get this branch back to when I got it with the scss.

Yes - only thing that's going on with these styles is to have 2 columns of checkboxes on medium+.

Taking a look at the page IRL would be really helpful, thanks. Added the review app to Aurora: https://dosomething-phoenix-de-pr-2125.herokuapp.com/us/campaigns/online-registration-drive/action (it's still building per latest update)

resources/assets/components/pages/VoterRegistrationDrivePage/Alpha/VotingReasonsQueryOptions.js
```diff
@@ -4,6 +4,8 @@ import React, { useEffect, useState } from 'react';
 
 import { votingReasons } from './config';
 
+import './voting-reasons-query-options.scss';
```
weerd, line 7: I think we want to avoid separate stylesheets unless absolutely necessary and I don't think this scenario requires it. Plus, adding styles by targeting an ID selector is not ideal. Is there a link to review app to visualize. I can help better guide if I can see the working example :) I think you could do something like what's in the cause filters (albeit it needs some cleanup) where you can easily use CSS Grid to accomplish this instead of flexbox which is trickier in this scenario.

weerd: oh! I think GitHub ate 😋 your comment @aaronschachter when you pushed up the changes to this section we were commenting on. I just saw it in Slack and checking the review app now 👯

weerd: ok! @aaronschachter remove the added stylesheet, you can avoid flex for this which just makes it a bit more complicated.

on the `<div id="voting-reasons-query-options">` containing the options just define a 2 column grid that takes effect on medium size and up!

```html
<div id="voting-reasons-query-options" class="md:grid md:grid-cols-2">
...
</div>
```

image

aaronschachter: These `grid` tailwind classes will now forever be burned into my brain. Thanks for the halp, @weerd !

mendelB: 🤩so beautifully simple!

resources/assets/components/actions/SocialDriveAction/SocialDriveAction.js
```diff
@@ -150,28 +152,41 @@ class SocialDriveAction extends React.Component {
...
-              <Embed url={this.state.longUrl} />
+              {queryOptions ? (
+                <a
+                  className="font-normal underline text-blurple-500"
+                  href={longUrl}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  preview your custom page
+                </a>
```
mendelB, line 163: Non blocking - but I think we could use [the `LinkButton`](https://github.com/DoSomething/phoenix-next/blob/1a78e8772e2a153d8c3d5ed94cf01866e9007706/resources/assets/components/utilities/Button/LinkButton.js) for this?

aaronschachter: I can take a look in the part 3 PR! Still a little more styling (and testing) to do for this feature.

PR #2126, opened 2020-05-11
Display personalized message using selected causes

Dinnall:
### What's this PR do?

This pull request...
Adds a personalized message to the beta's page with the causes the alpha selects 
"Voting is important for young people because we can affect change on issues we care about most like [cause 1], [cause 2], and [cause 3]."
' - [Alpha's First Name]

*
Screen Shot 2020-05-12 at 6 26 45 PM
Screen Shot 2020-05-12 at 6 27 07 PM
Screen Shot 2020-05-12 at 6 27 21 PM
Screen Shot 2020-05-12 at 6 27 30 PM

...

This PR isn't finished, just want to receive feedback before proceeding with writing cypress test 
...

### Relevant tickets
References [Pivotal #](https://www.pivotaltracker.com/n/projects/2417735/stories/172087475).

### Checklist

- [x ] This PR has been added to the relevant Pivotal card.

### DIFF
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -186,7 +186,7 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     cy.get(
       '[data-test=beta-voter-registration-drive-page-quote-text]',
     ).contains(
-      'Voting is important for young people because we can effect change on issues we care about most like , COVID-19 relief and climate change.',
+      'Voting is important for young people because we can effect change on issues we care about most like COVID-19 relief and climate change.',
     );
     cy.get(
       '[data-test=beta-voter-registration-drive-page-quote-byline]',
@@ -210,7 +210,7 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     cy.get(
       '[data-test=beta-voter-registration-drive-page-quote-text]',
     ).contains(
-      'Voting is important for young people because we can effect change on issues we care about most like , COVID-19 relief, student debt, mental health, gun violence and climate change.',
+      'Voting is important for young people because we can effect change on issues we care about most like COVID-19 relief,student debt,mental health,gun violence, and climate change.',
     );
     cy.get(
       '[data-test=beta-voter-registration-drive-page-quote-byline]',
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -204,13 +204,13 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     cy.visit(
       `${getBetaPagePathForUser(
         user,
-      )}&voting-reasons=covid-relief,student-debt,mental-health,gun-violence,climate-change`,
+      )}&voting-reasons=covid-relief,climate-change,healthcare`,
     );
 
     cy.get(
       '[data-test=beta-voter-registration-drive-page-quote-text]',
     ).contains(
-      'Voting is important for young people because we can effect change on issues we care about most like COVID-19 relief, student debt, mental health, gun violence, and climate change.',
+      'Voting is important for young people because we can effect change on issues we care about most like COVID-19 relief, climate change, and healthcare.',
     );
     cy.get(
       '[data-test=beta-voter-registration-drive-page-quote-byline]',
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -204,13 +204,13 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     cy.visit(
       `${getBetaPagePathForUser(
         user,
-      )}&voting-reasons=covid-relief,climate-change,healthcare`,
+      )}&voting-reasons=mental-health,climate-change,healthcare,covid-relief`,
     );
 
     cy.get(
       '[data-test=beta-voter-registration-drive-page-quote-text]',
     ).contains(
-      'Voting is important for young people because we can effect change on issues we care about most like COVID-19 relief, climate change, and healthcare.',
+      'Voting is important for young people because we can effect change on issues we care about most like mental health, climate change, healthcare, and COVID-19 relief',
     );
     cy.get(
       '[data-test=beta-voter-registration-drive-page-quote-byline]',
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -210,7 +210,7 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     cy.get(
       '[data-test=beta-voter-registration-drive-page-quote-text]',
     ).contains(
-      'Voting is important for young people because we can effect change on issues we care about most like COVID-19 relief,student debt,mental health,gun violence, and climate change.',
+      'Voting is important for young people because we can effect change on issues we care about most like COVID-19 relief, student debt, mental health, gun violence, and climate change.',
     );
     cy.get(
       '[data-test=beta-voter-registration-drive-page-quote-byline]',
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -149,6 +149,74 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     ).contains(`- ${user.firstName}`);
   });
 
+  it('Beta OVRD quote displays one cause query when found in voting-options query', () => {
+    const user = userFactory();
+
+    cy.mockGraphqlOp('BetaVoterRegistrationDrivePageQuery', {
+      user,
+      campaignWebsite,
+    });
+
+    cy.visit(`${getBetaPagePathForUser(user)}&voting-reasons=student-debt`);
+
+    cy.get(
+      '[data-test=beta-voter-registration-drive-page-quote-text]',
+    ).contains(
+      'Voting is important for young people because we can effect change on issues we care about most like student debt.',
+    );
+    cy.get(
+      '[data-test=beta-voter-registration-drive-page-quote-byline]',
+    ).contains(`- ${user.firstName}`);
+  });
+
+  it('Beta OVRD quote displays two causes when found in voting-options query', () => {
+    const user = userFactory();
+
+    cy.mockGraphqlOp('BetaVoterRegistrationDrivePageQuery', {
+      user,
+      campaignWebsite,
+    });
+
+    cy.visit(
+      `${getBetaPagePathForUser(
+        user,
+      )}&voting-reasons=covid-relief,climate-change`,
+    );
+
+    cy.get(
+      '[data-test=beta-voter-registration-drive-page-quote-text]',
+    ).contains(
+      'Voting is important for young people because we can effect change on issues we care about most like , COVID-19 relief and climate change.',
+    );
+    cy.get(
+      '[data-test=beta-voter-registration-drive-page-quote-byline]',
+    ).contains(`- ${user.firstName}`);
+  });
+
+  it('Beta OVRD quote displays three or more causes when found in voting-options query', () => {
+    const user = userFactory();
+
+    cy.mockGraphqlOp('BetaVoterRegistrationDrivePageQuery', {
+      user,
+      campaignWebsite,
+    });
+
+    cy.visit(
+      `${getBetaPagePathForUser(
+        user,
+      )}&voting-reasons=covid-relief,student-debt,mental-health,gun-violence,climate-change`,
+    );
+
+    cy.get(
+      '[data-test=beta-voter-registration-drive-page-quote-text]',
+    ).contains(
+      'Voting is important for young people because we can effect change on issues we care about most like , COVID-19 relief, student debt, mental health, gun violence and climate change.',
+    );
+    cy.get(
+      '[data-test=beta-voter-registration-drive-page-quote-byline]',
+    ).contains(`- ${user.firstName}`);
+  });
+
   it('Beta OVRD HeroSection displays scholarship info in blurb', () => {
     const user = userFactory();
 
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
@@ -24,22 +24,18 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
 
   const formatQuote = () => {
     const votingReasonsQuery = query('voting-reasons');
+
     if (!votingReasonsQuery) {
       return null;
     }
+
     const votingReasonsValues = votingReasonsQuery.split(',');
 
     if (votingReasonsValues.length === 1) {
       return ` like ${votingReasons[votingReasonsValues[0]]}`;
     }
 
-    if (votingReasonsValues.length === 2) {
-      return ` like ${votingReasons[votingReasonsValues[0]]} and ${
-        votingReasons[votingReasonsValues[1]]
-      }`;
-    }
-
-    if (votingReasonsValues.length > 2) {
+    if (votingReasonsValues.length > 1) {
       let result = '';
 
       votingReasonsValues.forEach((value, index) => {
@@ -49,6 +45,7 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
           result = `${result}, ${votingReasons[value]}`;
         }
       });
+
       return ` like ${result}`;
     }
 
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
@@ -1,8 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-
+import QueryString from 'query-string';
 import CoverImage from '../../../utilities/CoverImage/CoverImage';
 import CampaignInfoBlock from '../../../blocks/CampaignInfoBlock/CampaignInfoBlock';
+import votingReasons from '../../../pages/VoterRegistrationDrivePage/Beta/config';
 
 const HeroSection = ({ user, campaignInfo, modalToggle }) => {
   const { firstName } = user;
@@ -18,6 +19,13 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
    * TODO: Check for a voting-reasons query parameter, and render values in quote if present.
    * @see https://www.pivotaltracker.com/story/show/172087475
    */
+
+  QueryString.parse(location.search);
+  /*
+Query url
+check for votingReasons
+append to p tag
+*/
   return (
     <div className="hero-landing-page">
       <CoverImage
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
@@ -35,17 +35,23 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
       return ` like ${votingReasons[votingReasonsValues[0]]}`;
     }
 
-    if (votingReasonsValues.length > 1) {
+    if (votingReasonsValues.length === 2) {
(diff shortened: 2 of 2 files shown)


resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -18,6 +19,25 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
    * TODO: Check for a voting-reasons query parameter, and render values in quote if present.
    * @see https://www.pivotaltracker.com/story/show/172087475
    */
+
+  const formatEndofBetaCauseSentence = () => {
+    let userCauses = query('voting-reasons').split(',');
```
lindsayrmaher, line 24: i think we want to do `', '` on the split rather than just a comma.

lindsayrmaher:
This looks good to me aside from the string rendered once we pull out the causes from the query params. It also looks like there's a few small syntax/build errors from circleCI!

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
Dinnall: Resolved the circleCI issues and added a space when I split the array

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -18,6 +19,25 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
    * TODO: Check for a voting-reasons query parameter, and render values in quote if present.
    * @see https://www.pivotaltracker.com/story/show/172087475
    */
+
+  const formatEndofBetaCauseSentence = () => {
+    const userCauses = query('voting-reasons').split(', ');
```
mendelB, line 24: Will the query value be divided by a comma _and_ space here?

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -18,6 +19,25 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+    if (userCauses && userCauses.length) {
+      userCauses
+        .reduce((accumulator, currentCause, idx) => {
+          return (
+            accumulator +
+            currentCause +
+            (idx === userCauses.length - 2 ? 'and' : ', ')
+          );
+        }, '')
+        .slice(0, -2);
```
mendelB, line 35: I wonder if we could replace the `reduce` (I personally find them really hard to grok), with a conditional something roughly like the following:
```js
if (userCauses.length >= 2) {
  userCauses[userCauses.length-1] = `and ${lodash.last(userCauses)}`;
}

return userCauses.join(', ');
```

Perhaps subjective, but to me it feels a bit more explicit about our special case, and we get away with a simple `join` for the regular comma formatting?

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
mendelB: Also, we'll probably need to use a lodsh `get` here to ensure we don't call `split` on an `undefined` value
```suggestion
    const userCauses = get(query('voting-reasons'), '').split(', ');
```

Then we could even remove the first condition directly below since we're guaranteed an (at least empty) array!

mendelB: Ohhh and also, I wonder if we should whitelist the accepted causes and filter out anything else before we process the array to avoid any hardcoded naughty values?

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
aaronschachter: We do not want to split on `', '` . This line is looking for the `voting-reasons` query parameter passed from the alpha page, which will look like `/my-voter-registration-drive?referrer_user_id={userId}&voting-reasons=voting-reasons?gun-control,covid-relief,healthcare`. It should split by only the comma.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
aaronschachter: If a voting-reason / cause doesn't have a corresponding key in our config, it should not be displayed - so we don't have to worry about users customizing their `voting-reasons` values. Refs https://github.com/DoSomething/phoenix-next/pull/2126#pullrequestreview-409558496

Dinnall: Thanks for the feedback. I pushed the revised changes.

Currently, the code works for all test cases, except when there are no elements in `voter-reasons` param. I tried @mendelB suggestion for using `get` before splitting the array but it didn't work. It returns `Cannot read property 'split' of undefined` which makes sense. I also tried to check with a simple `if(userCauses && userCauses.length)` before splitting the array but it breaks the previously passing cases 
cc: @aaronschachter

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import { query } from '../../../../helpers';
+import { votingReasons } from './config';
```
aaronschachter, line 5: Nit - let's order these by [ascending line length](https://github.com/DoSomething/communal-docs/tree/master/Contributing#imports).

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -18,6 +20,34 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
    * TODO: Check for a voting-reasons query parameter, and render values in quote if present.
    * @see https://www.pivotaltracker.com/story/show/172087475
    */
+
+  const formatEndOfBetaCauseSentence = () => {
+    let result = [];
```
aaronschachter, line 25: First check to see if `query('voting-reasons')` is set. If it's not, we can return `null` here -- that will account for the scenario when the query doesn't eixst.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -18,6 +20,34 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+    let lastCause;
```
aaronschachter, line 26: We don't need to declare `result` or `lastCause` until we get to our check for length > 2.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -18,6 +20,34 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+      return ` like ${votingReasons[userCauses[0]]}`;
+    }
+
+    if (userCauses.length === 2) {
+      return ` like ${votingReasons[userCauses[0]]} and ${
+        votingReasons[userCauses[1]]
+      }`;
+    }
+
+    if (userCauses.length >= 2) {
```
aaronschachter, line 39: At this point we don't need a length check, we should have handled when the length is 0 (per my suggestion to first check whether its set), 1, and 2

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -18,6 +20,34 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+      userCauses.forEach(cause => {
+        result.push(votingReasons[cause]);
+      });
+      lastCause = result[result.length - 1];
+      result = result.slice(0, -1);
+      result.join(', ');
+      return ` like ${result} and ${lastCause}`;
```
aaronschachter, line 46: This could be simplified by concatenating our end result vs building an array during this `forEach` loop:

```
let result = '';
votingReasonsValues.forEach((value, index) => {
   if (index === votingReasonsValues.length - 1) {
      result = `${result} and ${value}
   } else {
      result = `${result}, `
   }
});
```

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -18,6 +20,34 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
    * TODO: Check for a voting-reasons query parameter, and render values in quote if present.
    * @see https://www.pivotaltracker.com/story/show/172087475
    */
+
+  const formatEndOfBetaCauseSentence = () => {
+    let result = [];
+    let lastCause;
+    const userCauses = query('voting-reasons').split(',');
```
aaronschachter, line 27: Can we call this something like `votingReasonValues`? I know we're referring to this as "causes" a lot in business speak, but I've made it a point to not use the term in these components, because a "cause" in our codebase generally refers to completely separate values we store on a user's profile (and correspond to the cause pages, e.g. (dosomething.org/us/causes/voter-registration)

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -18,6 +20,34 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
    * TODO: Check for a voting-reasons query parameter, and render values in quote if present.
    * @see https://www.pivotaltracker.com/story/show/172087475
    */
+
+  const formatEndOfBetaCauseSentence = () => {
```
aaronschachter, line 24: Let's add function comments to indicate this function returns a string. Can we rename to just `formatQuote` ? This name feels confusing because technically it's our alpha's quote to share the reasons why they feel it's important to vote, and want the beta to register to vote as well.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -18,6 +20,34 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+    if (userCauses.length >= 2) {
+      userCauses.forEach(cause => {
+        result.push(votingReasons[cause]);
+      });
+      lastCause = result[result.length - 1];
+      result = result.slice(0, -1);
+      result.join(', ');
+      return ` like ${result} and ${lastCause}`;
+    }
+    return '';
```
aaronschachter, line 48: We generally always add a blank line after our `return` statements. I can add this to the contributing wiki.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
lindsayrmaher: 🤦‍♀️ - I did not think through that response.. (probs shouldn't review things at the EOD). Meant to say we should add a space when we join the values back together to be rendered

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
Dinnall: Are you saying to check `query('voting-reasons')` outside the function?

1. I tried @mendelB  suggestions using `get`. Ran into a few issues, I mentioned above ^
2. is there another you want me to check if `query('voting-reasons')` the empty. I also tried using `isEmpty`

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
Dinnall: I moved there here because they were causing linting issues.
I can move them back.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -15,9 +17,44 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+   * Query url for voting-reasons param.
+   *
+   * @return {String}
    */
+
+  const formatQuote = () => {
+    const votingReasonsQuery = query('voting-reasons');
+    if (!votingReasonsQuery) {
+      return null;
+    }
```
aaronschachter, line 29: Let's add new lines before and after this to help readability. I've opened up a pull request in the Contributing wiki to help formalize this preference: https://github.com/DoSomething/communal-docs/pull/36

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -15,9 +17,44 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+      let result = '';
+
+      votingReasonsValues.forEach((value, index) => {
+        if (index === votingReasonsValues.length - 1) {
+          result = `${result} and ${votingReasons[value]}`;
+        } else {
+          result = `${result}, ${votingReasons[value]}`;
+        }
+      });
+      return ` like ${result}`;
```
aaronschachter, line 52: Let's also add a new line before returning this.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -15,9 +17,44 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+      return ` like ${votingReasons[votingReasonsValues[0]]}`;
+    }
+
+    if (votingReasonsValues.length === 2) {
+      return ` like ${votingReasons[votingReasonsValues[0]]} and ${
+        votingReasons[votingReasonsValues[1]]
+      }`;
+    }
+
+    if (votingReasonsValues.length > 2) {
```
aaronschachter, line 42: We can remove this check entirely, we've already accounted for all instances less than 2.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -15,9 +17,44 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+        if (index === votingReasonsValues.length - 1) {
+          result = `${result} and ${votingReasons[value]}`;
+        } else {
+          result = `${result}, ${votingReasons[value]}`;
+        }
+      });
+      return ` like ${result}`;
+    }
+
+    return ' ';
```
aaronschachter, line 55: Don't need this either if we remove the check for length > 2.

aaronschachter:
To add new tests with the `voting-options` query, you'll want to append it to the URL we visit:

```
 cy.visit(`${getBetaPagePathForUser(user)}&voting-options=student-debt`);
```

Dinnall:
> To add new tests with the `voting-options` query, you'll want to append it to the URL we visit:
> 
> ```
>  cy.visit(`${getBetaPagePathForUser(user)}&voting-options=student-debt`);
> ```
switched `voting-options` to `voting-reasons`

cypress/integration/beta-voter-registration-drive-page.js
```diff
@@ -149,6 +149,74 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
...
+    cy.visit(
+      `${getBetaPagePathForUser(
+        user,
+      )}&voting-reasons=covid-relief,climate-change`,
+    );
+
+    cy.get(
+      '[data-test=beta-voter-registration-drive-page-quote-text]',
+    ).contains(
+      'Voting is important for young people because we can effect change on issues we care about most like , COVID-19 relief and climate change.',
```
aaronschachter, line 189: This quote shouldn't contain a comma, it should read as:

>  like COVID-19 relief and climate change.

cypress/integration/beta-voter-registration-drive-page.js
```diff
@@ -149,6 +149,74 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
...
+    cy.visit(
+      `${getBetaPagePathForUser(
+        user,
+      )}&voting-reasons=covid-relief,student-debt,mental-health,gun-violence,climate-change`,
+    );
+
+    cy.get(
+      '[data-test=beta-voter-registration-drive-page-quote-text]',
+    ).contains(
+      'Voting is important for young people because we can effect change on issues we care about most like , COVID-19 relief, student debt, mental health, gun violence and climate change.',
```
aaronschachter, line 213: This is also a bug. Instead of:

> like , COVID-19 relief, student debt, mental health, gun violence and climate change.

It should read as:
> like COVID-19 relief, student debt, mental health, gun violence, and climate change.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -15,9 +17,41 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+    if (votingReasonsValues.length === 1) {
+      return ` like ${votingReasons[votingReasonsValues[0]]}`;
+    }
+
+    if (votingReasonsValues.length > 1) {
+      let result = '';
+
+      votingReasonsValues.forEach((value, index) => {
+        if (index === votingReasonsValues.length - 1) {
+          result = `${result} and ${votingReasons[value]}`;
```
aaronschachter, line 43: This is missing its trailing comma, which we could easily fix with:
```suggestion
          result = `${result}, and ${votingReasons[value]}`;
```

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -15,9 +17,41 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+        } else {
+          result = `${result}, ${votingReasons[value]}`;
```
aaronschachter, line 45: Looks like we need another conditional to avoid adding the space and comma here if our index is equal to 1 - so we can avoid that leading comma

cypress/integration/beta-voter-registration-drive-page.js
```diff
@@ -149,6 +149,74 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
...
+    cy.visit(
+      `${getBetaPagePathForUser(
+        user,
+      )}&voting-reasons=covid-relief,student-debt,mental-health,gun-violence,climate-change`,
+    );
+
+    cy.get(
+      '[data-test=beta-voter-registration-drive-page-quote-text]',
+    ).contains(
+      'Voting is important for young people because we can effect change on issues we care about most like COVID-19 relief,student debt,mental health,gun violence, and climate change.',
```
aaronschachter, line 213: This needs a space after the commas:

Instead of
```
COVID-19 relief,student debt,mental health,gun violence, and climate change.
```
it should read as: 
```
COVID-19 relief, student debt, mental health, gun violence, and climate change.
```

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/HeroSection.js
```diff
@@ -15,9 +17,47 @@ const HeroSection = ({ user, campaignInfo, modalToggle }) => {
...
+      let result = '';
+
+      votingReasonsValues.forEach((value, index) => {
+        if (index === votingReasonsValues.length - 1) {
+          result = `${result} and ${votingReasons[value]}`;
+        } else {
+          result = `${result}${votingReasons[value]},`;
+        }
+      });
+      console.log('result:', result);
```
aaronschachter, line 54: Let's remove this debugging.

Dinnall: I completely missed this 😩.
Removed it

cypress/integration/beta-voter-registration-drive-page.js
aaronschachter: I see this conversation marked as resolved in the PR, but this test is not what our expected output should be when there are more than 2 causes per last [comment](https://github.com/DoSomething/phoenix-next/pull/2126/files#r424093252).

Dinnall: I added the space after the comma, but my cypress test keeps failing. I even copied the exact message you have here. I Just pushed my last commit. Can you take a look when you get a chance?

PR #2127, opened 2020-05-11
Button docs and updates part trois

weerd:
### What's this PR do?

This pull request updates some remaining components to switch out the old button components for the new ones. It also officially removes the old `ButtonLink` component and further updates the "General Buttons" documentation.

👀 

### Relevant tickets

### Checklist


resources/assets/components/CampaignPageNavigation/CampaignPageNavigation.js
```diff
@@ -34,7 +34,7 @@ const CampaignPageNavigation = ({
   return campaignPages.length ? (
     <PageNavigation pages={campaignPages}>
       {isAffiliated ? null : (
-        <SignupButtonContainer className="-inline nav-button" />
+        <SignupButtonContainer className="inline-block -inline nav-button" />
```
weerd, line 37: Since I don't believe this shows up on campaign page nav anywhere since we don't really use unaffiliated community style pages anymore I just popped this class on so it behaves as what the Forge `-inline` class is setting. I suspect though that we can likely remove the `SignupButtonContainer` from even needing to output here.

resources/assets/components/actions/LinkAction/templates/CtaTemplate.js
```diff
@@ -50,9 +37,19 @@ const CtaTemplate = ({
...
+        onClick={() =>
+          trackAnalyticsEvent('clicked_link_action', {
+            action: 'button_clicked',
+            category: EVENT_CATEGORIES.campaignAction,
+            label: 'link_action',
+            context: { ...context, url: href },
+          })
+        }
+        text={buttonText}
+      />
```
weerd, line 52: I went ahead and updated this component even though we'll be deprecating it, so that it would allow us to delete the `<Button>` component in an upcoming PR!

resources/assets/components/blocks/CurrentSchoolBlock/CurrentSchoolForm.js
```diff
@@ -22,32 +22,35 @@ const CurrentSchoolForm = ({ description, userId }) => {
   const [updateUserSchool] = useMutation(USER_SCHOOL_MUTATION);
 
   return (
-    <div className="school-finder-form">
-      {description ? <p className="p-3">{description}</p> : null}
-      <div className="select-state p-3">
+    <div className="p-3" data-test="school-finder-form">
```
mendelB, line 25: I think the corresponding Cypress test needs to be updated to select for a `data-test` instead of the class

weerd: yeep!

PR #2128, opened 2020-05-11
Add trackingId to affirmation modal

katiecrane:
### What's this PR do?

This pull request adds a `trackingId` to the affirmation modal (aka `<PostSignupModal>`) which is only rendered in one place, `CampaignRoute.js`.

Is this where this should be and what it should be called?

We want to be able to better track when this modal opens to try to figure out what's going on with a bug where it doesn't show up sometimes.

### Relevant tickets

References [Pivotal #172157205](https://www.pivotaltracker.com/story/show/172157205).

### Checklist

### DIFF
--- a/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
+++ b/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
@@ -34,7 +34,10 @@ const CampaignRoute = props => {
   return (
     <>
       {shouldShowAffirmation ? (
-        <Modal onClose={clickedHideAffirmation}>
+        <Modal
+          onClose={clickedHideAffirmation}
+          trackingId="SIGNUP_AFFIRMATION_MODAL"
+        >
           <PostSignupModal
             affirmation={affirmation || undefined}
             onClose={clickedHideAffirmation}


katiecrane:
@weerd I don't know much about this - is the issue with this that we'd have to add it each separate place a modal is rendered?

mendelB:
I always thought of the `trackingId` as an optional convenience prop. We have the option of utilizing it to automatically invoke the 'open' and 'close' analytics events in the hardcoded format. But otherwise, we're still free to refrain from passing through that prop and manually adding an open event whenever we open the modal, and/or a 'close' event via the `onClose` callback.

weerd:
Yep, sorry. @mendelB is right. I think moreso the name of the property is confusing. I think maybe we rename to `analyitcsLabel` or something to more accurately represent what it is and why we have it. ID is always a bit cryptic and can apply to a number of different use cases. When I first read the PR title, my 🧠 immediately thought of a tracking ID to track the user journey type of thing. 🤷‍♂️ I also hadn't had my coffee yet, so maybe that was it.

mendelB:
> Yep, sorry. @mendelB is right. I think moreso the name of the property is confusing. I think maybe we rename to `analyitcsLabel` or something to more accurately represent what it is and why we have it. ID is always a bit cryptic and can apply to a number of different use cases. When I first read the PR title, my 🧠 immediately thought of a tracking ID to track the user journey type of thing. 🤷‍♂️ I also hadn't had my coffee yet, so maybe that was it.

Oooh yeah agreed! - especially with our new analytics scheme - this could read clearer as `analyticsLabel`.

weerd:
Especially since we seem to uppercase the name, which in the analytics helper function we always lowercase so that also threw me off. Probably don't need to uppercase it and make it represent how we apply labels and other such analytics info across the app.

mendelB:
> Additionally confusing as to what this "ID" is for at first glance is that we seem to uppercase the name, which in the analytics helper function we always lowercase so that also threw me off. Probably don't need to uppercase it and make it represent how we apply labels and other such analytics info across the app.

Ah good call as well! We might just need to then start uppercasing the value when we pass it in as `context.modalType` since that's how we've tracked that value historically?
https://github.com/DoSomething/phoenix-next/blob/3c5572cc32f20ef9aa44c78b108a51eb853e4330/resources/assets/components/utilities/Modal/Modal.js#L73

weerd:
Ya, I think maybe a ticket to address this in the future 🤔 My thoughts are we should likely lowercase all things in context to just standardize it. Would suck if we can't track some data on our platform cause we happen to capitalize one letter or something in the value 😬

PR #2123, opened 2020-05-11
Retire Legacy Landing Page

mendelB:
### What's this PR do?

This pull request formally deprecates the Legacy Landing Page (which employed the `MosaicTemplate` from the `LedeBanner`) and it's accompanied logic.

It also removes the deprecated Campaign `template` field.

This opens us up to some exciting 🗑️and restructuring in follow-up work to come! (Keeping this PR contained to just retiring the rendering functionality. But will follow up with trashing legacy components and restructuring and simplifying some of the Landing Page / Lede Banner logic.)

### How should this be reviewed?
Perhaps see if you happen to spy anything being trashed that shouldn't be, but this should all be straightforward!

### Any background context you want to provide?
We've ported over all lingering campaigns activating the `featureFlagUseLegacyTemplate` to the new Landing Page & `HeroTemplate` ([Pivotal #170879500](https://www.pivotaltracker.com/story/show/170879500)).

### Relevant tickets

References [Pivotal #171776980](https://www.pivotaltracker.com/story/show/171776980).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2129, opened 2020-05-12
Contentful Script to remove signupArrowContent field from campaigns

mendelB:
### What's this PR do?

This pull request adds a Contentful Management API script to remove all the deprecated `additionalContent.SignupArrowContent` fields from campaigns.

👀 
https://dosomething.gitbook.io/phoenix-documentation/development/contentful/content-management-api-scripts

We've deprecated the legacy template in #2123 which means this feature is 👋 

I did think twice before writing this script. But I felt like
1. A looooot of campaigns had this field (we used this on most database campaigns)
2. We've got some remnant `additionalContent` fields which both clutter and confuse. So this felt like the right move

Side tangent but it's lame that most of the busy work on these scripts tends to be taking care of updating/publishing/archiving which feel automatable? It felt particularly lame on this script since we're actually doing so little, but it still feels like a ton of work. Food for thought. (Maybe another ole helper method if we see this continuing to happen?)

### Relevant tickets

### Checklist

### DIFF
--- a/contentful/management-api-scripts/2020_05_11_remove_signup_arrow_content_field_from_campaigns.js
+++ b/contentful/management-api-scripts/2020_05_11_remove_signup_arrow_content_field_from_campaigns.js
@@ -1,0 +1,86 @@
+const { omit } = require('lodash');
+
+const { contentManagementClient } = require('./contentManagementClient');
+const {
+  constants,
+  createLogger,
+  getField,
+  processEntries,
+  sleep,
+  withFields,
+} = require('./helpers');
+
+const { LOCALE } = constants;
+
+const logger = createLogger('remove_signup_arrow_content_field_from_campaigns');
+
+const removeSignupArrowContentField = async (environment, campaignEntry) => {
+  const campaignInternalTitle = getField(campaignEntry, 'internalTitle');
+  const campaignAdditionalContent = getField(
+    campaignEntry,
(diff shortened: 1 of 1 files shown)


contentful/management-api-scripts/2020_05_11_remove_signup_arrow_content_field_from_campaigns.js
```diff
@@ -0,0 +1,86 @@
...
+  publishedUpdatedCampaign = await updatedCampaign.publish();
+
+  if (!publishedUpdatedCampaign) {
+    logger.info(`☓ Unable to publish updated ${campaignInternalTitle}.`);
+    return;
+  }
+
+  logger.info(`✔ Published updated ${campaignInternalTitle}.`);
+
+  await sleep(500);
```
Vmack, line 69: everything looks great just a question about this timer, why does it need to be added here?

mendelB: This is just to avoid bombarding the Contentful API with too many requests all at once (we pause to space it out a bit), so we don't get throttled.

Vmack: ok makes sense!

PR #2136, opened 2020-05-14
Starts on QuizResultPage feature

aaronschachter:
### What's this PR do?

This pull request adds a new site path, `/us/quiz-results/:id`, and displays a new `QuizResultPage` component. It also adds a new `quiz_result_page` feature flag, which when enabled -- redirects a user to the respective quiz result page after completing the quiz. 

The `QuizResultPage` is minimally styled and will be polished up in future PR's, so not looking for feedback there. I'm looking for feedback to whether the docs make this hopefully easy for any dev to get a feel for how this feature works and contribute to the quiz result pages (without needing to know the nuts and bolts of how quizzes work)

Here are some example pages from the review app:
* https://dosomething-phoenix-de-pr-2136.herokuapp.com/us/quiz-results/347iYsbykgQe6KqeGceMUk
* https://dosomething-phoenix-de-pr-2136.herokuapp.com/us/quiz-results/1lvJHhlJqQSgKgwIwUymQ8

* I decided to define the Asset to display in the banner within a config file vs adding a new field to the `LinkAction` type, which would only be used for this one instance. It felt leaner vs figuring out what to name the field when it's only used for this single Quiz Result template (and possibly never again after this year's election) and then requiring corresponding changes to GraphQL, the `LinkAction` GraphQL fragment, etc. I manually added 3 of the assets into our dev environment so we can build against the same images we want to display in production.

* I spent a good amount of time trying to write a Cypress test to mock displaying a quiz block, hoping to test the redirect based on feature flag. I was copying by example from other Cypress tests, but the Quiz Block kept returning ` Technical Details: GraphQL error: Please return a __typename in "Block" Error: GraphQL error: Please return a __typename in "Block", 1589419807481, ID: 5ebc9f1d3b32bf55b2c8994`.  I tried both visiting the quiz at `/us/blocks/:id` and also created a example fixture of the quiz campaign, but both approaches resulted in that error within the quiz block.

### Relevant tickets

References [Pivotal #172752983](https://www.pivotaltracker.com/story/show/172752983).

### Checklist

### DIFF
--- a/docs/development/features/voter-registration.md
+++ b/docs/development/features/voter-registration.md
@@ -100,13 +100,9 @@ Ungated: `/us/campaigns/ready-vote/quiz/ready` - user can take quiz but must sig
 
 ### Quiz Result Page
 
-We're working on a new release to redirect users to a new `QuizResultPage` component after completing the quiz if the `DS_ENABLE_QUIZ_RESULT_PAGE` configuration variable is set to `true`.
+We're working on a new release to redirect users to `/us/quiz-results/:id` to view their quiz result if the `DS_ENABLE_QUIZ_RESULT_PAGE` configuration variable is set to `true`.
 
-```
-/us/quiz-results/:id
-```
-
-The page expects the `:id` route parameter to be the ID of one of the Link Action entries referenced by the Quiz's `resultBlocks` multi-value reference field.
+The page displays a new `QuizResultPage` component, and expects the `:id` route parameter to be the ID of one of the Link Action entries referenced by the Quiz's `resultBlocks` multi-value reference field.
 
 The `QuizResultPage` displays a static `GalleryBlock` for all of the different result ID's.
 
--- a/docs/development/features/voter-registration.md
+++ b/docs/development/features/voter-registration.md
@@ -102,14 +102,29 @@ The 2020 version will redirect users to a `QuizResultPage` after completing the 
 /us/quiz-results/:id
 ```
 
-The `:id` route parameter corresponds to the Contentful entry containing the quiz result content. Please only edit the quiz entries that link to these entries only if necessary, as they are configured delicately by manually entering JSON.
+The `:id` route parameter corresponds to the Contentful `LinkBlock` entry containing the quiz result content. Please only edit the quiz entries that link to these entries only if necessary, as they are configured delicately by manually entering JSON.
 
-**Dev**:
+**Production**
 
 Quiz results:
 
-| id                     | type      | internalTitle    | title |
-| ---------------------- | --------- | ---------------- | ----- |
-| 347iYsbykgQe6KqeGceMUk | LinkBlock | Super Motivated  |       |
-| 1lvJHhlJqQSgKgwIwUymQ8 | LinkBlock | Social Voter     |       |
-| 2KfkCOTi7u4CqAyyCuGyci | LinkBlock | Election Dabbler |       |
+| id                     | title               | internalTitle      | assetId                |
+| ---------------------- | ------------------- | ------------------ | ---------------------- |
+| p7hqjSP4Y1U6ad0UDz4iS  | Shell-tered Voter   | Vote By Mail       | 49Y4ucuGbJbgZL7IDDfxG0 |
+| 1giTEF3B2hO2CyccmhlVDm | Hare Who Dares      | In-Person Voting   | 2f2kgaHl9w5VtdswKkaBWT |
+| 21PDBge2bKCTWMe5f9eo1H | Sloth At a Loss     | Unsure of Voting   | 1YomtHAeqXJ3qbjQNgsM0v |
+| 14KfeAs265httjNMf1jwTw | Moral Support Panda | Ineligible to Vote | 3WjT0QGNnJEPPz2yMd3inj |
+
+Gallery Block: 78WaGsvDEzAxnreEvNx3Za
+
+**Dev**
+
+Quiz results:
+
+| id                     | title               | internalTitle    | assetId                |
+| ---------------------- | ------------------- | ---------------- | ---------------------- |
+| 347iYsbykgQe6KqeGceMUk | Moral Support Panda | Super Motivated  | 6J13jUL4YGGC1fyYMNEfbc |
+| 1lvJHhlJqQSgKgwIwUymQ8 | Shell-tered Voter   | Social Voter     | 3iLKsRlFQ1k9ddQbRb3RN8 |
+| 2KfkCOTi7u4CqAyyCuGyci | Hare Who Dares      | Election Dabbler | 3uB88eZmTNEaoFxV9pZ8hX |
+
+Gallery Block: 2VGFq3XBcqCfKOA8mC5mP4
--- a/docs/development/features/voter-registration.md
+++ b/docs/development/features/voter-registration.md
@@ -87,3 +87,29 @@ Content:
 ## Tracking Source
 
 TODO: Document how the `r` query parameter is used by the Chompy import.
+
+## Quiz
+
+Our Voter Registration Quiz campaign uses the legacy Quiz content type, which has been deprecated for Typeform except for this one warrior entry, living at path:
+
+```
+/us/campaigns/ready-vote/ready
+```
+
+The 2020 version will redirect users to a `QuizResultPage` after completing the quiz if the `DS_ENABLE_QUIZ_RESULT_PAGE` configuration variable is set to `true`.
+
+```
+/us/quiz-results/:id
+```
+
+The `:id` route parameter corresponds to the Contentful entry containing the quiz result content. Please only edit the quiz entries that link to these entries only if necessary, as they are configured delicately by manually entering JSON.
+
+**Dev**:
+
+Quiz results:
+
+| id                     | type      | internalTitle    | title |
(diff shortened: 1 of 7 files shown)


docs/development/features/voter-registration.md
```diff
@@ -87,3 +87,44 @@ Content:
 ## Tracking Source
 
 TODO: Document how the `r` query parameter is used by the Chompy import.
+
+## Quiz
+
+Our Voter Registration Quiz campaign uses the legacy Quiz content type, which has been deprecated for Typeform except for this one warrior entry, living on all environments at the path:
+
+```
+/us/campaigns/ready-vote/ready
```
mendelB, line 96:
```suggestion
/us/campaigns/ready-vote/quiz/ready
```

docs/development/features/voter-registration.md
```diff
@@ -87,3 +87,44 @@ Content:
...
+
+The 2020 version will redirect users to a `QuizResultPage` after completing the quiz if the `DS_ENABLE_QUIZ_RESULT_PAGE` configuration variable is set to `true`.
+
+```
+/us/quiz-results/:id
+```
+
+The page expects the `:id` route parameter to be the ID of one of the Link Action entries referenced by the Quiz's `resultBlocks` multi-value reference field. Please avoid editing the Quiz entries if possible, as [they are delicately configured](https://github.com/DoSomething/phoenix-next/blob/8b5a97fdd973c8eb925191f78b36c2f676d2707a/docs/content-publishing/quiz.md#adding-available-choices-for-question) (deleting one of the `LinkAction` entries referenced by the `resultBlocks` field would not be pretty).
+
+**Production**
```
mendelB, line 107: Maybe worth calling out Production/QA here?

docs/development/features/voter-registration.md
```diff
@@ -87,3 +87,44 @@ Content:
...
+/us/campaigns/ready-vote/ready
+```
+
+The 2020 version will redirect users to a `QuizResultPage` after completing the quiz if the `DS_ENABLE_QUIZ_RESULT_PAGE` configuration variable is set to `true`.
+
+```
+/us/quiz-results/:id
+```
+
+The page expects the `:id` route parameter to be the ID of one of the Link Action entries referenced by the Quiz's `resultBlocks` multi-value reference field. Please avoid editing the Quiz entries if possible, as [they are delicately configured](https://github.com/DoSomething/phoenix-next/blob/8b5a97fdd973c8eb925191f78b36c2f676d2707a/docs/content-publishing/quiz.md#adding-available-choices-for-question) (deleting one of the `LinkAction` entries referenced by the `resultBlocks` field would not be pretty).
```
mendelB, line 105: I _think_ it might be helpful to also explicitly specify (though I know the table makes it somewhat clear) that each LinkAction result is tied to a particular asset and each group to a particular Gallery Block, and _why_ that is (banner and gallery on results page).

resources/assets/components/Quiz/QuizResult.js
```diff
@@ -27,7 +29,11 @@ const QuizResult = ({ id, campaignId }) => {
     }
   }, [id]);
 
-  return <ContentfulEntryLoader id={id} />;
```
mendelB, line 30: Are we maintaining the authentication gating + signup for the results redirect (within the `useEffect` above)? Or is that just still pending discussion.

resources/assets/components/pages/QuizResultPage/config.js
```diff
@@ -0,0 +1,45 @@
...
+      // Ineligible to vote / Panda
+      '14KfeAs265httjNMf1jwTw': {
+        assetId: '3WjT0QGNnJEPPz2yMd3inj',
+      },
+    },
+  },
+  development: {
+    galleryBlockId: '2VGFq3XBcqCfKOA8mC5mP4',
+    results: {
+      // Super Motivated
```
mendelB, line 26: Are these descriptors different than prod just for kicks?

mendelB:
> I spent a good amount of time trying to write a Cypress test to mock displaying a quiz block, hoping to test the redirect based on feature flag. I was copying by example from other Cypress tests, but the Quiz Block kept returning Technical Details: GraphQL error: Please return a __typename in "Block" Error: GraphQL error: Please return a __typename in "Block", 1589419807481, ID: 5ebc9f1d3b32bf55b2c8994. I tried both visiting the quiz at /us/blocks/:id and also created a example fixture of the quiz campaign, but both approaches resulted in that error within the quiz block.

Strange! The only thing I can think of offhand -- were you perhaps not defining the `...on LinkBlock` or whatever for the nested block fields?

docs/development/features/voter-registration.md
aaronschachter: This is what I've been using to test, the path works: https://qa.dosomething.org/us/campaigns/ready-vote/ready

But I guess the better path maybe is https://qa.dosomething.org/us/campaigns/ready-vote/action since this is what users are directed to after signup?

docs/development/features/voter-registration.md
aaronschachter: > each group to a particular Gallery Block

I'm not sure I follow what you mean by group, do you mean quiz results?  The plan is to display a single Gallery Block to display for all quiz results, not per quiz result.

docs/development/features/voter-registration.md
aaronschachter: Hmm, I think it should be generally known that QA id's equal production ids and would like to avoid repeating it in any instance we talk about production ID's. The OVRD section in this wiki also describes the two as `production` and `development`, matching the config file.

docs/development/features/voter-registration.md
mendelB: Ahhh ok so it's from the campaign directly -- in that case, agreed that `/action` seems right! (I was under the impression from [this whole thing](https://dosomething.slack.com/archives/C09ANFQLA/p1586209006090200?thread_ts=1586208677.087900&cid=C09ANFQLA) that we were redirecting users to the _quiz_ path specifically).

docs/development/features/voter-registration.md
mendelB: Exactly, that's what I mean by group. I guess my confusion might have been that we explicitly outline the `LinkAction` entries being referenced by the quiz as `resultBlocks` but don't explain that all results will display alongside an asset and a gallery. (something that could be gleaned by the config and the component, but it was confusing to me in the table)

resources/assets/components/Quiz/QuizResult.js
aaronschachter: As far as I know, this feature is only for changing the way we display the quiz result: breaking it out of the campaign page layout. Having a specific URL for these makes it much easier to develop/test (vs taking quiz every time to spot check each result).

So... the answer is... yes? we're maintaining it? I have to do some manual testing to make sure the signup mutation still works as expected, but figured it'd be pretty safe since we aren't returning this `Redirect` component until the mutation has completed.

resources/assets/components/pages/QuizResultPage/config.js
aaronschachter: These result blocks are from the quiz that was already set up on the Voter Registration Quiz Campaign (us/campaigns/ready-vote), and wasn't feeling like going through the manual effort of changing the `LinkAction` titles and content to try to match what we have on prod. If we need to, we can change titles and the comments/docs.

docs/development/features/voter-registration.md
aaronschachter: Sounds like it makes sense to document all the different entry points, because that whole thing is/was something isn't very clear (or documented, I'm guessing)

Does this sound right?

* https://qa.dosomething.org/us/campaigns/ready-vote/quiz/ready - An ungated page, where anonymous users can take the quiz. They need to create an account (which also creates a signup) to see their result.

 (I'm assuming this is the case because this is what happened when I tested incognito on QA, I was hit with the authentication paywall)

* https://qa.dosomething.org/us/campaigns/ready-vote/action - If user's entry point was the campaign URL ( https://qa.dosomething.org/us/campaigns/ready-vote) and they signup for the campaign, they are directed to this action page to take the quiz

resources/assets/components/Quiz/QuizResult.js
mendelB: Yeah, ok gotcah - this sounds good to me! Especcially re flow #1 here https://github.com/DoSomething/phoenix-next/pull/2136#discussion_r425277676 and in Tej's doc.

docs/development/features/voter-registration.md
mendelB: Smart -- makes sense!

docs/development/features/voter-registration.md
aaronschachter: Cleaned this up a bit to hopefully clarify.

resources/assets/components/Quiz/QuizResult.js
aaronschachter: Verified that authentication and signup still works as expected from the ungated quiz (had to set a valid `campaignId` in the `additionalContent` of the dev Contentful `quiz` entry first 😅)

mendelB: 👏 ohhh right right, tricky re the `campaignId`, sorry I forgot to call that one out!

PR #2139, opened 2020-05-14
Adds placeholder content for QuizResultPage

aaronschachter:
### What's this PR do?

This pull request will display placeholder content on the `QuizResultPage`, unless a `preview` query parameter is set, e.g. `/quiz-results/p7hqjSP4Y1U6ad0UDz4iS?preview=true`. 

We'll need this to test our work against QA reliably, because our current production content for the voting quiz results contains markup to display the banner image -- so it gets displayed twice. Example: https://qa.dosomething.org/us/quiz-results/p7hqjSP4Y1U6ad0UDz4iS

Once this feature is ready to go live, we can edit our production content to remove the image - as well as removing the placeholder copy, and just displaying the GraphQL content by default.

It also starts on some baseline tests for the new `QuizResultPage` component, added in #2136 

Verify expected behavior:

* By default, we show placeholder content: https://dosomething-phoenix-de-pr-2139.herokuapp.com/us/quiz-results/347iYsbykgQe6KqeGceMUk

* We show GraphQL content when `preview` query is set: https://dosomething-phoenix-de-pr-2139.herokuapp.com/us/quiz-results/347iYsbykgQe6KqeGceMUk?preview=true

When I started on these tests, I was getting the same  `Please return a __typename in "Block"` error that I got in https://github.com/DoSomething/phoenix-next/pull/2136#issuecomment-628667720, when I tried to add coverage for visiting a quiz -- either by visiting the quiz campaigns' action page, or the BlockPage (`/us/blocks/:id`)

With @mendelB's clue around not "defining the ...on LinkBlock or whatever for the nested block fields" - it got me thinking about the GraphQL fixtures we have, and adding a `LinkBlock` in there fixed the GraphQL query.

I kept adding more types to try to get the `ContentfulBlockQuery` to properly mock the Quiz Result Page's `GalleryBlock` -- but still kept seeing the `Please return a __typename in "Block`. error. I can push up that commit in a different branch, wanted to keep this branch in a good place to review/merge.
 
### Relevant tickets

References [Pivotal #172752983](https://www.pivotaltracker.com/story/show/172752983).

### Checklist

### DIFF
--- a/cypress/fixtures/graphql.js
+++ b/cypress/fixtures/graphql.js
@@ -113,6 +113,10 @@ export const mocks = {
     id: args.id,
     __typename: args.__typename || blocks[args.id],
   }),
+  LinkBlock: args => ({
+    title: args.title || faker.lorem.title(),
+    content: args.content || faker.lorem.paragraphs(),
+  }),
   // By default, we'll assume most optional block fields aren't set so
   // that they can just use the defaults in the component itself:
   TextSubmissionBlock: () => ({
--- a/cypress/integration/quiz-result-page.js
+++ b/cypress/integration/quiz-result-page.js
@@ -1,0 +1,62 @@
+/// <reference types="Cypress" />
+
+import faker from 'faker';
+
+const quizResultPagePath = '/us/quiz-results/';
+const quizResultId = '347iYsbykgQe6KqeGceMUk';
+
+/**
+ * @param String id
+ * @param Boolean preview
+ * @return String
+ */
+function getQuizResultPath(id) {
+  return `/us/quiz-results/${id}`;
+}
+
+const linkBlock = {
+  id: quizResultId,
+  __typename: 'LinkBlock',
+  title: faker.company.bsBuzz(),
+  content: faker.lorem.sentence(),
+};
+
+describe('Quiz Result Page', () => {
+  beforeEach(() => cy.configureMocks());
+
+  it('Renders NotFound if GraphQL query does not return a block', () => {
+    cy.mockGraphqlOp('QuizResultPageQuery', {
+      block: null,
+    });
+
+    cy.visit(getQuizResultPath('55767606a59dbf3c7a8b4571'));
+
+    cy.get('[data-test=not-found-page]').should('have.length', 1);
+    cy.get('[data-test=quiz-result-page]').should('have.length', 0);
+  });
+
+  it('Renders placeholder content if preview query is not passed', () => {
+    cy.mockGraphqlOp('QuizResultPageQuery', {
+      block: linkBlock,
+    });
+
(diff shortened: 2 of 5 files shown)


aaronschachter:
Checking the console logs in Cypress, I can see there are a few warnings that look they may be clues?

Screen Shot 2020-05-14 at 4 25 42 PM

cypress/fixtures/graphql.js
```diff
@@ -113,6 +113,10 @@ export const mocks = {
     id: args.id,
     __typename: args.__typename || blocks[args.id],
   }),
+  LinkBlock: args => ({
+    title: args.title || faker.lorem.title(),
+    content: args.content || faker.lorem.paragraphs(),
+  }),
```
aaronschachter, line 119: What's still got me 🤔 here about why this stops the error in mocking the `QuizResultPageQuery` below is it seems like we need anything we reference as a fragment in any query defined here. Ideally I'd prefer to be mocking this in the test below (esp if we want to write tests for components like our Quiz or GalleryBlock that will require quite a bit of fragments)

cypress/integration/quiz-result-page.js
```diff
@@ -0,0 +1,62 @@
...
+    cy.mockGraphqlOp('QuizResultPageQuery', {
+      block: linkBlock,
+    });
+
+    cy.visit(getQuizResultPath(quizResultId));
+
+    cy.get('[data-test=quiz-result-page]').should('have.length', 1);
+    cy.get('h1').should('contain', linkBlock.title);
+    cy.get('[data-test=quiz-result-page-content]').contains(
+      'Saepe cupiditate non. Facere velit vitae corporis.',
```
mendelB, line 48: Could it be cleaner (or somewhat more reliable perhaps) to import the content from the config, to just assert against it in the test without manually typing it out here?

mendelB:
> I kept adding more types to try to get the ContentfulBlockQuery to properly mock the Quiz Result Page's GalleryBlock -- but still kept seeing the Please return a __typename in "Block. error. I can push up that commit in a different branch, wanted to keep this branch in a good place to review/merge.

I wonder if we can add a `Showcaseable` type to the `mocks`, and have it return some defaults for `showcaseable[Title|Description|Image`, and a `GalleryBlock` that mocked a list of random Showcase-able's 🤔 

Or... I wonder if we'd need to (also?) add a `CampaignWebsite` and `Page` to the list

cypress/integration/quiz-result-page.js
aaronschachter: I had tried this -- but it gets rendered as HTML.. which I think is useful to see three paragraphs or so to mock what the real page will look like. Felt simple enough to just test for the first few words, since we'll eventually scrap this query parameter when we're ready to enable the feature anyway.

PR #2141, opened 2020-05-15
Voter Reg Form Updates

lindsayrmaher:
### What's this PR do?

This pull request refactors our form for starting voter registration on the Beta OVRD page to use the form `action` attribute and `GET` method rather than relying on changing the `window.location` in the `handleSubmit`

It also adds hidden inputs with the static values of the other query params added so the URL builds as expected and redirects users to RTV.

There is a small update to the SUMMARY.md file to add the sitewide banner documentation to gitbooks, which I missed in a previous PR!

I just have a question around some of the characters in the URL getting encoded when it redirects and whether that's something we need to be concerned with. 

Expected URL: `https://register.rockthevote.com/registrants/new?partner=37187&source=user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true&email_address=${email}&home_zip_code=${zip}`

URL from form redirect: `https://register.rockthevote.com/registrants/new?partner=37187&source=user%3A5e5edabd4362a2432578c0d3%2Csource%3Aweb%2Csource_details%3Aonlinedrivereferral%2Creferral%3Dtrue&email_address=l%40test.com&home_zip_code=11233`

### Relevant tickets

References [Pivotal # 172439461](https://www.pivotaltracker.com/story/show/172439461).

### Checklist

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -33,6 +33,7 @@
   - [General Buttons](development/features/general-buttons.md)
   - [Paginated Campaign Gallery](development/features/paginated-campaign-gallery.md)
   - [Refer A Friend](development/features/referral-pages.md)
+  - [Sitewide Banner](development/features/sitewide-banner.md)
   - [Sixpack A/B Testing](development/features/sixpack/sixpack-a-b-testing.md)
     - [Code Tests](development/features/sixpack/sixpack-code-tests.md)
     - [Contentful Tests](development/features/sixpack/sixpack-contentful-tests.md)
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
@@ -11,11 +11,12 @@ import PrimaryButton from '../../../utilities/Button/PrimaryButton';
 const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
   const [email, setEmail] = useState('');
   const [zip, setZip] = useState('');
+  // const urlSourceDetails = `user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true`;
 
   const isDisabled = !zip || !email;
   const handleChange = event => {
     const { name, value } = event.target;
-    return name === 'email' ? setEmail(value) : setZip(value);
+    return name === 'email_address' ? setEmail(value) : setZip(value);
   };
 
   const handleSubmit = event => {
@@ -49,7 +50,7 @@ const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
                 className="text-field"
                 required
                 type="email"
-                name="email"
+                name="email_address"
                 value={email}
                 onChange={handleChange}
                 data-id="voter-registration-email-field"
@@ -63,7 +64,7 @@ const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
               <input
                 className="text-field"
                 type="text"
-                name="zip"
+                name="home_zip_code"
                 value={zip}
                 onChange={handleChange}
                 required
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
@@ -11,9 +11,10 @@ import PrimaryButton from '../../../utilities/Button/PrimaryButton';
 const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
   const [email, setEmail] = useState('');
   const [zip, setZip] = useState('');
-  const urlSourceDetails = `user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true`;
 
+  const urlSourceDetails = `user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true`;
   const isDisabled = !zip || !email;
+
   const handleChange = event => {
     const { name, value } = event.target;
     return name === 'email_address' ? setEmail(value) : setZip(value);
@@ -45,7 +46,9 @@ const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
           className="form p-3"
         >
           <input type="hidden" name="partner" value="37187" />
+
           <input type="hidden" name="source" value={urlSourceDetails} />
+
           <div className="form-item stretched">
             <label htmlFor="email" className="font-bold">
               Email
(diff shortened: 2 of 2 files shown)


weerd:
In regards to the URL encoding question, it's a best practice to encode the url in general to replace potentially unsafe characters with safe ones. So it makes sense for the browser to do this automatically.

The only potential problem could be if Rock The Vote for whatever reason can read the encoded URL but I take it that's not the case since you've tested this form submitting to RTV and it works, correct?

If you need to test against the encoded URL to make sure the end result matches what's intended, in your tests you can use [`encodeURI()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI) to encode the example to test against so you can be sure it matches. Or alternatively, just `decodeURI()` the resulting redirect URL the browser constructs and compare against that 😄

aaronschachter:
I know this approach may be semantically preferred - but does this get us further from being able to writing a Cypress test for what the URL is that we'll send a user to? @Dinnall and I were trying to get a `data-url` property working per the #dev-applications conversations over in https://dosomething.slack.com/archives/CUQMU4Q6B/p1589486079246800?thread_ts=1589483455.242300&cid=CUQMU4Q6B.

The test is especially important for one of our stories next sprint, where the URL parameters need to change slightly when we embed this form on the Quiz Result page.

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -41,15 +38,22 @@ const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
...
       >
-        <form onSubmit={handleSubmit} className="form p-3">
+        <form
+          action="https://register.rockthevote.com/registrants/new"
+          method="GET"
+          onSubmit={handleSubmit}
+          className="form p-3"
+        >
+          <input type="hidden" name="partner" value="37187" />
+          <input type="hidden" name="source" value={urlSourceDetails} />
```
aaronschachter, line 48: Although I maybe we can write the Cypress test to check that this input value is as expected upon entering input? I think that might do the trick.

weerd:
> I know this approach may be semantically preferred - but does this get us further from being able to writing a Cypress test for what the URL is that we'll send a user to?

I was chatting with Mendel, and I suspect there should be a way to snag a Cypress form request submission to a third-party and then you should just be able to check against the request URL. I don't know the specifics off-hand but seems like it should be doable.

Maybe @DFurnes knows offhand from his experience w/ Cypress?

weerd:
> I know this approach may be semantically preferred

Just to clarify and make sure we're all on the same page, it's not really a semantics thing. 

We were hacking in some behavior to override how a form naturally works just because of some initial work that was done that lead us down that path. I think we all got into the mindset of needing JS to handle _all the things_™, when in this case we didn't need to and just forgot about that; myself included.

How an HTML form naturally works actually hits all the points that we wanted to address, without having to hack it and sets a better example for future forms that need to behave in a similar fashion.

docs/SUMMARY.md
```diff
@@ -33,6 +33,7 @@
   - [General Buttons](development/features/general-buttons.md)
   - [Paginated Campaign Gallery](development/features/paginated-campaign-gallery.md)
   - [Refer A Friend](development/features/referral-pages.md)
+  - [Sitewide Banner](development/features/sitewide-banner.md)
```
weerd, line 36: haha, sneaking this in 😉

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -41,15 +38,22 @@ const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
...
           <div className="form-item stretched">
```
weerd, line 49: Lets add some carriage return spaces between sibling elements... let'em breathe! 👾

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
weerd: Ya, if snagging the form submission to check the request URL doesn't work out, this feels like a reasonable alternative 👍

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
```diff
@@ -11,16 +11,15 @@ import PrimaryButton from '../../../utilities/Button/PrimaryButton';
 const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
   const [email, setEmail] = useState('');
   const [zip, setZip] = useState('');
+  const urlSourceDetails = `user:${referrerUserId},source:web,source_details:onlinedrivereferral,referral=true`;
```
weerd, line 14: Lets add a carriage return above this `urlSourceDetails` variable to visually separate from the React hooks stuff above it :)

resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js
weerd: @aaronschachter with this approach using the `GET` method, if we can't snag the request to check it and instead check the inputs, they should be set with data when the form is loaded! Which means there shouldn't be any weird order of operations issues with the Cypress tests, waiting for the form submit click to populate variables and running into issues like you and Michael were experiencing in the prior tests! Much simpler tests! Win win :D

DFurnes:
> > I know this approach may be semantically preferred - but does this get us further from being able to writing a Cypress test for what the URL is that we'll send a user to?
> 
> I was chatting with @mendelB, and I suspect there should be a way to snag a Cypress form request submission to a third-party and then you should just be able to check against the request URL. I don't know the specifics off-hand but seems like it should be doable.
> 
> Maybe @DFurnes knows offhand from his experience w/ Cypress?

This is, unfortunately, one of Cypress's [limitations](https://docs.cypress.io/guides/references/trade-offs.html#Same-origin). The way we've solved this in the past is by using an event handler to ["intercept" the redirect](https://github.com/DoSomething/phoenix-next/blob/04c199e5271317260fde8eed58f395871b7d39fe/resources/assets/middleware/requiresAuthentication.js#L26-L31) & then [make assertions based on that](https://github.com/DoSomething/phoenix-next/blob/2523855fbed2eaaea74fb5553fb6110187ed796b/cypress/support/commands.js#L86-L99).

This may be hard with the built-in `<form>` behavior (since the browser adds the query parameters, and we don't have a way to "peek" into that until it happens). We could maybe use an event handler on the form itself to hijack the redirect in tests, but I'm not sure without trying it out. (And that, of course, brings us back to writing JavaScript for this UI!)

I don't have a strong preference on how we implement this, but definitely on board with testing!

weerd:
> This is, unfortunately, one of Cypress's limitations.

Bummer. Ok, well maybe what @aaronschachter suggested with testing the data in the inputs in Cypress gets us in a confident-enough place with the tests for now! 😄

PR #2142, opened 2020-05-15
Adds ScholarshipCard and gallery support

lkpttn:
### What's this PR do?

This pull request adds a new `*Card` component, the `ScholarshipCard` and adds support for it to be displayed in a gallery block. It also lays some foundations for `Gallery` to accept a block type prop to determine how it should render different components that stem from the same content type.  

I.E. - The campaign gallery and scholarship gallery will both be fed from the campaign/`CampaignWebsite` since it has all the fields we need for both cards. We'll use the `blocktype` prop to handle the logic of displaying one versus the other

You can run a gallery locally by altering the default prop from `null` to `'scholarship'` to preview any gallery as a scholarship gallery. It will only render cards for campaigns that have campaigns.

We're using this snippet to sniff out the blocktype if one isn't provided as a prop to keep the galleries working until we can run a script to assign a gallery type to existing galleries.

```
renderBlock(
    blockType || block.__typename,
    block,
    imageAlignment.toLowerCase(),
    imageFit.toLowerCase(),
)
```

### Relevant tickets

References [Pivotal #172203401](https://www.pivotaltracker.com/story/show/172203401).

### Checklist

Screen Shot 2020-05-14 at 1 16 55 PM
Screen Shot 2020-05-15 at 12 09 27 PM
Screen Shot 2020-05-15 at 12 09 09 PM


resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
```diff
@@ -40,14 +43,12 @@ export const GalleryBlockFragment = gql`
   }
 `;
 
-const renderBlock = (block, imageAlignment, imageFit) => {
-  // We finesse the block type and fields to support both
+const renderBlock = (blockType, block, imageAlignment, imageFit) => {
+  // blockType is used to determine what kind of gallery this is going to be
+  // Person, Campaign, Page, Scholarship, Content Block (External)
+
   // GraphQL ('Showcasable' interface) and Phoenix-backend (legacy) queried blocks.
```
mendelB, line 50: Since we've essentially deprecated support for PHP blocks below, we can remove this comment. Maybe we can replace it with a TODO to clean away the PHP specific lower case type `case`s & the manual field mapping below?

resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
```diff
@@ -66,8 +68,12 @@ const renderBlock = (block, imageAlignment, imageFit) => {
         />
       );
 
+    case 'scholarship':
```
mendelB, line 71: re the above comment regarding the types being defined with a lower case as well to account for PHP -- I wonder if to be backward compatible with Gallery Blocks with no `blockType` assigned, we'll have to either
A. hold on to both the lower-case and capitalized types
B. use an upper case in Contentful
C. map to an all caps ENUM in GraphQL and use that for the `blockType` (while retaining the capitalized GraphQL types for backward compatibility).

🤔

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
```diff
@@ -0,0 +1,97 @@
...
+      id
+      scholarshipAmount
+      scholarshipDeadline
+      staffPick
+      url
+    }
+    ... on StoryPageWebsite {
+      id
+      url
+    }
```
mendelB, line 30: I wonder if it's worth removing this from the query to avoid confusion until we support these in this component? It would break as of yet since they don't support scholarship fields. (Though as we discussed IRL, this would also break for campaigns without those fields, so if we're comfortable having it just kaboom for faulty block references...)

lkpttn: I don't think we want it to kaboom, but maybe fail gracefully with a verbose message lol. Right now blocks without correct field just won't show a card at all.

mendelB: Noted, that makes more sense! But hmm as it stands, wouldn't it break since we're invoking methods on the fields below? (`toLocaleString` & `getHumanFriendlyDate` on the scholarship values).

resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
katiecrane: If there's no `blockType` doesn't this return `null`?

resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
```diff
@@ -105,16 +111,27 @@ const galleryTypes = {
...
 
   return (
     <div className="gallery-block">
       {title ? <SectionHeader underlined title={title} /> : null}
+
       <Gallery type={galleryType} className="-mx-3 mt-3">
         {blocks.map(block =>
           renderBlock(
+            // Baby hack to allow scholarship galleries to overwrite the initial blocktype
+            // until a script is run to assign types to all galleries.
```
katiecrane, line 133: Should this be a `@TODO`?

resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
mendelB: Exactly, though we do have a ton of live gallery blocks using the conventional 'type' methodology so I figure we'd have to tread carefully to ensure those are still being rendered properly.

katiecrane: I'm confused I think - if when there's no `blockType` we return `null`, then how do we have Gallery Blocks with no `blockType`? Or am I reading your first comment wrong?

mendelB: Ah no that's a perfectly reasonable question!

So up till this point, we've been relying on each specific block's GraphQL `__typename` to determine how to render it (with the default being `null`). Notabely, these would be the capitalized version (`Campaign`) as opposed to the legacy PHP rendered `campaign` which we don't need to support any longer per #1900 

Here, we're adding the lowercased `scholarship` type, which will read from the _parent_ `GalleryBlock` entry's `blockType` field to determine how to render _all_ blocks in the gallery consistently. (Deprecating support for galleries with mixed block types).

My concern here is that we need to be consistent about our casing for backwards compatibility as we add this new `blockType` option to Contentful for new Gallery Blocks going forward. 

So should they be enum'ed `CAMPAIGN` etc. via a GraphQL resolver, or consistently native GraphQL type friendly `Campaign` or the current lowercased `campaign`...

mendelB: I realize that this essay...is hard to decipher. I think this is a bit of a nuanced string of ideas to articulate properly in text, so happy to hop on a call to discuss if that would be wiser!

katiecrane: Oooooo hmmm I think I'm mostly following now but yeah maybe you and me and @lkpttn could have a quick chat about this?

mendelB: Very much down! I can leave this to @lkpttn to schedule pending what his timeline is with this work, or happy to just throw something on the calendar!

katiecrane: I think it makes most sense for @lkpttn to schedule based on his priorities 👍

lkpttn: I will schedule this.

katiecrane:
@lkpttn ping me when this is ready for a final review so I don't forget! thanks!

lkpttn:
@katiecrane I'm rerunning the circleci build, but this should be ready for re-review. Graphql has been deployed with changes to support the galleryType.

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
katiecrane: @lkpttn I think this question still stands! ^

schema.json
```diff
@@ -14062,18 +14236,6 @@
...
-            "description": "Toggles the display of the page views info card adjacent to the block",
-            "args": [],
-            "type": {
-              "kind": "SCALAR",
-              "name": "Boolean",
-              "ofType": null
-            },
-            "isDeprecated": false,
-            "deprecationReason": null
-          },
```
katiecrane, line 14076: I'm not familiar with this file, does this get generated automatically?

lkpttn: I'm not familiar either, I followed [these instructions](https://dosomething.gitbook.io/phoenix-documentation/installation-+-usage/usage#cypress) to make cypress use the updated schema.

katiecrane: Ah perfect!

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
lkpttn: What I've done is update the return values of the card to display "N/A" if the date or amount is unavailable. This makes it fail nicely if it's a campaign without a scholarship or a story page placed into the gallery.

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
```diff
@@ -0,0 +1,110 @@
...
+        <p className="flex-grow">{showcaseDescription}</p>
+        <div className="pt-4">
+          <div className="float-left pr-8">
+            <h4 className="font-bold uppercase text-gray-600">Amount</h4>
+            <p className="">
+              {scholarshipAmount
+                ? `$${scholarshipAmount.toLocaleString()}`
+                : 'N/A'}
+            </p>
+            {/* <p className="">{`$${scholarshipAmount.toLocaleString()}`}</p> */}
```
katiecrane, line 87: I think this and another line below are leftover!

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
```diff
@@ -0,0 +1,110 @@
...
+          </a>
+        </h1>
+
+        <p className="flex-grow">{showcaseDescription}</p>
+        <div className="pt-4">
+          <div className="float-left pr-8">
+            <h4 className="font-bold uppercase text-gray-600">Amount</h4>
+            <p className="">
+              {scholarshipAmount
+                ? `$${scholarshipAmount.toLocaleString()}`
```
katiecrane, line 84: What are the backtick and `$$` for here?

PR #2132, opened 2020-05-13
PhotoSubmissionAction component button updates, docs and tests! Oh my! 🦁🐯🐻

weerd:
### What's this PR do?

This pull request updates the Photo Submission Action to use the new buttons (last straggler) and fixes up the structure a bit due to the button changes. This last update allows us to remove the older `Button` component.

It also includes an update to the docs to help clarify the current reasoning behind why the `btn` composed class needs to exist.

Lastly, it adds some tests for all the new button components! It utilizes a new testing utility library called [Testing Library](https://testing-library.com/) that works really well with Jest and helps with encouraging good testing practices! I really dug it 🎉 

### How should this be reviewed?

👀 

### Any background context you want to provide?

This has been quite a journey with buttons on Phoenix. I'm tired... 💤 

### Relevant tickets

References [Pivotal #172184401](https://www.pivotaltracker.com/story/show/172184401).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Documentation added for new features/changed endpoints.
- [x] Added appropriate feature/unit tests.

PR #2145, opened 2020-05-15
Fix for broken Ghost Inspector test.

weerd:
### What's this PR do?

This pull request adds a `data-testid` in the photo submission action to make it easier to query this particular item in Ghost Inspector. In #2132, I removed the named classes in favor of using Tailwind utility classes and turns out a test in GI depended on it.

Best to be using `data-testid` anyway for testing purposes!

👀 

The [Testing Library framework](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid) has built in support for querying test IDs as data attributes that are named as `data-testid`, which is slightly different from our current `data-test`. I think it makes sense to switch to this library's recommendation since it has the same support for Cypress tests, etc.

I suspect this would mean updating the [Communal Docs](https://github.com/DoSomething/communal-docs/tree/master/Contributing#data-attributes) to follow the Testing Library's naming scheme.

### DIFF
--- a/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
+++ b/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
@@ -393,7 +393,10 @@ class PhotoSubmissionAction extends PostForm {
                                 total items
                               </span>
 
-                              <span className="block font-league-gothic leading-none text-4xl">
+                              <span
+                                className="block font-league-gothic leading-none text-4xl"
+                                data-testid="quantity-total"
+                              >
                                 {quantity}
                               </span>
                             </div>


DFurnes:
> The [Testing Library framework](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid) has built in support for querying test IDs as data attributes that are named as `data-testid`, which is slightly different from our current `data-test`. I think it makes sense to switch to this library's recommendation since it has the same support for Cypress tests, etc.
> 
> I suspect this would mean updating the [Communal Docs](https://github.com/DoSomething/communal-docs/tree/master/Contributing#data-attributes) to follow the Testing Library's naming scheme.

Yeah, I think this makes great sense! 👍

PR #2146, opened 2020-05-15
More Legacy Landing Page Cleanup - Ruthless Purge

mendelB:
### What's this PR do?

This pull request removes a bunch of deprecated logic/props associated with the legacy landing page and uses the opportunity to clean up the `CampaignPage` & `CampaignPageContent` a bit.

Most notably, thanks to a point @aaronschachter made in [this thread](https://dosomething.slack.com/archives/CUQMU4Q6B/p1589571767013100?thread_ts=1589569201.008200&cid=CUQMU4Q6B), this simplifies the way we somewhat hackily render a Quiz from a campaign page. Which opens us up to further refactoring if we so desire. (Right now the benefit of keeping the quiz here is that we get some free Campaign contextual 'not found' structuring by default through the `CampaignPageContent` for faulty quiz routes (an arguably superfluous benefit TBF)).

This seems like an armful, but I promise, commit-by-commit this and it'll be a straightforward joy! (Round of shots on me if this stresses you out (a win-win for me 🙃)).

Hmmm the main thing I'll call out is that similarly to #2143, this replaces the styling from the `campaign-page.scss` with...nothing really beyond a `leading-normal` and `text-large` to almost perfectly imitate our scss rules from there - but otherwise nil because we already retired `.primary...` here. But it does use Tailwind grid for the Not Found block in https://github.com/DoSomething/phoenix-next/commit/634e91c5c7579e5cace13a8905e67fb661c62fc8 
(I feel like this paragraph didn't make sense but It's Friday afternoon, so apologies!)

Not Found before:
image

After:
image

### Relevant tickets


resources/assets/components/pages/CampaignPage/CampaignPageContent.js
```diff
@@ -1,37 +1,40 @@
...
     page.type === 'page' ? page.fields.slug.endsWith(match.params.slug) : false,
   );
 
   if (!subPage) {
-    return <NotFound />;
+    return (
+      <div className="grid grid-cols-12 px-3 py-6 md:px-12">
+        <NotFound className="col-span-12 lg:col-start-2 lg:col-span-8" />
+      </div>
+    );
```
mendelB, line 25: I didn't want to go too crazy here. This is as barebones of a basic improvement as my energy could muster right now, but I think it sets us up pretty nicely!

resources/assets/components/pages/CampaignPage/CampaignPageContainer.js
```diff
@@ -1,38 +1,23 @@
...
-
-  if (has(ownProps, 'match.params', null)) {
-    const { id, slug } = ownProps.match.params;
-
-    // @TODO: temporary retrieval of single campaign page (quiz) based on matched id or slug.
-    entryContent = findContentfulEntry(state, id || slug);
-  }
+  // If we're on a nested quiz route (/campaigns/:slug/quiz/:slug),
+  // try and find the associated quiz in state.
+  const quizEntry = state.campaign.quizzes.find(
```
weerd, line 12: Is it possible that `state.campaign.quizzes` could be `undefined` and thus throw error when trying to run `.find()`?

resources/assets/components/pages/CampaignPage/CampaignPageContent.js
```diff
@@ -27,7 +27,7 @@ const CampaignPageContent = props => {
     '/us/campaigns/online-registration-drive/action';
 
   return (
-    <div className="campaign-page__content" id={subPage.id}>
+    <div className="leading-normal text-base" id={subPage.id}>
```
weerd, line 35: I suspect there was a reason why the stylesheet needed to ensure the source sans font family was specified. Not sure if maybe somewhere it gets overwritten and needed to be specified in stylesheet to ensure the body copy was source sans 🤷‍♂️ Just want to make sure it's been double-checked!

resources/assets/components/pages/CampaignPage/CampaignPageContent.js
weerd: Does this replicate the `base-12-grid` or meant to work inside it? Otherwise why not use the `base-12-grid` class? I haven't inspected the DOM in review app, just pondering aloud.

mendelB: This replicates the `base-12-grid`, I explained this more in #2143 where I implemented it initially for the Landing Page per your recommendation of moving away from our custom grid classes.

resources/assets/components/pages/CampaignPage/CampaignPageContainer.js
mendelB: I initially used a `lodash.get` here, but realized that per the way we parse the quizzes in the PHP Campaign Entity, it would always be at least an empty array:
https://github.com/DoSomething/phoenix-next/blob/a419fed0456052870fc27b38766446766f28344d/app/Entities/Campaign.php#L67-L69

resources/assets/components/pages/CampaignPage/CampaignPageContent.js
mendelB: The rule seems to be from wayyyyyyy back when we initially added Pages to Phoenix, so I suspect it's outdated. I did double-check that things look the same on the Action page, and things did line up!

resources/assets/components/pages/CampaignPage/CampaignPageContainer.js
weerd: Ah gotcha! Ok, that's what I wasn't sure of, but thanks for confirming!

resources/assets/components/pages/CampaignPage/CampaignPageContent.js
weerd: Ah right right. Ok! 👌

weerd: Oh wait! I think we should use `base-12-grid` if it applies, but then use the tailwind column span classes. Since `base-12-grid` sets up the same foundation across our pages, less likely that padding changes could be introduced etc...

mendelB: Ahh ok, I can update! Just confirming re padding though - weren't we going to remove the padding rules from the `base-12-grid` to allow more customization? (I could be misremembering)

weerd: Yep, but we [already did that](https://github.com/DoSomething/phoenix-next/pull/2008). It was just for the vertical padding, but we ALWAYS want the horizontal padding to be consistent and shouldn't change!

mendelB: Ah I thought that was an April fools

mendelB: lol. Ok will update! I also realize I forgot to account for the `grid-gap`. It'll be nice to get that for free from the class as well!

mendelB: Updated in https://github.com/DoSomething/phoenix-next/pull/2146/commits/5c72610abcb71bb53727c60fe6e0bf8c39c28f31

PR #2151, opened 2020-05-18
Cypress beta url refactor

Dinnall:
### What's this PR do?

This pull request adds a cypress test to @lindsayrmaher [PR ](https://github.com/DoSomething/phoenix-next/pull/2141) refractor work for the beta form,
with this test we account for coverage on the expected URL we'll redirect a user to.

...

With the suggestions from various conversations in [slack](https://dosomething.slack.com/archives/CTVPG6L4R/p1589575301148900) and [PR's](https://github.com/DoSomething/phoenix-next/pull/2141), this appears to be a reasonable solution to best the URL on the beta page and the coming work for the quiz page since the URL will be slightly different. 
...

### Relevant tickets

References [Pivotal #](https://www.pivotaltracker.com/n/projects/2417735/stories/172829662).

### Checklist

### DIFF
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -289,6 +289,10 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
 
     cy.get('[data-id=voter-registration-email-field]').type('text@test.com');
     cy.get('[data-id=voter-registration-zip-field]').type('12345');
+    cy.get('[data-id=voter-registration-sourceDetails').type(
+      `user:${user.id},source:web,source_details:onlinedrivereferral,referral=true`,
+      { force: true },
+    );
 
     cy.get('[data-test=voter-registration-submit-button]').should('be.enabled');
   });
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -289,9 +289,9 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
 
     cy.get('[data-id=voter-registration-email-field]').type('text@test.com');
     cy.get('[data-id=voter-registration-zip-field]').type('12345');
-    cy.get('[data-id=voter-registration-sourceDetails').type(
+    cy.get('[data-id=voter-registration-source-details]').should(
+      'have.value',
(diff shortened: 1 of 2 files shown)


cypress/integration/beta-voter-registration-drive-page.js
```diff
@@ -289,6 +289,10 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
 
     cy.get('[data-id=voter-registration-email-field]').type('text@test.com');
     cy.get('[data-id=voter-registration-zip-field]').type('12345');
+    cy.get('[data-id=voter-registration-sourceDetails').type(
+      `user:${user.id},source:web,source_details:onlinedrivereferral,referral=true`,
+      { force: true },
+    );
```
aaronschachter, line 295: I'm not sure about why we're testing typing into an input field here -- I was thinking we'd use something `like contains` to test that our hidden field value gets updated upon changing the input.

Does this work? (Also fixes the missing closing bracket in line 292)
 ```
cy.get('[data-id=voter-registration-sourceDetails]').contains(`user:${user.id},source:web,source_details:onlinedrivereferral,referral=true`)
```

File: resources/assets/components/pages/VoterRegistrationDrivePage/Beta/StartVoterRegistrationForm.js [hunk 2]
```diff
@@ -47,7 +47,12 @@ const StartVoterRegistrationForm = ({ campaignId, referrerUserId }) => {
         >
           <input type="hidden" name="partner" value="37187" />
 
-          <input type="hidden" name="source" value={urlSourceDetails} />
+          <input
+            type="hidden"
+            name="source"
+            value={urlSourceDetails}
+            data-id="voter-registration-sourceDetails"
```
aaronschachter, line 54: Let's use [kebab-case](https://wiki.c2.com/?KebabCase) for all testing id's to keep consistent.

cypress/integration/beta-voter-registration-drive-page.js
lindsayrmaher: that makes sense @aaronschachter, we were going off the syntax of the other input tests. but this is clearer!

Dinnall: Currently, this test pass, your suggestions of using `contains` should work. I would still need to add `{ force: true }`  because the element is being hidden with CSS. Will give this a try

PR #2155, opened 2020-05-18
Renamed data attribute

Dinnall:
### What's this PR do?

This pull request makes a change to the data attribute being called for testing in [PR](https://github.com/DoSomething/phoenix-next/pull/2151) to `data-testid` being used [here](https://github.com/DoSomething/communal-docs/tree/master/Contributing#tests) for testing.

### How should this be reviewed?

...

### Any background context you want to provide?

...

### Relevant tickets

References [Pivotal #]().

### Checklist

- [ x ] This PR has been added to the relevant Pivotal card.
- [ x ] Added appropriate feature/unit tests.

PR #2156, opened 2020-05-18
Adding Testing Library support for Cypress

weerd:
### What's this PR do?

This pull request adds support for using [Testing Library](https://testing-library.com/docs/cypress-testing-library/intro) utilities in our Cypress tests!

It also updates the `StartVoterRegistrationForm.js` to use the new `data-testid` attribute for testing to then allow us to use the `findByTestId()` helper function in our Cypress tests.

Some of these changes were made in #2155 but they were merged into a non-master branch 😬 

👀

### DIFF
--- a/cypress/integration/beta-voter-registration-drive-page.js
+++ b/cypress/integration/beta-voter-registration-drive-page.js
@@ -43,6 +43,7 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     });
   });
 
+  /** @test */
   it('Beta OVRD page displays NotFoundPage if referrer user ID not present', () => {
     cy.visit(betaPagePath);
 
@@ -53,6 +54,7 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     );
   });
 
+  /** @test */
   it('Beta OVRD page displays NotFoundPage if referrer user not found', () => {
     const user = userFactory();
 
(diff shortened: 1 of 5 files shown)


weerd:
cc @Dinnall

cypress/integration/beta-voter-registration-drive-page.js
```diff
@@ -43,6 +43,7 @@ describe('Beta Voter Registration Drive (OVRD) Page', () => {
     });
   });
 
+  /** @test */
```
mendelB, line 46: Are the `@test` annotations a new convention we want to introduce? I figure maybe worth throwing into the communal docs, or where might the best place be?

Dinnall:
> cc @Dinnall

🤯 I just remembered I didn't merge that branch into master. I should have just made another commit in the same branch.

cypress/integration/beta-voter-registration-drive-page.js
weerd: oh ya, I was trying it out related to this [conversation](https://github.com/DoSomething/phoenix-next/pull/2132#discussion_r425995637).

I was planning on updating the Communal Docs after this now that the Testing Library helper can be used with Cypress :)

weerd:
> 🤯 I just remembered I didn't merge that branch into master. I should have just made another commit in the same branch.

No worries! It should be updated now after merging this PR so just pull down and you should be all set 👯

cypress/integration/beta-voter-registration-drive-page.js
mendelB: Ah the ole Diego breathing room ™️ :) Thanks for that context link!

PR #2161, opened 2020-05-19
Adds Cypress Test to Scholarship Modal for Unauthorized User 

Vmack:
### What's this PR do?

This pull request adds a cypress test to the scholarship modal to assure that unauthorized users who visit from a partner site like fastweb, efficiently visit a campaign page 

👀
This test should be pretty straight forward. It only checks for the campaign and the 'Apply Now' button 

### Relevant tickets

References [Pivotal #172546172](https://www.pivotaltracker.com/story/show/172546172).

### Checklist

### DIFF
--- a/cypress/integration/campaign-signup.js
+++ b/cypress/integration/campaign-signup.js
@@ -137,4 +137,17 @@ describe('Campaign Signup', () => {
     cy.findByTestId('campaign-banner-signup-button').should('not.exist');
     cy.get('.card.affirmation').should('not.exist');
   });
+
+  /** @test */
+  it('Visits a campaign page from scholarship partner, as an unauthenticated user', () => {
+    //Visit the campaign pitch page
+    cy.withState(exampleCampaign).visit(
+      '/us/campaigns/test-example-campaign?utm_campaign=fastweb&utm_source=scholarship',
+    );
+    //We should see the Apply Now button in the modal
+    cy.findByTestId('campaign-info-block-scholarship-details').contains(
+      'button',
+      'Apply Now',
+    );
+  });
 });
--- a/package-lock.json
+++ b/package-lock.json
@@ -3408,7 +3408,7 @@
     "@types/zen-observable": {
       "version": "0.8.0",
       "resolved": "https://registry.npmjs.org/@types/zen-observable/-/zen-observable-0.8.0.tgz",
-      "integrity": "sha512-te5lMAWii1uEJ4FwLjzdlbw3+n0FZNOvFXHxQDKeT0dilh7HOzdMzV2TrJVUzq8ep7J4Na8OUYPRLSQkJHAlrg=="
+      "integrity": "sha1-i2OrfxqlMhJIqtWsiQpIVlbc6k0="
     },
     "@webassemblyjs/ast": {
       "version": "1.9.0",
@@ -3605,7 +3605,7 @@
     "@xtuc/ieee754": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/@xtuc/ieee754/-/ieee754-1.2.0.tgz",
-      "integrity": "sha512-DX8nKgqcGwsc0eJSqYt5lwP4DH5FlHnmuWWBRy7X0NcaGR0ZtuyeESgMwTYVEtxmsNGY+qit4QYT/MIYTOTPeA==",
+      "integrity": "sha1-7vAUoxRa5Hehy8AM0eVSM23Ot5A=",
       "dev": true
     },
     "@xtuc/long": {
@@ -4228,7 +4228,7 @@
     "apollo-link-persisted-queries": {
       "version": "0.2.2",
       "resolved": "https://registry.npmjs.org/apollo-link-persisted-queries/-/apollo-link-persisted-queries-0.2.2.tgz",
-      "integrity": "sha512-YL7XBu/5QsSbbYaWUXgm87T2Hn/2AQZk5Wr8CLXGDr3Wl3E/TRhBhKgQQTly9xhaTi7jgBO+AeIyTH5wCBHA9w==",
+      "integrity": "sha1-FWWXyyWbe7Vs9Olnp74DEpVPRZE=",
       "requires": {
         "apollo-link": "^1.2.1",
         "hash.js": "^1.1.3"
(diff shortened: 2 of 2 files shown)


cypress/integration/hero-campaign-signup.js
```diff
@@ -199,4 +199,19 @@ describe('Hero Template Landing Page', () => {
...
+    //Visit the campaign pitch page
+    cy.withState(exampleCampaign).visit(
+      '/us/campaigns/test-example-campaign?utm_campaign=fastweb&utm_source=scholarship',
+    );
+
+    cy.contains('Example Campaign');
+    cy.contains('This is an example campaign for automated testing');
+
+    //We should see the Apply Now button in the modal
+    cy.contains('button', 'Apply Now');
```
katiecrane, line 215: I think we want to test specifically for the scholarship info block showing up like in [this test](https://github.com/DoSomething/phoenix-next/blob/d03dbcbb69fea84aa3b91dbba146736729a9c8e9/cypress/integration/campaign-info-block.js#L15-L18) with this:
```js
    cy.get('[data-test=campaign-info-block-scholarship-details]').should(
      'have.length',
      1,
    );
```
We want to do that because the signup button on the page is also going to say "Apply Now", but the scholarship info block _only_ appears in the modal so if we see that, we know for sure the modal is showing up!

cypress/integration/hero-campaign-signup.js
```diff
@@ -199,4 +199,19 @@ describe('Hero Template Landing Page', () => {
...
+  it('Visits a campaign page from scholarship partner, as an unauthenticated user', () => {
+    const user = userFactory();
+
+    //Visit the campaign pitch page
+    cy.withState(exampleCampaign).visit(
+      '/us/campaigns/test-example-campaign?utm_campaign=fastweb&utm_source=scholarship',
+    );
+
+    cy.contains('Example Campaign');
+    cy.contains('This is an example campaign for automated testing');
```
katiecrane, line 212: We don't need to check for either of these texts on the page - while we do expect them there, this test only cares about the scholarship modal

cypress/integration/hero-campaign-signup.js
```diff
@@ -199,4 +199,19 @@ describe('Hero Template Landing Page', () => {
     cy.get('mosaic-lede-banner__signup').should('not.exist');
     cy.get('.card.affirmation').should('not.exist');
   });
+
+  it('Visits a campaign page from scholarship partner, as an unauthenticated user', () => {
+    const user = userFactory();
```
katiecrane, line 204: Since we're unauthenticated, we don't need a user at all!

cypress/integration/hero-campaign-signup.js
weerd: @katiecrane @Vmack just FYI, we've documented the new approach to the `data-testid` for use in testing including Cypress tests. It's the thing I linked about in Slack earlier this week. Please update to the new format so we have less to update in the future! https://github.com/DoSomething/communal-docs/tree/master/Contributing#data-attributes

katiecrane: AH this is the old one I see. So we need to update in the test to be `data-testid` and in the actual component. Thanks for flagging @weerd !

weerd: Sure thing! 🚦 👮

weerd: Also you can use the new `cy.findByTestId()` method in the cypress test!

cypress/integration/hero-campaign-signup.js
Vmack: ok!

cypress/integration/hero-campaign-signup.js
Vmack: ok good to know!

cypress/integration/hero-campaign-signup.js
Vmack: thanks @weerd so if I use `cy.findByTestId()` I would still have to make sure the url is visited, correct?

weerd: Not sure I follow:

It's really just a syntax change:
```js
// old:
cy.get('[data-test=campaign-info-block-scholarship-details]')...

// new:
cy.findByTestId('campaign-info-block-scholarship-details')...
```

Vmack: I placed the `data-testid="campaign-info-block-scholarship-details` where the attributes are defined in the `ScholarshipInfoBlock` but when running the test it's still unable to find it 😬is that the correct location? I've tried other locations and I'm getting the same 'unable to find test id' error.

weerd: In the updated code I'm seeing it doesn't show it as `data-testid` but as `data-test`.

katiecrane: @Vmack if you search for `campaign-info-block-scholarship-details` you should be able to find the spot where it was before, which just so happens to be [right here](https://github.com/DoSomething/phoenix-next/blob/0d6c6221ddd8f1d6c31b76f65b7b1785d361c31f/resources/assets/components/CampaignBanner/CampaignBanner.js#L138)

mendelB:
@Vmack - sorry this is annoying - but with #2165 we've deleted the `hero-campaign-signup` file, and instead are relying on the canonical `campaign-signup` (also some of the old `data-test` were updated to `data-testid`). It should be really easy to consolidate though!

If you just copy your changes from `campaign-hero-signup`, rebase off of master, and just paste it into `campaign-signup`!

cc @katiecrane

katiecrane:
hmm @Vmack @mendelB I'm not sure what is happening but this PR now has a bunch of unexpected changes and I see Mendel's commit showing up in the history? (I didn't think that usually happened with rebase?) 🤔

katiecrane:
@Vmack nice all that extra stuff is gone now!

cypress/integration/hero-campaign-signup.js
katiecrane: Nice so @mendelB had already taken care of this change in his PR so this part is already done!

Vmack:
@katiecrane yea @mendelB helped me fix it! 🙌🏽I was freaking out 😩

katiecrane:
good teamwork!! there's always a way out with Git 😇

PR #2165, opened 2020-05-20
Campaign Signup Test Cleanup

mendelB:
### What's this PR do?

This pull request removes the specialized `hero-campaign-signup` Cypress suite we'd set up while juggling Mosaic & Hero templates.

Does this structure make sense? I tried to make as few new decisions as possible, and personally have a few questions about Cypress testing philosophy, but this does hopefully just organizes things a bit better for now.

I think this might cause some merge conflicts with #2161 @Vmack I'd love to chat about how we can consolidate our work (it might just mean simply copy-pasting your assertions into the `campaign-signup` file).

We're extracting some `CampaignInfoBlock` assertions into the `campaign-info-block` specific suite and adding a few simple cases + cleaning things up. We also move some `CampaignBanner` assertions into the main `campaign-signup` suite and update some `data-test`s to `data-testid` per our [new conventions](https://github.com/DoSomething/communal-docs/tree/master/Contributing#tests)!

### Relevant tickets


cypress/integration/campaign-info-block.js
```diff
@@ -2,22 +2,160 @@
...
+
+  const reportbackActionTwo = {
+    actionLabel: 'Sign a Petition',
+    timeCommitmentLabel: '15 minutes - 1 hour',
+    scholarshipEntry: false,
+    reportback: true,
+  };
+
+  const plainAction = {
+    actionLabel: 'Call Tej',
```
DFurnes, line 27: ☎️

PR #2177, opened 2020-05-28
Campaign Info Bar - Zendesk Form Bug

mendelB:
### What's this PR do?

This pull request ensures we don't render a `ZendeskForm` from the `CampaignInfoBar` on _non_ campaign pages where the form wouldn't work. 

- https://github.com/DoSomething/phoenix-next/commit/c558ecd376fba08527b6d891f8877bce0a6133ba adds a helper to allow us to obtain the campaign from window.STATE (and avoid a Redux selector). This will help us determine if we're on a campaign page (and hopefully be useful otherwise for obtaining campaign data without redux).
- https://github.com/DoSomething/phoenix-next/commit/9e071c0dc9ae02e02ccab563ee5ca7f3df514dc7 will ensure we're on a campaign page before rendering the `ZendeskForm` in the `CampaignInfoBar`, using the helper.
- https://github.com/DoSomething/phoenix-next/commit/9ccac5d58883ce848f116f84514f5a4c0294e11a uses a `pageTitle` prop instead of `campaignTitle` to assign into the help email `mailto` link's subject line. (Since we're using this on Story Pages this seems logical and hopefully more clear). We adjust the logic to make it a required prop since we shouldn't default to `'Campaign'` anymore as the subject.
- https://github.com/DoSomething/phoenix-next/commit/ef90904f991dafa6d9fe2f5737a1b332e6c80f16 pass a `pageTitle` to the `CampaignInfoBar` directly. We don't need the container since that's campaign specific.

commit-by-commit! Hopefully this all makes simple logical sense.

We're seeing the help link fail for auth'd users on Story Pages due to an unrelated bug to be fixed in a follow-up PR. This will avoid that bug in the first place since it would only occur on non campaign pages that render the `ZendeskForm`.

### Relevant tickets

References [Pivotal #173063909](https://www.pivotaltracker.com/story/show/173063909).

### DIFF
--- a/resources/assets/components/CampaignInfoBar/CampaignInfoBar.js
+++ b/resources/assets/components/CampaignInfoBar/CampaignInfoBar.js
@@ -11,7 +11,7 @@ const CampaignInfoBar = ({
   affiliateCreditText,
   affiliateSponsors,
   affiliatePartners,
-  campaignTitle,
+  pageTitle,
   contactEmail,
 }) => {
   const [showZendeskModal, setShowZendeskModal] = useState(false);
@@ -45,7 +45,7 @@ const CampaignInfoBar = ({
           ) : (
             <a
               href={encodeURI(
-                `mailto:${contactEmail}?subject=Question About ${campaignTitle}`,
+                `mailto:${contactEmail}?subject=Question About ${pageTitle}`,
               )}
             >
               Email Us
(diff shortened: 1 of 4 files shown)


mendelB:
> Great writeup, and dig the helper. The one thing that through me for a loop was trying to figure out why'd we show a `CampaignInfoBar` on a `StoryPage` - but that's not in scope for a 🐛 🧹

Thanks! And lol very much agreed!! I was debating renaming this to `InfoBar`, maybe that's worth incorporating into #2178?

aaronschachter:
I think it'd be easier to review as separate PR. Maybe `PageInfoBar` vs `InfoBar`?

PR #2178, opened 2020-05-28
Zendesk Form Bug - FAQ helper fix

mendelB:
### What's this PR do?

This pull request branches off of #2177 to update the logic for obtaining a Campaign FAQ page path for the Zendesk Form by updating the Redux selector to a standard helper method and adding some tests.

Do you agree that this is better as a helper? Does the logic check out?

We noticed Zendesk Forms failing to render on Story Pages (addressed in #2177). This cleans up the helper we're using to remove the buggy logic. (Kept this in a separate PR since merging #2177 is higher priority!)

### Relevant tickets

References [Pivotal #173063909](https://www.pivotaltracker.com/story/show/173063909).

### Checklist


resources/assets/helpers/campaign.js
```diff
@@ -25,4 +25,26 @@ export function prepareCampaignPageSlug(campaignSlug, pageSlug) {
...
+export function getCampaignFaqsPath() {
+  const campaign = getCampaign();
+
+  if (!campaign) {
+    return undefined;
+  }
+
+  // Find the FAQs page & grab its slug value.
+  const faqsSlug = get(
+    campaign.pages.find(page => get(page, 'fields.slug', '').match(/faqs?$/)),
```
aaronschachter, line 42: Does this catch `faq` as well as `faqs`? Could be worth a clarifying comment to avoid a web search on regex to answer my (and maybe others, one day) question

resources/assets/helpers/campaign.js
```diff
@@ -25,4 +25,26 @@ export function prepareCampaignPageSlug(campaignSlug, pageSlug) {
  */
 export const getCampaign = () => get(window.STATE, 'campaign');
 
+/**
+ * Get path to Campaign FAQs page.
+ *
+ * @return {String|Undefined}
+ */
+export function getCampaignFaqsPath() {
```
aaronschachter, line 33: Similarly, I wonder about just naming this `Faq` but kinda sounds like a 🖕

resources/assets/helpers/campaign.test.js
```diff
@@ -0,0 +1,28 @@
...
+    campaign: { pages: [{ fields: { slug: 'puppet-party/faq' } }] },
+  };
+
+  /** @test */
+  it('returns fully formed FAQ path when the campaign has an faq page', () => {
+    expect(getCampaignFaqsPath()).toEqual(`/us/campaigns/puppet-party/faq`);
+  });
+
+  /** @test */
+  it('returns fully formed FAQ path when the campaign has an faqs page with a trailing "s"', () => {
```
aaronschachter, line 14: Ah and here's the answer to that question up there

resources/assets/selectors/campaign.js
```diff
@@ -32,23 +32,4 @@ export function getCampaignDataForNorthstar(state) {
...
-    state.campaign.pages.find(page =>
-      get(page, 'fields.slug', '').match(/faqs?$/),
-    ),
-    'fields.slug',
-  );
-
-  // If found, return fully formed path to the FAQs page.
-  return faqsSlug ? `/us/campaigns/${faqsSlug}` : undefined;
-}
-
```
aaronschachter, line 53: What the hay is this file i've never even looked at

resources/assets/helpers/campaign.js
mendelB: Ah good call! I'll add a comment

resources/assets/helpers/campaign.js
mendelB: LOL agreed, this should indeed be `Faq`, (at least then it's only a singular 🖕lol)

mendelB:
Sorry @aaronschachter mind taking one more 👀at this? I added a few commits below to address review + more FAQ`s` + one quick bug fix in https://github.com/DoSomething/phoenix-next/pull/2178/commits/8818a4a833d3e4219c6b5f11c88cf6b5121d759b

resources/assets/components/CampaignInfoBar/CampaignInfoBarContainer.js
```diff
@@ -13,7 +13,7 @@ const mapStateToProps = state => {
     affiliateSponsors: state.campaign.affiliateSponsors,
     affiliatePartners: state.campaign.affiliatePartners,
     contactEmail: get(state, 'campaign.campaignLead.fields.email', undefined),
-    campaignTitle: get(state, 'campaign.title', 'Campaign'),
+    pageTitle: get(state, 'campaign.title', 'Campaign'),
```
aaronschachter, line 16: Whoops, I missed this too! That makes more sense. Forgive my ignorance, but does the current path get sent over in the zendesk ticket? it could be nice to default to window.pathname vs campaign

resources/assets/helpers/campaign.js
```diff
@@ -25,4 +25,30 @@ export function prepareCampaignPageSlug(campaignSlug, pageSlug) {
...
+  if (!campaign) {
+    return undefined;
+  }
+
+  // Find the FAQ page & grab its slug value.
+  const faqsSlug = get(
+    campaign.pages.find(page =>
+      get(page, 'fields.slug', '')
+        // Matches a page with an ".../faq" or ".../faqs" slug.
+        .match(/faqs?$/),
```
aaronschachter, line 45: What also is tricky is that the const is called `faqsSlug`, which does make you wonder if you're looking for FAQ or FAQs if you're not a regex head looking at this. Thanks for adding the comment!

resources/assets/components/CampaignInfoBar/CampaignInfoBarContainer.js
mendelB: This one's actually just for unauthenticated users or non campaign pages where we _don't_ render a zendesk form and instead just use a `mailto` for a standard email. 
https://github.com/DoSomething/phoenix-next/blob/0d7c46b8719b60afea238e3c11a91c75f719bf3a/resources/assets/components/CampaignInfoBar/CampaignInfoBar.js#L46-L52

It gives some context in the subject line of the email per page title.

mendelB: Interestingly, I _don't_ think we actually send over the current path. I don't recall this being a requested attribute when we did this work, but it could probably be useful!

resources/assets/helpers/campaign.js
mendelB: oh my word I seem to miss more and more of these. I'm gonna update that to `faqSlug` before merging!

mendelB: (We funny enough, only added this more generous regex because some campaign faq slugs were mistakenly set to `/faqs`)

mendelB: Well I'll be damned if I'm not more confused after reading this https://english.stackexchange.com/questions/13215/correct-usage-of-faq

resources/assets/selectors/campaign.js
mendelB: It's an ole[Redux convention](https://redux.js.org/recipes/computing-derived-data), using 'selectors' to obtain data from Redux state. We def used to use these more heavily back when, and still are littered with these, probably mostly in Redux container components.

PR #2188, opened 2020-06-01
Updating the SiteWide Banner to exclude Quiz Results pages

lindsayrmaher:
### What's this PR do?

This pull request adds a helper function to check our pathnames that shouldn't display the howdyBar 🤠 

Could the helper be refactored in anyway? I tried to keep it succinct but i didn't know how to avoid da loop and the if statements.

We want to exclude the howdyBar from all 4 quiz results pages, so we need to be able to check for the pathname without the quiz ID.

### Relevant tickets

References [Pivotal # 172898540](https://www.pivotaltracker.com/story/show/172898540).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
+++ b/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
@@ -4,21 +4,17 @@ import React, { useRef, useEffect } from 'react';
 import excludedPaths from './config';
 import SitewideBannerContent from './SitewideBannerContent';
 
-const checkExcludedPathname = pathname => {
-  for (let i = 0; i < excludedPaths.length; i += 1) {
-    if (excludedPaths[i] === pathname) {
-      return true;
+const isExcludedPath = pathname => {
+  return excludedPaths.find(excludedPath => {
+    if (excludedPath.includes('*')) {
+      return (
+        pathname.includes(excludedPath.slice(0, -1)) &&
+        pathname.length > excludedPath.length
+      );
     }
-    if (excludedPaths[i].includes('*')) {
-      if (
-        pathname.includes(excludedPaths[i].slice(0, -1)) &&
-        pathname.length > excludedPaths[i].length
-      ) {
-        return true;
-      }
-    }
-  }
-  return false;
+
+    return excludedPath === pathname;
+  });
 };
 
 const SitewideBanner = props => {
@@ -39,7 +35,7 @@ const SitewideBanner = props => {
 
   const target = usePortal('banner-portal');
 
-  return !checkExcludedPathname(window.location.pathname)
+  return !isExcludedPath(window.location.pathname)
     ? createPortal(children, target)
     : null;
 };
--- a/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
+++ b/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
@@ -10,7 +10,10 @@ const checkExcludedPathname = pathname => {
       return true;
     }
     if (excludedPaths[i].includes('*')) {
-      if (pathname.includes('/us/quiz-results/')) {
+      if (
+        pathname.includes(excludedPaths[i].slice(0, -1)) &&
+        pathname.length > excludedPaths[i].length
+      ) {
         return true;
       }
     }
--- a/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
+++ b/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
@@ -4,6 +4,20 @@ import React, { useRef, useEffect } from 'react';
 import excludedPaths from './config';
 import SitewideBannerContent from './SitewideBannerContent';
 
+const checkExcludedPathname = pathname => {
+  for (let i = 0; i < excludedPaths.length; i += 1) {
+    if (excludedPaths[i] === pathname) {
+      return true;
+    }
+    if (excludedPaths[i].includes('*')) {
+      if (pathname.includes('/us/quiz-results/')) {
+        return true;
+      }
+    }
+  }
+  return false;
+};
+
 const SitewideBanner = props => {
   const usePortal = id => {
     const rootElem = useRef(document.createElement('div'));
@@ -22,7 +36,7 @@ const SitewideBanner = props => {
 
   const target = usePortal('banner-portal');
 
-  return !excludedPaths.includes(window.location.pathname)
+  return !checkExcludedPathname(window.location.pathname)
     ? createPortal(children, target)
     : null;
 };
--- a/cypress/integration/sitewide-banner.js
+++ b/cypress/integration/sitewide-banner.js
@@ -60,6 +60,29 @@ describe('Site Wide Banner', () => {
   });
 
   /** @test */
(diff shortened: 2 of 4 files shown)


resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -4,6 +4,20 @@ import React, { useRef, useEffect } from 'react';
...
+      return true;
+    }
+    if (excludedPaths[i].includes('*')) {
+      if (pathname.includes('/us/quiz-results/')) {
+        return true;
+      }
+    }
+  }
+  return false;
+};
```
mendelB, line 19: What if we changed `'/us/quiz-results/*'` to just `'/us/quiz-results/'` below in the `excludedPaths` (formalizing that entire pathname _categories_ could be excluded), and then we can simplify this logic using [`String.startsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith):
```js
for (let i = 0; i < excludedPaths.length; i += 1) {
  if (pathname.startsWith(excludedPathnames[i])) {
    return true;
  }
}
```

mendelB: And hmmm, if we're already looping, I _think_ using a [`Array.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) here could be nice!

```js
const checkExcludedPathname = pathname => excludedPaths.find(path => pathname.startsWith(path))
```

(We could even inline this logic below!)

aaronschachter: > What if we changed '/us/quiz-results/*' to just '/us/quiz-results/'

I think it'd be cleaner to require admins to add the wildcard. This suggestion could have unexpected behavior where we might want to hide the banner on "Explore Campaigns" at `/us/campaigns/`, but would end up hiding the banner on all campaign pages at `/us/campaigns/:slug`.

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -4,6 +4,20 @@ import React, { useRef, useEffect } from 'react';
 import excludedPaths from './config';
 import SitewideBannerContent from './SitewideBannerContent';
 
+const checkExcludedPathname = pathname => {
+  for (let i = 0; i < excludedPaths.length; i += 1) {
+    if (excludedPaths[i] === pathname) {
+      return true;
+    }
+    if (excludedPaths[i].includes('*')) {
+      if (pathname.includes('/us/quiz-results/')) {
```
aaronschachter, line 13: Instead of hardcoding the quiz results path here, we should be checking for the substring before the `*` :
```suggestion
      if (pathname.includes(excludedPaths[i].slice(0, -1)) {
```

aaronschachter: This assuming that the wildcard is appearing at the end of the string, which I think is fine for our purposes -- we'll just need to spell it out in docs that we only support wildcards at the end (not something like `us/account/*/subscriptions`)

mendelB: Wouldn't this to the contrary of https://github.com/DoSomething/phoenix-next/pull/2188#discussion_r433518187 potentially include a base category (`/us/campaigns/`) when we want to exclude nested pages (`/us/campaigns/*`)?

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
mendelB: Ah good point!

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
mendelB: Hmm maybe we could use some form of regex to be extra sure in this case?
Something roughly like: 
```js
excludedPaths[i].slice(0, -1).match(`pathname${(\/)?(^[\?]).+}`)
```

Which would make sure the pathname isn't a base category 🤔

aaronschachter: > Wouldn't this to the flip side of #2188 (comment) potentially include a base category (/us/campaigns/) when we want to exclude nested pages (/us/campaigns/*)

Yep, good catch. regex sounds like itcould work...  i also wonder about checking the length of `pathname` in this conditional to check whether it's a nested page to avoid using regex (in interest of keeping this readable)

mendelB: That's an interesting idea! So basically ensuring that it's longer than the excluded pathname which means it _has_ a sub path?

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
lindsayrmaher: So to clarify, I should keep the wildcard?

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
lindsayrmaher: @aaronschachter so essentially `pathname.includes(excludedPaths[i].slice(0, -1) && pathname.length > excludedPaths[i].length`

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
aaronschachter: yup!

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
aaronschachter: that looks like it should work!

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -4,6 +4,23 @@ import React, { useRef, useEffect } from 'react';
...
+    if (excludedPaths[i].includes('*')) {
+      if (
+        pathname.includes(excludedPaths[i].slice(0, -1)) &&
+        pathname.length > excludedPaths[i].length
+      ) {
+        return true;
+      }
+    }
+  }
+  return false;
```
mendelB, line 21: It could (arguabely) still be nice to use a `.find` here:
```js
return excludedPaths.find(excludedPath => {
  if (excludedPath.includes('*')) {
    return pathname.includes(excludedPath.slice(0, -1)) && pathname.length > excludedPath.length;
  }
    
  return excludedPath === pathname;
}
```

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -4,6 +4,23 @@ import React, { useRef, useEffect } from 'react';
 import excludedPaths from './config';
 import SitewideBannerContent from './SitewideBannerContent';
 
+const checkExcludedPathname = pathname => {
```
aaronschachter, line 7: It'd be nice to rename this to clarify what we're returning, e.g. `isExcludedPath`

PR #2197, opened 2020-06-03
Adds the Quiz Campaign pathname to list that suppresses the HowdyBar

lindsayrmaher:
### What's this PR do?

This pull request suppresses the sitewide banner from the quiz in general, rather than just from the quiz results page. 

👀 

Having issues with the way the react Router redirect method is interacting with the React portal that's created for the sitewide banner and when it checked the pathnames on redirects!

### Relevant tickets

References [Pivotal # 172898540](https://www.pivotaltracker.com/story/show/172898540).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
+++ b/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
@@ -7,9 +7,10 @@ import SitewideBannerContent from './SitewideBannerContent';
 const isExcludedPath = pathname => {
   return excludedPaths.find(excludedPath => {
     if (excludedPath.includes('*')) {
+      const pathWithoutAsterisk = excludedPath.slice(0, -1);
       return (
         pathname.includes(excludedPath.slice(0, -1)) &&
-        pathname.length > excludedPath.length
+        pathname.length >= pathWithoutAsterisk.length
       );
     }
 
--- a/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
+++ b/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
@@ -8,9 +8,10 @@ const isExcludedPath = pathname => {
   return excludedPaths.find(excludedPath => {
     if (excludedPath.includes('*')) {
       const pathWithoutAsterisk = excludedPath.slice(0, -1);
+
       return (
-        pathname.includes(excludedPath.slice(0, -1)) &&
-        pathname.length >= pathWithoutAsterisk.length
+        pathname.includes(pathWithoutAsterisk) &&
+        pathname.length > pathWithoutAsterisk.length
       );
     }
 
--- a/docs/development/features/sitewide-banner.md
+++ b/docs/development/features/sitewide-banner.md
@@ -18,5 +18,6 @@ If you want to suppress the banner from a page that may have multiple sub paths 
 
 ### Pathnames Currently Excluded:
 
+- `/us/campaigns/ready-vote*`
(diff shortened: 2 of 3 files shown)


resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -7,9 +7,10 @@ import SitewideBannerContent from './SitewideBannerContent';
 const isExcludedPath = pathname => {
   return excludedPaths.find(excludedPath => {
     if (excludedPath.includes('*')) {
+      const pathWithoutAsterisk = excludedPath.slice(0, -1);
       return (
         pathname.includes(excludedPath.slice(0, -1)) &&
-        pathname.length > excludedPath.length
+        pathname.length >= pathWithoutAsterisk.length
```
mendelB, line 13: I think by allowing them to be _equal_ length as well we'd be introducing the possibility of excluding a base path in addition to a categorized path? `/us/campaigns/*` would include a `/us/campaigns/` pathname (like in https://github.com/DoSomething/phoenix-next/pull/2188#discussion_r433872229). Do you think it'd make sense to add both the categorized _and_ base path explicitly? `/us/campaigns/ready-vote` + `/us/campaigns/ready-vote/*`

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -7,9 +7,10 @@ import SitewideBannerContent from './SitewideBannerContent';
 const isExcludedPath = pathname => {
   return excludedPaths.find(excludedPath => {
     if (excludedPath.includes('*')) {
+      const pathWithoutAsterisk = excludedPath.slice(0, -1);
       return (
         pathname.includes(excludedPath.slice(0, -1)) &&
```
mendelB, line 12: Would be sweet to use the `pathWithoutAsterisk` here as well!
```suggestion
        pathname.includes(pathWithoutAsterisk) &&
```

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -7,9 +7,10 @@ import SitewideBannerContent from './SitewideBannerContent';
 const isExcludedPath = pathname => {
   return excludedPaths.find(excludedPath => {
     if (excludedPath.includes('*')) {
+      const pathWithoutAsterisk = excludedPath.slice(0, -1);
       return (
```
aaronschachter, line 11: Nit - this const could use a [new line break](https://github.com/DoSomething/communal-docs/tree/master/Contributing#line-breaks) for some space 🪐

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
aaronschachter: I do think that makes sense @mendelB. I suppose it's not a biggie if we display the banner on `/us/campaigns/ready-vote`, since there's no CTA to register (we don't prompt until user signs up to take the quiz) but likely one for our PM to decide.

lindsayrmaher: I included both the exact path here and the asterisk version on this commit. @aaronschachter do you think it would be better to just go with the asterisk path for now?

aaronschachter: I don't have a preference, but sounds like a call Tej or Diana can make (whether to exclude from landing page)

PR #2200, opened 2020-06-04
Require group select on signup if groups campaign

aaronschachter:
### What's this PR do?

This PR adds support to require a user to select a group before they can signup for a campaign from the landing page, if a `group_type_id` query parameter is passed. Eventually, we'll check to see whether the campaign has a `group_type_id` set in order to determine when to display the Group Select. 

Upon selecting a group and submitting the form, the `group_id` is saved to the signup.

**Scholarship Modal**

Screen Shot 2020-06-08 at 1 22 51 PM

I'll add this review app to Aurora to make it testable, but the signup's group ID won't be visible yet (will add this in a future Rogue PR)

* Half of the changes in this PR are for updates to the GraphQL `schema.json`, to test against the recent addition of `Group` and `GroupType` types in https://github.com/DoSomething/graphql/pull/234 and https://github.com/DoSomething/graphql/pull/237

* I thought it'd make sense to rename the `SignupButton` component as `CampaignSignupForm`, but held off to avoid too many big changes and to hopefully help keep this reviewable.

* Also holding off on documentation for now to keep this reviewable (and because ideally we want to document that we set the group select via Rogue campaign, not query parameter)

### Relevant tickets

References [Pivotal #172541970](https://www.pivotaltracker.com/story/show/172541970).

### Checklist


aaronschachter:
Here's an example: https://dosomething-phoenix-de-pr-2200.herokuapp.com/us/campaigns/online-registration-drive?group_type_id=1

resources/assets/components/SignupButton/SignupButton.js
```diff
@@ -58,19 +66,45 @@ const SignupButton = props => {
...
+    );
+  }
 
   return (
-    <PrimaryButton
-      className={className}
-      onClick={handleSignup}
-      text={isCampaignClosed(endDate) ? 'Notify Me' : buttonCopy}
-    />
+    <div className="mt-3 mb-3" data-testid="join-group-signup-form">
```
mendelB, line 82:
```suggestion
    <div className="my-3" data-testid="join-group-signup-form">
```

resources/assets/components/SignupButton/SignupButton.js
```diff
@@ -44,6 +49,9 @@ const SignupButton = props => {
     storeCampaignSignup(campaignId, {
       body: {
         details: JSON.stringify(details),
+        group_id: groupId,
+        // TODO: When should get this uncommented?
+        // referrer_user_id: query('referrer_user_id'),
```
mendelB, line 54: We do have a ticket to start tracking this for RAF v2 in [Pivotal #172747771](https://www.pivotaltracker.com/story/show/172747771)

resources/assets/components/SignupButton/GroupSelect.js
```diff
@@ -0,0 +1,58 @@
...
+`;
+
+const GroupSelect = ({ groupTypeId, onChange }) => {
+  const { loading, error, data } = useQuery(SEARCH_GROUPS_QUERY, {
+    variables: { groupTypeId },
+  });
+
+  if (loading) {
+    return <Placeholder />;
+  }
```
mendelB, line 30: The Group Select seems to be expanding a bunch when this is loading:
Screen Shot 2020-06-08 at 4 50 03 PM

resources/assets/components/SignupButton/GroupSelect.js
```diff
@@ -0,0 +1,58 @@
...
+    return <ErrorBlock error={error} />;
+  }
+
+  /**
+   * Passing id and instanceId props to the Select for use in our Cypress tests.
+   * @see https://react-select.com/props#select-props
+   */
+  return (
+    <Select
+      id="select-group-dropdown"
```
mendelB, line 42: Are we able to use the `data` prop here for a `data-testid`?

resources/assets/components/CampaignBanner/CampaignBanner.js
```diff
@@ -82,7 +85,7 @@ const CampaignBanner = ({
           >
             {!isAffiliated ? (
               <div
-                data-testid="campaign-banner-signup-button"
+                data-testid="campaign-banner-signup-form"
```
mendelB, line 88: Just a heads up that we'll need to update some Ghost Inspector tests per this new ID. (I _think_ just [this module](https://app.ghostinspector.com/tests/5e5d7511c676e442da8bfcfe)).

resources/assets/components/SignupButton/GroupSelect.js
aaronschachter: I noticed this too. Removed the `minHeight` in b8b137c15fd51bfcc0c878251edf2ffd1beb8ece to get this looking slightly better:
Screen Shot 2020-06-08 at 3 24 39 PM

resources/assets/components/SignupButton/GroupSelect.js
aaronschachter: I don't think so, I can't seem to find a `data` prop in the [React Select API](https://react-select.com/props#select-props )-- unless there's something I'm missing? Setting a `data-testid` prop here does not add it into the source HTML.

resources/assets/components/CampaignBanner/CampaignBanner.js
aaronschachter: Thanks for the heads up -- just to confirm, does this sounds like the correct order of operations?
* Merge this PR
* Change test and run QA suite. At this point, our prod tests will fail for this ID
* If all is good, deploy
* Run production tests, all should be ✅

resources/assets/components/SignupButton/GroupSelect.js
mendelB: Oooh, maybe worth just using the `Spinner` directly? 

https://github.com/DoSomething/phoenix-next/blob/832fcbf80e1c5a616dce0365a5c7dd976a961033/resources/assets/components/PaginatedQuery.js#L21

resources/assets/components/SignupButton/GroupSelect.js
mendelB: Oops sorry, I'd spotted one but it was not for the `<Select>` component!

resources/assets/components/SignupButton/GroupSelect.js
aaronschachter: I did try that.. and it did look better.. but figured I'd try the Placeholder to DRY with min height. Still doesn't look 💯 so I'm up for adding the Spinner instead.

resources/assets/components/CampaignBanner/CampaignBanner.js
mendelB: Since this is a shared module (at least the one I found), you'll need to add support for both QA & prod. What I've been doing is using multi selectors (`[data-testid="x"], [dataid="y"]) to cover both changes on QA & maintain green on prod. Then once deployed you can just remove the legacy selector & re-run!

aaronschachter: I see. I think I'll revert this change for now and revisit when renaming the `SignupButton` component to `CampaignSignupForm` (I think that's a better name, esp with this group select within the `SignupButton` component, no?)

mendelB: That makes sense to me too! Maybe we can even eventually couple the Affiliate opt-in stuff into this form (not more work for you now but in theory lol)
https://github.com/DoSomething/phoenix-next/blob/832fcbf80e1c5a616dce0365a5c7dd976a961033/resources/assets/components/CampaignBanner/CampaignBanner.js#L105-L109

PR #2201, opened 2020-06-04
Globally Unsubscribe from Emails on Phoenix

Vmack:
### What's this PR do?

This pull request sets up the interface for a user to globally unsubscribe from all emails and notifications 

This is a draft but some questions as I am working on this:
How's the naming of the component? Does everything make logical sense? 

This is part 2 of the gql work done for global unsubscribe [here](https://github.com/DoSomething/graphql/pull/225) 

**Update**
I also added a custom getter method in northstar [here](https://github.com/DoSomething/northstar/pull/1029)(suggested by @mendelB 🙌🏽) to fix a bug that triggers when the unsubscribe link is clicked by the user. With both of these changes, a user should now be able to re-subscribe after unsubscribing without issue.

### Relevant tickets

References [Pivotal #172444740](https://www.pivotaltracker.com/story/show/172444740).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
+++ b/resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
@@ -2,7 +2,8 @@ import React from 'react';
 import gql from 'graphql-tag';
 import { useQuery, useMutation } from '@apollo/react-hooks';
 
-import LinkButton from '.../utilities/Button/LinkButton';
+import LinkButton from '../../../utilities/Button/LinkButton';
+import Spinner from '../../../artifacts/Spinner/Spinner';
 
 const EMAIL_SUBSCRIPTION_STATUS = gql`
   query EmailSubscriptionStatus($userId: String!) {
@@ -39,26 +40,28 @@ const CancelEmailSubscription = () => {
     return <p>Something went wrong!</p>;
   }
 
+  if (loading) {
+    return Spinner;
+  }
   return (
     <div>
-      <p>
-        "Need a break?
-        <LinkButton
-          attributes={attributes}
-          className={classes}
-          href=""
-          onClick={() =>
-            updateEmailSubscriptionStatus({
-              variables: {
-                userId,
-                subscribed: false,
-              },
-            })
-          }
-          text="Unsubscribe"
-        />
-        from all newsletters and account notification emails"
-      </p>
+      {data.user.EMAIL_SUBSCRIPTION_STATUS === true ? (
+        <p>
+          Need a break?
+          <LinkButton
+            onClick={() =>
+              updateEmailSubscriptionStatus({
+                variables: {
+                  userId,
+                  subscribed: false,
+                },
+              })
+            }
+            text="Unsubscribe"
+          />
+          from all newsletters and account notification emails
+        </p>
+      ) : null}
     </div>
   );
 };
--- a/resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
+++ b/resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
@@ -1,0 +1,52 @@
+import React from 'react';
+import gql from 'graphql-tag';
+import { useQuery, useMutation } from '@apollo/react-hooks';
+
+import LinkButton from '.../utilities/Button/LinkButton';
+
+const EMAIL_SUBSCRIPTION_STATUS = gql`
+  query EmailSubscriptionStatus($userId: String!) {
+    user(id: $userId) {
+      id
+      emailSubscriptionStatus
+    }
+  }
+`;
+const EMAIL_SUBSCRIPTION_STATUS_MUTATION = gql`
+  mutation EmailSubscriptionStatus(
+    $userId: String!
+    $emailSubscriptionStatus: Boolean!
+  ) {
+    updateEmailSubscriptionStatus(
+      id: $userId
+      emailSubscriptionStatus: $emailSubscriptionStatus
+    ) {
+      id
+      emailSubscriptionStatus
+    }
+  }
+`;
+const CancelEmailSubscription = () => {
+  const options = { variables: { userId: window.ondurationchange.id } };
+
+  const { data, loading, error } = useQuery(EMAIL_SUBSCRIPTION_STATUS, options);
+  const [updateEmailSubscriptionStatus, { loading: modifying }] = useMutation(
+    EMAIL_SUBSCRIPTION_STATUS_MUTATION,
+    options,
+  );
+
+  if (error) {
+    return <p>Something went wrong!</p>;
+  }
+
+  return (
+    <div>
+      <p>
+        "Need a break? Unsubscribe <LinkButton /> from all newsletters and
+        account notification emails"
+      </p>
+    </div>
+  );
+};
+
+export default CancelEmailSubscription;
--- a/resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
+++ b/resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
@@ -1,8 +1,8 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
 import { useQuery, useMutation } from '@apollo/react-hooks';
 
-import LinkButton from '../../../utilities/Button/LinkButton';
 import Spinner from '../../../artifacts/Spinner/Spinner';
 
 const EMAIL_SUBSCRIPTION_STATUS = gql`
@@ -27,11 +27,12 @@ const EMAIL_SUBSCRIPTION_STATUS_MUTATION = gql`
     }
   }
 `;
-const CancelEmailSubscription = () => {
-  const options = { variables: { userId: window.ondurationchange.id } };
+const CancelEmailSubscription = props => {
+  const userId = props.user.id;
+  const options = { variables: { userId } };
 
   const { data, loading, error } = useQuery(EMAIL_SUBSCRIPTION_STATUS, options);
-  const updateEmailSubscriptionStatus = useMutation(
+  const [updateEmailSubscriptionStatus] = useMutation(
     EMAIL_SUBSCRIPTION_STATUS_MUTATION,
     options,
   );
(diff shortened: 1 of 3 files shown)


resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -0,0 +1,79 @@
...
+    EMAIL_SUBSCRIPTION_STATUS_MUTATION,
+    options,
+  );
+
+  if (error) {
+    return <p>Something went wrong!</p>;
+  }
+
+  if (loading) {
+    return Spinner;
```
mendelB, line 45: @Vmack One thing that jumps out at me is this -- Per Diego's comment about that error message in the screenshots, I think we'd want to output the rendered component and not the reference?
```suggestion
    return <Spinner />;
```

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -0,0 +1,79 @@
...
+      emailSubscriptionStatus: $emailSubscriptionStatus
+    ) {
+      id
+      emailSubscriptionStatus
+    }
+  }
+`;
+const CancelEmailSubscription = props => {
+  const userId = props.user.id;
+  const options = { variables: { userId } };
```
mendelB, line 32: @Vmack per [Aaron's comment](https://dosomething.slack.com/archives/CUQMU4Q6B/p1593189750008500?thread_ts=1593187616.006000&cid=CUQMU4Q6B), if we add a `refetchQueries: ['EmailSubscriptionsQuery']` here, this should force those individual email topic subscription queries to refresh which should update without a page refresh!
```suggestion
  const options = { variables: { userId }, refetchQueries: ['EmailSubscriptionsQuery'] };
```

mendelB: We'd also need the `EmailSubscriptionItem` to list this query here as a `refetch` so that toggling an individual topic subscription, eventually causes the cancel link to re-appear when they're _resubscribed_ to email. (But also blocked by [this bug](https://dosomething.slack.com/archives/CUQMU4Q6B/p1593190746009400?thread_ts=1593187616.006000&cid=CUQMU4Q6B))
https://github.com/DoSomething/phoenix-next/blob/1194d7d2e0a3187eed2cea8c244ed9fffe41f4a2/resources/assets/components/pages/AccountPage/Subscriptions/EmailSubscriptionItem.js#L41

`const options = { variables: { userId: window.AUTH.id }, refetchQueries: ['EmailSubscriptionStatus` }; `

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -0,0 +1,79 @@
...
+          Need a break?
+          <button
+            className="px-1 text-blue-500 pb-4"
+            type="button"
+            onClick={() =>
+              updateEmailSubscriptionStatus({
+                variables: {
+                  userId,
+                  emailSubscriptionStatus: false,
+                },
```
mendelB, line 60: And oops, https://github.com/DoSomething/phoenix-next/pull/2201#discussion_r446295352 actually applies here! For the mutation:
```suggestion
              updateEmailSubscriptionStatus({
                variables: {
                  userId,
                  emailSubscriptionStatus: false,
                },
                refetchQueries: ['EmailSubscriptionsQuery'],
```

Vmack: @mendelB should I also add refetchQueries to emailSubscriptionTopics? There's an error with re-subscribing to individual topics after globally unsubscribing

mendelB: I think we will need to refetch the `EmailSubscriptionsQuery` so that we can re-display the cancel subscriptions link once we fix the bug described [here](https://dosomething.slack.com/archives/CUQMU4Q6B/p1593191113010300?thread_ts=1593187616.006000&cid=CUQMU4Q6B). But I'm not quite sure what this error is about 👀

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
DFurnes: I think we're taking advantage of this now, but just noting for future reference that including `id` on each of these queries should let Apollo take care of updating these two UIs for us, without explicitly tying them to each other!

(This works because Apollo keeps an in-memory representation of each object that's returned from a query with an ID, e.g. `User:5543dfd6469c64ec7d8b46b3`. It then updates that in-memory cache whenever it gets new information with the same ID, which will trigger a re-render of any components that depend on that data).

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -0,0 +1,80 @@
...
+  const options = { variables: { userId } };
+
+  const { data, loading, error } = useQuery(EMAIL_SUBSCRIPTION_STATUS, options);
+  const [updateEmailSubscriptionStatus] = useMutation(
+    EMAIL_SUBSCRIPTION_STATUS_MUTATION,
+    options,
+  );
+
+  if (error) {
+    return <p>Something went wrong!</p>;
```
DFurnes, line 41: I'm thinking we should make an `<ErrorMessage>` at some point to DRY these up.

Made [#173628149](https://www.pivotaltracker.com/story/show/173628149) for a future cleanup sprint.

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  if (loading) {
+    return <Spinner />;
+  }
+  return (
```
DFurnes, line 47: Nitpick: Let's include a newline between this conditional and the `return` to give it some breathing room:

```suggestion

  return (
```

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -0,0 +1,80 @@
...
+      {data.user.emailSubscriptionStatus ? (
+        <p>
+          Need a break?
+          <button
+            className="px-1 text-blue-500 pb-4"
+            type="button"
+            onClick={() =>
+              updateEmailSubscriptionStatus({
+                variables: {
+                  userId,
```
DFurnes, line 58: I'm pretty sure you can omit `userId` here since it's set in the default variables on the query, above. (So we'd only have to specify what things are unique to this particular time we're running it – the status boolean!)

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
DFurnes: Oh I see, this is all down here! I'm surprised this doesn't update the other query's results without this argument. 🤔 No need to block merging, but there should be a way to avoid explicitly tying these two together...

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
mendelB: Ooooh I remember you mentioning something like this in that dev roundtable / sprint demo a while back re the causes/subscription components, but I couldn't for the life of me recall explicitly if that would work here 🤦

mendelB: Wait but @DFurnes aren't we indeed including the user ID in the query/mutation above? (Looks like we are in the Topics as well). Should that automatically cause these two to update / is that precisely your point in https://github.com/DoSomething/phoenix-next/pull/2201#discussion_r449085475?

DFurnes: Yeah, my understanding is that this should "just work" since we're including the `id`!

mendelB: Hrmmmm we may need to page Houston here 👩‍🚀  👨‍🚀

Vmack: yea I thought so too which is why I'm wondering why it doesn't work

mendelB: @DFurnes @Vmack Quick update per Diego D's bug [noted in Slack](https://dosomething.slack.com/archives/CP2D7UGAU/p1594821771010600?thread_ts=1594657236.003600&cid=CP2D7UGAU). It seems that we're still not getting an automatic refresh of the subscription status even with our explicit `refetchQueries` solution.

But per Dave's point here, I was wondering why this actually might be the case. 

To me, it _seems_ from the Apollo Mutation docs [here](https://www.apollographql.com/docs/react/data/mutations/#updating-a-single-existing-entity) that we'd actually need to return the modified field explicitly within our query in order to update the cache (emphasis mine):

> To do so, the mutation must return the id of the modified entity, _along with the values of the fields that were modified_. Conveniently, mutations do this by default in Apollo Client.

So our issue might just be that we're not including the 'secondarily' modified `emailSubscriptionStatus`, or contrarily the `emailSubscriptionTopics` from each respective query!

Indeed, on my local dev, when I included those fields explicitly, things updated automatically! _Without_ the `refreshQueries` option.

Vmack: oooohh so then that would mean that I need to just get rid of the redundant `refreshQueries` and just add the modified `emailSubscriptionTopic` to each query?

mendelB: Exactly! I think we'd just need to add the `emailSubscriptionTopic` to the `CancelEmailSubscription` mitation since that's the field that's updated as a 'side effect'
https://github.com/DoSomething/phoenix-next/blob/941dea7923cc451678a29926ef051d361e3baa13/resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js#L17-L28

And add the `emailSubscriptionStatus` field to the `EmailSubscriptionTopic` mutation:
https://github.com/DoSomething/phoenix-next/blob/941dea7923cc451678a29926ef051d361e3baa13/resources/assets/components/pages/AccountPage/Subscriptions/EmailSubscriptionItem.js#L19-L32

Vmack: fantastic

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -0,0 +1,80 @@
...
+const EMAIL_SUBSCRIPTION_STATUS = gql`
+  query EmailSubscriptionStatus($userId: String!) {
+    user(id: $userId) {
+      id
+      emailSubscriptionStatus
+    }
+  }
+`;
+const EMAIL_SUBSCRIPTION_STATUS_MUTATION = gql`
+  mutation EmailSubscriptionStatus(
```
aaronschachter, line 17: I'm attempting to run `npm run schema:fetch` to add a new content type added over in https://github.com/DoSomething/graphql/pull/265, and get an error:


> Error in "Loading queries for dosomething-graphql-dev-lambda": Error: ️️There are multiple definitions for the `EmailSubscriptionStatus` operation. Please rename or remove all operations with the duplicated name before continuing.

It seems like this is where we have multiple definitions -- We can change line 9 to `EmailSubscriptionStatusQuery` (not sure off-hand what our naming convention is for the mutations) and I'd think that'd fix (along with changing the name of the query `EmailSubscriptionStatus` in other spots is used)

PR #2196, opened 2020-06-03
Refactor BaseFigure out of Voter Registration Referrals Block

mendelB:
### What's this PR do?

This pull request replaces the `BaseFigure` in our `VoterRegistrationReferralsBlock` component with inline Tailwind classes.

### How should this be reviewed?
Makes sense? Looks good?

### Any background context you want to provide?
The `BaseFigure` was giving us all sort of headaches (See [Pivotal #172608156](https://www.pivotaltracker.com/n/projects/2441250/stories/172608156)). Toward the effort of potentially extracting this gallery to be used for the RAF tab ([Pivotal #172748788](https://www.pivotaltracker.com/story/show/172748788)), this felt like a good time to try and wrangle out the `BaseFigure`. This allowed removing some float logic in the parent `<div>`s

For the most part, what we were getting from BaseFigure seemed to be 
- centered text for the label (replaced with `text-center`)
- margin-spaced-out nodes (replaced with `justify-around` for auto spacing)
- widths on the nodes themselves (replaced with the series of `w-x` on each item)

I also added a few extra breakpoint width adjustments for the nodes so they'll take up a bit more space on the in-between breakpoints & added a custom `grid-wide` class for when this component is rendered on campaign pages.

### Relevant tickets

References [Pivotal #172748788](https://www.pivotaltracker.com/story/show/172748788).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [x] Added screenshots of front-end changes on small, medium, and large screens.
- [x] Added appropriate feature/unit tests.

Kapture 2020-06-03 at 14 15 47

- [Large](https://user-images.githubusercontent.com/12417657/83672857-f6ad6600-a5a4-11ea-8877-781b5b6583ca.png)
- [Medium](https://user-images.githubusercontent.com/12417657/83672914-0cbb2680-a5a5-11ea-9969-5b4652636d21.png)
- [Small](https://user-images.githubusercontent.com/12417657/83672965-2197ba00-a5a5-11ea-9e6f-99937f33d4ba.png)

PR #2202, opened 2020-06-04
Extract VR Referrals Gallery to Utility component

mendelB:
### What's this PR do?

This pull request - on the heels of #2196 & #2198  - extracts the VR Referrals gallery markup & utility components into a general `ReferralsGallery` utility component.

The big change is that we now expect `referrals` (a list of users) instead of `posts` with nested users. This ensures this can work with posts _or_ signups (RAF will be tracking signup referrals).

It also adds the tests we have implemented in the Cypress `voter-registration-referrals-block.js` suite in a local React Testing Library file (so that we'll have coverage for the RAF implementation as well... I wasn't particularly sure this was explicitly needed, but it felt a bit more central now that it's a general utility).

👀  Not much craziness here

We'll be using this utility for the RAF referrals gallery!

### Relevant tickets

References [Pivotal #172748788](https://www.pivotaltracker.com/story/show/172748788).

### Checklist

- [ ] Documentation added for new features/changed endpoints. - _I'll try and throw some documentation for this new utility in a follow-up_

### DIFF
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/VoterRegistrationReferralsBlock.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/VoterRegistrationReferralsBlock.js
@@ -8,7 +8,7 @@ import { getUserId } from '../../../helpers/auth';
 import EmptyRegistrationImage from './empty-registration.svg';
 import CompletedRegistrationImage from './completed-registration.svg';
 import SectionHeader from '../../utilities/SectionHeader/SectionHeader';
-import VoterRegistrationReferralsList from './VoterRegistrationReferralsList';
+import ReferralsList from '../../utilities/ReferralsGallery/ReferralsList';
 
 export const VoterRegistrationReferralsBlockFragment = gql`
   fragment VoterRegistrationReferralsBlockFragment on VoterRegistrationReferralsBlock {
@@ -58,7 +58,7 @@ const VoterRegistrationReferralsBlock = ({ title }) => (
               </div>
             )}
             <div className="md:flex">
-              <VoterRegistrationReferralsList
+              <ReferralsList
                 referralPosts={data.posts}
                 referralIcon={CompletedRegistrationImage}
                 placeholderIcon={EmptyRegistrationImage}
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/VoterRegistrationReferralsBlock.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/VoterRegistrationReferralsBlock.js
@@ -59,7 +59,7 @@ const VoterRegistrationReferralsBlock = ({ title }) => (
             )}
 
             <ReferralsGallery
-              referralPosts={data.posts}
+              referrals={data.posts.map(post => post.user)}
               referralIcon={CompletedRegistrationImage}
               placeholderIcon={EmptyRegistrationImage}
             />
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/VoterRegistrationReferralsBlock.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/VoterRegistrationReferralsBlock.js
@@ -8,7 +8,7 @@ import { getUserId } from '../../../helpers/auth';
 import EmptyRegistrationImage from './empty-registration.svg';
 import CompletedRegistrationImage from './completed-registration.svg';
 import SectionHeader from '../../utilities/SectionHeader/SectionHeader';
-import ReferralsList from '../../utilities/ReferralsGallery/ReferralsList';
+import ReferralsGallery from '../../utilities/ReferralsGallery/ReferralsGallery';
 
 export const VoterRegistrationReferralsBlockFragment = gql`
   fragment VoterRegistrationReferralsBlockFragment on VoterRegistrationReferralsBlock {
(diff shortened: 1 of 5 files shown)


resources/assets/components/blocks/VoterRegistrationReferralsBlock/VoterRegistrationReferralsBlock.js
```diff
@@ -57,21 +57,12 @@ const VoterRegistrationReferralsBlock = ({ title }) => (
...
-                  data-testid="additional-referrals-count"
-                  className="text-center md:text-left pt-8 md:pt-16 font-bold uppercase text-gray-600"
-                >
-                  {`+ ${numberOfReferrals - 3} more`}
-                </div>
-              ) : null}
-            </div>
+
+            <ReferralsGallery
+              referrals={data.posts.map(post => post.user)}
```
aaronschachter, line 62: One thing that's tripping me up a little bit here is that a list of signup referrals will have different data from a list of post referrals (or even just user registration referrals). And all we need out of this `post.user` is for the sake of a `ReferralsGallery` is their `displayName` field value.

One idea that comes to mind is just calling this `referralLabels` and passing the list of users' `displayName` field values. 

Another idea that goes a bit off the deep end, but curious what you think --  I wonder about whether this thing is a Badges Gallery essentially (that's what I was looking at when [starting with the `Figure`](https://dosomething.slack.com/archives/CUQMU4Q6B/p1588022785073100) for building this component initially)

We could pass an optional `displayLimit` of 3 for when we want to display it for referrals. Our badges profile tab doesn't work well how this is built now because each list item has a different badge... but maybe that could be determined by sending an `items` array prop, that expects label and image source properties...e.g. 

Badges: 
```
items: [{ label: 'Coffee Drinker', src: 'coffee-cup.png'}, { label: 'Pizza Fiend', src: 'empty-pizza- box.png'}]
```

Referrals:
```
items: [{ label: 'Mickey M.', src: 'completed-registration.png'}, { label: 'Donald D.', src: 'completed-registration.png'}]
```

mendelB: @aaronschachter I'm keen on the `referralLabels` idea, since agreed, it's really just a list of `displayName`s after all for labeling the icons... 

The reason I'd be hesitant to merge this with Badges re the deep-end idea is that there's a follow up to this work to allow expanding/collapsing the referrals list to include _all_ referrals, which would presumably entail a bunch of custom referral gallery logic (you can see deets in the designs in this [Pivotal #172748788](https://www.pivotaltracker.com/story/show/172748788)). I _think_ that would be a strong argument for keeping this separate?

mendelB: But to your badges point does -- the `Badge.js` at least, is essentially a `ReferralsListItem.js` (circle icon with label), with empty/complete state, so I wonder if we can't extract the list item as an abstract utility and use it for both 🤔

mendelB: @aaronschachter Just to follow-up here -sorry, I know things are crazy- but if you have a chance to give this another gander today I'd appreciate it! 

I was also hoping that whatever our confirmed plan would be could be addressed in a follow-up PR if this is ok as-is.

One other thought: this new general badge/referral utility node in whatever form, could perhaps end up being the beta `Figure` replacement that we're been [thinking about for a while](https://www.pivotaltracker.com/story/show/172608156), but I also wonder, if, since it's pretty cut & dry right now (icon with two states, label underneath), if it would be simpler to abstract this for referrals & badges and call it a day?

aaronschachter: Abstracting for both sounds good to me, and updating badges for sure doesn't need to be a part of this PR. Are you thinking to hold on renaming `referrals` to `referralLabels` just because we'll eventually have just a `NewFigureGallery` type of utility component? (whatever we call it)

mendelB: Awesome! I was only thinking to address changing to `referralLabels` in a follow-up PR since I jumped the gun and branched  off of this one in #2204 🙃

mendelB: Updated the `referrals` prop to `referralLabels` in https://github.com/DoSomething/phoenix-next/pull/2207.

But oy, I started messing around with using the `ReferralListItem` in the [`Badge` component](https://github.com/DoSomething/phoenix-next/blob/995316236cdd4ce8448e179c49cfd049b8d209b8/resources/assets/components/pages/AccountPage/Badges/Badge.js) and hit a bit of a wall with the `showLock` prop that toggles an overlayed _second_ image on top of the badge image that we use on modals 😖 

I wonder if it's prudent to push off this cleanup until we formally tackle replacing `Figure` with full rigor?

aaronschachter: Agreed! I also didn't realize until megagroom today that there's content associated with each badge -- something we don't need for these referral galleries.

PR #2213, opened 2020-06-10
Groups Voter Registration Drive Action - Part 1/2

aaronschachter:
### What's this PR do?

This pull request modifies the `VoterRegistrationDriveAction` component to query GraphQL for the signup of the campaign that the alpha user is viewing a `VoterRegistrationDriveAction` from. If the signup contains a `group`, a `group_id` query parameter should be added to their OVRD page's share link. Example:

```
/us/my-voter-registration-drive?group_id=7&referrer_user_id=5547be89469c64ec7d8b518d
```

Part 2 will entail appending a `group_id` key/value to the RTV tracking source if found on a OVRD page, so it eventually makes its way onto the voter-reg post's `group_id` field when imported (refs https://github.com/DoSomething/chompy/pull/176)

### How should this be reviewed?

Is this readable, maintainable, and does it make sense? I'll add the review app for this PR into Aurora, and provide links to manually test and spot check that a `group_id` is present if joining a group upon signup.

### Any background context you want to provide?

Created new additions in the `campaign` helper because we'll eventually need to query for a signup from the `VoterRegistrationReferralsBlock` as well, as we'll need to display different data and template if the user is viewing a `VoterRegistrationReferralsBlock` from a groups campaign.

Holding off on too many changes for now to help keep this reviewable (at least half of the changes in this PR are updates to the `schema.json` we used for testing, per GraphQL changes in https://github.com/DoSomething/graphql/pull/244)

### Relevant tickets

References [Pivotal #173019820](https://www.pivotaltracker.com/story/show/173019820).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [x] Added appropriate feature/unit tests.

PR #2217, opened 2020-06-12
Groups Voter Registration Drive Action - Part 2/2

aaronschachter:
### What's this PR do?

This pull request continues #2213 to include a `group_id` key/value in the RTV tracking source if a valid `group_id` query parameter is found on an OVRD page. 

This example URL:
```
/us/my-voter-registration-drive?group_id=4&referrer_user_id=55767606a59dbf3c7a8b4571`
```
will pass the tracking source:
```
user:55767606a59dbf3c7a8b4571,source:web,source_details:onlinedrivereferral,group_id=4,referral=true
```

which will end up in a voter-reg post imported with a `group_id` value of 4 per https://github.com/DoSomething/chompy/pull/176.

If a `group_id` query parameter is passed and there is no group found, the OVRD page will display a `NotFoundPage`. This is because the Chompy request to the Rogue `POST /posts` API endpoint would fail with a `foreign key constraint fails` error -- refs https://github.com/DoSomething/rogue/pull/1042#discussion_r437736621.

👀 

This PR also consolidates a few of the OVRD page Cypress tests that visit a valid OVRD page and test for expected content into a single "View expected info" test. This makes it a little quicker to run all of the tests, because there's less of them -- and I think helps keep the file a bit more maintainable.

### Relevant tickets

Finishes [Pivotal #173019820](https://www.pivotaltracker.com/story/show/173019820).

### Checklist

### DIFF
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/VoterRegistrationDrivePage.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/VoterRegistrationDrivePage.js
@@ -20,6 +20,7 @@ const VOTER_REGISTRATION_DRIVE_PAGE_QUERY = gql`
   query VoterRegistrationDrivePageQuery(
     $referrerUserId: String!
     $voterRegistrationDriveCampaignWebsiteId: String!
+    $groupId: Int!
   ) {
     user(id: $referrerUserId) {
       id
@@ -38,11 +39,17 @@ const VOTER_REGISTRATION_DRIVE_PAGE_QUERY = gql`
       title
       url
     }
+
+    group(id: $groupId) {
+      id
+      name
+    }
   }
 `;
 
 const VoterRegistrationDrivePage = () => {
   const referrerUserId = query('referrer_user_id');
+  const groupId = query('group_id');
 
   const config = isDevEnvironment()
     ? gqlVariables.development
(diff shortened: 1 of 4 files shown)


resources/assets/components/pages/VoterRegistrationDrivePage/VoterRegistrationDrivePage.js
```diff
@@ -67,6 +74,8 @@ const VoterRegistrationDrivePage = () => {
       variables: {
         referrerUserId,
         voterRegistrationDriveCampaignWebsiteId,
+        // This is a hack to avoid passing a null groupId, which is required for the query.
+        groupId: groupId || 0,
```
mendelB, line 78: This feels like a really good use case for the `@include` directive (which per [this thread](https://dosomething.slack.com/archives/CUQMU4Q6B/p1584643484049700) from moons ago, we could activate with a simple update in our GraphQL app). 

But also per that thread, I wonder if we could get away with two separate queries, and the `skip` parameter like we do in the `ScholarshipInfoBlock`
https://github.com/DoSomething/phoenix-next/blob/74d5f4ded3c982f8d0b2dfdcff007b7a6ad52cc7/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js#L64

aaronschachter: I remember trying to get the `@include` to work but think I didn't realize it'd need to be a change on the GraphQL side. I definitely think it's worth cleaning this up in a future PR, thanks for the great suggestions (and code reviews!)

PR #2226, opened 2020-06-22
Progress bar OVRD Groups

lindsayrmaher:
### What's this PR do?

This pull request adds a progress bar to the OVRD groups page so users can easily see how far their group has gotten with registrations. 

👀 
Screen Shot 2020-06-22 at 4 17 32 PM

Screen Shot 2020-06-22 at 4 18 06 PM

N/A

### Relevant tickets

References [Pivotal # 173290162](https://www.pivotaltracker.com/story/show/173290162).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
@@ -59,7 +59,7 @@ const GroupTemplate = ({ group }) => {
         {data => (
           <>
             <ProgressBar
-              goalProgess={data.groupReferrals.length || 35}
+              goalProgess={data.groupReferrals.length}
               goalTotal={group.goal || 50}
               testId="group-progress"
             />
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
@@ -59,7 +59,7 @@ const GroupTemplate = ({ group }) => {
         {data => (
           <>
             <ProgressBar
-              goalProgess={data.groupReferrals.length || 30}
+              goalProgess={data.groupReferrals.length}
               goalTotal={group.goal || 50}
               testId="group-progress"
             />
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
@@ -59,7 +59,7 @@ const GroupTemplate = ({ group }) => {
         {data => (
           <>
             <ProgressBar
-              goalProgess={data.groupReferrals.length}
+              goalProgess={data.groupReferrals.length || 30}
               goalTotal={group.goal || 50}
               testId="group-progress"
             />
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
@@ -59,7 +59,7 @@ const GroupTemplate = ({ group }) => {
         {data => (
           <>
             <ProgressBar
-              goalProgess={data.groupReferrals.length}
+              goalProgess={data.groupReferrals.length || 57}
               goalTotal={group.goal || 50}
               testId="group-progress"
             />
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
@@ -1,8 +1,10 @@
 import React from 'react';
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
+import { css } from '@emotion/core';
 
 import Query from '../../../Query';
+import { tailwind } from '../../../../helpers';
 import { getUserId } from '../../../../helpers/auth';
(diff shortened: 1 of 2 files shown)


resources/assets/components/utilities/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,49 @@
...
+    width: 350px;
+    border-radius: 50px;
+    border: 1px solid #fff;
+  `;
+
+  const progressBar = css`
+    background: ${tailwindYellow['500']};
+    height: 100%;
+    border-radius: inherit;
+    transition: width 5s ease-in;
```
weerd, line 23: Ok, so without testing this out, I think part of the issue if the transition is not working is that first, the width would have to be declared as `0` somewhere as a starting point, cause at the moment, it only looks like the width is declared as the value in `barWidth` but the transition doesn't know what to transition _from_ to this dynamic value. I think the second thing could be that typically with transitions they are triggered on a state change... so like a `hover` or something, or maybe when scrolling into view (depends on where this shows up).

I think first try and see if declaring a starting value for the width and seeing if that helps. Additionally you could try to switch to using CSS Keyframe Animation. I found a pretty cool example via stackoverflow that linked to a nice [JSFiddle example](http://jsfiddle.net/ySj7t/).

lindsayrmaher: I will keep fiddling with this, but i'm not going to worry about it for now since it's not needed for the PR to be ready for review! Thanks for the tips, Diego!

resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
```diff
@@ -57,6 +58,11 @@ const GroupTemplate = ({ group }) => {
       >
         {data => (
           <>
+            <ProgressBar
+              goalProgess={data.groupReferrals.length || 57}
```
aaronschachter, line 62: Is this default `57` a leftover from testing that we should remove?

resources/assets/components/utilities/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,48 @@
...
+      <div className="relative bg-gray-200" css={progressBarContainer}>
+        <div css={progressBar} style={{ width: `${barWidth}%` }} />
+      </div>
+    </div>
+  );
+};
+
+ProgressBar.propTypes = {
+  goalProgess: PropTypes.number.isRequired,
+  goalTotal: PropTypes.number.isRequired,
```
aaronschachter, line 44: This has a typo - `Progess`. What do you think about simplifying the prop names to `target` and `completed`? Don't want to over-build, but I wonder about `target` vs `goal` if we were to ever use this progress bar in other places, e.g. profile completion.

resources/assets/components/utilities/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,48 @@
...
+  testId: PropTypes.string.isRequired,
```
aaronschachter, line 45: Will there be a followup PR with test coverage? It's a nice to have, but if it's not coming any time soon I'd recommend removing this until we use it.

resources/assets/components/utilities/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,48 @@
...
+    height: 100%;
+    border-radius: inherit;
+  `;
+
+  const label =
+    percentCompleted > 100
+      ? `You're ${percentCompleted}% to your goal!`
+      : `${percentCompleted}% To Your Goal!`;
+
+  const barWidth = percentCompleted > 100 ? 100 : percentCompleted;
```
aaronschachter, line 30: Not needed for merge, but ideally we should display the overage if a group exceeds their goal (relevant comments in https://www.pivotaltracker.com/story/show/173290162/comments/215181034, https://www.pivotaltracker.com/story/show/173290162/comments/215351623)

resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
lindsayrmaher: yah yah! i thought i had deleted that 🤦‍♀️

resources/assets/components/utilities/ProgressBar/ProgressBar.js
lindsayrmaher: No i was considering that too, how re usable they would feel. I like `target` and `completed`

resources/assets/components/utilities/ProgressBar/ProgressBar.js
lindsayrmaher: I want to add a couple tests in a follow up PR tomorrow!

resources/assets/components/utilities/ProgressBar/ProgressBar.js
lindsayrmaher: Oh gotcha, I had interpretted that to mean the bar would stay at 100% width and we would display in the copy how much over they were.

If we want to display the overage, maybe i should follow up with Davis on how that should look along with the finalized copy.

aaronschachter: I think the bar would stay at 100%, just the label changes to "🎉 You're at 140% of your goal 🎉 " per the [comment](https://www.pivotaltracker.com/story/show/173290162/comments/215351623).

PR #2230, opened 2020-06-23
Add analytics link to gallery block

Dinnall:
### What's this PR do?

This pull request adds analytics tracking to `ScholarshipCard`

...

We are adding tracking so when a user clicks on one of the blocks from the gallery block it can be tracked by analytics

### Relevant tickets

References [Pivotal # 173132311](https://www.pivotaltracker.com/n/projects/2401401/stories/173132311).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
+++ b/resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
@@ -11,6 +11,7 @@ import {
 import {
   EVENT_CATEGORIES,
   trackAnalyticsEvent,
+  getPageContext,
 } from '../../../helpers/analytics';
 
 // Write a graphql query to get campaign information for a specific id
@@ -62,9 +63,7 @@ const ScholarshipCard = ({ campaign }) => {
       context: {
         url: path,
         campaignId,
-        pageId:
-          get(window, 'STATE.campaign.id', null) ||
-          get(window, 'STATE.page.id', null),
+        pageId: getPageContext(),
       },
     });
   };
--- a/resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
+++ b/resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
@@ -62,8 +62,8 @@ const ScholarshipCard = ({ campaign }) => {
       label: 'scholarship_gallery_card',
       context: {
         url: path,
-        campaignId,
         ...getPageContext(),
+        campaignId,
       },
     });
   };
--- a/resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
+++ b/resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
@@ -7,7 +7,6 @@ import {
   contentfulImageSrcset,
   contentfulImageUrl,
   getHumanFriendlyDate,
-  handleScholarshipCardShareClick,
 } from '../../../helpers';
 
 // Write a graphql query to get campaign information for a specific id
--- a/resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
+++ b/resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
@@ -63,7 +63,7 @@ const ScholarshipCard = ({ campaign }) => {
       context: {
         url: path,
         campaignId,
-        pageId: getPageContext(),
+        ...getPageContext(),
       },
     });
   };
--- a/resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
+++ b/resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
@@ -43,6 +43,7 @@ const ScholarshipCard = ({ campaign }) => {
     scholarshipAmount,
     scholarshipDeadline,
     showcaseImage,
+    campaignId,
     staffPick,
     path,
   } = campaign;
@@ -54,10 +55,17 @@ const ScholarshipCard = ({ campaign }) => {
   ]);
 
   const handleScholarshipCardShareClick = () => {
-    trackAnalyticsEvent('clicked_scholarshipCard_action', {
+    trackAnalyticsEvent('clicked_scholarship_gallery_block_apply_now', {
       action: 'button_clicked',
       category: EVENT_CATEGORIES.siteAction,
       label: 'scholarship_gallery_card',
+      context: {
+        url: path,
+        campaignId,
+        pageId:
+          get(window, 'STATE.campaign.id', null) ||
+          get(window, 'STATE.page.id', null),
+      },
     });
   };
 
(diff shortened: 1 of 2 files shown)


resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
```diff
@@ -49,6 +53,14 @@ const ScholarshipCard = ({ campaign }) => {
     { height: 820, width: 1460 },
   ]);
 
+  const handleScholarshipCardShareClick = () => {
+    trackAnalyticsEvent('clicked_scholarshipCard_action', {
```
mendelB, line 57: It looks like the event name in the ticket is `phoenix_clicked_scholarship_gallery_block_apply_now` which means here, we want to paste that event name exactly, but we could just strip off the `phoenix` bit since that's added automatically:
```suggestion
    trackAnalyticsEvent('clicked_scholarship_gallery_block_apply_now', {
```

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
```diff
@@ -49,6 +53,14 @@ const ScholarshipCard = ({ campaign }) => {
...
+      action: 'button_clicked',
+      category: EVENT_CATEGORIES.siteAction,
+      label: 'scholarship_gallery_card',
```
mendelB, line 60: It looks like in the ticket the `label` is specified as `scholarship_gallery_block`? But I agree with you here that `scholarship_gallery_card` makes more sense, since we're tracking clicks on individual cards and not the gallery block generally.

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
```diff
@@ -49,6 +53,14 @@ const ScholarshipCard = ({ campaign }) => {
...
+    });
```
mendelB, line 61: We'll need to also add a `context` property with `url, campaignId, pageId`:
```js
context: {
  url: path,
  campaignId,
  pageId,
{
```

Though `path` is the only property currently in scope here. 

I suppose for the `pageId`, we could do something similar to this logic in the analytics helper:
https://github.com/DoSomething/phoenix-next/blob/e384d9bfa8db10a3328e178b988674c33f68c18f/resources/assets/helpers/analytics.js#L189-L191

Though this will only work for pages queried via our PHP backend, which excludes e.g. cause & collection pages.

For the `campaignId`, I think we'd need to add the `campaignId` to the list of requested fields in the GraphQL fragment for `campaignWebsite`s in the GalleryBlock here:
https://github.com/DoSomething/phoenix-next/blob/e384d9bfa8db10a3328e178b988674c33f68c18f/resources/assets/components/blocks/GalleryBlock/GalleryBlock.js#L35-L40

And then passing that along to the rendered `ScholarshipCard` here as a new prop:
https://github.com/DoSomething/phoenix-next/blob/e384d9bfa8db10a3328e178b988674c33f68c18f/resources/assets/components/blocks/GalleryBlock/GalleryBlock.js#L74

Dinnall: Thanks for the suggestions Mendel, regarding this comment do you mean to pass it along within the `campaign` prop or make a new prop specific to `campaignId`

mendelB: Ooh that's a good point! We wouldn't actually need to make a new prop since it'll be available under the `campaign` prop just like the other fields.

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
Dinnall: I agree I wasn't sure why block was used here instead of `card` or `action`

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
mendelB: We can just grab it along with the other fields when we destructure the `campaign`
https://github.com/DoSomething/phoenix-next/blob/e384d9bfa8db10a3328e178b988674c33f68c18f/resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js#L35-L44

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
mendelB: Maybe just a good idea to confirm the change with Diego D so he's in the loop?

Dinnall: will do

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
```diff
@@ -49,6 +55,19 @@ const ScholarshipCard = ({ campaign }) => {
...
 
+  const handleScholarshipCardShareClick = () => {
+    trackAnalyticsEvent('clicked_scholarship_gallery_block_apply_now', {
+      action: 'button_clicked',
+      category: EVENT_CATEGORIES.siteAction,
+      label: 'scholarship_gallery_card',
+      context: {
+        url: path,
+        campaignId,
+        pageId: getPageContext(),
```
mendelB, line 66: I don't think we can full on use the `getPageContext` helper here since it returns an object with `campaignId` & `pageId` (and it's pulling the `campaignId` from the `window.STATE.campaign` which wouldn't be populated on most other pages). 

But hmm what we can actually do here is use the spread operator (`...`) to add both fields returned from `getPageContext` directly as fields on the `context` object, and then overwrite the `campaignId` below it:
```suggestion
        ...getPageContext(),
        campaignId,
```

Dinnall: I love this suggestion! Done. 
I tested it and it did return an object

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
```diff
@@ -49,6 +55,19 @@ const ScholarshipCard = ({ campaign }) => {
...
 
+  const handleScholarshipCardShareClick = () => {
+    trackAnalyticsEvent('clicked_scholarship_gallery_block_apply_now', {
+      action: 'button_clicked',
+      category: EVENT_CATEGORIES.siteAction,
+      label: 'scholarship_gallery_card',
+      context: {
+        url: path,
+        campaignId,
+        ...getPageContext(),
```
mendelB, line 66: Awesome! And last thing, this just needs to be flipped around so that the `campaignId` we're providing here manually, will overwrite the one returned from `...getPageContext()`:
```suggestion
        ...getPageContext(),
        campaignId,
```

PR #2232, opened 2020-06-25
Track Referrer User ID in Posts and Signups

mendelB:
### What's this PR do?

This pull request appends the `referrer_user_id` per query parameter of the same name to Campaign Signups (moving it from the `source_details` to a top-level field) as well as Rogue Posts.

👀 

All posts to Rogue from 'Action blocks' should be invoking the `formatPostPayload` where we perform all our furnishing work (location information, UTMs for the source details, etc.), so I added the `referrer_user_id` there.

I was thinking about testing and wasn't sure of the best direction. To me, unit tests on these helpers seem like a good idea and perhaps something for the freezer. Of course though, there's the option of adding a Cypress integration test specifically for this, to see if hitting a campaign with a `?referrer_user_id...` ended up in the signup request payload, but I wasn't sure if that was overkill? Would love thoughts!

### Relevant tickets

References [Pivotal #172747771](https://www.pivotaltracker.com/story/show/172747771).

### Checklist

### DIFF
--- a/docs/development/features/referral-pages.md
+++ b/docs/development/features/referral-pages.md
@@ -53,7 +53,7 @@ The campaign URL that the Beta Page links to will include the alpha's user ID as
 https://www.dosomething.org/us/campaigns/teens-jeans?referrer_user_id=5547be89469c64ec7d8b518d
 ```
 
-This `referrer_user_id` query parameter will be added to the `source_detail` of any new user accounts created by betas.
+This `referrer_user_id` query parameter will be added to the `referrer_user_id` of any new user accounts created by betas, as well as appended to the campaign signup or reportback submitted when the query parameter is attached.
 
 ## Iterations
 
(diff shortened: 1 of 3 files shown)


mendelB:
> > the option of adding a Cypress integration test specifically for this, to see if hitting a campaign with a ?referrer_user_id... ended up in the signup request payload
> 
> I think this would be perfect -- I remember thinking about this for adding the `group_id` to the signup request payload and can't remember if I got stuck or didn't have the time to add the test.

Oh fabulous! Ok I'll add that in a follow-up and can throw in an assertion for groupId as well, which would be at the end of this test, right?
https://github.com/DoSomething/phoenix-next/blob/27b5056e7c868e396f380b91ac6d3deb762fbe62/cypress/integration/campaign-signup.js#L187-L217

aaronschachter:
Yep!

PR #2234, opened 2020-06-25
Rename SignupButton to CampaignSignupForm

aaronschachter:
### What's this PR do?

This pull request simply renames the `SignupButton` components to `CampaignSignupForm`, now that the component contains a "Select Group" dropdown for groups campaigns (the `GroupSelect` component), and will soon additionally contain a "Select State"  dropdown if displaying for a groups campaign that should `filterByState` (refs https://github.com/DoSomething/graphql/pull/250)

👀 

📶 

### Relevant tickets

References https://github.com/DoSomething/phoenix-next/pull/2200#discussion_r437430274, [#173231765](https://www.pivotaltracker.com/story/show/173231765)

### Checklist


weerd:
@aaronschachter  Just noting here that I find the `CampaignSignupForm` name onfusing since there's not really a form per se and it's just a collection of 2 UI elements. If the component uses the term "form" in the name, which is a bit of a loaded term in that I would expect that there's an HTML `<form />` tag involved in the output of the component and related behavior.

I've seen this a bit in other components where "form" is used as a means to just describe a grouping of UI elements and I don't think it's a pattern we should follow. This isn't urgent but something I think we should fix over time; also not sure what would be a more ideal name for the component, maybe `CampaignSignupInterface` (?), but I can help with figuring out a better name if desired 😄

aaronschachter:
I've been using the `Form` naming for components that have buttons for executing `POST` API requests, like this component, the [PostForm](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/actions/PostForm.js) parent class, [CurrentSchoolForm](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/blocks/CurrentSchoolBlock/CurrentSchoolForm.js) -- which is why it doesn't bug me that there isn't any `form` HTML elements found, but I understand how it could be confusing. I'm open to changing the names of these components... could be a good dev round table topic in tandem with revisiting components to clean up or abstract.

weerd:
Ah gotcha. Yeah, I was taking a look at those and it wasn't readily apparent that there's a `POST` API request happening, and instead I was looking for whether there was some `<form />` specified in there.

Early on in Phoenix I know we established (for actions that utilized a form submit, which were fewer and more isolated than maybe we have now so maybe it was a bit easier to categorize) a naming convention (likely not documented or was perviously in some now lost location) of using the term `Submission` as a textual clue that the component included a `<form />` that would be submitted. So components like the `PhotoSubmissionAction`, `TextSubmissionAction`, `PetitionSubmissionAction`, etc. The `Action` suffix of it was meant to imply it "counted" as an action in Rogue that could be counted as a reportback. So for example the `LinkAction` was a link click that would be counted as an action but didn't involve a form submit.

We tried to specifically refrain from using the term "form" since it is a loaded keyword of sorts in web dev, and since we'd likely end up using it to very specifically describe form related components like the `FormValidation` component.

I definitely dig your idea of this potentially being a good Dev Round Table topic, and something we could address in future clean up! I think it'd be good to work through as a team and then document as a reference for how we decide on certain component naming :)

I can add it as a future DRT topic! Thanks for the suggestion!

PR #2235, opened 2020-06-26
Starts on GroupFinder component

aaronschachter:
### What's this PR do?

This pull request moves the `GroupSelect` component rendered in the `CampaignSignupForm` into a new `GroupFinder` component, in anticipation of adding an additional [UsaStateSelect](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/components/utilities/UsaStateSelect.js) utility component (refs #2238). 

The `UsaStateSelect` will be used filter the group search by selected state, if the group type's `filterByState` attribute is to `true` ~~(this is blocked from merge until https://github.com/DoSomething/graphql/pull/250 is deployed)~~ UPDATE: It's been deployed

### How should this be reviewed?

:eyes:

### Any background context you want to provide?

No functional changes yet, this is just prep work for adding the state dropdown, and to help keep the PR's reviewable.

### Relevant tickets

References [Pivotal #173231765](https://www.pivotaltracker.com/story/show/173231765).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2238, opened 2020-06-26
Adds UsaStateSelect to GroupFinder

aaronschachter:
### What's this PR do?

This pull request modifies the Group Finder (added in #2235) to require user to select a US state first before selecting a group, if the campaign group type has a truthy `filter_with_state` attribute.

It also renames the `filterByState` prop in the existing `SchoolSelect` component to `schoolState`, to avoid confusion with the newly added `filterByState` field on our `GroupType` type in Rogue.

:eyes:

* Similar to https://github.com/DoSomething/rogue/pull/1058, this won't be testable in dev or QA until we get final files to import NASSP groups (which will have group types that filter by state, along with `city` and `state` values populated in the groups).

* I'll be adding documentation for Groups in a future PR to help keep this reviewable.

### Relevant tickets

References [Pivotal #173231765](https://www.pivotaltracker.com/story/show/173231765).

### Checklist

- ~~[ ] Documentation added for new features/changed endpoints.~~ -- will be provided in a future PR

### DIFF
--- a/cypress/integration/campaign-signup.js
+++ b/cypress/integration/campaign-signup.js
@@ -212,9 +212,9 @@ describe('Campaign Signup', () => {
     );
   });
 
-  // TODO: Add context
+  // TODO: Use cypress context to better group this test once #2238 is merged.
   /** @test */
-  it('Signup button is enabled after selecting a group', () => {
+  it('If campaign group type does not filter by state, signup button is enabled after selecting group', () => {
     const user = userFactory();
 
     cy.mockGraphqlOp('SearchGroupsQuery', {
--- a/cypress/integration/campaign-signup.js
+++ b/cypress/integration/campaign-signup.js
@@ -265,77 +265,78 @@ describe('Campaign Signup', () => {
     cy.wait('@signupRequest')
       .its('request.body.group_id')
       .should('equal', 1);
-  }),
-    context(
-      'Campaign ID configured with group type that filters by state',
-      () => {
-        /** @test */
-        it('Signup button is enabled after selecting state, then group', () => {
-          const user = userFactory();
+  });
 
-          cy.mockGraphqlOp('SearchGroupsQuery', {
-            groups: [
-              { id: 1, name: 'New York', state: 'NY' },
-              { id: 2, name: 'Philadelphia', state: 'PA' },
-              { id: 3, name: 'San Francisco', state: 'CA' },
-            ],
-          });
+  context(
+    'Campaign ID configured with group type that filters by state',
+    () => {
+      /** @test */
+      it('Signup button is enabled after selecting state, then group', () => {
+        const user = userFactory();
 
-          cy.mockGraphqlOp('CampaignBannerQuery', {
-            campaign: {
-              id: campaignId,
-              groupTypeId: 1,
-              groupType: {
-                id: 1,
-                filterByState: true,
-              },
-            },
-          });
-
-          // Visit the campaign pitch page
-          cy.withState(exampleCampaign).visit(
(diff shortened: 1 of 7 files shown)


resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
```diff
@@ -63,7 +63,7 @@ const CampaignSignupForm = props => {
   };
 
   const handleGroupFinderChange = selected => {
-    setGroupId(selected.id);
+    setGroupId(selected ? selected.id : null);
```
mendelB, line 66: Does this have to do with the clear state bugginess you mentioned?

cypress/integration/campaign-signup.js
```diff
@@ -264,4 +265,74 @@ describe('Campaign Signup', () => {
...
+        cy.findByTestId('join-group-signup-form').should('have.length', 1);
+        cy.findByTestId('campaign-banner-signup-button').contains(
+          'button',
+          'Join Group',
+        );
+        cy.findByTestId('join-group-signup-button').should('be.disabled');
+        cy.get('#select-state-dropdown').should('have.length', 1);
+        cy.get('#select-group-dropdown').should('have.length', 0);
+        cy.get('#select-state-dropdown').click();
+        cy.get('#react-select-select-state--input').type('new');
```
mendelB, line 306: Small fry and I don't even know if this makes sense, but I wonder for readability if we can drop the individual `select` prefix per Select since they're always prefixed with `react-select` and are always a selection of some sort?

cypress/integration/campaign-signup.js
```diff
@@ -264,4 +265,74 @@ describe('Campaign Signup', () => {
...
+        // Select "New York"
+        cy.get('#react-select-select-state--option-32').click();
+        cy.findByTestId('join-group-signup-button').should('be.disabled');
+        cy.get('#select-group-dropdown').should('have.length', 1);
+        // Forcing these actions because this element gets covered by the SitewideBanner.
```
mendelB, line 311: Yikes, what's going on here with the `SitewideBanner`?

resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
aaronschachter: It wasn't the bug -- although technically our school finder needs this change in order to support the `isClearable` prop in `AsyncSelect` we're passing.

The bug I found when clearing/retrying input was fixed by https://github.com/DoSomething/phoenix-next/pull/2238/commits/c4a81eeefe8cd22be532543d2625a8b50a56faa2.

cypress/integration/campaign-signup.js
aaronschachter: I've tried commenting out the content in the `landingPage` field for the example campaign fixture, hoping that it's the extra scrolling that causes the bug -- but still occurs without a `landingPage`

mendelB: Ahh dang that's no good. Would [`scrollIntoView`](https://docs.cypress.io/api/commands/scrollintoview.html#Syntax) work here? My thinking is that if this really isn't able to be reached for users due to a blocking element we'd want to know if possible. (I also wonder if we should - someday - bake in some way to optionally hide the sitewide banner e.g. for tests like this 🤔 )

cypress/integration/campaign-signup.js
aaronschachter: I think it makes sense, I can look into making this change in an upcoming documentation / cleanup PR

PR #2239, opened 2020-06-29
Site Feature Config Env Variables

mendelB:
### What's this PR do?

This pull request adds a formal `feature-config`, config category, `window.ENV.FEATURE_CONFIG` variable, and  `featureConfig` helper to expose configuration variables controlling feature specific functionality.

How's the naming? Would this be better as just `site-config`? Does this make sense in general?

Per this conversation https://github.com/DoSomething/phoenix-next/pull/2080#discussion_r417793699 where @aaronschachter pointed out that the `featureFlag` et al functionality, which applies nicely to feature flags, doesn't quite cover the general configuration variables we might need across the app.

This is in anticipation of [Pivotal #172627879](https://www.pivotaltracker.com/story/show/172627879), where we'll need to introduce a feature configuration which would be subject to change and would be convenient to edit per environment via UI.

### Relevant tickets

References [Pivotal #172627879](https://www.pivotaltracker.com/story/show/172627879).

### Checklist

- [ ] Documentation added for new features/changed endpoints. - _if this sounds good, I'll add some documentation_


app/Providers/AppServiceProvider.php
```diff
@@ -49,6 +49,7 @@ public function boot(Router $router)
                 'SIXPACK_TIMEOUT' => config('services.sixpack.timeout'),
                 'CONTENTFUL_USE_PREVIEW_API' => config('contentful')['delivery.preview'],
                 'FEATURE_FLAGS' => config('features'),
+                'FEATURE_CONFIG' => config('feature-config'),
```
aaronschachter, line 52: > Would this be better as just `site-config`?

At first thought, I do prefer `site-config`.. I think it's a little bit more clear on what it's used for. Some of these settings could be used within multiple features or "feature-less" sitewide config.

mendelB: I think I confused myself, tripping up over this `site-config` vs the general `config.php` and associated ENV variables also being 'site configuration'. But agreed with your point above, and I _think_ the distinction is clear enough? Updated in https://github.com/DoSomething/phoenix-next/pull/2239/commits/acd9d790a5d832fdb92dab4682a3783dbb9c70cd

weerd: What about just `SETTINGS`? 🤷‍♂️

DFurnes: This is super nitpicky, but I wonder if `site.php` is better? I definitely see the need for something like this and can see how "feature" might be too limited scope, but it feels weird to have `config/*-config.php`!

DFurnes: My mental model here is that `config/app.php` configures the _Laravel_ application, `config/features.php` configures boolean feature flags, and `config/site.php` configures our website-specific settings.

weerd: ooooo, ya I dig `site.php` better @DFurnes!

aaronschachter: I like that too. I wonder if it's worth explicitly renaming `features.php` to `feature-flags.php`, to help clarify these are settings that one day will pass (either graduating to `site.php` or ☠️ )

DFurnes: I like that! 👍

mendelB: This makes great sense!! Updated to `site` in https://github.com/DoSomething/phoenix-next/pull/2239/commits/6fd62034b46df056d11d6f71f03f4f8b2bfc5fb4 and `feature-flags` in https://github.com/DoSomething/phoenix-next/pull/2239/commits/116ed7984220417b6ce2155dd848818f12e7be50

mendelB: I did keep the helper names `siteConfig` though, which I think makes sense for clarity?

DFurnes: Yeah, that makes sense to me to me too. 👍

PR #2242, opened 2020-06-29
Beta Referral Page - bring back that - Secondary Campaign Referral Link

mendelB:
### What's this PR do?

This pull request displays a secondary (default?) referral campaign link on the RAF Beta page. Also adds/fixes the site config for 'default referral campaign ID', and adds support in Cypress to override the site config.

With a sensitive but strong eye to make sure this logic checks out.

I added the feature flag and a default campaign ID config of `9001` to the [review app](https://dosomething-phoenix-de-pr-2242.herokuapp.com/us/join?user_id=5bbd00271f1c570099415d02&campaign_id=9002).

This is feature-flagged behind the RAF V2.

I started reconsidering this site config `default_` prefix since we use this as the secondary campaign on the Beta page, and will use it (it's main purpose) as the referral `campaign_id` from the account alpha page. But I didn't have any better ideas. `primary`? *Plus, we _may_ end up using this as a default/fallback for the alpha/beta pages if the provided ID is invalid/missing. 

*[I did promise](https://www.pivotaltracker.com/story/show/172627879/comments/215681911) this much for the Alpha page, but I'm actually reconsidering whether this would generally be a good idea since it introduces a fair amount of complexity.

### Relevant tickets

References [Pivotal #172627879](https://www.pivotaltracker.com/story/show/172627879).

### Checklist

- [ ] Documentation added for new features/changed endpoints. - _I'm holding off on documentation for a bit since there are more moving pieces_

**Secondary Campaign Link**
image

**Not if the default config ID matches the provided `campaign_id`**
image

### DIFF
--- a/resources/assets/components/pages/ReferralPage/Beta/BetaPage.js
+++ b/resources/assets/components/pages/ReferralPage/Beta/BetaPage.js
@@ -3,10 +3,10 @@ import gql from 'graphql-tag';
 
 import Query from '../../../Query';
 import ErrorPage from '../../ErrorPage';
-import { query } from '../../../../helpers';
 import MoneyHandImage from './money-hand.svg';
 import CampaignLink from './BetaPageCampaignLink';
 import ArticleHeader from '../../../utilities/ArticleHeader';
+import { featureFlag, query, siteConfig } from '../../../../helpers';
 import { getReferralCampaignId } from '../../../../helpers/refer-friends';
 
 const REFERRAL_PAGE_USER = gql`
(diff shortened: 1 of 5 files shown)


resources/assets/components/pages/ReferralPage/Beta/BetaPage.js
```diff
@@ -22,6 +22,12 @@ const BetaPage = () => {
   const userId = query('user_id');
   const campaignId = getReferralCampaignId();
 
+  const defaultCampaignId = siteConfig('default_referral_campaign_id');
+
+  // We'll display a secondary campaign link option for RAF V2.
+  const displaySecondaryCampaign =
+    featureFlag('refer_friends_v2') && defaultCampaignId !== campaignId;
+
```
Dinnall, line 30: Everything looks good to me. Looking at one of your other files I know `default_referral_campaign_id` is set as a string, I'm curious to know if this is checking the same type here. Does `campaignId ` return a number or string?

mendelB: Good call! This actually tripped me up because I originally set the `default_referral_campaign_id ` as an integer in the Cypress config. 

Both are expected to be strings -- the `campaignId` is being parsed from the URL as a string, and the config `defaultCampaignId` (along with any plain value we set from `.env` by default I believe) would always a string as well.

Dinnall: Cypress is a weird world in itself, but thanks. This makes sense.

PR #2249, opened 2020-07-01
add a br tag

Dinnall:
### What's this PR do?

This pull request adds a `<br />` tag to add space for actions 

...

Jen noticed a missing space while qa'ing.
add br tag to break 
Screen Shot 2020-07-01 at 2 29 05 PM

...

### Relevant tickets

References [173152297](https://www.pivotaltracker.com/n/projects/2401401/stories/173152297).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Badges/BadgesTab.js
+++ b/resources/assets/components/pages/AccountPage/Badges/BadgesTab.js
@@ -56,6 +56,7 @@ const badgeModalContent = {
     unearnedText: (
       <span>
         Unlock this badge by
+        <br />
         {exploreCampaignsLink('signing up for a campaign')}.
       </span>
     ),
@@ -79,6 +80,7 @@ const badgeModalContent = {
     unearnedText: (
       <span>
         Want to earn your second Action badge?
+        <br />
         {exploreCampaignsLink('Rock another DoSomething campaign')}.
       </span>
     ),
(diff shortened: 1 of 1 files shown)


resources/assets/components/pages/AccountPage/Badges/BadgesTab.js
```diff
@@ -56,6 +56,7 @@ const badgeModalContent = {
     unearnedText: (
       <span>
         Unlock this badge by
+        <br />
         {exploreCampaignsLink('signing up for a campaign')}.
```
mendelB, line 60: I wonder if we can achieve a space without a line break by adding a literal space string {' '}?. (Something I've seen Prettier formatter do for instances [like this,](https://github.com/DoSomething/phoenix-next/blob/4c20abeab4f7e00177ae60d07b54763892a6b966/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js#L169-L173) but I suppose here they couldn't catch it).
```suggestion
        Unlock this badge by{' '}{exploreCampaignsLink('signing up for a campaign')}.
```

Dinnall: That's an interesting approach. I just added a new commit to reflect this.
I would have prefered to turn the entire `unearnedText` into a template literal, but that wouldn't work here

mendelB: Nice! And confirming that it's showing up correctly for both badges now? 

The reason I think this is neat is that it gives us a space as opposed to a line break which might not be applicable across all viewport sizes. Another option I've used is to add an HTML space code to ensure the space is rendered `&nbsp;`

Dinnall: Super odd, `{" "}` worked but `&nbsp;` looked better, so I substituted it.

PR #2252, opened 2020-07-02
add PhotoSubmissionBlock to contenful entry

Dinnall:
### What's this PR do?

This pull request Widen grid for photo uploader on story pages.

👍 
...

...

### Relevant tickets

References [Pivotal # 173152297](https://www.pivotaltracker.com/n/projects/2401401/stories/173152297).

### Checklist

Screen Shot 2020-07-02 at 12 01 21 PM

### DIFF
--- a/resources/assets/components/pages/StoryPage/StoryPage.js
+++ b/resources/assets/components/pages/StoryPage/StoryPage.js
@@ -6,7 +6,11 @@ import PropTypes from 'prop-types';
 import ContentfulEntry from '../../ContentfulEntry';
 import PageInfoBar from '../../PageInfoBar/PageInfoBar';
 import SiteFooter from '../../utilities/SiteFooter/SiteFooter';
-import { contentfulImageUrl, withoutNulls } from '../../../helpers';
+import {
+  contentfulImageUrl,
+  withoutNulls,
+  isAuthenticated,
+} from '../../../helpers';
 import SocialShareTray from '../../utilities/SocialShareTray/SocialShareTray';
 import SiteNavigationContainer from '../../SiteNavigation/SiteNavigationContainer';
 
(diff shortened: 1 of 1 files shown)


mendelB:
@Dinnall  One quick question -- how does this look for unauthenticated users where we display the auth button. I have a feeling that might not work well in the wide grid 🤔  (We may have to apply this custom class conditionally if the user is autheticated)

Dinnall:
> @Dinnall One quick question -- how does this look for unauthenticated users where we display the auth button. I have a feeling that might not work well in the wide grid 🤔 (We may have to apply this custom class conditionally if the user is authenticated)

Currently, the uploader isn't visible unless you're an authenticated user, if you click the button to add a photo it redirects to the login page

mendelB:
> > @Dinnall One quick question -- how does this look for unauthenticated users where we display the auth button. I have a feeling that might not work well in the wide grid 🤔 (We may have to apply this custom class conditionally if the user is authenticated)
> 
> Currently, the uploader isn't visible unless you're an authenticated user, if you click the button to add a photo it redirects to the login page

Ah so the button itself doesn't stretch across the wide grid?

Dinnall:
> > > @Dinnall One quick question -- how does this look for unauthenticated users where we display the auth button. I have a feeling that might not work well in the wide grid 🤔 (We may have to apply this custom class conditionally if the user is authenticated)
> > 
> > 
> > Currently, the uploader isn't visible unless you're an authenticated user, if you click the button to add a photo it redirects to the login page
> 
> Ah so the button itself doesn't stretch across the wide grid?

There is a button, but the uploader itself isn't visible. Here is an image of the current logout state 
Screen Shot 2020-07-02 at 12 40 36 PM

mendelB:
Ahh yeah ok I don't think we'd want that stretched out on the wide grid. I think what we can do is import the [`isAuthenticated` helper method](https://github.com/DoSomething/phoenix-next/blob/28ff4ce032b993c8b5849295c14e54b99e7f1a2a/resources/assets/helpers/auth.js#L76-L81), and only attach the `grid-wide` _if_ they're authenticated:
`PhotoSubmissionBlock: isAuthenticated() ? 'grid-wide' : 'grid-narrow'`

Dinnall:
> isAuthenticated() ? 'grid-wide' : 'grid-narrow'

This definitely looks better. Added another commit to reflect this change
Screen Shot 2020-07-02 at 1 07 01 PM

PR #2246, opened 2020-07-01
Display group voter registrations on OVRD page if group_id query exists

aaronschachter:
### What's this PR do?

This pull request displays the `VoterRegistrationReferralsBlock` Group template instead of the `CampaignInfoBlock` in the `VoterRegistrationDrivePageBanner` if a `group_id` query parameter exists (and the group is valid). It tweaks the Group template to support passing a `user` parameter, which corresponds to the required `referrer_user_id` query parameter used on a OVRD page.

Group OVRD page

### How should this be reviewed?

Once the review app builds, I'll add example URL's of an OVRD page with and without a `group_id` query to verify expected behavior.

### Any background context you want to provide?

After this is merged, I'll cleanup the TODOs in the Cypress tests that have been waiting for https://github.com/DoSomething/graphql/pull/252 to properly mock separate counts for individual and group voter registration counts. 

### Relevant tickets

References [Pivotal #173044099](https://www.pivotaltracker.com/story/show/173044099).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [x] Added appropriate feature/unit tests.

PR #2253, opened 2020-07-02
Fix styles for OVRD group registration stats

aaronschachter:
### What's this PR do?

This pull request branches from #2250 to add some of the style changes needed in the `VoterRegistrationReferralsBlock` Group template when displaying it on an OVRD page with a `group_id` (added in #2246) per [design review](https://dosomething.slack.com/archives/CTVPG6L4R/p1593635632146500?thread_ts=1593634325.143200&cid=CTVPG6L4R). This PR adds support for slight differences that the OVRD page requires:

* The percentage label should be larger text, and black, not gray
* The stat number should appear to the left of each label

It also removes the white border around the progress bar that is not in design, per [Slack request](https://dosomething.slack.com/archives/CTVPG6L4R/p1593635816147200?thread_ts=1593634325.143200&cid=CTVPG6L4R).

**OVRD page**

OVRD group stats

**Action Page** (no visual changes made)

Action page group stats

👀 

* This PR removes the percentage label "You're 20% to your goal" out of the `ProgressBar` component and into the `Group` template of the `VoterRegistrationReferralsBlock`, so it'll be easier to reuse the `ProgressBar`in other places should we ever need to

* The one style change I struggled with and held on for now is vertically centering the stat label with the stat number on the OVRD page view. I tried using Tailwind's [`align-middle`](https://tailwindcss.com/docs/vertical-align/) class but it didn't do the trick. Held off on this for now to get the majority of needed changes in

* This is blocked from merge until #2250 is merged, and I'll need to remember to edit this PR to merge into master once that happens before this is approved/merged.

### Relevant tickets

References [Pivotal #173044099](https://www.pivotaltracker.com/n/projects/2417735/stories/173044099).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
@@ -29,7 +29,10 @@ const StatBlock = ({ amount, isVertical, label, testId }) => {
   );
 
   return (
-    <div className={`pt-3 ${isVertical ? 'flex' : null}`} data-testid={testId}>
+    <div
+      className={`pt-3 ${isVertical ? 'pb-3 flex' : null}`}
+      data-testid={testId}
+    >
       {isVertical ? null : statLabel}
       <h1
         className={`font-normal font-league-gothic text-3xl ${
@@ -80,7 +83,11 @@ const GroupTemplate = ({ group, user }) => {
           return (
             <>
               <div data-testid="group-progress" className="py-3">
-                <span className="font-bold uppercase text-gray-600">
+                <span
+                  className={`font-bold uppercase ${
+                    user ? 'text-lg' : 'text-gray-600'
+                  }`}
+                >
                   {percentage > 100
                     ? `🎉 You're at ${percentage}% of your goal! 🎉`
                     : `${percentage}% to your goal!`}
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
@@ -17,11 +17,11 @@ const GROUP_VOTER_REGISTRATION_REFERRALS_QUERY = gql`
   }
 `;
 
-const StatBlock = ({ amount, label, testId, condensed }) => {
+const StatBlock = ({ amount, isVertical, label, testId }) => {
   const statLabel = (
     <div
       className={`font-bold uppercase text-gray-600 ${
-        condensed ? 'w-5/6 pb-2' : ''
+        isVertical ? 'w-5/6 pb-2' : ''
       }`}
     >
       {label}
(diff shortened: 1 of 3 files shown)


mendelB:
> This is blocked from merge until #2250 is merged, and I'll need to remember to edit this PR to merge into master once that happens before this is approved/merged.

I _think_ this happens automatically btw once the current destination branch is merged!

aaronschachter:
Ah, you're right it does... and I'm trying to fix the merge conflict 😿

mendelB:
Oh lordy. Godspeed! 🙌

aaronschachter:
Merge conflict resolved, ready for 👀 when you have a chance @mendelB

resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
```diff
@@ -17,15 +17,40 @@ const GROUP_VOTER_REGISTRATION_REFERRALS_QUERY = gql`
...
+      {label}
+    </span>
+  );
+
+  return (
+    <div
+      className={`pt-3 ${isVertical ? 'pb-3 flex' : null}`}
+      data-testid={testId}
+    >
+      {isVertical ? null : statLabel}
```
mendelB, line 36: 🤔 Could we use [`flex-direction: row-reverse`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction) for `isVertical` (or the opposite), so that we don't need to render this twice?

resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
```diff
@@ -55,17 +80,27 @@ const GroupTemplate = ({ group, user }) => {
...
             <>
-              <ProgressBar
-                completed={groupTotal}
-                target={groupGoal}
-                testId="group-progress"
-              />
+              <div data-testid="group-progress" className="py-3">
+                <span
+                  className={`font-bold uppercase ${
+                    user ? 'text-lg' : 'text-gray-600'
```
mendelB, line 90: Small fry thought -- maybe for the future as we further formalize this -- to perhaps use a more explicit prop to govern which display we use, instead of the implicit derivation from having an authenticated user 🤔

aaronschachter: Yeah, I think that'd be helpful. Maybe just use the same`isVertical` name for a boolean prop here for now?

mendelB: Ah totally -- that makes sense to me!

resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
aaronschachter: Great suggestion! 🔥 Updated in 25dcf16209840702ac0fea83c6973c8439e05e30.

resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
```diff
@@ -17,23 +17,40 @@ const GROUP_VOTER_REGISTRATION_REFERRALS_QUERY = gql`
...
+      {label}
+    </span>
+
+    <h1
+      className={`font-normal font-league-gothic text-3xl ${
+        isVertical ? ' w-1/6' : ''
+      }`}
+    >
+      {amount}
+    </h1>
```
weerd, line 39: So just to help clarify here for future reference, I suspect that `align-middle` gave you trouble because it only works on [inline level items](https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align). Here you're using one inline level item by default (the `<span>`) and one block level item by default (the `<h1>`) so the rule would never really apply here.

Side note, the use of an `<h1>` (the highest level heading possible) here seems weird and unnecessary. This component already probably will sit within another component with a high level heading (likely an `<h2>`) but then this one which is a label for a value is higher than that. Maybe a future refactor could be to use a definition list or maybe just switch this to another span since it's not really meant to act as a heading (it just visually in the design _looks_ like a heading cause its bold and uppercase, etc). Switching it to span would also likely make it so `align-middle` would then work.

PR #2259, opened 2020-07-07
Fix typo in RAF soft launch copy

mendelB:
### What's this PR do?

In another exciting edition of Mendel's Copy Adventures, this pull request fixes a typo in a link on the Beta Page flagged by Diego D. EDIT: And one more typo.

(It's not lost on me that this is another prime example of the benefits of GUI or config editable copy or some such convenience which we've been discussing. But for another day. This one's on me.)


mendelB:
image

PR #2264, opened 2020-07-09
OVRD Goals Styling Updates

lindsayrmaher:
### What's this PR do?

This pull request adds some additional styling updates to the goals section of the beta OVRD pages based on feedback.

👀 

Current:
Screen Shot 2020-07-09 at 2 56 16 PM

New: 
Screen Shot 2020-07-09 at 2 55 04 PM

https://app.abstract.com/share/93b12c91-10b4-4144-b551-e0322be9a51c?collectionId=ca9d4751-035f-4a87-bd85-8e301e5a0734&collectionLayerId=13b7fb5b-5e55-4aa7-9212-e918ee130a3c&mode=build&sha=f6c7ebddf2128d005dca1d84e48cdc9cb3abd987

### Relevant tickets

References [Pivotal # 173044099](https://www.pivotaltracker.com/story/show/173044099).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
@@ -32,13 +32,13 @@ const StatBlock = ({ amount, isVertical, label, testId }) => (
       {label}
     </span>
 
-    <span
+    <h2
       className={`font-normal font-league-gothic text-3xl ${
         isVertical ? ' w-1/6 mb-0' : ''
       }`}
     >
       {amount}
-    </span>
+    </h2>
   </div>
 );
 
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
@@ -24,7 +24,7 @@ const StatBlock = ({ amount, isVertical, label, testId }) => (
   >
     <span
       className={`font-bold uppercase text-gray-600 ${
-        isVertical ? 'w-5/6 pb-2' : ''
+        isVertical ? 'w-5/6' : ''
       }`}
     >
       {label}
@@ -54,7 +54,10 @@ const GroupTemplate = ({ group, isVertical, user }) => {
   const groupDescription = `${group.groupType.name}: ${group.name}`;
 
   return (
-    <div data-testid="group-voter-registration-referrals-block">
+    <div
+      data-testid="group-voter-registration-referrals-block"
+      className="mx-3"
+    >
       {user ? null : (
         <>
           <SectionHeader title={groupDescription} />
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
@@ -19,7 +19,9 @@ const GROUP_VOTER_REGISTRATION_REFERRALS_QUERY = gql`
 
 const StatBlock = ({ amount, isVertical, label, testId }) => (
   <div
-    className={`pt-3 ${isVertical ? 'pb-3 flex flex-row-reverse' : null}`}
+    className={`pt-3 ${
+      isVertical ? 'pb-3 flex flex-row-reverse items-center' : null
+    }`}
     data-testid={testId}
   >
     <span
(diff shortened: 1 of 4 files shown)


resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
```diff
@@ -19,20 +19,22 @@ const GROUP_VOTER_REGISTRATION_REFERRALS_QUERY = gql`
 
 const StatBlock = ({ amount, isVertical, label, testId }) => (
   <div
-    className={`pt-3 ${isVertical ? 'pb-3 flex flex-row-reverse' : null}`}
+    className={`pt-3 ${
+      isVertical ? 'pb-3 flex flex-row-reverse items-center' : null
```
weerd, line 23: What's the `isVertical` prop?

aaronschachter: I added it in #2253. It's set to `true` when we're rendering this component instead of a `CampaignInfoBlock` on an OVRD page. 
It's default `false` view is used when we're rendering the component in a Page `blocks` (#2253 has screenshots of each)

weerd: Ok, to make sure I understand it's for this visual alignment:

image

versus

image

weerd: Or am I referring to the wrong thing? Just trying to get the full context here.

aaronschachter: It's used for aligning the number to the left of the label -- but also to render percentage label in be larger black text vs gray (mentioned in #2253).

Does sound like a design question -- initially they were the same in https://github.com/DoSomething/phoenix-next/pull/2246 because I missed that they were different, and we received feedback to modify per design (comments in the Pivotal story)

weerd: Ya, for future reference I think this is something the tech team should push back on based on the additional complexity for little payoff and I think it just looks better all as one line. I would understand if this was specifically for mobile, but then it would be the case that it applies to both types of pages and thus would just use responsive classes to deal with the change in layout.

weerd: @aaronschachter I also left a comment on your PR regarding the `align-middle` and a suggestion _not_ to use `<h1>` in this which seems unnecessary and would likely fix your `align-middle` woes 😄

resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
```diff
@@ -19,20 +19,22 @@ const GROUP_VOTER_REGISTRATION_REFERRALS_QUERY = gql`
...
     <span
       className={`font-bold uppercase text-gray-600 ${
-        isVertical ? 'w-5/6 pb-2' : ''
+        isVertical ? 'w-5/6' : ''
       }`}
     >
       {label}
     </span>
 
     <h1
```
weerd, line 35: Suggestion, but I'd recommend not using an `<h1>` here. Maybe a lower level heading or just another span.

aaronschachter: Not sure from exactly which component, but I know I copied the `h1` by example from some other components that have display the similar large number. I see that this was updated in cf7e89d -- thanks for making the change  @lindsayrmaher !

resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Group.js
lindsayrmaher: i updated to a `span` tag on the number for consistencies sake, but i still can't get them to align properly. @aaronschachter unless you're worried about this, for the sake of time i'm going to merge this in.

aaronschachter: I'm not worried! Feel free to ✅

aaronschachter: Although I think https://github.com/DoSomething/phoenix-next/pull/2264/commits/cf7e89ddf6a0f4da773deee87fd75cb3e2d95bf6 breaks tests...

PR #2267, opened 2020-07-10
Modify individual VoterRegistrationReferralsBlock to retrieve all referrals

aaronschachter:
### What's this PR do?

This pull request modifies the GraphQL query in the individual template of the `VoterRegistrationReferralsBlock` to retrieve all referrals, regardless of status -- instead of only completed referrals. 

It doesn't make any corresponding UI changes -- the component still only displays completed referrals. This work is in preparation for displaying the total list of an alpha's registrations, including those who haven't completed, per [Pivotal #173230283](https://www.pivotaltracker.com/story/show/173230283).

👀 

🥍 

### Relevant tickets

References [Pivotal #173230283](https://www.pivotaltracker.com/story/show/173230283).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Individual.js
+++ b/resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Individual.js
@@ -30,11 +30,7 @@ const INDIVIDUAL_VOTER_REGISTRATION_REFERRALS_QUERY = gql`
 const parseVoterRegistrationReferrals = voterRegPosts => {
   const result = { complete: {}, incomplete: {} };
 
-  /**
-   * @param {String} status
-   * @param {Object} user
-   */
-  const parseVoterRegistrationReferral = ({ status, user }) => {
+  voterRegPosts.forEach(({ status, user }) => {
     const userId = user.id;
 
     // If status is incomplete:
@@ -55,9 +51,7 @@ const parseVoterRegistrationReferrals = voterRegPosts => {
     }
 
     result.complete[userId] = user;
-  };
-
-  voterRegPosts.forEach(parseVoterRegistrationReferral);
+  });
 
   return result;
 };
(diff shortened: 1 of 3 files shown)


resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Individual.js
```diff
@@ -12,19 +12,56 @@ import ReferralsGallery from '../../../utilities/ReferralsGallery/ReferralsGalle
...
+ * @return {Object}
+ */
+const parseVoterRegistrationReferrals = voterRegPosts => {
+  const result = { complete: {}, incomplete: {} };
+
+  /**
+   * @param {String} status
+   * @param {Object} user
+   */
+  const parseVoterRegistrationReferral = ({ status, user }) => {
```
DFurnes, line 37: This is subjective, but the nested function declaration kinda throws me for a loop. I think I'd prefer if we pulled this out (if we think it'd be useful on it's own, and should be in the module scope) or inline it into the `forEach`:

```
voterRegPosts.forEach(function parseVoterRegistrationReferral({ status, user }) {
  // ...
});
```

resources/assets/components/blocks/VoterRegistrationReferralsBlock/templates/Individual.js
```diff
@@ -14,17 +14,53 @@ const INDIVIDUAL_VOTER_REGISTRATION_REFERRALS_QUERY = gql`
   query IndividualVoterRegistrationReferralsQuery($referrerUserId: String!) {
     posts(
       referrerUserId: $referrerUserId
+      status: [REGISTER_FORM, REGISTER_OVR, STEP_1, STEP_2, STEP_3, STEP_4]
```
aaronschachter, line 17: I've added this check to avoid retrieving posts we wouldn't want to appear in the list at all (when the status is either `REJECTED` OR `UNDER-18`). 

Eventually we may need to switch this to a paginated query to display the full list of referrals. Because we want completed referrals to show up first, it'd likely be best to split this into two different GraphQL queries, one for complete `REGISTER-*` statuses and the other for in-progress  `STEP-*` statuses.

PR #2268, opened 2020-07-13
Remove Page Sidebar Field

mendelB:
### What's this PR do?

This pull request branches off of #2266 to remove references to the deprecated Page#sidebar field. It also adds the result of removing the field from Contentful Page and generating a new migration script. (Also removes two legacy Landing Page fields from the PHP Entity).

👀 

We've deprecated any usage of this field in #2266 but broadly haven't been using this field for a hot minute. We'll also be removing the field from GraphQL in an accompanying PR.

### Relevant tickets

References [Pivotal #173738645](https://www.pivotaltracker.com/story/show/173738645). (Also deprecates two legacy Landing Page fields from the PHP Entity).

### Checklist


app/Entities/LandingPage.php
```diff
@@ -19,8 +19,6 @@ public function jsonSerialize()
             'fields' => [
                 'internalTitle' => $this->internalTitle,
                 'content' => $this->content,
-                'sidebar' => $this->parseBlocks($this->sidebar),
-                'blocks' => $this->parseBlocks($this->blocks),
```
katiecrane, line 23: Why is `blocks` removed here as well?

cypress/fixtures/contentful/exampleVoterRegistrationDriveCampaign.js
```diff
@@ -122,8 +120,6 @@ export default {
       fields: {
         internalTitle: 'Online Registration Drive Landing Page',
         content: null,
-        sidebar: [],
-        blocks: [],
```
katiecrane, line 126: Same Q here about blocks being removed!

contentful/content-types/page.js
```diff
@@ -267,22 +246,14 @@ module.exports = function(migration) {
   });
 
   page.changeFieldControl('metadata', 'builtin', 'entryLinkEditor', {});
-
-  page.changeFieldControl('authors', 'builtin', 'entryLinksEditor', {
-    bulkEditing: false,
-  });
+  page.changeFieldControl('authors', 'builtin', 'entryLinksEditor', {});
```
katiecrane, line 249: Is the update here an old one that got rolled in?

app/Entities/LandingPage.php
mendelB: These fields were removed from the Landing Page content type in #1931 but I think we just forgot to remove them from the Entity here.

contentful/content-types/page.js
mendelB: Yeah 🤷 I assume one of those formatting things that seem to get rolled in as we update these older scripts

PR #2269, opened 2020-07-13
Added Cypress Test for Scholarship Modal - Authenticated User Already Signed up via Partner 

Vmack:
### What's this PR do?

This pull request adds a cypress test to the scholarship modal to test the flow for a user coming from a scholarship partner website 

Is this the correct flow for the user coming from a scholarship partner?
Is there a more simple way to do this?

### Relevant tickets

References [Pivotal #172546539](https://www.pivotaltracker.com/story/show/172546539).

### Checklist

### DIFF
--- a/cypress/integration/campaign-signup.js
+++ b/cypress/integration/campaign-signup.js
@@ -212,6 +212,25 @@ describe('Campaign Signup', () => {
     );
   });
 
+  /** @test */
+  it('Visits a campaign page from a scholarship partner, as an authenticated user', () => {
+    const user = userFactory();
+
+    //Log in and visit the campaign pitch page
+    cy.authVisitCampaignWithSignup(user, exampleCampaign).visit(
+      '/us/campaigns/test-example-campaign?utm_campaign=fastweb&utm_source=scholarship',
+
+      //Find the example campaign
+      cy.contains('Example Campaign'),
+      cy.contains('This is an example campaign for automated testing.'),
+
+      // We shouldn't see the "Join Now" button or affirmation modal,
+      // since the user is already signed up for this campaign:
+      cy.findByTestId('campaign-banner-signup-button').should('not.exist'),
+      cy.get('.card.affirmation').should('not.exist'),
+    );
+  });
+
   // TODO: Use cypress context to better group this test once #2238 is merged.
   /** @test */
   it('If campaign group type does not filter by state, signup button is enabled after selecting group', () => {
--- a/cypress/integration/campaign-signup.js
+++ b/cypress/integration/campaign-signup.js
@@ -217,18 +217,20 @@ describe('Campaign Signup', () => {
     const user = userFactory();
 
     //Log in and visit the campaign pitch page
-    cy.authVisitCampaignWithSignup(user, exampleCampaign).visit(
+    cy.authVisitCampaignWithSignup(user, exampleCampaign);
+
+    // Visit the campaign pitch page
+    cy.withState(exampleCampaign).visit(
       '/us/campaigns/test-example-campaign?utm_campaign=fastweb&utm_source=scholarship',
+    );
 
-      //Find the example campaign
-      cy.contains('Example Campaign'),
+    //Find the example campaign
+    cy.contains('Example Campaign'),
       cy.contains('This is an example campaign for automated testing.'),
-
       // We shouldn't see the "Join Now" button or affirmation modal,
       // since the user is already signed up for this campaign:
       cy.findByTestId('campaign-banner-signup-button').should('not.exist'),
-      cy.get('.card.affirmation').should('not.exist'),
-    );
+      cy.get('.card.affirmation').should('not.exist');
   });
 
   // TODO: Use cypress context to better group this test once #2238 is merged.
--- a/cypress/integration/campaign-signup.js
+++ b/cypress/integration/campaign-signup.js
@@ -217,20 +217,21 @@ describe('Campaign Signup', () => {
     const user = userFactory();
 
     //Log in and visit the campaign pitch page
-    cy.authVisitCampaignWithSignup(user, exampleCampaign);
-
-    // Visit the campaign pitch page
-    cy.withState(exampleCampaign).visit(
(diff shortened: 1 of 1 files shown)


cypress/integration/campaign-signup.js
```diff
@@ -212,6 +212,27 @@ describe('Campaign Signup', () => {
     );
   });
 
+  /** @test */
+  it('Visits a campaign page from a scholarship partner, as an authenticated user', () => {
+    const user = userFactory();
+
+    //Log in and visit the campaign pitch page
+    cy.authVisitCampaignWithSignup(user, exampleCampaign);
```
katiecrane, line 220: oooo you know what, I don't think we can use this helper because we need to visit the campaign with those utms. But we CAN look at what the helper does behind the scenes and modify it so that we can use it. The [authVisitCampaignWithSignup](https://github.com/DoSomething/phoenix-next/blob/9eb5183fd72d8f8bc4f0d0a574d912dd404748a6/cypress/support/commands.js#L186-L199) helper does this:
```
  cy.login(user)
    .withState(contentfulCampaign)
    .withSignup(contentfulCampaign.campaign.campaignId)
    .visit(`${campaignPath}${contentfulCampaign.campaign.slug}`);
```
so I think we can replace this line with that block, but make a few changes so it's doing exactly what we want.
- The login part is good since we're using `user` as well
- Each time `contentfulCampaign` appears, we'll want to instead use `exampleCampaign`
- We'll also sub in the path you have down below, so `visit` will take `'/us/campaigns/test-example-campaign?utm_campaign=fastweb&utm_source=scholarship'` as the argument instead of `${campaignPath}${contentfulCampaign.campaign.slug}`

Does that make sense?

cypress/integration/campaign-signup.js
```diff
@@ -212,6 +212,27 @@ describe('Campaign Signup', () => {
...
+
+    // Visit the campaign pitch page
+    cy.withState(exampleCampaign).visit(
+      '/us/campaigns/test-example-campaign?utm_campaign=fastweb&utm_source=scholarship',
+    );
```
katiecrane, line 225: Since we'll be visiting the campaign page in the above block, this will be able to be removed

cypress/integration/campaign-signup.js
```diff
@@ -212,6 +212,27 @@ describe('Campaign Signup', () => {
...
+
+    //Find the example campaign
+    cy.contains('Example Campaign'),
+      cy.contains('This is an example campaign for automated testing.'),
+      // We shouldn't see the "Join Now" button or affirmation modal,
+      // since the user is already signed up for this campaign:
+      cy.findByTestId('campaign-banner-signup-button').should('not.exist'),
+      cy.get('.card.affirmation').should('not.exist');
```
katiecrane, line 233: We want to make sure that we test specifically for the "Apply Now" button not showing up. The `campaign-banner-signup-button` is the one on the landing page, so we don't need to worry about that here, or the affirmation bit. We want to basically do the opposite of this assertion from the test above:
```
    cy.findByTestId('campaign-info-block-scholarship-details').contains(
      'button',
      'Apply Now',
    );
```
but I'm actually not sure how you do "does not contain" in Cypress and couldn't find an example of it in their docs or within our tests. @DFurnes is there a way to do "does not contain" via Cypress?

DFurnes: I think you could write this as:

```js
cy.findByTestId('campaign-info-block-scholarship-details').should('not.contain', 'Apply Now')
```

Vmack: so `.should('not.exist')` won't work?

cypress/integration/campaign-signup.js
Vmack: that makes great sense, thank you!

cypress/integration/campaign-signup.js
katiecrane: I think that won't work because I believe you have to call `.should('not.exist')` on an element itself [like in this example](https://docs.cypress.io/guides/references/assertions.html#Existence). Here we're concerned with the content of `'campaign-info-block-scholarship-details'`, and we want to make sure that content does not contain "Apply Now" so I think that's why we want to use `.should('not.contain', 'Apply Now')`

cypress/integration/campaign-signup.js
```diff
@@ -212,6 +212,28 @@ describe('Campaign Signup', () => {
...
+    cy.login(user)
+      .withState(exampleCampaign)
+      .withSignup(exampleCampaign.campaign.campaignId)
+      .visit(
+        '/us/campaigns/test-example-campaign?utm_campaign=fastweb&utm_source=scholarship',
+      );
+
+    //Find the example campaign
+    cy.contains('Example Campaign'),
+      cy.contains('This is an example campaign for automated testing.'),
```
katiecrane, line 229: Let's make this line up with the line above it (goes for the lines below it too)

cypress/integration/campaign-signup.js
```diff
@@ -212,6 +212,28 @@ describe('Campaign Signup', () => {
...
+      // We shouldn't see the "Join Now" button or affirmation modal,
```
katiecrane, line 230: Let's change this to say "Apply Now" button and remove the part about the affirmation modal

cypress/integration/campaign-signup.js
```diff
@@ -212,6 +212,28 @@ describe('Campaign Signup', () => {
...
+      // since the user is already signed up for this campaign:
+      cy
+        .findByTestId('campaign-info-block-scholarship-details')
```
katiecrane, line 233: This can go on the previous line so we end up with `cy.findByTestId('campaign-info-block-scholarship-details')` on the first line here

Vmack: prettier keeps reverting the changes 😬

katiecrane: oh welp! I guess we're fine then?

Vmack: I guess...it's so odd though

PR #1919, opened 2020-02-10
Fix content widths on general pages.

DFurnes:
### What's this PR do?

This pull request fixes an issue [flagged by @mendelB in Slack](https://dosomething.slack.com/archives/C3ASB4204/p1581355562011000): now that Content Blocks handle their own layout, they're appearing too narrow on Article/Fact pages:

Screen Shot 2020-02-10 at 1 24 01 PM

As a quick fix, I've added a class that lets us conditionally "force" these full-width on articles:

Screen Shot 2020-02-10 at 1 24 16 PM

### How should this be reviewed?

🙈 

### Any background context you want to provide?

This is definitely not a good long-term solution, but it should unblock today's deploy!

### Relevant tickets

References [Pivotal #171190791](https://www.pivotaltracker.com/story/show/171190791).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2272, opened 2020-07-13
Content Block Content Span Update / propsByEntry option for ContentfulEntryLoader

mendelB:
### What's this PR do?

This pull request:
- adds support for a new `propsByEntry` option to allow assigning custom props per Content Type when using the `ContentfulEntryLoader` (similar to the `classNameByEntry` option).
- Uses this setting to toggle the Content Block's new `fullWidthContent` on General Pages and Modal routes so that we can avoid the explicit `scss` styling.

- First of all, does the `propsByEntry` make sense in principle, generally speaking? I feel like this idea has come up before. Conceptually, it feels very alike to the `classNameByEntry` and could allow for some neat flexibility & customization from the top level.
- Secondly, how do we feel about this refactored version for the Content Block's content width HACK? The obvious advantage is that we no longer rely on CSS rules and instead support a formal toggle, but this subjectively complicates things a bit.
- Finally, how's the naming? (for the dear old life of me I couldn't think of anything better than `fullWidthContent` but I'm slightly disappointed with that effort). Is the documentation clear?

This work broadly came up via #2266 

See #1919 for more context on the background for this hack. I suppose the danger of formalizing this here is that we're committing more formally to supporting Content Blocks on these Modal routes and Articles where we'd prefer to start using inline Rich text for content with no accompanying image (See [thread](https://dosomething.slack.com/archives/C3ASB4204/p1581355562011000) from #1919), but we're a ways away from the perfect Rich Text world so perhaps this will do for now?

### Relevant tickets

References [Pivotal #173738645](https://www.pivotaltracker.com/story/show/173738645).

### Checklist


resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
```diff
@@ -173,7 +174,10 @@ const ContentfulEntryLoader = ({
       data-contentful-id={id}
       className={classnames(className, entryClassNames)}
     >
-      <ContentfulEntry json={data.block} />
+      <ContentfulEntry
+        json={data.block}
+        propsByEntry={get(propsByEntry, blockType)}
```
DFurnes, line 179: I like `propsByEntry` for the `ContentfulEntryLoader`, but it feels a little awkward to continue to name it "by entry" when passing it along to the `ContentfulEntry`... since now it's just the props!

Would `props` (woah, maybe that's weird) or `customProps` work for this?

DFurnes: We could consider also just taking all the props that aren't `json` provided to `<ContentfulEntry {...props} />` and passing them along to the child (which feels kinda elegant, if perhaps a bit more brittle).

So here we'd just say:

```jsx
<ContentfulEntry json={data.block} {...get(propsByEntry, blockType)} />
```

And then in our component we could pull `const ContentfulEntry({ json, ...props }) => { … }`

DFurnes:
> First of all, does the `propsByEntry` make sense in principle, generally speaking? I feel like this idea has come up before. Conceptually, it feels very alike to the `classNameByEntry` and could allow for some neat flexibility & customization from the top level.

👍 👍 

> Secondly, how do we feel about this refactored version for the Content Block's content width HACK? The obvious advantage is that we no longer rely on CSS rules and instead support a formal toggle, but this subjectively complicates things a bit.

Yes!! I think this is a much more robust solution for this problem. I'm always in favor of a more verbose, but explicit approach.

> How's the naming? (for the dear old life of me I couldn't think of anything better than `fullWidthContent` but I'm slightly disappointed with that effort).

Naming! I'm okay with it. We could maybe shorten to `fullWidth` if you think it's a bit long-winded?

mendelB:
Thanks for this awesome feedback here @DFurnes! (And for parsing through this very "extra" PR description 🙃 ) These ideas are fire, and will work in an update!

resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js
mendelB: I incorporated these updates in #2277! Hopefully faithfully.

PR #2273, opened 2020-07-14
add test for scholarship modal "Apply Now"

Dinnall:
### What's this PR do?

This pull request test authenticated user who is signed up for the campaign modal & test if the "Apply Now" button is. visible.

...

...
### Relevant tickets

References [Pivotal 172561855](https://www.pivotaltracker.com/n/projects/2401401/stories/172561855).

### Checklist

### DIFF
--- a/cypress/integration/campaign-info-block.js
+++ b/cypress/integration/campaign-info-block.js
@@ -142,7 +142,7 @@ describe('Campaign Info Block', () => {
   });
 
   /** @test */
-  it('Visit campaign, as an authenticated user, show scholarship modal without "Apply Now" ', () => {
+  it('Visit campaign, as an affiliated user, show scholarship modal without "Apply Now" ', () => {
     const user = userFactory();
 
     // Auth user visit a campaign page:
--- a/cypress/integration/campaign-info-block.js
+++ b/cypress/integration/campaign-info-block.js
@@ -141,6 +141,20 @@ describe('Campaign Info Block', () => {
   });
 
   /** @test */
+  it('Visit campaign, as an authenticated user, show scholarship modal without "Apply Now" ', () => {
+    const user = userFactory();
+
+    // Auth user visit a campaign page:
+    cy.authVisitCampaignWithSignup(user, exampleCampaign);
+
+    cy.findByTestId('campaign-info-block-container').within(() => {
+      cy.contains('button', 'View Scholarship Details').click();
+    });
+
+    cy.findByTestId('campaign-banner-signup-button').should('not.exist');
+  });
+
+  /** @test */
   it('Opens the scholarship modal when "View Scholarship Details" is clicked, and closes it when the X button is clicked.', () => {
     cy.anonVisitCampaign(exampleCampaign);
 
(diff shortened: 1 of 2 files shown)


cypress/integration/campaign-info-block.js
```diff
@@ -140,6 +141,20 @@ describe('Campaign Info Block', () => {
...
+    const user = userFactory();
+
+    // Auth user visit a campaign page:
+    cy.authVisitCampaignWithSignup(user, exampleCampaign);
+
+    cy.findByTestId('campaign-info-block-container').within(() => {
+      cy.contains('button', 'View Scholarship Details').click();
+    });
+
+    cy.findByTestId('campaign-banner-signup-button').should('not.exist');
```
mendelB, line 155: I think we want to test for [this div & enclosed signup button](https://github.com/DoSomething/phoenix-next/blob/6272d589e59a7f2049312d418731f1ece02032fc/resources/assets/components/CampaignBanner/CampaignBanner.js#L178-L189) not showing up since it's the one particular to the scholarship modal that we're testing. 

It might make sense to add a `data-testid` there as we did on the other signup button to help us assert against it.

cypress/integration/campaign-info-block.js
```diff
@@ -140,6 +141,20 @@ describe('Campaign Info Block', () => {
     });
   });
 
+  /** @test */
+  it('Visit campaign, as an authenticated user, show scholarship modal without "Apply Now" ', () => {
```
mendelB, line 145: How's about:
```suggestion
  it('Visit campaign, as an affiliated user, show scholarship modal without "Apply Now" ', () => {
```
Since we're trying to explicitly test the signup button when they're affiliated?

resources/assets/components/CampaignBanner/CampaignBanner.js
```diff
@@ -175,7 +175,10 @@ const CampaignBanner = ({
             }
           >
             {!isAffiliated ? (
-              <div className={`pt-6 ${!groupType ? 'w-2/3 sm:w-1/2' : null}`}>
+              <div
+                data-testid="campaign-banner-affiliated"
```
mendelB, line 179: How's about something like 
```suggestion
                data-testid="scholarship-modal-signup-button"
```

To complement the `campaign-banner-signup-button` above?

Dinnall:
> Looks great! 👍 Just two quick naming related recommendations!

Like the name suggestions

cypress/integration/campaign-info-block.js
```diff
@@ -140,6 +141,20 @@ describe('Campaign Info Block', () => {
...
+    const user = userFactory();
+
+    // Auth user visit a campaign page:
+    cy.authVisitCampaignWithSignup(user, exampleCampaign);
+
+    cy.findByTestId('campaign-info-block-container').within(() => {
+      cy.contains('button', 'View Scholarship Details').click();
+    });
+
+    cy.findByTestId('campaign-banner-affiliated').should('not.exist');
```
mendelB, line 155: We just need to update this test per the new name and were golden!

mendelB: Oops @Dinnall we need to update this test to look for the `scholarship-modal-signup-button` since we've renamed it. (The `campaign-banner-affiliated` will never appear because it doesn't exist 😅 ). 

Totally fine for a follow-up PR! 

```suggestion
    cy.findByTestId('scholarship-modal-signup-button').should('not.exist');
```

PR #2277, opened 2020-07-15
Content Block Content Span Update / propsByEntry option for ContentfulEntryLoader

mendelB:
### What's this PR do?

This pull request is really this pull request https://github.com/DoSomething/phoenix-next/pull/2272


resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -92,13 +92,13 @@ class ContentfulEntry extends React.Component {
...
         return (
           <ContentBlock
-            className={className}
             {...withoutNulls({
               ...json,
               // Resolves the aliases used in the ContentBlockFragment.
               content: json.contentBlockContent,
               imageAlignment: json.contentBlockImageAlignment,
             })}
+            {...this.props}
```
DFurnes, line 101: This would also send the `ContentfulEntry` props to the child component (which isn't the worst thing, but could be confusing). What if we use a rest parameter when restructuring above so we only forward along any additional props that aren't `json`, `className`, `classNameByEntry`, or `classNameByEntryDefault`?

```js
    const {
      json = DEFAULT_BLOCK,
      className = null,
      classNameByEntry = {},
      classNameByEntryDefault = null,
      ...additionalCustomProps
    } = this.props;

    // ...

    <ContentBlock
       {...additionalCustomProps}
    />
```

resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -241,12 +241,14 @@ ContentfulEntry.propTypes = {
   className: PropTypes.string,
   classNameByEntry: PropTypes.object, // eslint-disable-line
   classNameByEntryDefault: PropTypes.string,
+  customProps: PropTypes.object,
```
DFurnes, line 244: It looks like this isn't used here, so I think we can remove! 👾

PR #2284, opened 2020-07-20
Remove Facebook share button on affirmation modal 

Dinnall:
### What's this PR do?

This pull request removes  the Facebook share button on affirmation modal 

...

...

### Relevant tickets

References [Pivotal # 172631182](https://www.pivotaltracker.com/n/projects/2401401/stories/172631182).

### Checklist

Screen Shot 2020-07-19 at 10 50 24 PM

### DIFF
--- a/resources/assets/components/Affirmation/Affirmation.js
+++ b/resources/assets/components/Affirmation/Affirmation.js
@@ -83,9 +83,6 @@ const Affirmation = ({
                 <h3>{callToActionHeader}</h3>
                 <p>{callToActionDescription}</p>
               </FlexCell>
-              <FlexCell className="p-3" width="half">
-                <Share variant="blue" parentSource="affirmation" />
-              </FlexCell>
             </Flex>
 
             {author ? (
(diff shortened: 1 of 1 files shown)


Dinnall:
> ```diff
>  </div>
> ```

when the

Dinnall:
> I wonder if we could also remove the surrounding `<Flex`> (replace it with like a `div`) and the other `<FlexCell>` to allow the CTA content to flow across the modal 🤔
> 
> ```diff
> -            <Flex className="items-center">
> +            <div>
> -              <FlexCell className="affirmation__cta p-3" width="half">
>                 <h3>{callToActionHeader}</h3>
>                 <p>{callToActionDescription}</p>
> -              </FlexCell>
> -            </Flex>
> +            </div>
> ```

when `Flex` is removed it doesn't stretch across the modal as intended. Some of the words are almost not visible within the modal

mendelB:
Hmm I don't _think_ it needs to stretch across the whole modal, but it should take up more than half of the space, I'd assume, since it has more room to breath 🤔 

I think it's getting some padding currently so we'd need to keep that in which I think would fix the issue you're describing (there's also a styling class I think we'd need to keep):
```diff
-            <Flex className="items-center">
+            <div className="affirmation__cta p-3">
-              <FlexCell className="affirmation__cta p-3" width="half">
                <h3>{callToActionHeader}</h3>
                <p>{callToActionDescription}</p>
-              </FlexCell>
-            </Flex>
+            </div>
```

PR #2285, opened 2020-07-20
Remove the Social Override from Phoenix

mendelB:
### What's this PR do?

This pull request completely retires the legacy/deprecated Social Override field from Campaigns/Pages/Campaign Updates, and all associated code.

Just a whole lotta 🗑 ! commit-by-commit could be convenient.
I did my best to confirm that nothing active is being deprecated, but if you spot something give a shout! 

The initiative to replace this field with the Metadata content type happened a whole while back starting around https://github.com/DoSomething/phoenix-next/pull/1137, this just purges the code relics!

I got [confirmation](https://dosomething.slack.com/archives/C09ANFQLA/p1595258188069000) that we're ok to retire the only existing (_non functional) instances of these in Campaign Updates so I'll remove those from the Contentful space.

### Relevant tickets

References [Pivotal #173771700](https://www.pivotaltracker.com/story/show/173771700).

### Checklist


mendelB:
Ahhh flip. I forgot that I branched https://github.com/DoSomething/phoenix-next/pull/2286 off of this one, and merged it into here! If you commit-by-commit this one, just ignore the final commit with the additional Campaign Update changes!

contentful/content-types/campaign.js
```diff
@@ -431,8 +418,8 @@ module.exports = function(migration) {
         nodes: {},
       },
       {
-        enabledMarks: ['bold', 'underline', 'italic'],
-        message: 'Only bold, underline, and italic marks are allowed',
+        enabledMarks: ['bold', 'italic', 'underline'],
+        message: 'Only bold, italic, and underline marks are allowed',
```
Dinnall, line 422: Curious to know why these were changed here

resources/assets/components/ContentfulEntry/ContentfulEntry.js
```diff
@@ -80,7 +80,7 @@ class ContentfulEntry extends React.Component {
 
       case 'CampaignUpdateBlock':
         return (
-          <CampaignUpdateContainer
+          <CampaignUpdate
```
Dinnall, line 83: What do you think of the name `UpdateCampaign` here?

mendelB: We're rendering the [`CampaignUpdate` component](https://github.com/DoSomething/phoenix-next/blob/40c785c3cfa5ee507c51519bcc49fdffe9fde121/resources/assets/components/blocks/CampaignUpdate/CampaignUpdate.js) here so we need to invoke it as named. (This is one of our older blocks which allows Campaign Leads to embed an update / link / announcement within campaign pages via Contentful so we called it 'Campaign Update')

contentful/content-types/campaign.js
mendelB: haha this drove me crazy bc I couldn't figure out why the heck the order kept on changing when I generated the migration from the `dev` & `master` spaces on Contentful! Something must be stored in a different order somehow which causes the discrepancy 🤷‍♂️

resources/assets/components/ContentfulEntry/ContentfulEntry.js
Dinnall: ah- makes sense

PR #2286, opened 2020-07-20
Cleanup Campaign Update

mendelB:
### What's this PR do?

This pull request branches off of #2285 to do some quick follow-up cleanup to the Campaign Update. We had some legacy deprecated support for linking to individual Campaign Update blocks on /campaign/:slug/blocks|modal/:id routes which we've since stopped using.

commit-by-commit

Legacy 🗑️ 

### Relevant tickets

References [Pivotal #173771700](https://www.pivotaltracker.com/story/show/173771700).

### Checklist


resources/assets/components/blocks/CampaignUpdate/CampaignUpdate.js
```diff
@@ -80,27 +80,6 @@ const CampaignUpdate = props => {
...
-
-CampaignUpdate.defaultProps = {
-  affiliateLogo: null,
-  link: null,
-  bordered: true,
-  author: null,
-  content: null,
-  closeModal: null,
-};
+CampaignUpdate.propTypes = propType(CampaignUpdateBlockFragment).isRequired;
```
weerd, line 83: Noice!

weerd: Not sure if there's a solve for this, I do like that this is cleaner and all the props defined already in the fragment, but the one thing we lose here is that it's not as easily scannable to know what type (string, boolean, etc) a property will be or what the default values are from a glance.... 🤷‍♂️

mendelB: Yeah that's a really good point! I suppose ideally, we'd have this info accessible via the schema itself, so they'd work together as an interdependency and we'd be able to know what each value is since GraphQL is strictly typed anyhow?

But I also did need to add back in a few props that aren't from GraphQL itself in https://github.com/DoSomething/phoenix-next/pull/2286/commits/4e53d3e390ecbc8ff53e04c73063a4167f155855

weerd: Ya totes, the schema is the place to see it, but meant more at a glance without having to jump out of the file. Granted I think if you have the right editor extension it might do that for you, but not everyone has those set up.

weerd:
> But I also did need to add back in a few props that aren't from GraphQL itself in 4e53d3e

👍

resources/assets/components/blocks/CampaignUpdate/CampaignUpdate.js
mendelB: Hmm yeah agreed (e.g. I don't have the extension set up 😬 🙃 ). Which do you think is the better trade-off? I totally ripped this off of the [CampaignCard](https://github.com/DoSomething/phoenix-next/blob/a26f7ec498699e7179c0ae704cce19f82a2162d7/resources/assets/components/utilities/CampaignCard/CampaignCard.js#L77-L79) so this is really all your fault Diego 😛

weerd: Haha, I ripped it off something else so.... 

Not sure. It's a balance between convenience and readability. Maybe something to bring up with the team?

weerd: But I don't think it should block merging! Maybe a dev roundtable or retro convo :)

mendelB: That's a good idea! Hmm, I can try to throw something in #application-devs to kick off a light discussion!

PR #2288, opened 2020-07-21
Update CancelEmailSubscription.js

Dinnall:
### What's this PR do?

This pull request track clicks on the global unsubscribe

...

...

### Relevant tickets

References [Pivotal # 173901873](https://www.pivotaltracker.com/n/projects/2401401/stories/173901873).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
+++ b/resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
@@ -3,6 +3,11 @@ import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
 import { useQuery, useMutation } from '@apollo/react-hooks';
 
+import {
+  EVENT_CATEGORIES,
+  trackAnalyticsEvent,
+  getPageContext,
+} from '../../../../helpers/analytics';
 import Spinner from '../../../artifacts/Spinner/Spinner';
 
 const EMAIL_SUBSCRIPTION_STATUS = gql`
@@ -46,6 +51,20 @@ const CancelEmailSubscription = props => {
     return <Spinner />;
   }
 
+  const handleGlobalUnsubscribe = () => {
+    trackAnalyticsEvent(
+      'phoenix_clicked_link_action_global_email_unsubscribe',
+      {
+        action: 'link_clicked',
+        category: EVENT_CATEGORIES.siteAction,
+        label: 'global_email_unsubscribe',
+        context: {
+          ...getPageContext(),
+        },
+      },
+    );
+  };
+
   return (
     <div>
       {data.user.emailSubscriptionStatus ? (
@@ -59,6 +78,7 @@ const CancelEmailSubscription = props => {
                 variables: {
                   emailSubscriptionStatus: false,
                 },
+                handleGlobalUnsubscribe,
               })
             }
           >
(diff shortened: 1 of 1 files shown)


resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -59,6 +78,7 @@ const CancelEmailSubscription = props => {
                 variables: {
                   emailSubscriptionStatus: false,
                 },
+                handleGlobalUnsubscribe,
```
mendelB, line 81: It looks like we're passing this method as an argument to the `updateEmailSubscriptionStatus` mutation call, but I don't think the [mutation](https://www.apollographql.com/docs/react/data/mutations/#executing-a-mutation) accepts a callback as a parameter?

I think what we could do though is perhaps move _both_ of these calls into a general e.g. `handleGlobalUnsubscribeClick` method 🤔 

```js
const handleGlobalUnsubscribeClick = () => {
    trackAnalyticsEvent(
     ...
    );

   updateEmailSubscriptionStatus({
     variables: {
       emailSubscriptionStatus: false,
     }});
  };
```

and then right here we just pass that to the `onClick`:
```js
...
onClick={handleGlobalUnsubscribeClick}
...
```

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -46,6 +51,20 @@ const CancelEmailSubscription = props => {
     return <Spinner />;
   }
 
+  const handleGlobalUnsubscribe = () => {
+    trackAnalyticsEvent(
+      'phoenix_clicked_link_action_global_email_unsubscribe',
```
mendelB, line 56: The `phoenix` part is automatically prefixed within [the helper](https://github.com/DoSomething/phoenix-next/blob/23f295bed7925cdeb4fa56ee2365f526bce1f6ee/resources/assets/helpers/analytics.js#L310) so we can leave that part out.

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -46,6 +51,20 @@ const CancelEmailSubscription = props => {
...
+      {
+        action: 'link_clicked',
+        category: EVENT_CATEGORIES.siteAction,
+        label: 'global_email_unsubscribe',
+        context: {
+          ...getPageContext(),
+        },
```
mendelB, line 63: I don't think we need the page context data here since this is singularly on the account subscriptions tab (this data could generally be relevant for e.g. action blocks that could appear on all sorts of Campaigns or Pages etc. where the specific Campaign ID or Page ID would help define context). But in this case, I think we can just commit the `context` property completely 😌

resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -46,6 +51,20 @@ const CancelEmailSubscription = props => {
     return <Spinner />;
   }
 
+  const handleGlobalUnsubscribeClick = () => {
+    trackAnalyticsEvent('clicked_link_action_global_email_unsubscribe', {
+      action: 'link_clicked',
+      category: EVENT_CATEGORIES.siteAction,
+      label: 'global_email_unsubscribe',
+      context: {},
```
mendelB, line 59: We can just omit this `context` property entirely since it's empty.

PR #2289, opened 2020-07-21
Adding More Causes to VR Social Share

lindsayrmaher:
### What's this PR do?

This pull request adds two more cause options to the social share on OVRD's per request of NASSP partnership.

👀 

**Alpha:**
Screen Shot 2020-07-21 at 11 13 19 AM

**Beta:**
Screen Shot 2020-07-21 at 11 13 10 AM

NASSP requested that we add `food insecurity` and `job creation` to this list.

### Relevant tickets

References [Pivotal # 173897788](https://www.pivotaltracker.com/story/show/173897788).

### Checklist

- [X] This PR has been added to the relevant Pivotal card.
- [X] Added screenshots of front-end changes on small, medium, and large screens.

### DIFF
--- a/resources/assets/components/actions/VoterRegistrationDriveAction/QueryOptions.js
+++ b/resources/assets/components/actions/VoterRegistrationDriveAction/QueryOptions.js
@@ -15,6 +15,8 @@ const votingReasons = {
   'homelessness-and-poverty': 'Homelessness & Poverty',
   education: 'Education Reform',
   'gender-equality': 'Gender Equality',
+  'food-insecurity': 'Food Insecurity',
+  'job-creation': 'Job Creation',
 };
 
 const QueryOptions = ({ onChange }) => {
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/config.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/config.js
@@ -41,4 +41,6 @@ export const votingReasons = {
   'homelessness-and-poverty': 'homelessness and poverty',
   education: 'education reform',
   'gender-equality': 'gender equality',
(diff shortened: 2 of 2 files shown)


resources/assets/components/actions/VoterRegistrationDriveAction/QueryOptions.js
```diff
@@ -15,6 +15,8 @@ const votingReasons = {
   'homelessness-and-poverty': 'Homelessness & Poverty',
   education: 'Education Reform',
   'gender-equality': 'Gender Equality',
+  'food-insecurity': 'Food Insecurity',
+  'job-creation': 'Job Creation',
```
DFurnes, line 19: What's the difference between this object and the one in `config.js` below?

lindsayrmaher: When we originally built out the alpha and beta pages, they both had their own config files so that it was clear what was for which page. And even though we changed the implementation, i think that's why there's still these two separate objects, since the one config is specific to the beta OVRD page and the `QueryOptions` are specific to the social share for the alpha campaign. @aaronschachter might be able to shed more light on that!

aaronschachter: The difference is the capitalization in each value of the key/value pair. I created these as two separate objects because some key/values don't have consistent rules (e.g. acronyms like COVID and LGTBQ) that we'd need to add edge cases for if we converted the option text to lowercase, but also as a precaution in case the option text wouldn't match the beta page text (e.g. if we wanted the beta page to read "creating jobs" for the "Job Creation" option)

PR #2294, opened 2020-07-23
Add Link to Groups Campaign Page on Beta OVRD page

lindsayrmaher:
### What's this PR do?

This pull request adds a link to the group campaign page for Beta users in case they're interested in learning more or in signing up via the group campaign page vs individual. The text reads `What's INSERT_GROUPTYPE_NAME` and links out to a different campaign page.

Let me know if there's any refactoring that could happen here, naming makes sense, etc!

Desktop:
Screen Shot 2020-07-23 at 4 50 08 PM

Mobile:
Screen Shot 2020-07-23 at 4 50 21 PM

We're providing more context to users who are visiting a beta page and may not have background info on the group their friend is a part of.

### Relevant tickets

References [Pivotal # 173410336](https://www.pivotaltracker.com/story/show/173410336).

### Checklist

### DIFF
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/GroupTypeLink.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/GroupTypeLink.js
@@ -61,6 +61,8 @@ const GroupTypeLink = ({ group }) => {
           ) : (
             <a
               href={groupCampaignPath}
+              target="_blank"
+              rel="noopener noreferrer"
               data-test="voter-registration-drive-page-group-campaign-link"
             >
               What&lsquo;s {group.groupType.name}?
--- a/resources/assets/components/pages/VoterRegistrationDrivePage/GroupTypeLink.js
+++ b/resources/assets/components/pages/VoterRegistrationDrivePage/GroupTypeLink.js
@@ -1,0 +1,84 @@
+import React from 'react';
+import { get } from 'lodash';
+import gql from 'graphql-tag';
+import PropTypes from 'prop-types';
+import { useQuery } from '@apollo/react-hooks';
+
+import ErrorBlock from '../../blocks/ErrorBlock/ErrorBlock';
+import PlaceholderText from '../../utilities/PlaceholderText/PlaceholderText';
+
+const GROUPS_CAMPAIGN_QUERY = gql`
+  query GroupsCampaignQuery($groupTypeId: Int!, $causes: [String]) {
+    campaigns(groupTypeId: $groupTypeId, causes: $causes) {
+      id
+    }
+  }
+`;
+
+const GROUPS_CAMPAIGN_WEBSITE_QUERY = gql`
+  query GroupsCampaignWebsiteQuery($campaignId: String!) {
+    campaignWebsiteByCampaignId(campaignId: $campaignId) {
+      id
+      path
+      url
+    }
+  }
+`;
+
+const GroupTypeLink = ({ group }) => {
+  const { loading, error, data } = useQuery(GROUPS_CAMPAIGN_QUERY, {
+    variables: {
+      groupTypeId: get(group, 'groupType.id'),
+      causes: ['voter-registration'],
+    },
+  });
+  const groupTypeCampaignId = get(data, 'campaigns[0].id', null);
+
+  const {
+    loading: campaignWebsiteLoading,
+    error: campaignWebsiteError,
+    data: campaignWebsiteData,
+  } = useQuery(GROUPS_CAMPAIGN_WEBSITE_QUERY, {
+    skip: groupTypeCampaignId == null,
+    variables: {
+      campaignId: String(groupTypeCampaignId),
+    },
+  });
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/VoterRegistrationDrivePage/GroupTypeLink.js
```diff
@@ -0,0 +1,86 @@
...
+};
+
+GroupTypeLink.propTypes = {
+  group: PropTypes.shape({
+    groupType: PropTypes.shape({
+      id: PropTypes.number,
+      name: PropTypes.string,
+    }),
+  }).isRequired,
+};
```
DFurnes, line 84: It might be nice to swap these to be two explicit `id` and `name` props. This way we don't have a sort of "hidden" implicit dependency on the shape of the `group` prop being passed in here... e.g.

```diff
- {group ? <GroupTypeLink group={group} /> : null}
+ {group ? <GroupTypeLink id={group.groupType.id} name={group.groupType.id} /> : null}
```

resources/assets/components/pages/VoterRegistrationDrivePage/GroupTypeLink.js
```diff
@@ -0,0 +1,86 @@
...
+  } = useQuery(GROUPS_CAMPAIGN_WEBSITE_QUERY, {
+    skip: groupTypeCampaignId == null,
+    variables: {
+      campaignId: String(groupTypeCampaignId),
+    },
+  });
+  const groupCampaignPath = get(
+    campaignWebsiteData,
+    'campaignWebsiteByCampaignId.path',
+    null,
```
DFurnes, line 50: Is there a fallback URL we'd want to use here? Or should we show the error block if this happens?

aaronschachter: If a campaign doesn't exist for the group, it could because it's not published. I'd think we'd be okay to just return `null`

resources/assets/components/pages/VoterRegistrationDrivePage/GroupTypeLink.js
lindsayrmaher: 😍  great idea!

lindsayrmaher: updates on fce9d46

resources/assets/components/pages/VoterRegistrationDrivePage/GroupTypeLink.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+`;
+
+const GROUPS_CAMPAIGN_WEBSITE_QUERY = gql`
+  query GroupsCampaignWebsiteQuery($campaignId: String!) {
+    campaignWebsiteByCampaignId(campaignId: $campaignId) {
+      id
+      path
+      url
```
aaronschachter, line 23: Looks like we're not using this `url` and can remove? (or link to `url` vs `path`, I'm forgetting if `path` returns the `/us/campaigns`

resources/assets/components/pages/VoterRegistrationDrivePage/GroupTypeLink.js
```diff
@@ -0,0 +1,82 @@
...
+        <ErrorBlock error={error || campaignWebsiteError} />
+      ) : (
+        <div className="mt-3">
+          {loading || campaignWebsiteLoading ? (
+            <PlaceholderText size="medium" />
+          ) : (
+            <a
+              href={groupCampaignPath}
+              target="_blank"
+              rel="noopener noreferrer"
```
aaronschachter, line 65: I'd imagine we don't want this, since it's an internal URL?

aaronschachter: Oops, sorry read the request in Pivotal to open in a new tab.

lindsayrmaher: I had assumed we wouldn't want to lead beta's away from the page entirely since they could still take the action of registering to vote. But I can take it out if that's not a concern.

resources/assets/components/pages/VoterRegistrationDrivePage/VoterRegistrationDrivePageBanner.js
```diff
@@ -98,6 +100,12 @@ const VoterRegistrationDrivePageBanner = ({
               After you register, share with your friends to enter to win a $
               {`${scholarshipAmount.toLocaleString()}`} scholarship!
             </p>
+            {group ? (
+              <GroupTypeLink
+                id={get(group, 'groupType.id', null)}
+                name={get(group, 'groupType.name', null)}
```
aaronschachter, line 106: Nit, but we don't need to use `get` here because this component does expect these within the shape Proptype, and in line 103 we've checked that the `group` exists

PR #2293, opened 2020-07-23
Adds tests for voter-registration helper

aaronschachter:
### What's this PR do?

This pull request adds unit tests for the `voter-registration` helper introduced in #2292 , and moves the `getVoterRegistrationTrackingSource` into it as well to chip away at breaking up the massive index helper into smaller ones.

### How should this be reviewed?

👀 

### Any background context you want to provide?

🧮 

### Relevant tickets

References [Pivotal #173807000](https://www.pivotaltracker.com/story/show/173807000).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2299, opened 2020-07-28
Connect hola bar with user vr status,

Dinnall:
### What's this PR do?

This pull request connect HolaBar with user VR status

... 🥇 

users that are unregistered to vote will continue to see the Holabar
users that are ineligible or completed registration will not see the banner
...

### Relevant tickets

References [Pivotal # 172147696](https://www.pivotaltracker.com/n/projects/2401401/stories/172147696).

### Checklist

USER REGISTRATION STATUS: registration_complete
Screen Shot 2020-07-27 at 10 44 07 PM

USER REGISTRATION STATUS: UNREGISTERED

Screen Shot 2020-07-27 at 10 47 33 PM

### DIFF
--- a/cypress/integration/sitewide-banner.js
+++ b/cypress/integration/sitewide-banner.js
@@ -5,18 +5,6 @@ import faker from 'faker';
 import { userFactory } from '../fixtures/user';
 import exampleCampaign from '../fixtures/contentful/exampleCampaign';
 
-const exampleRegisteredUser = {
-  id: '3401458',
-  displayName: 'Michael',
-  voterRegistrationStatus: 'REGISTRATION_COMPLETE',
-};
-
-const exampleUnregisteredUser = {
-  id: '3401458',
-  displayName: 'Michael',
-  voterRegistrationStatus: 'UNREGISTERED',
-};
-
 describe('Site Wide Banner', () => {
   beforeEach(() => {
     cy.configureMocks();
@@ -157,12 +145,13 @@ describe('Site Wide Banner', () => {
     if ('#banner-portal > .wrapper > [data-test=site-wide-banner]') {
       cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
 
-      cy.findByTestId('sitewide-banner-button').should('have.length', 1);
-      cy.findByTestId('sitewide-banner-button').should(
-        'have.attr',
-        'href',
-        `https://vote.dosomething.org/?r=user:${user.id},source:web,source_details:hellobar`,
-      );
-    }
+    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
+
+    cy.findByTestId('sitewide-banner-button').should('have.length', 1);
+    cy.findByTestId('sitewide-banner-button').should(
+      'have.attr',
+      'href',
+      `https://vote.dosomething.org/?r=user:${user.id},source:web,source_details:hellobar`,
+    );
   });
 });
--- a/cypress/integration/sitewide-banner.js
+++ b/cypress/integration/sitewide-banner.js
@@ -140,13 +140,8 @@ describe('Site Wide Banner', () => {
       },
     });
 
-    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
-
     if ('#banner-portal > .wrapper > [data-test=site-wide-banner]') {
       cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
-
-      cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
-
       cy.findByTestId('sitewide-banner-button').should('have.length', 1);
       cy.findByTestId('sitewide-banner-button').should(
         'have.attr',
--- a/cypress/integration/sitewide-banner.js
+++ b/cypress/integration/sitewide-banner.js
@@ -5,6 +5,18 @@ import faker from 'faker';
 import { userFactory } from '../fixtures/user';
 import exampleCampaign from '../fixtures/contentful/exampleCampaign';
 
+const exampleRegisteredUser = {
+  id: '3401458',
+  displayName: 'Michael',
+  voterRegistrationStatus: 'REGISTRATION_COMPLETE',
+};
+
+const exampleUnregisteredUser = {
+  id: '3401458',
+  displayName: 'Michael',
+  voterRegistrationStatus: 'UNREGISTERED',
+};
+
 describe('Site Wide Banner', () => {
   beforeEach(() => {
     cy.configureMocks();
@@ -23,9 +35,17 @@ describe('Site Wide Banner', () => {
   it('The Site Wide Banner is displayed on campaign pages', () => {
     cy.anonVisitCampaign(exampleCampaign);
 
+    if ('#banner-portal > .wrapper > [data-test=site-wide-banner]') {
+      cy.get('#banner-portal > .wrapper > [data-test=site-wide-banner]').should(
+        'have.length',
+        1,
+      );
+    }
+  });
+
+  it('Site Wide Banner is not displayed on any pages when data-test=site-wide-banner is not visible', () => {
     cy.get('#banner-portal > .wrapper > [data-test=site-wide-banner]').should(
-      'have.length',
-      1,
+      'not.exist',
     );
   });
 
--- a/cypress/integration/sitewide-banner.js
+++ b/cypress/integration/sitewide-banner.js
@@ -43,12 +43,6 @@ describe('Site Wide Banner', () => {
     }
   });
 
-  it('Site Wide Banner is not displayed on any pages when data-test=site-wide-banner is not visible', () => {
-    cy.get('#banner-portal > .wrapper > [data-test=site-wide-banner]').should(
-      'not.exist',
-    );
-  });
-
   it('The Site Wide Banner is not displayed on the beta voter registration (OVRD) drive page', () => {
     const user = userFactory();
 
@@ -106,12 +100,14 @@ describe('Site Wide Banner', () => {
   it('The Site Wide Banner CTA URL is correct for an unauthenticated user', () => {
     cy.anonVisitCampaign(exampleCampaign);
 
-    cy.findByTestId('sitewide-banner-button').should('have.length', 1);
-    cy.findByTestId('sitewide-banner-button').should(
-      'have.attr',
-      'href',
-      'https://vote.dosomething.org/?r=source:web,source_details:hellobar',
-    );
+    if ('#banner-portal > .wrapper > [data-test=site-wide-banner]') {
+      cy.findByTestId('sitewide-banner-button').should('have.length', 1);
+      cy.findByTestId('sitewide-banner-button').should(
+        'have.attr',
+        'href',
+        'https://vote.dosomething.org/?r=source:web,source_details:hellobar',
+      );
+    }
   });
 
   /** @test */
@@ -119,13 +115,16 @@ describe('Site Wide Banner', () => {
     const user = userFactory();
 
     // Log in & visit the campaign pitch page:
-    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
 
-    cy.findByTestId('sitewide-banner-button').should('have.length', 1);
-    cy.findByTestId('sitewide-banner-button').should(
-      'have.attr',
-      'href',
-      `https://vote.dosomething.org/?r=user:${user.id},source:web,source_details:hellobar`,
-    );
+    if ('#banner-portal > .wrapper > [data-test=site-wide-banner]') {
+      cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
+
+      cy.findByTestId('sitewide-banner-button').should('have.length', 1);
+      cy.findByTestId('sitewide-banner-button').should(
+        'have.attr',
+        'href',
+        `https://vote.dosomething.org/?r=user:${user.id},source:web,source_details:hellobar`,
+      );
+    }
   });
 });
--- a/cypress/integration/sitewide-banner.js
+++ b/cypress/integration/sitewide-banner.js
@@ -136,8 +136,9 @@ describe('Site Wide Banner', () => {
       },
     });
 
+    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
+
     if ('#banner-portal > .wrapper > [data-test=site-wide-banner]') {
-      cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
       cy.findByTestId('sitewide-banner-button').should('have.length', 1);
       cy.findByTestId('sitewide-banner-button').should(
         'have.attr',
--- a/cypress/integration/sitewide-banner.js
+++ b/cypress/integration/sitewide-banner.js
@@ -145,13 +145,14 @@ describe('Site Wide Banner', () => {
     if ('#banner-portal > .wrapper > [data-test=site-wide-banner]') {
       cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
 
-    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
+      cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
 
-    cy.findByTestId('sitewide-banner-button').should('have.length', 1);
-    cy.findByTestId('sitewide-banner-button').should(
-      'have.attr',
-      'href',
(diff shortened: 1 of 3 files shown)


resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -1,9 +1,27 @@
...
 import { createPortal } from 'react-dom';
 import React, { useRef, useEffect } from 'react';
+import { useQuery } from '@apollo/react-hooks';
 
+import { getUserId } from '../../../helpers/auth';
 import excludedPaths from './config';
 import SitewideBannerContent from './SitewideBannerContent';
 
+const VOTER_REGISTRATION_STATUS = gql`
+  query VoterRegBadgeQuery($userId: String!) {
```
katiecrane, line 11: I believeee each query needs to have a unique name so we should change this - maybe to something like `VoterRegSitewideBannerQuery`?

cypress/integration/sitewide-banner.js
```diff
@@ -5,6 +5,18 @@ import faker from 'faker';
...
+  id: '3401458',
+  displayName: 'Michael',
+  voterRegistrationStatus: 'REGISTRATION_COMPLETE',
+};
+
+const exampleUnregisteredUser = {
+  id: '3401458',
+  displayName: 'Michael',
+  voterRegistrationStatus: 'UNREGISTERED',
+};
```
katiecrane, line 18: The content here is great but we want to return these from GraphQL mocks, so something like:
```js
cy.mockGraphqlOp('VoterRegSitewideBannerQuery', {
    user: {
        id: '3401458',
        firstName: 'Michael',
        voterRegistrationStatus: 'UNREGISTERED',
    },
});
```
[Here's an example of a similar mock](https://github.com/DoSomething/phoenix-next/blob/master/cypress/integration/campaign-signup.js#L151-L156)
And we'll want one inside the test for registered users and one in the test for unregistered, so they'll need to move down into the corresponding tests

Dinnall: ah - I was a little confused about the `mockGraphqlOp`, with this query would 
`user .voterRegistrationStatus` be updated according to the user status in the database? or would I need to create one for a registered user ass well?

katiecrane: We're not actually going to touch any databases - we're just going to fake what we get back for the `VoterRegSitewideBannerQuery` query in `SitewideBanner.js`. So we'll have two separate test cases, and for each one we'll create a new mock with the info we want so we'll end up with one mock for `REGISTERED` and a separate mock for `REGISTRATION_COMPLETE`

Dinnall:
@mendelB @katiecrane I wanted to try if statements would work for the 3 failing cypress test, curious to know why this approach didn't work before I revert back to using `mockGraphqlOp`

mendelB:
> @mendelB @katiecrane I wanted to try if statements would work for the 3 failing cypress test, curious to know why this approach didn't work before I revert back to using `mockGraphqlOp`

I think the reason the `if` statements aren't working is because they're evaluating a populated String as the condition, which will _always_ be 'truthy' and thus running the nested assertions which fail.

But I think in this case we don't want to rely on optionally running these assertions (which is an interesting approach! Kinda reminds me of Ghost Inspector's [conditional steps](https://ghostinspector.com/docs/conditional-steps/)), but rather we can create the correct environment for ourselves more explicitly via the GraphQL mocks so that we can actively assert whether or not the banner is showing up, depending on the user's `voterRegistrationStatus`!

resources/assets/components/App.js
```diff
@@ -54,6 +54,22 @@ const App = ({ store, history }) => {
...
+            context={{ contextSource: 'voter_registration' }}
+            render={(handleClose, handleComplete) => (
+              <SitewideBanner
+                cta="Get Started"
+                description="Make your voice heard. Register to vote in less than 2 minutes."
+                handleClose={handleClose}
+                handleComplete={handleComplete}
+                link={`https://vote.dosomething.org/?r=${getVoterRegistrationTrackingSource(
+                  'hellobar',
+                )}`}
```
mendelB, line 69: Interesting! So it looks like we renamed this method to `getTrackingSource` in https://github.com/DoSomething/phoenix-next/pull/2293, so perhaps we rebased here and now we need to update this method call. 

I'm now seeing that this rebase must have re-added the sitewide banner above as well! Looks like it's duplicated above on lines 40-55.

cypress/integration/sitewide-banner.js
```diff
@@ -23,9 +23,41 @@ describe('Site Wide Banner', () => {
   it('The Site Wide Banner is displayed on campaign pages', () => {
     cy.anonVisitCampaign(exampleCampaign);
 
+    if ('#banner-portal > .wrapper > [data-test=site-wide-banner]') {
+      cy.get('#banner-portal > .wrapper > [data-test=site-wide-banner]').should(
+        'have.length',
+        1,
+      );
+    }
```
mendelB, line 31: Let's remove the conditional here since we'd always expect this banner to appear for anonymous users on a campaign page.
```suggestion
    cy.get('#banner-portal > .wrapper > [data-test=site-wide-banner]').should(
      'have.length',
      1,
    );
```

cypress/integration/sitewide-banner.js
```diff
@@ -86,26 +118,41 @@ describe('Site Wide Banner', () => {
...
-      'https://vote.dosomething.org/?r=source:web,source_details:hellobar',
-    );
+    if ('#banner-portal > .wrapper > [data-test=site-wide-banner]') {
+      cy.findByTestId('sitewide-banner-button').should('have.length', 1);
+      cy.findByTestId('sitewide-banner-button').should(
+        'have.attr',
+        'href',
+        'https://vote.dosomething.org/?r=source:web,source_details:hellobar',
+      );
+    }
```
mendelB, line 128: Let's remove this conditional as well. We'd always expect the banner in this scenario.
```suggestion
    cy.findByTestId('sitewide-banner-button').should('have.length', 1);
    cy.findByTestId('sitewide-banner-button').should(
      'have.attr',
      'href',
      'https://vote.dosomething.org/?r=source:web,source_details:hellobar',
    );
```

cypress/integration/sitewide-banner.js
```diff
@@ -86,26 +118,41 @@ describe('Site Wide Banner', () => {
...
+
+      cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
+
+      cy.findByTestId('sitewide-banner-button').should('have.length', 1);
+      cy.findByTestId('sitewide-banner-button').should(
+        'have.attr',
+        'href',
+        `https://vote.dosomething.org/?r=user:${user.id},source:web,source_details:hellobar`,
+      );
+    }
```
mendelB, line 156: Same with this conditional.
```suggestion
    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);

    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);

    cy.findByTestId('sitewide-banner-button').should('have.length', 1);
    cy.findByTestId('sitewide-banner-button').should(
      'have.attr',
      'href',
      `https://vote.dosomething.org/?r=user:${user.id},source:web,source_details:hellobar`,
    );
```

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -1,9 +1,28 @@
...
 import { createPortal } from 'react-dom';
+import { useQuery } from '@apollo/react-hooks';
 import React, { useRef, useEffect } from 'react';
 
 import excludedPaths from './config';
+import { getUserId } from '../../../helpers/auth';
 import SitewideBannerContent from './SitewideBannerContent';
 
+const VOTER_REGISTRATION_STATUS = gql`
+  query VoterRegSiteWideBannerQuery($userId: String!) {
```
mendelB, line 12: I think we can downcase the w here:
```suggestion
  query VoterRegSitewideBannerQuery($userId: String!) {
```

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -1,9 +1,28 @@
...
+    user(id: $userId) {
+      id
+      voterRegistrationStatus
+    }
+  }
+`;
+
+const shouldDisplayRegistrationBanner = [
```
mendelB, line 20: I wonder if we can be slightly more descriptive with this variable, e.g. `unregisteredStatuses`?

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -1,9 +1,28 @@
...
 
 import excludedPaths from './config';
+import { getUserId } from '../../../helpers/auth';
 import SitewideBannerContent from './SitewideBannerContent';
 
+const VOTER_REGISTRATION_STATUS = gql`
+  query VoterRegSiteWideBannerQuery($userId: String!) {
+    user(id: $userId) {
+      id
+      voterRegistrationStatus
```
mendelB, line 15: This makes me wonder about how neat it would be if we could add a graphQL resolved boolean field indicating whether the user is or isn't registered so we can encapsulate all this logic! (Though maybe even better if this was computed in Northstar itself either as a persisted field or in the API response 🤔 )

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
Dinnall: Like this suggestion

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
katiecrane: @mendelB Oooo I love the idea of not having to think about this everytime! Though, I wonder if what we consider "registered" may change based on the use - for instance in this case, we are treating "confirmed" members as unregistered because we know that many of them are not, but in the user profile if we were to surface their status it'd be registered.

Dinnall: Is there an example within the codebase, that I can take a look at? 
or should we create a ticket to do this within Northstar when I comeback 🤔

Dinnall: Also ` VoterRegistrationStatus` returns a few different statuses, so would we have to account for each one?
To @katiecrane point, I agree, it would be nice it know if those statuses might change 

cc @mendelB

katiecrane: I think this is just brainstorming for now - no work to be done yet!

cypress/integration/sitewide-banner.js
```diff
@@ -99,6 +129,17 @@ describe('Site Wide Banner', () => {
...
+    cy.mockGraphqlOp('VoterRegSiteWideBannerQuery', {
+      user: {
+        voterRegistrationStatus: 'UNREGISTERED',
+      },
+    });
+
+    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
+
+    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
+
```
katiecrane, line 142: Looks like this test is currently failing and I'm wondering if it has something to do with `    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);` being called 3 times instead of just once?

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
mendelB: Hmm yeah, good points! I suppose if it really does very a lot case-by-case a formalization wouldn't be all that useful. I suppose we can keep an eye on this as we integrate more features with the voter reg status!

cypress/integration/sitewide-banner.js
```diff
@@ -99,13 +129,22 @@ describe('Site Wide Banner', () => {
...
-      `https://vote.dosomething.org/?r=user:${user.id},source:web,source_details:hellobar`,
-    );
+    if ('#banner-portal > .wrapper > [data-test=site-wide-banner]') {
+      cy.findByTestId('sitewide-banner-button').should('have.length', 1);
+      cy.findByTestId('sitewide-banner-button').should(
+        'have.attr',
+        'href',
+        `https://vote.dosomething.org/?r=user:${user.id},source:web,source_details:hellobar`,
+      );
+    }
```
mendelB, line 148: Looks like the conditional snuck in again 👾

Dinnall: I removed the if statement here, but it fails stating `Timed out retrying: Unable to find an element by: [data-testid="sitewide-banner-button"]`

katiecrane: Hmm if you manually go to a campaign page while logged in does it perform as expected? I'm wondering if it's an issue with the test, or if the failing test is actually correct

cypress/integration/sitewide-banner.js
```diff
@@ -99,8 +129,14 @@ describe('Site Wide Banner', () => {
     const user = userFactory();
 
     // Log in & visit the campaign pitch page:
-    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
 
+    cy.mockGraphqlOp('VoterRegSiteWideBannerQuery', {
```
mendelB, line 133: Ahh I think this test is failing because we need to update our mock query per the rename we did
```suggestion
    cy.mockGraphqlOp('VoterRegSitewideBannerQuery', {
```

Same with the ones in the earlier tests!

cypress/integration/sitewide-banner.js
mendelB: I think this is the culprit! https://github.com/DoSomething/phoenix-next/pull/2299#discussion_r466654515

cypress/integration/sitewide-banner.js
```diff
@@ -29,6 +29,36 @@ describe('Site Wide Banner', () => {
     );
   });
 
+  it('The Site Wide Banner is displayed on campaign pages for authenticated users who are not registered to vote', () => {
+    const user = userFactory();
+
+    cy.mockGraphqlOp('VoterRegSiteWideBannerQuery', {
```
mendelB, line 35:
```suggestion
    cy.mockGraphqlOp('VoterRegSitewideBannerQuery', {
```

cypress/integration/sitewide-banner.js
```diff
@@ -29,6 +29,36 @@ describe('Site Wide Banner', () => {
...
+    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
+    cy.get('#banner-portal > .wrapper > [data-test=site-wide-banner]').should(
+      'not.exist',
+    );
+  });
+
+  it('The Site Wide Banner is not displayed on campaign pages for authenticated users who are registered to vote', () => {
+    const user = userFactory();
+
+    cy.mockGraphqlOp('VoterRegSiteWideBannerQuery', {
```
mendelB, line 50:
```suggestion
    cy.mockGraphqlOp('VoterRegSitewideBannerQuery', {
```

cypress/integration/sitewide-banner.js
katiecrane: oh GREAT eye Mendel!!

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -1,9 +1,24 @@
...
+const VOTER_REGISTRATION_STATUS = gql`
+  query VoterRegSitewideBannerQuery($userId: String!) {
+    user(id: $userId) {
+      id
+      voterRegistrationStatus
+    }
+  }
+`;
+
+const unregisteredStatuses = ['UNCERTAIN', 'CONFIRMED', 'UNREGISTERED'];
```
lindsayrmaher, line 20: Hi! Wondering why `CONFIRMED` is included in this? Are we only suppressing this for people who have gone through the RTV registration process?

mendelB: I'm afraid I don't have the background on this one but per Katie's point here https://github.com/DoSomething/phoenix-next/pull/2299#discussion_r466581565 it definitely looks like we're trying indeed to rely solely on confirmed RTV registrations!

lindsayrmaher: 🙈 yeah i just read the ticket and realized this was clarified. my bad!

PR #2306, opened 2020-07-30
Group Finder tweaks

aaronschachter:
### What's this PR do?

This pull request makes the Join Group button just a liiiitle bit smaller on mobile, and renders the help text in a single `p`

Before

Screen Shot 2020-07-30 at 2 06 10 PM

After

Screen Shot 2020-07-30 at 2 05 36 PM

👀 

🦈 

### Relevant tickets

References [Slack thread](https://dosomething.slack.com/archives/CTVPG6L4R/p1596132412220100).

### Checklist

### DIFF
--- a/resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
+++ b/resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
@@ -97,24 +97,27 @@ const CampaignSignupForm = props => {
     <div className="my-3" data-testid="join-group-signup-form">
       <Card title="Join a group" className="rounded bordered">
         <div className="p-3">
+          <p className="text-sm text-gray-500 pt-3 md:pt-0">
+            Can&apos;t find your group?
+          </p>
+
+          <p className="text-sm text-gray-500 mt-0">
+            Email tej@dosomething.org for help.
+          </p>
+
           <GroupFinder
             context={{ campaignId, pageId }}
             groupType={groupType}
             onChange={handleGroupFinderChange}
           />
+
           <PrimaryButton
             attributes={{ 'data-testid': 'join-group-signup-button' }}
-            className={className}
+            className={`${className} py-2 md:py-3`}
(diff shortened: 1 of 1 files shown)


resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
```diff
@@ -102,18 +102,17 @@ const CampaignSignupForm = props => {
             groupType={groupType}
             onChange={handleGroupFinderChange}
           />
+
           <PrimaryButton
             attributes={{ 'data-testid': 'join-group-signup-button' }}
-            className={className}
+            className={`${className} py-2 md:py-3`}
```
weerd, line 108: You should probably import and use the `classnames()` helper function here.

Also not really sure why we're dropping the height of a single button since it would be inconsistent with the rest of the site. 🤷‍♂️

lindsayrmaher: Also wondering whether we should only add these for mobile since this is only an issue for mobile screens!

resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
```diff
@@ -102,18 +102,17 @@ const CampaignSignupForm = props => {
...
             onClick={handleSignup}
             text="Join Group"
           />
+
           <p className="text-sm text-gray-500 pt-3 md:pt-0">
-            Can&apos;t find your group?
-          </p>
-          <p className="text-sm text-gray-500 mt-0">
-            Email tej@dosomething.org for help.
+            Can&apos;t find your group? Email tej@dosomething.org for help.
```
lindsayrmaher, line 115: I'm guessing that we're prioritizing space on mobile vs design on medium sided screens for this text. But just flagging here that on medium screens the text here wraps to a new line, which is why we had decided to go with two `p` tags orignally.

PR #2310, opened 2020-08-03
Hide Campaigns from Website Discovery V1

mendelB:
### What's this PR do?

This pull request adds a bit of a hack for a first pass at hiding campaigns from website discovery (this currently means any `PaginatedCampaignGallery`'s pulling campaigns via Rogue (Explore Campaigns page & Cause Pages)).

This feature is feature-flagged for now.

We simply reject any Groups campaigns (campaign with a `group_type_id`), and then additionally reject campaigns who's ID's are listed in the `DS_HIDE_CAMPAIGN_IDS` site configuration variable.

👀  Is this readable? Does the logic make sense? 

We eventually want to enable toggling a campaign as hidden from the website formally on the Rogue end. (And pending whether we'll continue to reject all groups campaigns but that should also be done via the API formally). Until then, since we've already launched a few Group campaigns & specialized campaigns we want to be hidden from the Explore Campaigns page, we're starting with this hack.

([We've confirmed](https://www.pivotaltracker.com/story/show/173991883/comments/216683338) that we're ok with the UI consequence - since we're filtering the retrieved list of campaigns on the front-end - wherein we won't always be displaying the clean e.g. 36 campaigns nodes per paginated batch (since we might be rejecting some of those per our local filtering)).

### Relevant tickets

References [Pivotal #173991883](https://www.pivotaltracker.com/story/show/173991883).

### Checklist

- [ ] Added appropriate feature/unit tests. _I'm having some trouble with the testing so will opt for a follow-up PR so we can get this deployed ASAP!_

e.g.
- [Without hide feature enabled](https://user-images.githubusercontent.com/12417657/89201259-341b7900-d57f-11ea-9b0c-317aaa05a633.png)
- [With hide feature enabled & campaign assigned](https://user-images.githubusercontent.com/12417657/89201215-249c3000-d57f-11ea-9be6-5378607651a7.png)


resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
```diff
@@ -78,8 +79,22 @@ const PaginatedCampaignGallery = ({
...
+          !siteConfig('hide_campaign_ids').includes(
+            // Our filter will be a list of Strings but the Campaign IDs will be integers.
+            String(get(campaign, 'node.id')),
+          ),
+      )
+    : campaigns;
+
   // Parse out the nested campaign website nodes so we can pass them to the Gallery Block.
-  const campaigns = get(data, 'campaigns.edges', []).map(edge =>
+  const campagnWebsites = filteredCampaigns.map(edge =>
```
aaronschachter, line 97:
```suggestion
  const campaignWebsites = filteredCampaigns.map(edge =>
```

resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
```diff
@@ -102,7 +117,7 @@ const PaginatedCampaignGallery = ({
   return (
     <div className={className} data-ref="paginated-campaign-gallery">
       <GalleryBlock
-        blocks={campaigns}
+        blocks={campagnWebsites}
```
aaronschachter, line 120:
```suggestion
        blocks={campaignWebsites}
```

resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js
mendelB: Ahhh good catch! Thanks for making this suggested change on both ends, I can just conveniently commit 😌

PR #2313, opened 2020-08-04
Beta Page Redesign

lindsayrmaher:
### What's this PR do?

This pull request is a bit hefty, but essentially we're removing the cover image from the beta OVRD page and updating the layout of some of the page elements! This is happening in the hopes that moving these items "above the fold" will help users to take action, aka register!

I need some help getting a background image to work on the top 3rd of the page. There is an image the team wants to use rather than the plain blurple background (see pivotal comments), and I think i'm just not using the `emotion css` correctly. Any insight there would be really helpful. 

Current page: https://www.dosomething.org/us/my-voter-registration-drive?referrer_user_id=5547be89469c64ec7d8b518d&group_id=172

Desktop Redesign:
Screen Shot 2020-08-04 at 5 13 49 PM

Tablet Redesign:
Screen Shot 2020-08-04 at 5 13 19 PM

Screen Shot 2020-08-04 at 2 28 55 PM

Mobile Redesign:
Screen Shot 2020-08-04 at 5 12 36 PM
Screen Shot 2020-08-04 at 5 12 54 PM
Screen Shot 2020-08-04 at 2 27 41 PM

More context in the pivotal ticket.

### Relevant tickets

References [Pivotal # 173994331](https://www.pivotaltracker.com/story/show/173994331).

### Checklist


resources/assets/components/pages/VoterRegistrationDrivePage/VoterRegistrationDrivePage.js
```diff
@@ -112,29 +120,29 @@ const VoterRegistrationDrivePage = () => {
   return (
     <>
       <SiteNavigationContainer />
-      <main data-test="voter-registration-drive-page">
+      <main css={linkStyles} data-test="voter-registration-drive-page">
```
mendelB, line 123: Are we targeting a specific component/element here? I wonder if we can either nest this style more inline with the component where we want to override our general markdown styling, or perhaps inline this styling in the `markdown.scss` (or even keep this styling in here but/) and use a specific classname for targeting so that this is more explicitly defined 🤔  (lol confusing but In other words, I'm mulling over how we can be careful in overwriting our defined Markdown styling in a clear way)

resources/assets/components/pages/VoterRegistrationDrivePage/VoterRegistrationDrivePageBanner.js
```diff
@@ -67,58 +55,48 @@ const VoterRegistrationDrivePageBanner = ({
...
-            </p>
-            {group ? (
-              <GroupTypeLink
-                id={group.groupType.id}
-                name={group.groupType.name}
-              />
-            ) : null}
+        <div
+          css={css`
+            background-image: url('./VoterReg_Patterns-01.png');
```
mendelB, line 61: I had to do something similar in the `CtaReferralPageBanner`, and what worked was [_importing_](https://github.com/DoSomething/phoenix-next/blob/03716ad136a3d00b5a31e36bf6bceba99e306118/resources/assets/components/utilities/CtaReferralPageBanner/CtaReferralPageBanner.js#L8) the asset, and then assigning that value as the `background` in a template string:
https://github.com/DoSomething/phoenix-next/blob/03716ad136a3d00b5a31e36bf6bceba99e306118/resources/assets/components/utilities/CtaReferralPageBanner/CtaReferralPageBanner.js#L54

resources/assets/components/utilities/StartVoterRegistrationForm/StartVoterRegistrationForm.js
```diff
@@ -46,75 +45,66 @@ const StartVoterRegistrationForm = ({
...
-                name="email_address"
-                value={email}
-                onChange={handleChange}
-                data-testid="voter-registration-email-field"
-              />
-            </label>
-          </div>
-
-          <div className="form-item stretched">
-            <label htmlFor="zip" className="font-bold">
```
mendelB, line 85: Are there any accessibility concerns with removing the `<label>`? (I ask this with ignorance, and I do imagine this is not our first input without a label TBF).

resources/assets/components/pages/VoterRegistrationDrivePage/VoterRegistrationDrivePage.js
lindsayrmaher: totally fair! yeah it's specifically for the a tag link inside the div with the markdown class. i tried adding the styling more directly to it but it was getting overridden by the `markdown.scss`. i'll give it a go with a more explicit classname defined within the scss file and we can decide if that's more straighforward!

resources/assets/components/pages/VoterRegistrationDrivePage/VoterRegistrationDrivePageBanner.js
lindsayrmaher: yeeeees ok this was what i was trying to remember lol

resources/assets/components/utilities/StartVoterRegistrationForm/StartVoterRegistrationForm.js
lindsayrmaher: i was wondering that myself. the updated designs only show placeholder text. would it make sense to keep the label wrapper but remove the actual label? so remove the "Email" and "Zip Code" but keep the label element?

mendelB: That's a thought! An and/or: I feel like I've seen conflicting opinions on Aria attributes, but [this article](https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html) seems to suggest using an `aria-label` for such a case.

resources/assets/components/pages/VoterRegistrationDrivePage/VoterRegistrationDrivePageBanner.js
lindsayrmaher: updated successfully on a7a7fe2 🚀

resources/assets/components/utilities/StartVoterRegistrationForm/StartVoterRegistrationForm.js
lindsayrmaher: Oh nice! added on 9fc282e

resources/assets/components/pages/VoterRegistrationDrivePage/VoterRegistrationDrivePage.js
lindsayrmaher: updated on 5bbdcb8

resources/assets/components/pages/VoterRegistrationDrivePage/GroupTypeLink.js
```diff
@@ -55,14 +60,13 @@ const GroupTypeLink = ({ id, name }) => {
...
-          {loading || campaignWebsiteLoading ? (
-            <PlaceholderText size="medium" />
-          ) : (
+          {loading || campaignWebsiteLoading ? null : (
             <a
               href={groupCampaignUrl}
               target="_blank"
               rel="noopener noreferrer"
               data-testid="voter-registration-drive-page-group-campaign-link"
+              css={groupsLink}
```
mendelB, line 69: Non-blocking I'm just curious at this point TBH -- if we use a `style` here, we _should_ be able to override without an `!important` (Emotion also uses classes to assign the styles so I guess it still ends up being lower on the hierarchy then the `.markdown` 🤔 )

```suggestion
              style={{ color: 'white', fontWeight: 700 }}
```

resources/assets/components/pages/VoterRegistrationDrivePage/VoterRegistrationDrivePageBanner.js
```diff
@@ -1,28 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { css } from '@emotion/core';
 
 import { votingReasons } from './config';
 import { query } from '../../../helpers';
 import GroupTypeLink from './GroupTypeLink';
 import ReferralsInfo from './ReferralsInfo';
+import backgroundVrImage from './VoterReg_Patterns-01.png';
```
mendelB, line 9: minor - I think our naming convention is dash-case for these asset filenames?

lindsayrmaher: updated! i meant to rename anyway so thanks for the flag 🎏

resources/assets/components/pages/VoterRegistrationDrivePage/GroupTypeLink.js
lindsayrmaher: fair q, i was able to update the color with emotion but not the font weight, but it does work with a direct style attribute which is still readable and makes sense so i made that change!

PR #2314, opened 2020-08-04
Adds GalleryBlockSignup Component For One-Click Campaign Signup

Vmack:
### What's this PR do?

This pull request adds a graphql mutation in order to redirect an authenticated user to a campaign page with one click. The new `GalleryBlockSignup` uses a helper to check if the user is authenticated and then redirects them to the campaign page. I also added the`GalleryBlockSignup` to the scholarship card "apply now" button to ensure that it works but this is only the beginning so let me know if this is not the best logical approach.

Is this the right approach? Did I add the component to a feasible location? How's the naming? (I'm terrible at naming things 🙈)

See pivotal ticket for more info

### Relevant tickets

References [Pivotal #173228830](https://www.pivotaltracker.com/story/show/173228830).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -13,22 +13,38 @@ export const CREATE_SIGNUP_MUTATION = gql`
     }
   }
 `;
-const GalleryBlockSignup = ({ campaignId }) => {
-  const [handleSignup] = useMutation(CREATE_SIGNUP_MUTATION, {
-    variables: { campaignId },
-  });
+const GalleryBlockSignup = ({ campaignId, slug }) => {
+  const [handleSignup, { called, error }] = useMutation(
+    CREATE_SIGNUP_MUTATION,
+    {
+      variables: { campaignId },
+    },
+  );
 
-  return <SecondaryButton
-    className="w-full"
-    text="Apply Now"
-    onClick={handleSignup}
-  /> ? (
-    <Redirect to={`/us/campaigns/${campaignId}`} />
-  ) : null;
+  if (called && !error) {
+    return <Redirect to={`/us/campaigns/${slug}`} />;
+  }
+
+  if (error) {
+    // TODO: show error state…
+  }
+
+  return (
+    <SecondaryButton
+      className="w-full"
+      text="Apply Now"
+      onClick={handleSignup}
+    />
+  );
 };
 
 GalleryBlockSignup.propTypes = {
-  campaignId: PropTypes.number.isRequired,
+  campaignId: PropTypes.number,
+  slug: PropTypes.string,
+};
+GalleryBlockSignup.defaultProps = {
+  campaignId: null,
+  slug: null,
 };
 
 export default GalleryBlockSignup;
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -5,6 +5,7 @@ import { useMutation } from '@apollo/react-hooks';
 
 import SecondaryButton from '../Button/SecondaryButton';
 import Spinner from '../../artifacts/Spinner/Spinner';
+import ErrorBlock from '../../blocks/ErrorBlock/ErrorBlock';
 
 export const CREATE_SIGNUP_MUTATION = gql`
   mutation CampaignSignupMutation($campaignId: Int!) {
@@ -28,7 +29,7 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
   }
 
   if (error) {
-    // TODO: show error state…
+    return <ErrorBlock error={error} />;
   }
 
   return (
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -13,7 +13,7 @@ export const CREATE_SIGNUP_MUTATION = gql`
     }
   }
 `;
-const GalleryBlockSignup = ({ campaignId, slug }) => {
+const GalleryBlockSignup = ({ campaignId, path }) => {
   const [handleSignup, { called, error }] = useMutation(
     CREATE_SIGNUP_MUTATION,
     {
@@ -22,7 +22,7 @@ const GalleryBlockSignup = ({ campaignId, slug }) => {
   );
 
   if (called && !error) {
-    return <Redirect to={`/us/campaigns/${slug}`} />;
+    return <Redirect to={path} />;
   }
 
   if (error) {
@@ -40,11 +40,11 @@ const GalleryBlockSignup = ({ campaignId, slug }) => {
 
 GalleryBlockSignup.propTypes = {
   campaignId: PropTypes.number,
-  slug: PropTypes.string,
+  path: PropTypes.string,
 };
 GalleryBlockSignup.defaultProps = {
   campaignId: null,
-  slug: null,
+  path: null,
 };
 
 export default GalleryBlockSignup;
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -13,7 +13,7 @@ export const CREATE_SIGNUP_MUTATION = gql`
     }
   }
 `;
-const GallleryBlockSignup = ({ id, campaignId }) => {
+const GalleryBlockSignup = ({ id, campaignId }) => {
   const [handleSignup] = useMutation(CREATE_SIGNUP_MUTATION, {
     variables: { campaignId },
   });
@@ -27,9 +27,9 @@ const GallleryBlockSignup = ({ id, campaignId }) => {
   return <Redirect to={`/us/campaigns/${campaignId}`} />;
 };
 
-GallleryBlockSignup.propTypes = {
+GalleryBlockSignup.propTypes = {
   id: PropTypes.string.isRequired,
   campaignId: PropTypes.number.isRequired,
 };
 
-export default GallleryBlockSignup;
+export default GalleryBlockSignup;
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -4,7 +4,7 @@ import React, { useEffect } from 'react';
 import { Redirect } from 'react-router-dom';
 import { useMutation } from '@apollo/react-hooks';
 
-import { isAuthenticated } from '../../../helpers/auth';
+import SecondaryButton from '../Button/SecondaryButton';
 
 export const CREATE_SIGNUP_MUTATION = gql`
   mutation CampaignSignupMutation($campaignId: Int!) {
@@ -18,11 +18,12 @@ const GalleryBlockSignup = ({ id, campaignId }) => {
     variables: { campaignId },
   });
 
-  useEffect(() => {
-    if (isAuthenticated()) {
-      handleSignup();
-    }
-  }, [id]);
+  <SecondaryButton
+    className="w-full"
+    href={path}
+    text="Apply Now"
+    onClick={handleSignup}
+  />;
 
   return <Redirect to={`/us/campaigns/${campaignId}`} />;
 };
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -1,6 +1,6 @@
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
-import React, { useEffect } from 'react';
+import React from 'react';
 import { Redirect } from 'react-router-dom';
 import { useMutation } from '@apollo/react-hooks';
 
@@ -13,23 +13,21 @@ export const CREATE_SIGNUP_MUTATION = gql`
     }
   }
 `;
-const GalleryBlockSignup = ({ id, campaignId }) => {
+const GalleryBlockSignup = ({ campaignId }) => {
   const [handleSignup] = useMutation(CREATE_SIGNUP_MUTATION, {
     variables: { campaignId },
   });
 
-  <SecondaryButton
+  return <SecondaryButton
     className="w-full"
-    href={path}
     text="Apply Now"
     onClick={handleSignup}
-  />;
-
-  return <Redirect to={`/us/campaigns/${campaignId}`} />;
+  /> ? (
+    <Redirect to={`/us/campaigns/${campaignId}`} />
+  ) : null;
 };
 
 GalleryBlockSignup.propTypes = {
-  id: PropTypes.string.isRequired,
   campaignId: PropTypes.number.isRequired,
 };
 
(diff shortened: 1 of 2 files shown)


resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
```diff
@@ -56,6 +57,7 @@ const ScholarshipCard = ({ campaign }) => {
   ]);
 
   const handleScholarshipCardShareClick = () => {
+    GalleryBlockSignup();
```
katiecrane, line 60: I think we'll want to include the `<GalleryBlockSignup>` component down below (line 119) where we have the `<SecondaryButton>`. But, we only want to display the `<GalleryBlockSignup>` if the user is authenticated, otherwise we'll still show the `<SecondaryButton>`.

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -0,0 +1,35 @@
...
+
+import { isAuthenticated } from '../../../helpers/auth';
+
+export const CREATE_SIGNUP_MUTATION = gql`
+  mutation CampaignSignupMutation($campaignId: Int!) {
+    createSignup(campaignId: $campaignId) {
+      id
+    }
+  }
+`;
```
katiecrane, line 15: This mutation looks great!!!

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -0,0 +1,35 @@
...
+
+  return <Redirect to={`/us/campaigns/${campaignId}`} />;
+};
+
+GallleryBlockSignup.propTypes = {
+  id: PropTypes.string.isRequired,
+  campaignId: PropTypes.number.isRequired,
+};
+
+export default GallleryBlockSignup;
```
katiecrane, line 35: There's a typo here and elsewhere in this file that might be causing issues when you try to import it into the other file (though it is consistent in this file so you might not have seen errors here!):
```suggestion
export default GalleryBlockSignup;
```

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -0,0 +1,35 @@
...
+const GallleryBlockSignup = ({ id, campaignId }) => {
+  const [handleSignup] = useMutation(CREATE_SIGNUP_MUTATION, {
+    variables: { campaignId },
+  });
+
+  useEffect(() => {
+    if (isAuthenticated()) {
+      handleSignup();
+    }
+  }, [id]);
```
katiecrane, line 25: This is looking excellent so far! I think this part is going to be slightly different from the Quiz example - instead of `useEffect()` I believe we'll want to have a button that when clicked, triggers `handleSignup()`. The button will be similar to the `<SecondaryButton>` that already exists in the `<ScholarshipCard> so I would copy over that one and edit! And bonus: we don't need to check if the user is authenticated because we'll be handling that in `<ScholarshipCard>` as well!

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
Vmack: good eye!

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
Vmack: ahhh gotcha!

katiecrane:
@Vmack I feel like this is at a stage to be tested with the review app yay!! You have a few linting errors which is causing the review app to not build though - let me know if I can help with those I know sometimes they're gnarly!

Vmack:
> @Vmack I feel like this is at a stage to be tested with the review app yay!! You have a few linting errors which is causing the review app to not build though - let me know if I can help with those I know sometimes they're gnarly!

yes! I noticed. I have a question about the id. Do we need to capture the user id in the `galleryblocksignup` ? @katiecrane

katiecrane:
@Vmack oooo great question and you know what, I don't think we actually do need the user id! Since they're logged in, when we run the mutation to create the signup it'll already know who is signing up since they'll be the one making the request

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -0,0 +1,34 @@
...
+    variables: { campaignId },
+  });
+
+  return <SecondaryButton
+    className="w-full"
+    text="Apply Now"
+    onClick={handleSignup}
+  /> ? (
+    <Redirect to={`/us/campaigns/${campaignId}`} />
+  ) : null;
```
katiecrane, line 27: Ahhhh I think this is what's causing issues! We'll never display the `<SecondaryButton>` but will always just immediately redirect a user to the campaign page here because of how this conditional is setup. We only want to redirect them _after_ they've clicked the button and the mutation has run. @DFurnes I know you did the work similar to this for the quiz pages but I'm not sure if we can apply the exacttt same approach here, what do you think about how we handle the redirect?

DFurnes: A ha! 🕵️‍♀️ Yeah, this is a little trickier than the time I'd done this.

We should be able to handle this by reading the result from [`useMutation`](https://www.apollographql.com/docs/react/data/mutations/#usemutation-api), like this:

```js
const [handleSignup, { called, error }] = useMutation(CREATE_SIGNUP_MUTATION, {
  variables: { campaignId },
});

if (called && !error) {
  return <Redirect to={`/us/campaigns/${campaignId}`} />;
}

if (error) {
  // TODO: show error state…
}

return (
  <SecondaryButton
     className="w-full"
     text="Apply Now"
     onClick={handleSignup}
  />
)
```

Vmack: oh wow! thanks a ton @katiecrane @DFurnes

Vmack: Added it! So there's good news and bad news. The good news is that I am finally able to access the easy-scholarships page however, when I click "Apply Now" it breaks again and the same CampaignRoute error appears again. @katiecrane @DFurnes

mendelB: It looks like we're using the `campaignId` in the campaign path, but we need to use the `slug`
```diff
- return <Redirect to={`/us/campaigns/${campaignId}`} />;
+ return <Redirect to={`/us/campaigns/${slug}`} />;
```

We'll need to add that to the GraphQL query in the `ScholarshipCard` though (like we did with the `campaignId`), and pass that through as a prop (maybe call it `campaignSlug`?) to this `GalleryBlockSignup` component.

DFurnes: Could that be from the `/us/campaigns/9019` path that you're redirecting to? The `/campaigns` route expects a "slug" (like `/us/campaigns/teens-for-jeans`), which is used to load the campaign page contents.

DFurnes: I believe if you pass in the `path` from the query in `ScholarshipCard.js`, that should work here.

DFurnes: Oh jeez sorry, just saw Mendel beat me to it! 🙈

mendelB: Nono thank you! I rambled about needing a new `slug` field, but the `path` is the way to go!

Vmack: thank you guys! I pushed the changes you suggested. However, I'm still getting the campaignRoute error. Maybe it's something I missed? Let me know if you notice anything.

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -0,0 +1,50 @@
...
+  );
+};
+
+GalleryBlockSignup.propTypes = {
+  campaignId: PropTypes.number,
+  slug: PropTypes.string,
+};
+GalleryBlockSignup.defaultProps = {
+  campaignId: null,
+  slug: null,
```
katiecrane, line 47: This is passed through as `campaignSlug` - try updating all the references in this file from `slug` to `campaignSlug` and see if that fixes things!

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
katiecrane: I left a comment on this outside of this thread! https://github.com/DoSomething/phoenix-next/pull/2314#discussion_r492328282

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
Vmack: I tried that and it said that it couldn't query `campaignSlug`

Vmack: should the slug be required in props validation? also should I add it as a variable to the sign up mutation in `GalleryBlockSignup` component?

mendelB: My apologies! I goofed when I said we should add the `slug` because we already are querying the `path`! See https://github.com/DoSomething/phoenix-next/pull/2314#discussion_r492276195

We can simply pass through the `path` from the `scholarshipCard` (and we can update this naming, but keeping it direct here):
`<GalleryBlockSignup campaignId={campaignId} path={path} />`

And then in the `GalleryBlockSignup` use it as the path:
```js
const GalleryBlockSignup = ({ campaignId, path }) => {
...
return <Redirect to={path} />;
```

Which should guarantee us the right redirect, plus no new query properties!

Vmack: thanks a ton! maybe I'm missing something because it's still breaking after clicking 'Apply Now' 😣

mendelB: The scholarship gods are still not placated! Would you mind posting a screenshot or copy/paste of the error output? Things seem to look good to me here!

katiecrane: I'm also pulling down the branch to poke around and see if I can figure anything out!

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
```diff
@@ -114,12 +118,16 @@ const ScholarshipCard = ({ campaign }) => {
           </div>
         </div>
 
-        <SecondaryButton
-          className="w-full"
-          href={path}
-          text="Apply Now"
-          onClick={handleScholarshipCardShareClick}
-        />
+        {isAuthenticated ? (
```
mendelB, line 121: This is actually exported as a function, so we'll need to invoke it to determine the authenticated status (otherwise it'll always evaluate to 'truthy').
```suggestion
        {isAuthenticated() ? (
```

resources/assets/components/utilities/ScholarshipCard/ScholarshipCard.js
```diff
@@ -13,6 +14,7 @@ import {
   trackAnalyticsEvent,
   getPageContext,
 } from '../../../helpers/analytics';
+import { isAuthenticated } from '../../../selectors/user';
```
mendelB, line 17: This is a touch confusing, but the `selectors` are specific to our Redux implementation and rely on the global `state` (and a container) to function properly. 

We can import the shinier `isAuthenticated` from the `/helpers/auth.js`file which will let us simply evaluate the `window.AUTH` object without Redux magic.
```suggestion
import { isAuthenticated } from '../../../helpers/auth';
```

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
katiecrane: Aside from the `isAuthenticated()` stuff that @mendelB pointed out below I didn't find much yet! I do get redirected to something like `http://phoenix.test/us/campaigns/ready-vote/` (which looks right!) but still getting all sorts of errors from `CampaignRoute`. However, when I refresh the page, the errors clear and the page loads!! Is there some issue with the way we use `<Redirect>`? @DFurnes did you run into anything like this when you did the quiz stuff?
image

mendelB: Oh that's probably it! For the quiz, we didn't need a full page refresh for the page we were redirecting to, but in our case, it's the mega campaigns page, which probably needs a refresh for everything to load properly from the backend 🤔 

We probably need to use a `window.location = path` here instead of the `<Redirect/>`!

katiecrane: That seems to work! I'll push up a couple of commits shortly, I have some meetings now

Vmack: excellent!! Good eye @mendelB ! It's working now 🙌🏽

Vmack: The `<Redirect/>` works for me but I can swap it so that it's consistent. Thanks so much you guys!

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -0,0 +1,49 @@
...
+const GalleryBlockSignup = ({ campaignId, path }) => {
+  const [handleSignup, { called, error }] = useMutation(
+    CREATE_SIGNUP_MUTATION,
+    {
+      variables: { campaignId },
+    },
+  );
+
+  if (called && !error) {
+    return (window.location = { path });
```
mendelB, line 24: I think it would be neat to return a `Spinner` component, now that the user will have to wait for a page refresh. It should make the UX seamless, so it feels like the signup is 'processing' while the campaign page loads! (We also need to assign `path` as a string value):
```suggestion
    window.location = path;
    return <Spinner className="flex justify-center p-4" />
```

We'll need to import the Spinner component as well.

Vmack: makes perfect sense, added! thanks @mendelB I assigned `path` as a string value in props validation. Is that what you meant?

mendelB: Awesome! I was referring to the value we're assigning to the `window.location` which now looks perfect!

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -0,0 +1,52 @@
...
+  );
+
+  if (called && !error) {
+    window.location = path;
+
+    return <Spinner className="flex justify-center p-4" />;
+  }
+
+  if (error) {
+    // TODO: show error state…
```
katiecrane, line 31: I think adding in the error state here is all that's left! I tested the functionality (as a logged in user and not) on the review app and that all is working great!

```suggestion
     return <ErrorBlock error={error} />;
```

katiecrane: @Vmack let me know if I can help with this!

Vmack: @katiecrane I just pushed the change, let me know if it looks good?

katiecrane (approved):
this is a go from me!!!

PR #2322, opened 2020-08-10
Hide Hellobar on Groups pages

lindsayrmaher:
### What's this PR do?

This pull request adds a graphql query to check for a set groupTypeId on the campaign. If it's set, we hide the hellobar!

I am struggling with getting all the Cypress tests to pass and I'm not clear on why. The first test will intermittently fail, and the last test is consistently failing. Any insights there would be super helpful!

We're hiding the helloBar on groups campaigns because of the screen real estate it takes up when users are on mobile.

Current Groups Page:
Screen Shot 2020-08-10 at 11 24 56 AM

Hidden HB Groups Page:
Screen Shot 2020-08-10 at 11 22 31 AM

### Relevant tickets

References [Pivotal #174164390](https://www.pivotaltracker.com/story/show/174164390).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
+++ b/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
@@ -70,8 +70,6 @@ const SitewideBanner = props => {
   );
   const campaignGroupTypeId = get(campaignData, 'campaign.groupType.id', null);
 
-  console.log(campaignGroupTypeId);
-
   const usePortal = id => {
     const rootElem = useRef(document.createElement('div'));
 
--- a/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
+++ b/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
@@ -6,6 +6,7 @@ import React, { useRef, useEffect } from 'react';
 
 import excludedPaths from './config';
 import { getUserId } from '../../../helpers/auth';
+import { getCampaign } from '../../../helpers/campaign';
 import SitewideBannerContent from './SitewideBannerContent';
 
 const VOTER_REGISTRATION_STATUS = gql`
@@ -17,7 +18,18 @@ const VOTER_REGISTRATION_STATUS = gql`
   }
 `;
 
+const CAMPAIGN_GROUPTYPE_QUERY = gql`
+  query CampaignSitewideBannerQuery($campaignId: Int!) {
+    campaign(id: $campaignId) {
+      id
+      groupTypeId
+    }
+  }
+`;
+
 const unregisteredStatuses = ['UNCERTAIN', 'CONFIRMED', 'UNREGISTERED'];
+
+const campaign = getCampaign();
 
 const isExcludedPath = pathname => {
   return excludedPaths.find(excludedPath => {
@@ -36,12 +48,21 @@ const isExcludedPath = pathname => {
 
 const SitewideBanner = props => {
   const userId = getUserId();
-  const options = { variables: { userId }, skip: !userId };
-  const { data, loading } = useQuery(VOTER_REGISTRATION_STATUS, options);
+  const campaignId = campaign ? parseInt(campaign.campaignId, 10) : null;
 
+  const options = { variables: { userId }, skip: !userId };
+  const { data } = useQuery(VOTER_REGISTRATION_STATUS, options);
   const unregistered = unregisteredStatuses.includes(
     get(data, 'user.voterRegistrationStatus'),
   );
+
+  const { data: campaignData } = useQuery(CAMPAIGN_GROUPTYPE_QUERY, {
+    variables: {
+      campaignId,
+    },
+    skip: !campaignId,
+  });
+  const campaignGroupTypeId = get(campaignData, 'campaign.groupTypeId', null);
 
   const usePortal = id => {
     const rootElem = useRef(document.createElement('div'));
@@ -61,6 +82,7 @@ const SitewideBanner = props => {
 
   const target = usePortal('banner-portal');
 
+<<<<<<< HEAD
   if (loading) {
     return null;
   }
@@ -72,6 +94,13 @@ const SitewideBanner = props => {
   }
 
   return createPortal(children, target);
+=======
+  return !isExcludedPath(window.location.pathname) &&
+    (!userId || unregistered) &&
+    !campaignGroupTypeId
+    ? createPortal(children, target)
+    : null;
+>>>>>>> add a query for a campaigns groupTypeId, and only show hellobar if it does not have one
 };
 
 export default SitewideBanner;
--- a/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
+++ b/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
@@ -82,25 +82,11 @@ const SitewideBanner = props => {
 
   const target = usePortal('banner-portal');
 
-<<<<<<< HEAD
-  if (loading) {
-    return null;
-  }
-
-  if (isExcludedPath(window.location.pathname) || (userId && !unregistered)) {
-    target.setAttribute('data-testid', 'sitewide-banner-hidden');
-
-    return null;
-  }
-
-  return createPortal(children, target);
-=======
   return !isExcludedPath(window.location.pathname) &&
     (!userId || unregistered) &&
     !campaignGroupTypeId
     ? createPortal(children, target)
     : null;
->>>>>>> add a query for a campaigns groupTypeId, and only show hellobar if it does not have one
 };
 
 export default SitewideBanner;
--- a/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
+++ b/resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
@@ -22,7 +22,6 @@ const CAMPAIGN_GROUPTYPE_QUERY = gql`
   query CampaignSitewideBannerQuery($campaignId: Int!) {
     campaign(id: $campaignId) {
       id
-      groupTypeId
       groupType {
         id
       }
@@ -51,7 +50,7 @@ const isExcludedPath = pathname => {
 
 const SitewideBanner = props => {
   const userId = getUserId();
-  const campaignId = campaign ? parseInt(campaign.campaignId, 10) : null;
+  const campaignId = campaign ? Number(campaign.campaignId) : null;
 
   const options = { variables: { userId }, skip: !userId };
   const { data, loading } = useQuery(VOTER_REGISTRATION_STATUS, options);
@@ -68,7 +67,7 @@ const SitewideBanner = props => {
       skip: !campaignId,
     },
   );
-  const campaignGroupTypeId = get(campaignData, 'campaign.groupType.id', null);
+  const isGroupCampaign = !!get(campaignData, 'campaign.groupType.id');
 
   const usePortal = id => {
     const rootElem = useRef(document.createElement('div'));
@@ -95,7 +94,7 @@ const SitewideBanner = props => {
   if (
     isExcludedPath(window.location.pathname) ||
     (userId && !unregistered) ||
-    campaignGroupTypeId
+    isGroupCampaign
   ) {
     target.setAttribute('data-testid', 'sitewide-banner-hidden');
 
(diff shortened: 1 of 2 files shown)


mendelB:
@lindsayrmaher we might need to consolidate this with the changes in #2319 if possible since there was a bit of a test bug we resolved there. (Should be pretty simple to align though!)

mendelB:
> @lindsayrmaher we might need to consolidate this with the changes in #2319 if possible since there was a bit of a test bug we resolved there. (Should be pretty simple to align though!)

Ahh and just caught

> I am struggling with getting all the Cypress tests to pass and I'm not clear on why. The first test will intermittently fail, and the last test is consistently failing. Any insights there would be super helpful! 

If you have a minute to review #2319 I can merge and you can rebase to fix this up!

lindsayrmaher:
I'm now failing more tests than i was before 🤦‍♀️  @mendelB

mendelB:
> I'm now failing more tests than i was before 🤦‍♀️ @mendelB

Hmm we'd probably need to mock the `CampaignSitewideBannerQuery` in some of the other tests to ensure we're setting ourselves up to show/hide the banner based on the group campaign data.

lindsayrmaher:
I'm trying to figure out why all of the campaigns are getting set with the join group signup vs a normal signup, but that seems off to me. That makes it seem like the example campaign has a groupType set on it..

mendelB:
> I'm trying to figure out why all of the campaigns are getting set with the join group signup vs a normal signup, but that seems off to me. That makes it seem like the example campaign has a groupType set on it..

Cypress will be mocking the return of the `CampaignSitewideBannerQuery` query per test run, so unless we explicitly mock it ourselves, I assume it'll just give us a `groupTypeId` arbitrarily per test!

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -36,13 +51,25 @@ const isExcludedPath = pathname => {
 
 const SitewideBanner = props => {
   const userId = getUserId();
+  const campaignId = campaign ? parseInt(campaign.campaignId, 10) : null;
```
mendelB, line 54: Looks like we use `Number` for this in the `CampaignBanner`?
https://github.com/DoSomething/phoenix-next/blob/86f800ec09764d9517c5aed1e13bce0fa5f021b0/resources/assets/components/CampaignBanner/CampaignBanner.js#L62

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -36,13 +51,25 @@ const isExcludedPath = pathname => {
...
+  const { data: campaignData, loading: campaignLoading } = useQuery(
+    CAMPAIGN_GROUPTYPE_QUERY,
+    {
+      variables: {
+        campaignId,
+      },
+      skip: !campaignId,
+    },
+  );
+  const campaignGroupTypeId = get(campaignData, 'campaign.groupType.id', null);
```
mendelB, line 71: I wonder if for readability below (that conditional is getting hefty lol), we can rename this to `isGroupCampaign` or something and render this as a Boolean:
```suggestion
  const isGroupCampaign = !!get(campaignData, 'campaign.groupType.id');
```

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -17,8 +18,22 @@ const VOTER_REGISTRATION_STATUS = gql`
...
 `;
 
+const CAMPAIGN_GROUPTYPE_QUERY = gql`
+  query CampaignSitewideBannerQuery($campaignId: Int!) {
+    campaign(id: $campaignId) {
+      id
+      groupTypeId
+      groupType {
+        id
+      }
```
mendelB, line 28: Any reason we need to query the `groupType` here?

lindsayrmaher: yeah so at first I thought I could just use the `groupTypeId`, but the groupType prop on the campaign page is actually based on whether there's a `groupType` so I changed it and that's how I was able to get all the cypress tests to work as expected. I'll remove `groupTypeId` now so it's less confusing.

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
lindsayrmaher: yah! I had that originally, but then was having another issue and changed it hoping that was maybe the problem. It wasn't so i'll change it back :)

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
mendelB: Shouldn't this be independant of the campaign page logic though? (I assume there it runs off of a separate GraphQL query) but here and in our test mocks we can determine this via the `groupTypeId` in isolation and prevent the nested more expensive association.

🤔  If tests are failing regardless that might be a separate issue! But I could be missing something here.

lindsayrmaher: So I just double checked this, and all the tests where the banner should show up fail if I just mock the groupTypeId to be false and check the groupTypeId in the `SiteWideBanner` component itself.

lindsayrmaher: I feel like maybe I'm not understanding something...

mendelB: Hmmmm that is mysterious! I'm gonna take a peek locally. I feel like there's some mischief afoot!

lindsayrmaher: I had initially thought it had something to do with the way the campaign pages were rendering within cypress bc the "join group" signup is showing up on all 8 tests even though that's not set up as part of the example campaign and that's determined by whether the `groupType` object is present vs the `groupTypeId`.

lindsayrmaher: i had assumed in theory that all the campaigns used in the `sitewide-banner` test file would have the regular signup except for the one with a group defined. but i think i'm misunderstanding that.

mendelB: If I understand this correctly, our `SitewideBanner` logic should be completely de-coupled from the actual campaign signup -- not in practice of course since we're reading from the same Rogue campaign in both the regular campaign & banner queries, but within the tests, since we're stubbing out the GraphQL queries and thus should be seeing an independent result per query unless we explicitly stub it ourselves!

mendelB: @lindsayrmaher I'm officially stumped! Something seems to be kicking off an _extra_ `CampaignSitewideBannerQuery` which evades out test stub and returns a `groupTypeId` causing the banner to disappear. (even after a valid appearance per e.g. the first test where we stub it to not be a group campaign).

Something must be somehow causing this extra re-render, and it seems to have to do with the `getCampaign` call we're making because when I remove that, the re-render doesn't happen. But I'm not sure why, what's being updated, and how this extra query result keeps on evading our manual stubs (I even tried stubbing this formally [here](https://github.com/DoSomething/phoenix-next/blob/6eacc2924505860f615faf47a617a913a3177033/cypress/fixtures/graphql.js#L40-L70) but no dice!)

image

I suspect that the reason the tests are passing now is because the query itself is failing out due to us stubbing the `group` with a `null` ID:
image

When I update the stub to be a `group: null` we end up with the same issue as before with the phantom extra query result which doesn't respect the mock:
image

mendelB: Putting the logic to fetch the campaign inside a useEffect solved this for me - e.g.
```js
  let campaignId;
  const userId = getUserId();

  useEffect(() => {
    const campaign = getCampaign();
    campaignId = campaign ? Number(campaign.campaignId) : null;
  }, []);
```

🤷‍♂️   There's something afoot with `window.STATE` & the campaign or _something_ that seems to be unique to our Cypress flow and is causing this headache. I feel like with this light hack you're at least unblocked for now, but there's certainly an underlying issue here.

lindsayrmaher: HMMMMMMMMMMM thanks for all of this investigating, @mendelB! I'll try out moving the campaign stuff to useEffect(). Do you think this is worth making an ice box ticket to investigate? I can if so!

mendelB: That sounds like it could be worthwhile! (it would be neat to link to the ticket with a clarifying comment from the `useEffect`!)

lindsayrmaher: When you ran this locally using `useEffect`, did you pass all the tests? I'm now having an issue where for the groups specific test, I'm failing bc it's not recognizing the `campaignId` value change outside the scope of the useEffect func 😭 😭 😭

mendelB: Well I'll be damned, apologies. I didn't properly test out my theory and that might have just been a flat out wrong idea! I can try taking another look after megagroom!

aaronschachter: I'm wondering if making 2 GraphQL queries could be causing problem here, and if might be simpler to use a single query for both checks?
```
  query SitewideBannerQuery($userId: String, $campaignId: Int) {
    user(id: $userId) {
      id
      voterRegistrationStatus
    }
    campaign(id: $campaignId) {
      id
      groupType {
        id
      }
  }
}
```

lindsayrmaher: 💡 🤯  - didn't think of this, but will try!

mendelB: I thought we had to maintain these separately because of their independent nature 🤔 there might not be a userId but it's a group campaign, or there is a userId but it's not a campaign page

mendelB: This is where the @skip [directive](https://graphql.org/learn/queries/#directives) would be handy!

lindsayrmaher: yeah we're using the skip directive on both of the queries separately, so perhaps we combine and see what happens

mendelB: In our case though, we're using the [Apollo Client skip](https://www.apollographql.com/docs/react/data/queries/#options). 

I think the issue is that we'd need to skip _one_ and _not_ the other query at times. e.g. if we're on a campaign page as an unauthenticated user, we want to run the Campaign query but not the User one. If we bunch them together, we'd be either passing an invalid `null` value to one of the `id` query parameters which would cause the query to fail, or would be skipping both queries when we'd want one of them to run.

mendelB: (If we could make use of the GraphQL `@skip` directive, we could effectively bunch both queries together, and use skip individually per user/voter reg queries from our query schema itself!)

mendelB: I'll be damned once again -- I think this 'phantom' extra campaign query is somehow coming through the `CampaignBannerQuery` [here](https://github.com/DoSomething/phoenix-next/blob/6eacc2924505860f615faf47a617a913a3177033/resources/assets/components/CampaignBanner/CampaignBanner.js#L70-L72)! (You alluded to this @lindsayrmaher re the campaign itself not being a groups campaign).

Mocking this query as well and returning `null` passed the tests for me!

mendelB: What's driving me absolutely bananas is _why_ the hell this separate query interferes with our isolated `campaignData`. I'm chalking it up to the cache and the inner workings of Apollo. It must update the local cache for the banner query with the _unstubbed_ value, and this in turn causes our own `VoterRegSitewideBannerQuery` to update with those cached results. ffs. I'm not completely sure but assume it's something akin to this.

lindsayrmaher: OYYYYYYYY. can confirm this solves the cypress issues, I just pushed up an update for now so we can merge 🤞 Thanks for all this investigation, Mendel!

PR #2323, opened 2020-08-10
Add ActionStatsBlock as available type for Page and StoryPage blocks fields

aaronschachter:
### What's this PR do?

This pull request updates the content type migrations of the `Page` and `StoryPage` fields to allow saving `ActionStatsBlock` entries (introduced in #2312, and component-ized in #2320).

It also starts on minimal documentation for what an `ActionStatsBlock` is, and links to documentation for what an [Action Stat](https://github.com/DoSomething/rogue/blob/master/docs/endpoints/action-stats.md) is.

The documentation updates can be previewed in Gitbook [here](https://dosomething.gitbook.io/phoenix-documentation/v/docs%2Faction-stats-block/development/content-types/action-stats-block)

* Our linter has added all of those extra spaces in the SUMMARY.md, and I have idea why. All I did was add the one line linking to the newly added markdown file.

* I migrated these content types from our `master` environment, as that seems to be the most reliable source of truth for what fields a content type should have.  That being said, this PR adds the `socialOverride` field, which seems like it was added in way back in 2018 per https://github.com/DoSomething/phoenix-next/pull/969. This is a 🤷‍♂️ cc @mendelB 

### Relevant tickets

References [Pivotal #174021616](https://www.pivotaltracker.com/story/show/174021616).

### Checklist


mendelB:
Not sure why that `socialOverride` field didn't make it into `dev`, but let's chalk it up to a drunk Mendel. I've bumped into instances where `dev`-`master` spaces didn't match, and I typically just try and re-align them where possible. (once (if?) we have [this worked out](https://www.pivotaltracker.com/story/show/172900020) this should be a non-issue)

contentful/content-types/storyPage.js
```diff
@@ -130,34 +130,34 @@ module.exports = function(migration) {
 
       validations: [
         {
-          linkContentType: ['sectionBlock'],
+          linkContentType: ['actionStatsBlock', 'sectionBlock'],
```
mendelB, line 133: Oops. @aaronschachter I missed this before, but I think we'd want to support the `actionStatsBlock ` in the `sectionBlock` as a valid embedded entry, and not here. At the top level, Story pages should be using Section Blocks which in turn contain rich text content.

PR #2324, opened 2020-08-11
Query groups by location instead of state

aaronschachter:
### What's this PR do?

This pull request queries the Groups resource by location instead of state within the Group Finder per https://github.com/DoSomething/graphql/pull/263, and also checks for whether a Group Type should filter by location (vs filter by state) per https://github.com/DoSomething/graphql/pull/267. No user facing changes.

This cleanup work is being done to eventually drop the state column from the groups table once this is merged, in an effort to keep how we query for an entity's location consistent in different types, per https://github.com/DoSomething/rogue/pull/1079#issuecomment-666484121.

👀 

🌎 

### Relevant tickets

References [Pivotal #174021616](https://www.pivotaltracker.com/story/show/174021616).

### Checklist

### DIFF
--- a/resources/assets/components/CampaignSignupForm/GroupFinder/GroupSelect.js
+++ b/resources/assets/components/CampaignSignupForm/GroupFinder/GroupSelect.js
@@ -6,19 +6,23 @@ import AsyncSelect from 'react-select/async';
 import { useApolloClient } from '@apollo/react-hooks';
 
 const SEARCH_GROUPS_QUERY = gql`
-  query SearchGroupsQuery($groupTypeId: Int!, $name: String!, $state: String) {
-    groups(groupTypeId: $groupTypeId, name: $name, state: $state) {
+  query SearchGroupsQuery(
+    $groupTypeId: Int!
+    $name: String!
+    $location: String
+  ) {
+    groups(groupTypeId: $groupTypeId, name: $name, location: $location) {
       id
       name
       city
-      state
+      location
     }
   }
 `;
 
 const GroupSelect = ({
   groupLabel,
-  groupState,
+  groupLocation,
   groupTypeId,
   onChange,
   onFocus,
@@ -35,8 +39,8 @@ const GroupSelect = ({
       name: searchString,
     };
 
-    if (groupState) {
-      variables.state = groupState;
+    if (groupLocation) {
+      variables.location = groupLocation;
     }
 
     client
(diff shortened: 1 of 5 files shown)


resources/assets/components/CampaignSignupForm/GroupFinder/GroupSelect.js
```diff
@@ -53,11 +57,11 @@ const GroupSelect = ({
     <AsyncSelect
       getOptionLabel={group =>
         group.city
-          ? `${group.name} - ${group.city}, ${group.state}`
+          ? `${group.name} - ${group.city}, ${group.location.substring(3)}`
```
katiecrane, line 60: Had to think on this line for a minute! But `substring(3)` will output from the 3rd character to the end right, so `US-CA` becomes `CA`?

DFurnes: I wonder if this might be worth extracting into a `iso3166ToStateAbbreviation` helper? Another potential readability bonus would be to split this on the `-`, like `[countryCode, regionCode] = isoCode.split('-')`.

aaronschachter: I've been wondering about making this a calculated GraphQL property, since I'm doing the same thing on the Rogue admin side of things, and we're going to run into this same issue once we update the School GraphQL type to query by `location` instead of `state` eventually.

 But it's weird, when we don't want to just call it `group.state` .. and would call it something like `groupLocationAbbreviation`?

DFurnes: Yeah... this is definitely a tough naming problem!

It could be nice to handle this like we do the [`location`](https://github.com/DoSomething/graphql/blob/fce278b5117ace99c8993044efad17aac2f77f88/src/schema/rogue.js#L574-L575) field in Rogue, where we could pass [`ISO_FORMAT`](https://github.com/DoSomething/graphql/blob/fce278b5117ace99c8993044efad17aac2f77f88/src/schema/rogue.js#L40-L45), `ISO_REGION`, `ISO_COUNTRY`, etc. to parse out the different pieces as needed... then it's always location but we're making it clear we're just formatting it?

aaronschachter: I like that approach, and can add that in a followup PR. 

Per my question in [Slack yesterday](https://dosomething.slack.com/archives/CUQMU4Q6B/p1597084491013900?thread_ts=1597084133.013400&cid=CUQMU4Q6B), seems like this should be ok to merge and deploy without a Fastly cache change. When we eventually delete all instances of the `state` location/filter the Group type/queries, that's when we'd need to clear the cache.

PR #2327, opened 2020-08-12
Only sets fixed class when on a campaign page without a groupType

lindsayrmaher:
### What's this PR do?

This pull request adds a check for whether we're on a groups campaign before making the signup button sticky on mobile screens.

👀 

**Before:**
Screen Shot 2020-08-12 at 11 16 26 AM

**After:**
Screen Shot 2020-08-12 at 11 15 05 AM

This is the second PR to clean up the mobile experience for users on groups campaigns. ([see slack thread](https://dosomething.slack.com/archives/CTVPG6L4R/p1596132412220100))

### Relevant tickets

References [Pivotal #174164406](https://www.pivotaltracker.com/story/show/174164406).

### Checklist

### DIFF
--- a/resources/assets/components/CampaignBanner/CampaignBanner.js
+++ b/resources/assets/components/CampaignBanner/CampaignBanner.js
@@ -111,8 +111,8 @@ const CampaignBanner = ({
               <div
                 data-testid="campaign-banner-signup-button"
                 className={classnames(
-                  'bg-white bottom-0 md:bottom-auto left-0 md:left-auto p-3 md:p-0 w-full md:w-auto z-10 md:z-auto',
-                  { 'fixed md:static': !groupType, static: groupType },
+                  'bg-white bottom-0 md:bottom-auto left-0 md:left-auto p-3 md:static md:p-0 w-full md:w-auto z-10 md:z-auto',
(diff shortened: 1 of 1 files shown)


resources/assets/components/CampaignBanner/CampaignBanner.js
```diff
@@ -109,7 +110,10 @@ const CampaignBanner = ({
             {!isAffiliated ? (
               <div
                 data-testid="campaign-banner-signup-button"
-                className="bg-white bottom-0 md:bottom-auto left-0 md:left-auto p-3 md:p-0 fixed md:static w-full md:w-auto z-10 md:z-auto"
+                className={classnames(
+                  'bg-white bottom-0 md:bottom-auto left-0 md:left-auto p-3 md:p-0 w-full md:w-auto z-10 md:z-auto',
+                  { 'fixed md:static': !groupType, static: groupType },
```
weerd, line 115: I think you just really need to override the `fixed` class and decide whether to output that or not... the `md:static` can just always be there, it's not a big deal if it duplicates since the default behavior is for the element to be static.

So can probs just use:

```
{'fixed': !groupType}
```

PR #2330, opened 2020-08-19
Filter ActionStatsBlock by school

aaronschachter:
### What's this PR do?

This pull request adds a school finder to the `ActionStatsBlock`, which will appear if a user selects a US state. It still needs to be styled, but I'll be adding them in a followup PR to help keep this one reviewable.

Some details on the changes:

* Refactors the `ActionStatsBlock` to use the `UsaStateSelect` component that both the school and group finders use, vs the `SelectLocationDropdown` component I initially used in hopes to eventually deprecate the `UsaStateSelect` component. The `UsaStateSelect` component style matches the `SchoolSelect`'s style because they both use [React Select](https://react-select.com/), which makes this easier to get out the door for next week, and also gives us some UX wins like filtering list of locations as user types, and having a button to clear selection (via the [isClearable](https://react-select.com/props#select-props) prop).

* Refactors the `UsaStateSelect` to return a location value instead of a state abbreviation, and refactors the `SchoolSelect` component to query schools by the `location` field instead of `state`, per https://github.com/DoSomething/graphql/pull/268.

👀 

Hoping to add test coverage in a a follow up PR as well, although ideally we'd want to mock different GraphQL results for the same query but different variables. I'm not sure if this is possible with [cypress-graphql-mock](https://github.com/tgriesser/cypress-graphql-mock), but haven't dug in too deep yet to see if it is.

### Relevant tickets

References [Pivotal #174301697](https://www.pivotaltracker.com/story/show/174301697).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/ActionStatsBlock/ActionStatsBlock.js
+++ b/resources/assets/components/blocks/ActionStatsBlock/ActionStatsBlock.js
@@ -4,7 +4,8 @@ import React, { useState } from 'react';
 
 import ActionStatsTable from './ActionStatsTable';
 import ActionStatsLeaderboard from './ActionStatsLeaderboard';
-import SelectLocationDropdown from '../../utilities/SelectLocationDropdown/SelectLocationDropdown';
+import SchoolSelect from '../CurrentSchoolBlock/SchoolSelect';
+import SchoolLocationSelect from '../../utilities/UsaStateSelect';
 
 export const ActionStatsBlockFragment = gql`
   fragment ActionStatsBlockFragment on ActionStatsBlock {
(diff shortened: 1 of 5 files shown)


resources/assets/components/blocks/CurrentSchoolBlock/SchoolSelect.js
```diff
@@ -11,39 +11,46 @@ import {
...
+}) => {
   const client = useApolloClient();
+
+  const schoolNotAvailableOption = includeSchoolNotAvailableOption
+    ? [
+        {
+          id: SCHOOL_NOT_AVAILABLE_SCHOOL_ID,
+        },
+      ]
+    : null;
```
mendelB, line 37: I forget, is this so we can store this on the user NS profile so we know that their school isn't available? ~When would we _not_ want to include this option?~ Ah from the `ActionStatsBlock`!

resources/assets/components/blocks/ActionStatsBlock/ActionStatsBlock.js
```diff
@@ -13,22 +14,36 @@ export const ActionStatsBlockFragment = gql`
...
-          locationList="domestic"
-          onSelect={event => setSchoolLocation(event.target.value)}
-          selectedOption={schoolLocation || ''}
-        />
+      <div className="flex pb-3">
+        <div className="md:w-1/4 pb-3">
+          <SchoolLocationSelect
+            isClearable
+            onChange={selected =>
+              setSchoolLocation(selected ? selected.value : null)
```
mendelB, line 29: This is minor, and potentially all wrong but if we flip this, would we prevent a state refresh / re-render or would it not make a difference since the value doesn't change when it's `null`?
```suggestion
              selected.value ? setSchoolLocation(selected.value) : null
```

resources/assets/components/blocks/CurrentSchoolBlock/SchoolSelect.js
aaronschachter: Exactly, when we ran TFJ we wanted users to be able to indicate they couldn't find their school in the school finder and save it to their profile...but there's really no use for selecting a "My school's not available" option for the leaderboard view.

resources/assets/components/blocks/ActionStatsBlock/ActionStatsBlock.js
aaronschachter: This is intentional -- we do want to set the `schoolLocation` to `null` in order to get the clearable "X" button working.

PR #2345, opened 2020-08-28
Current Club Block

mendelB:
### What's this PR do?

This pull request adds a new `CurrentClubBlock` component rendered via embedded Current Club Block Contentful entries.

The Block - a la the [`CurrentSchoolBlock`](https://github.com/DoSomething/phoenix-next/blob/96183a8c9acbc289fe4662ca33212dba10777a4b/resources/assets/components/blocks/CurrentSchoolBlock/CurrentSchoolBlock.js) - displays an authenticated user's current club if they've joined one, or acts as a form selector for them to find and join their club. It supports lazy authentication thanks to our in-house `useGate` hook @DFurnes added in https://github.com/DoSomething/phoenix-next/pull/1854.

### How should this be reviewed?
I'd recommend commit-by-commit. If it's helpful, keep the [`CurrentSchoolBlock`](https://github.com/DoSomething/phoenix-next/blob/96183a8c9acbc289fe4662ca33212dba10777a4b/resources/assets/components/blocks/CurrentSchoolBlock) directory handy for quick references for where we're following by example.

Notice any holes in the logic, or oddities? You'll probably notice areas rife for DRY-ness between this, the School block & the Groups block, which I'm sure @aaronschachter has more good thoughts on. 

I don't think we've documented our `useGate` hook functionality, so happy to chat through that flow!

The designs [are here](https://app.abstract.com/projects/fadf0790-f1f9-11e6-9a30-b54ba7e8e705/branches/75907f4b-8dff-48c9-a076-58df31f0273e/commits/latest/files/05f5faae-3c13-4947-b624-f9e0ea97822c/layers/0868842E-5A4B-4936-B046-FB26B098EDC1?collectionId=f7ec00df-516b-4e9d-9243-19e82d1ff5fc&collectionLayerId=de6acd75-c101-4805-b312-71309adbf9c7&mode=build).

### Any background context you want to provide?
Documentation & Tests to follow!

Just noting that if a user has an existing club, even if they visit the form unauthenticated and pick a different club, upon redirect, they'll still see they're original club since the block always attempts to find the existing one when it loads and withholds the form if it finds one, so the pending mutation won't run.

### Relevant tickets

References [Pivotal #174340202](https://www.pivotaltracker.com/story/show/174340202).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [x] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

[...]

PR #2347, opened 2020-08-31
Current Club Block Docs & Tests

mendelB:
### What's this PR do?

This pull request adds some documentation and Cypress tests for the new Current Club Block added in #2345 

👀 ([Docs on Gitbooks](https://dosomething.gitbook.io/phoenix-documentation/v/docs%2Fcurrent-club-block-+-tests/development/content-types/current-club-block)).

I had much trouble trying to test the 'lazy auth' flow for unauthenticated users.

[We have](https://github.com/DoSomething/phoenix-next/blob/17165454b0f30abc35a5ef724a22fe7c3a384855/cypress/support/commands.js#L81-L102) a `handleLogin` Cypress command, which _should_ take care of everything after adding the Cypress logic we use in our Redux [lazy auth](https://github.com/DoSomething/phoenix-next/blob/17165454b0f30abc35a5ef724a22fe7c3a384855/resources/assets/middleware/requiresAuthentication.js#L26-L31) to the [Hooks flow](https://github.com/DoSomething/phoenix-next/blob/17165454b0f30abc35a5ef724a22fe7c3a384855/resources/assets/helpers/auth.js#L112-L115). But for some reason, Cypress fails [to find](https://github.com/DoSomething/phoenix-next/blob/17165454b0f30abc35a5ef724a22fe7c3a384855/cypress/support/commands.js#L89) the `data-test="redirect"` in the DOM though it clearly seemed to be in there when I was attempting this.

In any case, I'm not positive how we'd mock the core functionality which is the `updateClubId` mutation running with the preselected and 'flashed' `clubId` in session storage, since I don't think our [Cypress GraphQL Mock](https://github.com/tgriesser/cypress-graphql-mock) is sophisticated enough to allow that? So I left this test as a TODO.

### Relevant tickets

References [Pivotal #174340202](https://www.pivotaltracker.com/story/show/174340202).

### Checklist


resources/assets/components/blocks/CurrentClubBlock/CurrentClubForm.js
```diff
@@ -46,17 +46,20 @@ const CurrentClubForm = () => {
 
   // If the mutation was successful, hide the form while the CurrentClubBlock
   // updates to display the user's current club.
-  if (data) {
+  if (isAuthenticated() && (data || flash.clubId)) {
```
katiecrane, line 49: Okay I see how if this is loading then the user must be authenticated - but when would we have only `data` or `flash.clubId`?

mendelB: Yeah this one escaped me on the first run! But if the user's doing lazy auth, there will be a couple of seconds potentially, when they've returned from Northstar and the mutation is gonna be triggered or is triggered and now loading, and we don't want to display the form -- and we'd catch this by seeing that there's a `flash.clubId`!

mendelB: Hmm perhaps a cooler UI would be to populate the form with their preselected ClubId and use the `loading` to make it look more like a loading form, but I think this is the quicker bet for now.

cypress/integration/current-club-block.js
```diff
@@ -0,0 +1,105 @@
...
+      cy.findByTestId('current-club').within(() => {
+        cy.findByTestId('current-club-name').contains(club.name);
+      });
+    });
+  });
+
+  context('Unauthenticated users', () => {
+    // @TODO: Fix this test to authenticate the user and see if the mutation runs with their selected club ID.
+    /** @test */
+    xit('Can join a club via "lazy authentication"', () => {
```
katiecrane, line 73: Does `xit` mean it doesn't run?

resources/assets/components/blocks/CurrentClubBlock/CurrentClubForm.js
katiecrane: Oooooo so before we have `data` back we need to be in a loading state and at that time we'll only have `flash.clubId`?

katiecrane:
>I had much trouble trying to test the 'lazy auth' flow for unauthenticated users.

This test sounds like a doozy! I think we should be sure to flag this as a specific flow to test when we're doing QA since we don't have Cypress support for it

resources/assets/components/blocks/CurrentClubBlock/CurrentClubForm.js
mendelB: Exactly! 👌

mendelB:
> > I had much trouble trying to test the 'lazy auth' flow for unauthenticated users.
> 
> This test sounds like a doozy! I think we should be sure to flag this as a specific flow to test when we're doing QA since we don't have Cypress support for it

Oooh good flag -- will do!

resources/assets/components/blocks/CurrentClubBlock/CurrentClubForm.js
katiecrane: Okay now I might be confusing myself - but doesn't `data` being truthy imply that we have a result back from the mutation and if that is so, won't this _always_ show the spinner even once we have a result? This also might just be Monday morning brain

mendelB: No, you're absolutely right and that's intentional! Once we have a result, we want the spinner to show until the `UserClubQuery` runs again and the `CurrentClubBlock` [swaps](https://github.com/DoSomething/phoenix-next/blob/17165454b0f30abc35a5ef724a22fe7c3a384855/resources/assets/components/blocks/CurrentClubBlock/CurrentClubBlock.js#L48-L70) the current club view in and gets rid of the form!

katiecrane: AH I was just missing the last step - thanks for walking me through it!!

PR #2350, opened 2020-09-03
Update SitewideBanner.js

Dinnall:
### What's this PR do?

This pull request checks for auth user and if the user is registered to vote, then we will display a refer a friend banner instead of VR banner.

### Relevant tickets

References [Pivotal #173796215](https://www.pivotaltracker.com/story/show/173796215).

Users with voter registration statuses for : "uncertain", "unregistered", and "confirmed".
Screen Shot 2020-09-04 at 1 21 46 PM

Users with voter registration statuses for: "registration complete" or "ineligible
Screen Shot 2020-09-04 at 1 20 54 PM

### Checklist

- [ x ] Added screenshots of front-end changes.

### DIFF
--- a/cypress/integration/sitewide-banner.js
+++ b/cypress/integration/sitewide-banner.js
@@ -130,7 +130,7 @@ describe('Site Wide Banner', () => {
     });
 
     cy.anonVisitCampaign(exampleCampaign);
-
+    cy.wait(2000);
     cy.findByTestId('sitewide-banner-hidden').should('have.length', 0);
   });
 
--- a/cypress/integration/sitewide-banner.js
+++ b/cypress/integration/sitewide-banner.js
@@ -122,6 +122,8 @@ describe('Site Wide Banner', () => {
 
   /** @test */
   it('The Site Wide Banner is not displayed on groups campaign pages', () => {
+    const user = userFactory();
+
     cy.mockGraphqlOp('CampaignSitewideBannerQuery', {
       campaign: {
         id: campaignId,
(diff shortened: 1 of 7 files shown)


mirie:
@Dinnall So I did some more sleuthing after dinner and I really couldn't find any problem with your code. I decided to check out this `package-lock.json` change and that seemed to be the problem. Now I'm not sure what exactly in the lock file update caused the analyticsWaypoint error - if you like, you can sort out that a bit more.

I also undid the `PropTypes.array,` changes because they seemed irrelevant to your original goal, but I could be wrong there. 

It looks like the CI issue is now resolved.

lindsayrmaher (approved):
these updates look good to me!

cypress/integration/sitewide-banner.js
```diff
@@ -73,34 +73,6 @@ describe('Site Wide Banner', () => {
...
-    cy.mockGraphqlOp('UserSitewideBannerQuery', {
-      user: {
-        voterRegistrationStatus: 'REGISTRATION_COMPLETE',
-      },
-    });
-
-    cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
-
-    cy.findByTestId('sitewide-banner-hidden').should('have.length', 1);
-  });
```
aaronschachter, line 102: Instead of deleting these tests completely, they should have been updated to verify the Refer Friends link is rendered within the banner.

Dinnall: ah - so just adding a check within these tests to see if the context matches?
- If yes, I think that would have been a good start, but the text for RAF changed, so what we would be checking wouldn't always match.

aaronschachter: We could check that the button renders the correct link when an authenticated user is ineligible or registration complete, similar to what we do in this test:

https://github.com/DoSomething/phoenix-next/blob/d301a678dee506c5017c1f00c0c6e09ae54a9820/cypress/integration/sitewide-banner.js#L176-L180

as well as check for the expected copy within the banner. This PR removes the existing tests for these status values completely (yes, because the functionality changes -- but it should be pretty straight forward to update the tests to check the banner content, instead of deleting them).

PR #2351, opened 2020-09-03
Fix issue building review apps.

DFurnes:
### What's this PR do?

This pull request fixes an issue building review apps for Phoenix.

I'd accidentally typo'd the environment variable that is provided by the [JawsDB Heroku add-on](https://elements.heroku.com/addons/jawsdb-maria) as `JAWS_MARIA_URL` (instead of `JAWSDB_MARIA_URL`) when [upgrading this application to Laravel 6.0](https://github.com/DoSomething/phoenix-next/commit/37ebe61e275c6ccaccd3f43c581edf66e2c8a339#diff-8a025ad5d4f4fade612727b1bb0e974f). This caused review apps to stop functioning.

I've corrected this to use the right environment variable name, removed a redundant `database` configuration line, and added some options that we use on our production database config for closer dev/prod parity. I've also removed the unused [ClearDB](https://elements.heroku.com/addons/cleardb) connection, since it was unclear at a glance which we actively used.

### How should this be reviewed?

I've broken up the changes into atomic commits. And hopefully the review app works now! :v:

### Any background context you want to provide?

This makes it easier to test changes before merging code in!

### Relevant tickets

References [Pivotal #174579215](https://www.pivotaltracker.com/story/show/174579215).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2352, opened 2020-09-04
Airtable Embeds in Company Pages

mendelB:
### What's this PR do?

This pull request adds `https://airtable.com/embed/...` as a valid URL for our Contentful Embed content type and `EmbedBlock` block in Phoenix. It also allows embedding an embed ( ♻️ ) in Company Page `content`.

👀 

https://dosomething-phoenix-de-pr-2352.herokuapp.com/us/about/who-we-are

The regex for the Embed `url` field is officially out of control now lol. We should def revisit for a better way to validate this on the editor side.

This was more rapid POC then official feature PR, but this is technically ready to ship! It would be nice to have documentation + tests for this though at some point.

One additional bit of work that could be helpful if we so choose would be to allow customizing the `height` of the `IframeEmbed` via props so that we could extend the height of say Airtable embeds if we want it to be more spacious.

### Relevant tickets

References [Pivotal #174201121](https://www.pivotaltracker.com/story/show/174201121).

### Checklist

image
image

### DIFF
--- a/contentful/content-types/companyPage.js
+++ b/contentful/content-types/companyPage.js
@@ -133,6 +133,7 @@ module.exports = function(migration) {
                 'actionStatsBlock',
                 'callToAction',
                 'contentBlock',
+                'embed',
                 'galleryBlock',
                 'imagesBlock',
                 'linkAction',
--- a/contentful/content-types/embed.js
+++ b/contentful/content-types/embed.js
@@ -24,12 +24,12 @@ module.exports = function(migration) {
       {
         regexp: {
           pattern:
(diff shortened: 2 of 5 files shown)


aaronschachter (approved):
>  We should def revisit for a better way to validate this on the editor side.

Indeed! Scary stuff 🎃

mendelB:
The review app broke again, though I'm _pretty sure_ I was up to date with #2351 😭 . Merging this so we can test against dev/qa. This is an isolated feature in any case, so worse comes to worse, we don't need to update the validation on Contentful `master` to allow Airtable if until this works.

PR #2353, opened 2020-09-09
Add click events to account nav tabs

Dinnall:

...

Adds tracking for user click events for each of the Account pages Nav items (tabs).

...

### Relevant tickets

References [Pivotal # 171523391](https://www.pivotaltracker.com/n/projects/2401401/stories/171523391).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Account/AccountNavigation.js
+++ b/resources/assets/components/pages/AccountPage/Account/AccountNavigation.js
@@ -22,20 +22,20 @@ const AccountNavigation = props => (
       <NavigationLink
         exact
         to="/us/account"
-        onClick={handleAccountNavTabClick('account')}
+        onClick={() => handleAccountNavTabClick('account')}
       >
         Account
       </NavigationLink>
       <NavigationLink
         to="/us/account/campaigns"
-        onClick={handleAccountNavTabClick('campaigns')}
+        onClick={() => handleAccountNavTabClick('campaigns')}
       >
         Campaigns
       </NavigationLink>
       {props.user.hasBadgesFlag ? (
         <NavigationLink
           to="/us/account/badges"
-          onClick={handleAccountNavTabClick('badges')}
+          onClick={() => handleAccountNavTabClick('badges')}
         >
           Badges
         </NavigationLink>
@@ -43,7 +43,7 @@ const AccountNavigation = props => (
       {featureFlag('volunteer_credits') ? (
         <NavigationLink
           to="/us/account/credits"
-          onClick={handleAccountNavTabClick('credits')}
+          onClick={() => handleAccountNavTabClick('credits')}
         >
           Credits
         </NavigationLink>
@@ -51,20 +51,20 @@ const AccountNavigation = props => (
       {featureFlag('cause_preferences') ? (
         <NavigationLink
           to="/us/account/interests"
-          onClick={handleAccountNavTabClick('interests')}
+          onClick={() => handleAccountNavTabClick('interests')}
         >
           Interests
         </NavigationLink>
       ) : null}
       <NavigationLink
         to="/us/account/subscriptions"
-        onClick={handleAccountNavTabClick('subscriptions')}
+        onClick={() => handleAccountNavTabClick('subscriptions')}
       >
         Subscriptions
       </NavigationLink>
       <NavigationLink
         to="/us/account/refer-friends"
-        onClick={handleAccountNavTabClick('refer-friends')}
+        onClick={() => handleAccountNavTabClick('refer-friends')}
       >
         Refer a Friend
       </NavigationLink>
--- a/resources/assets/components/pages/AccountPage/Account/AccountNavigation.js
+++ b/resources/assets/components/pages/AccountPage/Account/AccountNavigation.js
@@ -9,7 +9,7 @@ import {
 } from '../../../../helpers/analytics';
 
 const handleAccountNavTabClick = tabName => {
-  trackAnalyticsEvent(`clicked_account_navlink_${tabName}}`, {
+  trackAnalyticsEvent(`clicked_account_nav_link_${tabName}}`, {
     action: 'link_clicked',
     category: EVENT_CATEGORIES.navigation,
     label: `account_${tabName}`,
--- a/resources/assets/components/pages/AccountPage/Account/AccountNavigation.js
+++ b/resources/assets/components/pages/AccountPage/Account/AccountNavigation.js
@@ -3,27 +3,69 @@ import PropTypes from 'prop-types';
 
 import { featureFlag } from '../../../../helpers';
 import NavigationLink from '../../../utilities/NavigationLink/NavigationLink';
+import {
+  EVENT_CATEGORIES,
+  trackAnalyticsEvent,
+} from '../../../../helpers/analytics';
+
+const handleAccountNavTabClick = tabName => {
+  trackAnalyticsEvent(`clicked_account_navlink_${tabName}}`, {
+    action: 'link_clicked',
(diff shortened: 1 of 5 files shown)


mirie:
@Dinnall based on the issue I found in your other PR, your CI test failure complaint is a result of a dependency in the updated `package-lock.json`

lindsayrmaher:
@Dinnall sounds like we should revert the `package-lock.json` as mai did on #2350

Dinnall:
> @Dinnall sounds like we should revert the `package-lock.json` as mai did on #2350
sounds like a good plan. @mirie how did you revert the package.json file 🤔

resources/assets/components/pages/AccountPage/Account/AccountRoute.js
```diff
@@ -11,35 +11,68 @@ import UserPostsQuery from '../Campaigns/UserPostsQuery';
...
+    label: `account_${tabName}`,
+    context: {},
+  });
+};
 
 const AccountRoute = props => (
   <Switch>
     <Route
       path="/us/account/campaigns"
+      onClick={handleAccountNavTabClick}
```
katiecrane, line 32: I think this one is missing passing through the `tabName`

Dinnall: it was 😫 . I updated the PR.

resources/assets/components/pages/AccountPage/Account/AccountRoute.js
```diff
@@ -11,35 +11,68 @@ import UserPostsQuery from '../Campaigns/UserPostsQuery';
...
+  EVENT_CATEGORIES,
+  trackAnalyticsEvent,
+} from '../../../../helpers/analytics';
+
+const handleAccountNavTabClick = tabName => {
+  trackAnalyticsEvent(`clicked_account_navlink_${tabName}}`, {
+    action: 'link_clicked',
+    category: EVENT_CATEGORIES.navigation,
+    label: `account_${tabName}`,
+    context: {},
```
katiecrane, line 24: If we're not passing through any `context` I don't think we need to include this field

katiecrane: Ah good eye! Yeah we don't need it in this event or `CancelEmailSubscription` because like you said we omit it when we're not passing anything

mendelB:
@Dinnall Have you been able to trigger these new events locally? I'm wondering if these events need to be in the [`AccountNavigation`](https://github.com/DoSomething/phoenix-next/blob/d301a678dee506c5017c1f00c0c6e09ae54a9820/resources/assets/components/pages/AccountPage/Account/AccountNavigation.js) component on the `NavigationLink`s where the actual tab links are rendered, rather then the `Route` definitions which define the routes only.

katiecrane:
@Dinnall I just tested with the review app and I'm not sure why this is happening, but when I click on one of the tabs I see events fire as if I had clicked on ALL the tabs - are you seeing that happen locally?

Dinnall:
> @Dinnall Have you been able to trigger these new events locally? I'm wondering if these events need to be in the [`AccountNavigation`](https://github.com/DoSomething/phoenix-next/blob/d301a678dee506c5017c1f00c0c6e09ae54a9820/resources/assets/components/pages/AccountPage/Account/AccountNavigation.js) component on the `NavigationLink`s where the actual tab links are rendered, rather then the `Route` definitions which define the routes only.

This makes more sense, currently appears all of them are firing. Let me look into this

mendelB:
> @Dinnall I just tested with the review app and I'm not sure why this is happening, but when I click on one of the tabs I see events fire as if I had clicked on ALL the tabs - are you seeing that happen locally?

Hmm I think this is because we're invoking the `onClick` value as a function call `onClick={function()}` as opposed to passing the function as a reference `onClick={() => function()}`

Dinnall:
> @Dinnall I just tested with the review app and I'm not sure why this is happening, but when I click on one of the tabs I see events fire as if I had clicked on ALL the tabs - are you seeing that happen locally?

all are firing at once 🤔 , I just saw @mendelB suggestions of moving the handle click to `NavigationLinks `, instead of having them attached to the `Routes`.

Dinnall:
> @Dinnall Have you been able to trigger these new events locally? I'm wondering if these events need to be in the [`AccountNavigation`](https://github.com/DoSomething/phoenix-next/blob/d301a678dee506c5017c1f00c0c6e09ae54a9820/resources/assets/components/pages/AccountPage/Account/AccountNavigation.js) component on the `NavigationLink`s where the actual tab links are rendered, rather then the `Route` definitions which define the routes only.

Moving the handleClick to the `AccountNavigation` file doesn't stop firing all clicks at once. I also tested with not invoking the function immediately. For example `onClick={handleAccountNavTabClick('campaigns')}` vs `onClick={ ()=> handleAccountNavTabClick('campaigns')}`

cc @katiecrane

resources/assets/components/pages/AccountPage/Account/AccountNavigation.js
```diff
@@ -3,27 +3,69 @@ import PropTypes from 'prop-types';
...
+};
 
 const AccountNavigation = props => (
   <nav className="base-12-grid page-navigation py-3 md:py-6 -no-fade">
     <div className="grid-wide nav-items -mx-3">
-      <NavigationLink exact to="/us/account">
+      <NavigationLink
+        exact
+        to="/us/account"
+        onClick={handleAccountNavTabClick('account')}
```
mendelB, line 25: Per https://github.com/DoSomething/phoenix-next/pull/2353#issuecomment-692894379, I'm surprised that changing this to pass a function reference didn't resolve the issue, but I think we should make that change regardless (here and in the `onClick`s below) since otherwise the function would necessarily be invoked for each render.
```suggestion
        onClick={() => handleAccountNavTabClick('account')}
```

Dinnall: I went ahead and reverted the function back to having an arrow function. No change 🤔

mendelB: Looks like the preview app is still building based off the new changes, so we won't see the effect yet! Once it finishes building, I'm assuming the issue should be resolved, but... 🤞

Dinnall: The page is built now. Also figured out why more than one event was firing at once.I still have the events firing from the `AccountRoute` as well.

resources/assets/components/pages/AccountPage/Account/AccountNavigation.js
```diff
@@ -3,27 +3,69 @@ import PropTypes from 'prop-types';
 
 import { featureFlag } from '../../../../helpers';
 import NavigationLink from '../../../utilities/NavigationLink/NavigationLink';
+import {
+  EVENT_CATEGORIES,
+  trackAnalyticsEvent,
+} from '../../../../helpers/analytics';
+
+const handleAccountNavTabClick = tabName => {
+  trackAnalyticsEvent(`clicked_account_navlink_${tabName}}`, {
```
mendelB, line 12: I think `navlink` is meant to be separated as `nav_link`? The Pivotal card seems to have mistakenly formatted the surrounding underscores into italics, but on the events [spreadsheet](https://docs.google.com/spreadsheets/d/1iw5bqzRFdOTIW93UehsAZtrd5oZHCzmgoRcM3kBtpmM/edit#gid=118684449&range=95:95) it looks separated.

Dinnall: ah - Small change. Done

mendelB (approved):
🔬

PR #2355, opened 2020-09-10
Voter reg form scholarship modal

lindsayrmaher:
### What's this PR do?

This pull request adds a new modal option to the campaign banner. This is intended to be the initial modal users see before they take the quiz, but the regular scholarship details modal should display if they've dismissed. (that work will come in a follow up PR). 

Does this implementation make sense? Is the naming reasonable? 

**Screenshots:**
Screen Shot 2020-09-10 at 4 09 59 PM

Screen Shot 2020-09-10 at 4 09 44 PM

I struggled a bit with what would be the least repetitive and most efficient implementation option, so if you have opinions let me know!

### Relevant tickets

References [Pivotal #173229594](https://www.pivotaltracker.com/story/show/173229594).

### Checklist


resources/assets/helpers/index.js
```diff
@@ -1009,3 +1009,25 @@ export function getMillisecondsFromDays(days) {
...
+
+      return (
+        pathname.includes(pathWithoutAsterisk) &&
+        pathname.length > pathWithoutAsterisk.length
+      );
+    }
+
+    return path === pathname;
+  });
+};
```
katiecrane, line 1033: Love that you extracted this out into a helper!

cypress/integration/sitewide-banner.js
```diff
@@ -132,7 +132,7 @@ describe('Site Wide Banner', () => {
     });
 
     cy.authVisitCampaignWithoutSignup(user, exampleCampaign);
-    cy.findByTestId('sitewide-banner-hidden').should('have.length', 0);
+    cy.findByTestId('sitewide-banner-hidden').should('have.length', 1);
```
katiecrane, line 135: Was this previously testing the opposite of what it was supposed to?

resources/assets/components/blocks/ScholarshipReferralVoterRegistrationBlock/ScholarshipReferralVoterRegistrationBlock.js
```diff
@@ -0,0 +1,123 @@
...
+  affiliateSponsors,
+  attributes,
+  campaignId,
+  utmLabel,
+}) => {
+  const { loading, error, data } = useQuery(SCHOLARSHIP_AFFILIATE_QUERY, {
+    skip: utmLabel === null,
+    variables: {
+      utmLabel,
+      preview: env('CONTENTFUL_USE_PREVIEW_API'),
```
katiecrane, line 37: Oooo I haven't seen this before! This allows us to test updates to the affiliate in this context before making them live?

katiecrane: Like on the preview site I guess?

cypress/integration/sitewide-banner.js
lindsayrmaher: i don't think it was testing the opposite per say, but it was failing yesterday on one of Michael's PRs.

resources/assets/components/blocks/ScholarshipReferralVoterRegistrationBlock/ScholarshipReferralVoterRegistrationBlock.js
lindsayrmaher: yeah i think it's saying whether we want to be able to preview this particular affiliate?

resources/assets/components/blocks/ScholarshipReferralVoterRegistrationBlock/ScholarshipReferralVoterRegistrationBlock.js
```diff
@@ -0,0 +1,123 @@
...
+          {isLoaded ? (
+            <strong className="text-lg">
+              Welcome
+              {affiliateTitle
+                ? ` from ${affiliateTitle.toUpperCase()}`
+                : ' to DoSomething.org!'}
+              !
+            </strong>
+          ) : (
+            <PlaceholderText size="large" />
```
katiecrane, line 83: What's this `<PlaceholderText>` component do? I don't think I've run into this one before

resources/assets/components/blocks/ScholarshipReferralVoterRegistrationBlock/ScholarshipReferralVoterRegistrationBlock.js
```diff
@@ -0,0 +1,123 @@
...
+          )}{' '}
+          <Fragment>
+            <p>
+              Before you take the quiz, make sure you’re registered to vote!
+              Voting is one of the most impactful ways to make a difference on
+              the issues that matter to you.
```
katiecrane, line 89: I'm kind of concerned that this copy makes it sounds like you HAVE to register to vote before you're able to complete the quiz and be entered to win the scholarship?

cypress/integration/sitewide-banner.js
katiecrane: Isn't this switching from testing if that test id is showing up vs. is not?

resources/assets/components/blocks/ScholarshipReferralVoterRegistrationBlock/ScholarshipReferralVoterRegistrationBlock.js
katiecrane: I know you didn't write this copy though - should I comment on the Pivotal card?

cypress/integration/sitewide-banner.js
lindsayrmaher: yes, but what's tricky is the `sitewide-banner-hidden` testId is added if the banner is hidden, otherwise it is set to `sitewide-banner`. so the length should be 1, even though nothing shows up on the page. if that makes sense.

resources/assets/components/blocks/ScholarshipReferralVoterRegistrationBlock/ScholarshipReferralVoterRegistrationBlock.js
lindsayrmaher: It loads a gray rectangle where the title text should be. It's essentially an alternative to the spinner while the graphql query loads.

resources/assets/components/blocks/ScholarshipReferralVoterRegistrationBlock/ScholarshipReferralVoterRegistrationBlock.js
lindsayrmaher: Yeah i think that would be helpful! I agree re: the implications the wording has.

resources/assets/components/blocks/ScholarshipReferralVoterRegistrationBlock/ScholarshipReferralVoterRegistrationBlock.js
katiecrane: That's super cool!!

cypress/integration/sitewide-banner.js
katiecrane: OH got it so this WAS asserting the wrong thing when it had `0` instead of `1`?

lindsayrmaher: that's what i'm guessing yah!

katiecrane: gotcha!

katiecrane (approved):
This is good to go once we resolve the question about the copy! Happy to take another look at that point if you'd like or you can feel free to merge

resources/assets/helpers/index.js
```diff
@@ -1009,3 +1009,25 @@ export function getMillisecondsFromDays(days) {
...
+/**
+ * Checks if given pathname matches an entry in a paths config.
+ *
+ * @param {Array} paths
+ * @param {String} pathname
+ * @return {Boolean}
+ */
+export const isExcludedPath = (paths, pathname) => {
+  return paths.find(path => {
+    if (path.includes('*')) {
```
aaronschachter, line 1022: Now that this lives in a separate helper (which I agree with @katiecrane is 👍 ), I think this function name seems a little less readable than it could be. It's checking whether the pathname matches one of the patterns in paths -- but it's the other components that call it which use it or the sake of excluding. One day we might want to use it for including something (vs excluding it) based on a path.

What do you think about refactoring to something like `isCurrentPathInPaths(paths)`, and DRY passing the ` window.location.pathname` as an argument?


```suggestion
/**
 * Checks if current path matches an item in given paths array.
 *
 * @param {Array} paths
 * @return {Boolean}
 */
export const isCurrentPathInPaths = (paths) => {
  const pathname = window.location.pathname;
    
  return paths.find(path => {
    if (path.includes('*')) {
```

resources/assets/components/blocks/ScholarshipReferralVoterRegistrationBlock/ScholarshipReferralVoterRegistrationBlock.js
```diff
@@ -0,0 +1,123 @@
...
+import PlaceholderText from '../../utilities/PlaceholderText/PlaceholderText';
+import StartVoterRegistrationForm from '../../utilities/StartVoterRegistrationForm/StartVoterRegistrationForm';
+
+import '../ScholarshipInfoBlock/scholarshipInfoBlock.scss';
+
+/**
+ * The GraphQL query to load data for this component.
+ */
+const SCHOLARSHIP_AFFILIATE_QUERY = gql`
+  query ScholarshipAffiliateQuery($utmLabel: String!, $preview: Boolean!) {
```
aaronschachter, line 20: This query name is already used in the [ScholarshipInfoBlock](https://github.com/DoSomething/phoenix-next/blob/d03dbcbb69fea84aa3b91dbba146736729a9c8e9/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js#L27) -- it needs to be unique to avoid breaking our [schema fetching for cypress tests.](https://dosomething.gitbook.io/phoenix-documentation/installation-+-usage/usage#cypress)

What do you think about creating a new utility component to DRY this GraphQL request and the conditional rendering of the title per `isLoading`,`PlaceholderText` etc?

aaronschachter (changes requested):

resources/assets/helpers/index.js
lindsayrmaher: agree on the name change! and like the suggestion for setting the pathname var inside the helper

aaronschachter (approved):
Looking good! Just a few cleanup requests for line breaks that I missed during the first pass. It'd be helpful to see an example URL in the review app to help spot check functionality (or add some tests in a follow up PR...)

PR #2358, opened 2020-09-18
Adds optional CivicEngine Voting Widget embed to ContentBlock

aaronschachter:
### What's this PR do?

This pull request adds support to embed a [CivicEngine Voting Widget](https://developers.civicengine.com/docs/widget/) in a `ContentBlock` by setting a `footerType` property to `VotingWidgetBlock` in the `additionalContent` JSON field. 

This voting widget will only be used for voter registration -- it seemed a bit overkill to create a new content type in Contentful and go through the workflow of adding it to GraphQL and our ContentfulEntryLoader/ContentfulEntry components. We have a similar feature request for this sprint that @lindsayrmaher is working on, for a "Get Out The Vote" block type that will only be used for a one-off campaign -- that could be another `footerType` value supported (e.g. `footerType: 'GetOutTheVoteBlock'`) 

I'll add more info about the `VotingWidgetBlock` to documentation in a follow-up PR if this all looks 👍 

Mobile:
Screen Shot 2020-09-18 at 2 06 39 PM

Desktop:
Screen Shot 2020-09-18 at 2 06 27 PM

### How should this be reviewed?

👀 

### Any background context you want to provide?

I followed the "Silver Level" instructions for embedding the Voting Widget, which supports auto-resizing, and sending along a UTM parameter. When I'm logged into the DS site in Chrome, the widget will fail to load locally. There's a console error 

> DevTools failed to load SourceMap: Could not parse content for https://app.[removed].civicengine.com/iframeResizer.map: Unexpected token < in JSON at position 0

It works correctly incognito in Chrome locally, as well as Safari (anonymous, and authenticated). I sent an email to the CivicEngine support to ask about this issue, as it seems to be on their end, and related to https://github.com/davidjbradshaw/iframe-resizer/issues/136.

### Relevant tickets

References [Pivotal #174199686](https://www.pivotaltracker.com/story/show/174199686).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Documentation added for new features/changed endpoints.
- [x] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2359, opened 2020-09-21
Gotv Block Initial Structure

lindsayrmaher:
### What's this PR do?

This pull request creates the basic structure for the `GetOutTheVoteBlock` we'll be featuring in a GOTV campaign. I'm using the same additional content method of rendering that was implemented in #2358 and will be putting up another PR to clean up stying and add in the toggle functionality featured in the mocks.

Does the structure seem ok? Please excuse the horrid styling and spacing

Screen Shot 2020-09-22 at 10 52 21 AM

Right now, this form doesn't submit or save anything. And that's on purpose for the moment. We're anticipating this to be a one and done action for users, and that they won't need to return to the page to see their friends saved. But because that is a possibility, there is another ticket to save this information to local storage, so that we can compromise on saving unnecessary data. 

### Relevant tickets

References [Pivotal #174219622](https://www.pivotaltracker.com/story/show/174219622).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/ContentBlock/ContentBlockFooter.js
+++ b/resources/assets/components/blocks/ContentBlock/ContentBlockFooter.js
@@ -2,11 +2,14 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import VoterWidgetBlock from '../VoterWidgetBlock/VoterWidgetBlock';
+import GetOutTheVoteBlock from '../GetOutTheVoteBlock/GetOutTheVoteBlock';
 
 const ContentBlockFooter = ({ type }) => {
   switch (type) {
     case 'VoterWidgetBlock':
       return <VoterWidgetBlock />;
+    case 'GetOutTheVoteBlock':
+      return <GetOutTheVoteBlock />;
 
     default:
       return null;
--- a/resources/assets/components/blocks/ContentBlock/ContentBlockFooter.js
+++ b/resources/assets/components/blocks/ContentBlock/ContentBlockFooter.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import ErrorBlock from '../ErrorBlock/ErrorBlock';
 import VoterWidgetBlock from '../VoterWidgetBlock/VoterWidgetBlock';
 import GetOutTheVoteBlock from '../GetOutTheVoteBlock/GetOutTheVoteBlock';
 
@@ -12,7 +13,7 @@ const ContentBlockFooter = ({ type }) => {
       return <GetOutTheVoteBlock />;
 
     default:
-      return null;
+      return <ErrorBlock error={`ContentBlock is unable to render ${type}.`} />;
   }
 };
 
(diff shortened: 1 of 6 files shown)


resources/assets/components/blocks/GetOutTheVoteBlock/GetOutTheVoteBlock.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+
+import Embed from '../../utilities/Embed/Embed';
+import WhoAreYourPeopleForm from './WhoAreYourPeopleForm';
+
+const GetOutTheVoteBlock = () => {
+  return (
+    <>
```
mendelB, line 8: What's neat is that as long as a component has a single top-level element being rendered, we don't need to wrap it in a fragment, so I think we can ditch this one and the one in the people form!

resources/assets/components/blocks/GetOutTheVoteBlock/WhoAreYourPeopleForm.js
```diff
@@ -0,0 +1,71 @@
...
+              className="border border-dashed rounded border-gray-600 p-3"
+              type="text"
+              placeholder="Friend's First Name"
+            />
+          </div>
+
+          <div className="w-2/3 flex">
+            <p className="w-1/3">Voting in-person</p>
+
+            <label className="w-1/3" htmlFor="voting-location">
```
mendelB, line 19: Any reason we surround these inputs with `label`s since we don't actually label them? Maybe worth imitating the VR form where we also used `aria-label`s for accessibility.

https://github.com/DoSomething/phoenix-next/blob/9afc0b112606507ec293752173cce2b9931cc854/resources/assets/components/utilities/StartVoterRegistrationForm/StartVoterRegistrationForm.js#L71-L83

resources/assets/components/blocks/ContentBlock/ContentBlockFooter.js
```diff
@@ -2,11 +2,14 @@ import React from 'react';
...
 
 const ContentBlockFooter = ({ type }) => {
   switch (type) {
     case 'VoterWidgetBlock':
       return <VoterWidgetBlock />;
+    case 'GetOutTheVoteBlock':
+      return <GetOutTheVoteBlock />;
 
     default:
       return null;
```
mendelB, line 15: I know this isn't a new change, but maybe worth rendering an error block here which could be super helpful if we're debugging or typo the value?

mendelB (approved):
Left some minor comments, but the structure seems great to me!

mendelB:
I know the styling isn't quite done but could you happen to have screenshots for medium / small viewports handy? I can't seem to find this via the review app

resources/assets/components/blocks/GetOutTheVoteBlock/GetOutTheVoteBlock.js
lindsayrmaher: ah good call, i'll update!

lindsayrmaher:
> I know the styling isn't quite done but would you happen to have screenshots for medium / small viewports handy? I can't seem to find this via the review app

Screen Shot 2020-09-22 at 1 42 10 PM

Screen Shot 2020-09-22 at 1 42 21 PM

PR #2360, opened 2020-09-22
Generate updated migration for contentful Embed

mendelB:
Updates validation messages and fixes the URL validation regex

### What's this PR do?

This pull request is a quick addendum to #2352 to update the contentful migration script.
🙈 

### Relevant tickets
https://dosomething.slack.com/archives/C09ANFQLA/p1600804456032400

### DIFF
--- a/contentful/content-types/embed.js
+++ b/contentful/content-types/embed.js
@@ -24,7 +24,7 @@ module.exports = function(migration) {
       {
         regexp: {
           pattern:
-            '^(https:\\/\\/dosomething\\\\.(?:(?:carto\\\\.com\\/builder\\/[a-z0-9-]+\\/embed)|(?:typeform\\.com\\/to\\/[a-zA-Z0-9-]+)))|(https:\\/\\/airtable.com\\/embed\\/[a-zA-Z0-9-\\/]+)\\/?',
+            '^(https:\\/\\/dosomething\\.(?:(?:carto\\.com\\/builder\\/[a-z0-9-]+\\/embed)|(?:typeform\\.com\\/to\\/[a-zA-Z0-9-]+)))|(https:\\/\\/airtable.com\\/embed\\/[a-zA-Z0-9-\\/]+)\\/?',
           flags: null,
         },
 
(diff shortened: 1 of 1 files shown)


Vmack (approved):
sorry I'm late! This looks good Mendel!

contentful/content-types/embed.js
```diff
@@ -24,7 +24,7 @@ module.exports = function(migration) {
       {
         regexp: {
           pattern:
-            '^(https:\\/\\/dosomething\\\\.(?:(?:carto\\\\.com\\/builder\\/[a-z0-9-]+\\/embed)|(?:typeform\\.com\\/to\\/[a-zA-Z0-9-]+)))|(https:\\/\\/airtable.com\\/embed\\/[a-zA-Z0-9-\\/]+)\\/?',
+            '^(https:\\/\\/dosomething\\.(?:(?:carto\\.com\\/builder\\/[a-z0-9-]+\\/embed)|(?:typeform\\.com\\/to\\/[a-zA-Z0-9-]+)))|(https:\\/\\/airtable.com\\/embed\\/[a-zA-Z0-9-\\/]+)\\/?',
```
aaronschachter, line 27: What's the rationale for restricting the URL values here?

mendelB: This is actually a fantastic question. I _think_ originally we weren't filtering the URL's in Phoenix, so this was our line of defense. Otherwise, this might just be editor convenience to ensure noones attempting to work with a non supported URL. But now that we [filter in Phoenix](https://github.com/DoSomething/phoenix-next/blob/6b8ab8f53c0f67db2a410bdd87ccae5fa6ceb6b7/resources/assets/components/blocks/EmbedBlock/EmbedBlock.js#L10-L14), plus we can just use the field description to list the supported URLs, we can ditch this Regex!

aaronschachter: What's the point of restricting URL's in general though? To avoid fielding questions about why it looks funny, etc?

mendelB: I suppose precisely that. Plus it's an embed, so I feel like filtering is a good idea in general? There's an argument to be made that giving editorial freedom to embed whatever is more convenient, but I'd figure the technical restriction and guidance here is ultimately safer and _more_ convenient. (That's at least my hunch here).

PR #2362, opened 2020-09-23
Changes CivicEngine voting widget embed type to bronze level (iframe)

aaronschachter:
### What's this PR do?

~~This pull request refactors the configuration of the CivicEngine Voting Widget introduced in #2358 to support the three different embed types [documented](https://developers.civicengine.com/docs/widget/). I still get errors using Gold/Silver embed types, which link out to an external script -- but can successfully load with the Bronze embed type, which is just an iframe.~~ Update: removed support for gold/silver embed types which weren't working, for simplicity sake.

Ballot Ready emailed me back with a response that we should be able to send along the UTM source within the URL of the iframe in the bronze level, so we'll go with that for now.

* Does this code make sense? Readable?
* Does the widget load correctly? https://dosomething-phoenix-de-pr-2362.herokuapp.com/us/blocks/772nliHy7544iwCIT78x8

Ballot Ready is still trying to replicate the error we get when we load the widget via Gold/Silver embed types:

> DevTools failed to load SourceMap: Could not parse content for https://app.[removed].civicengine.com/iframeResizer.map: Unexpected token < in JSON at position 0

~~I don't think there's a harm in leaving this PR as-is, as it'd be great to use those embed types to get automatic resizing if we ever find out the source of the bug. They suggested it could potentially be something else in our source code interfering, but for now - let's use the bronze embed to move on to other stories.~~ Update: Per @mendelB's suggestion in this PR, we can get pretty close to automatic resizing by setting a width of 100%. Seems simple enough to commit to using the iframe / bronze level to simplify the logic in the code.

### Relevant tickets

References [Pivotal #174199686](https://www.pivotaltracker.com/story/show/174199686).

### Checklist


resources/assets/components/blocks/VoterWidgetBlock/VoterWidgetBlock.js
```diff
@@ -2,24 +2,60 @@ import React, { useEffect } from 'react';
...
+  }
+
+  const widget = (
+    <div className="civicengine-address" data-utm-source={utmSource} />
+  );
+
+  // If gold level, we've already added the required script tags into the <HEAD> element.
+  if (embedType === 'gold') {
+    return widget;
+  }
```
mendelB, line 41: Would using the [`useScript`](https://www.npmjs.com/package/@charlietango/use-script) hook here like we do for the [`SoftEdgeBlock`](https://github.com/DoSomething/phoenix-next/blob/46687a171f32b93973397c7ce25c3d895f416ce7/resources/assets/components/actions/SoftEdgeBlock.js#L46) work here? There's probably benefit to only loading that `<script>` if necessary instead of on every page load.

resources/assets/components/blocks/VoterWidgetBlock/VoterWidgetBlock.js
```diff
@@ -2,24 +2,60 @@ import React, { useEffect } from 'react';
...
+
+  // If silver level, we need to include the script tags.
+  // @see https://developers.civicengine.com/docs/widget/#silver-level-widget-instructions
```
mendelB, line 44: Why do we need to support the silver level if we can simply inject the script? (This might actually answer my question above in the gold level, with this being the method to only inject the script _when_necessary -- meaning we could actually ditch the _gold_ level in favor of this one? (and perhaps use `useScript` for this one?))

mendelB:
> Does the widget load correctly? https://dosomething-phoenix-de-pr-2362.herokuapp.com/us/blocks/772nliHy7544iwCIT78x8

It's not loading for me on the preview app:
image

resources/assets/components/blocks/VoterWidgetBlock/VoterWidgetBlock.js
```diff
@@ -2,24 +2,60 @@ import React, { useEffect } from 'react';
...
+   * Although the docs claim bronze level does not support UTM parameters, Ballot Ready says
+   * this should still send through the UTM source.
+   * @see https://developers.civicengine.com/docs/widget/#bronze-level-widget-instructions
+   */
+  if (embedType === 'bronze') {
+    return (
+      <iframe
+        src={`${baseUri}w/address?utm_source=${utmSource}`}
+        title="Voting Widget"
+      />
```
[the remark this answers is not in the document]
aaronschachter: Yes! This addition makes me think it's worth just scrapping the Gold/Silver level embeds completely, it gets us close enough to automatic resizing - and simplifies the code significantly.

resources/assets/components/blocks/VoterWidgetBlock/VoterWidgetBlock.js
aaronschachter: Indeed, that's what I was thinking I was doing here with the `useEffect` but didn't do enough testing. Going to scrap gold/silver embeds per https://github.com/DoSomething/phoenix-next/pull/2362#discussion_r494015621.

aaronschachter:
Thanks for reviewing and testing, @mendelB ! Verified working on Chrome and Safari (both mobile/desktop)

https://dosomething-phoenix-de-pr-2362.herokuapp.com/us/blocks/772nliHy7544iwCIT78x8

config/site.php
```diff
@@ -16,9 +16,15 @@
 
     // These Group Type IDs have a 'chapter' group label instead of the default in Group Finders.
     'chapter_group_type_ids' => explode(',', env('DS_CHAPTER_GROUP_TYPE_IDS')),
+    /**
+     * The CivicEngine vars are used forVoting Widget configuration.
+     * @see https://developers.civicengine.com/docs/widget
+     */
+    'civic_engine_client_name' => env('CIVIC_ENGINE_CLIENT_NAME'),
+    'civic_engine_embed_type' => env('CIVIC_ENGINE_EMBED_TYPE'),
```
mendelB, line 24: 👾 
```suggestion
```

mendelB (approved):
🖼️

PR #2366, opened 2020-09-24
Add a '?chromeless' mode for embeds.

DFurnes:
### What's this PR do?

This pull request adds a `?chromeless` query string that hides the navigation, footer, and HolaBar _and_ forces all links to open in new tabs. This fixes two issues with the embed at the end of the voter registration flow:
1. We should no longer need to whitelist any links someone could follow from that content page, since any links on the page (whether via Markdown, a component like the gallery, etc.) will be forced to open in a new tab.
2. We won't have the awkwardness of having a second navigation chrome within the tiny Rock The Vote embed, so things will look more like they belong! This is also a clear way to indicate the "chromeless" mode so we don't start abusing this functionality...

Screen Shot 2020-09-24 at 2 33 28 PM

(I'd show the embed within the actual Rock The Vote page but it doesn't like our non-secure `http://phoenix.test` or the self-signed Homestead certificate on `https://phoenix.test` – boo, security!!! But this should work there too! 😜)

I'd recommend reviewing this [without whitespace](https://github.com/DoSomething/phoenix-next/pull/2366/files?w=1) since some of the indentation changes are distracting!

My goal here is to fix this issue with a priority flow in a way that allows us to work around this issue in the future without continuing to litter the codebase to references to specific one-off experiences.

### Relevant tickets

References [Pivotal #]().

### Checklist

### DIFF
--- a/resources/assets/components/App.js
+++ b/resources/assets/components/App.js
@@ -9,11 +9,11 @@ import { Router, Route, Switch } from 'react-router-dom';
 import graphqlClient from '../graphql';
 import ErrorPage from './pages/ErrorPage';
 import Modal from './utilities/Modal/Modal';
-import { env, featureFlag } from '../helpers';
 import { initializeStore } from '../store/store';
 import HomePage from './pages/HomePage/HomePage';
 import BlockPage from './pages/BlockPage/BlockPage';
 import CausePage from './pages/CausePage/CausePage';
+import { env, featureFlag, query } from '../helpers';
 import NewHomePage from './pages/HomePage/NewHomePage';
 import CompanyPage from './pages/CompanyPage/CompanyPage';
 import CampaignContainer from './Campaign/CampaignContainer';
@@ -34,10 +34,15 @@ import VoterRegistrationDrivePage from './pages/VoterRegistrationDrivePage/Voter
 
 const App = ({ store, history }) => {
   initializeStore(store);
+
   return (
     <ReduxProvider store={store}>
       <ErrorBoundary FallbackComponent={ErrorPage}>
         <ApolloProvider client={graphqlClient(env('GRAPHQL_URL'))}>
+          {query('chromeless') ? (
+            /* If we're in "chromeless" mode, open all links in new windows: */
+            <base target="_blank" />
+          ) : null}
           <DismissableElement
             name="sitewide_banner_call_to_action"
             daysToReRender={7}
--- a/resources/assets/components/SiteNavigation/SiteNavigation.js
+++ b/resources/assets/components/SiteNavigation/SiteNavigation.js
@@ -11,6 +11,7 @@ import SiteNavigationFeature from './SiteNavigationFeature';
 import CloseButton from '../artifacts/CloseButton/CloseButton';
 import ProfileIcon from '../artifacts/ProfileIcon/ProfileIcon';
 import DoSomethingLogo from '../utilities/DoSomethingLogo/DoSomethingLogo';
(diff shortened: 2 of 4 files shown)


resources/assets/components/App.js
```diff
@@ -34,10 +34,15 @@ import VoterRegistrationDrivePage from './pages/VoterRegistrationDrivePage/Voter
...
 const App = ({ store, history }) => {
   initializeStore(store);
+
   return (
     <ReduxProvider store={store}>
       <ErrorBoundary FallbackComponent={ErrorPage}>
         <ApolloProvider client={graphqlClient(env('GRAPHQL_URL'))}>
+          {query('chromeless') ? (
+            /* If we're in "chromeless" mode, open all links in new windows: */
+            <base target="_blank" />
```
DFurnes, line 44: This is a [pretty nifty HTML element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) that basically sets "defaults" for all links on a page! By setting `target="_blank"` here, it's basically like setting it on all the links we generate in JSX, Markdown, Blade templates, etc – without the tedious work of doing so!

resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -169,6 +170,11 @@ class SiteNavigation extends React.Component {
   };
 
   render() {
+    // Hide navigation if we're in "chromeless" mode, e.g. for an embed:
+    if (query('chromeless')) {
+      return null;
+    }
```
DFurnes, line 176: This (and the change to the footer & HolaBar, below) fixes this sorta awkward "double navigation" displayed on this embed. Instead, this embed will just show the "main" content of whatever page you're on!

Screen Shot 2020-09-24 at 2 47 41 PM

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
```diff
@@ -63,6 +63,11 @@ const SitewideBanner = props => {
     return null;
   }
 
+  // If we're in "chromeless" mode, e.g. an embed, hide this bar:
+  if (query('chromeless')) {
+    return null;
+  }
```
DFurnes, line 69: I thought it made sense to hide the HolaBar alongside other "site chrome" for `?chromeless` mode, mainly because we currently use the HolaBar to prompt voter registration... and if you just registered to vote then you certainly don't need to be prompted again!

resources/assets/components/SiteNavigation/SiteNavigation.js
DFurnes: For illustration, this is what this embed (with `&chromeless=1` appended to the URL) would look like now:

Screen Shot 2020-09-24 at 2 50 52 PM

resources/assets/components/utilities/SitewideBanner/SitewideBanner.js
ngjo: ooooh sweet, we had a ticket for this https://www.pivotaltracker.com/story/show/174779151 !

DFurnes: Synergy!! ⚡

katiecrane (approved):
Love this, such a clean way to handle these cases!

mendelB:
This is really cool, but 

> I'd recommend reviewing this without whitespace since some of the indentation changes are distracting!

just changed my life. I did _not_ know about this setting. what the heck 🤯

PR #2367, opened 2020-09-24
Update Contentful Embed migration script - remove regex validation

mendelB:
### What's this PR do?

This pull request removes the flabbergasting Regex validation from the Embed URL field on Contentful.

👀 

We [filter per hostname](https://github.com/DoSomething/phoenix-next/blob/8a854e19bf8602c81c2caae7c941451b29cbcc49/resources/assets/components/blocks/EmbedBlock/EmbedBlock.js#L10-L14) on our end, and at this point, maintaining this Regex validation has proven more trouble then it's worth.

### Relevant tickets
This (ongoing) discussion https://github.com/DoSomething/phoenix-next/pull/2360#discussion_r494444901, and note #2352.

### DIFF
--- a/contentful/content-types/embed.js
+++ b/contentful/content-types/embed.js
@@ -13,25 +13,13 @@ module.exports = function(migration) {
     .validations([])
     .disabled(false)
     .omitted(false);
-
   embed
     .createField('url')
     .name('URL')
     .type('Symbol')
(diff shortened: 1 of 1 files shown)


aaronschachter (approved):
👋

DFurnes (approved):

PR #2373, opened 2020-09-28
Rename as CivicEngineVoterWidget, remove PollLocation component

aaronschachter:
### What's this PR do?

This pull request renames the `RequestBallotBlock` as a `CivicEngineVoterWidget`, and updates docs to detail how the widget will function both as a ballot request form and a polling locator form based on the configuration controlled by TMC (details in the Pivotal story).

It also removes the `PollLocator` component, since we won't be using it for this election year. Removing it seems like the easiest way to avoid confusion, we can always restore the commit that deletes it should we want to revisit.

👀 

I manually changed the `footerType` in the `additionalContent` of our test dev `ContentBlock` entry -- will do the same in QA and prod once this is merged.

### Relevant tickets

References [Pivotal #174199686](https://www.pivotaltracker.com/story/show/174199686).

### Checklist


Dinnall (approved):
Looks good to go!

Questions, instead of removing the`PollLocator` component, couldn't we have just left it for reference. Instead of trying to find the commit to reintroduce it, if it was to be used again?

aaronschachter:
Thanks for reviewing! We generally prefer to remove code that we know we aren't using. It took me quite a bit of Slack coordination to learn that we currently don't use or plan to use the `PollLocator` this year because we're using a different polling location service. I updated the docs to link out to this PR in case we decide we ever do want to revisit (and also document when we did use, and what we switched to)

Dinnall:
> Thanks for reviewing! We generally prefer to remove code that we know we aren't using. It took me quite a bit of Slack coordination to learn that we currently don't use or plan to use the `PollLocator` this year because we're using a different polling location service. I updated the docs to link out to this PR in case we decide we ever do want to revisit (and also document when we did use, and what we switched to)

Noted, I did see the mention of the new polling service and the reference of the old one. Cleared up my question, thanks

PR #2385, opened 2020-10-08
Adds documentation for Badges experience

katiecrane:
### What's this PR do?

This pull request adds docs about badges! Though this has code in Northstar too, I opted to put these docs here with the rest of our Phoenix stuff, the same way we did for the Refer-A-Friend docs.

Do you have any questions while reading the docs that are not answered? Anything missing? Also, check out the rich diff!

✍️ 

### Relevant tickets

References [Pivotal #173686138](https://www.pivotaltracker.com/story/show/173686138).

### Checklist


docs/development/features/badges.md
```diff
@@ -0,0 +1,66 @@
+# Badges
+
+## Overview
+
+Badges are awarded to users based on actions they have taken. Currently, badges is still technically a test, with 70% of _new_ users opted-in to the badges experience, with some exceptions.
```
DFurnes, line 5: I felt like this second sentence read a little awkwardly with so many commas. What if we summarized a bit more high-level here and linked to the more detailed "Who Gets Badges" section below for specifics?

```suggestion
Badges are awarded to users based on actions they have taken. The "badges" experience is still a test, and only [a subset of new users](#who-gets-badges) are opted-in.
```

docs/development/features/badges.md
```diff
@@ -0,0 +1,66 @@
...
+
+The `DS_BADGES_TEST` feature flag in _Northstar_ is what toggles the currently running badge test on and off. _Phoenix_ also checks if a user is opted-in to the badges experiment before displaying any badge related content.
```
DFurnes, line 7: It might be worth further clarifying that, unlike many (all?) of our other feature flags, this enables the _test_ and not the feature itself. And (correct me if I'm wrong), disabling this flag will _not_ disable the feature for anyone who has been opted-in already.

docs/development/features/badges.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Who Gets Badges
+
+Only new users get the badge experience, but there are some exceptions. All of the following logic takes place in _Northstar_.
+
+The following users will NOT have a chance to be opted-in to the badges experience:
+
+-   Users who have `utm_source:clubs` in their `source_detail`
```
DFurnes, line 15: I'm not sure if this is outside the scope of this documentation, but it might be nice to explain why these folks are opted out! I'm assuming it's because we think they're more incentivized and so they might skew data?

DFurnes (approved):
Love a good doc! 📚

docs/development/features/badges.md
katiecrane: love this suggestion!!

docs/development/features/badges.md
katiecrane: VERY true - I will add in that distinction!

docs/development/features/badges.md
katiecrane: Excellent question, I omitted because I didn't remember the history here 😂 but good flag, I can certainly dig up and add that in

katiecrane: I had to ask in slack about this one, I couldn't find a reason! We'll see if anyone else remembers!

docs/development/features/badges.md
```diff
@@ -0,0 +1,66 @@
+# Badges
+
+## Overview
+
+Badges are awarded to users based on actions they have taken. The "badges" experience is still a test, and only [a subset of new users](#who-gets-badges) are opted-in.
+
+The `DS_BADGES_TEST` feature flag in _Northstar_ is what toggles the currently running badge _test_ on and off, it does _not_ toggle the badge feature itself. The badge feature is controlled by the `badges` feature flag that lives on a user in Northstar. When `DS_BADGES_TEST` is `true`, that means that some users in Northstar are getting `badges` added to their `feature_flags`. Users who have the badges feature flag will _always_ get the badges experience. _Phoenix_ checks if a user has the badges feature flag before displaying any badge related content.
```
katiecrane, line 7: @DFurnes I updated this section to explicitly outline how this bit works, but is it confusing now?

DFurnes: I think this reads clearly! 👍

DFurnes (approved):
Looooovely 📚!

PR #2382, opened 2020-10-07
Voter Widget style tweaks

aaronschachter:
### What's this PR do?

This pull request decreases the `height` value of the iframe used to render the Voter Widget, and adjusts its width to be the full width of its parent `ContentBlock`.

StoryPage example

### How should this be reviewed?

I've created a sample story page on dev, `/stories/voter-widget`, and will link to the review app upon building.

### Any background context you want to provide?

The addresses do get cut off with this height change, but seems better than having such a large vertical space in-between elements on the page. Anything smaller than this height change will begin to cut off the bottom of the red CTA button.

Screen Shot 2020-10-07 at 1 24 02 PM

### Relevant tickets

References [Pivotal #175148561](https://www.pivotaltracker.com/story/show/175148561).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2386, opened 2020-10-08
Fixes Voter Widget width on campaign pages

aaronschachter:
### What's this PR do?

This pull request adds a check for the current URL to determine whether a `CivicEngineVoterWidget` is being rendered on a campaign page. If it is, render the parent `ContentBlock` at 2/3 instead of full width. Fixes a bug introduced in #2382 

Campaign page

👀 

I'm a little stuck on why the test for rendering a `GetOutTheVoteBlock` is failing:

> Error: Uncaught [Invariant Violation: Could not find "client" in the context or passed in as an option. Wrap the root component in an <ApolloProvider>, or pass an ApolloClient instance in via options.]

I've looked in https://github.com/apollographql/react-apollo/issues/2900 a bit but no luck, our versions seem okay. Another error occurs too - coming from the `GetOutTheVoteBlock`. I don't see this error though when inspecting the console on [a campaign page](https://qa.dosomething.org/us/campaigns/your-plan-your-vote/action).

>   Warning: Failed prop type: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`.

Because this bug is blocking a deploy, I think it's fine to leave fixing the test for later.

### Relevant tickets

References [Pivotal #]().

### Checklist

### DIFF
--- a/resources/assets/components/blocks/ContentBlock/ContentBlock.js
+++ b/resources/assets/components/blocks/ContentBlock/ContentBlock.js
@@ -39,7 +39,7 @@ const ContentBlock = ({
   const { footerType } = additionalContent;
   const fullWidthFooter =
     footerType === 'GetOutTheVoteBlock' ||
-    // Only display CivicEngineVoterWidget as full width when not rendered in a campaign page.
+    // HACK: Only display CivicEngineVoterWidget as full width when not rendered in a campaign page.
     (footerType === 'CivicEngineVoterWidget' &&
       !window.location.pathname.includes('/us/campaigns/'));
 
--- a/resources/assets/components/blocks/ContentBlock/ContentBlock.js
+++ b/resources/assets/components/blocks/ContentBlock/ContentBlock.js
@@ -37,9 +37,9 @@ const ContentBlock = ({
 }) => {
   const contentNode = content ? <TextContent>{content}</TextContent> : null;
   const { footerType } = additionalContent;
-  // The CivicEngineVoterWidget should only be full width if not rendered on a campaign page.
   const fullWidthFooter =
     footerType === 'GetOutTheVoteBlock' ||
+    // Only display CivicEngineVoterWidget as full width when not rendered in a campaign page.
     (footerType === 'CivicEngineVoterWidget' &&
       !window.location.pathname.includes('/us/campaigns/'));
 
--- a/resources/assets/components/blocks/ContentBlock/ContentBlock.js
+++ b/resources/assets/components/blocks/ContentBlock/ContentBlock.js
@@ -37,6 +37,11 @@ const ContentBlock = ({
 }) => {
   const contentNode = content ? <TextContent>{content}</TextContent> : null;
   const { footerType } = additionalContent;
+  // The CivicEngineVoterWidget should only be full width if not rendered on a campaign page.
+  const fullWidthFooter =
+    footerType === 'GetOutTheVoteBlock' ||
+    (footerType === 'CivicEngineVoterWidget' &&
+      !window.location.pathname.includes('/us/campaigns/'));
 
   return (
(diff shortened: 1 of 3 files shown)


aaronschachter:
Review app links:
* [Story page](https://dosomething-phoenix-de-pr-2386.herokuapp.com/us/stories/voter-widget) -- full width
* [Campaign page](https://dosomething-phoenix-de-pr-2386.herokuapp.com/us/campaigns/test-groups-campaign/voter-widget) -- 2/3 (must be authenticated)

resources/assets/components/blocks/ContentBlock/ContentBlock.test.js
```diff
@@ -68,11 +68,58 @@ describe('ContentBlock component', () => {
       );
     });
 
+    /** @test */
+    /*
+    // @TODO: Fix this failing test: Invariant Violation: Could not find "client" in the context or passed in as an option. Wrap the root component in an <ApolloProvider>, or pass an ApolloClient instance in via options.
```
mendelB, line 73: This is probably coming from the [GraphQL query](https://github.com/DoSomething/phoenix-next/blob/51d8409e3e2e10088d2d0b7cffd81aef69c87a29/resources/assets/components/utilities/Embed/Embed.js#L61) we're making in the [rendered `Embed` components](https://github.com/DoSomething/phoenix-next/blob/51d8409e3e2e10088d2d0b7cffd81aef69c87a29/resources/assets/components/blocks/GetOutTheVoteBlock/GetOutTheVoteBlock.js#L48-L52).

We might be able to get away with this using `shallow` to render the component, but with the transition to react testing libraries `render`, I think we need to wrap this in a `MockedProvider` like we do [here](https://github.com/DoSomething/phoenix-next/blob/51d8409e3e2e10088d2d0b7cffd81aef69c87a29/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.test.js#L46).

resources/assets/components/blocks/ContentBlock/ContentBlock.test.js
```diff
@@ -68,11 +68,58 @@ describe('ContentBlock component', () => {
...
+        'col-span-3',
+      );
+    });
+
+    /** @test */
+    test('displays across two-thirds of the row if footerType is CivicEngineVoterWidget and viewed on a campaign page', () => {
+      delete window.location;
+      window.location = new URL(
+        'https://dev.dosomething.org/us/campaigns/test-campaign',
+      );
```
mendelB, line 102: I'm not positive that this is better practice, but I've been interacting with the `jsdom` to reconfigure the `location` (is that even the context for react testing library tests?)

https://github.com/DoSomething/phoenix-next/blob/51d8409e3e2e10088d2d0b7cffd81aef69c87a29/resources/assets/helpers/refer-friends.test.js#L30

resources/assets/components/blocks/ContentBlock/ContentBlock.js
```diff
@@ -37,6 +37,11 @@ const ContentBlock = ({
 }) => {
   const contentNode = content ? <TextContent>{content}</TextContent> : null;
   const { footerType } = additionalContent;
+  const fullWidthFooter =
+    footerType === 'GetOutTheVoteBlock' ||
+    // Only display CivicEngineVoterWidget as full width when not rendered in a campaign page.
```
mendelB, line 42: I've found it helpful (or maybe just plain gotten a kick out of) when we sort of call out the `Hack:` in the comments formally, e.g. https://github.com/DoSomething/phoenix-next/blob/51d8409e3e2e10088d2d0b7cffd81aef69c87a29/webpack.config.js#L39

mendelB (approved):
📏

PR #2380, opened 2020-10-02
Hide leaderboard Load More Button for now

aaronschachter:
### What's this PR do?

This pull request hides the load more button until we fix pagination in a Rogue `GET /action-stats` API request when a group type filter is set, which [causes an integrity constraint violation](https://github.com/DoSomething/rogue/pull/1125#issuecomment-702849789):

```
production.ERROR: SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous (SQL: select * from `action_stats` inner join `groups` on `action_stats`.`school_id` = `groups`.`school_id` where `action_id` = 954 and `groups`.`group_type_id` = 5 and (`impact` < 6 or (`impact` = 6 and `id` > 11955)) order by `impact` desc, `action_stats`.`id` asc limit 11 offset 0)
```

### How should this be reviewed?

👀 

### Any background context you want to provide?

🛺 

### Relevant tickets

References [Pivotal #175102031](https://www.pivotaltracker.com/story/show/175102031).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2396, opened 2020-10-15
Restores Load More button on ActionStatsTable, starts on SchoolSelect tests

aaronschachter:
### What's this PR do?

This pull request reverts #2380 (and is blocked from merging until fix, https://github.com/DoSomething/rogue/pull/1140, is deployed) to restore the "Load More" button in the ActionStatsTable component.

Also adds some cleanup: 
* Fixes warning for required `schoolLocation` prop in SchoolSelect component, and refactors it to simplify (instead of sending an additional `isDisabled` prop, disable the `AsyncSelect` if the `schoolLocation` prop is null)
* Moves the SchoolSelect component out of the blocks/CurrentSchoolBlock dir, and into a new utilities/SchoolSelect dir

And starts on tests. I'm stuck one of them for now (commented out in a TODO).

👀 

I'll keep digging a bit to try to find why this tests fails -- unsure if it's an Enzyme limitation. 

### Relevant tickets

References [Pivotal #175272211](https://www.pivotaltracker.com/story/show/175272211).

### Checklist


resources/assets/components/utilities/SchoolSelect/SchoolSelect.test.js
```diff
@@ -0,0 +1,69 @@
...
+
+      expect(asyncSelect.prop('isDisabled')).toBe(true);
+    });
+  });
+
+  /**
+  TODO: FIX ME
+
+  describe('With a schoolLocation prop', () => {
+    // This line fails with React.Children.only expected to receive a single React element child.
```
DFurnes, line 55: Strange! Like, it thinks `<SchoolSelect>` is not a single React element? 🤔 Even stranger that it works above...

DFurnes: I wonder if this is a result of mocking the `react-select/async` package above and returning a string. What if we returned a `<div>mock</div>` there instead? (And maybe it's not failing above since the query is disabled there?)

aaronschachter: Changing the mock to return `<div>mock</div>` returns:
```
 TypeError: Cannot read property 'jsx' of undefined

      3 | import { debounce } from 'lodash';
      4 | import PropTypes from 'prop-types';
    > 5 | import AsyncSelect from 'react-select/async';
```
`'<div>mock</div>'` returns:
```
  ● SchoolSelect component › Without a schoolLocation prop › encountered a declaration exception

    InvalidCharacterError: "<div>mock</div>" did not match the Name production: Unexpected syntax in top
    Line 1:
    <div>mock</div>
```

Somewhat a tangent, because ideally it'd be great to know how to fix this -- but I figured a jest test would work well for this component since it's used in two places... although it's only this `ActionStatsBlock` component that makes use of disabling the `AsyncSelect`, which could make the case for writing a Cypress tests for an ActionStatsBlock via blocks URL to test this new behavior (toggling disabled per `schoolLocation` prop). Buut because `AsyncSelect` returns div's instead of an input -- I think it'd be tougher to test via Cypress test that it's disabled.

DFurnes: Huh! The `cannot read property 'jsx' of undefined` would make me think that `React` isn't being imported (since I believe that's referencing that it can't read `React.jsx()`, which is what those JSX tags are transformed into). But we're importing it at the top & using JSX just fine elsewhere in this same file... 🤔 

The other `InvalidCharacterError` error you mentioned here does make me think that JSX transform isn't running, which is maybe something special with Jest (and perhaps due to how `jest.mock` is a bit more special than other code that runs "within" the test file...?).

mendelB: Hmmm I recall having trouble with this -- and don't remember the details, but something like this seemed to work
https://github.com/DoSomething/phoenix-next/blob/156d20ce975355231c29c9e43ab3300b0a6329ac/resources/assets/components/pages/AccountPage/Credits/VolunteerCreditsQuery.test.js#L15-L17

aaronschachter: I'm experimenting with something -- I think creating a `MockedProvider` twice within the file is the problem...

DFurnes: @mendelB Huh, from the GitHub issue linked in the comment there comment I would've assumed it should be `() => 'VolunteerCreditsTable'` (since it's not a HTML component, which I think is where the lower-case comes in). Thinking, for example, of how we handle [mocking Redux container components](https://github.com/DoSomething/phoenix-next/blob/865a8407c69902edf470a371393ac3d81f90c6f9/resources/assets/components/actions/ShareAction/ShareAction.test.js#L8)... but I see we seem to use both approaches successfully in different places. 🤔

mendelB: Yeah that's true! Hmm I _think_ I remember having trouble with `'VolunteerCreditsTable'` which is why I ended up with `'volunteer-credits-table'`. Not helpful lol, I'm admittedly guilty of sometimes just being so frustrated with the tests not working that I just brute force whatever works and call it a day.

aaronschachter: > I'm admittedly guilty of sometimes just being so frustrated with the tests not working that I just brute force whatever works and call it a day.

Speaking of which -- I've updated this branch to get tests to pass by by creating a separate test file for each `MockedProvider` variation --avoiding `React.Children.only expected to receive a single React element child` error that the second instance would throw (regardless of whether the `schoolLocation` prop was passed)

DFurnes (approved):
Looks great! 👍

PR #2404, opened 2020-10-20
Adds Documentation for Newsletter Popover

Vmack:
### What's this PR do?

This pull request adds documentation for the newsletter popover or `CtaPopoverEmailForm`

### How should this be reviewed?

👀 is this suffice?

### Any background context you want to provide?

N/A

### Relevant tickets

References [Pivotal #172682176](https://www.pivotaltracker.com/story/show/172682176).

### Checklist

- [ x ] This PR has been added to the relevant Pivotal card.
- [ x ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2406, opened 2020-10-20
Upgrade dependencies, including Cypress 5.

DFurnes:
### What's this PR do?

This pull request upgrades this application's dependencies, including Cypress 5.x! This is an important part of keeping this application up-to-date (and free of security issues!) and also gets us some neat new features (like automatic test retries, flaky test insights, and GitHub "status check" integration.

I broke this into individual commits for each step of the process. I ran into a lot of unexpected issues from running a simple `npm update` (it had been a while), and had to do some fixes to get the build back to "green" after that initial step.

There's some sort of issue with batched GraphQL queries that I still haven't sorted out. I think this is fine to merge as-is for now, and we can revisit what might be causing that issue later on. (I'm curious to see what the performance impact is, if any!)

🆙

### Relevant tickets

References [Pivotal #174581418](https://www.pivotaltracker.com/story/show/174581418).

### Checklist


cypress.json
```diff
@@ -1,6 +1,7 @@
 {
   "baseUrl": "http://phoenix.test",
   "projectId": "ayzqmy",
+  "retries": 3,
```
DFurnes, line 4: This instructs Cypress to retry tests 3 times if they fail. If a test fails & then passes from a retry, it'll be marked as "flaky" and tracked in their dashboard (we can then create tickets to fix these up in a cleanup sprint).

See Cypress's [Test Retries documentation](https://docs.cypress.io/guides/guides/test-retries.html) for more details.

cypress/integration/campaign-post.js
```diff
@@ -94,11 +94,7 @@ describe('Campaign Post', () => {
 
     cy.get('.photo-submission-action').within(() => {
       // Choose an image to upload as a photo post:
-      const fileName = 'upload.jpg';
-      cy.fixture(fileName).then(fileContent => {
-        const payload = { fileContent, fileName, mimeType: 'image/jpeg' };
-        cy.get('input[type="file"]').upload(payload, { subjectType: 'input' });
-      });
+      cy.get('input[type="file"]').attachFile('upload.jpg');
```
DFurnes, line 97: This is the _much_ nicer syntax in the newest version of the [cypress-file-upload](https://www.npmjs.com/package/cypress-file-upload) plugin!

cypress/plugins/index.js
```diff
@@ -4,6 +4,14 @@
...
+const webpackPreprocessor = require('@cypress/webpack-preprocessor');
+const defaults = webpackPreprocessor.defaultOptions;
+
+/**
+ * Use our Babel preset when compiling Cypress test files.
+ * @see https://git.io/JTRF1
+ */
+module.exports = on => {
+  delete defaults.webpackOptions.module.rules[0].use[0].options.presets;
+  on('file:preprocessor', webpackPreprocessor(defaults));
```
DFurnes, line 16: Because we rely on the `tailwind()` helper function in one of our tests, Cypress's built-in Babel configuration was choking on a language feature in our code (class properties) that they don't support out-of-the-box. This deletes their provided preset so that it will use what we define in [our own Babel configuration](https://github.com/DoSomething/phoenix-next/blob/bed543941ebf0828a808867d5adeac16f066b6d5/package.json#L33-L47).

resources/assets/components/utilities/AnalyticsWaypoint/AnalyticsWaypoint.test.js
```diff
@@ -1,32 +1,34 @@
 import React from 'react';
-import { mount } from 'enzyme';
+import { render, screen } from '@testing-library/react';
```
DFurnes, line 2: This test was failing after I ran `npm update`. I wasn't able to figure out _whyyyy_ that was the case, but swapping this to use [React Testing Library](https://testing-library.com/docs/react-testing-library/intro) (which we've started using on other new tests) seems to have solved it.

resources/assets/graphql.js
```diff
@@ -51,7 +51,7 @@ export default uri => {
 
   // Create the HTTP link! This is our terminating link that batches up
   // GraphQL queries and makes the actual HTTP request to our server.
-  const httpLink = new BatchHttpLink({ uri });
+  const httpLink = new HttpLink({ uri });
```
DFurnes, line 54: This is another spooky bug from the `npm update` that I was only able to whittle down by a lot of trial and error. For some reason, the `BatchHttpLink` causes any Cypress tests that would have batched requests to fail (real traffic is unaffected).

For now, I think it's fine to swap this to the normal HTTP link & make a follow-up ticket to investigate further.

DFurnes:
One of my favorite new features that I just want to shout out is the GitHub status check integration! Rather than having to dig into the CircleCI step to get the recording URL, you can just click through directly from the PR! 😌 

Screen Shot 2020-10-20 at 11 44 44 AM

resources/assets/graphql.js
DFurnes: Created [#175357934](https://www.pivotaltracker.com/story/show/175357934) as a follow-up ticket.

cypress/integration/campaign-post.js
katiecrane: 😱 wayyyyy better!

katiecrane (approved):
looks good! so excited for those automatic retries and tracking flaky tests!

DFurnes:
^ Rebased this to pull in latest changes... seeing some weird behavior on my local.

DFurnes:
Interestingly, I'm only seeing this issue on my local – perhaps because I'm running Chrome there, and we're using Electron when running the test suite in CI. Something to look into a bit later... it'd be nice to use a "real" browser across the board.

Anyways, while Cypress [automatically clears `localStorage` between tests](https://docs.cypress.io/api/commands/clearlocalstorage.html#Syntax), it [does _not_ automatically clear `sessionStorage`](https://git.io/JT0og) (which we use to keep track of UTMs throughout a session, amongst some other things). And so we would continue to see the `_scholarship_fastweb` source from [an earlier test](https://github.com/DoSomething/phoenix-next/blob/12096ae7400389a24663157bc57e0974405ee64d/cypress/integration/campaign-signup.js#L274) when we read it from `getUtms` later on!! 🙅 

I've updated our "on load" event to clear session storage in 2df7dfa so this should no longer be an issue.

cypress/support/index.js
```diff
@@ -12,8 +12,11 @@ import { defaultReferralCampaignId } from '../fixtures/constants';
 Cypress.on('window:before:load', window => {
   const document = window.document;
 
+  // Clear session storage between tests:
+  window.sessionStorage.clear();
+
   // Custom ENV variables for the testing environment.
-  window.ENV = {
+  window.window.ENV = {
```
katiecrane, line 19: is this supposed to have the double `window`?

DFurnes: It is NOT! Thanks for catching that! Fixed this in 23d7150. I was wondering why this worked at all, and I guess since the "window" is the [global object](https://developer.mozilla.org/en-US/docs/Glossary/Global_object), it does _technically_ seem to work... `window.window.window.window...` 😵 🌀

katiecrane (approved):
another approval, just for good measure!

cypress/support/index.js
katiecrane: oh no it can just go on forever 😳 😧

aaronschachter:
I'm wondering if this PR's package updates are what's caused the `Prop type `array` is forbidden` warnings to appear in console. From what I can tell in our `.eslintrc.js` -- they always should have been firing the warning?

https://github.com/DoSomething/phoenix-next/blob/af2bb9e12786c13e61bedc78262fc13d082d0d6a/.eslintrc.js#L14

DFurnes:
I would assume we've always been warning on these too! Is there a chance we'd just been ignoring them (since warnings won't fail the build), or did they only start popping up more recently? Looks like we bumped from 7.19.0 to 7.21.4 here, which does include [a handful of fixes to that linter](https://github.com/yannickcr/eslint-plugin-react/blob/master/CHANGELOG.md), so that could defs be the culprit...

aaronschachter:
I always 👀  on the logs, and haven't done much work in this repo for a bit -- so I do think they started popping up recently...

DFurnes:
Yeah, this was my first time back in here in a while too so I just assumed they were already there! I wonder if not...

DFurnes:
Hmm, yeah – looking back at [this build](https://app.circleci.com/pipelines/github/DoSomething/phoenix-next/4636/workflows/c49ed68a-4a03-40a3-8848-b1e538844496/jobs/4891) for #2404 before these changes got merged in & I don't see these warnings. So definitely new with this update! Maybe it's from this "warn on destructured values as well" change in [7.20.1](https://github.com/yannickcr/eslint-plugin-react/blob/master/CHANGELOG.md#7201---2020-06-28).

PR #2411, opened 2020-10-21
Gallery Block "Content Block" gallery type

mendelB:
### What's this PR do?

This pull request adds a new `CONTENT_BLOCK` gallery type to the `GalleryBlock`. We'll retire the legacy `EXTERNAL_LINK` which this is replacing once we deploy https://github.com/DoSomething/graphql/pull/291 which will convert legacy blocks to the new format for backward compatibility.

It also adds a very basic Gallery Block test to cover this backward compatibility. I wanted to add a more robust suite and some documentation, but alas the relentless waves of time.

### How should this be reviewed?
👀 

### Any background context you want to provide?
This has been a point of confusion for editors!

### Relevant tickets

References [Pivotal #174048432](https://www.pivotaltracker.com/story/show/174048432).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [x] Added appropriate feature/unit tests.

PR #2412, opened 2020-10-22
Retire Gallery Block "External Link" gallery type

mendelB:
### What's this PR do?

This pull request follows up on #2411 to retire the legacy `EXTERNAL_LINK` gallery type. It updates our schema per the new changes in GraphQL https://github.com/DoSomething/graphql/pull/291 and updates the test.

👀 

We kept this alive to maintain backward compatibility with existing entries, but that's being resolved via GraphQL now. (Plus we've updated existing entries on Contentful, so we'll be sunsetting the GraphQL resolution shortly!)

### Relevant tickets

References [Pivotal #174048432](https://www.pivotaltracker.com/story/show/174048432).

### Checklist

### DIFF
--- a/cypress/integration/gallery-block.js
+++ b/cypress/integration/gallery-block.js
@@ -4,36 +4,18 @@ import { userFactory } from '../fixtures/user';
 describe('Gallery Block', () => {
   beforeEach(() => cy.configureMocks());
 
-  const blockId = 'abcdefghi123456789';
-
   /** @test */
-  it.only('renders a ContentBlockGalleryItem component for "EXTERNAL_LINK" blocks', () => {
+  it.only('renders a ContentBlockGalleryItem component for "CONTENT_BLOCK" blocks', () => {
     cy.mockGraphqlOp('ContentfulBlockQuery', {
       block: {
         __typename: 'GalleryBlock',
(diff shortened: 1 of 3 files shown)


cypress/integration/gallery-block.js
```diff
@@ -4,36 +4,18 @@ import { userFactory } from '../fixtures/user';
...
       block: {
         __typename: 'GalleryBlock',
-        galleryType: 'EXTERNAL_LINK',
+        galleryType: 'CONTENT_BLOCK',
         blocks: [{ __typename: 'ContentBlock' }],
       },
     });
 
-    cy.visit(`us/blocks/${blockId}`);
+    cy.visit('us/blocks/abcdefghi123456789');
```
Dinnall, line 17: why was changed to a hardcoded id?

mendelB: We're only running one assertion here, so figured I could just inline it for now 🤷‍♂️ (the ID shouldn't actually matter since we're mocking the GraphQL request which would otherwise be using the ID to query for the block).

Dinnall: thanks for that insight

Dinnall (approved):

PR #2408, opened 2020-10-20
Refactoring Sitewide Banner to Include CTA popover rendering

lindsayrmaher:
### What's this PR do?

This pull request refactors the `SiteWideBanner` and `CompanyPage` to merge the checks for displaying the hellobar vs the ctaPopover on certain pages. This will make it easier for us to check that one is displayed over the other within the same file and test for all expected outcomes from the same cypress test suite.

The `CtaPopover` is also rendered within the `GeneralPage` component based on an `additionalContent` prop. This PR doesn't account for removing that yet, we're waiting to confirm what pathnames that relates to and can add those to our config file once a list is finalized. 

### How should this be reviewed?

I'd love some thoughts on the best way to approach the cypress tests. Should all tests related to Hellobar and CtaPopover be handled in the sitewide-banner test suite? Or should they continue to live in the files they started in?

Desktop:
Kapture 2020-10-21 at 11 29 05

Mobile:
Kapture 2020-10-21 at 11 30 30

### Any background context you want to provide?

We'll need to update the documentation from #2404 once we turn the feature flag off. I think this will be one of two PR's for this work, with the follow up documenting and possibly consolidating cypress test files. 

### Relevant tickets

References [Pivotal #175211087](https://www.pivotaltracker.com/story/show/175211087).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [x] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2413, opened 2020-10-22
Sitewide banner/Cta Popover Structure and Documentation Changes

lindsayrmaher:
### What's this PR do?

This pull request creates a new component called the `PopoverDispatcher` which replaces the original `SitewideBanner` component. This acts as the origin point for all types of popover elements, which are currently the SitewideBanner and the CtaPopover. This restructuring helps to clarify when elements are or aren't being displayed in conjunction with one another, and keeps track of all specific rendering needs in one config file.

This also creates documentation for the `PopoverDispatcher` and updates the `SitewideBanner` and `NewsletterPopover` documentation to reflect this change.

Is the documentation clear? And are we ok with the new file naming and structuring? Also was thinking more about the tests and wondering is it's ok to keep them as is right now vs creating a new popover file. I could also see the case of just re naming the sitewide-banner test file and adding in a test for the cta popover there on the `easy-scholarships` page.

This is part two of #2408 

### Relevant tickets

References [Pivotal #](https://www.pivotaltracker.com/story/show/175211087).

### Checklist


resources/assets/components/App.js
```diff
@@ -48,7 +48,7 @@ const App = ({ store, history }) => {
             daysToReRender={7}
             context={{ contextSource: 'voter_registration' }}
             render={(handleClose, handleComplete) => (
-              <SitewideBanner
+              <PopoverDispatcher
                 cta="Get Started"
                 description="Make your voice heard. Register to vote in less than 2 minutes."
```
aaronschachter, line 53: Thinking ahead to when the [default text to display will get more complicated](https://www.pivotaltracker.com/n/projects/2401401/stories/175406026), what do you think about refactoring the `PopoverDispatcher` to only expect `handleClose` and `handleComplete` props, and defining the `cta`, `description`, and `link` props for the default SitewideBanner [inline](https://github.com/DoSomething/phoenix-next/blob/eb9b6223eb0f73b6b1af78d8f24f93c46bb00049/resources/assets/components/utilities/PopoverDispatcher/PopoverDispatcher.js#L173) when we return the default `SitewideBanner` if no other conditions match? 

I'm also considering now that we may need to store different storage variables to separate closing the SidewideBanner vs closing the Newsletter widget? (which also may suggest further refactoring of the props). It's tough to tell from the code whether or not this is the case, may need to manually test...

docs/development/features/sitewide-banner.md
```diff
@@ -6,22 +6,20 @@ This is a banner feature that is displayed site wide when turned on. In the past
...
 
 Before the banner renders on the page, we check the config for any pages that should be suppressed.
 
-If you want to suppress the banner from a page that may have multiple sub paths but share the same base path, you'll need to add a `*` to the end of the pathname.
-
 ## Displaying Sitewide banner content
 
-The content that will be displayed on the banner currently depends on a few variables. We check for users that have self reported either confirmed or uncertain when registering with us. If they have either as their voter registration status, they'll see a specific copy to encourage them to double check their status.
+The content that will be displayed on the banner currently depends on a few variables. We check for users that have self reported either `confirmed` or `uncertain` when registering with us. If they have either as their voter registration status, they'll see a specific copy to encourage them to double check their status.
```
aaronschachter, line 19: What do you think about this edit to hopefully keep a little more succinct?

The content that will be displayed on the banner currently depends on a few variables. If a user has a self reported voter registration status of either `confirmed` or `uncertain`, they'll see a specific copy to encourage them to double check their status.
```

File: docs/development/features/sitewide-banner.md [hunk 3]
```diff
@@ -6,22 +6,20 @@ This is a banner feature that is displayed site wide when turned on. In the past
...
 
-Besides self reporting, we check if the user is authenticated and their voter registration status. If the user has a voter registration status of `uncertain`, `unregistered`, or `confirmed`, the banner will display content prompting the user to register to vote.
+Besides self reporting, we check if the user is authenticated and their voter registration status. If the user has a voter registration status of `unregistered` or they have no status saved to their profile, the banner will display content prompting the user to register to vote.
```
aaronschachter, line 21: Any of the `step-*` statuses will also get this prompt. Our best bet might be to try to get this wording to match our logic within the component -- this is the default status if user's status isn't self reported or verified to be complete/ineligible.

docs/development/features/popover-dispatcher.md
```diff
@@ -0,0 +1,15 @@
+# Popover Dispatcher
+
+## Overview
+
+This is a feature used to display various banners and popover elements throughout the site. It checks what type of page we're on based on it's pathname, and from there displays (or does not display) the correct element.
```
aaronschachter, line 5: Grammari nit
```suggestion
This is a feature used to display various banners and popover elements throughout the site. It checks what type of page we're on based on its pathname, and from there displays (or does not display) the correct element.
```

aaronschachter (commented):
This is looking great -- just had a question about checking in on expected functionality per refactor from the last PR... (also makes me wonder about local storage + Cypress tests, which would be a nice-to-have if doesn't exist)

docs/development/features/sitewide-banner.md
aaronschachter: I don't think we need to spend too much time on the voter reg documentation logic though -- because it's all about to get [deleted anyway](https://www.pivotaltracker.com/story/show/175293569)...

resources/assets/components/App.js
lindsayrmaher: Deffo makes sense to me re moving the cta, description, and link props to be the default inline. 

As far as closing the two, maybe we refactor to take a `dismissableElement` and `contextSource` prop? My only issue there is that we'd need to wrap every returned "widget" with a `DismissableElement` parent container which feels messy as well.

lindsayrmaher: @aaronschachter they both have separate local storage variables, but do you mean we should make sure that both aren't triggered by one being rendered? I think they should be separate because they're handled by different handleClose functions, but I'll look into it more.

aaronschachter: Ah I think if they are different local storage variables, and you've just spot checked that closing the sitewide banner doesn't affect displaying a newsletter widget, we're good!

resources/assets/components/utilities/PopoverDispatcher/config.js
```diff
@@ -1,13 +1,11 @@
...
- * This file may eventually turn into config variables to avoid hardcoding when to suppress.
  *
  * Because the SitewideBanner CTA is hardcoded to redirect to vote.dosomething.org, we want to hide
  * the banner on certain pages based on content, or when authenticated user is already registered or
  * ineligible to vote.
  */
-
 // An asterik can be used to exclude all children of a path, e.g. /us/campaigns/ready-vote/*
-export const excludedPaths = [
+export const siteWideBannerExcludedPaths = [
```
aaronschachter, line 8: Such a nit, but can we change the capitalization to `sitewideBanner` to keep consistent with the `SitewideBanner` component casing (vs `SiteWideBanner`)

aaronschachter (approved):
Lovely! 🖼️

resources/assets/components/App.js
lindsayrmaher: mmmm I think I spoke too soon..it the widget is suppressed if you dismiss the banner from another page.

lindsayrmaher: I think perhaps I need to move the dismissable element wrapper into the `SitewideBanner` component, or do i need to wrap each returned `SitewideBanner` within the `PopoverDispatcher`? 😭

aaronschachter: Do you mind linking to the dev URL that renders a scholarships newsletter? Offhand I don't know without starting to dig into the code, which I can :eyes: a little later today

lindsayrmaher: Yeah! https://dev.dosomething.org/us/about/easy-scholarships

aaronschachter: Thanks! Yeah, I can replicate the bug by dismissing the sitewide banner, and then verifying the easy scholarships doesn't display the CTA -- because in the `App` component, we're wrapping the `PopoverDispatcher` in a `DismissableElement` with the name `sitewide_banner_call_to_action`. 
 
The cleanest fix sounds like wrapping each `SitewideBanner` instance with its own `DismissableElement`, because each `SitewideBanner` instance has a separate CTA (if someone dismissed the banner CTA to register, and then they register -- I'd assume they should get a new CTA for refer friends vs waiting another X days from dismissing the old CTA).

However... we're about to scrap all of the voter reg logic anyway for https://www.pivotaltracker.com/story/show/175293569 -- which means we'd essentially only need to wrap the `SitewideBanner` once because we wont' have any more variations (for now). I suggest we merge this as is (since the bug was introduced in https://github.com/DoSomething/phoenix-next/pull/2408 -- apologies for not catching sooner), and then open a new PR to fix the bug, and also simplify the logic to simply display the Refer Friends CTA.  What do you think about that approach?

lindsayrmaher: That approach makes good sense to me! Are you thinking of making the PR this week or waiting until after the 3rd? And then second q, so the logic would be either widget or the default `sitewidebanner` which will be Refer a Friend?

aaronschachter: We have a [story](https://www.pivotaltracker.com/story/show/175293569) this sprint for yep -- either widget or RAF sitewide banner. It also documents we're okay to make the change before Nov 3

PR #2415, opened 2020-10-26
Grab `affirmationContent` from `PhotoSubmissionAction`

Vmack:
### What's this PR do?

This pull request will allow affirmation copy to be editable in Contentful by providing a `submissionActionId`and allowing content to be queried according to the Id

👀

This is part 1b of the Related Campaigns work

### Relevant tickets

References [Pivotal #174885003](https://www.pivotaltracker.com/story/show/174885003).

### Checklist

### DIFF
--- a/resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
+++ b/resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
@@ -1,27 +1,24 @@
 import React from 'react';
 import gql from 'graphql-tag';
-import { useQuery } from '@apollo/react-hooks';
 
-import Card from '../../utilities/Card/Card';
-import Spinner from '../../artifacts/Spinner/Spinner';
+import { query } from '../../../helpers';
+import Query from '../../Query';
 import SiteFooter from '../../utilities/SiteFooter/SiteFooter';
 import SiteNavigationContainer from '../../SiteNavigation/SiteNavigationContainer';
 
 export const GET_AFFIRMATION_CONTENT = gql`
   query submissionActionIdQuery($id: String!) {
-    block(id: $submissionActionId)
+    block(id: $id) {
+      id
+      ... on PhotoSubmissionBlock {
+        affirmationContent
+      }
+    }
   }
 `;
 
-const ShowSubmissionPage = id => {
-  const { loading, error, data } = useQuery(GET_AFFIRMATION_CONTENT, {
-    variables: { id },
-    skip: !id,
-  });
-
-  <Card className="bordered rounded" title={title}>
-    {loading ? <Spinner className="flex justify-center p-16" /> : null}
-  </Card>;
+const ShowSubmissionPage = () => {
+  const id = query('submissionActionId');
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
```diff
@@ -1,9 +1,26 @@
...
+export const GET_AFFIRMATION_CONTENT = gql`
+  query submissionActionIdQuery($id: String!) {
+    block(id: $id) {
+      id
+      ... on PhotoSubmissionBlock {
+        affirmationContent
+      }
+    }
+  }
+`;
```
aaronschachter, line 19: This GraphQL query could use TitleCase to follow the convention we use for other queries.. but I'm noticing that the [`CONTENTFUL_BLOCK_QUERY` ](https://github.com/DoSomething/phoenix-next/blob/319c3513e76e98d3cdaaca543953504859e631a1/resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js#L38) constant will already grab this field for a `PhotoSubmissionBlock` (it's using a fragment defined here:) https://github.com/DoSomething/phoenix-next/blob/ccf850ba8e15a748afa1e4974bdff77273618d6f/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js#L49

What do you think about exporting the GraphQL query from that file, to avoid repeating the GraphQL query/fragment?   We'd need to add an `export` to when declaring the constant in order to import it here:
https://github.com/DoSomething/phoenix-next/blob/319c3513e76e98d3cdaaca543953504859e631a1/resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js#L38

☝️ This would turn into
```
export const CONTENTFUL_BLOCK_QUERY = gql`
```
which would allows us to import the constant in this file
```
import { CONTENTFUL_BLOCK_QUERY } from '../../utilities/ContentfulEntryLoader/ContentfulEntryLoader';
```
and then pass it to our `Query` component later on:
```
<Query query={CONTENTFUL_BLOCK_QUERY} variables={{ id }}>
```
This would also future-proof this submission page if we wanted to use it for submitting other post types as well.

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
```diff
@@ -13,11 +30,22 @@ const ShowSubmissionPage = () => {
           <h1 className="uppercase text-3xl md:text-4xl font-league-gothic font-normal">
             We Got Your Submission
           </h1>
-          <p>
-            Thanks for joining the movement! After we review your submission,
-            we&apos;ll add it to the public gallery alongside submissions from
-            all the other members taking action in this campaign.
-          </p>
+          <Query query={GET_AFFIRMATION_CONTENT} variables={{ id }}>
```
aaronschachter, line 33: Maybe nothing we need to worry about for this PR, but will this return an error if a `submissionActionId` query parameter is not present?

aaronschachter:
I think rebasing this branch with master should resolve the Cypress errors.

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
Vmack: wow that's a great alternative!!

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
```diff
@@ -1,9 +1,16 @@
...
 import SiteFooter from '../../utilities/SiteFooter/SiteFooter';
+import TextContent from '../../utilities/TextContent/TextContent';
 import SiteNavigationContainer from '../../SiteNavigation/SiteNavigationContainer';
+import { CONTENTFUL_BLOCK_QUERY } from '../../utilities/ContentfulEntryLoader/ContentfulEntryLoader';
 
 const ShowSubmissionPage = () => {
+  const id = query('submissionActionId');
+  const defaultContent =
+    'Thanks for joining the movement! After we review your submission, we&apos;ll add it to the public gallery alongside submissions from all the other members taking action in this campaign.';
   return (
```
aaronschachter, line 14: Nit: adding a newline before `return`
```suggestion

  return (
```

aaronschachter (approved):

PR #2417, opened 2020-10-29
Adjust Dismissible Element for Sitewide banner

lindsayrmaher:
### What's this PR do?

This pull request accomplishes a few things! 

1. We had a 🐛  with the `PopoverDispatcher` updates that ended up suppressing the newsletter CTA popover if the SitewideBanner had been dismissed. So we moved the `DismissibleElement` wrapper inside of the dispatcher itself.

2. Now that Voter Registration efforts are winding down, we're able to make the Refer A Friend CTA our default Sitewide Banner, and remove some of the other checks for specific Voter Reg banners to be displayed.

3. We needed to be accurately tracking that the Refer a Friend banner was being seen and dismissed or interacted with vs the Voter Registration CTA, so this PR also updates the `contextSource` on the DismissibleElement wrapping the default sitewideBanner.

4. Because of number 2 & 3, this PR also updates the documentation for the sitewide banner and the cypress tests.

Can anything else be removed or adjusted? Does anymore detail need to be added to documentation?

Some additional work went into this PR because it was easier to accomplish it all in one fell swoop instead of opening 3 separate PRs.

### Relevant tickets

References [Pivotal #175211087](https://www.pivotaltracker.com/story/show/175211087),[Pivotal #175293569](https://www.pivotaltracker.com/story/show/175293569), and [Pivotal #175154024](https://www.pivotaltracker.com/story/show/175154024).

### Checklist


resources/assets/components/utilities/PopoverDispatcher/PopoverDispatcher.js
```diff
@@ -130,53 +127,21 @@ const PopoverDispatcher = props => {
...
-  }
-
   return createPortal(
-    <SitewideBanner
-      cta="Get Started"
-      description="Make your voice heard. Register to vote in less than 2 minutes."
-      link={`https://vote.dosomething.org/?r=${getTrackingSource('hellobar')}`}
-      {...props}
+    <DismissableElement
+      name="sitewide_banner_call_to_action"
```
aaronschachter, line 132: What do you think about naming this as `sitewide_banner_refer_frends`? That way, if we want to display different CTA's later on based on different attributes (e.g.https://www.pivotaltracker.com/story/show/175406026) -- we'd start with a clean slate.

resources/assets/components/utilities/PopoverDispatcher/PopoverDispatcher.js
```diff
@@ -130,53 +127,21 @@ const PopoverDispatcher = props => {
...
+      daysToReRender={7}
+      context={{ contextSource: 'refer_a_friend' }}
+      render={(handleClose, handleComplete) => (
+        <SitewideBanner
+          contextSource="refer_a_friend"
```
aaronschachter, line 137: Do we use the prefix `refer_a_friend` anywhere? If not, it might be cleanest to keep as `refer-friends` (per the feature flag name, URL path) to keep a little shorter/readable)

cypress/integration/sitewide-banner.js
```diff
@@ -174,34 +109,7 @@ describe('Site Wide Banner', () => {
...
-
-    cy.findByTestId('sitewide-banner-button').should('have.length', 1);
-    cy.findByTestId('sitewide-banner-button').should(
-      'have.attr',
-      'href',
-      `https://vote.dosomething.org/?r=user:${user.id},source:web,source_details:hellobar`,
+      '/us/account/refer-friends',
     );
   });
 });
```
aaronschachter, line 115: I think it might be helpful to rename this file as `popover-manager.js`. A great nice to have would be to add a test that when closing the sitewidebanner and then visiting one of the newsletter pages, that we still see the newsletter. We could do this by using any of the fixtures and just modifying its path to be one of the paths in config. This would be fine for a followup PR too.

aaronschachter (approved):
Looks good! Just a few minor suggested changes.

resources/assets/components/utilities/PopoverDispatcher/PopoverDispatcher.js
lindsayrmaher: See my other comment, but I used this based on what was in the pivotal ticket for updating the analytics. I'll check in about changing 👍

cypress/integration/sitewide-banner.js
lindsayrmaher: Like this idea! I'll look into it while we're finalizing the analytics pieces.

resources/assets/components/utilities/PopoverDispatcher/PopoverDispatcher.js
lindsayrmaher: I think that makes sense with the way we refactored this. I need to check in with Diego D to see if this change and the change you mention for the contextSource are fine for him as far as analytics go. If so then i'm cool with this!.

lindsayrmaher: Also wondering if we need to update (or even include) the contextSource if we are calling out with CTA it is within the name.

aaronschachter: Totally understand analytics implications of changing the`contextSource` -- but isn't the `name` prop only used for identifying local storage variables? This is where I think it'd be useful to storage based on the sitewide banner's content (so when the sitewide banner CTA is changed, essentially everyone sees it the first time)

PR #2419, opened 2020-10-29
Equip scholarship gallery type with 1 click signup

Dinnall:
### What's this PR do?

This pull request adds an auth flow for unauthorized users to redirect them if they try to sign up for a scholarship on the Easy Scholarship page. 

...

 I removed the `handleScholarshipCardShareClick` from `ScholarshipCard` to the `GalleryBlockSignup`
to continue to track the events.
...

### Relevant tickets

References [Pivotal #173308345](https://www.pivotaltracker.com/n/projects/2401401/stories/173308345).

### Checklist

Auth users :


resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -39,8 +75,9 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
   return (
     <SecondaryButton
       className="w-full"
+      href={path}
```
mendelB, line 78: I kinda dig adding the `href` here! On the one hand I was thinking it's misleading because this doesn't actually _function_ as a link but as a standard `<button>`, on the other hand, it does allow a preview of the actual ultimate destination being the campaign 🤔 

But just confirming this doesn't affect any of the functionality?

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -39,8 +75,9 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
...
       text="Apply Now"
-      onClick={handleSignup}
+      onClick={handleScholarshipCardShareClick}
```
mendelB, line 80: Have we flagged for Diego D. / Data that we'll now be tracking this event for unauthenticated _and_ authenticated users? (Previously we were only tracking for the former) ...Or is that our intention here to begin with?

Dinnall: this was flagged in the ticket

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
Dinnall: Adding the path here doesn't affect the functionality.
It can be removed because clicking the campaign title can lead to the individual page while clicking the `apply now` signs a user up

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
mendelB: Ah gotcha, I don't recall @ddonizeth responding to [this comment](https://www.pivotaltracker.com/story/show/173308345/comments/219198179) in the ticket, but was this confirmed elsewhere?

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
mendelB: Oh right good point! Then that's probably a good idea, so we we get the `<button>` here instead of an `<a>`.

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
Dinnall: No answer, I will ping @ddonizeth on the ticket for a final decision

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -22,6 +29,35 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
...
+        url: path,
+        ...getPageContext(),
+        campaignId,
+      },
+    });
+
+    if (!isAuthenticated()) {
+      authenticate({ campaignId });
+    }
+    handleSignup();
```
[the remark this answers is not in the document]
Dinnall: I do like that the ternary is a more concise suggestion, but I also can't think of a time that `handleSignup` will be fired before reaching the if statement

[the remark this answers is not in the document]
mendelB: I don't think it would be able to be fired _before_ we hit the if statement, but I was thinking about afterward, while the `authenticate` function is running, we'll continue to process through the end of this method which means we'll also attempt to run the `handleSignup` function potentially?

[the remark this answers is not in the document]
Dinnall: You're correct we won't be able to fire the `handleSingup` before the if statement, that's what we want.
We want to always check if they have an account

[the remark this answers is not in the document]
mendelB: Sorry - just to clarify what I mean, here's a similar code snippet stubbing out the functionality here. You can see that even though we're not authenticated, the `signup` method ends up being executed:

image

But if we add a strict condition it won't be able to run:

image

(I don't think this is a very big deal in practice, because the signup will fail without an authentication token anyhow, but it could interfere with analytics if we get a lot of those errors).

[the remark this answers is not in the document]
Dinnall: Thanks for the visual, I can commit the suggested ternary here

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -22,6 +29,33 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
...
+        ...getPageContext(),
+        campaignId,
+      },
+    });
+
+    return isAuthenticated() ? handleSignup() : authenticate({ campaignId });
+  };
+
+  useEffect(() => {
+    if (isAuthenticated() && flash.campaignId === campaignId) {
```
DFurnes, line 54: Since this logic jumps around a bit due to the auth flow, it might be nice to add a comment giving context:

```suggestion
    // If we're returning from the authentication flow with a "flashed" campaign ID, complete the signup:
    if (isAuthenticated() && flash.campaignId === campaignId) {
```

DFurnes (approved):
Looks great to me!

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
DFurnes: We should be able to differentiate between these two types of clicks based on whether or not the payload has a `userId` parameter, right? If so, it should be possible to differentiate these two "types" of clicks.

I'm curious if we want to also fire an event in the `useEffect` below, like a `register_completed` or something, to easily track the "funnel" of unauthenticated folks who click the button, finish the registration form, and return here? (Or perhaps do we already get enough information on that from other events we fire in Northstar or here?)

mendelB: Good point re differentiation! I know Diego was after an event akin to our `phoenix_completed_signup` event we track for the current main signup flow, so maybe worth firing something like that once the mutation is successful? (Which would ofc imply that previously anonymous users completed registration as well).

Dinnall: That would be great, just as @mendelB mentioned it would imply they finished the registration flow. I guess it wouldn't hurt to again check if they are registered before proceeding 🤔

PR #2429, opened 2020-11-09
Replace 'AbsoluteUrl' scalar with 'URL'.

DFurnes:
### What's this PR do?

This pull request updates usage of `AbsoluteUrl` with `URL`, per DoSomething/graphql#298.

👀

We swapped the package that we'd gotten this scalar type from in DoSomething/graphql#297 to address a security issue. It turns out that the new package's equivalent type, `URL`, can't be aliased to the old `AbsoluteUrl` name we'd been using even though they're functionally the same! So, unfortunately, this needs to be a breaking change. 😩

We, luckily, don't need to change the `Date` or `DateTime` types since their names stayed the same!

References DoSomething/graphql#297 and DoSomething/graphql#298.

### Relevant tickets

References [Pivotal #175625159](https://www.pivotaltracker.com/story/show/175625159).

### Checklist


lindsayrmaher (approved):
approve!

PR #2430, opened 2020-11-09
Progress Bar Sixpack Test

lindsayrmaher:
### What's this PR do?

This pull request adds a sixpack test to see whether showing the progress bar on the 'Go Greener' campaign improves our signup rates. I'm having trouble following the documentation and making sure everything is set up correctly/I'm missing something on the sixpack interface side of things.

Help with making sure everything is set up correctly so that I can test everything out.

**Landing Page Banner with Experiment Running:**
Screen Shot 2020-11-10 at 3 27 58 PM

Screen Shot 2020-11-10 at 3 28 12 PM

We're using the progress bar that initially was made for `VoterRegistrationDrivePageBanner`! We're also creating two environment variables rather than pulling from Rogue because the exact data point we want isn't available yet via graphql and we decided it was smarter to do less work on an experiment.

### Relevant tickets

References [Pivotal #175488432](https://www.pivotaltracker.com/story/show/175488432).

### Checklist

### DIFF
--- a/config/feature-flags.php
+++ b/config/feature-flags.php
@@ -23,4 +23,6 @@ return [
     'refer_friends_incentive' => env('DS_ENABLE_REFER_FRIENDS_INCENTIVE', false),
     'sitewide_nps_survey' => env('DS_ENABLE_SITEWIDE_NPS_SURVEY', false),
     'volunteer_credits' => env('DS_ENABLE_VOLUNTEER_CREDITS', false),
+    'go_greener_campaign_goal' => env('DS_GO_GREENER_GOAL', null),
+    'go_greener_campaign_quantity' => env('DS_GO_GREENER_QUANTITY', 0),
 ];
--- a/config/services.php
+++ b/config/services.php
@@ -58,4 +58,9 @@ return [
         'url' => 'https://app.contentful.com',
         'cache' => env('CONTENTFUL_CACHE', true),
     ],
+
+    'go-greener-campaign' => [
+        'goal' => env('DS_GO_GREENER_GOAL'),
+        'quantity' => env('DS_GO_GREENER_QUANTITY'),
+    ]
 ];
--- a/config/services.php
+++ b/config/services.php
@@ -62,5 +62,5 @@ return [
     'go-greener-campaign' => [
         'goal' => env('DS_GO_GREENER_GOAL'),
         'quantity' => env('DS_GO_GREENER_QUANTITY'),
-    ]
+    ],
 ];
--- a/config/services.php
+++ b/config/services.php
@@ -58,9 +58,4 @@ return [
         'url' => 'https://app.contentful.com',
         'cache' => env('CONTENTFUL_CACHE', true),
     ],
-
-    'go-greener-campaign' => [
-        'goal' => env('DS_GO_GREENER_GOAL'),
-        'quantity' => env('DS_GO_GREENER_QUANTITY'),
-    ],
 ];
--- a/resources/assets/components/CampaignBanner/CampaignBanner.js
+++ b/resources/assets/components/CampaignBanner/CampaignBanner.js
@@ -108,7 +108,6 @@ const CampaignBanner = ({
                 internalTitle="Campaign Progress Bar Experiment"
                 convertableActions={['signup']}
                 alternatives={[
-                  // Displaying the progress bar is the default, or control test alternative.
                   <div className="mb-6" testName="Progress Bar Visible">
                     <ProgressBar percentage={percentage} />
                     <p className="text-lg">
--- a/resources/assets/components/CampaignBanner/CampaignBanner.js
+++ b/resources/assets/components/CampaignBanner/CampaignBanner.js
@@ -17,6 +17,8 @@ import Spinner from '../artifacts/Spinner/Spinner';
 import CampaignHeader from '../utilities/CampaignHeader';
 import ErrorBlock from '../blocks/ErrorBlock/ErrorBlock';
 import CoverImage from '../utilities/CoverImage/CoverImage';
+import { getGoalInfo } from '../../helpers/voter-registration';
+import ProgressBar from '../utilities/ProgressBar/ProgressBar';
 import TextContent from '../utilities/TextContent/TextContent';
 import { SCHOLARSHIP_SIGNUP_BUTTON_TEXT } from '../../constants';
 import CampaignInfoBlock from '../blocks/CampaignInfoBlock/CampaignInfoBlock';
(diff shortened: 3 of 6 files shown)


config/services.php
```diff
@@ -58,4 +58,9 @@
         'url' => 'https://app.contentful.com',
         'cache' => env('CONTENTFUL_CACHE', true),
     ],
+
+    'go-greener-campaign' => [
+        'goal' => env('DS_GO_GREENER_GOAL'),
+        'quantity' => env('DS_GO_GREENER_QUANTITY'),
+    ],
```
DFurnes, line 65: Nitpick: let's put this in `feature-flags.php` (since it doesn't really fit under the third-party service umbrella). One nice side-effect there is we include the entirety of that file on the front-end automatically, under `env('FEATURE_FLAGS')`, so there's no extra "wiring up" necessary that way!

resources/assets/components/CampaignBanner/CampaignBanner.js
```diff
@@ -95,6 +105,24 @@ const CampaignBanner = ({
             data-testid="campaign-banner-primary-content"
             className="grid-wide-7/10 mb-6"
           >
+            {numCampaignId === 9109 || numCampaignId === 9006 ? (
+              <SixpackExperiment
+                title="Progress Bar Experiment"
+                convertableActions={['signup']}
+                alternatives={[
+                  // The displaying the progress bar is the default, or control test alternative.
```
DFurnes, line 113: I'm having trouble reconciling the docs with the source code for this component - my understanding from skimming the code is that the control is specified as the `control` prop and if a single alternative is provided like this, then the experiment will be between an "empty" control & a single "test" alternative.

Is that what you're seeing when you test this out?

DFurnes:
@lindsayrmaher This looks good to me! What sorts of problems are you running into?

resources/assets/components/CampaignBanner/CampaignBanner.js
lindsayrmaher: That was what I assumed as well re: the "empty" control and the single "test" alternative. And it's what I'm expecting since the test is specifically see nothing or see the progress bar.

lindsayrmaher:
> @lindsayrmaher This looks good to me! What sorts of problems are you running into?

So basically I'm trying to force the alternative using the query params that are mentioned in the docs

> When testing to confirm that an experiment is set up properly, you can force a specific test alternative by passing a URL parameter in the URL

But I feel like maybe I have to actually set up the test in sixpack? I can't remember how to access the dashboard, and the last sixpack test I ran was in northstar so the setup was a bit different.

DFurnes:
You can access the Sixpack dashboard at http://experiments-qa.dosomething.org (we don't have a "dev" instance for this app, at the moment). The username & password can be found in Lastpass under the "Sixpack [QA/Thor" entry in the "Shared-TechDatabase" vault.

As far as I remember, tests are automatically set up in Sixpack the first time they're "run" in code, so you should see the test (and preliminary results from your first visit) in the dashboard as soon as you've set this up. You can also double-check that Sixpack is enabled for your local environment by checking [these variables](https://github.com/DoSomething/phoenix-next/blob/004702f93331e8edc16b8e11e816a602530f4d73/.env.example#L54-L57) in your `.env` file.

If that doesn't work, let me know & we can keep digging!

lindsayrmaher:
> As far as I remember, tests are automatically set up in Sixpack the first time they're "run" in code, so you should see the test (and preliminary results from your first visit) in the dashboard as soon as you've set this up. You can also double-check that Sixpack is enabled for your local environment by checking [these variables](https://github.com/DoSomething/phoenix-next/blob/004702f93331e8edc16b8e11e816a602530f4d73/.env.example#L54-L57) in your `.env` file.

I was able to log into the dashboard and double check the env variables for sixpack. I noticed in some other [Sixpack experiments](https://github.com/DoSomething/phoenix-next/pull/1509) we've done in the past an `additionalContent` value was necessary on the campaign in order to run the test. Is that something we need here as well? 

Also noticed we have a [ticket in the icebox](https://www.pivotaltracker.com/story/show/170964251) for applications that references not being able to test on the Landing Page right now but I wasn't sure if that just pertained to tests in contentful.

cc @DFurnes and @mendelB

mendelB:
> Also noticed we have a [ticket in the icebox](https://www.pivotaltracker.com/story/show/170964251) for applications that references not being able to test on the Landing Page right now but I wasn't sure if that just pertained to tests in contentful.

@lindsayrmaher correct! This should only affect our ability to run tests on the Landing Page from Contentful, but tests in code should be perfectly fine 😌 

> I noticed in some other Sixpack experiments we've done in the past an additionalContent value was necessary on the campaign in order to run the test. Is that something we need here as well?

I believe in the past, we've used an explicit value in `additionalContent` to opt-in campaigns to sixpack tests so that we could run them selectively. If this test is _only_ for a singular campaign though, I'd think it's safe to hardcode the ID or something!

lindsayrmaher:
@DFurnes can you just give this one more 👀  before I merge?

resources/assets/components/CampaignBanner/CampaignBanner.js
```diff
@@ -95,6 +103,24 @@ const CampaignBanner = ({
             data-testid="campaign-banner-primary-content"
             className="grid-wide-7/10 mb-6"
           >
+            {numCampaignId === 9109 || numCampaignId === 9001 ? (
+              <SixpackExperiment
+                internalTitle="Campaign Progress Bar Experiment"
+                convertableActions={['signup']}
+                alternatives={[
+                  // Displaying the progress bar is the default, or control test alternative.
```
DFurnes, line 111: Let's remove this comment, since it implies that this is the experimental control.

DFurnes (approved):
Looks great!

config/feature-flags.php
```diff
@@ -23,4 +23,6 @@
     'refer_friends_incentive' => env('DS_ENABLE_REFER_FRIENDS_INCENTIVE', false),
     'sitewide_nps_survey' => env('DS_ENABLE_SITEWIDE_NPS_SURVEY', false),
     'volunteer_credits' => env('DS_ENABLE_VOLUNTEER_CREDITS', false),
+    'go_greener_campaign_goal' => env('DS_GO_GREENER_GOAL', null),
+    'go_greener_campaign_quantity' => env('DS_GO_GREENER_QUANTITY', 0),
```
mendelB, line 27: @lindsayrmaher I'm late with this, and this is not a biggie at all -- but we've set up a `config/site.php` file to configure env variables more related to a site configuration then a feature-flagged feature per se (I _think_ this would fall under the _config_ category), which can be accessed on the front-end via the [`site_config` helper](https://github.com/DoSomething/phoenix-next/blob/0aba76c58d5ec95e71fe0b3c912620256314c12c/resources/assets/helpers/index.js#L514-L523)! 

(helpful to clearly differentiate the two use-cases in code 😌)

lindsayrmaher: oh nice! i can make this update in a small PR i'm making now. i think it makes more sense to keep them here than the `feature-flags`

PR #2440, opened 2020-11-24
Fixes issue rendering unpublished Call To Action blocks.

DFurnes:
### What's this PR do?

This pull request fixes an issue rendering unpublished Call To Action blocks on [preview.dosomething.org](https://preview.dosomething.org/us).

Here's an example of an unpublished block that does not render on preview:

Screen Shot 2020-11-24 at 3 28 26 PM

After making this change, it does!

Screen Shot 2020-11-24 at 3 31 36 PM

Since GraphQL is able to query either Contentful's standard or "preview" APIs from the same endpoint, we choose which to read from using an argument passed to the `block()` query. In the future, this could allow us to enable admins to preview items within the normal "production" site...!

### Relevant tickets

References [Pivotal #173016057](https://www.pivotaltracker.com/story/show/173016057).

### Checklist

### DIFF
--- a/resources/assets/components/blocks/CallToActionBlock/CallToActionBlock.js
+++ b/resources/assets/components/blocks/CallToActionBlock/CallToActionBlock.js
@@ -5,13 +5,13 @@ import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
 import { useQuery } from '@apollo/react-hooks';
 
-import { tailwind } from '../../../helpers';
+import { env, tailwind } from '../../../helpers';
 import Spinner from '../../artifacts/Spinner/Spinner';
 import PrimaryButton from '../../utilities/Button/PrimaryButton';
 
 const CALL_TO_ACTION_QUERY = gql`
-  query CallToActionBlockQuery($id: String!) {
-    block(id: $id) {
+  query CallToActionBlockQuery($id: String!, $preview: Boolean!) {
+    block(id: $id, preview: $preview) {
       id
       ... on CallToActionBlock {
         superTitle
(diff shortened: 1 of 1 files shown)


resources/assets/components/blocks/CallToActionBlock/CallToActionBlock.js
```diff
@@ -28,17 +29,21 @@ const CALL_TO_ACTION_QUERY = gql`
...
+    return <ErrorBlock error={error} />;
   }
 
   if (loading) {
     return <Spinner className="flex justify-center p-2" />;
   }
 
+  if (!data.block) {
+    return <ErrorBlock error="Block does not exist." />;
+  }
```
DFurnes, line 45: The "null is not an object (evaluating 'c.superTitle')" we were seeing was _really_ because we'd return `null` for this block (as it was unpublished), and this component doesn't know how to handle that. In many cases, this can be fixed by checking for an unpublished block but if the entry _truly_ does not exist, we'll return this error block.

resources/assets/components/blocks/CallToActionBlock/CallToActionBlock.js
```diff
@@ -28,17 +29,21 @@ const CALL_TO_ACTION_QUERY = gql`
 
 const CallToActionBlock = ({ id }) => {
   const { data, loading, error } = useQuery(CALL_TO_ACTION_QUERY, {
-    variables: { id },
+    variables: { id, preview: env('CONTENTFUL_USE_PREVIEW_API', false) },
   });
 
   if (error) {
-    return <p>Something went wrong!</p>;
+    return <ErrorBlock error={error} />;
```
DFurnes, line 36: As you can see in the screenshot, we _were_ seeing this error block before but not from here! It was rendering from an error boundary that caught when we tried to read from the null `data.block` response. This `if (error)` condition would only trigger for errors that fail the GraphQL request (like a network error or query syntax issue).

Dinnall: Love this change, more descriptive and can be helpful while debugging in the future

Dinnall (approved):
Looks good to me 💯

PR #2442, opened 2020-11-25
Update GalleryBlockSignup.js

Dinnall:
### What's this PR do?

This pull request adds a query to determine the text displayed on the scholarship buttons. If a user hasn't signed up for a campaign the button will display the text `Apply Now`, if they signed up for the campaign already it will say `View Application`

🔢 
...

...

### Relevant tickets

References [Pivotal #173454908](https://www.pivotaltracker.com/n/projects/2401401/stories/173454908).

### Checklist

Not signed up:
Screen Shot 2020-11-25 at 12 25 23 PM

Campaigns associated with this user:
Screen Shot 2020-11-25 at 12 24 08 PM

### DIFF
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -105,7 +105,7 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
   return (
     <SecondaryButton
       className="w-full"
-      text={textToDisplay(campaignData.signups[0].campaignId)}
+      text={textToDisplay(campaignData.signups.length)}
       href={path}
       onClick={handleScholarshipCardShareClick}
     />
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -40,6 +40,7 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
       userId: window.AUTH.id.toString(),
       campaignId: campaignId.toString(),
     },
+    skip: !window.AUTH.id,
   });
 
   const [handleSignup, { loading, data, error }] = useMutation(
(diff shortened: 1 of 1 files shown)


resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -67,14 +99,15 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
...
-  if (error) {
+  if (error || errorCampaign) {
     return <ErrorBlock error={error} />;
   }
 
   return (
     <SecondaryButton
       className="w-full"
-      text="Apply Now"
+      text={textToDisplay(campaignData.signups[0].campaignId)}
```
DFurnes, line 109: Will this cause an issue if the user doesn't have a signup? I'm thinking if `campaignData` returns `[]` (if there isn't a signup for this user/campaign), then this would end up trying to read `.campaignId` from undefined...

If so, could we instead use `campaignData.signups.length` as our check here?

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -21,18 +21,46 @@ export const CREATE_SIGNUP_MUTATION = gql`
...
+  }
+`;
+
 const GalleryBlockSignup = ({ campaignId, path }) => {
+  const {
+    data: campaignData,
+    loading: loadingCampaign,
+    error: errorCampaign,
+  } = useQuery(SEARCH_USER_CAMPAIGN_QUERY, {
+    variables: {
```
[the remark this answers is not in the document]
Dinnall: Good additional!

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
Dinnall: Good question,  leaving it as is or using`campaignData.signups.length` is fine, if  `campaignData.signups[0].campaignId` return undefined its considered a falsy value so it would default to 
`Apply Now`

DFurnes: You're right that `undefined` would be falsy, but I think our problem is that the thing we're reading _from_ (when we try to read `.campaignId` is undefined), and JavaScript doesn't like that. It'd be like trying to read `false.campaignId`.

Here's what I see when I visit this page as a user without a signup on [the review app](https://dosomething-phoenix-de-pr-2442.herokuapp.com/us/about/easy-scholarships) for this PR:

Screen Shot 2020-11-30 at 11 58 57 AM

Dinnall: ah I see now, I went ahead and switched the param for `textToDisplay()`

DFurnes (approved):
Looks good to me! ⛵

PR #2445, opened 2020-12-02
Adding Photo to Submission Page

lindsayrmaher:
### What's this PR do?

This pull request adds a query for the post's photo URL if it exists! We want to add the photo from the post so that users who completed a photo report back can see what they submitted.

👀 

I have two questions I'd like feedback on.

1. Is there a better way to size the photos based on screensize? I went with width in TW because it's easy and seems to work well enough, but not sure if we should be more explicit in defining width etc.

2. On medium screens, when the related campaigns are displayed one per row, I feel like it looks wonky compared to the size of the reportback photo. Thoughts?

Large:
Screen Shot 2020-12-03 at 12 02 10 PM

Small:
Screen Shot 2020-12-03 at 12 03 33 PM

Medium:
Screen Shot 2020-12-03 at 12 03 45 PM

This is only relevant for photo submission reportbacks

### Relevant tickets

References [Pivotal #173236812](https://www.pivotaltracker.com/story/show/173236812).

### Checklist

### DIFF
--- a/resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
+++ b/resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
@@ -46,8 +46,8 @@ const ShowSubmissionPage = ({ match }) => {
     <>
       <SiteNavigationContainer />
 
-      <div className="base-12-grid bg-white">
-        <main className="grid-wide lg:flex">
+      <main className="base-12-grid">
+        <div className="grid-wide lg:flex bg-white">
           {postImageUrl ? (
             <div className="w-1/2 md:w-1/4 pt-6 lg:p-6">
               {loading ? (
@@ -86,11 +86,13 @@ const ShowSubmissionPage = ({ match }) => {
               <TextContent className="mb-6">{defaultContent}</TextContent>
             )}
           </div>
-        
+        </div>
 
-        <RecommendedCampaignsGallery />
+        <div className="grid-wide">
+          <RecommendedCampaignsGallery />
+        </div>
       </main>
-      </div>
+
       <SiteFooter />
     </>
   );
--- a/resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
+++ b/resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { get } from 'lodash';
 import gql from 'graphql-tag';
+import classnames from 'classnames';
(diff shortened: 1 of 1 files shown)


resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
```diff
@@ -1,52 +1,105 @@
...
-            </Query>
-          ) : (
-            <p>
+      <main className="base-12-grid">
+        <div className="grid-wide lg:flex bg-white">
+          {postImageUrl ? (
+            <div className="w-1/2 md:w-1/4 pt-6 lg:p-6">
+              {loading ? (
+                <Spinner />
+              ) : (
```
mendelB, line 55: I don't think we can ever actually hit this state? Since the `postImageUrl` will not be assigned until after we've finished loading. Although I'm not sure offhand what a good alternative would be here since we probably don't want to show the spinner for non-image posts. 

Maybe the concession is just no spinner 🤔

mendelB (commented):
> Is there a better way to size the photos based on screensize? I went with width in TW because it's easy and seems to work well enough, but not sure if we should be more explicit in defining width etc.

I think this is perfectly fine! A related example I can think of is that we don't do any formatting on the [`PostCard`](https://github.com/DoSomething/phoenix-next/blob/66e61dda9e3cc13b29fae6845decf2817c7c4564/resources/assets/components/utilities/PostCard/PostCard.js#L63) for post images in galleries

> On medium screens, when the related campaigns are displayed one per row, I feel like it looks wonky compared to the size of the reportback photo. Thoughts?

Hmmm I wonder if this might be the question of whether we want to keep displaying two cards per row until a smaller viewport? It seems like that's what the design calls for here, but that would need to be a global change across all the galleries.

Also, we're still missing padding on this which might at least help a little 🤞 

Perhaps @LeahAngle can weigh in here?

lindsayrmaher:
> Also, we're still missing padding on this which might at least help a little

@mendelB where do we need to add padding? figured maybe I should add on this PR!

mendelB:
> > Also, we're still missing padding on this which might at least help a little
> 
> @mendelB where do we need to add padding? figured maybe I should add on this PR!

Thanks @lindsayrmaher!  From the designs it just looks like there's supposed to be some margin or padding wrapping the gallery itself, particular to the viewport width (so maybe we wrap it in a `div` with some spacing, or allow the gallery to receive a `className` prop?)

mendelB (approved):
Gallery spacing aside, this looks great! 👍

PR #2446, opened 2020-12-02
Add Recommended Campaigns Gallery Component - Show Submission Page

mendelB:
### What's this PR do?

This pull request adds a new `RecommenedCampaignsGallery` component rendered by the `ShowSubmissionPage` which displays a gallery of the 3 most recent scholarship campaigns to the user as a recommended list to tackle next!

👀 
How's the name? I'm really bad at naming.

This all utilizes existing mechanics set up in Algolia / GraphQL / Rogue in [Pivotal ID#173817560](https://www.pivotaltracker.com/story/show/173817560) which allows us to incorporate scholarship campaigns filtering via the Algolia `searchCampaigns` query (among the other backfilled search parameters).

I opted to not make use of the [`PaginatedCampaignsGallery` component](https://github.com/DoSomething/phoenix-next/blob/5f556d66f4b950f13b331250ae427e4c16f61d49/resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js) nor its query - though that was our original thought - since it seemed dead simple to just statically write up our own query for this gallery. (But LMK if you have thoughts).

More filtering to come in [Pivotal #175358477](https://www.pivotaltracker.com/story/show/175358477) to exclude group campaigns / site excluded campaigns / the current user completed campaign.

I don't think we can add tests since this page will be gated to the user which disqualifies Cypress for now. I can try to finagle this in Jest, but perhaps this will be better suited to a more holistic suite of tests for this page.

### Relevant tickets

References [Pivotal #173236860](https://www.pivotaltracker.com/story/show/173236860).

### Checklist

Not much happening new visually -- it's just the standard Gallery Block with Scholarship Cards:
image

### DIFF
--- a/resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
+++ b/resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
@@ -1,0 +1,44 @@
+import React from 'react';
+import gql from 'graphql-tag';
+
+import Query from '../../Query';
+import GalleryBlock from '../../blocks/GalleryBlock/GalleryBlock';
+import { scholarshipCardFragment } from '../../utilities/ScholarshipCard/ScholarshipCard';
+
+const SCHOLARSHIP_CAMPAIGNS_QUERY = gql`
+  query ScholarshipCampaignsQuery {
+    campaigns: searchCampaigns(
+      perPage: 3
+      hasWebsite: true
+      hasScholarship: true
+      isOpen: true
+      orderBy: "start_date,desc"
+    ) {
+      edges {
+        node {
+          id
+          campaignWebsite {
+            ...ScholarshipCard
+          }
+        }
+      }
+    }
+  }
+
+  ${scholarshipCardFragment}
+`;
+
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
```diff
@@ -0,0 +1,44 @@
...
+  <Query query={SCHOLARSHIP_CAMPAIGNS_QUERY}>
+    {result => (
+      <GalleryBlock
+        blocks={result.campaigns.edges.map(edge => edge.node.campaignWebsite)}
+        galleryType="SCHOLARSHIP"
+        itemsPerRow={3}
+        imageAlignment="LEFT"
+      />
+    )}
+  </Query>
```
katiecrane, line 41: This component is so cool! I agree that it probably is cleaner to have this on its own rather than adding a bunch of conditionals to the `PaginatedCampaignsGallery`

katiecrane (approved):
it's happening!!!

PR #2447, opened 2020-12-02
Update Query component to useQuery hook

mendelB:
### What's this PR do?

This pull request updates the in-house `Query` component to use the `useQuery` hook instead of the older Apollo `Query` component

👀 

I don't think we need the detailed `networkStatus` right? I swapped that for the `loading` variable which I'm pretty sure is the updated equivalent of the status code we were using here (`loading: 1`).

I used this component in #2446 and felt like this might be a nice time to update.

### Checklist

### DIFF
--- a/resources/assets/components/Query.js
+++ b/resources/assets/components/Query.js
@@ -1,32 +1,27 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Query as ApolloQuery } from 'react-apollo';
+import { useQuery } from '@apollo/react-hooks';
 
-import { NetworkStatus } from '../constants';
-import ErrorBlock from './blocks/ErrorBlock/ErrorBlock';
 import Spinner from './artifacts/Spinner/Spinner';
+import ErrorBlock from './blocks/ErrorBlock/ErrorBlock';
 
 /**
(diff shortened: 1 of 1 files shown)


resources/assets/components/Query.js
```diff
@@ -1,32 +1,27 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Query as ApolloQuery } from 'react-apollo';
+import { useQuery } from '@apollo/react-hooks';
 
-import { NetworkStatus } from '../constants';
```
aaronschachter, line 5: Just checking -- we use this anywhere else? If not, let's remove this from the `constants` ✂️

aaronschachter (approved):
Looks great! We used the `Query` component in a few voter reg components too, thanks for taking this on.

resources/assets/components/Query.js
mendelB: We're still using it in the [`PaginatedQuery`](https://github.com/DoSomething/phoenix-next/blob/5f556d66f4b950f13b331250ae427e4c16f61d49/resources/assets/components/PaginatedQuery.js#L20), but come to think of it, that can be refactored to hooks as well! I can tackle that in a follow-up!

PR #2448, opened 2020-12-03
add analytics to view_application and apply_now button

Dinnall:
### What's this PR do?

This pull request adds analytics to the `View Application` and `Apply Now` button

🖲️ 
...

...

### Relevant tickets

References [Pivotal #220141604](https://www.pivotaltracker.com/n/projects/2401401/stories/175155765/comments/220141604).

References [Pivotal #220142613](https://www.pivotaltracker.com/n/projects/2401401/stories/175604649/comments/220142613).

### Checklist

- [  X  ] This PR has been added to the relevant Pivotal card.

### DIFF
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -55,7 +55,6 @@ const GalleryBlockSignup = ({ path, campaignId, campaignTitle }) => {
           category: EVENT_CATEGORIES.campaignAction,
           label: campaignTitle,
           context: {
-            url: path,
             contextSource: 'scholarship_card',
             ...getPageContext(),
           },
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -88,23 +88,16 @@ const GalleryBlockSignup = ({ path, campaignId, campaignTitle }) => {
         });
   };
 
-  const handleScholarshipButtonAnalytic = event => {
-    event.preventDefault();
-
-    if (campaignData && campaignData.signups.length) {
-      trackAnalyticsEvent(
-        'clicked_scholarship_gallery_block_view_application',
-        {
-          action: 'button_clicked',
-          category: EVENT_CATEGORIES.siteAction,
-          label: 'scholarship_gallery_card',
-          context: {
-            url: path,
-            campaignId,
-          },
-        },
-      );
-    }
+  const handleScholarshipButtonAnalytic = () => {
+    trackAnalyticsEvent('clicked_scholarship_gallery_block_view_application', {
+      action: 'button_clicked',
+      category: EVENT_CATEGORIES.siteAction,
+      label: 'scholarship_gallery_card',
+      context: {
+        url: path,
+        campaignId,
+      },
+    });
   };
 
   useEffect(() => {
@@ -138,7 +131,9 @@ const GalleryBlockSignup = ({ path, campaignId, campaignTitle }) => {
       }
       href={path}
       onClick={() => {
-        campaignData && campaignData.signups.length ? handleScholarshipButtonAnalytic() : handleScholarshipCardShareClick();
+        campaignData && campaignData.signups.length
+          ? handleScholarshipButtonAnalytic()
+          : handleScholarshipCardShareClick();
       }}
     />
   );
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -30,7 +30,7 @@ export const SEARCH_USER_CAMPAIGN_QUERY = gql`
   }
 `;
 
-const GalleryBlockSignup = ({ campaignId, path }) => {
+const GalleryBlockSignup = ({ path, campaignId, campaignTitle }) => {
   const {
     data: campaignData,
     loading: loadingCampaign,
@@ -48,6 +48,17 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
     {
       variables: {
         campaignId,
+      },
+      onCompleted: () => {
+        trackAnalyticsEvent('completed_signup', {
+          action: 'signup_completed',
+          category: EVENT_CATEGORIES.campaignAction,
+          label: campaignTitle,
+          context: {
+            url: path,
+            contextSource: 'scholarship_card',
+          },
+        });
       },
     },
   );
@@ -75,6 +86,25 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
       : authenticate({
           campaignId,
         });
+  };
+
+  const handleScholarshipButtonAnalytic = event => {
+    event.preventDefault();
+
+    if (campaignData && campaignData.signups.length) {
+      trackAnalyticsEvent(
+        'clicked_scholarship_gallery_block_view_application',
+        {
+          action: 'button_clicked',
+          category: EVENT_CATEGORIES.siteAction,
+          label: 'scholarship_gallery_card',
+          context: {
+            url: path,
+            campaignId,
+          },
+        },
+      );
+    }
   };
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -77,6 +88,25 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
...
+          category: EVENT_CATEGORIES.siteAction,
+          label: 'scholarship_gallery_card',
+          context: {
+            url: path,
+            campaignId,
+          },
+        },
+      );
+    }
+  };
```
mendelB, line 108: I'd recommend that if the user is already signed up for the campaign, we should redirect them to the campaign straight away instead of triggering another signup.

We can do this by either only triggering _this_ method as the click event on the button when `campaignData && campaignData.signups.length` (like the button text logic), and then adding a redirect (`window.location = path;`) after we track the analytics above.

Dinnall: `handleScholarshipButtonAnalytic` doesn't trigger another signup, it checks that the user is clicking a campaign they already signed up for.

mendelB: Sorry, I'm referring to the `handleScholarshipCardShareClick` which we trigger on the button click regardless of signup status. I think Ii's important also because we'll be triggering a `clicked_scholarship_gallery_block_apply_now` event even when the button is `View Application`.

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -107,17 +137,22 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
           : 'Apply Now'
       }
       href={path}
-      onClick={handleScholarshipCardShareClick}
+      onClick={() => {
+        handleScholarshipCardShareClick();
+        handleScholarshipButtonAnalytic();
+      }}
```
mendelB, line 143: @Dinnall Per https://github.com/DoSomething/phoenix-next/pull/2448#discussion_r535346053 - I think what we should do here is something like the following:
```suggestion
      onClick={() => {
        campaignData && campaignData.signups.length ? handleScholarshipButtonAnalytic() : handleScholarshipCardShareClick();
      }}
```

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -77,6 +88,25 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
...
+          action: 'button_clicked',
+          category: EVENT_CATEGORIES.siteAction,
+          label: 'scholarship_gallery_card',
+          context: {
+            url: path,
+            campaignId,
+          },
+        },
+      );
+    }
```
mendelB, line 107: And per https://github.com/DoSomething/phoenix-next/pull/2448/files#r535347695, we can do something like:
```suggestion
      trackAnalyticsEvent(
        'clicked_scholarship_gallery_block_view_application',
        {
          action: 'button_clicked',
          category: EVENT_CATEGORIES.siteAction,
          label: 'scholarship_gallery_card',
          context: {
            url: path,
            campaignId,
          },
        },
      );
      
      window.location = path;
```

Dinnall: ah - I see, move the place of the if statement, and use a ternary. I like the suggestions. I don't think adding `window.location = path` is needed in the place you suggested because a `href={path}` is already in `SecondaryButton`.

mendelB: Ahhh good call! So if we remove the `event.preventDefault` at the top we'll be golden 😌

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -49,6 +49,17 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
...
       },
+      onCompleted: () => {
+        trackAnalyticsEvent('completed_signup', {
+          action: 'signup_completed',
+          category: EVENT_CATEGORIES.campaignAction,
+          label: campaignTitle,
+          context: {
+            url: path,
+            contextSource: 'scholarship_card',
+          },
```
mendelB, line 60: Perhaps worth adding the page context as in the current click event?
```suggestion
          context: {
            url: path,
            contextSource: 'scholarship_card',
            ...getPageContext(),
          },
```

(I wonder if we should start just including this in the canonical [track event helper](https://github.com/DoSomething/phoenix-next/blob/66e61dda9e3cc13b29fae6845decf2817c7c4564/resources/assets/helpers/analytics.js#L287) 🤔 )

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -77,6 +88,18 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
...
 
+  const handleScholarshipButtonAnalytic = () => {
+    trackAnalyticsEvent('clicked_scholarship_gallery_block_view_application', {
+      action: 'button_clicked',
+      category: EVENT_CATEGORIES.siteAction,
+      label: 'scholarship_gallery_card',
+      context: {
+        url: path,
+        campaignId,
+      },
```
mendelB, line 99:
```suggestion
      context: {
        url: path,
        campaignId,
        ...getPageContext(),
      },
```

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -77,6 +88,18 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
         });
   };
 
+  const handleScholarshipButtonAnalytic = () => {
```
mendelB, line 91: Can we call this something like `handleViewApplicationButtonClick` so it's more specific to the triggering context?

mendelB (commented):
Looking fab! A couple suggestions.

Looks like some formatting changes crept though on the `ScholarshipCard` component -- perhaps worth git resetting that file?

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
Dinnall: Love this suggestion

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -49,6 +49,18 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
       variables: {
         campaignId,
       },
+      onCompleted: () => {
+        trackAnalyticsEvent('completed_signup', {
+          action: 'signup_completed',
+          category: EVENT_CATEGORIES.campaignAction,
+          label: campaignTitle,
+          context: {
+            url: path,
```
mendelB, line 58: Oops sorry missed this, but we shouldn't need the `url` property in the context here since we automatically have the page href via Snowplow / GTM. 

We typically will use `url` in events where we want to denote the specific URL being handled by the action, for instance for [sharing a link on Facebook](https://github.com/DoSomething/phoenix-next/blob/66e61dda9e3cc13b29fae6845decf2817c7c4564/resources/assets/components/actions/ShareAction/ShareAction.js#L102) where the link is separate of the actual page URL. 

(I'm assuming the intention here is to track the page URL and not the campaign path?)

```suggestion
```

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -77,6 +89,19 @@ const GalleryBlockSignup = ({ campaignId, path }) => {
         });
   };
 
+  const handleViewApplicationButtonClick = () => {
+    trackAnalyticsEvent('clicked_scholarship_gallery_block_view_application', {
+      action: 'button_clicked',
+      category: EVENT_CATEGORIES.siteAction,
+      label: 'scholarship_gallery_card',
+      context: {
+        url: path,
```
mendelB, line 98: Same here per above
```suggestion
```

And I know it's not a new change - but I think we should remove it from the `clicked_scholarship_gallery_block_apply_now` event above as well.

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
Dinnall: Interesting didn't know this, and correct the intention here is to track the page URL.

mendelB:
@Dinnall Were you able to reset the changes to `ScholarshipCard.js`? It looks like some formatting changes snuck in that we probably don't want with the PR.

Dinnall:
> @Dinnall Were you able to reset the changes to `ScholarshipCard.js`? It looks like some formatting changes snuck in that we probably don't want with the PR.

No, because of the formatting plugin I have installed

DFurnes:
I wonder if those changes are from a newer Prettier release! We might not have run `npm format` since updating...

DFurnes (approved):
This all looks great to me!

mendelB (approved):
🔥 Love how this all came together!

PR #2449, opened 2020-12-03
Update PaginatedQuery component to use the useQuery hook

mendelB:
### What's this PR do?

This pull request updates the `PaginatedQuery` component to utilize the newer `useQuery` React hook.

👀 

Following up on #2447 

This admittedly doesn't clean much.... Though it doesn't hurt to be hooked into the hook.

I initially wanted to tackle this here to see if we could scrap the network status mapping constant per https://github.com/DoSomething/phoenix-next/pull/2447#discussion_r534649070, but turns out we actually still need it here to differentiate between the initial data fetch and the subsequent "fetch more" requests to either render a plain spinner or just notify the child that we're in loading state.

### DIFF
--- a/resources/assets/components/PaginatedQuery.js
+++ b/resources/assets/components/PaginatedQuery.js
@@ -1,63 +1,62 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Query } from 'react-apollo';
+import { useQuery } from '@apollo/react-hooks';
 
 import { NetworkStatus } from '../constants';
-import ErrorBlock from './blocks/ErrorBlock/ErrorBlock';
 import Spinner from './artifacts/Spinner/Spinner';
+import ErrorBlock from './blocks/ErrorBlock/ErrorBlock';
 
 /**
- * Fetch results via GraphQL using a query component.
+ * Fetch results via GraphQL using a useQuery hook.
  */
-const PaginatedQuery = ({ query, queryName, variables, count, children }) => (
-  <Query
-    query={query}
-    variables={{ ...variables, count, page: 1 }}
-    notifyOnNetworkStatusChange
-  >
-    {result => {
-      // On initial load, just display a loading spinner.
-      if (result.networkStatus === NetworkStatus.LOADING) {
-        return <Spinner className="flex justify-center p-6" />;
-      }
+const PaginatedQuery = ({ query, queryName, variables, count, children }) => {
+  const { error, loading, data, fetchMore, networkStatus } = useQuery(query, {
+    variables: {
+      ...variables,
+      count,
+      page: 1,
+    },
+    notifyOnNetworkStatusChange: true,
+  });
 
-      if (result.error) {
-        console.error(`${queryName} ERROR: ${result.error}`);
-        return <ErrorBlock error={result.error} />;
-      }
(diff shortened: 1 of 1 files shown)


resources/assets/components/PaginatedQuery.js
```diff
@@ -1,63 +1,62 @@
...
+
+          return {
+            ...previousResult,
+            [queryName]: [
+              ...previousResult[queryName],
+              ...fetchMoreResult[queryName],
+            ],
+          };
+        },
+      }),
```
DFurnes, line 57: It might be nice to extract this into a helper [like we did in Rogue](https://github.com/DoSomething/rogue/blob/9c83e89167f6e1853711e3a91440abc36500f754/resources/assets/helpers.js#L97-L111), so it could be used on "custom" paginated components too! (I'm not sure how much we do that in Phoenix, but it was really helpful [in that app](https://github.com/DoSomething/rogue/search?q=updateQuery).

DFurnes (approved):
Love it! 🎣

DFurnes:
> I initially wanted to tackle this here to see if we could scrap the network status mapping constant per [#2447 (comment)](https://github.com/DoSomething/phoenix-next/pull/2447#discussion_r534649070), but turns out we actually still need it here to differentiate between the initial data fetch and the subsequent "fetch more" requests to either render a plain spinner or just notify the child that we're in loading state.

Ugh yeahhhh, and one of the most frustrating things about this was that they're exposed as integer enums that were only described in their documentation (and not exported in code anywhere), so we had to hard-code this [in our constants](https://github.com/DoSomething/phoenix-next/blob/66e61dda9e3cc13b29fae6845decf2817c7c4564/resources/assets/constants/index.js#L10-L19) file. I wonder if that ever got cleaned up...

resources/assets/components/PaginatedQuery.js
mendelB: Waiiiiiit a second, we already have one here in Phoenix too! Updated in https://github.com/DoSomething/phoenix-next/pull/2449/commits/42f11d560a1313c5bc72ff018ffe927812f02b3c

mendelB:
> > I initially wanted to tackle this here to see if we could scrap the network status mapping constant per [#2447 (comment)](https://github.com/DoSomething/phoenix-next/pull/2447#discussion_r534649070), but turns out we actually still need it here to differentiate between the initial data fetch and the subsequent "fetch more" requests to either render a plain spinner or just notify the child that we're in loading state.
> 
> Ugh yeahhhh, and one of the most frustrating things about this was that they're exposed as integer enums that were only described in their documentation (and not exported from the library anywhere), so we had to hard-code this [in our constants](https://github.com/DoSomething/phoenix-next/blob/66e61dda9e3cc13b29fae6845decf2817c7c4564/resources/assets/constants/index.js#L10-L19) file. I wonder if that ever got cleaned up...

Blurge, [per the docs](https://www.apollographql.com/docs/react/api/react/hooks/#result), it seems like this still is the case 😞

resources/assets/components/PaginatedQuery.js
DFurnes: Oh even better!! 👏

PR #2452, opened 2020-12-04
Update GalleryBlockSignup.js

Dinnall:
### What's this PR do?

This pull request removes `event.preventDefault`. Now trackAnalyticsEvent is recorded correctly

Our `View Application` button worked correctly, but because `event.preventDefault` was still present it prevented
`trackAnalyticsEvent` from being fired.  

🚀 
...

...

### Relevant tickets

References [Pivotal #220141604](https://www.pivotaltracker.com/n/projects/2401401/stories/175155765/comments/220141604).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
+++ b/resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
@@ -67,9 +67,7 @@ const GalleryBlockSignup = ({ path, campaignId, campaignTitle }) => {
     `OneClickSignupCampaignId:${campaignId}`,
   );
 
-  const handleScholarshipCardShareClick = event => {
-    event.preventDefault();
-
+  const handleScholarshipCardShareClick = () => {
     trackAnalyticsEvent('clicked_scholarship_gallery_block_apply_now', {
       action: 'button_clicked',
       category: EVENT_CATEGORIES.siteAction,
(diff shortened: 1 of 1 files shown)


resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -67,9 +67,7 @@ const GalleryBlockSignup = ({ path, campaignId, campaignTitle }) => {
     `OneClickSignupCampaignId:${campaignId}`,
   );
 
-  const handleScholarshipCardShareClick = event => {
-    event.preventDefault();
-
+  const handleScholarshipCardShareClick = () => {
```
mendelB, line 70: I think removing this would cause the button click to redirect to the campaign which could prevent the signup from going through. I think we could solve the bug here by either passing through the event parameter in the callback functions:
```
onClick={(event) => {
        if (campaignData && campaignData.signups.length) {
          handleViewApplicationButtonClick();
        } else {
          handleScholarshipCardShareClick(event);
        }
      }}
```

Or, I think we can even have the parameter automatically passed through by simplifying it as such:
```
onClick={
        campaignData && campaignData.signups.length
          ? handleViewApplicationButtonClick
          : handleScholarshipCardShareClick
      }
```

DFurnes: +1, love the idea of using a ternary to swap out the event handler!

Dinnall: +1 the idea of using a ternary. Done

mendelB (approved):
⚡

PR #2456, opened 2020-12-07
Update Recommend Campaigns Gallery Filters

mendelB:
### What's this PR do?

This pull request adds new filtering to the `RecommendedCampaignsGallery` used by the `ShowSubmissionPage` to exclude 
- env-variable-configured campaigns to hide
- the pages submission post's campaign ID (we don't want to recommend the campaign they've just completed!)
- group campaigns (which we typically want to exclude from site galleries)

It also adds some basic integration tests. I know we anticipate gating this page which could cause Cypress issues, but I'm hoping that we can address [Pivotal #175313574](https://www.pivotaltracker.com/n/projects/2401401/stories/175313574) as we add the gate so this test will still work.

commit-by-commit if you so fancy!
Perhaps keep an eye on the logic we're using to exclude the campaign IDs, I think that's the most interesting bit!

We've added these new filters to our Algolia `searchCampaigns` query (see [here](https://git.io/JI48K)). A follow up task is to add these filters to the `PaginatedCampaignGallery` as well.

### Relevant tickets

References [Pivotal #175358477](https://www.pivotaltracker.com/story/show/175358477).

### Checklist

### DIFF
--- a/resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
+++ b/resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
@@ -1,19 +1,22 @@
 import React from 'react';
 import { get } from 'lodash';
 import gql from 'graphql-tag';
+import PropTypes from 'prop-types';
 
 import Query from '../../Query';
+import { siteConfig } from '../../../helpers';
 import GalleryBlock from '../../blocks/GalleryBlock/GalleryBlock';
 import { scholarshipCardFragment } from '../../utilities/ScholarshipCard/ScholarshipCard';
 
 const SCHOLARSHIP_CAMPAIGNS_QUERY = gql`
-  query ScholarshipCampaignsQuery {
+  query ScholarshipCampaignsQuery($excludeIds: [Int]) {
     campaigns: searchCampaigns(
       perPage: 3
       hasWebsite: true
       hasScholarship: true
       isOpen: true
       orderBy: "start_date,desc"
+      excludeIds: $excludeIds
     ) {
       edges {
         node {
@@ -29,8 +32,16 @@ const SCHOLARSHIP_CAMPAIGNS_QUERY = gql`
   ${scholarshipCardFragment}
 `;
 
-const RecommendedCampaignsGallery = () => (
-  <Query query={SCHOLARSHIP_CAMPAIGNS_QUERY}>
+const RecommendedCampaignsGallery = ({ variables }) => (
+  <Query
+    query={SCHOLARSHIP_CAMPAIGNS_QUERY}
+    variables={{
+      excludeIds: [
+        ...(variables.excludeIds || []),
+        ...siteConfig('hide_campaign_ids', []).map(id => Number(id)),
+      ],
+    }}
+  >
     {result => (
       <GalleryBlock
         blocks={(get(result, 'campaigns.edges') || []).map(
@@ -43,5 +54,15 @@ const RecommendedCampaignsGallery = () => (
     )}
   </Query>
 );
+
+RecommendedCampaignsGallery.propTypes = {
+  variables: PropTypes.shape({
+    excludeIds: PropTypes.arrayOf(PropTypes.number),
+  }),
+};
+
+RecommendedCampaignsGallery.defaultProps = {
+  variables: {},
+};
 
 export default RecommendedCampaignsGallery;
--- a/resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
+++ b/resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
@@ -14,6 +14,7 @@ const SCHOLARSHIP_CAMPAIGNS_QUERY = gql`
       perPage: 3
       hasWebsite: true
       hasScholarship: true
+      isGroupCampaign: false
       isOpen: true
       orderBy: "start_date,desc"
       excludeIds: $excludeIds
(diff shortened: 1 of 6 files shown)


resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
```diff
@@ -28,11 +33,21 @@ const SCHOLARSHIP_CAMPAIGNS_QUERY = gql`
...
 
-const RecommendedCampaignsGallery = () => (
-  <Query query={SCHOLARSHIP_CAMPAIGNS_QUERY}>
+const RecommendedCampaignsGallery = ({ variables }) => (
+  <Query
+    query={SCHOLARSHIP_CAMPAIGNS_QUERY}
+    variables={{
+      excludeIds: [
+        ...(variables.excludeIds || []),
+        ...siteConfig('hide_campaign_ids', []).map(id => Number(id)),
```
mendelB, line 42: I wonder if we ought to coerce these into Integers in the [config file itself](https://github.com/DoSomething/phoenix-next/blob/3d8b335192a336d8f6aa9e46c46642c9825ce90e/config/site.php#L22) since I assume we'll always want this to be a list of numbers 🤔

aaronschachter: That sounds like a good idea! Definitely would DRY should we never need the `hide_campaign_ids` anywhere else (which we do, I think?)

resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
```diff
@@ -28,11 +33,21 @@ const SCHOLARSHIP_CAMPAIGNS_QUERY = gql`
...
+      ],
+    }}
+  >
     {result => (
       <GalleryBlock
-        blocks={result.campaigns.edges.map(edge => edge.node.campaignWebsite)}
+        blocks={(get(result, 'campaigns.edges') || []).map(
```
aaronschachter, line 48: Nit - any reason we can't specify the empty array as the default value? I think it'd be a slight improvement for readability...  
```suggestion
        blocks={get(result, 'campaigns.edges', []).map(
```

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
```diff
@@ -35,7 +36,7 @@ const ShowSubmissionPage = ({ match }) => {
     },
   });
 
-  const postImageUrl = get(postData, 'post.url', null);
+  const { url: postImageUrl, campaignId } = get(postData, 'post') || {};
```
aaronschachter, line 39: Could this use the same default value?
```suggestion
  const { url: postImageUrl, campaignId } = get(postData, 'post', {});
```

resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
```diff
@@ -41,4 +56,14 @@ const RecommendedCampaignsGallery = () => (
   </Query>
 );
 
+RecommendedCampaignsGallery.propTypes = {
+  variables: PropTypes.shape({
+    excludeIds: PropTypes.arrayOf(PropTypes.number),
+  }),
```
aaronschachter, line 62: Curious about the reasoning behind sending a `variables` shape instead of just a single array prop like `excludeCampaignIds`.

aaronschachter (approved):

resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
mendelB: Surprisingly, (Or I suppose this actually makes sense), `get` will not return the default value if the object's value is assigned to `null` (which is the edge case here when GraphQL returns `null` results instead of the empty array we'd expect).

```js
> const lodash = require('lodash')
undefined
> lodash.get({}, 'nothing', [])
[]
> lodash.get({nothing: null}, 'nothing', [])
null
```

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
mendelB: I experienced the same issue here per https://github.com/DoSomething/phoenix-next/pull/2456#discussion_r538770345

resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
mendelB: I was thinking it might be prudent to give room for additional variables to be specified in the future -- do you think that's overkill? My rationale was also that this is similar to the way we've handle variables passed through to the [`PaginatedCampaignGallery`](https://github.com/DoSomething/phoenix-next/blob/e7b602ddaa1b5da21e150fe372692883343ae291/resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js#L187-L192), but I could hear the argument that individual props in general would be a better idea?

resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
mendelB: Yeah! And I did end up needing to repeat this logic in #2457. I'll clean them both up in a follow-up!

resources/assets/components/pages/ShowSubmissionPage/RecommendedCampaignsGallery.js
aaronschachter: My preference would be for the individual prop since it's all we need for now and I think would keep the code a little simpler... but really not a biggie.

mendelB: I think it's a good idea! I can tackle this in the same follow-up per https://github.com/DoSomething/phoenix-next/pull/2456#discussion_r538774199

PR #2460, opened 2020-12-09
Add Show Submission Page Redirect Following Photo Post Submission

mendelB:
### What's this PR do?

This pull request adds a redirect to the Show Submission Page (`/us/posts/:id`) following a successful photo-post submission.

👀 

I toyed with using a `<Redirect />` instead of the full page refresh but the UX win was dampened by the window scroll not being adjusted on the next page, and it also complicated the logic a little bit. (I also wasn't sure about the analytics ramifications).

I didn't DRY up the photo post logic in the tests (we're running the photo post submission logic twice now) since we'll be combining the two once this launches.

### Relevant tickets

References [Pivotal #176056956](https://www.pivotaltracker.com/story/show/176056956).

### Checklist

Kapture 2020-12-09 at 15 30 36

### DIFF
--- a/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
+++ b/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
@@ -4,6 +4,7 @@ import React from 'react';
 import gql from 'graphql-tag';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import { Redirect } from 'react-router-dom';
 import { get, has, invert, mapValues } from 'lodash';
 
 import PostForm from '../PostForm';
@@ -66,15 +67,13 @@ class PhotoSubmissionAction extends PostForm {
       // Resetting the submission item so that this won't be triggered continually for further renders.
       nextProps.resetPostSubmissionItem(nextProps.id);
 
-      if (featureFlag('post_confirmation_page')) {
-        // Redirect the user to the post submission page.
-        window.location = `/us/posts/${response.data.id}?submissionActionId=${nextProps.id}`;
-      }
+      // If the feature is toggled on, we'll redirect to the show submission page instead of displaying the affirmation modal.
+      const redirectToSubmissionPage = featureFlag('post_confirmation_page');
 
       return {
         shouldResetForm: true,
-        // Don't display the modal if we're redirecting the user to the post submission page.
-        showModal: !featureFlag('post_confirmation_page'),
+        redirectToSubmissionPage,
+        showModal: !redirectToSubmissionPage,
       };
     }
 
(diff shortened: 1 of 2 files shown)


resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -66,9 +66,15 @@ class PhotoSubmissionAction extends PostForm {
       // Resetting the submission item so that this won't be triggered continually for further renders.
       nextProps.resetPostSubmissionItem(nextProps.id);
 
+      if (featureFlag('post_confirmation_page')) {
+        // Redirect the user to the post submission page.
+        window.location = `/us/posts/${response.data.id}?submissionActionId=${nextProps.id}`;
```
aaronschachter, line 71: I know that when we added the Quiz Result page, @DFurnes suggested we use the `Redirect` and not this method. I think potentially because this method does affect analytics negatively? I don't remember what the rationale was.

mendelB: Hmm, I figured the analytics consequence would've been via the `Redirect` as opposed to a typical page visit through this method? There was some wonkiness to the page scroll since the uploader's typically all the way down and this persists through the `Redirect` which was my main reason for opting out of it (though that could be something we address globally, I saw that they address this in their [React Router docs](https://reactrouter.com/web/guides/scroll-restoration))

aaronschachter: Makes sense.. we can always change this up later if we need the `Redirect` here.

aaronschachter (approved):

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
DFurnes: We prefer the `<Redirect>` because it gets us client-side navigation (with no full page reload, so this page will load a lot quicker). It looks like we could get the expected "scroll to top" behavior for client-side routing by following [this example](https://reactrouter.com/web/guides/scroll-restoration).

DFurnes: Both should work the same for analytics, so definitely not a high priority fix!

mendelB: Unfortunately, I don't think their simple scroll suggestion wouldn't cut it for us wholesale since we use tabbed navigation for Campaign pages (and some scroll concierge stuff - also in Quizzes). They do have a recommendation for individual use scrolling. 

But in better news, the work in https://github.com/DoSomething/phoenix-next/commit/031f707ab860671794c0a90e9c003570d7827410 which added a loading state `<Placeholder />` to this component rendered the scroll issue moot!

So big UI win overall. Updated in https://github.com/DoSomething/phoenix-next/pull/2460/commits/cc217ca8bdb3ae51e667d4b9f85b9c89f140818b.

aaronschachter (approved):
Sweet!

PR #2461, opened 2020-12-10
gated and ungated sixpack

Dinnall:
### What's this PR do?

This pull request adds a sixpack A/B test on gated vs ungated campaign

ungated : 
http://phoenix.test/us/campaigns/test-teens-for-jeans/action?sixpack-force-ungated_or_gated_campaign=ungated_campaign

gated:

http://phoenix.test/us/campaigns/test-teens-for-jeans/action?sixpack-force-ungated_or_gated_campaign=gated_campaign
...

...

### Relevant tickets

References [Pivotal #175560547](https://www.pivotaltracker.com/n/projects/2401401/stories/175560547).

### Checklist

### DIFF
--- a/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
+++ b/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
@@ -103,8 +103,6 @@ const CampaignRoute = props => {
               );
             }
 
-            // @TODO: Add support for SixpackExperiment components (https://bit.ly/2T99sUl).
-
             if (campaignIdToNumber === 9108 || campaignIdToNumber === 9001) {
               return (
                 <SixpackExperiment
--- a/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
+++ b/resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
@@ -12,6 +12,23 @@ import LandingPage from '../../pages/LandingPage/LandingPage';
 import PostSignupModal from '../../pages/PostSignupModal/PostSignupModal';
 import CampaignClosedPage from '../../pages/CampaignPage/CampaignClosedPage';
 import CampaignPageContainer from '../../pages/CampaignPage/CampaignPageContainer';
+import SixpackExperiment from '../../utilities/SixpackExperiment/SixpackExperiment';
+
+export const UNGATED_SESSION_KEY = 'ungated_session';
+
+const UngatedCampaignRedirect = props => {
+  const { baseUrl } = props;
+
+  window.sessionStorage.setItem(UNGATED_SESSION_KEY, JSON.stringify('ungated'));
+  return (
+    <Redirect
+      to={{
+        pathname: join(baseUrl, 'action'),
+        search: window.location.search,
+      }}
+    />
+  );
+};
 
 const CampaignRoute = props => {
   const {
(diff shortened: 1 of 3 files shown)


resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
```diff
@@ -23,12 +40,12 @@ const CampaignRoute = props => {
     location,
     match,
     shouldShowAffirmation,
+    campaignId,
   } = props;
 
   const isClosed = isCampaignClosed(endDate);
-
+  const campaignIdToNumber = Number(campaignId);
```
DFurnes, line 47: Why do we need to cast this to a number? Is the type that it comes in as inconsistent?

resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
```diff
@@ -86,8 +103,29 @@ const CampaignRoute = props => {
               );
             }
 
+            // @TODO: Add support for SixpackExperiment components (https://bit.ly/2T99sUl).
```
DFurnes, line 106: We can probably remove this TODO - looks like the card it pointed at has been deleted/archived.

resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
```diff
@@ -12,6 +12,23 @@ import LandingPage from '../../pages/LandingPage/LandingPage';
...
 import CampaignClosedPage from '../../pages/CampaignPage/CampaignClosedPage';
 import CampaignPageContainer from '../../pages/CampaignPage/CampaignPageContainer';
+import SixpackExperiment from '../../utilities/SixpackExperiment/SixpackExperiment';
+
+export const UNGATED_SESSION_KEY = 'ungated_session';
+
+const UngatedCampaignRedirect = props => {
+  const { baseUrl } = props;
+
+  window.sessionStorage.setItem(UNGATED_SESSION_KEY, JSON.stringify('ungated'));
```
DFurnes, line 22: Making sure I follow: is the reason that we use session storage here so that we can customize the action page based on whether or not we're opted into this experiment? If so, might be nice to drop a quick comment explaining that & pointing to `CampaignBanner.js` for anyone else seeing this code out of context!

DFurnes (approved):
Really excited to see the results of this experiment! 📊

resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
DFurnes: This makes me wish that Sixpack had an endpoint to check _if_ you're participating in a given experiment without also opting in... like a `is_participating` equivalent of `participate`! I wonder if we could build that sort of API on our end...

Dinnall: Spot on! I do wish sixpack had that as well, but I do think we can probably create a helper in another ticket.

resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
Dinnall: Correct! just to make sure I used `typeof` and it was coming back as a string. To be on the safe side added `Number`

resources/assets/components/Campaign/CampaignRoute/CampaignRoute.js
mendelB, line 22: @Dinnall I think we should specify the Campaign ID as the value here and use that when we check for this value in the `CampaignBanner` otherwise, folks opted into the ungated experiment would be able to access _any_ campaign /action page as an ungated experience. 

(Admittedly sort of an edge case based on typical traffic (folks will typically not hit the /action page directly and would be met by the Landing Page), so not urgent!)

PR #2464, opened 2020-12-10
Bug fix - default value on hide campaign IDs config

mendelB:
### What's this PR do?

This pull request fixes a bug flagged by @lindsayrmaher where if the `DS_HIDE_CAMPAIGN_IDS` is not in the `.env`, it'll explode since `explode` is expecting a String not an Array.

💥 

### Relevant tickets
https://dosomething.slack.com/archives/CUQMU4Q6B/p1607619687006200


Dinnall (approved):
LGTM 👍   quick fix and a good catch by Lindsay!

PR #2474, opened 2020-12-15
Fix Show Submission Page Redirect Bug

mendelB:
### What's this PR do?

This pull request fixes a - bit of an edge case - bug wherein because we were redirecting with React `Redirect` to the show submission page, users on the show submission page who would navigate to a new refreshed page - e.g. a campaign in the recommended gallery - and then use the browser back button to go back to the campaign would face an error because the Redux campaign state was missing and the browser didn't force a redirect.

👀 

There are probably other alternative ways we can solve this, but this seems like the easiest quickest fix so we can deploy before the code freeze and maintain this new feature.

### Relevant tickets
https://dosomething.slack.com/archives/CP2D7UGAU/p1608052149017300?thread_ts=1608051501.016800&cid=CP2D7UGAU
https://github.com/DoSomething/phoenix-next/pull/2460#discussion_r539683322

Kapture 2020-12-15 at 14 17 25


Dinnall (approved):
Love this cleanup! Much easier to follow

PR #2481, opened 2020-12-18
Add rewards table to rewards tab

lindsayrmaher:
### What's this PR do?

This pull request adds a brand new table to the rewards tab in user's profiles! I will also be working on adding some Cypress tests for this new addition in a new PR.

Let me know whether there are less hacky ways to implement any of this! Tables just feel hacky to me, so it's hard for me to see spots we could DRY up or refactor. Let a girl know!!

Some screenshots:

**Large Screen:**
Screen Shot 2020-12-18 at 10 49 21 AM

Screen Shot 2020-12-18 at 10 23 34 AM

Screen Shot 2020-12-18 at 10 23 12 AM

**Medium Screen:**
Screen Shot 2020-12-18 at 10 21 17 AM

**Small Screen:**
Screen Shot 2020-12-18 at 10 20 03 AM

Right now it's based on a hardcoded number, but that will change once we are able to make updates to the backend to support this work. We are getting some front end pieces in place now while the backend project is getting worked on. 

### Relevant tickets

References [Pivotal #175855270](https://www.pivotaltracker.com/story/show/175855270).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Badges/RewardLevelsTable.js
+++ b/resources/assets/components/pages/AccountPage/Badges/RewardLevelsTable.js
@@ -26,7 +26,7 @@ const RewardLevelsTable = () => {
           }
         >
           Doer <br />{' '}
-          <span className="font-normal text-xs md:text-sm">2 badges</span>
+          <span className="font-normal text-xs md:text-base">2 badges</span>
         </TableCellCenter>
 
         <TableCellCenter
@@ -37,14 +37,14 @@ const RewardLevelsTable = () => {
           }
         >
           SuperDoer <br />{' '}
-          <span className="font-normal text-xs md:text-sm">4 badges</span>
+          <span className="font-normal text-xs md:text-base">4 badges</span>
         </TableCellCenter>
 
         <TableCellCenter
           className={badges >= 6 ? 'bg-yellow-500 bg-opacity-25' : 'bg-white'}
         >
           Legend <br />{' '}
-          <span className="font-normal text-xs md:text-sm">6 badges</span>
+          <span className="font-normal text-xs md:text-base">6 badges</span>
         </TableCellCenter>
       </tr>
     </TableHeader>
--- a/resources/assets/components/pages/AccountPage/Badges/RewardLevelsTable.js
+++ b/resources/assets/components/pages/AccountPage/Badges/RewardLevelsTable.js
@@ -3,17 +3,17 @@ import tw from 'twin.macro';
 
 import SectionHeader from '../../../utilities/SectionHeader/SectionHeader';
 
-const Table = tw.table`md:w-3/4 my-6`;
+const Table = tw.table`my-6 w-full lg:w-3/4`;
 const TableHeader = tw.thead`font-bold p-4 pr-6 text-center w-full border-solid border-b-4 border-gray-400`;
 const TableCellLeft = tw.td`p-3 text-sm text-left md:text-base border-solid border-b border-gray-400`;
 const TableCellLeftBottom = tw.td`p-2 text-sm text-left md:text-base`;
-const TableCellCenter = tw.td`p-2 text-sm text-center md:text-base border-solid border-l border-b border-gray-400 items-center justify-center`;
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/AccountPage/Badges/RewardLevelsTable.js
```diff
@@ -0,0 +1,163 @@
...
+    </TableHeader>
+  );
+
+  const userLevelLabel = badgeNumber => {
+    let userLevel = 'Doer';
+    if (badgeNumber >= 6) {
+      userLevel = 'Legend';
+    } else if (badgeNumber > 3) {
+      userLevel = 'SuperDoer';
+    }
```
katiecrane, line 59: Clever way to handle this logic here!

resources/assets/components/pages/AccountPage/Badges/RewardLevelsTable.js
```diff
@@ -0,0 +1,163 @@
...
+            <TableCellLeft>2x Scholarship entries</TableCellLeft>
+
+            <TableCellCenter
+              className={
+                badges >= 2 && badges < 4
+                  ? 'bg-teal-500 bg-opacity-25'
+                  : 'bg-white'
+              }
+            >
+              {badges >= 2 && badges < 4 ? <TableMarker /> : null}
```
katiecrane, line 86: I thought we would have these table markers even if it's not the user's current level, so they can see what other levels get. Did I interpret the designs wrong at some point?

resources/assets/components/pages/AccountPage/Badges/RewardLevelsTable.js
```diff
@@ -0,0 +1,163 @@
...
+      <Table>
+        {header}
+
+        <tbody data-testid="rewards-levels-table-body">
+          <tr className="w-full">
+            <TableCellLeft>2x Scholarship entries</TableCellLeft>
+
+            <TableCellCenter
+              className={
+                badges >= 2 && badges < 4
```
katiecrane, line 81: I'm not sure if it would be better or worse, but did you consider using something like this here? I was thinking it miiiight be easier to think about since you already have that logic above?
```suggestion
                {userLevelLabel(badges)} === 'Doer'
```

resources/assets/components/pages/AccountPage/Badges/RewardLevelsTable.js
lindsayrmaher: OH no I think maybe I misunderstood! That makes those easier lol 😂

katiecrane: Yeah, I think it's a little confusing because originally in the designs the markers were going to be cumulative, but because of the nature of the rewards we're doing we'll end up with one marker per level. I believe that's what we landed on!

lindsayrmaher: updated in 8bf9b39

resources/assets/components/pages/AccountPage/Badges/RewardLevelsTable.js
lindsayrmaher: i think this is a good update! i shall do.

katiecrane (approved):
looks great! (also hi welcome back!)

resources/assets/components/pages/AccountPage/Badges/RewardLevelsTable.js
```diff
@@ -0,0 +1,177 @@
...
+  const header = (
+    <TableHeader>
+      <tr>
+        <TableCellLeft />
+
+        <TableCellCenter
+          className={
+            userLevelLabel(badges) === 'Doer'
+              ? 'bg-teal-500 bg-opacity-25'
+              : 'bg-white'
```
mendelB, line 40: @lindsayrmaher I just bumped into a very interesting looking way to avoid repeating our logic across each cell [on SO!](https://stackoverflow.com/a/27234542/5657278). Looks like if we use the [`<colgroup>` tag](https://www.w3schools.com/tags/tag_colgroup.asp) we can define it in one singular place up top 🎉

lindsayrmaher: oh snap!! i will try it out

PR #2483, opened 2020-12-23
Helper cleanup for social media SDKs

weerd:
### What's this PR do?

This pull request does some cleanup in the **helper/index.js** file to simplify it a bit. It pulls out all the social media related SDKs and functions into their own respective helper file for each (Facebook, Twitter and Snapchat).

### How should this be reviewed?

👀  Nothing crazy, just moving stuff around and updating imports based on the new files.

### Relevant tickets

References [Pivotal #173019917](https://www.pivotaltracker.com/story/show/173019917).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.

PR #2484, opened 2020-12-23
Helper cleanup env

weerd:
### What's this PR do?

This pull request continues the work started in #2483 to clean up the **helpers/index.js** file. It separates out any helper functions that deal with the `window.ENV` object and puts them all in their own **helpers/env.js** file. Similar to how most (not all but cause we need to do some cleanup) of the `window.AUTH` helper methods are in the **helpers/auth.js**.

Since there was a test for `siteConfig()` helper method, I added additional tests for the rest of the functions in this new helper file. 🙌  🚥 

👀  Looks like a lot of files, but the brunt of it is all just fixing up import statements in lots of files to match the new location for the imported functions.

### Relevant tickets

### Checklist


DFurnes (approved):
Love it!

PR #2487, opened 2021-01-04
Moving URL helpers into separate file.

weerd:
### What's this PR do?

This pull request continues the work from #2483 and #2484 to clean up the **helpers/index.js** file. It separates out any helper functions that deal with URLs, including queries, paths and UTMs.

### How should this be reviewed?

👀  Looks like a lot of files, but the brunt of it is all just fixing up import statements in lots of files to match the new location for the imported functions.

### Relevant tickets

References [Pivotal #173019917](https://www.pivotaltracker.com/story/show/173019917).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.

PR #2488, opened 2021-01-04
Moving Contentful helpers into separate file.

weerd:
### What's this PR do?

This pull request continues the work from #2483, #2484 and #2487 to clean up the **helpers/index.js** file. It separates out any helper functions that deal with Contentful.

👀  Looks like a lot of files, but the brunt of it is all just fixing up import statements in lots of files to match the new location for the imported functions.

### Relevant tickets

### Checklist


resources/assets/helpers/index.js
```diff
@@ -489,6 +445,7 @@ export function openDialog(href, callback, width = 550, height = 420) {
  * @param  {Object} state
  * @param  {String} identifier
  * @return {Object|Undefined}
+ * @todo rename to findContentfulEntryInCampaign and move to helpers/campaign.js
```
weerd, line 448: This might be a bit subjective... although the method name would seem to imply it should belong in the **helpers/contentful.js**, the contentful entries being searched for are within the `campaign` object in state and thus tightly coupled to it. So my thoughts were to rename it to better explain where it's searching for the entries and that it should be moved to **helpers/campaign.js**.

lindsayrmaher: This makes sense to me, i think the re naming makes it clearer _where_ we are looking and that it's not necessarily a contentful helper. I plus one the re name and moving to the campaign helpers.

lindsayrmaher (approved):
looking good to me!

PR #2489, opened 2021-01-05
Updating Authentication to Front End for Account Pages

lindsayrmaher:
### What's this PR do?

This pull request removes middleware from a php route and adds an authentication gate around the "account" route. This should allow us to check whether a user is logged in before navigating to their account page successfully.

Am I missing anything here? It seems to work as expected. Does it makes sense to create a couple of tests just to see that we can visit the page if we're logged in and we get redirected if we don't? I'm guessing the answer to this Q is yes lol

In order to add Cypress tests for _any_ of the account pages, we need to be able to authenticate via the front end so that cypress can mock that behavior. this is something we really wanted to update since we have quite a few tabs now in the account page.

### Relevant tickets

References [Pivotal #](https://www.pivotaltracker.com/story/show/175313574).

### Checklist

### DIFF
--- a/cypress/integration/profile-page.js
+++ b/cypress/integration/profile-page.js
@@ -1,0 +1,18 @@
+import { userFactory } from '../fixtures/user';
+
+describe('User Profile Page', () => {
+  beforeEach(() => cy.configureMocks());
+  /** @test */
+  it('Renders Voter Registration Call to Action in Registration Status', () => {
+    const user = userFactory();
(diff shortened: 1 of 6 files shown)


DFurnes:
> Does it makes sense to create a couple of tests just to see that we can visit the page if we're logged in and we get redirected if we don't? I'm guessing the answer to this Q is yes lol

Yeah, that'd be great!!

DFurnes (approved):
🔏

PR #2490, opened 2021-01-06
Auth helper cleanup OR "There Can Only Be One isAuthenticated()" ⚔️

weerd:
### What's this PR do?

This pull request continues the work from the following PRs to clean up the **helpers/index.js** file. 
- #2483
- #2484
- #2487 
- #2488

It separates out any helper functions that deal `window.AUTH`. A **helpers/auth.js** file already existed, but there were a few helper methods in the **helpers/index.js** that still needed to be moved.

There were also 3 different `isAuthenticated()` methods; two that did the exact same thing and another that just selected the `isAuthenticated` from the Redux state. This has now been cleaned up and reduced to the single helper function in **helpers/auth.js**.

👀  Looks like a lot of files, and your eyes are not fooling you!

The brunt of it is all just fixing up import statements in lots of files to match the new location for the imported functions. 

Along with consolidating the `isAuthenticated()` Redux selector function, I was also able to consolidate the `getUserId()` Redux selector function (both from **selectors/user.js**) and remove them to only use the functions in **helpers/auth.js** with the same name.

The aforementioned consolidation showed that all the prior calls to these 2 functions to set state in component container files were not really necessary (the state never changes as evidenced by the [empty reducer](https://github.com/DoSomething/phoenix-next/blob/712a2f734747ec0ec89a2d088f705be0104cf7c7/resources/assets/reducers/user.js) and could just be called directly in the component files themselves. So the other large majority of the changes are edits to remove those items from component container files and call the helper functions directly in the corresponding component files.

As a side bonus, we're aiming to slowly move away from using Redux, so this helps a bit in a teeny-tiny way by removing some cruft in the container files and even removed the need for 2 container files entirely! ✨ 

### Relevant tickets

### Checklist


resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -120,6 +121,8 @@ class PhotoSubmissionAction extends PostForm {
       whyParticipatedValue: '',
       hoursValue: '',
     };
+
+    this.userId = getUserId();
```
weerd, line 125: There are a few other instances of similar updates in other components. My thinking is that I wanted to make it very apparent in the component code that the `userId` is not part of the `props` and gets assigned via the helper method. Figured it was still helpful to have it as a variable throughout the component code instead of multiple calls to `getUserId()`, and decided assigning it in the constructor made sense.

DFurnes (approved):
Epic cleanup! ⚔️

PR #2495, opened 2021-01-07
Helper cleanup File and Analytics

weerd:
### What's this PR do?

This pull request continues the work from the following PRs to clean up the **helpers/index.js** file. 
- #2483
- #2484
- #2487 
- #2488
- #2490 
- #2492 
- #2493 

It separates out any helper functions that deal with file/image manipulation into the **helpers/file.js** and the Sixpack related helper into the **helpers/analytics.js**.

### How should this be reviewed?

👀 

### Relevant tickets

References [Pivotal #173019917](https://www.pivotaltracker.com/story/show/173019917).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.

PR #2493, opened 2021-01-07
Helper cleanup display

weerd:
### What's this PR do?

This pull request continues the work from the following PRs to clean up the **helpers/index.js** file. 
- #2483
- #2484
- #2487 
- #2488
- #2490 
- #2492 

It separates out any helper functions that deal with display related needs into a **helpers/display.js** file. I debated what to call this... I originally thought **ui.js**, but figured **display.js** is more obvious. Lemme know if you have thoughts.

### How should this be reviewed?

👀 

### Relevant tickets

References [Pivotal #173019917](https://www.pivotaltracker.com/story/show/173019917).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.

PR #2492, opened 2021-01-06
Moving straggling campaign helpers into dedicated file.

weerd:
### What's this PR do?

This pull request continues the work from the following PRs to clean up the **helpers/index.js** file. 
- #2483
- #2484
- #2487 
- #2488
- #2490 

It separates out any helper functions that deal with campaigns. A **helpers/campaign.js** already existed but there were a few straggling helper functions in the **helpers/index.js** that needed to be moved as well.

### How should this be reviewed?

👀 

### Relevant tickets

References [Pivotal #173019917](https://www.pivotaltracker.com/story/show/173019917).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.

PR #2496, opened 2021-01-08
Helper cleanup for data manipulation helpers

weerd:
### What's this PR do?

This pull request continues the work from the following PRs to clean up the **helpers/index.js** file. 
- #2483
- #2484
- #2487 
- #2488
- #2490 
- #2492 
- #2493 
- #2495 

It separates out any helper functions that deal with data manipulation into a dedicated **helpers/data.js** file. Initially I added the helpers that dealt more with objects and their properties but ended up including ones that also manipulate strings because they technically are manipulating data and would keep us from having a separate **string.js** as well for a small number of methods.

👀 

### Relevant tickets

### Checklist

### DIFF
--- a/resources/assets/helpers/data.js
+++ b/resources/assets/helpers/data.js
@@ -78,7 +78,7 @@ export function generateUniqueId() {
 /**
  * Return a boolean indicating whether the provided argument is an empty array.
  *
- * @param  {Mixed}  data
+ * @param  {*} data
  * @return {Boolean}
  */
 export function isEmptyArray(data) {
@@ -92,7 +92,7 @@ export function isEmptyArray(data) {
 /**
  * Return a boolean indicating whether the provided argument is an empty string.
  *
- * @param  {Mixed}  data
+ * @param  {*} data
  * @return {Boolean}
  */
 export function isEmptyString(data) {
--- a/resources/assets/helpers/data.js
+++ b/resources/assets/helpers/data.js
@@ -12,8 +12,8 @@ import {
 } from 'lodash';
 
 /**
- * Convert an int to a string. (Supports 0-10)
- * @param  {int} number
+ * Convert an number to a string. (Supports 0-10)
+ * @param  {Number} number
  * @return {String}
  */
 export function convertNumberToWord(number) {
--- a/resources/assets/helpers/data.js
+++ b/resources/assets/helpers/data.js
@@ -1,3 +1,5 @@
+/* global Date, JSON, Math, Object */
+
 import {
   isArray,
   isEmpty,
--- a/resources/assets/helpers/data.js
+++ b/resources/assets/helpers/data.js
@@ -1,0 +1,198 @@
+import {
+  isArray,
+  isEmpty,
+  isNil,
+  isNull,
+  isObjectLike,
+  isUndefined,
+  mapValues,
+  omitBy,
+} from 'lodash';
+
+/**
+ * Convert an int to a string. (Supports 0-10)
+ * @param  {int} number
+ * @return {String}
+ */
+export function convertNumberToWord(number) {
+  switch (number) {
+    case 0:
+      return 'zero';
+    case 1:
+      return 'one';
+    case 2:
+      return 'two';
+    case 3:
+      return 'three';
+    case 4:
+      return 'four';
+    case 5:
+      return 'five';
+    case 6:
+      return 'six';
+    case 7:
+      return 'seven';
+    case 8:
+      return 'eight';
+    case 9:
+      return 'nine';
+    case 10:
+      return 'ten';
+    default:
+      throw new Error('Number out of range');
+  }
+}
+
+/**
+ * Return a string with tokens replaced by specified values.
+ *
+ * @param  {String} string
+ * @param  {Object} tokens
+ * @return {String}
+ */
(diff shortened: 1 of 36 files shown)


weerd:
Set this to draft; waiting on #2493 & #2495 to get merged since this was branched off of it. Once set the number of file changes will drop! 😅

resources/assets/helpers/display.js
```diff
@@ -8,6 +8,32 @@ import tailwindVariables from '../../../tailwind.variables';
...
   'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7';
 
+/**
+ * Get a formatted name (small, medium, large)
+ * of the current display size.
+ *
+ * @param  {int} width - Defaults to the screen width
+ * @return {String}
+ */
+export function getFormattedScreenSize(screenWidth = window.innerWidth) {
```
weerd, line 18: One I had missed moving into *helpers/display.js*

resources/assets/helpers/data.js
```diff
@@ -0,0 +1,200 @@
...
+  isNull,
+  isObjectLike,
+  isUndefined,
+  mapValues,
+  omitBy,
+} from 'lodash';
+
+/**
+ * Convert an int to a string. (Supports 0-10)
+ * @param  {int} number
```
aaronschachter, line 16:
```suggestion
 * @param  {Number} number
```
Is this intentionally an `{int}`?

resources/assets/helpers/data.js
```diff
@@ -0,0 +1,200 @@
...
+}
+
+/**
+ * Return a boolean indicating whether the provided argument is an empty string.
+ *
+ * @param  {Mixed}  data
+ * @return {Boolean}
+ */
+export function isEmptyString(data) {
+  return data === '';
```
aaronschachter, line 99: Minor, but I thought declaring the argument as `string` made this a little more readable...

resources/assets/helpers/data.js
```diff
@@ -0,0 +1,200 @@
...
+/**
+ * Pluralize a noun.
+ *
+ * @param {Number} quantity
+ * @param {String} singular
+ * @param {String} plural
+ */
+export function pluralize(quantity, singular, plural) {
+  return quantity === 1 ? singular : plural;
+}
```
aaronschachter, line 138: Out of scope, but curious why we didn't go with [pluralize](https://www.npmjs.com/package/pluralize) as I think we do in our GraphQL app...

aaronschachter (approved):
Looks good to me. I'm curious what you think about using `util` instead of `data` -- both are pretty vague, we use `util` over in very similar looking [Gambit helpers](https://github.com/DoSomething/gambit/blob/main/lib/helpers/util.js).

resources/assets/helpers/data.js
weerd: Oh I think this is from me playing around and when I was initially debating whether to have both a `data.js` file and `string.js` file and this function felt like it belonged more with the other similar "data" ones. But eventually decided to put all the string ones into "data" anyhoo... so I can change this back so it's clearer!

resources/assets/helpers/data.js
weerd: 🤷‍♂️  Not really sure and/or don't remember well. Mostly moving stuff around for this cleanup. I can check and see if I notice anything in code history as to reasoning 🤔

resources/assets/helpers/data.js
weerd: I just copy-🍝 the code, but good catch! Will update.

resources/assets/helpers/data.js
weerd: Kinda going back and forth on this one since it's technically `mixed` data and no guarantee that what is passed in is going to be a `string`, so I'm leaning towards `data` being the more appropriate argument name 🤔  Thoughts?

resources/assets/helpers/data.js
weerd: This is the [PR](https://github.com/DoSomething/phoenix-next/pull/234) where it was introduced. I suspect maybe the logic needed was just a couple lines and got the job done so not a need to pull in a package. But not really sure.

resources/assets/helpers/data.js
aaronschachter: Re: mixed data, I haven't really seen that `Mixed` type before. According to JSDoc, it looks like `{*}`is recommended for params of any type.

Because some of other functions that operate on strings have the `string` name, I thought it'd help with readability. I don't have strong opinions on this one though.

weerd:
> I'm curious what you think about using util instead of data -- both are pretty vague, we use util over in very similar looking Gambit helpers.

@aaronschachter 🤔  I get the impression `util` is generic enough where any of these helpers, even ones in other files would fit in there. The idea with `data` was moreso for data manipulation helpers, but maybe that doesn't come across entirely.

I still lean towards `data`, but I can change it if you feel very strong about it or if others would prefer it. 😄

resources/assets/helpers/data.js
weerd: Oh good to know! I think that's some of my PHP doc knowledge leaking into JS 😬  I'll update it to `*`!

PR #2501, opened 2021-01-11
Profile and Badges Cypress Tests

lindsayrmaher:
### What's this PR do?

This pull request adds long awaited front-end testing for the profile and badges tabs in a user account! Most of these are pretty basic, but I tried to make sure were testing things that were dependent on specific queries.

👀 

Anything you think we could add in? 

I will add in more testing for the rewards tab with the table and progress bar (not yet added to the front end) once we're ready to connect to the backend with real data and have a query we can work with.

### Relevant tickets

References [Pivotal #176438356](https://www.pivotaltracker.com/n/projects/2401401/stories/176438356).

### Checklist

### DIFF
--- a/cypress/fixtures/graphql.js
+++ b/cypress/fixtures/graphql.js
@@ -76,6 +76,7 @@ export const mocks = {
   // Custom scalars:
   URL: () => 'https://www.example.com/',
   DateTime: () => new Date().toISOString(),
+  Date: () => new Date().toISOString(),
   JSON: () => ({}),
 
   // If the field is a string & not mocked, default to showing the
--- a/cypress/integration/account-profile-tab.js
+++ b/cypress/integration/account-profile-tab.js
@@ -7,15 +7,17 @@ describe('User Account Profile Tab', () => {
     const user = userFactory();
     cy.mockGraphqlOp('AccountQuery', {
       user: {
+        firstName: user.firstName,
         lastName: 'Tester',
-        birthdate: Date(),
         email: 'tester@mail.com',
-        emailSubscriptionTopics: ['COMMUNITY'],
+        birthdate: '10/12/1994',
       },
     });
     cy.login(user);
     cy.visit(`/us/account`);
     cy.findByTestId('user-email').should('contain', 'tester@mail.com');
+    cy.findByTestId('user-name').should('contain', `${user.firstName} Tester`);
+    cy.findByTestId('user-birthdate').should('contain', '10/12/1994');
   });
 
   /** @test */
@@ -27,12 +29,6 @@ describe('User Account Profile Tab', () => {
   /** @test */
   it('Shows the user as unregistered based on their voter registration status', () => {
     const user = userFactory();
-    cy.mockGraphqlOp('AccountQuery', {
-      user: {
-        birthdate: Date(),
-        email: 'tester@mail.com',
-      },
-    });
 
     cy.mockGraphqlOp('UserVoterRegistrationStatusQuery', {
       user: {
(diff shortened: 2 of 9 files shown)


cypress/integration/account-profile-tab.js
```diff
@@ -0,0 +1,92 @@
...
+  beforeEach(() => cy.configureMocks());
+  /** @test */
+  it('Renders user Profile Tab if that user is logged in', () => {
+    const user = userFactory();
+    cy.mockGraphqlOp('AccountQuery', {
+      user: {
+        lastName: 'Tester',
+        birthdate: Date(),
+        email: 'tester@mail.com',
+        emailSubscriptionTopics: ['COMMUNITY'],
```
mendelB, line 13: Nit but do we need to custom stub out the other properties here besides for the `email` which we're asserting against? On the other hand, it could be nice to check that the name, birthdate show up as well!

cypress/integration/account-rewards-tab.js
```diff
@@ -0,0 +1,73 @@
...
+        birthdate: Date(),
+        email: 'tester@mail.com',
+      },
+    });
+
+    cy.login(user);
+    cy.visit(`/us/account/badges`);
+
+    cy.findByTestId('signup-badge').click();
+    cy.findByTestId('badges-modal').should('have.length', 1);
```
mendelB, line 37: One other testing idea is to ensure that the badge toggles the "unearned text" based on whether the user has earned the badge or not. (We can mock this [signup count query](https://github.com/DoSomething/phoenix-next/blob/03716ad136a3d00b5a31e36bf6bceba99e306118/resources/assets/components/pages/AccountPage/Badges/BadgesTab.js#L19-L23) to return signups or not and test against that.)

mendelB (approved):
🧪

cypress/integration/account-profile-tab.js
lindsayrmaher: So it's a bit convoluted, but when I first was building this out the `AccountQuery` was throwing an error about a Date being undefined, and when I added the birthday it resolved. So then I added values for all the things the `AccountQuery` pulls lol I think we could pare (par?) it down a little.

cypress/integration/account-rewards-tab.js
lindsayrmaher: I was wondering how to test that, I'll add it in!

cypress/integration/account-profile-tab.js
mendelB: Iiinteresting 🤔 

I wonder if this might be because we don't mock the `Date` Scalar which is used for the [birthdate field](https://github.com/DoSomething/graphql/blob/80c0bce1a5f4bdcef8f06313d0481c92f87a311c/src/schema/northstar.js#L110) in GraphQL. We only mock the `DateTime` over in our [GraphQL fixture](https://github.com/DoSomething/phoenix-next/blob/03716ad136a3d00b5a31e36bf6bceba99e306118/cypress/fixtures/graphql.js#L78).

lindsayrmaher: > I wonder if this might be because we don't mock the Date Scalar which is used for the birthdate field in GraphQL. We only mock the DateTime over in our GraphQL fixture.

This was 100% the issue! I added it to the fixture and it's running just fine now without that query added.

cypress/fixtures/graphql.js
```diff
@@ -76,6 +76,7 @@ export const mocks = {
   // Custom scalars:
   URL: () => 'https://www.example.com/',
   DateTime: () => new Date().toISOString(),
+  Date: () => new Date().toISOString(),
```
lindsayrmaher, line 79: @mendelB does this make sense? I figured it would be similar but not sure 🤔

mendelB: I think this is fine! If we want to match the [expected format](https://github.com/DoSomething/graphql/blob/80c0bce1a5f4bdcef8f06313d0481c92f87a311c/src/schema/northstar.js#L109) (`YYYY-MM-DD`) perfectly, we can just append a `slice` to grab only this part of the string I think.

```suggestion
  Date: () => new Date().toISOString().slice(0, 10),
```

`# '2021-01-12T16:42:47.230Z' -> '2021-01-12'`

PR #2506, opened 2021-01-14
Tooltip Component for Campaign Info Block

mendelB:
### What's this PR do?

This pull request adds a new `Tooltip` utility component and renders it in the `CampaignInfoBlock` for the Time Commitment and Volunteer Credit fields.

- Does the implementation of the library for the new Tooltip component make sense?
- Does the implementation of the Tooltip for the CampaignInfoBlock make sense? (The copy is kinda stubbed out for now per the designs, so I'll update to the final copy before merging)

I'll like to the review app once it builds, that'd be an easy place to see this in action.

I'm going to follow up with documentation/tests.

So this one turned out to be a bit of a doozy.

I started by trying to implement a tooltip with our own markup/CSS following along with the [w3schools tutorial](https://www.w3schools.com/css/css_tooltip.asp).

But custom code presented a few issues:
- Granted our desired scenario ([designs](https://app.abstract.com/share/3e468b7b-8e3a-4a7c-a708-5d4c7040efea?collectionId=c5a15e60-b0e3-4bea-aed1-10667303a908&collectionLayerId=55126c71-99dd-44fd-ae1f-b4bb17185bd1&mode=build&sha=194a57e6723f8acbe559a95168f62a6ed7ede6cb)) is somewhat straightforward, but having this work in multiple directions (not just as a "top" tooltip) would require more involved styling.
- This actually became relevant, since on very small screens, the values in the info block end up [floating to the _left_](https://user-images.githubusercontent.com/12417657/104645398-ffa6ff80-567c-11eb-8537-1c397dc7e4e8.png) of the screen which would require different arrow placement than in the designs.
- Most pressingly, on mobile screens, the tooltip would break out of the viewport boundary (since it's `absolute`ly placed). The logic to keep it inside is just more work.
- For a robust re-usable tooltip - this really didn't feel like an in-house approach would be fitting -- why re-invent the wheel.

I audited a few different packages:
- [Popper](https://popper.js.org/docs/v2/) seemed to be pretty popular (sorry), but was pretty complicated to implement, and seemed a tad buggy. That could have just been my local, but the headache was large.
- [React Laag](https://www.react-laag.com/) was pretty cool but was also a bit difficult to implement and couldn't seem to handle responsive-ness perfectly.
- [React Tooltip Lite](https://github.com/bsidelinger912/react-tooltip-lite) seemed a bit too lite, and I couldn't get it to work perfectly either.

Finally, I settled on [React Tiny Popover](https://github.com/alexkatz/react-tiny-popover) a pretty lightweight solution focused on _placement_ which is our core problem for this tooltip. Most importantly it was pretty intuitive to implement on a basic level and just...worked, straight off the bat.

I added a `utilities/Tooltip` component so we can use this wherever / however. We can further abstract away, customize, etc. as use-cases arise, but I think we're in a pretty good place. 

I opted to keep the trigger content implementation (in this case a question-mark SVG) _outside_ the actual tooltip, for now, keeping it more agnostic about its children.

Finally, the gnarly part of the implementation was figuring out how to trigger/hide the tooltip across desktop devices while allowing hover ability over the tooltip itself as well, mobile devices that need touchscreen interactivity, and accessibility for focus tabbing, and ENTER triggering. The most interesting bit is a 1-second timer to hide the tooltip on larger screens, allowing the user the opportunity to hover over the actual tooltip which would cancel the hide event.

### Relevant tickets

References [Pivotal #176371444](https://www.pivotaltracker.com/story/show/176371444).

### Checklist

📸 
- [Large](https://user-images.githubusercontent.com/12417657/104656716-c0cd7580-568d-11eb-97ef-c4b0db43f215.png)
- [Medium](https://user-images.githubusercontent.com/12417657/104656729-c7f48380-568d-11eb-9894-0f688b7ed3d0.png)
- [Small](https://user-images.githubusercontent.com/12417657/104656787-e0fd3480-568d-11eb-9552-b64d57b8a463.png)
- [XS](https://user-images.githubusercontent.com/12417657/104656817-ece8f680-568d-11eb-9271-b493d5ea6ed7.png)


mendelB:
UPDATE: https://github.com/DoSomething/phoenix-next/pull/2506/commits/9450b3cc321c42f385c3fbd35a440f9223696cb5 adds logic for the tooltip to disappear when the user scrolls on mobile screens per [QA feedback](https://www.pivotaltracker.com/story/show/176371444/comments/221192082).

resources/assets/components/utilities/Tooltip.js
```diff
@@ -0,0 +1,75 @@
...
+        tabIndex={0}
+        onMouseEnter={() => setIsOpen(true)}
+        onMouseLeave={() =>
+          !isSmallScreen
+            ? // Slight delay before closing tooltip to enable hovering over the tooltip itself which will keep it open.
+              setCloseTimeout(setTimeout(() => setIsOpen(false), 1000))
+            : {}
+        }
+        // Enables using the ENTER key to open the tooltip for accessibility.
+        onKeyDown={event => (event.keyCode === 13 ? setIsOpen(true) : {})}
```
lindsayrmaher, line 59: love!

lindsayrmaher (commented):
This seems to be working as expected for me! I think the library makes sense to use, and I like that it addresses accessibility options as well. I did run into [an error](https://reactjs.org/docs/error-decoder.html/?invariant=300) when using the review app, I'm not sure if this is getting triggered by these changes or some of the other features in Teans for Jeans

mendelB:
> I did run into [an error](https://reactjs.org/docs/error-decoder.html/?invariant=300) when using the review app, I'm not sure if this is getting triggered by these changes or some of the other features in Teans for Jeans

Iiinteresting! I was able to reporoduce this error by toggling between mobile and desktop views. I'll look into this -- thanks for flagging! In the meantime, you (hopefully!) should be able to test this without errors by sticking to one of those views throughout the session.

mendelB:
> > I did run into [an error](https://reactjs.org/docs/error-decoder.html/?invariant=300) when using the review app, I'm not sure if this is getting triggered by these changes or some of the other features in Teans for Jeans
> 
> Iiinteresting! I was able to reproduce this error by toggling between mobile and desktop views. I'll look into this -- thanks for flagging! In the meantime, you (hopefully!) should be able to test this without errors by sticking to one of those views throughout the session.

Ahhh this had to do with the way we were conditionally rendering the `useEffect` hook. Fixed in https://github.com/DoSomething/phoenix-next/pull/2506/commits/3ac30dd0473a8ecedf81fa07f40fd587476659d1!

lindsayrmaher (approved):
looking good!

PR #2502, opened 2021-01-12
Campaigns Tab cypress test

lindsayrmaher:
### What's this PR do?

This pull request adds tests to the Campaigns tab of a user account.

### How should this be reviewed?

👀 

### Any background context you want to provide?

Again doing some backfill from before we had auth setup via the front-end

### Relevant tickets

References [Pivotal #176438356](https://www.pivotaltracker.com/story/show/176438356).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Added appropriate feature/unit tests.

PR #2507, opened 2021-01-14
Subscription center cypress tests

lindsayrmaher:
### What's this PR do?

This pull request adds a test to make sure our subscription center displays the correct status for a users choices. I think we should be able to close #1657 since it is essentially an older version of this! (and also maybe #1733 which was covered in #2502 - although i like the idea of testing multiple posts so maybe we should port that over).

👀 

Ongoing testing updates

### Relevant tickets

References [Pivotal #176438356](https://www.pivotaltracker.com/story/show/176438356).

### Checklist

### DIFF
--- a/cypress/integration/account-subscriptions-tab.js
+++ b/cypress/integration/account-subscriptions-tab.js
@@ -12,20 +12,32 @@ describe('User Account Email Subscriptions Tab', () => {
 
     cy.mockGraphqlOp('AccountQuery', {
       // Get back a particular user and subscription topics
-      user: root => ({
+      user: {
         firstName: 'Delilah',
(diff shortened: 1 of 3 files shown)


cypress/integration/account-subscriptions-tab.js
```diff
@@ -0,0 +1,43 @@
...
+  beforeEach(() => cy.configureMocks());
+
+  it('View existing subscriptions', () => {
+    // This user is just for auth purposes
+    const user = userFactory();
+
+    cy.mockGraphqlOp('AccountQuery', {
+      // Get back a particular user and subscription topics
+      user: {
+        firstName: 'Delilah',
```
DFurnes, line 16: Deliiiilahhhhh! 🎵

DFurnes (approved):

PR #2512, opened 2021-01-19
add benefits to site navigation

Dinnall:
### What's this PR do?

This pull request removes the `Scholarship` tab and adds the `Benefits` to the site navigation

I also bundled the relevant analytics events for the sub-navigation of the benefits tab 

🔎 
...

...

### Relevant tickets

References [Pivotal #176401113](https://www.pivotaltracker.com/n/projects/2401401/stories/176401113).

### Checklist

- [  x ] This PR has been added to the relevant Pivotal card.
- [  x ] Added screenshots of front-end changes on small, medium, and large screens.
- 
New Edition: 
Screen Shot 2021-01-19 at 4 23 22 PM

Mobile:
Screen Shot 2021-01-19 at 4 53 48 PM

### DIFF
--- a/resources/assets/components/SiteNavigation/SiteNavigation.js
+++ b/resources/assets/components/SiteNavigation/SiteNavigation.js
@@ -534,7 +534,7 @@ class SiteNavigation extends React.Component {
                             name: 'clicked_nav_link_benefits',
                             action: 'link_clicked',
                             category: EVENT_CATEGORIES.navigation,
-                            label: 'benefits_scholarships',
+                            label: 'benefits',
                           })
                         }
                       >
--- a/resources/assets/components/SiteNavigation/SiteNavigation.js
+++ b/resources/assets/components/SiteNavigation/SiteNavigation.js
@@ -585,7 +585,6 @@ class SiteNavigation extends React.Component {
                           </a>
                           <a
                             href="/us/about/volunteer-hours"
-                            className="block"
                             onClick={() => {
                               this.handleOnClickLink({
                                 name: 'clicked_subnav_link_benefits_volunteer',
--- a/resources/assets/components/SiteNavigation/SiteNavigation.js
+++ b/resources/assets/components/SiteNavigation/SiteNavigation.js
@@ -511,20 +511,133 @@ class SiteNavigation extends React.Component {
               ) : null}
             </li>
 
-            <li className="menu-nav__item">
-              <a
-                href="/us/about/easy-scholarships"
-                onClick={() =>
-                  this.handleOnClickLink({
-                    name: 'clicked_nav_link_scholarships',
-                    action: 'link_clicked',
-                    category: EVENT_CATEGORIES.navigation,
-                    label: 'scholarships',
-                  })
-                }
+            <li
+              className={classnames('menu-nav__item', {
+                'is-active': this.state.activeSubNav === 'BenefitsSubNav',
+              })}
+              onMouseEnter={() => this.handleMouseEnter('BenefitsSubNav')}
+              onMouseLeave={() => this.handleMouseLeave('BenefitsSubNav')}
+            >
+              <Media
+                queries={{
+                  large: '(min-width: 960px)',
+                }}
               >
-                Scholarships
-              </a>
+                {matches => (
+                  <>
+                    {matches.large ? (
+                      <a
+                        id="main-nav__benefits"
+                        href="/us/about/benefits"
+                        onClick={() =>
+                          this.handleOnClickLink({
+                            name: 'clicked_nav_link_benefits',
+                            action: 'link_clicked',
+                            category: EVENT_CATEGORIES.navigation,
+                            label: 'benefits',
+                          })
+                        }
+                      >
+                        Benefits
+                        <span className="main-nav__arrow" />
+                      </a>
+                    ) : (
+                      <a
+                        id="main-nav__benefits"
+                        href="/"
+                        onClick={e =>
+                          this.handleOnClickToggle(e, 'BenefitsSubNav', {
+                            name: 'clicked_nav_link_benefits',
+                            action: 'link_clicked',
+                            category: EVENT_CATEGORIES.navigation,
+                            label: 'benefits',
+                          })
+                        }
+                      >
+                        Benefits
+                        <span className="main-nav__arrow" />
+                      </a>
+                    )}
+                  </>
+                )}
+              </Media>
+
+              {this.state.activeSubNav === 'BenefitsSubNav' ? (
+                <div className="main-subnav menu-subnav">
+                  <div className="wrapper base-12-grid py-3 md:py-6">
+                    <section className="main-subnav__links-benefits menu-subnav__links menu-subnav__section">
+                      <h1>Benefits</h1>
+                      <ul>
+                        <li>
+                          <a
+                            href="/us/about/easy-scholarships"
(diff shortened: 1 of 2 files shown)


katiecrane:
I guess this really was a 2!! The only off thing that I'm noticing is where the text column is, it's all the way to the left and I think it should be positioned the same way the text is in the cause dropdown:

Jan-19-2021 15-59-01

Dinnall:
> I guess this really was a 2!! The only off thing that I'm noticing is where the text column is, it's all the way to the left and I think it should be positioned the same way the text is in the cause dropdown:
> 
> Jan-19-2021 15-59-01

lol 👀 didn't think anyone would review this today.

Regarding the position of the subtext, I didn't change it because that's how it appears in the abstract file 
@katiecrane

katiecrane:
haha @Dinnall it's because I'm on mountain time! to me it looks in the designs like the text is bumped in on the benefits menu too, unless maybe we're looking at different designs? 
image

weerd:
Just FYI, I'll aim to give this an 👁️ before my EOD (8pm est!)

weerd:
> Regarding the position of the subtext, I didn't change it because that's how it appears in the abstract file

I think the Abstract file is wrong then. You can see that the featured item lines up with a specified internal padding from the right, and in the cause menu it's the same amount on the left, pushing the links in. That's all intended and meant to align to the grid 😄 

Site Navigation

resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -511,20 +511,132 @@ class SiteNavigation extends React.Component {
...
+                      </a>
+                    )}
+                  </>
+                )}
+              </Media>
+
+              {this.state.activeSubNav === 'BenefitsSubNav' ? (
+                <div className="main-subnav menu-subnav">
+                  <div className="wrapper base-12-grid py-3 md:py-6">
+                    <section className="main-subnav__links-benefits menu-subnav__links menu-subnav__section">
```
weerd, line 568: To make the menu align to the grid properly, you'll need to add an item to the SCSS file. This component wasn't built with Tailwind and uses an attached style file :) Here's the relevant section you can copy from how it's done for the Causes subnav links:

https://github.com/DoSomething/phoenix-next/blob/65d216212f526c5528d4c07f0fb4c20eb38625b0/resources/assets/components/SiteNavigation/site-navigation.scss#L209-L215

resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -511,20 +511,132 @@ class SiteNavigation extends React.Component {
...
+                        <li>
+                          <a
+                            href="/us/about/easy-scholarships"
+                            onClick={() => {
+                              this.handleOnClickLink({
+                                name:
+                                  'clicked_subnav_link_benefits_scholarships',
+                                action: 'link_clicked',
+                                category: EVENT_CATEGORIES.navigation,
+                                label: 'benefits',
```
weerd, line 580: Maybe this was decided elsewhere, but if mimicking how it's done in the Causes submenu wouldn't the `label` be `benefits_scholarships`? Not 100% sure as this might have been discussed already!

resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -511,20 +511,132 @@ class SiteNavigation extends React.Component {
...
+                            Scholarships
+                          </a>
+                          <a
+                            href="/us/about/volunteer-hours"
+                            onClick={() => {
+                              this.handleOnClickLink({
+                                name: 'clicked_subnav_link_benefits_volunteer',
+                                action: 'link_clicked',
+                                category: EVENT_CATEGORIES.navigation,
+                                label: 'benefits',
```
weerd, line 593: Same question as above with this `label`.

resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -511,20 +511,132 @@ class SiteNavigation extends React.Component {
...
+                                  'clicked_subnav_link_benefits_scholarships',
+                                action: 'link_clicked',
+                                category: EVENT_CATEGORIES.navigation,
+                                label: 'benefits',
+                              });
+                            }}
+                          >
+                            Scholarships
+                          </a>
+                          <a
```
weerd, line 586: These `<a>`'s shouldn't be siblings; they should each be in their own `<li>`s, so:

```suggestion
                              </a>
                          </li>
                          <li>
                              <a
```

resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -511,20 +511,132 @@ class SiteNavigation extends React.Component {
...
+                            Volunteer Credits
+                          </a>
+                        </li>
+                      </ul>
+                    </section>
+
+                    <section className="main-subnav__featured menu-subnav__content menu-subnav__section">
+                      <SiteNavigationFeature
+                        imageSrc="https://images.ctfassets.net/81iqaqpfd8fy/5atz8qkRKqMd6RnNbu8TyS/e2856278db6218801d07ecc6326b3b81/strenth_though_service_header_photo_Artboard_1_copy_5.png"
+                        imageAlt="volunteer hours"
```
weerd, line 606: This might be for someone else to provide, but we should strive to have our `alt` attributes better describe the image... something like:

`Image with text in center that reads "Strength Through Service" and is surround by illustration of hands with hearts in their palms, reaching towards the text.`

I know this is more work, but we should do a better job in general with our accessibility!

resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -511,20 +511,132 @@ class SiteNavigation extends React.Component {
...
+                        title="Volunteer Credits"
+                        text="Earn a certificate verifying your volunteer hours"
+                        moreLinkText="Get Started"
+                        callback={() =>
+                          this.analyzeEvent({
+                            name:
+                              'clicked_subnav_link_feature_benefits_volunteer_hours',
+                            action: 'link_clicked',
+                            category: EVENT_CATEGORIES.navigation,
+                            label: 'benefits',
```
weerd, line 617: Same as questions above regarding the `label`.

resources/assets/components/SiteNavigation/SiteNavigation.js
ngjo: ahhh good catch @weerd this was inadvertently bad copy/pasting on my part from the spreadsheet to the [ticket](https://www.pivotaltracker.com/story/show/176421574). @Dinnall i've updated the ticket now! 

one difference to call out that we're ok with is that the featured slot for the benefits drop down has a slightly different analytic to help us easily know it's the benefits drop-down (i.e., `clicked_subnav_link_feature_benefits_volunteer_hours` vs the causes drop-down has `clicked_subnav_link_feature_racial_justice_collection` -- this is intentional). thanks!

weerd:
@Dinnall do you have [EditorConfig](https://editorconfig.org/) and Prettier set up on your computer? It's hard to see the changes in the SCSS file cause it modified a bunch of spacing etc when it shouldn't have 🤔

weerd:
This looks correct to me in the deployed review app! Just need to fix those tests and should be good to go! 😄

weerd (approved):
🎉

PR #2526, opened 2021-01-28
Update SiteNavigation.js

Dinnall:
### What's this PR do?

This pull request adds the tracking context field to our benefits tab analytics

...

The initial analytics was added in the [benefits dropdown](https://github.com/DoSomething/phoenix-next/pull/2512) PR
this ticket adds the last field 
...

### Relevant tickets

References [Pivotal #176421574](https://www.pivotaltracker.com/n/projects/2401401/stories/176421574).

### Checklist

- [ X  ] This PR has been added to the relevant Pivotal card.

### DIFF
--- a/resources/assets/components/SiteNavigation/SiteNavigation.js
+++ b/resources/assets/components/SiteNavigation/SiteNavigation.js
@@ -534,6 +534,10 @@ class SiteNavigation extends React.Component {
                             action: 'link_clicked',
                             category: EVENT_CATEGORIES.navigation,
                             label: 'benefits',
+                            context: {
+                              ...this.getContextData(),
+                              ...this.trackingData,
+                            },
                           })
                         }
                       >
@@ -550,6 +554,10 @@ class SiteNavigation extends React.Component {
                             action: 'link_clicked',
                             category: EVENT_CATEGORIES.navigation,
                             label: 'benefits_scholarships',
+                            context: {
+                              ...this.getContextData(),
+                              ...this.trackingData,
+                            },
                           })
                         }
                       >
@@ -577,6 +585,10 @@ class SiteNavigation extends React.Component {
                                 action: 'link_clicked',
                                 category: EVENT_CATEGORIES.navigation,
                                 label: 'benefits_scholarships',
+                                context: {
+                                  ...this.getContextData(),
+                                  ...this.trackingData,
+                                },
                               });
                             }}
                           >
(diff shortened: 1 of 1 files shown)


resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -534,6 +534,10 @@ class SiteNavigation extends React.Component {
                             action: 'link_clicked',
                             category: EVENT_CATEGORIES.navigation,
                             label: 'benefits',
+                            context: {
+                              ...this.getContextData(),
+                              ...this.trackingData,
+                            },
```
weerd, line 540: Not sure exactly what's going on in these changes.... 

The `SiteNavigation` class doesn't have a `getContextData()` method or a `trackingData` property. We do already add a bunch of context info when we call the `analyzeEvent()` method (via `this.handleOnClick()`):

https://github.com/DoSomething/phoenix-next/pull/2526/files#diff-60d09dddf46e11848350bf31ba3988233c48aea8c617787297dfd83fb2ae162eR48-R53

Mind clarifying? Maybe I'm just missing something 😬

Dinnall: I don't mind clarifying. Withing the ticket `campaignId, pageId, utms` were included.
I added them because I didn't see any harm in including them although they are optional 🤔 .

weerd: Ah gotcha, ok.

All that stuff already gets added in! The `analyticsEvent()` method on the `SiteNavigation` class calls `getPageContext()` and `getUtmContext()` and appends it all to the `context` property :)

https://github.com/DoSomething/phoenix-next/blob/91029490fbf52565efffb18020e6999c853c8ffa/resources/assets/helpers/analytics.js#L163-L182

But also, I think the way you added it would only cause errors on the page and not actually add anything to the event since you're calling those methods via `this.*` and those don't exist on the class.

The `this.getContextData()` and `this.trackingData` only exist internally to the `ShareAction.js` component:

https://github.com/DoSomething/phoenix-next/blob/91029490fbf52565efffb18020e6999c853c8ffa/resources/assets/components/actions/ShareAction/ShareAction.js#L47

https://github.com/DoSomething/phoenix-next/blob/91029490fbf52565efffb18020e6999c853c8ffa/resources/assets/components/actions/ShareAction/ShareAction.js#L101

I don't think we have tests for this tracking stuff, so I think that's why all the checks passed and didn't otherwise pick up on this! 😓

weerd:
You can just close this PR since the default tracking should already include all that info! Magic ✨  🕺

Dinnall:
> You can just close this PR since the default tracking should already include all that info! Magic ✨ 🕺

Sounds good to me 👍🏾

PR #2528, opened 2021-02-01
Matching Gallery Block Header to HomePage

lindsayrmaher:
### What's this PR do?

This pull request updates the header of the `GalleryBlock` component to match the homepage!

I tried to keep things DRY, but i didn't want to entirely eliminate the current `SectionHeader` or the styling of the Homepage because there were specific differences between them all. (ie, homepage uses AnalyticsWayPoints and the `SectionHeader` is used in other places). That being said, are there other ways we could make this more reusable?

Request from the design team!
Current:
Screen Shot 2021-02-01 at 11 24 05 AM

Large Screen:
Screen Shot 2021-02-01 at 11 18 13 AM

Small Screen:
Screen Shot 2021-02-01 at 11 18 42 AM

### Relevant tickets

References [Pivotal #174220566](https://www.pivotaltracker.com/story/show/174220566).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/SectionHeader/GalleryBlockHeader.js
+++ b/resources/assets/components/utilities/SectionHeader/GalleryBlockHeader.js
@@ -13,28 +13,29 @@ export const centerHorizontalRule = css`
   }
 `;
 
-const GalleryBlockHeader = ({ title }) => (
+const GalleryBlockHeader = ({ bgColor, title }) => (
   <div className="grid-wide text-center">
-    {title ? (
-      <h2 className="mb-6 relative">
-        <span className="bg-gray-100 font-league-gothic font-normal leading-tight inline-block px-6 relative text-3xl md:text-4xl uppercase z-10">
-          {title}
-        </span>
-        <span
-          className="absolute bg-purple-500 block h-1 w-full z-0"
-          css={centerHorizontalRule}
-        />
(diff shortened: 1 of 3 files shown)


resources/assets/components/utilities/SectionHeader/GalleryBlockHeader.js
```diff
@@ -0,0 +1,40 @@
...
+export const centerHorizontalRule = css`
+  @media (min-width: ${tailwindScreens.md}) {
+    margin-top: -2px;
+    top: 50%;
+  }
+`;
+
+const GalleryBlockHeader = ({ title }) => (
+  <div className="grid-wide text-center">
+    {title ? (
```
mendelB, line 18: I wonder if there's any sense in outputting anything if there's no `title` prop provided 🤔 since otherwise we end up with a phantom `<div>` on the page. Maybe worth wrapping even the parent `<div>` in this ternary and making the prop required?

mendelB (approved):
💯 

> homepage uses AnalyticsWayPoints

If we pushed the waypoint just above the <div> in the homepage (and perhaps used two separate `div.grid-wide...`s for that gallery if the `GalleryBlockHeader`s one interferes) could we actually use the `GalleryBlockHeader` in the Homepage as well?

PR #2529, opened 2021-02-01
Fix Campaign Link for RAF Beta Page

mendelB:
Quick fix per #2525 where I mistakenly used the `url` property (which we aren't even querying via GraphQL) instead of the `path`.

https://www.pivotaltracker.com/story/show/176641967/comments/221606907


katiecrane (approved):
good catch!

PR #2533, opened 2021-02-01
update homepage a tags

Dinnall:
### What's this PR do?

This pull request adds an update to the anchor tags on the homepage.
Refer to Leah's last comment

...

I also added a small tweak to one of the benefits tracking labels.
Previously, the label appeared as `benefits_scholarships`, now it states `benefits_volunteer`
References [Pivotal #176421574](https://www.pivotaltracker.com/n/projects/2401401/stories/176421574).

...

### Relevant tickets

References [Pivotal #174220514](https://www.pivotaltracker.com/n/projects/2401401/stories/174220514).

### Checklist

### DIFF
--- a/resources/assets/components/pages/HomePage/HomePage.js
+++ b/resources/assets/components/pages/HomePage/HomePage.js
@@ -255,7 +255,7 @@ const HomePageTemplate = ({ articles, campaigns, coverImage, title }) => {
                   Choose a campaign below to make an impact,{' '}
                   <a
                     href="/us/about/easy-scholarships"
-                    className="font-normal text-blurple-500 underline text-teal-500 hover:text-blurple-400"
+                    className="font-normal underline text-teal-500 hover:text-blurple-400"
                     data-label="campaign_section_earn_scholarships"
                   >
                     win scholarships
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/HomePage/HomePage.js
```diff
@@ -255,15 +255,15 @@ const HomePageTemplate = ({ articles, campaigns, coverImage, title }) => {
...
+                    className="font-normal text-blurple-500 underline text-teal-500 hover:text-blurple-400"
                     data-label="campaign_section_earn_scholarships"
                   >
                     win scholarships
                   </a>
                   , and{' '}
                   <a
                     href="/us/about/volunteer-hours"
-                    className="font-normal text-blurple-500 hover:text-blurple-300 underline hover:text-teal-500"
+                    className="font-normal text-blurple-500 underline text-teal-500 hover:text-blurple-400"
```
weerd, line 266: I don't think we need the first `text-plurple-500` if right after we're setting the text to `text-teal-500`. The text is getting colored as `blurple-500` initially, then `teal-500` after since it comes next in the list; so the first class is just redundant.

Dinnall: updated

weerd (approved):
👍

PR #2539, opened 2021-02-04
Campaign Signup Form - GraphQL instead of Redux

mendelB:
### What's this PR do?

This pull request updates the `CampaignSignupForm` component to use GraphQL to signup the user instead of via Redux actions. 

- The core of the refactoring happens in https://github.com/DoSomething/phoenix-next/commit/19af1bd4690a367dcdcdd80fdf54edaa8ec80b64
- https://github.com/DoSomething/phoenix-next/commit/1d5a3d5d161ee1a1dc3e47f32532b597b89fb84f moves the affiliate opt in logic into the component

The rest of the work is mostly deleting unused and outdated Redux reducers/fields.

UPDATE: https://github.com/DoSomething/phoenix-next/pull/2539/commits/55f803664f693e5197e52d6dd308a34a8c569062 persists flashed session data via the `useGate` hook throughout a Cypress test login flow, and https://github.com/DoSomething/phoenix-next/pull/2539/commits/372e2cb1e2cecdebd59ee3bf9132eb973fb75292 copies over the tests from `campaign-signup` with the new feature flag toggled on and edited for the GraphQL flow.

I'd recommend commit-by-commit on this one!

This is part of the effort to migrate our signup functionality out of Redux and simplify it with GraphQL. The full shebang will require more work, but this is a solid start I think!

I've feature flagged this for now since this is _important_ functionality, so once this is live and thriving for a hot minute we can retire the old code.

~Tests have been a hot mess to try and add so far (main issue was trying to stash and restore our `useGate` flash storage). I've got something working for a separate PR. But we've got thorough 👻 tests we can also rely on to ensure the signup flow is working on site.~

### Relevant tickets

References [Pivotal #175291206](https://www.pivotaltracker.com/story/show/175291206).

### Checklist


schema.json
```diff
@@ -10574,7 +10574,7 @@
             "args": [
               {
                 "name": "campaignId",
-                "description": null,
+                "description": "The campaign ID for this signup.",
```
Dinnall, line 10577: Love the elaborating on the description here

resources/assets/components/CampaignSignupForm/CampaignSignupForm.test.js
```diff
@@ -1,50 +1,83 @@
...
 
 const props = {
   campaignId: '123',
   pageId: '456',
   storeCampaignSignup,
+  signupCreated,
 };
 
+// Mock the GraphQL query and response.
+const mocks = [
```
Dinnall, line 23: 👏🏾

resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
```diff
@@ -106,55 +287,53 @@ const CampaignSignupForm = props => {
...
-            className={`${className} py-2 md:py-3`}
-            isDisabled={!groupId}
-            onClick={handleSignup}
-            text="Join Group"
-          />
+            <PrimaryButton
+              attributes={{ 'data-testid': 'join-group-signup-button' }}
+              className={`${className} py-2 md:py-3`}
+              isDisabled={!groupId}
+              isLoading={loading || mutationLoading}
```
Dinnall, line 305: Curious, why do we need both `loading` and `mutationLoading` here?

mendelB: @mendelB good 👁️! I added this one in before I had the [`return` condition](https://github.com/DoSomething/phoenix-next/pull/2539/files#diff-cb3f69b2c5ddc99c8e4fcab6b9b92d29a3847410a597cc57af321e5aa9c6a8aaR162) set up. Removed in https://github.com/DoSomething/phoenix-next/pull/2539/commits/1f9705ddb02084ba7487b329e500d12ccc9ec830

config/feature-flags.php
```diff
@@ -22,4 +22,5 @@
     'refer_friends_incentive' => env('DS_ENABLE_REFER_FRIENDS_INCENTIVE', false),
     'sitewide_nps_survey' => env('DS_ENABLE_SITEWIDE_NPS_SURVEY', false),
     'rewards_levels' => env('DS_ENABLE_REWARDS_LEVELS', false),
+    'graphql_campaign_signup' => env('DS_ENABLE_GRAPHQL_CAMPAIGN_SIGNUP', false),
```
DFurnes, line 25: Smart to keep this under a feature flag! 👍

cypress/support/commands.js
```diff
@@ -87,13 +87,25 @@ Cypress.Commands.add('handleLogin', user => {
...
       const { actionId } = qs.parse(search);
 
       cy.login(user);
 
+      // Restore any flashed pre-authentication state.
+      cy.on('window:before:load', window => {
+        window.sessionStorage.setItem(
+          flashIdentifier,
+          JSON.stringify(flashState),
+        );
```
DFurnes, line 106: Would this work if we still stored this in `sessionStorage` in our application logic (rather than storing on `data-` and populating here)? I think that should be maintained between page refreshes, so the `sessionStorage.setItem` in the `authenticate` method should handle this.

resources/assets/helpers/auth.js
```diff
@@ -153,7 +153,20 @@ export const useGate = identifier => {
...
   const authenticate = newState => {
     window.sessionStorage.setItem(identifier, JSON.stringify(newState));
-    window.location = buildAuthRedirectUrl();
+
+    const redirect = buildAuthRedirectUrl();
+
+    // If we're running our test suite, don't automatically initiate
+    // the login redirect flow & leave something to assert on.
+    if (window.Cypress) {
+      document.body.innerHTML = `<div data-test="redirect" data-url="${redirect}" data-flash-identifier="${identifier}" />`;
```
DFurnes, line 162: I wonder if we should make a `redirect(url)` helper that includes this test logic & sets `window.location` so we don't need to remember about this implementation details whenever we write a feature that includes redirects?

resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
```diff
@@ -1,41 +1,173 @@
...
   const [groupId, setGroupId] = useState(null);
 
-  // Decorate click handler for A/B tests & analytics.
-  const handleSignup = () => {
-    const details = {};
+  // We'll want to redirect anonymous users to the auth flow, while stashing their selected data to session storage.
+  const [flash, authenticate] = useGate(
+    // Only pull the stashed data once we're done loading the users signups so that
+    // we don't overwrite it when the component refreshes!
+    loading ? '' : `CampaignSignupForm:${campaignId}`,
```
DFurnes, line 97: Tricky! I wonder if we could add a `skip` option for this hook (like how we can skip queries with `useQuery`):

```js
const [flash, authenticate] = useGate(`CampaignSignupForm:${campaignId}`, { skip: loading });
```

DFurnes (approved):
This is so exciting! 🐬

cypress/support/commands.js
mendelB: I think the issue stems from our manual session storage wipe between tests:

https://github.com/DoSomething/phoenix-next/blob/8757e704e7dd5afb5552489b05cfcc7e8fb010f0/cypress/support/index.js#L15-L16

which I think is triggered when we reload in the stubbed auth flow here

https://github.com/DoSomething/phoenix-next/blob/8757e704e7dd5afb5552489b05cfcc7e8fb010f0/cypress/support/commands.js#L99

🤔 Could we perhaps build something in here to skip the page refresh (a `withoutRedux` or something argument?) since we don't need to trigger any Redux actions via the `actionId` query param?

resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
mendelB: 🔥 🔥 ooooh I'll try and refactor this!

resources/assets/helpers/auth.js
mendelB: 👌 That's a great idea!

resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
mendelB: I gave this a go in https://github.com/DoSomething/phoenix-next/pull/2539/commits/c0881975c71f2ee04b2eec492bca6a7a79d73507

cypress/support/commands.js
mendelB: Hmm or better yet, [it looks like](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests.html#Support-file) we can define a `beforeEach` within the support files!

I can put the session storage clear command in a beforeEach instead of in the `before:load` hook which should get us golden!

resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
DFurnes: Amazing!!! 😎

cypress/support/commands.js
DFurnes: Oh that's much better!!! Yeah, that sounds like a bug that it was clearing session storage within tests. 👞

resources/assets/helpers/auth.js
mendelB: So much cleaner! https://github.com/DoSomething/phoenix-next/pull/2539/commits/827791f0c76a2d0afde8040b84bc0d4758f3aa99

PR #2541, opened 2021-02-08
Add button carat toggle

lindsayrmaher:
### What's this PR do?

This pull request adds back in our favorite carat toggle for the filters on the campaigns page! 

👀 

Kapture 2021-02-08 at 17 33 21

The way this is implemented should allow for this toggle to be added on different buttons should it be necessary in the future, but right now we only want to add it on each filter button on the campaigns index page.

### Relevant tickets

References [Pivotal #173713718](https://www.pivotaltracker.com/n/projects/2401401/stories/173713718).

### Checklist

### DIFF
--- a/resources/assets/components/utilities/Button/ElementButton.js
+++ b/resources/assets/components/utilities/Button/ElementButton.js
@@ -4,6 +4,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
+import MenuCarat from '../../artifacts/MenuCarat/MenuCarat';
+
 /**
  * Basic button element (<button>) with minimal styles.
  * Pass classes to style pseudo states like :hover, :focus and :active, etc.
@@ -12,7 +14,10 @@ import classnames from 'classnames';
  */
 const ElementButton = ({
   attributes,
+  caratColor,
+  caratToggle,
   className,
+  displayCaratToggle,
   isDisabled,
   isLoading,
   onClick,
@@ -28,13 +33,19 @@ const ElementButton = ({
       {...attributes}
     >
       {text}
+      {displayCaratToggle ? (
+        <MenuCarat color={caratColor} cssStyles={caratToggle} />
+      ) : null}
     </button>
   );
 };
 
 ElementButton.propTypes = {
   attributes: PropTypes.object,
+  caratColor: PropTypes.string,
+  caratToggle: PropTypes.string,
   className: PropTypes.string,
+  displayCaratToggle: PropTypes.bool,
   isDisabled: PropTypes.bool,
   isLoading: PropTypes.bool,
   onClick: PropTypes.func,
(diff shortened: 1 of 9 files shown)


lindsayrmaher:
waiting to merge my other PR and rebase so this effects less files

resources/assets/components/utilities/Button/SecondaryButton.js
```diff
@@ -53,7 +56,10 @@ const SecondaryButton = props => {
   return (
     <ElementButton
       attributes={attributes}
+      caratColor={caratColor}
+      caratToggle={caratToggle}
       className={classes}
+      displayCaratToggle={displayCaratToggle}
```
DFurnes, line 62: I would love to find a way to implement this that doesn't "bake in" the caret functionality to our more generalized `SecondaryButton` and `ElementButton` components. What if we passed the caret in as a `decoration` prop from `FilterNavigation.js`? Maybe something like this?

```jsx
<SecondaryButton
  isActive={activeFilter === name}
  text={startCase(name)}
  decoration={<MenuCarat color="#322baa" cssStyles={activeFilter === name ? caratToggle : null}  />}
  onClick={handleMenuToggle}
/>
```

That way all that `SecondaryButton` and `ElementButton` need to know is that there can optionally be some additional components passed in to sit next to the text & we can customize all the details of what that element is directly in the component where it's used!

DFurnes: This actually makes me wonder if it might be nice to provide the text for this component as children, so that we could place a decorative element before _or_ after the component too (say, if one day we wanted to place an icon in a button). But that's a larger refactoring that's probably not worth tackling right now!

lindsayrmaher: I like the idea of a decoration prop!

resources/assets/components/utilities/Button/ElementButton.js
```diff
@@ -21,20 +22,24 @@ const ElementButton = ({
 }) => {
   return (
     <button
-      className={classnames('btn', className, { 'is-loading': isLoading })}
+      className={classnames('btn flex', className, {
```
DFurnes, line 25: I'm assuming the `flex` here is to help space out the text & decoration, right? Are there any downsides where this makes the layout appear weird for other cases? (I'm not sure there will be, but I feel like I've seen some weirdness before with flex so want to make sure it seems ok!)

resources/assets/components/utilities/Button/ElementButton.js
```diff
@@ -21,20 +22,24 @@ const ElementButton = ({
...
+        'is-loading': isLoading,
+      })}
       disabled={isDisabled || isLoading}
       onClick={onClick}
       type={type}
       {...attributes}
     >
       {text}
+      {decoration || null}
```
DFurnes, line 34: Nitpick: you can omit the `|| null` since it defaults to it down below!

DFurnes (approved):
Looks great!

resources/assets/components/utilities/Button/ElementButton.js
lindsayrmaher: ah yes!

resources/assets/components/utilities/Button/ElementButton.js
lindsayrmaher: yeah i'm always hesitant to use flex anymore because of the weirdness. but it seems ok on all screen sizes for now!

PR #2542, opened 2021-02-10
Re-arrange mock query on petition post-submission cypress test to hopefully fix the flakiness

mendelB:
### What's this PR do?

This pull request attempts to address the flakey errors we keep seeing from the Petition post-submission redirect Cypress test by re-arranging where we setup the mock GraphQL query to align with the other tests.

👀 

#2535 confirmed that the issue here is the post's `user.id` which is stubbed to match the authed user is _not_ yielding the stubbed result.
image


lindsayrmaher (approved):
fix the flakes!

PR #2543, opened 2021-02-11
wraps variables in withoutNulls to avoid graphql error

lindsayrmaher:
### What's this PR do?

This pull request wraps the variables passed to the search query for the campaigns page in our `withoutNulls` helper to avoid errors thrown by algolia.

👀 

@mendelB flagged as an issue with a failing ghosty test and an issue he's seen before with other algolia searches.

### Relevant tickets

References [Pivotal #173713718](https://www.pivotaltracker.com/story/show/173713718).

### Checklist

### DIFF
--- a/resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js
+++ b/resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js
@@ -2,7 +2,6 @@ import { get } from 'lodash';
 import React, { useState } from 'react';
 
 import { featureFlag } from '../../../helpers/env';
-import { withoutNulls } from '../../../helpers/data';
 import SiteFooter from '../../utilities/SiteFooter/SiteFooter';
 import FilterNavigation from './FilterNavigation/FilterNavigation';
 import SiteNavigationContainer from '../../SiteNavigation/SiteNavigationContainer';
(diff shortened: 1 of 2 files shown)


DFurnes (approved):
Makes sense!

mendelB:
@lindsayrmaher I think we'd need to put the `withoutNulls` [here](https://github.com/DoSomething/phoenix-next/blob/4e27dd549ff1ba67a9fd4b23711cc98f30d53028/resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js#L110) in the `PaginatedCampaignsGallery` itself, since otherwise, [this call](https://github.com/DoSomething/phoenix-next/blob/03716ad136a3d00b5a31e36bf6bceba99e306118/resources/assets/components/pages/CausePage/CausePage.js#L56-L66) from the Cause Page will fail since it's not utilizing all the variables!

mendelB:
Ugh, and now that I'm saying this out loud, I'm realizing this _does_ need to happen over in GraphQL!

DFurnes:
Ahh I was just about to ask that @mendelB! Yeah, handling this in GraphQL makes sense to me.

lindsayrmaher:
I'm going to close this since we are fixing in graphql instead!

PR #2555, opened 2021-02-22
Start Voter Registration Form - Customizable Button Color

mendelB:
### What's this PR do?

This pull request allows a `buttonColor` prop in the `StartVoterRegistrationForm` component to customize the button color.

👀  commit-by-commit
Does this approach make sense?

A requirement for the Voter Registration Marketing Page is to customize the button color. The challenge was to dynamically 'tint' the color on hover state. This would have been a cinch with Tailwind's [opacity utility](https://tailwindcss.com/docs/background-color#changing-opacity), but that wouldn't work with non Tailwind colors assigned via the `bg-[color]`. 

I found [this article](https://www.sitepoint.com/javascript-generate-lighter-darker-color/) that maps out a really neat way to do this though, so pasted over a simplified version as a helper method and we should be golden!

### Relevant tickets

References [Pivotal #176905654](https://www.pivotaltracker.com/story/show/176905654/comments/222091036).

### Checklist

**Default**
image
**Hovered**
image

### DIFF
--- a/resources/assets/components/utilities/StartVoterRegistrationForm/StartVoterRegistrationForm.js
+++ b/resources/assets/components/utilities/StartVoterRegistrationForm/StartVoterRegistrationForm.js
@@ -7,6 +7,7 @@ import {
   EVENT_CATEGORIES,
   trackAnalyticsEvent,
 } from '../../../helpers/analytics';
+import ElementButton from '../Button/ElementButton';
 import Spinner from '../../artifacts/Spinner/Spinner';
 import { tailwind, colorLuminance } from '../../../helpers/display';
 import { getTrackingSource } from '../../../helpers/voter-registration';
@@ -98,28 +99,31 @@ const StartVoterRegistrationForm = ({
         />
       </div>
 
-      <button
+      <ElementButton
+        className="w-full"
         data-testid="voter-registration-submit-button"
-        className="btn w-full flex justify-center"
-        css={css`
-          background-color: ${buttonColor};
+        attributes={{
+          css: css`
+            background-color: ${buttonColor};
 
-          :hover {
-            background-color: ${colorLuminance(buttonColor, 10)};
-          }
-        `}
+            :hover {
+              background-color: ${colorLuminance(buttonColor, 10)};
+            }
+          `,
+        }}
         disabled={isDisabled || submitted}
         type="submit"
-      >
-        {submitted ? (
-          <>
-            <Spinner />
-            <span className="pl-1 pt-1">Processing...</span>
-          </>
-        ) : (
-          buttonText
-        )}
-      </button>
+        text={
+          submitted ? (
+            <>
+              <Spinner />
+              <span className="pl-1 pt-1">Processing...</span>
+            </>
+          ) : (
+            buttonText
+          )
+        }
+      />
     </form>
   );
 };
--- a/resources/assets/components/utilities/StartVoterRegistrationForm/StartVoterRegistrationForm.js
+++ b/resources/assets/components/utilities/StartVoterRegistrationForm/StartVoterRegistrationForm.js
@@ -101,7 +101,6 @@ const StartVoterRegistrationForm = ({
 
       <ElementButton
         className="w-full"
-        data-testid="voter-registration-submit-button"
         attributes={{
           css: css`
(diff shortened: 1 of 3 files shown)


lindsayrmaher (approved):
this approach makes sense to me! also i have never figured out why the opacity property from TW doesn't work with our custom colors so thanks for digging into that.

mendelB:
Thanks for the 🚤 review @lindsayrmaher! Added one critical update in https://github.com/DoSomething/phoenix-next/pull/2555/commits/4ae134d99bed1795db423bcb1d3f54439b79c88e to use the `fuchshyeah` hex in the docblock cc @weerd

resources/assets/helpers/display.js
```diff
@@ -145,4 +145,30 @@ export function toggleClassHandler(button, target, toggleClass) {
...
+
+    const colorLuminated = Math.round(
+      Math.min(Math.max(0, colorAsInt + colorAsInt * luminosity), 255),
+    ).toString(16);
+
+    luminatedValue += `00${colorLuminated}`.substr(colorLuminated.length);
+  }
+
+  return luminatedValue;
+}
```
weerd, line 172: I didn't get a chance to reply to the Pivotal Tracker ticket, but this is something I was wondering about if we were gonna allow for custom colors, we couldn't just do opacity changes! Very cool 👍 

One note, I wonder if it would be easier to think about if the luminosity was provided from a range of -100 to 100 instead of the decimal and values and just convert to decimal in the code 🤔  I suspect it'd make it clear, but not 100% sure on that; just a thought!

mendelB: Awesome! 

> One note, I wonder if it would be easier to think about if the luminosity was provided from a range of -100 to 100 instead of the decimal and values and just convert to decimal in the code 🤔 I suspect it'd make it clear, but not 100% sure on that; just a thought!

Agreed - I think that's a nice little touch! will update.

mendelB: Updated in https://github.com/DoSomething/phoenix-next/pull/2555/commits/04948282bc9c790d621233f237185b8b92e9c35b!

weerd: Quick note, but looks great! 💯

resources/assets/helpers/display.js
```diff
@@ -145,4 +145,30 @@ export function toggleClassHandler(button, target, toggleClass) {
   button.addEventListener('mousedown', clickHandler, false);
 }
 
+/**
+ * Alters luminosity for provided hex color by specified percentage.
+ * https://www.sitepoint.com/javascript-generate-lighter-darker-color/
+ *
+ * @param   {String} hex — a valid six character hex color value such as “#FF00FF".
+ * @param   {String} luminosity — the luminosity factor, i.e. -10 is 10% darker, 20 is 20% lighter, etc.
```
weerd, line 153: Maybe specify it's from -100 to 100?

mendelB: haha word! updated in https://github.com/DoSomething/phoenix-next/pull/2555/commits/a97fcbebbc3ae76cc56fb702458d1c8cfff010dd

weerd (approved):
💯  💡

resources/assets/components/utilities/StartVoterRegistrationForm/StartVoterRegistrationForm.js
```diff
@@ -96,27 +98,34 @@ const StartVoterRegistrationForm = ({
...
+      >
+        {submitted ? (
+          <>
+            <Spinner />
+            <span className="pl-1 pt-1">Processing...</span>
+          </>
+        ) : (
+          buttonText
+        )}
+      </button>
```
weerd, line 122: Just to confirm, you couldn't use one of these here, passing in modifications right?

mendelB: But alas this was the perfect occasion for our trusty `ElementButton`! Updated in https://github.com/DoSomething/phoenix-next/pull/2555/commits/ac7943ad1b13321c500cd8e953c7a072325d1886

weerd (approved):
Haha! Glad `ElementButton` worked out 🎉

PR #2558, opened 2021-02-23
Add RAF sixpack test 

Dinnall:
### What's this PR do?

This pull request adds a sixpack test to the thank you submission page.

We will be testing how users interact with the social share container and our recommended scholarship campaigns  

...

...
Screen Shot 2021-02-23 at 6 21 49 PM

Screen Shot 2021-02-23 at 6 46 05 PM

### Relevant tickets

References [Pivotal #174778100](https://www.pivotaltracker.com/n/projects/2401401/stories/174778100).

### Checklist


resources/assets/components/blocks/GalleryBlock/GalleryBlock.js
```diff
@@ -52,7 +52,13 @@ export const GalleryBlockFragment = gql`
   ${ExternalLinkBlockFragment}
 `;
 
-const renderBlock = (blockType, block, imageAlignment, imageFit) => {
+const renderBlock = (
+  blockType,
+  block,
+  imageAlignment,
+  imageFit,
+  sixpackConvertOnClick,
```
DFurnes, line 60: Ideally, the only component that needs to know about this test is the `ShowSubmissionPage` where the experiment is run. By passing down this test-specific prop, we've sorta "baked in" the experiment to all these components that live between. Is there a way we can make this more general?

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
```diff
@@ -125,6 +150,7 @@ const ShowSubmissionPage = ({ match }) => {
 
             <RecommendedCampaignsGallery
               excludeCampaignIds={campaignId ? [Number(campaignId)] : []}
+              sixpackConvertOnClick
```
DFurnes, line 153: What if, to make this more re-usable in the future, we instead passed an `onClick` handler here that ran alongside the gallery items' default behavior? In this file, that might look something like this:

```jsx
<RecommendedCampaignsGallery
  excludeCampaignIds={campaignId ? [Number(campaignId)] : []}
  onClick={() => sixpack().convertOnAction('galleryBlockClick')}
/>
```

That would allow us to keep all the things related to running this A/B test inside this one file (both the "setup" for opting in to the experiment & the conversion event for completing it). Then, after the test is completed, we can even leave the `onClick` behind in case we need similar functionality in the future!

resources/assets/components/utilities/ScholarshipCard/GalleryBlockSignup.js
```diff
@@ -100,6 +106,15 @@ const GalleryBlockSignup = ({ path, campaignId, campaignTitle }) => {
     });
   };
 
+  const handleClick = () => {
+    if (sixpackConvertOnClick) {
+      sixpack().convertOnAction('galleryBlockClick');
+    }
```
DFurnes, line 112: If we were to pass down an `onClick` prop, this could instead look something like this:

```js
if (onClick) {
  onClick();
}
```

Dinnall: Wouldn't this not be needed anymore if we take this path?
```
<RecommendedCampaignsGallery
  excludeCampaignIds={campaignId ? [Number(campaignId)] : []}
  onClick={() => sixpack().convertOnAction('galleryBlockClick')}
/>
```

Also, this means we can remove the `sixpackConvertOnClick` prop from the `ScholarshipCard` component as well

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
```diff
@@ -125,6 +151,7 @@ const ShowSubmissionPage = ({ match }) => {
 
             <RecommendedCampaignsGallery
               excludeCampaignIds={campaignId ? [Number(campaignId)] : []}
+              onClick={() => sixpack().convertOnAction('galleryBlockClick')}
```
mendelB, line 154: This is a really neat approach! We'll just need to pass this new `onClick` prop through the chain -- `RecommendedCampaignsGallery` -> `GalleryBlock` -> `ScholarshipCard` -> `GalleryBlockSignup` -- and then invoke the `onClick` there per https://github.com/DoSomething/phoenix-next/pull/2558#discussion_r582111498 for this to work.

Dinnall: I talked it through with Dave and from what I see I was still able to participate in the experiment

mendelB: Ah, hmm were you able to _convert_ on the experiment when clicking on one of the gallery cards? I would have assumed that this `onClick` never takes effect since we're only passing it through as a prop to the `RecommendedCampaignsGallery` but never actually _invoke_ it within the `GalleryBlockSignup` when the button is clicked 🤔

DFurnes: Yeah, we went back on forth on what was the best solution here! Michael had a good point that by putting the event listener on the gallery itself, we'd have a lot fewer other components we need to touch to run this test (and it's probably unlikely for anyone to click something _other_ than the button inside this gallery).

I do want to triple-check that this listener works before we merge, though. I was under the impression that you could only attach `onClick` handlers to HTML elements (e.g. `<div>`, `<a>`, etc.) and that when passed to React components (like `<RecommendedCampaignsGallery>`), it'd just be treated as a normal prop.

If that's the case, we could try adding a wrapping `<div>` in RecommendedCampaignsGallery that we bind the function provided in this prop to, and then then any clicks within any children of the gallery component should bubble up to fire that event as well.

mendelB: Ohhhhh word 🤯  that's smart!! 

But yeah I was under the same impression `onClick` handlers and HTML elements -- but that wrapping `<div>` solution is neat!

Dinnall: Okay, the flaky test is gone. We can review the Heroku build, if it isn't participating I can go ahead and add the `div` solution

mendelB: I added a thread for testing [here](https://dosomething.slack.com/archives/CUQMU4Q6B/p1614278664001500) and it seems like we'll need to use the wrapping `<div>` method for this!

Dinnall: Not sure how the `div` would look because from what I remember we can't have a click handler on a `div` directly only on anchor tags and buttons  
@DFurnes @mendelB

mendelB: @Dinnall It should work with `<div>`s too! I gave it a go locally and it seems to be converting 🎉 

```jsx
const RecommendedCampaignsGallery = ({ excludeCampaignIds, onClick }) => (
  <Query
    query={SCHOLARSHIP_CAMPAIGNS_QUERY}
    variables={{
      excludeIds: [
        ...excludeCampaignIds,
        ...siteConfig('hide_campaign_ids', []),
      ],
    }}
  >
    {result => (
      <div onClick={onClick}>
        <GalleryBlock
          blocks={(get(result, 'campaigns.edges') || []).map(
            edge => edge.node.campaignWebsite,
          )}
          galleryType="SCHOLARSHIP"
          itemsPerRow={3}
          imageAlignment="LEFT"
        />
      </div>
    ...
);
```

Dinnall: Did you include a `role` as an attribute for the `div`? @mendelB

mendelB: I didn't when testing, but seems like a `role="button"` might be necessary for the linter?

Dinnall: `tabIndex` and a `role` fixed the issue

DFurnes (approved):
Looks great! 🎉

PR #2538, opened 2021-02-03
Update HomePage.js

Dinnall:
### What's this PR do?

This pull request adjust the base font color for the anchor tags on the home page 

### How should this be reviewed?

Here is a [Codepen link](https://codepen.io/leah-morgan-angle/pen/rNWVPGm) reference provided by Leah

...

### Any background context you want to provide?

Screen Shot 2021-02-04 at 3 46 00 PM
Screen Shot 2021-02-04 at 3 45 48 PM

...

### Relevant tickets

References [Pivotal #174220514](https://www.pivotaltracker.com/n/projects/2401401/stories/174220514).

### Checklist

- [  x ] This PR has been added to the relevant Pivotal card.
- [ x  ] Screenshots

PR #2562, opened 2021-03-02
Text Link style sitewide updates

lindsayrmaher:
### What's this PR do?

This pull request updates our default styling for links to reflect the style that was added to the homepage in #2538 

I decided to approach this by setting up the default in our `base.scss` file, and then making specific style updates as needed in places we didn't want to use the default. It felt better to start from the baseline and then deal with special cases than vice versa. Let me know if there are other suggestions on approach or any possibilities to simplify.

Links that show default styling on the site should now reflect updated style guides from the design team (see below - this is the hover state), but links that were intended to have a specific design on the site will stay as is.

Screen Shot 2021-03-02 at 5 48 11 PM

### Relevant tickets

References [Pivotal #176810777](https://www.pivotaltracker.com/story/show/176810777).

### Checklist


resources/assets/components/pages/AccountPage/Subscriptions/CancelEmailSubscription.js
```diff
@@ -70,7 +70,7 @@ const CancelEmailSubscription = props => {
         <p>
           Need a break?
           <button
-            className="px-1 text-blue-500 pb-4"
+            className="px-1 text-blue-500 hover:text-blue-300 pb-4"
```
lindsayrmaher, line 73: added this as i was looking through links, it felt like it might be nice to have a distinctive color change to indicate it's a link.

resources/assets/scss/base.scss
```diff
@@ -19,91 +19,100 @@ $forge-path: '../../../node_modules/@dosomething/forge/assets/';
...
 
+// Updated baseline styling for links
+a {
+    @apply font-normal underline text-blurple-500;
+
+    &:hover {
+        color: theme('colors.blurple.300');
+        text-decoration-color: theme('colors.teal.500');
+    }
+}
```
weerd, line 74: Mind moving this up, above the `a > *` declaration? Since it's related to anchor links and trying to keep it all together 😄

resources/assets/components/utilities/SiteFooter/SiteFooter.js
```diff
@@ -7,6 +7,8 @@ import {
 } from '../../../helpers/analytics';
 import { query } from '../../../helpers/url';
 
+import './site-footer.scss';
+
```
weerd, line 11: Instead of importing a scss file, could we just add the respective tailwind classes to the anchors in the html below?

resources/assets/components/utilities/SiteFooter/site-footer.scss
```diff
@@ -0,0 +1,12 @@
...
+.footer {
+    a {
+        font-weight: bold;
+        text-decoration: none;
+
+        &:hover {
+            text-decoration: underline;
+        }
+    }
+}
```
weerd, line 12: Related to comment in footer file, but can probs remove this and just use tailwind classes in the footer file!

resources/assets/components/utilities/PageNavigation/page-navigation.scss
```diff
@@ -5,77 +5,79 @@
 $light-background-color: #f7fafc; // theme('colors.gray.100')
```
weerd, line 5: No real changes in this file, just spacing changes due to Prettier? 🤔

weerd (commented):
Looks good overall! Just a couple questions with simple fixes :)

resources/assets/components/utilities/SiteFooter/site-footer.scss
lindsayrmaher: yeah, the only reason i went this route instead was to avoid repeating it on each link since it's they are each written out vs a mapped through array type situation. but that definitely works too!

weerd: Some fancy `cmd + d` in VSCode for `class="` could help make it easier. But worth switching to using tailwind and eliminating an extra file :)

resources/assets/components/utilities/PageNavigation/page-navigation.scss
lindsayrmaher: they get lost because of the prettier updates, but I did add a couple lines on 59-60 for the `nav-link` class!

weerd (approved):
Looks good to me! 💯

PR #2557, opened 2021-02-23
Voter Registration Marketing Page Via Contentful 

mendelB:
### What's this PR do?

This pull request adds a new Contentful content type for the Voter Registration Marketing Page and renders the page via a Contentful query. It also runs the route through the backend to allow us to query and render the page's metadata.

### How should this be reviewed?
Commit-by-commit

### Any background context you want to provide?
- We use a `SectionBlock` for the content of this page since we want the same customization-ability of the story page (albeit limited to one section)
- The Instapage VR marketing pages use custom social metadata, so we need to utilize the backend (copying the pre-existing logic from e.g. [pages](https://github.com/DoSomething/phoenix-next/blob/3bf1d6bfe7de4e9fa1c7ab46e9697698b3a1e66d/app/Http/Controllers/PageController.php)) to query and render the appropriate metadata via the page's referenced `metadata` entry.
- We'll need to use the `ContentBlock` for the markdown collapsable items for the vote FAQs on these pages, but want the content to be spread full-width. We copy the logic from the general page (more context in https://github.com/DoSomething/phoenix-next/pull/2272) to allow customizing props per Rich Text embedded entry a la the `classNameByEntry`.
- I've disabled the Cypress test pending the accompanying GraphQL PR with the updated schema

### Relevant tickets

References [Pivotal #176895627](https://www.pivotaltracker.com/story/show/176895627).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints. _to come_
- [ ] Added screenshots of front-end changes on small, medium, and large screens. _unchanged_
- [ ] Added appropriate feature/unit tests. _to come_

PR #2564, opened 2021-03-03
Patch up the VR Marketing Page Migration Script

mendelB:
### What's this PR do?

This pull request fixes up the migration script added in #2557 which erroneously included migration script commands for all content types. (I must have mistakenly not specified the content type in one of the migration generation commands and committed this).


DFurnes (approved):
🙈

PR #2565, opened 2021-03-04
Refactoring Newsletter Subscriptions

weerd:
### What's this PR do?

This pull request does a bit of refactoring of the Newsletter Subscriptions (formerly known as _Email Subscriptions_) in preparation for added functionality with how the gallery of newsletter subscriptions shows up on the new _About_ page.

This PR mostly addresses the visual requirements and updates, but does not yet address the functionality. I created these as different components so they do not affect the existing "subscription center" on the profile and once ready can just swap in to replace the subscription center.

👀 

This is how it looks so far:
image

You should also be able to view on review app at: `{review_app_url}/us/about`.

The subscription center in the member profile area uses the old `gallery-grid` related classes which would not work great in the new layout of the About page that mimics the new Home page. Additionally, each email subscription item was technically made incorrectly, using the `Card` component which applies a slightly different design to the cards (border wraps all around, including the banner image). This does not represent the design correctly which follows the style of cards used on the homepage, where the border stops at the top banner image of the card.

While addressing these items and working to account for new functionality, it also presented an opportunity to refactor how these items are built; thus the new `NewsletterSubscriptionCard` and `NewsletterSubscriptionGallery` which more closely follows the approach used for the `CampaignCard`.

I'm also testing out an approach, since most of the content in the newsletter cards is hard-coded, I created a new `NewsletterSubscriptionCard/config.js` that contains constants for the content, banner, and topics. Ideally, thus simplifies using these newsletter cards anywhere on the site, and most of the hard-coded items can still be overridden using props.

### Relevant tickets

### Checklist


resources/assets/components/utilities/NewsletterSubscriptionCard/NewsletterSubscriptionCard.js
```diff
@@ -0,0 +1,60 @@
...
+        ) : null}
+      </div>
+    </article>
+  );
+};
+
+NewsletterSubscriptionCard.propTypes = {
+  description: PropTypes.string,
+  subtitle: PropTypes.string,
+  topic: PropTypes.oneOf(Object.keys(NEWSLETTER_TOPICS)).isRequired,
```
weerd, line 50: This is kinda cool! I can use the newletter topics to ensure only the existing ones are supplied. Throws an error in the console otherwise.

resources/assets/components/utilities/NewsletterSubscriptionCard/config.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Newsletter topic constants.
+ */
+export const NEWSLETTER_TOPICS = {
+  community: 'community',
+  lifestyle: 'lifestyle',
+  news: 'news',
+  scholarships: 'scholarships',
+};
```
weerd, line 14: Similar to the approach used with analytics `EVENT_CATEGORIES`, this ensures we're using the correct topics and avoids simple mistakes/typos.

weerd:
Next PR will tackle the functionality for when member is a guest vs authenticated and related functionality!

resources/assets/components/utilities/NewsletterSubscriptionCard/config.js
lindsayrmaher: i like this!!

lindsayrmaher (approved):
I really like this!! It feels a lot easier to digest vs having everything in one file. I'd like to refactor the rewards page in user profiles to do something similar :)

PR #2567, opened 2021-03-05
Defaults affiliate opt in to null in campaign signup form

mendelB:
### What's this PR do?

This pull request is a quick tweak to #2539 to assign the default `affiliateMessagingOptIn` state value to `null` so that we can exclude it from our Signup mutation payload by default instead of including `affiliateOptIn: false` in our `source_details` which seems lame.

https://github.com/DoSomething/phoenix-next/blob/acd285898e101e6a3702ad34a12e235577ab11a0/resources/assets/components/CampaignSignupForm/CampaignSignupForm.js#L182-L186


weerd (approved):
🙌

PR #2572, opened 2021-03-11
Newsletter signup UI updates

weerd:
### What's this PR do?

This pull request unifies the two different potential UIs for the Newsletter Signup component depending on whether a user is an authenticated member or a guest.

I refactored my prior refactor a bit, switched the `NewsletterSubscriptionGallery.js` to now be a `NewsletterSubscriptionForm` which more accurately describes what the interface is, an interactive form that behaves a little differently depending on authentication status.

There's still a bit of UI tweaks to do, but this gets the brunt of it done for the component and the two contexts. The functionality for subscribing to newsletters when authenticated works exactly as it does in the Profile Subscriptions. The functionality for guests is currently limited (can select newsletters, enter email and submit, but it doesn't do anything just yet) with it addressed in upcoming PR.

While it seems like a decent number of files, seven of them are either SVGs or a deleted file! 😅

One significant update, is that I changed around how the [`NewsletterSubscriptionCard` component](https://github.com/DoSomething/phoenix-next/pull/2572/files#diff-f1f48f177cd1f0602d928ca35e69e4d277e3e38a4536da1f289285d677d601dbR69-R74) works when used, where it now behaves more like a shell and you can essentially pass a button component or a checkbox component based on the authentication context. 

I also ended up creating a new `NewsletterSubscriptionFormInput` component that helps streamline the form markup a bit and does the auth check to decide which type of input to show (a button or a checkbox).

https://user-images.githubusercontent.com/105849/110868118-fcec1380-8295-11eb-8e46-96158ed50174.mp4

### Relevant tickets

### Checklist


weerd:
@DFurnes pinging you since I mentioned I would and @katiecrane for future reference on changes coming to the Subscription Center; once this is all ready, I'll swap them but all functionality should work as expected!

resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
```diff
@@ -0,0 +1,120 @@
...
+
+import { env } from '../../../helpers/env';
+import { NEWSLETTER_TOPICS } from './config';
+import PrimaryButton from '../Button/PrimaryButton';
+import { isAuthenticated } from '../../../helpers/auth';
+import NewsletterSubscriptionCard from './NewsletterSubscriptionCard';
+import NewsLetterSubscriptionFormInput from './NewsletterSubscriptionFormInput';
+
+const NewsletterSubscriptionForm = () => {
+  const [subscriptions, setSubscription] = useState([]);
```
mendelB, line 12: nit, but should the update method be pluralized as well?

```suggestion
  const [subscriptions, setSubscriptions] = useState([]);
```

resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
```diff
@@ -0,0 +1,120 @@
...
+  const handleOnChange = event => {
+    event.preventDefault();
+
+    setEmailValue(event.target.value);
+  };
+
+  const handleOnFocus = event => {
+    event.preventDefault();
+
+    console.log('handling input focus...');
```
mendelB, line 41: Will we be tracking an analytics event here?

resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
```diff
@@ -0,0 +1,120 @@
...
+      <p>Thanks for signing up!</p>
+      <p>
+        Want even more? Activate your account! We&apos;ve sent you an email
+        where you can set your password, valid for 72 hours.
+      </p>
+    </div>
+  ) : (
+    <form onSubmit={handleSubmit}>
+      <div className="newsletter-gallery mt-0 gap-8 grid grid-cols-1 md:grid-cols-2 xxl:grid-cols-4">
+        <NewsletterSubscriptionCard topic={NEWSLETTER_TOPICS.community}>
```
mendelB, line 69: Could we just map over the `NEWSLETTER_TOPICS` (via `Object.keys()`) and render a card for each value, or do you prefer the explicit rendering of each one?

resources/assets/components/utilities/NewsletterSubscription/ToggleSubscriptionButton.js
```diff
@@ -0,0 +1,86 @@
...
+      subscribed: $subscribed
+    ) {
+      id
+      emailSubscriptionTopics
+      emailSubscriptionStatus
+    }
+  }
+`;
+
+const ToggleSubscriptionButton = ({ topic }) => {
```
mendelB, line 38: Do we place on consolidating this with the [`EmailSubscriptionsItem`](https://github.com/DoSomething/phoenix-next/blob/504cf69016a080a66efe49071ac7976a19960a6b/resources/assets/components/pages/AccountPage/Subscriptions/EmailSubscriptionItem.js) component?

resources/assets/components/utilities/NewsletterSubscription/ToggleSubscriptionButton.js
```diff
@@ -0,0 +1,86 @@
...
+      deactivateText="Unsubscribe"
+      isDisabled={loading || modifying}
+      isLoading={loading || modifying}
+      isToggled={topics.includes(selectedTopic)}
+      onClick={() => {
+        console.log([topics, selectedTopic, !topics.includes(selectedTopic)]);
+
+        updateSubscription({
+          variables: {
+            userId: getUserId(),
```
mendelB, line 70: I don't think we need to pass this variable again since we assigned it in the initial `options` above?

```suggestion
```

mendelB (approved):
🌟 ☑️ 🗞️

mendelB:
@weerd You can rebase per https://github.com/DoSomething/phoenix-next/pull/2573 to get those failing tests to pass.

resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
weerd: yes! good 👁️

resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
weerd: Ya, I haven't added any analytics tracking yet. Upcoming PRs...

weerd: Meant to add a comment as such, but missed it! Will add.

resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
weerd: Ya, I was debating that. Still a possibility, and likely what I end up doing but this PR was more about just getting the functionality working and the UI in a better place. But ya, was thinking something similar for the final final*-official-final-ok-last-update (photoshop file name-style) version. 👍

resources/assets/components/utilities/NewsletterSubscription/ToggleSubscriptionButton.js
weerd: Potentially... but I needed to mimic the behavior and since the visual is way different than the newsletter signup for this build I didn't want to mess with it too much. I will need to update it for when I make a change on NS's API endpoint to accept an array apart from a string of a single newsletter.

resources/assets/components/utilities/NewsletterSubscription/ToggleSubscriptionButton.js
weerd: Ah fair. This was mostly a copy-🍝  from the Email Popover! But good catch as well 👍

PR #2570, opened 2021-03-11
Campaign Signup Form - Check Authentication Before Running Flashed Mutation

mendelB:
### What's this PR do?

This pull request adds a condition to ensure the user is authenticated before running the signup mutation for users ostensibly returning from the auth flow (using the flashed data).

### How should this be reviewed?
👀 

### Any background context you want to provide?
When spot checking this on the Apollo Studio app, [I see](https://studio.apollographql.com/graph/dosomething-graphql-lambda/operations?query=bea1d614135f336d05256365646004275f4a9425&queryName=CampaignSignupFormMutation&range=lastWeek&search=campaign&tab=traces&trace=dosomething-graphql-lambda~2021%3A03%3A11~7897174ce678e71ad8885c32de4d5e8e797e601cd52a4377962b2761c0602f98&traceBucket&variant=current) a bunch of failed mutations with the error:
`An access token is required for this query/mutation.`

My hunch is that these are users coming back from the auth flow without completing it, causing the mutation to fire off with a missing access token.

### Relevant tickets

References [Pivotal #175291206](https://www.pivotaltracker.com/story/show/175291206).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added screenshots of front-end changes on small, medium, and large screens.
- [ ] Added appropriate feature/unit tests.

PR #2573, opened 2021-03-12
Update Company Page Cypress test Stub to use intercept

mendelB:
### What's this PR do?

This pull request updates the stubbed POST requests in the company-page Cypress tests which were using the [deprecated](https://docs.cypress.io/api/commands/route.html) `cy.route()` and thus actually hitting our API (which was resulting in errors per an open issue discussed in [Slack](https://dosomething.slack.com/archives/CUQMTP0RM/p1615483885001600)) to use [`cy.intercept()`](https://docs.cypress.io/api/commands/intercept.html).

👀 

TL;DR: due to an unrelated update in Northstar, these scholarship POST requests started failing due to a CORS issue. This caused this test to fail (which seemed strange https://github.com/DoSomething/phoenix-next/pull/2570#issuecomment-796910304) but turns out it's bc it's actually hitting our API!

I'll add a Ghost Inspector test to cover this submission so that we have live test coverage. This one should be stubbed since it's solely testing the Phoenix functionality.

### DIFF
--- a/cypress/integration/company-page.js
+++ b/cypress/integration/company-page.js
@@ -44,11 +44,14 @@ describe('Company Page', () => {
 
     cy.visit(`/us/about/easy-scholarships`);
 
-    cy.route({
-      method: 'POST',
-      url: 'https://identity-dev.dosomething.org/v2/subscriptions',
-      status: 200,
-    });
+    cy.intercept(
+      'POST',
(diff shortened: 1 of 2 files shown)


lindsayrmaher (approved):
love cypress test improvements!!

PR #2575, opened 2021-03-18
Newsletter endpoint connection & validation errors UI.

weerd:
### What's this PR do?

This pull request connects the Newsletter Subscription component when viewed as a guest user to the updated [Northstar `/v2/subscriptions` endpoint](https://github.com/DoSomething/northstar/pull/1158).

It also does some cleanup and enables showing validation errors to the user:

https://user-images.githubusercontent.com/105849/111688783-a6d12000-8801-11eb-8db8-ac02e588a5a0.mp4

👀 

### Relevant tickets

### Checklist

### DIFF
--- a/resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
+++ b/resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
@@ -60,14 +60,10 @@ const NewsletterSubscriptionForm = () => {
         source: 'phoenix-next',
         source_detail: 'newsletter_subscriptions-about-page',
       })
-      .then(response => {
+      .then(() => {
         setShowConfirmation(true);
-
-        console.log(response);
       })
       .catch(error => {
-        console.log(error.response);
-
         setErrors(error.response.error);
       });
   };
--- a/resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
+++ b/resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
@@ -1,5 +1,7 @@
-import fetch from 'node-fetch';
+import { get } from 'lodash';
+import classnames from 'classnames';
 import React, { useState } from 'react';
+import { RestApiClient } from '@dosomething/gateway';
 
 import { env } from '../../../helpers/env';
 import { NEWSLETTER_TOPICS } from './config';
@@ -14,6 +16,8 @@ const NewsletterSubscriptionForm = () => {
   const [subscriptions, setSubscriptions] = useState([]);
 
   const [emailValue, setEmailValue] = useState('');
+
+  const [errors, setErrors] = useState(null);
 
   const [showConfirmation, setShowConfirmation] = useState(false);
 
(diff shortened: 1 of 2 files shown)


resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
```diff
@@ -68,53 +82,65 @@ const NewsletterSubscriptionForm = () => {
...
+              key={`newsletter-${key}`}
+              topic={NEWSLETTER_TOPICS[key]}
+            >
+              <NewsLetterSubscriptionFormInput
+                topic={NEWSLETTER_TOPICS[key]}
+                updateSubscriptions={updateSubscriptions}
+              />
+            </NewsletterSubscriptionCard>
+          );
+        })}
```
weerd, line 108: @mendelB went with the `.map()` 😄

resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
```diff
@@ -68,53 +82,65 @@ const NewsletterSubscriptionForm = () => {
...
-            className="block border-2 border-gray-300 border-solid leading-none outline-none focus:outline-2 focus:outline-blurple-100 focus:outline-solid px-4 py-3 placeholder-gray-400 rounded text-base w-full"
-            onChange={handleOnChange}
-            onFocus={handleOnFocus}
-            placeholder="Enter your email address"
-            type="email"
-            value={emailValue}
-          />
+        <div
+          className={classnames('md:flex md:max-w-xl mt-8 mx-auto', {
+            'pb-10': get(errors, 'fields.email', null),
```
weerd, line 120: This give the div a bit of room at the bottom to ensure it can show the absolutely-positioned error message below the input and not overlap into the next section!

resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
```diff
@@ -68,53 +78,65 @@ const NewsletterSubscriptionForm = () => {
...
+                updateSubscriptions={updateSubscriptions}
+              />
+            </NewsletterSubscriptionCard>
+          );
+        })}
       </div>
 
+      {get(errors, 'fields.email_subscription_topic', null) ? (
+        <p className="p-4 text-red-500">
+          Please select at least one newsletter to sign up for!
```
mendelB, line 109: Are we safe hardcoding this error message instead of customizing the Laravel response? (I suppose the edge case I'm thinking of is some sort of newsletter validation error other than 'required')

mendelB (approved):
🤩

resources/assets/components/utilities/NewsletterSubscription/NewsletterSubscriptionForm.js
weerd: I _thiiiiiiink_ so. For the most part not selecting a newsletter is the only error the user will see. Anything else would mean there's something royally wrong with the code.

That being said it is a good point and better to use Laravel's messaging. I didn't in this case cause it wouldn't make sense (something to the effect of "email_subscription_topic is required"). It would be ideal to override that and provide something better without creating a separate `FormRequest` class to handle it, but not sure offhand. Definitely something I'll look into though on the Northstar side of things.

PR #2579, opened 2021-03-24
About Page Header

weerd:
### What's this PR do?

This pull request builds out and styles the header section for the new About page.

https://user-images.githubusercontent.com/105849/112365688-539e1800-8cae-11eb-8d97-ec6ecafd1859.mp4

👀 

Design will be using a different cover image, so that will eventually get switched out to match the mocks.

### Relevant tickets

References [Pivotal #177168757](https://www.pivotaltracker.com/story/show/177168757).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AboutPage/AboutPage.js
+++ b/resources/assets/components/pages/AboutPage/AboutPage.js
@@ -4,6 +4,7 @@ import { Fragment, React } from 'react';
 import content from './about-page-content.json';
 import { tailwind } from '../../../helpers/display';
 import { isAuthenticated } from '../../../helpers/auth';
+import LinkButton from '../../utilities/Button/LinkButton';
 import SiteFooter from '../../utilities/SiteFooter/SiteFooter';
 import { contentfulImageUrl } from '../../../helpers/contentful';
 import PrimaryButton from '../../utilities/Button/PrimaryButton';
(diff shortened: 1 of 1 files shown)


resources/assets/components/pages/AboutPage/AboutPage.js
```diff
@@ -13,6 +16,37 @@ import StrikeThroughHeader from '../../utilities/SectionHeader/StrikeThroughHead
...
+        @media (min-width: ${tailwindScreens.lg}) {
+          background-image: url(${contentfulImageUrl(
+            content.coverImage.url,
+            '1440',
+            '539',
+            'fill',
+          )});
+        }
+      `
+    : null;
```
lindsayrmaher, line 48: could we[ export these styles from the homepage](https://github.com/DoSomething/phoenix-next/blob/04153bdbbeaee7b419e7a62ee77e0cf1b781f567/resources/assets/components/pages/HomePage/HomePage.js#L154) or save them somewhere to be reused rather than repeating here?

lindsayrmaher (approved):
I think this looks good, just a question about DRYing it up a lil!

resources/assets/components/pages/AboutPage/AboutPage.js
weerd: Ya! I think that would be ideal, just need to figure out best approach for that, maybe as a component or some helper function for visuals 🤔  I still have a bit of cleanup to finalize the About so I can figure out something and pop it in for an upcoming PR! Thanks for the suggestion @lindsayrmaher 🙌

PR #2580, opened 2021-03-24
Adds graphql query for users badges on Rewards tab!

lindsayrmaher:
### What's this PR do?

This pull request updates the `RewardsTab` component to a functional component using hooks vs a class and adds in a single query for a user and their badges vs calculating them when the page loads! 

The functionality and visuals on the page should stay the same, but we're using real data now instead of a dummy number in the URL 🎉 

What are thoughts around keeping the legacy file? I thought maybe it would be good for reference but also think it would make sense to just delete (def easier to read the PR).

Right now this will not calculate for badges that users earned before we added the field (that is pending with [this NS PR](https://github.com/DoSomething/northstar/pull/1171))

Added a fourth action badge!
Screen Shot 2021-03-25 at 1 28 58 PM

### Relevant tickets

References [Pivotal #175855166](https://www.pivotaltracker.com/story/show/175855166).

### Checklist


DFurnes:
> What are thoughts around keeping the legacy file? I thought maybe it would be good for reference but also think it would make sense to just delete (def easier to read the PR).

Let's delete it! 🔥 We can always reference this PR's diff or the Git history if we need to look back.

resources/assets/components/pages/AccountPage/Rewards/RewardsTab.js
```diff
@@ -18,27 +20,11 @@ import {
...
-`;
-
-const NEWSLETTER_BADGE = gql`
-  query SubscriptionTopicsBadgeQuery($userId: String!) {
+const USER_BADGE_QUERY = gql`
+  query UserBadgeCountQuery($userId: String!) {
     user(id: $userId) {
-      emailSubscriptionTopics
+      id
+      badges
```
DFurnes, line 27: This is so much simpler now! 🎉

resources/assets/components/pages/AccountPage/Rewards/RewardsTab.js
```diff
@@ -129,23 +124,14 @@ const badgeModalContent = {
   },
 };
 
-class RewardsTab extends React.Component {
-  constructor(props) {
-    super(props);
+const RewardsTab = ({ userId }) => {
+  const [modalName, setModalName] = useState('');
+  const [modalEarned, setModalEarned] = useState(false);
```
DFurnes, line 129: This feels like duplicating data in state that we could read elsewhere from the "source of truth" (the `badges` variable we receive from GraphQL). Could we instead name the modal by the badge's enum & then use that to read this data?

```jsx
onClick={() => showModal('SIGNUP')}
```

```jsx
{modalName ? (
        <BadgeModal onClose={closeModal} earned={badges.includes(modalName)} name={modalName}>
          <h1 className="font-league-gothic text-2xl">
            {badgeModalContent[modalName].title}
          </h1>
          <p
            data-testid={
              badges.includes(modalName) ? 'earned-badge-text' : 'unearned-badge-text'
            }
          >
            {modalEarned
              ? badgeModalContent[modalName].earnedText
              : badgeModalContent[modalName].unearnedText}
          </p>
        </BadgeModal>
      ) : null}
```

Then we could also rename the `badgeModalContent` and `BadgeImages` objects to be keyed by the enum, like `SIGNUP` (instead of `signupBadge`), but that might make things a tiny bit more readable/searchable too!

DFurnes (approved):
Looking great!!

resources/assets/components/pages/AccountPage/Rewards/RewardsTab.js
lindsayrmaher: I like this! I might flag it as a follow up just because I think changing the name might effect analytics tags as well and I want to run it by data/product before updating :)

PR #2581, opened 2021-03-24
About Page Intro Content 

weerd:
### What's this PR do?

This pull request builds out the hard-coded intro content section which includes a youtube video.

https://user-images.githubusercontent.com/105849/112393568-ab01af80-8cd1-11eb-85a5-cbfdbdad928e.mp4

👀 

I also removed the placeholder images for the hard-coded campaigns and added the correct URLs.

### Relevant tickets

References [Pivotal #177168863](https://www.pivotaltracker.com/story/show/177168863).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AboutPage/AboutPage.js
+++ b/resources/assets/components/pages/AboutPage/AboutPage.js
@@ -1,6 +1,7 @@
 import { Fragment, React } from 'react';
 
-import content from './about-page-content.json';
+import data from './about-page-content.json';
+import Embed from '../../utilities/Embed/Embed';
 import { tailwind } from '../../../helpers/display';
 import { isAuthenticated } from '../../../helpers/auth';
 import SiteFooter from '../../utilities/SiteFooter/SiteFooter';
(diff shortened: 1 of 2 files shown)


resources/assets/components/pages/AboutPage/AboutPage.js
```diff
@@ -27,6 +28,26 @@ const AboutPageTemplate = () => {
...
+          <section className="base-12-grid bg-white py-12">
+            <Embed
+              className="col-span-4 md:col-span-8 lg:col-span-10 xl:col-span-4 lg:col-start-2 xl:col-start-8 xl:mt-12 xl:row-start-1 self-start"
+              url={data.content.media.url}
+            />
+
+            <div className="col-span-4 md:col-span-8 lg:col-span-10 xl:col-span-6 lg:col-start-2 xl:col-start-2 xl:row-start-1 mt-4 xl:mt-0">
+              <h2 className="mb-0 text-lg text-purple-400 uppercase">
+                About Us
+              </h2>
```
weerd, line 41: If we end up using this grid split of 6/10 (left) and 4/10 (right) we may want to add a custom component class like we do for some of the other layouts so there's less need to mess with all the changing `col-*` classes, etc.

resources/assets/components/pages/AboutPage/about-page-content.json
```diff
@@ -49,12 +49,19 @@
...
       "path": "/"
     }
   ],
+  "content": {
+    "title": "A digital platform powering real-world impact.",
+    "copy": "As the largest not-for-profit exclusively for young people and social change, DoSomething’s millions of members represent every US area code and 131 countries. Using our digital platform, DoSomething members join our volunteer, social change, and civic action campaigns to make real-world impact on causes they care about. Sorry in advance for the #humblebrag, but here are just a few of our greatest campaigns.",
+    "media": {
+      "url": "https://www.youtube.com/embed/Ph1Rb1I5tmg?controls=0"
+    }
+  },
```
weerd, line 64: Been digging having this JSON file with the hard-coded content. It should make it easier in the future if we decide to make this an editable page on Contentful, and in general it's also easy to edit for devs to jump right to the string in the JSON file than potentially need to awkwardly traverse prettified JSX.

mendelB (approved):
🖼️

PR #2584, opened 2021-03-26
Call To Action Banner component

weerd:
### What's this PR do?

This pull request creates a new utility component directory called `CallToAction` which will house the different call to action components needed (similar to the pattern established by the `Gallery` component directory).

First up is the new `BannerCallToAction`, which is a full width page banner CTA. There's some flexibility built into the component:

- Use the `stacked` prop to indicate that even on large screens, the alignment of the text and button should remain stacked in the center; otherwise it will have the message text on the left and button on the right.
- The button is passed as a child component to the `BannerCallToAction`, so that any kind of button can be used and does not have to be predefined by this component; the button can define its own colors, url, etc...
- Lastly, the background and text colors can be customized via class names using the `colorClasses` object prop, with `background` and `text` keys.

I also switched out the page cover image URL with the new one from Creative!

👀 

Stacked Banner CTA:
image

Side-by-side Banner CTA:
image

Random Colors Stacked Banner CTA Example:
image

This should help @lindsayrmaher with the upcoming Articles page work!

@mendelB I also made an update to the `AnalyticsWaypoint` component to allow passing classes. I ended up using this to better position the waypoint both in the markup and via styling to avoid conflicts we've seen before with it conflicting with the grid and taking up a spot in a column. Essentially I aboslutely positioned them with utility classes at the top & bottom of the component they are working in. I tested using the DS Analytics Debugger and had the waypoints fire exactly as expected! 🎉 

### Relevant tickets

References [Pivotal #177468614](https://www.pivotaltracker.com/story/show/177468614).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AboutPage/AboutPage.js
+++ b/resources/assets/components/pages/AboutPage/AboutPage.js
@@ -146,17 +146,13 @@ const AboutPageTemplate = () => {
           </section>
 
           <BannerCallToAction
-            colorClasses={{
-              background: 'bg-purple-500',
-              text: 'text-teal-500',
-            }}
             message="Earn scholarships through community service"
             title="Easy Scholarships"
             stacked
           >
             <PrimaryButton
               attributes={{ 'data-label': 'scholarships_cta_authorize' }}
-              className="mt-8 xl:m-0 py-4 px-16 text-lg xl:ml-auto"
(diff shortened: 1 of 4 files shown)


lindsayrmaher (approved):
looking goooooooood!

PR #2593, opened 2021-04-02
Fix Flakey Questionnaire Test

mendelB:
### What's this PR do?

This pull request - hopefully - fixes what seems to be a flakey test in the Questionnaire Action Cypress suite. I think the issue is that we're stubbing the POST request after the login flow which might be too late for the triggered POST request that happens after the login.

👀


lindsayrmaher (approved):
seems reasonable to me, and i hate a flaky test!

PR #2594, opened 2021-04-02
Create Articles Page content type

lindsayrmaher:
### What's this PR do?

This pull request creates a new content type in Contentful called `articlesPage` with fields that editors will use to build out the new in house articles page to replace Instapage!

Anything to add in documentation? Anything required that doesn't need to be? Anything missing that needs to be added?

I removed having an explicit `title` field because I felt like we could pull the title from the article itself. The featured article should be in the latest articles gallery similar to how we have the homepage setup, and it felt duplicative to have editors add a title there. I'm wondering if the same would be true about the cover image. [Here are the designs](https://app.abstract.com/projects/fadf0790-f1f9-11e6-9a30-b54ba7e8e705/branches/a9688e06-15bd-4d39-92e8-a7e688f9645c/commits/c50a405cb08fdf5bb6d367c44c554ebedf46d9af/files/c8cfdbb7-44da-408d-adeb-c88499272f2a/layers/CE85B7FC-0F17-4677-AD1B-6674A676E6F2?collectionId=e7dc0263-183e-47eb-890a-ac94286d2f94&collectionLayerId=684f5031-d92e-490f-b22a-6c899fd844c4) if that's helpful to reference.

### Relevant tickets

References [Pivotal #177518391](https://www.pivotaltracker.com/story/show/177518391).

### Checklist

### DIFF
--- a/contentful/content-types/articlesPage.js
+++ b/contentful/content-types/articlesPage.js
@@ -13,15 +13,6 @@ module.exports = function(migration) {
     .validations([])
     .disabled(false)
     .omitted(false);
-  articlesPage
-    .createField('title')
-    .name('Title')
-    .type('Symbol')
-    .localized(true)
-    .required(true)
-    .validations([])
-    .disabled(false)
-    .omitted(false);
 
   articlesPage
     .createField('coverImage')
@@ -234,7 +225,6 @@ module.exports = function(migration) {
     .disabled(false)
     .omitted(false);
   articlesPage.changeFieldControl('internalTitle', 'builtin', 'singleLine', {});
-  articlesPage.changeFieldControl('title', 'builtin', 'singleLine', {});
 
   articlesPage.changeFieldControl('coverImage', 'builtin', 'assetLinkEditor', {
     helpText:
--- a/contentful/content-types/articlesPage.js
+++ b/contentful/content-types/articlesPage.js
@@ -36,8 +36,8 @@ module.exports = function(migration) {
     .linkType('Asset');
 
   articlesPage
-    .createField('latestArticlesGalleryTitle')
-    .name('Latest Articles Gallery Title')
+    .createField('featuredArticlesGalleryTopTitle')
+    .name('Featured Articles Gallery Top Title')
     .type('Symbol')
     .localized(false)
     .required(true)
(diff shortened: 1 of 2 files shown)


contentful/content-types/articlesPage.js
```diff
@@ -0,0 +1,325 @@
...
+    .createField('latestArticlesGallery')
+    .name('Latest Articles Gallery')
+    .type('Array')
+    .localized(false)
+    .required(true)
+    .validations([
+      {
+        size: {
+          min: 7,
+          max: 7,
```
mendelB, line 58: In the design it looks like we're displaying 6. Is the idea that the first one going to be pulled in as the 'featured' article for the page header? (I'm gathering this from your point in the description:)

> I removed having an explicit title field because I felt like we could pull the title from the article itself. The featured article should be in the latest articles gallery similar to how we have the homepage setup, and it felt duplicative to have editors add a title there. I'm wondering if the same would be true about the cover image. Here are the designs if that's helpful to reference.

Hmmm I have two thoughts:
1. I wonder if we could instead create a more formal `headerArticle` (or `featuredArticle`) reference field, and pull the title/cover image/link from there!
2. But since this is the page header and thus a pretty critical component, part of me wonders if it's prudent to allow editor customization and this keep them as separate static fields.
3. Or..... both! Maybe we allow overriding the `headerArticle` fields via separate fields for the header, which if empty will just be auto filled via the `headerArticle`?

contentful/content-types/articlesPage.js
```diff
@@ -0,0 +1,325 @@
...
+          max: 20971520,
+        },
+      },
+    ])
+    .disabled(false)
+    .omitted(false)
+    .linkType('Asset');
+
+  articlesPage
+    .createField('latestArticlesGalleryTitle')
```
mendelB, line 39: Could it make sense to keep this (and perhaps the `elevenFactsGalleryTitle` as well) more open-ended like the `topicArticlesGalleryOneTitle`? (So that we're not bound to these galleries being 'latest' and 11-facts, but instead like maybe a 'featured gallery top' and 'featured gallery bottom' or something?)

...and this is just a thought -- could be overkill if we're intent on always sticking to this format!

mendelB (approved):
👏 for documentation! I left a couple of thoughts as well!

contentful/content-types/articlesPage.js
lindsayrmaher: @katiecrane do you have any opinions here? I kind of like the idea of a space to just put the main article reference, but am wondering about the level of customization.

katiecrane: great questions! I think they'll want full customization available for everything in the header (photo, text, etc) but I'm gonna double check with Leah

lindsayrmaher:
@katiecrane and @mendelB pushed up changes based on the slack thread! I created the headerTitle field and then created a `headerArticle` field that we can pull the link from and a cover image and headline as a backup from the custom!

mendelB (approved):
👏 (I think we just need to update the documentation to reflect the changes!)

PR #2595, opened 2021-04-06
Photo Uploader Custom File Dimension Validation Message

mendelB:
### What's this PR do?

This pull request adds a custom message for `The file has invalid image dimensions.` errors on RB photo files.

👀 

This is a hack! As discussed in the [Slack thread](https://dosomething.slack.com/archives/CP2D7UGAU/p1614785550000700) linked in Pivotal, we don't have a neat way to insert links in validation messages and thus need to customize it from our end. 

👉  **Q:** My arbitrary call was to also customize the message on our end once we were inserting a link since it's very RB specific, but I could see it making sense for us to just customize that text bit from Northstar itself?

I should note that @weerd had [an interesting idea](https://dosomething.slack.com/archives/CP2D7UGAU/p1614816924013900?thread_ts=1614785550.000700&cid=CP2D7UGAU) for implementation via Laraval (And edit: can we include machine friendly codes per https://github.com/DoSomething/phoenix-next/pull/2595#issuecomment-814218762) -- but I figured this might be too involved for this build as it'll require some general changes in the way we parse these responses from the API on Phoenix, and could wait for a larger refactor (perhaps when we start speaking directly to the Northstar API, or using GraphQL for Posts).

Next up -- an update in Northstar to ensure we return the standard validation message for this field [in the first place](https://dosomething.slack.com/archives/CP2D7UGAU/p1614787402003800?thread_ts=1614785550.000700&cid=CP2D7UGAU)! 

### Relevant tickets

References [Pivotal #177189106](https://www.pivotaltracker.com/story/show/177189106).

### Checklist

photo-submission-action-dimension-errors


resources/assets/components/utilities/Form/FormValidation.js
```diff
@@ -15,6 +16,30 @@ const FormValidation = ({ response }) => {
   const statusMessage = getStatusMessage(response);
   const errorMessages = getFieldErrorMessages(response);
 
+  const renderErrorMessage = errorMessage => {
+    // @HACK: We render a link to the Help Center along with a customized validation message
+    // for file dimension errors. Can we somehow formalize this?
+    if (errorMessage === 'The file has invalid image dimensions.') {
```
DFurnes, line 22: For a hack, this is awfully clean & well-documented! 😸

docs/development/content-types/photo-submission-action.md
```diff
@@ -11,11 +11,17 @@ Once the user submits the form successfully, they will be redirected to the Show
 ## Content Type Fields
 
 -   **Internal Title**: This is for our internal Contentful organization and will be how the block shows up in search results, etc.
--   **Action ID**: The Action ID associated with this action in Rogue.
+-   **Action ID**: The Action ID associated with this action in Northstar.
```
DFurnes, line 14: 🌟

DFurnes (approved):
Looks good!

DFurnes:
> My arbitrary call was to also customize the message on our end once we were inserting a link since it's very RB specific, but I could see it making sense for us to just customize that text bit from Northstar itself?

Yeah, that'd be lovely since we'd be able to read the dimensions straight from the settings on the validator. However, that could also make the conditional to render this "linkified" version of the message on Phoenix a lot more brittle...

This really makes me wish we returned "machine-friendly" error codes alongside the human-readable text (like `dimensions:50,50,5000,4000` in the response) that we could then parse more easily in application code...

mendelB:
> > My arbitrary call was to also customize the message on our end once we were inserting a link since it's very RB specific, but I could see it making sense for us to just customize that text bit from Northstar itself?
> 
> Yeah, that'd be lovely since we'd be able to read the dimensions straight from the settings on the validator. However, that could also make the conditional to render this "linkified" version of the message on Phoenix a lot more brittle if we're matching on a longer string that might change if we change dimensions...
> 
> This really makes me wish we returned "machine-friendly" error codes alongside the human-readable text (like `dimensions:50,50,5000,4000` in the response) that we could then parse more easily in application code...

Yes! Hmmm, perhaps, since the only validation on the `file` we currently have _is_ the dimension validation, we could just rely on a more narrow string match like 'contains `Try copping your photo`' which I'd willing to bet is a string unique to this use case?

DFurnes:
Yeah, that could work! I think we'd want to leave a note in Northstar warning us that that validation message is something that we rely on in Phoenix's application logic so we don't get surprised if we ever want to adjust that language.

mendelB:
> Yeah, that could work! I think we'd want to leave a note in Northstar warning us that that validation message is something that we rely on in Phoenix's application logic so we don't get surprised if we ever want to adjust that language.

Wise! That does seem cleaner overall though -- will adjust!

PR #2602, opened 2021-04-08
Media Uploader Updates!

mendelB:
### What's this PR do?

This pull request refactors + embellishes the `MediaUploader` component with some UX updates for error handling.

commit-by-commit!

Notabely:
https://github.com/DoSomething/phoenix-next/commit/5bbafeca3282810103ce37dbf1209c2640a0ce67 Adds internal error handling. In case of unsupported file type uploads, we'd only ever console log the error. This will display the error within the uploader! We also make use of the [`accept` attribute](https://www.w3schools.com/tags/att_input_accept.asp) on the `<input>` to instruct browsers to only allow selecting supported filetypes in the first place! (This isn't actually validated & can be overwritten, so we still need to handle the rest of the logic ourselves).

https://github.com/DoSomething/phoenix-next/commit/8a24683395cc99c293aa73fc739c66eab09d4171 exciting! This will display an error (and reject the upload) if the file size is more than our 10MB limit. Previously, this would result in the photo uploader exploding since we don't explicitly handle the [HTTP `413` response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413). (My hunch is that this is one of the sources of those cryptic errors users were submitting Zendesk tickets for, [discussed here](https://dosomething.slack.com/archives/C03T8SDLY/p1614965219032900).)

https://github.com/DoSomething/phoenix-next/commit/22dc70528fb3ba24ea35146076bac3879bb1583a We hook into the [`invalid` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/invalid_event) on the `<input>` to display a more elegant error message when no photo is provided and the user submits. (This would previously [awkwardly display](https://user-images.githubusercontent.com/12417657/114101861-58f58800-9894-11eb-80b9-949965adc52c.png) the native HTML validation).

https://github.com/DoSomething/phoenix-next/commit/527a0d15a2fe6eaf2c87eba2fb3d145f023c8437, https://github.com/DoSomething/phoenix-next/commit/f7609ce6ab23ea7f1becb8490bdf2982c997bcbc When trying to add tests for this component led to some cryptic errors that seemed to have to do with our rather involved way of parsing the file values -- I dug around and found that the `File` object supports a [`type` property](https://developer.mozilla.org/en-US/docs/Web/API/File/type) to inspect the `'MIME' type` which allows for a drastically simpler file processing flow.

https://github.com/DoSomething/phoenix-next/commit/5e8e867ec16abca8b3308a3b0b55d8ffbcf95df1 Tests!

Cleanup sprint!

### Relevant tickets

References [Pivotal #177249428](https://www.pivotaltracker.com/story/show/177249428).

### Checklist

📸 
[Video Walkthrough](https://share.getcloudapp.com/nOuoe9gK)


DFurnes (approved):
Looks fantastic! I loved the video walkthrough & atomic commits! 🙇

PR #2608, opened 2021-04-13
Articles Page Cypress Test

lindsayrmaher:
### What's this PR do?

This pull request adds cypress tests for the newsletter signup on the Articles Page. It also updates the graphql schema to include the articles page query.

Does this seem like reasonable coverage?

Want to make sure the input field is working as expected.

### Relevant tickets

References [Pivotal #177518693](https://www.pivotaltracker.com/story/show/177518693).

### Checklist

### DIFF
--- a/cypress/integration/articles-page-newsletter-signup.js
+++ b/cypress/integration/articles-page-newsletter-signup.js
@@ -1,23 +1,10 @@
 /// <reference types="Cypress" />
 
-import faker from 'faker';
-
 describe('Articles Page', () => {
   beforeEach(() => cy.configureMocks());
 
   /** @test */
   it('Displays the newsletter signup section', () => {
-    cy.mockGraphqlOp('ArticlesPageQuery', {
-      page: {
-        coverImage: {
-          url: faker.image.imageUrl(),
-        },
-        internalTitle: 'testing title',
-        headerTitle: 'Testing',
-        headerLinkUrl: 'https://dosomething.org',
-      },
-    });
-
     cy.withFeatureFlags({ new_articles_page: true }).visit(`/us/articles`);
     cy.findByTestId('newsletter-section-articles-page').should(
       'have.length',
--- a/cypress/integration/articles-page-newsletter-signup.js
+++ b/cypress/integration/articles-page-newsletter-signup.js
@@ -1,43 +1,20 @@
 /// <reference types="Cypress" />
 
+import faker from 'faker';
+
 describe('Articles Page', () => {
   beforeEach(() => cy.configureMocks());
 
   /** @test */
   it('Displays the newsletter signup section', () => {
     cy.mockGraphqlOp('ArticlesPageQuery', {
-      articlesPage: {
-        id: '1',
+      page: {
+        coverImage: {
+          url: faker.image.imageUrl(),
+        },
         internalTitle: 'testing title',
         headerTitle: 'Testing',
         headerLinkUrl: 'https://dosomething.org',
-        featuredArticlesGalleryTopTitle: 'The Top Titles',
-        featuredArticlesGalleryTop: [
-          {
-            showcaseTitle: 'Article Title',
-            showcaseDescription: 'article description',
-          },
-          {
-            showcaseTitle: 'Article Title',
-            showcaseDescription: 'article description',
-          },
-          {
-            showcaseTitle: 'Article Title',
-            showcaseDescription: 'article description',
-          },
-          {
-            showcaseTitle: 'Article Title',
-            showcaseDescription: 'article description',
-          },
-          {
-            showcaseTitle: 'Article Title',
-            showcaseDescription: 'article description',
-          },
-          {
-            showcaseTitle: 'Article Title',
-            showcaseDescription: 'article description',
-          },
-        ],
       },
     });
 
--- a/cypress/integration/articles-page-newsletter-signup.js
+++ b/cypress/integration/articles-page-newsletter-signup.js
@@ -1,0 +1,69 @@
+/// <reference types="Cypress" />
+
+describe('Articles Page', () => {
+  beforeEach(() => cy.configureMocks());
+
+  /** @test */
+  it('Displays the newsletter signup section', () => {
+    cy.mockGraphqlOp('ArticlesPageQuery', {
+      articlesPage: {
+        id: '1',
+        internalTitle: 'testing title',
+        headerTitle: 'Testing',
+        headerLinkUrl: 'https://dosomething.org',
+        featuredArticlesGalleryTopTitle: 'The Top Titles',
+        featuredArticlesGalleryTop: [
+          {
+            showcaseTitle: 'Article Title',
+            showcaseDescription: 'article description',
+          },
+          {
+            showcaseTitle: 'Article Title',
+            showcaseDescription: 'article description',
+          },
+          {
+            showcaseTitle: 'Article Title',
+            showcaseDescription: 'article description',
+          },
+          {
+            showcaseTitle: 'Article Title',
+            showcaseDescription: 'article description',
+          },
+          {
+            showcaseTitle: 'Article Title',
+            showcaseDescription: 'article description',
+          },
+          {
+            showcaseTitle: 'Article Title',
+            showcaseDescription: 'article description',
+          },
+        ],
+      },
+    });
+
+    cy.withFeatureFlags({ new_articles_page: true }).visit(`/us/articles`);
+    cy.findByTestId('newsletter-section-articles-page').should(
+      'have.length',
+      1,
+    );
+  });
+
+  /** @test */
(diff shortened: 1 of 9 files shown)


cypress/integration/articles-page-newsletter-signup.js
```diff
@@ -0,0 +1,69 @@
...
+          {
+            showcaseTitle: 'Article Title',
+            showcaseDescription: 'article description',
+          },
+        ],
+      },
+    });
+
+    cy.withFeatureFlags({ new_articles_page: true }).visit(`/us/articles`);
+    cy.findByTestId('newsletter-section-articles-page').should(
```
weerd, line 45: Not sure if this is anything, but this `data-test` ID is present in the "Lifestyle Newsletter Signup Section" but not in the "Featured Gallery Section Top" that the above test is rendering, so it might just be failing because of this? I checked the `ArticlesPage.js` and looks like that "Featured Gallery Section Top" has a `data-test` ID of `campaigns-section` which should probably be updated as well!

weerd:
Separate note unrelated to the bug, but I wonder if there's a better name for the `NewsletterSubscriptionFormArticlesPage` component. If we ever use it outside the Articles page it'd be a bit weird and confusing, haha! Not crucial but lets think of a better name and could mean changing the name of the `NewsletterSubscriptionForm` as well to better describe what the differences are between the two 😄

resources/assets/components/utilities/NewsletterSubscription/ToggleSubscriptionButton.js
```diff
@@ -9,7 +9,7 @@ import { getUserId } from '../../../helpers/auth';
 import ToggleButton from '../Button/ToggleButton';
 
 const EMAIL_SUBSCRIPTION_QUERY = gql`
-  query EmailSubscriptionsQuery($userId: String!) {
+  query ToggleSubscriptionsButtonQuery($userId: String!) {
```
lindsayrmaher, line 12: had to update these due to an error with Apollo that came up when I tried to fetch the graphql schema! There is another component that had a query with the exact same name.

resources/assets/components/utilities/NewsletterSubscription/ToggleSubscriptionButton.js
```diff
@@ -18,7 +18,7 @@ const EMAIL_SUBSCRIPTION_QUERY = gql`
 `;
 
 const EMAIL_SUBSCRIPTION_MUTATION = gql`
-  mutation EmailSubscriptionTopic(
+  mutation EmailSubscriptionTopics(
```
lindsayrmaher, line 21: Same issue here re: the duplicate naming.

resources/assets/components/pages/ArticlesPage/ArticlesPage.js
```diff
@@ -201,12 +200,19 @@ const ArticlesLandingPage = ({
           </section>
 
           {/* Article Archive Call To Action Banner */}
-          <BannerCallToAction text={ctaText} title={ctaTitle} stacked>
+          <BannerCallToAction
+            text={
+              ctaText ||
+              'An index of all articles and videos by DoSomething.org'
```
lindsayrmaher, line 206: I was running into issues where even if the field was being sent over as `null` because it wasn't filled out, the default props weren't being added in. So i decided to take this approach instead. Let me know if there's a better way!

cypress/integration/articles-page.js
```diff
@@ -0,0 +1,59 @@
...
+            },
+          },
+        },
+      },
+    );
+
+    cy.findByTestId('articles-page-email-input').type('dskafjhdakj');
+    cy.findByTestId('articles-page-newsletter-signup-button').click();
+
+    cy.should('not.contain', 'Thanks for signing up!');
```
mendelB, line 57: Since the `<input>` has `type="email"`, I don't think this will even submit in the first place since it'll be validated in the browser as a valid email address. Maybe worth testing that the backend validation appears? e.g.

```js
cy.intercept(
      'POST',
      'https://identity-dev.dosomething.org/v2/subscriptions',
      {
        statusCode: 422,
        body: {
          error: {
            message: 'Failed validation.',
            fields: {
              email: ['Yikes! Our servers are on their lunch break.'],
            },
          },
        },
      },
    );

    cy.findByTestId('articles-page-email-input').type('test@example.com');
    cy.findByTestId('articles-page-newsletter-signup-button').click();

    cy.should('contain', 'Yikes! Our servers are on their lunch break.');
```

resources/assets/components/pages/ArticlesPage/ArticlesPage.js
```diff
@@ -13,7 +13,7 @@ import { coverImageMediaQueryStyles } from '../../../helpers/display';
 import BannerCallToAction from '../../utilities/CallToAction/BannerCallToAction';
 import SiteNavigationContainer from '../../SiteNavigation/SiteNavigationContainer';
 import StrikeThroughHeader from '../../utilities/SectionHeader/StrikeThroughHeader';
-import NewsletterSubscriptionFormArticlesPage from '../../utilities/NewsletterSubscription/NewsletterSubscriptionFormArticlesPage';
+import SingleNewsletterSubscriptionForm from '../../utilities/NewsletterSubscription/SingleNewsletterSubscriptionForm';
```
mendelB, line 16: Based on this naming, I might expect this component to accept an `emailSubscriptionTopic` prop and use that to assign the `email_subscription_topic` value in the payload -- may be worth making that quick update?

resources/assets/components/pages/ArticlesPage/ArticlesPage.js
mendelB: React uses `null` as a valid prop value and so wouldn't resort to using the default prop in the first place 😖 

What we've done in the past (e.g. [`CollectionPage`](https://github.com/DoSomething/phoenix-next/blob/3b5248353a488a6abba8ae52c94f7d26fbe0cbac/resources/assets/components/pages/CollectionPage/CollectionPage.js#L109)) is used the `withoutNulls` helper to avoid assigning any optional fields as `null`! Than the values are unassigned (or `undefined`) which triggers the default props 😌

resources/assets/components/utilities/NewsletterSubscription/ToggleSubscriptionButton.js
mendelB: I feel like the plural/singular naming here might get confusing 😖 maybe worth naming this in line with the above re-named query specific to this component? (e.g. `ToggleSubscriptions` or something.)

Or..... could we just export and re-use the existing query/mutation since they're identical 🤔

mendelB (approved):
🧪

resources/assets/components/utilities/NewsletterSubscription/ToggleSubscriptionButton.js
lindsayrmaher: yeah i was just re thinking this and might just export those two query/mutations @mendelB

resources/assets/components/utilities/NewsletterSubscription/ToggleSubscriptionButton.js
weerd: Ya, that's my bad. I thought I needed to use the same named query and got confused with that and actually needs to be uniquely named! 😖

resources/assets/components/utilities/NewsletterSubscription/ToggleSubscriptionButton.js
```diff
@@ -1,39 +1,15 @@
...
-      id
-      emailSubscriptionTopics
-      emailSubscriptionStatus
-    }
-  }
-`;
+import {
+  EMAIL_SUBSCRIPTION_QUERY,
+  EMAIL_SUBSCRIPTION_MUTATION,
+} from '../../pages/AccountPage/Subscriptions/EmailSubscriptionItem';
```
weerd, line 12: Could we actually switch this around and make the query in `ToggleSubscriptionButton.js` be the canonical one? The Email Subscription stuff sitting in the `AccountPage` directory is going to get removed and replaced with the new version I created for the About Page. I just didn't want to break functionality for the old version on the Account Page while the new one was in development and not launched yet!

weerd: I probably should have added a comment in the old version to make the intent more explicit. This is why I really want to establish a naming convention that quickly clues devs into intent... like if the `EmailSubscriptionItem` was renamed to `EmailSubscriptionItemLegacy` it'd be more obvious that it's going to go away.

weerd: Appreciate the switcheroo @lindsayrmaher! Now that this is merged I'll get a PR in this week to deprecate the old email subscription component in favor of the new one and switch the old one out! 🙌

PR #2609, opened 2021-04-14
Campaign Signup via Redux Cleanup

mendelB:
### What's this PR do?

This pull request cleans out code left over from our new feature flagged Campaign Signup via GraphQL work in #2539 by formalizing the new GraphQL flow and removing deprecated code from the old signup via Phoenix->Rogue/NS flow.

commit-by-commit!

This shouldn't be affecting anything live as since #2539 we've been using GraphQL and avoiding all of this code!

This isn't an exhaustive cleanse of the Redux related signup logic, just the things that are deprecated as part of the switch to signing up via GraphQL in the `CampaignSignupForm`.

### Relevant tickets

References [Pivotal #175291206](https://www.pivotaltracker.com/story/show/175291206).

### Checklist


resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
```diff
@@ -180,33 +182,19 @@ const CampaignSignupForm = props => {
       },
     });
 
+    sixpack().convertOnAction('signup');
```
lindsayrmaher, line 185: are we restarting an existing sixpack test? just wondering why we needed to add this in.

lindsayrmaher (approved):
this makes sense! I have a question regarding the sixpack test, but nothing else is standing out that seems concerning

resources/assets/components/CampaignSignupForm/CampaignSignupForm.js
mendelB: I added this as backward compatibility for what the conversions we did automatically in the old Redux signup function

https://github.com/DoSomething/phoenix-next/blob/1e27e09c67ac642d2c701c7c7bebe5b5782445a5/resources/assets/actions/signup.js#L106

There are no active tests that I'm aware of, but I think this serves to automatically trigger conversions for tests configured via [Contentful](https://app.contentful.com/spaces/81iqaqpfd8fy/content_types/sixpackExperiment/fields) with the `signup` option selected (though, I don't think we've run any of those for a while either 🤷 )

PR #2614, opened 2021-04-15
About Page Analytics

weerd:
### What's this PR do?

This pull request adds analytics tracking for the About Page.

👀  it's all pretty straight-forward

I also removed the HelloBar script from the main blade layout file because we no longer use the service, and the script was annoyingly throwing warnings in the console while I was debugging. How dare it. 👹

### Relevant tickets

References [Pivotal #177468614](https://www.pivotaltracker.com/story/show/177468614).
References [Pivotal #177494823](https://www.pivotaltracker.com/story/show/177494823)

### Checklist

### DIFF
--- a/resources/assets/components/pages/AboutPage/AboutPage.js
+++ b/resources/assets/components/pages/AboutPage/AboutPage.js
@@ -23,8 +23,11 @@ const AboutPageTemplate = () => (
     <main>
       <article data-test="about-page">
         {/* Page Header */}
-        <header role="banner" className="bg-white">
-          <AnalyticsWaypoint name="RENAME_section_top" />
+        <header role="banner" className="bg-white relative">
+          <AnalyticsWaypoint
+            className="absolute top-0 w-full"
+            name="page_header_section_top"
+          />
 
           <div
             className="bg-gray-400"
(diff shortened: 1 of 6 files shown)


resources/views/layouts/master.blade.php
```diff
@@ -52,7 +52,6 @@
     @endisset
 
     @stack('scripts')
-    <script src="https://my.hellobar.com/f70465b5f088ff5df39e838a358d27109b365641.js" type="text/javascript" charset="utf-8" async="async"></script>
```
weerd, line 55: 👋 🗑️

resources/assets/components/utilities/AnalyticsWaypoint/AnalyticsWaypoint.js
```diff
@@ -19,6 +21,7 @@ const AnalyticsWaypoint = ({ className, context, name }) => {
         category: EVENT_CATEGORIES.waypoint,
         label: name,
         context: {
+          url: window.location.href,
```
weerd, line 24: Adding the page URL so we have a better idea what page the analytics waypoint event fired on!

resources/assets/components/pages/AboutPage/AboutPage.js
```diff
@@ -23,8 +23,11 @@ const AboutPageTemplate = () => (
     <main>
       <article data-test="about-page">
         {/* Page Header */}
-        <header role="banner" className="bg-white">
-          <AnalyticsWaypoint name="RENAME_section_top" />
+        <header role="banner" className="bg-white relative">
+          <AnalyticsWaypoint
+            className="absolute top-0 w-full"
+            name="page_header_section_top"
+          />
```
weerd, line 30: So this is a pretty cool technique (possible after [adding support for `className`](https://github.com/DoSomething/phoenix-next/pull/2584/files#diff-1ab201a832b42d1f89e19430bcf5aac542843d74a8639822ba428a4735236fd1) to the `AnalyticsWaypoint` component.

It provides a better solution for dealing with waypoints when added inside an element that is acting as a grid. This allows the element to not be affected by the grid and actually align to either the top or bottom of the section it resides in. Prior to this approach, the waypoint divs would get mixed in as columns within the grid and there were scenarios where both top and bottom waypoints would fire at the same time since they existed as columns that spanned the entire height of the section! 😱

DFurnes: Smart!!

DFurnes (approved):
📊

PR #2617, opened 2021-04-15
Homepage Member Benefits Update

weerd:
### What's this PR do?

This pull request updates the Homepage and switches out the content in the Newsletter section and replaces it with content for Member Benefits!

👀 

https://user-images.githubusercontent.com/105849/114947456-31b23400-9e1b-11eb-9b1c-4da809a9bff9.mp4

I took this opportunity to add comments as sections headers to better navigate the long file of code. Also updated all the existing waypoints to utilize the same utility classes trick, positioning them `absolute`-ly, `relative` to the parent `section` container. Good times 🕺 

With the new pages currently in production to replace the old Instapage pages, most of the newsletter links in the Newsletter section on the homepage no longer apply. Thus this update replaces that old content with new content in that section and links to the new pages.

### Relevant tickets

References [Pivotal #177512175](https://www.pivotaltracker.com/story/show/177512175).

### Checklist


resources/assets/components/pages/HomePage/HomePage.js
```diff
@@ -152,8 +152,12 @@ const HomePageTemplate = ({
...
       <main>
         <article data-test="home-page">
-          <header role="banner" className="bg-white">
-            <AnalyticsWaypoint name="impact_section_top" />
+          {/* Page Header */}
+          <header role="banner" className="bg-white relative">
+            <AnalyticsWaypoint
+              className="absolute top-0 w-full"
+              name="impact_section_top"
+            />
```
DFurnes, line 160: Non-blocking: it could be handy to make an `align="top"` or boolean `top` prop that applies these classes for us, so we don't need to remember which combination to use, e.g:

```jsx
<AnalyticsWaypoint name="impact_section_top" align="top" />
```

resources/assets/components/pages/HomePage/HomePage.js
```diff
@@ -285,85 +296,117 @@ const HomePageTemplate = ({
...
+                    &:hover {
+                      text-decoration-color: ${tailwind('colors.teal.500')};
+                    }
+                  `}
+                  href="/authorize"
+                >
+                  create an account
+                </a>{' '}
+                to officially join our community and gain access to these
+                benefits.
```
weerd, line 349: This is a little odd, formatting-wise, but it all seems to work fine. It's a bit of a custom style for links on purple background. Could be we see this more elsewhere and maybe turn it into a link component of sorts so it makes it more readable and there's less potential from deviation in customization.

resources/assets/components/pages/HomePage/HomePage.js
```diff
@@ -285,85 +296,117 @@ const HomePageTemplate = ({
...
+                >
+                  campaign
+                </a>{' '}
+                or{' '}
+                <a
+                  className="font-normal text-white hover:text-white underline"
+                  css={css`
+                    &:hover {
+                      text-decoration-color: ${tailwind('colors.teal.500')};
+                    }
```
DFurnes, line 342: I wonder if we should make a [plugin](https://tailwindcss.com/docs/adding-new-utilities#using-a-plugin) to allow us to write this as `hover:text-decoration-color-teal-500`?

DFurnes: Also wondering if we want to make a class or component for this, since it seems like a common pattern?

DFurnes (approved):
Looks good!!

resources/assets/components/pages/HomePage/HomePage.js
weerd: oh! Love this idea! 💡 I'll pop a new PR in with this suggestion!

resources/assets/components/pages/HomePage/HomePage.js
weerd: Ya, [same thought](https://github.com/DoSomething/phoenix-next/pull/2617#discussion_r615051124)!

resources/assets/components/pages/HomePage/HomePage.js
```diff
@@ -285,85 +296,117 @@ const HomePageTemplate = ({
...
+                  <BenefitItem
+                    content="Find news, inspiring stories, and how-to guides for impacting your community."
+                    title="Content"
+                    image={MemberBenefitsImages.Content}
                     link={{
-                      copy: 'Get Inspired',
-                      url:
-                        'https://join.dosomething.org/#page-block-85qyo7ii8fh',
+                      copy: 'Read More',
+                      url: 'https://join.dosomething.org',
```
weerd, line 396: I'll update this once the Article page is launched!

PR #2633, opened 2021-04-23
Install & use phpredis extension.

DFurnes:
### What's this PR do?

This pull request installs the [Redis PHP extension](https://github.com/phpredis/phpredis). I'm thinking this may have better out-of-the-box TLS support (as the Predis extension seems to require providing your own certificate according to their documentation...). It also has better performance & is the "default" recommended in [laravel/laravel](https://github.com/laravel/laravel/blob/a6ffdbdf416d60c38443725807a260a84dca5045/config/database.php#L122) (and already used in Northstar).

I'll try deploying this to QA & upgrading that application's Redis add-on to a Premium-0 add-on (with TLS required).

I'm hoping this will address [some Redis issues](https://dosomething.slack.com/archives/C02BBP0CU/p1619121297014800) we experienced yesterday afternoon.

### Relevant tickets

References [Pivotal #177823411](https://www.pivotaltracker.com/story/show/177823411).

### Checklist

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -133,15 +133,21 @@ return [
 
     'redis' => [
 
-        'client' => env('REDIS_CLIENT', 'phpredis'),
+        'client' => 'phpredis',
 
         'options' => [
             'cluster' => env('REDIS_CLUSTER', 'redis'),
             'prefix' => env('REDIS_PREFIX', Str::slug(env('APP_NAME', 'laravel'), '_').'_database_'),
+
+            // We use Heroku Redis, which uses a self-signed certificate & thus cannot be verified
+            // by OpenSSL. Thus, we disable certificate validation per these instructions:
+            // <https://help.heroku.com/HC0F8CUS/redis-connection-issues>
+            'context' => [
+                'stream' => ['verify_peer' => false, 'verify_peer_name' => false],
+            ],
         ],
 
         'default' => [
-            'scheme' => env('REDIS_SCHEME', 'tls'),
             'url' => env('REDIS_URL'),
             'host' => env('REDIS_HOST', '127.0.0.1'),
             'password' => env('REDIS_PASSWORD', null),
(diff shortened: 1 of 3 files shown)


config/database.php
```diff
@@ -133,11 +133,18 @@
 
     'redis' => [
 
-        'client' => env('REDIS_CLIENT', 'predis'),
+        'client' => 'phpredis',
```
DFurnes, line 136: I decided to hard-code this, as we don't want to use different drivers in different environments.

config/database.php
```diff
@@ -133,11 +133,18 @@
...
         'options' => [
             'cluster' => env('REDIS_CLUSTER', 'redis'),
             'prefix' => env('REDIS_PREFIX', Str::slug(env('APP_NAME', 'laravel'), '_').'_database_'),
+
+            // We use Heroku Redis, which uses a self-signed certificate & thus cannot be verified
+            // by OpenSSL. Thus, we disable certificate validation per these instructions:
+            // <https://help.heroku.com/HC0F8CUS/redis-connection-issues>
+            'context' => [
+                'stream' => ['verify_peer' => false, 'verify_peer_name' => false],
+            ],
```
DFurnes, line 147: Jeez, this was awfully tricky – and not helped by the fact that the [referenced help article](https://help.heroku.com/HC0F8CUS/redis-connection-issues) links to PHP instructions that (unlike all the other language examples) don't actually disable SSL verification! This is also the only source I could find that actually explains _why_ all those other code samples disable SSL verification, since the reason that's necessary isn't ever actually spelled out in their [official docs](https://devcenter.heroku.com/articles/heroku-redis).

I was able to piece together how to pass these [SSL context options](https://www.php.net/manual/en/context.ssl.php) to phpredis (via the `context` option [shown here](https://laravel.com/docs/8.x/redis#phpredis)). This allowed me to successfully connect to a Heroku Redis `premium-0` add-on from my local machine.

DFurnes:
Tested on qa.dosomething.org & confirmed that it is now working with non-hobby Redis addons! Phew!

config/database.php
mendelB: Whoah 🕵️

mendelB (approved):
👏 🙏

PR #2636, opened 2021-04-29
Updating Homepage link to new Articles page & more!

weerd:
### What's this PR do?

This pull request updates a link on the Homepage so it points to the new `/us/articles` page instead of the now defunct Instapage page.

Also snuck in a quick update to the About page cover image to use a new one that was updated.

quick 👀


DFurnes (approved):
💯

PR #2652, opened 2021-05-14
Site Navigation Login link dropdown menu

mendelB:
### What's this PR do?

This pull request adds the profile dropdown menu when hovering over the login link (unauthed users) on large screens

👁️ 

commit-by-commit would allow spotting the newer additions a bit easier per the refactor in https://github.com/DoSomething/phoenix-next/pull/2652/commits/cf22b0fd57fb3b7c8d0bf7b182d341b4cfd54ac2

A little bit of finessing was necessary to add the right-hand border when the dropdown is activated without nudging the whole element to the left. 

It also arguably felt like a good idea to extract the menu dropdown itself with the subtle differences between auth/non-auth state toggled by the `isAuthenticated()` boolean.

### Relevant tickets

References [Pivotal #177933072](https://www.pivotaltracker.com/story/show/177933072).

### Checklist

image

image

### DIFF
--- a/cypress/integration/site-navigation.js
+++ b/cypress/integration/site-navigation.js
@@ -148,11 +148,17 @@ describe('Site Navigation', () => {
       cy.get('.main-subnav').should('not.exist');
     });
 
-    it('does not render the profile dropdown when hovering over the profile icon', () => {
+    it('does not render the profile dropdown when hovering over the profile icon or login link', () => {
       const user = userFactory();
 
       // Set the viewport:
       cy.viewport(size.width, size.height);
+
+      cy.visit('/us/facts/test-11-facts-about-testing');
+
+      cy.findByTestId('login-nav').trigger('mouseover');
+
+      cy.findByTestId('profile-dropdown').should('have.length', 0);
 
       // Log the user in.
       cy.login(user)
@@ -214,6 +220,25 @@ describe('Site Navigation', () => {
       });
 
       cy.findByTestId('account-profile-nav').trigger('mouseout');
+
+      cy.findByTestId('profile-dropdown').should('have.length', 0);
+    });
+
+    it.only('toggles the profile dropdown when hovering over the login icon', () => {
+      const user = userFactory();
+
+      // Set the viewport:
+      cy.viewport(size.width, size.height);
+
+      cy.visit('/us/facts/test-11-facts-about-testing');
+
+      cy.findByTestId('login-nav').trigger('mouseover');
+
+      cy.findByTestId('profile-dropdown').within(() => {
+        cy.findAllByTestId('profile-dropdown-link').should('have.length', 8);
+      });
+
+      cy.findByTestId('login-nav').trigger('mouseout');
 
       cy.findByTestId('profile-dropdown').should('have.length', 0);
     });
--- a/cypress/integration/site-navigation.js
+++ b/cypress/integration/site-navigation.js
@@ -224,7 +224,7 @@ describe('Site Navigation', () => {
       cy.findByTestId('profile-dropdown').should('have.length', 0);
     });
 
-    it.only('toggles the profile dropdown when hovering over the login icon', () => {
+    it.only('toggles the profile dropdown when hovering over the login link', () => {
       const user = userFactory();
 
       // Set the viewport:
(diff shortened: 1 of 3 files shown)


weerd:
> A little bit of finessing was necessary to add the right-hand border when the dropdown is activated without nudging the whole element to the left.

Still eyeing, but one approach I've used in the past is to give the element that is activated on hover (the login link) a transparent or white border with the width and style defined. Then you just change the color on hover to the gray, but it doesn't end up with wonky nudging on hover since the border technically is there and accounted for the whole time; just sneakily hidden.

weerd:
image

@mendelB at approx the width indicated in the screenshot the interface items are a bit wonky and break. Not sure if a huge deal since it's a small window of width size where it occurs. Just noting in case it needs to be addressed!

weerd:
Screen Shot 2021-05-17 at 2 15 40 PM

It looks like the `<a>` link doesn't span across the full width of the submenu. To help with UX it's best to make the subnav item span the full width and add padding around it, but no "non-hoverable" whitespace around the item (if that makes sense).

mendelB:
> > A little bit of finessing was necessary to add the right-hand border when the dropdown is activated without nudging the whole element to the left.
> 
> Still eyeing, but one approach I've used in the past is to give the element that is activated on hover (the login link) a transparent or white border with the width and style defined. Then you just change the color on hover to the gray, but it doesn't end up with wonky nudging on hover since the border technically is there and accounted for the whole time; just sneakily hidden.

@weerd ooooh interesting! 💡 I think that's more elegant than a negative margin. I'll give it a go!

mendelB:
> > > A little bit of finessing was necessary to add the right-hand border when the dropdown is activated without nudging the whole element to the left.
> > 
> > 
> > Still eyeing, but one approach I've used in the past is to give the element that is activated on hover (the login link) a transparent or white border with the width and style defined. Then you just change the color on hover to the gray, but it doesn't end up with wonky nudging on hover since the border technically is there and accounted for the whole time; just sneakily hidden.
> 
> @weerd ooooh interesting! 💡 I think that's more elegant than a negative margin. I'll give it a go!

This worked like a dream in https://github.com/DoSomething/phoenix-next/pull/2652/commits/3341931a78ce41a98dd9c69a82716bf695c7c854 🙌 😌

mendelB:
> Screen Shot 2021-05-17 at 2 15 40 PM
> 
> It looks like the `<a>` link doesn't span across the full width of the submenu. To help with UX it's best to make the subnav item span the full width and add padding around it, but no "non-hoverable" whitespace around the item (if that makes sense).
> 
> image
> Switch this left & right padding on the `<ul>` to be on the `<a>` instead and thus you'll get the full width effect 😄

👏  This is a great note! Updated in https://github.com/DoSomething/phoenix-next/pull/2652/commits/0c9cd00c38b203f5f084f3842e3cb99e2b1f47a0

mendelB:
> image
> 
> @mendelB at approx the width indicated in the screenshot the interface items are a bit wonky and break. Not sure if a huge deal since it's a small window of width size where it occurs. Just noting in case it needs to be addressed!

I used `whitespace: nowrap` to prevent the text from wrapping to the newline and breaking the UI in https://github.com/DoSomething/phoenix-next/pull/2652/commits/5ab02c020089160aee4b7ba3fad6c206ad322a61

cypress/integration/site-navigation.js
```diff
@@ -217,5 +223,24 @@ describe('Site Navigation', () => {
 
       cy.findByTestId('profile-dropdown').should('have.length', 0);
     });
+
+    it.only('toggles the profile dropdown when hovering over the login icon', () => {
```
weerd, line 227: The wording on this description feels a bit confusing... sounds like the dropdown only shows up when hovering on the carat icon but it's over the entire thing (text + icon).

weerd (approved):
👍  Looking good to me!

Just a quick comment on maybe rephrasing a test description that sounded slightly confusing.

cypress/integration/site-navigation.js
mendelB: I updated it to read `"login link"` in https://github.com/DoSomething/phoenix-next/pull/2652/commits/571c1692c88dfaa2459dac4250ebb9bd7aff60d9 (My thinking is this would intuitively imply the caret icon as well 🙏).

PR #2653, opened 2021-05-17
Progress Bar Add Impact Goal and Current Quantity

lindsayrmaher:
### What's this PR do?

This pull request updates the progress bar displayed on a campaign to:
- conditionally display for all campaigns with photo action reportback vs just one campaign
- pull the noun and verb from the primary action
- pull the impact goal and current impact quantity from the primary action
- always show the goal out of the nearest thousand unless there is an impact goal provided and we haven't surpassed it.

It also:
- Updates where the actions are queried for so that the action id is available for both the campaign info block and the progress bar.
- updates the campaign info block cypress tests to relfect this ^^ updates
- adds cypress tests for all 3 test cases for displaying the progress bar
- updates the graphql schema file with new fields and a couple of deprecated fields

These examples below show when an impact goal is set vs when it is not. The 2,000 is automatically set based on the current impact quantity. I think a review app would also be helpful here so I can add a link to that.

Screen Shot 2021-05-17 at 11 44 46 AM

Screen Shot 2021-05-17 at 11 44 07 AM

Now that we are going to display this for more than one campaign, we want to give editors the ability to set the parameters for the progress bar while setting some fallbacks if the info isn't provided. 

### Relevant tickets

References [Pivotal #177793716](https://www.pivotaltracker.com/story/show/177793716).

### Checklist

### DIFF
--- a/cypress/integration/campaign-progress-bar.js
+++ b/cypress/integration/campaign-progress-bar.js
@@ -1,6 +1,5 @@
 /// <reference types="Cypress" />
 import exampleCampaign from '../fixtures/contentful/exampleCampaign';
-// import { userFactory } from '../fixtures/user';
 
 describe('Campaign Progress Bar', () => {
   const scholarshipAndReportbackAction = {
@@ -46,12 +45,8 @@ describe('Campaign Progress Bar', () => {
     cy.anonVisitCampaign(exampleCampaign);
 
     cy.findByTestId('campaign-progress-bar-container').within(() => {
-      cy.contains(
-        `${scholarshipAndReportbackAction.currentImpactQuantity.toLocaleString()} ${
-          scholarshipAndReportbackAction.noun
-        } ${scholarshipAndReportbackAction.verb}.`,
-      );
-      cy.contains(`Help us get to 3,000!`);
+      cy.contains('1,200 petitions signed.');
+      cy.contains('Help us get to 3,000!');
     });
   });
 
@@ -74,11 +69,7 @@ describe('Campaign Progress Bar', () => {
     cy.anonVisitCampaign(exampleCampaign);
 
     cy.findByTestId('campaign-progress-bar-container').within(() => {
-      cy.contains(
-        `${scholarshipAndReportbackAction.currentImpactQuantity.toLocaleString()} ${
-          scholarshipAndReportbackAction.noun
-        } ${scholarshipAndReportbackAction.verb}.`,
-      );
+      cy.contains('1,200 petitions signed.');
       cy.contains(`Help us get to 2,000!`);
     });
   });
@@ -102,11 +93,7 @@ describe('Campaign Progress Bar', () => {
     cy.anonVisitCampaign(exampleCampaign);
 
     cy.findByTestId('campaign-progress-bar-container').within(() => {
-      cy.contains(
-        `${scholarshipAndReportbackAction.currentImpactQuantity.toLocaleString()} ${
-          scholarshipAndReportbackAction.noun
-        } ${scholarshipAndReportbackAction.verb}.`,
-      );
+      cy.contains('1,200 petitions signed.');
       cy.contains(`Help us get to 2,000!`);
     });
   });
--- a/cypress/integration/campaign-progress-bar.js
+++ b/cypress/integration/campaign-progress-bar.js
@@ -14,6 +14,17 @@ describe('Campaign Progress Bar', () => {
     currentImpactQuantity: 1200,
   };
 
+  const textPostTypeAction = {
+    actionLabel: 'Sign a Petition',
+    postType: 'text',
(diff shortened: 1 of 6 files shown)


cypress/integration/campaign-progress-bar.js
```diff
@@ -0,0 +1,134 @@
+/// <reference types="Cypress" />
+import exampleCampaign from '../fixtures/contentful/exampleCampaign';
+// import { userFactory } from '../fixtures/user';
```
DFurnes, line 3: 👾

cypress/integration/campaign-progress-bar.js
```diff
@@ -0,0 +1,134 @@
...
+      }),
+    });
+
+    cy.anonVisitCampaign(exampleCampaign);
+
+    cy.findByTestId('campaign-progress-bar-container').within(() => {
+      cy.contains(
+        `${scholarshipAndReportbackAction.currentImpactQuantity.toLocaleString()} ${
+          scholarshipAndReportbackAction.noun
+        } ${scholarshipAndReportbackAction.verb}.`,
```
DFurnes, line 80: I think I'd recommend hard-coding strings in these assertions. That makes it a bit easier to skim what this test is actually asserting against and reduces the chance that something going wrong with the mock object could cause us to have a false positive (e.g. still accepting "undefined petitions signed").

resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js
```diff
@@ -15,200 +14,171 @@ import {
...
+                    <button
+                      data-testid="campaign-info-block-scholarship-details-link"
+                      className="font-normal underline text-blurple-500 hover:text-blurple-300 pb-4"
+                      type="button"
+                      css={css`
+                        :hover {
+                           {
+                            text-decoration-color: ${tailwind(
+                              'colors.teal.500',
+                            )};
```
DFurnes, line 80: Is this for our new link styles (the blurple text with the blue underline)? If so, I think we should make a component that we can reuse for all these rather than copy-pasting the same Tailwind classes.

DFurnes (approved):
🌡️

resources/assets/components/utilities/CampaignProgressBar/CampaignProgressBar.js
```diff
@@ -0,0 +1,80 @@
...
+`;
+
+const CampaignProgressBar = ({ actionId }) => {
+  const { loading, error, data } = useQuery(CAMPAIGN_PROGRESS_BAR_QUERY, {
+    variables: { id: actionId },
+  });
+
+  const progressGoalCalculator = (currentTotal, overallGoal = 0) => {
+    if (!overallGoal || currentTotal >= overallGoal) {
+      return Math.ceil(currentTotal / 1000) * 1000;
```
katiecrane, line 31: I want to walk myself through an example here in order to review this part. Say current total is `1256`. Divide that by `1000` and get `1.256`. Take the ceiling of that and get `2`. Multiply that by `1000` to get `2000`. Perfect!

katiecrane: This is maybe too much of an edge case, BUT if the `currentTotal` is `0` we'll end up setting a goal of `0` (because `Math.ceil(0)` is `0`). Maybe we want to default to returning `1000` as the smallest possible goal if there isn't one set?

katiecrane (approved):
Looks great, so excited for this! I had one question about an edge case 🤓

resources/assets/components/utilities/CampaignProgressBar/CampaignProgressBar.js
lindsayrmaher: 🤔  good point @katiecrane so would you set one conditional for if there is no overallGoal && currentTotal is 0 and return `1000` and then keep this conditional for returning the Math.ceil()?

katiecrane: Yeah! I think you could do something like this: 
```suggestion
  const progressGoalCalculator = (currentTotal, overallGoal = 0) => {
    if (!overallGoal || currentTotal >= overallGoal) {
      if (currentTotal < 1) {
        return 1000;
      }

      return Math.ceil(currentTotal / 1000) * 1000;
```

katiecrane: That one I think mightttt be worth adding in (not necessarily exactly how I wrote it there are many ways you could choose to write it!) because I could see `currentTotal` being `0` right when a campaign launches.

Another edge case is if `currentTotal` is a multiple of `1000` we'll get a goal that is the same as the current total. That one seems not worth worrying about for now.

PR #2656, opened 2021-05-20
Swap articles for donate in nav

katiecrane:
### What's this PR do?

This pull request removes articles from the nav and replaces it with a link to the donate page. The idea is that down the road some of the nav categories will be re-worked and articles will be re-introduced.

image

Analytics events upon click:
image

Analytics look good? Anything else I missed?

All above and in ticket!

### Relevant tickets

References [Pivotal #178201214](https://www.pivotaltracker.com/n/projects/2401401/stories/178201214).

### Checklist

### DIFF
--- a/resources/assets/components/SiteNavigation/SiteNavigation.js
+++ b/resources/assets/components/SiteNavigation/SiteNavigation.js
@@ -337,7 +337,7 @@ class SiteNavigation extends React.Component {
 
             <li className="menu-nav__item">
               <a
-                href="us/about/donate"
+                href="/us/about/donate"
                 onClick={() =>
                   this.handleOnClickLink({
                     name: 'clicked_nav_link_donate',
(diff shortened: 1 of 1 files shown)


resources/assets/components/SiteNavigation/SiteNavigation.js
```diff
@@ -337,17 +337,17 @@ class SiteNavigation extends React.Component {
 
             <li className="menu-nav__item">
               <a
-                href="/us/articles"
+                href="us/about/donate"
```
mendelB, line 340: Do we need to preface the forward-slash here?

mendelB (approved):
⚡

resources/assets/components/SiteNavigation/SiteNavigation.js
katiecrane: Oh good call - it still works as is but looks like we do usually start with the slash, I'll add it in!

PR #2662, opened 2021-05-26
Update Login Profile Dropdown Links

mendelB:
### What's this PR do?

This pull request updates and uses the `buildAuthRedirectUrl` helper method for the "Log In" profile dropdown links on the site navigation bar to redirect users to their account pages via the Northstar login page for authentication instead of registration.

- https://github.com/DoSomething/phoenix-next/commit/d47b16aaf7ab8e5c50357760243a13b86f767823 swaps the helper to use a single object parameter, this allows us to more flexibly add & use options.
- https://github.com/DoSomething/phoenix-next/commit/6b8bdc22ddb0607327d8339161c6e8822f3c1a23 & https://github.com/DoSomething/phoenix-next/commit/7d72256b9fbff650f7286cb7e8c942bd92051af4 add the new `destination` option to allow customizing the user's redirect post-authentication. (In our use case, we want to redirect users to a specific account tab after logging in.)
- https://github.com/DoSomething/phoenix-next/commit/3eab724e0da29e8acdc0e86a5d35fdda4caa5a1c adds some tests
- https://github.com/DoSomething/phoenix-next/commit/67e8ad84116bbc5b8a4c2a4f9ee8f151150a18b8 does the work to use this new `destination` param for the 'Login' profile dropdown

Previously, we made use of the 'Authentication gate' surrounding the account pages to simply link users to the respective account pages. Leah [pointed out](https://www.pivotaltracker.com/story/show/177933072/comments/224415285) in QA that this was redirecting users to _register_ instead of the desired UX to _login_. 

### Relevant tickets

References [Pivotal #177933072](https://www.pivotaltracker.com/story/show/177933072).

### Checklist

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -137,9 +137,9 @@ class AuthController extends Controller
         // "Log In" in the top navigation), or to the path defined in the $redirectTo property.
         $defaultIntended = is_same_domain(url()->previous()) ? url()->previous() : $this->redirectTo;
 
-        // The post-login redirect will be to the intended page (if logging in to view a page
-        // protected by the 'auth' middleware), or to the default path determined above.
-        $intended = session()->pull('url.intended', $defaultIntended);
+        // The post-login redirect will be to the custom 'destination', or the intended page (if logging in
+        // to view a page protected by the 'auth' middleware), or to the default path determined above.
+        $intended = Arr::get($queryParams, 'destination') ?: session()->pull('url.intended', $defaultIntended);
 
         session(['login.intended' => $intended]);
 
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -122,6 +122,21 @@ class AuthController extends Controller
     }
 
     /**
+     * Get custom destination URL (redirect for post authentication) from query params, and parse out
+     * the relative path to avoid melicious redirects.
+     *
+     * @param array $queryParams
+     * @return string|null
+     *
+     */
+     protected function getCustomDestination($queryParams = [])
+     {
+        $customDestination = Arr::get($queryParams, 'destination');
+
+        return $customDestination ? parse_url($customDestination, PHP_URL_PATH) : null;
+     }
+
+    /**
      * Set session data used post-login and once user returns, redirected back to intended page.
      *
      * @param array $queryParams
@@ -139,7 +154,7 @@ class AuthController extends Controller
 
         // The post-login redirect will be to the custom 'destination', or the intended page (if logging in
         // to view a page protected by the 'auth' middleware), or to the default path determined above.
-        $intended = Arr::get($queryParams, 'destination') ?: session()->pull('url.intended', $defaultIntended);
+        $intended = $this->getCustomDestination($queryParams) ?: session()->pull('url.intended', $defaultIntended);
 
         session(['login.intended' => $intended]);
 
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -133,7 +133,14 @@ class AuthController extends Controller
      {
         $customDestination = Arr::get($queryParams, 'destination');
 
-        return $customDestination ? parse_url($customDestination, PHP_URL_PATH) : null;
+        if (! $customDestination) {
+            return null;
+        }
+
+        $path = parse_url($customDestination, PHP_URL_PATH);
+        $query = parse_url($customDestination, PHP_URL_QUERY);
+
+        return $path . (!empty($query) ? '?' . $query : '');
      }
 
     /**
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -127,10 +127,9 @@ class AuthController extends Controller
      *
      * @param array $queryParams
      * @return string|null
-     *
      */
-     protected function getCustomDestination($queryParams = [])
-     {
+    protected function getCustomDestination($queryParams = [])
+    {
         $customDestination = Arr::get($queryParams, 'destination');
 
         if (! $customDestination) {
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/AuthController.php
```diff
@@ -137,9 +137,9 @@ protected function setSessionData($queryParams = [])
         // "Log In" in the top navigation), or to the path defined in the $redirectTo property.
         $defaultIntended = is_same_domain(url()->previous()) ? url()->previous() : $this->redirectTo;
 
-        // The post-login redirect will be to the intended page (if logging in to view a page
-        // protected by the 'auth' middleware), or to the default path determined above.
-        $intended = session()->pull('url.intended', $defaultIntended);
+        // The post-login redirect will be to the custom 'destination', or the intended page (if logging in
+        // to view a page protected by the 'auth' middleware), or to the default path determined above.
+        $intended = Arr::get($queryParams, 'destination') ?: session()->pull('url.intended', $defaultIntended);
```
DFurnes, line 142: Do we need to do any sanitization on this URL to make sure an attacker couldn't use it to send a user through our login flow to a malicious destination? Could we limit this `destination` parameter to local paths only?

resources/assets/helpers/auth.js
```diff
@@ -81,15 +81,17 @@ export function getDataForAuthRedirect() {
...
  *
- * @param  {Null|Object} options
- * @param  {Null|String} actionId
+ * @param  {Undefined|Object} options
+ * @param  {Undefined|String} actionId
+ * @param  {Undefined|String} destination
  * @return {String}
  */
-export function buildAuthRedirectUrl(options = null, actionId = null) {
+export function buildAuthRedirectUrl({ options, actionId, destination } = {}) {
```
DFurnes, line 89: Love this change to named parameters!

resources/assets/helpers/auth.test.js
```diff
@@ -0,0 +1,25 @@
+import { buildAuthRedirectUrl } from './auth';
+
+describe('buildAuthRedirectUrl', () => {
```
DFurnes, line 3: Yay tests!! 🚥

resources/assets/components/SiteNavigation/SiteNavigationProfile.js
```diff
@@ -75,14 +75,24 @@ const DropdownMenu = () => (
...
-          <DropdownLink href={`/us/account/${slug}`} copy={copy} />
+          <DropdownLink
+            href={
+              isAuthenticated()
+                ? `/us/account/${slug}`
+                : buildAuthRedirectUrl({
+                    options: { mode: 'login' },
+                    destination: `/us/account/${slug}`,
+                  })
+            }
```
DFurnes, line 86: Where do we handle the existing redirect for these pages (that's currently bringing them to the registration page)? I wonder if we could use this same helper to trigger the redirect there, so that if we had other links (say, from a help article) they'd also trigger the same redirect flow.

mendelB: It's currently relying on the surrounding [`AuthGate`](https://github.com/DoSomething/phoenix-next/blob/07c206f0cc60772ead55618392898c8162a3b665/resources/assets/components/App.js#L88-#L90) which automatically redirects to the registration page. I feel like _typically_ the registration page _is_ the intended flow?

app/Http/Controllers/AuthController.php
mendelB: I added a helper in https://github.com/DoSomething/phoenix-next/pull/2662/commits/f93d2abc9a0b88e8a0e3535526ab51276a1ed523 to simply parse out the relative path so that we're always safe (and avoid the headache of reasoning about URL sanitization too deeply for this 😅 🙏 )

mendelB: ...And added logic to account for query params in https://github.com/DoSomething/phoenix-next/pull/2662/commits/8cfe61f8a8e8c611fd0d3726dab77d287c0543d0

app/Http/Controllers/AuthController.php
```diff
@@ -121,6 +121,27 @@ protected function getUrl()
...
     }
 
+    /**
+     * Get custom destination URL (redirect for post authentication) from query params, and parse out
+     * the relative path to avoid melicious redirects.
+     *
+     * @param array $queryParams
+     * @return string|null
+     */
+    protected function getCustomDestination($queryParams = [])
```
DFurnes, line 131: 👍 👍

resources/assets/components/SiteNavigation/SiteNavigationProfile.js
DFurnes: Mmm yeah, I'm curious if instead of setting these auth redirects here (in the site nav), we could instead add a `mode` prop to that `AuthGate` that chooses whether whatever routes are within the `AuthGate` should redirect to the login or registration flow when visited? 

Then we'd still just have "simple" links here & any other ways a user could get to account pages would still trigger the more expected login flow as well.

DFurnes (approved):
Left an idea for how we might streamline this, but no need to block merging!

resources/assets/components/SiteNavigation/SiteNavigationProfile.js
mendelB: Ahhh, see my assumption was that in all other cases we'd _still_ want to trigger the registration flow when visiting these pages, aside from e.g. this scenario where we're explicitly intending to log in, but I agree that it feels pretty darn sensible to instead globally assert a login flow when attempting to visit an _account_ page. And adding such functionality to the `AuthGate` would drastically simplify this!

I'll go ahead and merge this in so that we can deploy this intended flow before the weekend and work on this update in a follow-up!

mendelB: Updated in https://github.com/DoSomething/phoenix-next/pull/2663

PR #2673, opened 2021-06-09
Remove margin from VC signature

katiecrane:
### What's this PR do?

This pull request removes the margin on the VC signature so it gets positioned closer to the name below it. I think there's something different about this image and it looks better with no margin.

With margin (old):
image

Without margin (new):
image

Any consequences of removing the margin that I'm missing?

All above!

### Relevant tickets

References [Pivotal #178386171](https://www.pivotaltracker.com/story/show/178386171).

### Checklist

### DIFF
--- a/resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
+++ b/resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
@@ -253,10 +253,7 @@ const CertificateTemplate = ({ certificatePost }) => {
           <View style={{ fontSize: 10, marginTop: 20, lineHeight: 1.1 }}>
             <View style={styles.flex}>
               <View style={{ width: '50%' }}>
-                <Image
-                  src={signature}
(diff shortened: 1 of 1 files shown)


resources/assets/components/pages/AccountPage/Credits/CertificateTemplate.js
```diff
@@ -253,10 +253,7 @@ const CertificateTemplate = ({ certificatePost }) => {
           <View style={{ fontSize: 10, marginTop: 20, lineHeight: 1.1 }}>
             <View style={styles.flex}>
               <View style={{ width: '50%' }}>
-                <Image
-                  src={signature}
-                  style={{ width: 100, marginVertical: -12 }}
-                />
+                <Image src={signature} style={{ width: 100 }} />
```
DFurnes, line 256: Feels like a win to no longer need a negative-margin hack here, nice!

DFurnes (approved):
🖊️

PR #2690, opened 2021-07-23
Fix Timeouts on User Campaigns Query

mendelB:
### What's this PR do?

While testing the User Campaigns tabs, I seemed to be getting a lot of timeouts when loading my campaigns (I've got _tons_ of test signups, etc. on my account). I figured that it might be a good idea to batch up these heavily nested queries so that we're only loading - I settled on 20 - at a time (which should be enough for most users anyhow...)

commit-by-commit could be helpful. 
Does this approach make sense?

I copied this logic over from the [`CampaignsTable`](https://github.com/DoSomething/northstar/blob/f9bcd6343abb26b0b915f8f8f2e1e3c4831c8b46/resources/assets/admin/components/CampaignsTable.js#L85-L107) in Northstar where we use this to load the full list of campaigns matching an input filter.

### Relevant tickets

References [Pivotal #178248537](https://www.pivotaltracker.com/story/show/178248537).

### Checklist

_(https://github.com/DoSomething/phoenix-next/commit/2d1d28f6f110bf605a773035018face249eb7ece) I added some cute lil 'dynamic' messages to display per batch to at least keep us testers entertained? #cleanupSprint_

Kapture 2021-07-23 at 11 17 52


DFurnes:
> I added some cute lil 'dynamic' messages to display per batch to at least keep us testers entertained? #cleanupSprint

OH MY GOD, I _LOVE_ THIS. 🤩

resources/assets/components/pages/AccountPage/Campaigns/UserCampaigns.js
```diff
@@ -96,130 +109,169 @@ const UserCampaignsGallery = ({ description, signups, emptyStateImg }) => {
...
+
+        <nav
+          className="base-12-grid page-navigation pt-3 md:pt-6 -no-fade"
+          css={css`
+            border-bottom-width: 2px;
+            // offset some assigned styles for the .page-navigation class:
+            background-color: ${tailwind('colors.gray.100')};
+            border-top: 0;
+            padding-left: 0;
+            padding-right: 0;
```
DFurnes, line 187: Are these defined as CSS here in order to avoid battling for specificity with `.page-navigation`?

resources/assets/components/pages/AccountPage/Campaigns/UserCampaigns.js
```diff
@@ -96,130 +109,169 @@ const UserCampaignsGallery = ({ description, signups, emptyStateImg }) => {
...
+          `}
+        >
+          <div
+            className="grid-wide md:col-start-1 nav-items -mx-3"
+            style={{ float: 'none' }}
```
DFurnes, line 192: Could we use Tailwind's `float-none` here (or is this another specificity war)?

DFurnes (approved):
Looks great! ⛵

resources/assets/components/pages/AccountPage/Campaigns/UserCampaigns.js
mendelB: Yeah unfortunately this was a specificity war (with the `navigation` hierarchy of style rules). I tried with Tailwind and even Emotion to no avail 😢

resources/assets/components/pages/AccountPage/Campaigns/UserCampaigns.js
mendelB: Yeah this was the same as https://github.com/DoSomething/phoenix-next/pull/2690#discussion_r675656078. TBH I was thinking it might be prudent to just try and tailwindify, or otherwise organize that system so we can replicate and customize, but this whole feature felt so big, I couldn't muster the creative energy 🙈

resources/assets/components/pages/AccountPage/Campaigns/UserCampaigns.js
DFurnes: Blah, yeah that makes sense. 👍

DFurnes: If you want to use Tailwind classes here, Twin's [`!modifier`](https://github.com/ben-rogerson/twin.macro#features) should probably work to allow these to override Forge styles. (We could also consider setting up Tailwind's [JIT mode](https://tailwindcss.com/docs/just-in-time-mode), which also adds support for an [important modifier](https://tailwindcss.com/docs/just-in-time-mode#built-in-important-modifier). I'm less fond of the approach they took with scanning directories for class usage, but it's a first party solution which is nice.)

mendelB: Ohhh snap, this is great thank you! Imma bookmark this to try and circle back later in cleanup sprint! I do love how simple the Twin modifier is

resources/assets/components/pages/AccountPage/Campaigns/UserCampaigns.js
DFurnes: Totally fair! This seems like a good trade-off to get this out the door to me!

PR #2702, opened 2021-08-30
Formalize New User Campaigns Experience

mendelB:
### What's this PR do?

This pull request deprecates the `DS_ENABLE_USER_CAMPAIGNS` feature flag to formalize the new experience.
(https://github.com/DoSomething/phoenix-next/commit/beb7d3bf728abbb43791e15702878fe04ebf91ec updates the user signups stub to ensure we only load one batch of signups in the tests (this was causing a little flakiness).)

👀 

This is in anticipation of the new nested submissions page which will be feature flagged. The new experience has been live for a while now!

There were a few tests specific to the post block gallery that still seemed relevant so we extract those into a new file in https://github.com/DoSomething/phoenix-next/commit/ab3cabb665e649bfda5deb3672150b8bcb9e20bc

### Relevant tickets

References [Pivotal #178248537](https://www.pivotaltracker.com/story/show/178248537).

### Checklist


cypress/integration/account-campaigns-tab.js
```diff
@@ -8,205 +8,118 @@ describe('User Account Campaigns Tab', () => {
...
 
-      cy.findByTestId('complete-campaigns-tab').click();
+  /** @test */
+  it('Hides the Expired tab if there are no expired campaigns', () => {
+    const user = userFactory();
 
-      cy.location().should(loc => {
-        expect(loc.pathname).to.equal('/us/account/campaigns/complete');
-      });
+    // Mark the expired campaign fixture as 'evergreen'.
```
katiecrane, line 92: I'm a little confused by what this means, aren't we essentially creating a scenario where the user isn't signed up for any expired campaigns?

cypress/integration/post-gallery.js
```diff
@@ -0,0 +1,85 @@
...
+      ],
+    });
+
+    cy.authVisitBlockPermalink(user, '123');
+
+    cy.findByTestId('post-author').should('have.length', 1);
+  });
+
+  /** @test */
+  it('Does not display user name for a non anonymous post', () => {
```
katiecrane, line 36: I think this should be anonymous!

```suggestion
  it('Does not display user name for an anonymous post', () => {
```

katiecrane: OH WAIT this is in the profile right? So we DO want to see the name?

cypress/integration/account-campaigns-tab.js
mendelB: That's the idea! We're testing that in such a scenario, the expired tab is hidden (per the design)

cypress/integration/post-gallery.js
mendelB: My understanding is that we're testing here that for anonymous posts, the username does _not_ display to ensure anonymity? (This would be in a post gallery which are rendered via a contentful block but I think this logic applies globally to all post cards cross site).

mendelB: Omg I read your suggestion as the actual code! Yes indeed, thank! I'll commit this in. (This was all a copy pasta from the original so must have had a typo.

cypress/integration/account-campaigns-tab.js
katiecrane: Ah I follow that! I think I'm confused about the actual comment "Mark the expired campaign fixture as 'evergreen'"  - I would have expected something that's evergreen to always show up? Am I thinking about this the wrong way?

cypress/integration/post-gallery.js
```diff
@@ -0,0 +1,85 @@
...
+          user: {
+            firstName: user.firstName,
+          },
+        },
+      ],
+    });
+
+    cy.authVisitBlockPermalink(user, '123');
+
+    cy.findByTestId('anonymous-post-author').should('have.length', 1);
```
katiecrane, line 60: Sorry I think this might be a copy/paste thing but wouldn't we expect no name to show up on an anonymous post?

cypress/integration/account-campaigns-tab.js
mendelB: Ah ok gotcha! I totally see how this comment is misleading 😖 -- I'll re-word it to clarify

mendelB: Updated in https://github.com/DoSomething/phoenix-next/pull/2702/commits/c1462f5c67b54d24462a96b9ae04ae70b3a79f18

cypress/integration/post-gallery.js
katiecrane: OH I see now that it says `'anonymous-post-author'` and not just `post-author` - I think this is fine I just mis-read!

katiecrane (approved):
Looking great!!

PR #2710, opened 2021-10-01
Flesh out Campaign Submissions Page - Primary Action

mendelB:
### What's this PR do?

This pull request first extracts the logic we use for finding a campaign's "primary" action, then re-uses it for the new account submissions page to calculate this primary action and flesh out the campaign info card. It also cleans up some legacy code.

commit-by-commit!

- https://github.com/DoSomething/phoenix-next/commit/7925f3c60bddfd01c32588cb99165387751236a3, https://github.com/DoSomething/phoenix-next/commit/3fb51b5e83885c3fe98814f81d924e5940d91b65 we extract the logic - which is being essentially duplicated across two components - and add some tests. (The only difference was that in the `ScholarshipInfoBlock` we didn't bother looking for a fallback action if there was no action with both `scholarshipEntry` and `reportback`. But this component wouldn't be rendered in the first place without such an action!)
- https://github.com/DoSomething/phoenix-next/commit/98e0af8f4c6c23bf4cef02e00a14e71ead24b9cf While ensuring that we'd covered all use cases for the `ScholarshipInfoBlock` I noticed that this OVRD component contained outtaded logic to render this in a modal (the use case where this would be activated was deprecated in cleanup here https://github.com/DoSomething/phoenix-next/pull/2313)
- https://github.com/DoSomething/phoenix-next/commit/da75ac5df7b5061b5c37a1b0851cab21c446a166, https://github.com/DoSomething/phoenix-next/commit/3060b91d1f456f4ceb987f3bb3f8b7bdddc895cc, https://github.com/DoSomething/phoenix-next/commit/4844ea49f0f8840a595ea01b52949d0b106ff688 cover some simplification and cleanup of the `ScholarshipInfoBlock`. Most notably, we remove the query to load the campaign's actions. We're already running this query in the parent `CampaignBannner` so we can simply pass them along as a prop and avoid a duplicate query! Faster loading time for the modal and less code!
- Finally, we find the new campaign submission pages "primary" action in https://github.com/DoSomething/phoenix-next/commit/2ad1e5ca97f22172a12268a1a9cd68e91436b158, render some of the data for the info card in https://github.com/DoSomething/phoenix-next/commit/af49c7bbe6d2f26dd941372e857ccbe32621f037, and add an integration test in https://github.com/DoSomething/phoenix-next/commit/2332a2abde5b8f5833baea1aae2e4aa86904fdba

We still need to render one more data point covering the volunteer credit hours, and a link to download the certificate, which I'll cover in a follow-up PR.

### Relevant tickets

References [Pivotal #178932515](https://www.pivotaltracker.com/story/show/178932515).

### Checklist

Screen Shot 2021-10-01 at 16 15 09
Screen Shot 2021-10-01 at 16 15 04
Screen Shot 2021-10-01 at 16 14 59


weerd:
> Most notably, we remove the query to load the campaign's actions. We're already running this query in the parent CampaignBannner so we can simply pass them along as a prop and avoid a duplicate query! Faster loading time for the modal and less code!

Noice! 👏  Good find and cleanup!

resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js
```diff
@@ -104,11 +52,13 @@ const ScholarshipInfoBlock = ({
           <div className="float-left mr-4">
             <DoSomethingLogo className="h-16" />
           </div>
+
```
weerd, line 55: Sibling line breaks 😌

weerd (approved):
👏  Nice work; this looks great!

PR #2713, opened 2021-10-18
Account Submissions Page Posts Summary

mendelB:
### What's this PR do?

This pull request adds a summary/carousel of the user's posts to the account campaign submissions page.

commit-by-commit so you can easily review some extraction of post-card and volunteer credits logic for re-use in this new component.

We're still missing some of the custom views per specialized post type (questionnaires, soft-edge, petitions, displaying the contentful text prompt for text posts). But since we have a catch-all for these situations, and the vast majority of cases are photo / standard text -- this should put us in a good place to ship! And the rest we can build iteratively.

### Relevant tickets

References [Pivotal #178248559](https://www.pivotaltracker.com/story/show/178248559).

### Checklist

Kapture 2021-10-19 at 16 27 41

**Mobile:**

image

**Text post**
Screen Shot 2021-10-19 at 4 30 42 PM


resources/assets/helpers/post.js
```diff
@@ -0,0 +1,59 @@
...
+      )
+    : null;
+}
+
+/**
+ * Generates an author label for a post.
+ *
+ * @param  {Object}   post
+ * @param  {Boolean}  isAnonymous
+ * @return {String|null}
```
weerd, line 35: This always returns a `String` and never `null`, right? Seems like always will result in some string default.

resources/assets/components/artifacts/MenuCarat/MenuCarat.js
```diff
@@ -3,14 +3,28 @@ import PropTypes from 'prop-types';
...
+const directions = {
+  down: '0',
+  left: '90',
+  up: '180',
+  right: '270',
+};
+
+const MenuCarat = ({ className, color, height, direction, width, onClick }) => (
+  <div
+    className={classNames('menu-carat pl-2 my-auto', className)}
```
weerd, line 17: 🤔  will having a padding defined be ok? Depending on context used, would it be better for context to dictate that? I don't remember well off-hand all contexts. This also makes me wonder, now that it can be rotated, I'm pondering if the size of the SVG should be a square by default, otherwise, based on rotation, spacing around it might change? Not sure in practice though without digging in. Would be fine to do as a future update too.

weerd (approved):
This is looking great and exiting to have a carousel 🎠

resources/assets/components/pages/AccountPage/Submissions/PostSummary.js
```diff
@@ -0,0 +1,162 @@
...
+                currentIndex < posts.length - 1
+                  ? undefined
+                  : tailwind('colors.gray.400')
+              }
+            />
+          </div>
+        ) : null}
+      </div>
+
+      <div className="pl-4 py-3 sm:w-2/5">
```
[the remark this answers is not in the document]
mendelB: THANKS FOR POINTING THIS OUT! I totally overlooked the design here. I've updated things in https://github.com/DoSomething/phoenix-next/pull/2713/commits/938016ee13c6353c011bef570a798fe79224b015 and https://github.com/DoSomething/phoenix-next/pull/2713/commits/ea42e514e4edd68c39bca77713c8d3020f86c129 to align with the design -- fixed spacing for photo posts, and updated for all other posts as well.

resources/assets/components/artifacts/MenuCarat/MenuCarat.js
weerd: Sounds good. Mostly was noting for a future update :)

resources/assets/components/pages/AccountPage/Submissions/PostSummary.js
```diff
@@ -0,0 +1,162 @@
...
+  const authorLabel = getAuthorLabel(currentPost, isAnonymous);
+
+  return (
+    <Card className="sm:flex rounded bordered p-3">
+      <div className="sm:w-3/5">
+        <PostContent post={currentPost} authorLabel={authorLabel} />
+
+        {['photo', 'text'].includes(currentPost.type) && posts.length > 1 ? (
+          <div className="mt-5 mb-3 sm:mx-3 flex flex-wrap justify-center">
+            <MenuCarat
```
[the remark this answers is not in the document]
mendelB: Updated the naming and all other use cases in https://github.com/DoSomething/phoenix-next/pull/2713/commits/1a1ee93e6758a5f7281acd86d9b83fee95c9b4c6!

PR #2721, opened 2021-12-06
Featured campaign on thank you page #1 (feature flag, styling)

katiecrane:
### What's this PR do?

This pull request adds (behind a feature flag!) a hardcoded featured campaign on the thank you page and includes Cypress tests for those updates! Another update of note is that I added an optional prop to `<CampaignCardFeatured>` to allow the card to say something other than "Featured" - "Featured" is the default value for the optional prop though!

Check it out:
image

image

Looking good so far? Anything I should change to make the next step easier?

Next up will be allowed the featured campaign to set on a PhotoSubmissionAction in Contentful.

### Relevant tickets

References [Pivotal #177910660](https://www.pivotaltracker.com/story/show/177910660).

### Checklist


katiecrane:
circleci is failing on the "Report software versions" step which I know we added to help with debugging @weerd but I'm not sure why it's not working now...I pulled down `main` and rebased in case I was missing something but it still fails.

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
```diff
@@ -143,6 +162,35 @@ const ShowSubmissionPage = ({ match }) => {
...
+              // Display nothing if campaign to be featured is closed
+              if (!data.campaign.isOpen) {
+                return null;
+              }
+
+              return (
+                <div
+                  className="base-12-grid mb-16"
+                  data-testid="featured-campaign"
+                >
```
weerd, line 180: I think I need to better see the rendered markup, but a potential cleanup could be removing this `<div>` and placing this as a child of the `<div>` above that has the `base-12-grid` class already applied (again, _potentially_!).

resources/assets/components/utilities/CampaignCard/CampaignCardFeatured.js
```diff
@@ -104,10 +104,12 @@ const CampaignCardFeatured = ({ campaign, hasButton }) => {
 CampaignCardFeatured.propTypes = {
   campaign: propType(campaignCardFeaturedFragment).isRequired,
   hasButton: PropTypes.bool,
+  featuredText: PropTypes.string,
 };
 
 CampaignCardFeatured.defaultProps = {
   hasButton: false,
+  featuredText: 'Featured',
```
weerd, line 112: Maybe we call this `featureTitle` or `featureHeading` to better describe what it is and ideally a short string.

weerd (approved):
I think this is looking good! I'll take a gander at the CircleCI bug and see what's up!

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
katiecrane: Oooo are you referring to the `<div>` on line 82/101?

resources/assets/components/utilities/CampaignCard/CampaignCardFeatured.js
katiecrane: OOO yeah I like `featureHeading`! I'll update it to that, much clearer!

.circleci/config.yml
```diff
@@ -25,7 +25,7 @@ jobs:
             [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
             [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"
             nvm install v14
-            nvm alias default 14.18.1
+            nvm alias default 14.18
```
weerd, line 28: @katiecrane ok, I think I figured it out and looks like CircleCI is working a bit better now (I'll make the update on Northstar in my next PR as well since I foresee it breaking as well!).

Turns out when I was setting the alias I was being too specific, including the patch version. Seems like Node v14.18.x was recently updated (patched) and it went from `14.18.1` to `14.18.2` and my `alias` command was trying to find `14.18.1` which is no longer the latest, thus 💣 💥

katiecrane: AH that's tricky! I guess we have to stick to one decimal point only!

katiecrane: Thanks so much for looking into it, I didn't know where to start with it!

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
weerd: Yep yep. Maybe a future refactor/cleanup but I need to see the full rendered page to decide that a bit easier :)

.circleci/config.yml
weerd: It was tricky because the error was throwing in the follow-up command after the NVM/Node stuff, so a bit of a red herring! 🔴  🐟

resources/assets/components/utilities/CampaignCard/CampaignCardFeatured.js
katiecrane: updated in https://github.com/DoSomething/phoenix-next/pull/2721/commits/427187b3be8ac5a3333873b5231c025ad21f7072

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
katiecrane: Okay cool! We'll hold on this for now

PR #2722, opened 2022-01-12
Get the featured campaign pulled from Contentful displaying

katiecrane:
### What's this PR do?

This pull request builds out the new Featured Campaign on the thank you page! Updates made include:
- Corresponding update in GraphQL https://github.com/DoSomething/graphql/pull/349 (⚠️ need to make sure to deploy GraphQL FIRST) and corresponding `schema.json` updates
- Update to circleci config
- Added a new field on Contentful dev to allow adding a featured campaign on Photo Submission Blocks
- Additional Cypress tests
- Updated the `PhotoSubmissionBlockFragment`
- Refactored the `ShowSubmissionPage` `CONTENTFUL_BLOCK_QUERY` to be run in a `useQuery` hook so we can easily use the data we get back throughout the page
    - the query is skipped if we don't have a `submissionActionId`
- Update the featured campaign area to use the given `featuredCampaignId` instead of a hardcoded one

Contentful update:
image

image

Here's what it looks like:
image

Any missing Cypress tests? I'm also adding some questions in the code!

All in the ticket I think!

### Relevant tickets

References [Pivotal #177910660](https://www.pivotaltracker.com/story/show/177910660).

### Checklist


resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
```diff
@@ -47,6 +47,9 @@ export const PhotoSubmissionBlockFragment = gql`
     informationTitle
     informationContent
     affirmationContent
+    featuredCampaign {
+      campaignId
+    }
```
weerd, line 52: I'm not 100% sure this would work, but instead of in the `ShowSubmissionPage.js` querying the right action just to get the featured campaign id, if we potentially already have access to it here, could we just pass it into the redirect URL to the Submission Page as a param?

So essentially down on line 75:

```js
window.location = `/us/posts/${response.data.id}?submissionActionId=${nextProps.id}&featuredCampaign=${campaignId}`;
```

In the `ShowSubmissionPage` component it's already pulling out the `match` prop from React Router which will have any passed URL query params.

Not sure if this makes sense but something that popped into my head that could help simplify!
```

Comment (katiecrane) [reply resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js:52]: OOOOOO I didn't even think of doing it this way! Since the `submissionActionId` and the `featuredCampaignId` both live on the `PhotoSubmissionAction` I don't see why we shouldn't treat them the same way. Gonna try this and see how much it simplifies things!

Comment (weerd) [reply resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js:52]: Yais! Ok, sounds good. Keep me posted, cause I think it'll be a nice approach and help simplify and reduce the number of queries in the Submission Page component :)

File: resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js [hunk 2]
```diff
@@ -141,33 +163,35 @@ const ShowSubmissionPage = ({ match }) => {
...
-            variables={{ campaignId: 9001 }}
+            variables={{
+              campaignId: featuredCampaignId,
+            }}
           >
             {data => {
+              // Display nothing if we weren't able to grab the campaign
+              if (!data.campaign) {
+                return null;
+              }
```
katiecrane, line 194: Is there a cleaner way to handle these checks?

.circleci/config.yml
```diff
@@ -25,7 +25,7 @@ jobs:
             [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
             [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"
             nvm install v14
-            nvm alias default 14.18
+            nvm alias default 14
```
weerd, line 28: 🎈

contentful/content-types/photoSubmissionAction.js
```diff
@@ -137,132 +153,114 @@ module.exports = function(migration) {
...
-  );
+  photoSubmissionAction.changeEditorInterface('showQuantityField', 'boolean', {
+    helpText:
+      "Should the form ask for the quantity of items in member's photo submission?",
+    trueLabel: 'Yes',
+    falseLabel: 'No',
+  });
 
-  photoSubmissionAction.changeFieldControl(
+  photoSubmissionAction.changeEditorInterface(
```
weerd, line 177: I take Contentful changed their API method? This is from the generated content type migration right?

cypress/integration/show-submission-page.js
```diff
@@ -108,6 +130,37 @@ describe('Show Submission Page', () => {
...
+      featured_campaign_on_thank_you_page: true,
+    })
+      .login(user)
+      .visit('/us/posts/1?submissionActionId=abc');
+
+    // Featured campaign should NOT show on page
+    cy.findByTestId('featured-campaign').should('not.exist');
+
+    // Make sure whole page isn't broken!
+    cy.contains('We Got Your Submission');
```
weerd, line 161: 😏

weerd, line 161: 😏

contentful/content-types/photoSubmissionAction.js
katiecrane: Oh hmm yeah, I ran `contentful space generate migration` for this content type

resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js
katiecrane: Ended up not going this route since we need to pull other information from the `PhotoSubmissionAction` anyway!

weerd (approved):
Looks good to me! 👍

resources/assets/components/pages/ShowSubmissionPage/ShowSubmissionPage.js
katiecrane: That is cleaner!! I think I'll go with the first one because for me that's a little easier to follow at first glance

katiecrane:
@weerd I just added in some documentation updates/additions too if you want to check those out! Not mandatory though

PR #2729, opened 2022-02-16
NPM audit updates.

weerd:
### What's this PR do?

This pull request has package updates that occurred after running `npm audit fix`.

👀 

I approved one of the Dependabot updates for an NPM package, but saw that the CircleCI tests were failing. I think this is likely because when these PRs were created, they didn't include the update to the CircleCI config to not specify a minor version for Node `14.x`.

So instead of trying to fix each Dependabot PR, I figured best to run the audit of `main` (updated) branch, fix a few of the security issues and push it up as a single PR. The Dependabot PRs should close once they see the updates happened in this one.


katiecrane (approved):
smart move!

PR #2738, opened 2022-05-06
Word Cloud word processing and rendering

katiecrane:
### What's this PR do?

This pull request finishes fleshing out the Word Cloud feature! Screenshots down below!

If the feature flag is set to true AND the campaign id of the current campaign matches the one set, then the Word Cloud will display instead of the cover image.

Here is an overview of how it works:
1. We get the 100 most recent accepted posts for the given action id
2. We get the `text` from the posts and concatenate it all into one long string
3. The words get counted!
4. A little re-formatting happens
5. We filter out words only appearing once, words with only one letter, and words on our excluded word list
6. A little more re-formatting!
7. We render the Word Cloud with all our custom options!

Cypress tests and documentation included! Hopefully the documentation answers most questions about the feature.

Small:
Screen Shot 2022-05-06 at 2 57 34 PM

Medium:
Screen Shot 2022-05-06 at 2 57 46 PM

Large:
Screen Shot 2022-05-06 at 2 58 00 PM

Did I miss anything as a part of this feature? Any questions about the documentation?

You can check out the [technical spec](https://docs.google.com/document/d/1Pux8yqTx--Mwd_x8mBAhgB57HkJJhHdTTDzbdDNoZB0/edit) for further context!

### Relevant tickets

References [Pivotal #](https://www.pivotaltracker.com/story/show/181893750) & [Pivotal #](https://www.pivotaltracker.com/story/show/181893762)

### Checklist


resources/assets/components/utilities/WordCloud/WordCloud.js
```diff
@@ -0,0 +1,122 @@
...
+  };
+
+  const options = {
+    rotations: 2,
+    rotationAngles: [-90, 0],
+    colors: ['#3525AD', '#5887DA', '#A5C5ED'],
+    enableTooltip: false, // This has to be set to false otherwise tooltips will show up when hovering over the words
+    fontSizes: [12, 84], // If there is any weird overlapping happening, reduce the max size here
+    transitionDuration: 0,
+    fontFamily: 'helvetica', // Ideally we would use "montserrat" here
```
JasonAstuto-Amplify, line 94: It is mentioned that "montserrat" is the preffered fontFamily, but not why helvetica is being used. It would help to understand why choose this seemingly undesirable font.

JasonAstuto-Amplify (commented):
Overall great comments to explain the intent of the code where that is ambiguous. The changes appear to meet the stated goal of the project, and the code has no obvious bug, errors, or typos. Good Job!

resources/assets/components/utilities/WordCloud/WordCloud.js
katiecrane: Hi @JasonAstuto-Amplify, thanks for your help here! When I tried `'montserrat'` as the `fontFamily` it was NOT rendering montserrat but rather some sort of serif-having default font. Any idea why that might be?

I ended up with `'helvetica'` because it was the closest thing to `'montserrat'` that I could get to render as expected.

JasonAstuto-Amplify: It seems like the bold/pt weight is an issue for some, and with others, it is the way you import the custom google font. https://wordpress.org/support/topic/my-custom-font-doesnt-work/
https://stackoverflow.com/questions/54486873/my-montserrat-font-is-not-showing-the-correct-font-style

katiecrane: I did some digging and it turns out 'montserrat' is not actually one of the fonts we have installed! I definitely don't want to introduce a new typeface for this and we should use one we already have installed. We only have 2 installed, and Source Sans Pro looked closest to Montserrat to me, so I've updated to use that one!

katiecrane: Here's what it looks like with Source Sans Pro:
Screen Shot 2022-05-11 at 2 36 10 PM

JasonAstuto-Amplify: you could also reference the Monserrat FF directly from Google, and not install it.

katiecrane: Ooooo that's a good idea! It turns out Source Sans Pro was our designer's #2 choice anyway so gonna stick with this one for consistency!

katiecrane:
Now using the preferred cloud background image as well!

image

PR #2742, opened 2022-06-01
Adding fundraiseup donate cta

weerd:
### What's this PR do?

This pull request includes the new Fundraise Up script to utilize their donation button and UI/UX.

https://user-images.githubusercontent.com/105849/171491064-5edaa5fd-58b7-40a0-86ef-8969dae61fb3.mp4

The feature is currently limited to only showing up on the `/about/donate` page. It can be toggled on/off using the `DS_ENABLE_FUNDRAISEUP_DONATIONS` feature flag env variable.

👀 

In lieu of setting this up as a content type on Contentful, I decided it best to set it up as a component included in "CompanyPages", but based on the `/us/about/donate` slug. It is currently only used on the donate page, but if it is eventually used elsewhere, it may help to standardize it as a content type for editors to include on pages as needed.


dcrusoeds (approved):
Looks great. I've noted that Fundraising ID is hard-coded. In the future, maybe this gets componentized in something like Contentful (or elsewhere) and pushed as a changeable field so as to allow customization. All good for the meantime, though.

PR #2785, opened 2022-10-04
add e2e redirect test

FryderykDoSomething:
### What's this PR do?

This pull request adds an e2e test for redirects

Should work in CLI by itself

Once upon a time, there was a deployment wich crashed all the redirects :D


maciejziolkods (approved):
This need to be checked later whether is fired during deployment or only via CLI.

